Issue #2384, opened 2017-02-02
Solve Travis CI timeout for Clang


cbielow:
we always run into the time limit barrier.

multiple ways to resolve:
 - disable it (should be done in any case, until we have a new solution to see the green light CI tests again)
 - split the tests further (not ideal since its more build system hassle)
 - improve speed of building (compiler flags, caching of packages?)
 - use another CI

hroest:
basically it has to do with the fact that it is a debug-build with the address sanitizer. We have both a gcc build and a clang build in release mode and they both only take 30 minutes. So I think what we could do is to move the address sanitizer to cdash and only do a nightly test on it. I made a test build with clang-debug but without the address-sanitizer and lets see how well that does:

https://travis-ci.org/OpenMS/OpenMS/builds/213507338

hroest:
@jpfeuffer is there currently a nightly cdash that has the AdressSanitizer on?

hroest:
yes, pretty clear that the issue was the AdressSanitizer as this branch now runs pretty fast: https://github.com/OpenMS/OpenMS/commit/c08ed80e42c8f50f8e199585564a428b42045f96 and this is also what clang more or less specifies:
```
  Typical slowdown introduced by AddressSanitizer is 2x.
```
http://clang.llvm.org/docs/AddressSanitizer.html - I guess the slowdown is not just running but also building.

hroest:
well it is also a lot the tests:

```
Total Test time (real) = 154.76 sec
Total Test time (real) = 1373.97 sec
```

so basically 23 minutes vs 2.5 minutes to run all tests...

hroest:
fixed through #2567

PR #2476, opened 2017-03-23
Feature/de template

hroest:
A big push to de-template the mzML handler and the mzMLFile.

Overall, there were only 2 instances in the code base that actively used any other loading strategy than `MSExperiment<>`:

- src/topp/SpecLibSearcher.cpp
- src/utils/IDMassAccuracy.cpp

This simplifies the code base substantially, improves compile time and reduces compiled size for tools and utils:

before:
```
$ time make -j4 TOPP 
real    6m14.794s
user    22m36.968s
sys     1m15.672s
$ du -sh bin/
42M     bin/
```

after:
```
$ time make -j4 TOPP 
real    3m40.141s
user    12m55.000s
sys     1m1.736s
$ du -sh bin/
15M     bin/
```

so compile time is about half and compiled size is 2.8-fold reduced. The only downside is for the two tools mentioned above that try to load a RichPeakMap that there is a conversion cost as we load a PeakMap and then convert. Note that no data gets lost as the MzMLHandler anyways does not do anything with the RichPeak.


timosachsenberg:
nice. I think the PR by @enetz is also ready for review. I helped a bit in removing RichPeak* and there is not that much left that needs to be removed (but maybe in another PR). Without RichPeak* the way is free for removing quite a lot of templates in the core. One downside of using the DataArrays is that write/erase/swap... STL-like access to the vector of peaks in MSSpectrum is pretty unsafe if we have DataArrays and forget to manipulate them as well. Also, many of our filter algorithms that remove peaks didn't even bother about the data arrays. I converted a few of them but the are more.

hroest:
@timosachsenberg yes, I think that this will make quite a few things easier and RichPeak should be removed in the long run - I think the cases you describe are simply bugs and need to be fixed anyways since one cannot know whether a spectrum comes with DataArrays or not (they could be in the mzML file when reading). So I guess there is some work ahead ... fortunately this PR is pretty independent from that

hroest:
PS: I think total build time is now quite a bit lower (I only tested the new build) and the build uses much less space

```
$time make -j4 
real    14m34.621s
user    51m32.120s
sys     5m5.532s
$ du -sh 
106M    openms_release/src/tests/class_tests/bin/
537M    openms-contrib-all-old/src/tests/class_tests/bin/
```

hroest:
another nice feature: travis may run through now faster and this may also solve #2384

timosachsenberg:
+1
@cbielow @jpfeuffer @hendrikweisser any objections?

jpfeuffer:
No. +1 I like it!

hroest:
this is now superseded by #2477

PR #2479, opened 2017-03-24
[TEST] fix pyopenms test

hroest:


hroest:
merge as this is pyOpenMS only

PR #2480, opened 2017-03-24
[FEATURE] SQLite support for OpenSWATH

grosenberger:
This PR by @hroest and me provides support for SQLite-based file formats to OpenSWATH. Currently, the OpenSWATH workflow uses TraML, TSV and mzML to extract and transform data and to store results. Building on the standard formats works well for small-medium sized experiments, however with hundreds to thousands of runs, challenges in terms of storage and computation appear. These additions provide new data structures that hopefully alleviate this problems but maintain compatibility with the standards. Three new formats were implemented:

- PQP (peptide query parameter) is a structure consisting of different tables that represents TargetedExperiment/TraML. TargetedFileConverter (replacing both ConvertTraMLToTSV and ConvertTSVToTraML) converts between TSV, MRM, TraML and PQP files. PQP structures are more than 10x smaller than TraML and about 2x smaller than TSV. Importantly, they are easier to parse and access in downstream tools. PQP files could thus replace the TSV formats for downstream analysis while data should be continued to be shared by TraML.
- OSW (OpenSWATH report) can replace the TSV or featureXML reports generated by OpenSwathWorkflow and OpenSwathAnalyzer. They build on top of PQP and simply connect the detected and scored features with the peptide query parameters. The advantage is that all information (including PQP) is contained in the OSW files and this can be directly used by downstream algorithms like PyProphet, TRIC or TAPIR without the need for any other data.
- sqMass can replace/supplement chromatographic mzML, originating from the extracted data by OpenSwathWorkflow. When visualizing SWATH-MS data, TAPIR struggles from the need to use uncompressed mzML for random access. With hundreds to thousands of runs, a lot of space is required. sqMass files can be very easily connected to OSW reports and unnecessary data can be readily deleted. This should improve the scalability for future applications.

Importantly, these data structures are only thought for internal use by the workflows and to complement the PSI standards. They should not be used for data sharing, but rather to make the workflows more efficient.

The implementation is based on standard SQLite, please see also the corresponding `contrib` pull request. SQLite is a lightweight dependency and we could consider making it a dependency of OpenMS, since it is available by default in many distributions. Alternatively, Qt SQL provides a wrapped SQLite adapter. We could also consider building on top of that, but that might further blow up the dependencies.


src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
```diff
@@ -0,0 +1,981 @@
...
+      }
+
+      // Create SQL structure
+      char const *create_sql =
+
+        // data table
+        //  - compression is one of 0 = no, 1 = zlib, 2 = np-linear, 3 = np-slof, 4 = np-pic, 5 = np-linear + zlib, 6 = np-slof + zlib, 7 = np-pic + zlib
+        //  - data_type is one of 0 = mz, 1 = int, 2 = rt
+        //  - data contains the raw (blob) data for a single data array
+        "CREATE TABLE DATA(" \
```
timosachsenberg: is there a reason why both spectrum_id and chrom_id exist? Would having a single ID and a flag for chrom or spectrum be better?

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
```diff
@@ -0,0 +1,981 @@
...
+        "CHROMATOGRAM_ID INT," \
+        "CHARGE INT NULL," \
+        "ISOLATION_TARGET REAL NULL," \
+        "ISOLATION_LOWER REAL NULL," \
+        "ISOLATION_UPPER REAL NULL" \
+        ");" \
+
+        // precursor table
+        "CREATE TABLE PRECURSOR(" \
+        "SPECTRUM_ID INT," \
```
timosachsenberg: same here. is this for speed reasons?

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
```diff
@@ -0,0 +1,981 @@
...
+        "CHROMATOGRAM_ID INT," \
+        "CHARGE INT NULL," \
+        "PEPTIDE_SEQUENCE TEXT NULL," \
```
timosachsenberg: could this be more generic to also support small molecules?

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
```diff
@@ -0,0 +1,981 @@
...
+        {
+          // prevent writing of empty data which would throw an SQL exception
+          if (!data.empty())
+          {
+            prepare_statement.resize( prepare_statement.size() -1 ); // remove last ","
+            executeBlobBind_(db, prepare_statement, data);
+          }
+
+          data.clear();
+          prepare_statement = "INSERT INTO DATA(SPECTRUm_ID, DATA_TYPE, COMPRESSION, DATA) VALUES ";
```
timosachsenberg: SPECTRUm -> SPECTRUM

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
```diff
@@ -0,0 +1,981 @@
...
+
+        OpenMS::Precursor prec = chrom.getPrecursor();
+        // see src/openms/include/OpenMS/METADATA/Precursor.h for activation modes
+        int activation_method = -1;
+        if (!prec.getActivationMethods().empty() )
+        {
+          activation_method = *prec.getActivationMethods().begin();
+        }
+        String pepseq;
+        if (prec.metaValueExists("peptide_sequence"))
```
timosachsenberg: see above

timosachsenberg:
I think this is a very nice addition that might find use outside of OpenSWATH as well.
Especially visualization in KNIME or other downstream tools could benefit from reading a sqlite file.
+1 from my side.

timosachsenberg:
Chris and me discussed the level of integration and think that your approach is probably better then using the Qt interface.

hroest:
@timosachsenberg that is a good point, I think we could go two ways:

- have a single id for spectra and chromatograms, e.g. `entity_id`
- have separate ids for spectra and chromatograms

in my intuition, the first approach is somewhat more intuitive from a database building perspective, as it allows us to use the `PRIMARY KEY` directive. Consider the following with *separate ids*

``` 
INSERT INTO CHROMATOGRAM (ID) VALUES (1);
INSERT INTO CHROMATOGRAM (ID) VALUES (2);
INSERT INTO CHROMATOGRAM (ID) VALUES (3);
INSERT INTO SPECTRUM (ID) VALUES (1); // spectrum ids are different from chrom ids !
INSERT INTO SPECTRUM (ID) VALUES (2);
INSERT INTO SPECTRUM (ID) VALUES (3);
INSERT INTO SPECTRUM (ID) VALUES (3); // fails !
```

versus with *common ids*

``` 
INSERT INTO CHROMATOGRAM (ID) VALUES (1);
INSERT INTO CHROMATOGRAM (ID) VALUES (2);
INSERT INTO CHROMATOGRAM (ID) VALUES (3);
INSERT INTO SPECTRUM (ID) VALUES (1); 
// now we have a spectrum with id 1 and a chrom with id 1 but the DB doest know this and wont fail!
INSERT INTO SPECTRUM (ID) VALUES (2);
INSERT INTO SPECTRUM (ID) VALUES (3);
INSERT INTO SPECTRUM (ID) VALUES (3); // fails !
```

I think its just a DB design decision and something that prevents mistakes in general (and identifies bugs if some software tries to insert chroms/spectra with the same id).

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
hroest: see below, I dont really think that it would from a DB design perspective. Also it maps well to mzML ...

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
hroest: well again design reasons. Also it makes it clearer if there is a mistake and you can never join a spectrum with a precursor of a chromatogram and vice versa. I think it prevents many "dumb" mistakes

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
hroest: yes, sure! :-) I have not fully worked through what fields we would want but I think when I did the OpenSWATH transition for metabolomics I had `sum_formula` and `compound_name`. We would have to check whether that increases db size on disk if we have these fields and leave the empty but I think it wont affect it much...

hroest:
well this is clearly not working ... I cant really reproduce this on my own machine so I need to look into this more

hroest:
its possible that the version difference of sqlite screws this up -- precise has 3.7.9 which is quite dated by now (released late 2011)

https://launchpad.net/ubuntu/precise/+package/libsqlite3-dev

hroest:
yes, its pretty clear that the old sqlite version was the problem. I hacked together a version of OpenMS that now uses the newest 3.15 sqlite and this seems to work much better: https://github.com/OpenMS/OpenMS/tree/feature/sqlite2

hroest:
@grosenberger something is wrong with the clang builds:  TOPP_TargetedFileConverter_test_2_out1 fails only in the clang and not the gcc builds - any idea why ? http://cdash.openms.de/viewTest.php?onlyfailed&buildid=217339

hroest:
to check what is going on with regards to travis, please look at the branch which builds against the new sqlite library: https://github.com/OpenMS/OpenMS/branches/active

grosenberger:
@hroest I could not reproduce the problem yet on my machine, but will have a look at it ASAP.

hroest:
maybe it was just aeeb6c000f513c334af2479531b80f7d281ade35 ? lets see whether the branch runs through on travis

hroest:
yes, it seems that with aeeb6c0 we have a green travis!

https://github.com/OpenMS/OpenMS/commits/feature/sqlite2_hack

I just cherry-picked the commit so that its now also on this branch 5100ef2 - so we are now just waiting for https://github.com/OpenMS/contrib/pull/37 and the build system to be updated.

grosenberger:
Excellent, thank you @hroest!

hroest:
ok, I made the merge commit  and created the change in the build system to also build the new sqlite version on travis. If the nightlies are ready, we are good to merge

@jpfeuffer how are the nightly servers coming along, are they ready?

hroest:
@jpfeuffer @timosachsenberg are we ready for this? I would love to go ahead with the merge on this

timosachsenberg:
could you check why tests are failing on travis? Is the contrib not finished yet?

jpfeuffer:
Contrib should be ready for checking out freshly (e.g. on Travis).
I updated Mac and Linux nightlies. Win to come (hopefully today).

jpfeuffer:
Seems like gcc failed in the build quite early, while clang succeeded but timed out.

jpfeuffer:
http://cdash.openms.de/viewBuildError.php?buildid=220374
`/.../OpenMS/src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPReader.cpp:586:5: error: ‘cout’ is not a member of ‘std’
`
Maybe include iostream explicitly for g++?

timosachsenberg:
+1

pati-ni:
Trying to build after merging the develop branch and I'm getting:
```
/OpenMS/src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp: In member function ‘void OpenMS::Internal::MzMLSqliteHandler::executeSql_(sqlite3*, const stringstream&)’:
/home/main/projects/OpenMS/OpenMS/src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp:969:44: error: no match for ‘operator<<’ (operand types are ‘std::basic_ostream<char>’ and ‘const stringstream {aka const std::__cxx11::basic_stringstream<char>}’)
         std::cerr << "Prepared statement " << statement << std::endl;
```
I am using sqlite3-dev 3.16.2

hroest:
not sure where this is coming from, are you using a newer gcc and C++11 features?

pati-ni:
no I do not use c++11 features.
This happens also when I try to build the develop branch so I guess it is not something that I coded.

```
$ gcc --version
gcc (Debian 6.3.0-12) 6.3.0 20170406
```

jpfeuffer:
Seems like default mode in GCC6 is C++14. Did you explicitly specify -std=gnu++98?
But why does it break there and not everywhere else, too, then?

jpfeuffer:
You might be right, though, that the handling of the stream is not completely correct here (use .str() or .rdBuf()).
In this case I wonder why it worked in the nightlies and on Travis.

pati-ni:
@jpfeuffer turns out you are right too. After some testing passing -std=gnu++98 compiles the file.
However in that case it dumps to the stream a memory location and not the stringstream contents so I guess it is not the intended behavior.

jpfeuffer:
Aha! Thanks a ton for testing this. This probably would have gone unnoticed.

jpfeuffer:
@grosenberger Could you have a short look at this and create a small follow-up PR?

hroest:
I made a PR for the small fix

hroest:
see some more information about the reduction in file size in https://github.com/grosenberger/OpenMS/pull/6

PR #2484, opened 2017-03-28
[OBO] update to psi-ms 4.0.10

hroest:
has quite a selection of goodies, including some that we need

```
+[Term]
+id: MS:1002746
+name: MS-Numpress linear prediction compression followed by zlib compression
```

this will allow us to write substantially smaller mzML files. We have to check whether this breaks things, maybe @mwalzer can comment. If any of the names of the CV terms has changed, we may get warnings in our tests ...

### DIFF
--- a/share/OpenMS/CV/psi-ms.obo
+++ b/share/OpenMS/CV/psi-ms.obo
@@ -7112,7 +7112,7 @@ name: retention time(s)
 def: "OBSOLETE Retention time of the spectrum from the source file." [PSI:PI]
 comment: This term was made obsolete because scan start time (MS:1000016) should be used instead.
 xref: value-type:xsd\:double "The allowed value-type for this CV term."
-is_a: MS:1001105 ! peptide result details
+is_a: MS:1001105 ! peptide sequence-level identification attribute
 is_a: MS:1001405 ! spectrum identification result details
 relationship: has_units UO:0000010 ! second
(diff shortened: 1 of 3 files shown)


hroest:
it seems the following tests fail: `TOPP_IDFileConverter_9` anybody know why?

hroest:
it seems the culprit is here: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp#L1223

@mwalzer can you comment here, can we try and get the CV term by id instead of by name? that would seem more stable as people change the CV term name !

hroest:
PS: I also updated the obo to get rid of the warnings, fef948c so the obo is now *not* the one from the standard committee but it seems that one was a bit broken.

mwalzer:
@hroest that was requested time back for 'readability' ...  at some point I switched to by ```getTerm()``` followed by the term name as  inline comments

mwalzer:
release candidate 4.0.10_rc1 of the psi-ms.obo has all that fixed, release should be soon

hroest:
yes, lets wait a few more days for the final release of 4.0.10

mwalzer:
https://sourceforge.net/p/psidev/mailman/message/35760486/

hroest:
great, I now changed it to 4.0.10 the final version. The commit 3e7ee4f is minimal so lets merge!

PR #2387, opened 2017-02-02
Major update of FeatureFinderIdentification

hendrikweisser:
FeatureFinderIdentification (FFId) gets the ability to combine peptide IDs from multiple LC-MS/MS runs in a smarter fashion. "Internal" IDs (same LC-MS/MS run) are distinguished from "external" ones (other LC-MS/MS runs). A machine learning approach (SVM) is used to predict correct features for external IDs.

Introduces two helper classes, SimpleSVM (wrapper of LIBSVM for classification of features) and ElutionModelFitter (for fitting elution models to features). Tests for those classes are missing at the moment, but will be added.

PR #2486, opened 2017-03-28
FeatureFinderIdentification runtime optimization

hendrikweisser:
Following discussion here: https://github.com/OpenMS/OpenMS/pull/2387#issuecomment-286886978
This pull request refactors FeatureFinderIdentification to extract chromatograms (and then detect/score chromatographic peaks) for all peptides together, rather than one by one. This can save several hours of runtime.
The results should be the same as before (except for ordering).

There's a small addition to `ChromatogramExtractor::prepare_coordinates` for convenient use of varying RT start/end points.


timosachsenberg:
@hendrikweisser can you check why tests are failing?

timosachsenberg:
I restarted the builds... let's see if this helps..

hendrikweisser:
> can you check why tests are failing?

I'm on it. There's one test that consistently passes on my Mac, but fails on a Linux system. I'm looking into it.

timosachsenberg:
+1

PR #2487, opened 2017-03-28
fix msexperiment windows

FabianAicheler:
typedef typename caused error in VS 11 2012 when not part of template.

We'll need to check whether the fixed files work under linux


hroest:
:+1: works in travis, I therefore expect no problems! Thanks

PR #2489, opened 2017-03-29
[FIX] windows build fix

hroest:
may fix the windows build

we still have some issues with the windows build: http://cdash.openms.de/viewBuildError.php?buildid=217346

```
OpenSwathMRMFeatureAccessOpenMS_test.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __thiscall OpenMS::TransitionGroupOpenMS<class OpenMS::MSChromatogram<class OpenMS::ChromatogramPeak>,class OpenMS::ReactionMonitoringTransition>::TransitionGroupOpenMS<class OpenMS::MSChromatogram<class OpenMS::ChromatogramPeak>,class OpenMS::ReactionMonitoringTransition>(class OpenMS::MRMTransitionGroup<class OpenMS::MSChromatogram<class OpenMS::ChromatogramPeak>,class OpenMS::ReactionMonitoringTransition> &)" (__imp_??0?$TransitionGroupOpenMS@V?$MSChromatogram@VChromatogramPeak@OpenMS@@@OpenMS@@VReactionMonitoringTransition@2@@OpenMS@@QAE@AAV?$MRMTransitionGroup@V?$MSChromatogram@VChromatogramPeak@OpenMS@@@OpenMS@@VReactionMonitoringTransition@2@@1@@Z) referenced in function __catch$_main$14 [C:\dev\NIGHTLY\builds\win8-VS12_x86-Debug\src\tests\class_tests\openms\OpenSwathMRMFeatureAccessOpenMS_test.vcxproj]
```

I hope the attached patch fixes this. Note that eventually `MSSpectrum<ChromatogramPeak>` will go away....

### DIFF
--- a/src/openms/source/KERNEL/MRMTransitionGroup.cpp
+++ b/src/openms/source/KERNEL/MRMTransitionGroup.cpp
@@ -43,7 +43,9 @@ namespace OpenMS
 {
   MRMTransitionGroup<MSSpectrum<Peak1D>, ReactionMonitoringTransition> default_mrmtransitiongroup_;
   MRMTransitionGroup<MSSpectrum<Peak1D>, OpenSwath::LightTransition> default_openswath_mrmtransitiongroup_;
-  MRMTransitionGroup<MSSpectrum<ChromatogramPeak>, ReactionMonitoringTransition> default_chrompeak_mrmtransitiongroup_;
-  MRMTransitionGroup<MSSpectrum<ChromatogramPeak>, OpenSwath::LightTransition> default_chrompeak_openswath_mrmtransitiongroup_;
+  MRMTransitionGroup<MSSpectrum<ChromatogramPeak>, ReactionMonitoringTransition> default_chrompeak_mrmtransitiongroup_; // legacy, remove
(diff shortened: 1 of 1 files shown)


cbielow:
current develop compiles `OpenSwathMRMFeatureAccessOpenMS_test` on my VS2013 (VS14) and it seems rather weird that the symbols are missing since this is a template class, i.e. the test should see the full implementation and can simply instanciate it.
Nevertheless, lets try this fix and see what happens.

hroest:
yes, but I think we have had the same issues, see ec18ab32adabe3e8 back in 2012

cbielow:
the problem might acutally be something related:
You specified `struct OPENSWATHALGO_DLLAPI LightTransition` in TransisitionExperiment.h, yet the implementation of this class is within the header.
Therefore, any tool referencing it (including the MRMTransitionGroup) when linking against the library expects the implementation to be in the lib -- which it is not, since its header only.
So once you fix wrong usage of OPENSWATHALGO_DLLAPI, the problem should go away.

cbielow:
two ways to fix it: remove OPENSWATHALGO_DLLAPI, or move implementation to cpp.

PR #2490, opened 2017-03-30
[FIX] intensity_percentage_ and intensity_percentage_optional were swapped

DennisGoldfarb:
In FeatureFinderAlgorithmPicked, based on the documentation and code, intensity_percentage_ is meant to be a hard threshold on isotope probabilities (the default value FeatureFinderCentroided is 0.01) and intensity_percentage_optional_ was meant to be a soft threshold (the default value in FeatureFinderCentroided is 0.10).

A theoretical isotope whose probability is below intensity_percentage_ should be trimmed from the IsotopeDistribution, and then anything below intensity_percentage_optional_ should be labeled as optional in the TheoreticalIsotopePattern. However, right now intensity_percentage_optional_ is being used to trim the IsotopeDistribution, and intensity_percentage_ is being used for the "optional" annotations. So with the current defaults, anything below 0.1 is trimmed off completely, and then anything below 0.01 is labeled optional, but obviously nothing gets that label because they've already been trimmed off.

This PR just swaps the variables so they are used in the correct places and the unit tests were updated. They appear to be regression tests.

Here are the results of FeatureFinderCentroided using default values on a whole cell HeLa lysate:

Before bug fix:
=========================================================================
16:58:44 NOTICE: FeatureFinderCentroided of node #2 started. Processing ...
Found 227865 seeds for charge 1.
Found 42583 feature candidates for charge 1.
Found 133510 seeds for charge 2.
Found 33193 feature candidates for charge 2.
Found 59601 seeds for charge 3.
Found 12212 feature candidates for charge 3.
Found 8719 seeds for charge 4.
Found 1413 feature candidates for charge 4.
Removed 36039 overlapping features.
53362 features left.

Abort reasons during feature construction:
- Could not extend seed: 83498
- Could not find good enough isotope pattern containing the seed: 18611
- Feature quality too low after fit: 5618
- Invalid feature after fit - too few traces or peaks left: 107782
- Invalid fit: Center outside of feature bounds: 1123
- Invalid fit: Fitted model is bigger than 'max_rt_span': 18974
- Invalid fit: Less than 'min_rt_span' left after fit: 14
FeatureFinderCentroided took 09:10 m (wall), 09:02 m (CPU), 0.00 s (system), 09:02 m (user).

After bug fix
=========================================================================
17:10:04 NOTICE: FeatureFinderCentroided of node #2 started. Processing ...
Found 221279 seeds for charge 1.
Found 47125 feature candidates for charge 1.
Found 128414 seeds for charge 2.
Found 39522 feature candidates for charge 2.
Found 55138 seeds for charge 3.
Found 12799 feature candidates for charge 3.
Found 7297 seeds for charge 4.
Found 1346 feature candidates for charge 4.
Removed 44820 overlapping features.
55972 features left.

Abort reasons during feature construction:
- Could not extend seed: 82786
- Could not find good enough isotope pattern containing the seed: 12035
- Feature quality too low after fit: 6444
- Invalid feature after fit - too few traces or peaks left: 98872
- Invalid fit: Center outside of feature bounds: 1010
- Invalid fit: Fitted model is bigger than 'max_rt_span': 19406
- Invalid fit: Less than 'min_rt_span' left after fit: 10
FeatureFinderCentroided took 10:25 m (wall), 08:56 m (CPU), 0.00 s (system), 08:56 m (user).

So that's an improvement of almost 5% in total number of features!

### DIFF
--- a/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderAlgorithmPicked.cpp
+++ b/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderAlgorithmPicked.cpp
@@ -411,9 +411,9 @@ namespace OpenMS
         d.estimateFromPeptideWeight(0.5 * mass_window_width_ + index * mass_window_width_);
         //trim left and right. And store the number of isotopes on the left, to reconstruct the monoisotopic peak
         Size size_before = d.size();
-        d.trimLeft(intensity_percentage_optional_);
+        d.trimLeft(intensity_percentage_);
         isotope_distributions_[index].trimmed_left = size_before - d.size();
-        d.trimRight(intensity_percentage_optional_);
+        d.trimRight(intensity_percentage_);
 
         for (IsotopeDistribution::Iterator it = d.begin(); it != d.end(); ++it)
         {
@@ -429,7 +429,7 @@ namespace OpenMS
 
         for (Size i = 0; i < isotope_distributions_[index].intensity.size(); ++i)
         {
-          if (isotope_distributions_[index].intensity[i] < intensity_percentage_)
+          if (isotope_distributions_[index].intensity[i] < intensity_percentage_optional_)
           {
             if (!is_end && !is_begin) is_end = true;
             if (is_begin) ++begin;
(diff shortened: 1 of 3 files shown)


timosachsenberg:
nice catch!  +1

hroest:
:+1: looks great - maybe add a comment in there to explain the hard/soft thresholding?

DennisGoldfarb:
Ahh nevermind! I just looked at it with fresh eyes and I was completely wrong. The current code is correct.  It was my change that swapped the meaning, but the difference in feature IDs was real (when I accidentally used a hard threshold of 0.1). So maybe I need to optimize parameters for my experiments. Sorry about that.

timosachsenberg:
thanks for double checking!

Issue #2366, opened 2017-01-25
Error validating idXML file against XML schema


SharmeelaAshwin:
On running the "`fileinfo -v -in CC271.idXML`" on CC271.idXML file  below pasted errors are displayed

```
-- General information --

File name: CC271.idXML
File type: idXML

Validating idXML file against XML schema version 1.4
Validation error in file 'CC271.idXML' line 33 column 207: value 'controllerType
=0 controllerNumber=1 scan=26981' does not match regular expression facet '[+\-]
?[0-9]+'
Validation error in file 'CC271.idXML' line 64 column 207: value 'controllerType
=0 controllerNumber=1 scan=6989' does not match regular expression facet '[+\-]?
[0-9]+'
Validation error in file 'CC271.idXML' line 95 column 207: value 'controllerType
=0 controllerNumber=1 scan=17644' does not match regular expression facet '[+\-]
?[0-9]+'
Validation error in file 'CC271.idXML' line 126 column 208: value 'controllerTyp
e=0 controllerNumber=1 scan=22967' does not match regular expression facet '[+\-
]?[0-9]+'
Validation error in file 'CC271.idXML' line 157 column 207: value 'controllerTyp
e=0 controllerNumber=1 scan=13794' does not match regular expression facet '[+\-
]?[0-9]+'
Validation error in file 'CC271.idXML' line 188 column 208: value 'controllerTyp
e=0 controllerNumber=1 scan=24365' does not match regular expression facet '[+\-
]?[0-9]+'
Validation error in file 'CC271.idXML' line 219 column 208: value 'controllerTyp
e=0 controllerNumber=1 scan=17680' does not match regular expression facet '[+\-
]?[0-9]+'
Validation error in file 'CC271.idXML' line 250 column 207: value 'controllerTyp
e=0 controllerNumber=1 scan=29897' does not match regular expression facet '[+\-
]?[0-9]+'
Failed - errors are listed above!
FileInfo took 0.06 s (wall), 0.02 s (CPU), 0.00 s (system), 0.02 s (user).
```

Please find the link to the uploaded file CC271.idXML here : https://gist.github.com/SharmeelaAshwin/df58e22565c442ea29a495db3274a6fd

jpfeuffer:
Hi!

Where does the idXML come from? From conversion with IDFileConverter?
If so, please check the documentation and consider setting the scan_regex option of the tool.
[...]

PR #2491, opened 2017-03-30
Feature/id xml schema quickfix

mwalzer:
mitigating #2366


share/OpenMS/SCHEMAS/IdXML_1_4.xsd
```diff
@@ -249,7 +249,7 @@
 				<xs:documentation>significance threshold as calculated by the search engine</xs:documentation>
 			</xs:annotation>
 		</xs:attribute>
-		<xs:attribute name="spectrum_reference" type="xs:unsignedInt">
+                <xs:attribute name="spectrum_reference" type="xs:string">
```
timosachsenberg: could you comment on 1.4 change?

timosachsenberg: e.g. the documentation needs to be adapted to. also I don't know if we would need to increase the version number of idXML

src/openms/source/FORMAT/IdXMLFile.cpp
```diff
@@ -101,13 +101,13 @@ namespace OpenMS
 
     //write header
     os << "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n";
-    os << "<?xml-stylesheet type=\"text/xsl\" href=\"http://open-ms.sourceforge.net/XSL/IdXML.xsl\" ?>\n";
+    os << "<?xml-stylesheet type=\"text/xsl\" href=\"https://raw.githubusercontent.com/OpenMS/OpenMS/develop/share/OpenMS/XSL/IdXML.xsl\" ?>\n";
```
hroest: seems like a not very stable URL as the `develop` branch changes all the time - could we just use a release tag? e.g. if we link to release 2.1 then it would be a stable version...

share/OpenMS/SCHEMAS/IdXML_1_4.xsd
hroest: seems like it is not backward compatible, e.g. if there is an 1.4 file that has a string it cannot be read by a parser that uses the old 1.4 so there should be a 1.5 version imho

mwalzer: Please move discussion to the issue! #2366

mwalzer: albeit, we [read](https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/IdXMLFile.cpp#L566) and [wrote](https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/IdXMLFile.cpp#L283) whatever was in there as string.

src/openms/source/FORMAT/IdXMLFile.cpp
mwalzer: well, the schema should be stable, regardless of the branch. Has to start 'living' somewhere outside sourceforge ... - If you want to open a release 2.1 branch, we can put it there.

share/OpenMS/SCHEMAS/IdXML_1_4.xsd
timosachsenberg: I am, in principle, fine with having the change if it was a bug in the idXML schema file. But you should definitely update the schema documentation (still refers to an integer)

src/openms/source/FORMAT/IdXMLFile.cpp
timosachsenberg: every release tag can be directly accessed: e.g.,  https://github.com/OpenMS/OpenMS/tree/Release2.1.0

hroest: exactly, so please use https://raw.githubusercontent.com/OpenMS/OpenMS/Release2.1.0/share/OpenMS/XSL/IdXML.xsl

hroest: we can also of course put it somewhere here: http://ftp.mi.fu-berlin.de/pub/OpenMS/ if people are more comfortable with this

mwalzer: @timosachsenberg @jpfeuffer @hroest  so is it consent, that we point to an outdated file until the next release? Or the Berlin server? 
I tend to the latter with opening an issue to update all file writers to point to the next release tag.

timosachsenberg: Both is kind of suboptimal.
I would prefer a stable URL pointing to: www.openms.org/...
but I don't know how well redirects are handled or allowed.
also  the XSL should contain the version number of the idXML version it corresponds to.

mwalzer:
So, I think what @hroest said was quite true. Therefore, I subversioned 1.4 to 1.4.1 and made the respective changes. 
Pointing to openms.de (org?) and redirecting to ```raw.github...```` sounds like the best option we should persue. Will we change the TL-domain ending soon?

mwalzer:
so, pass-through seems to be _not_ working (```<html><head></head><body></body></html>``` response), github will always be visible, but that is better than nothing. 
The 301 redirect is now  ```/xml-schema/(.*).xsd -> https://raw.githubusercontent.com/OpenMS/OpenMS/develop/share/OpenMS/SCHEMAS/...``` similar for the stylesheets.
I've changed the respective attributes, and I think we should also add this to the release-how-to-steps to adjust the redirect endpoint to the correct release tag url upon any new release.

share/OpenMS/SCHEMAS/IdXML_1_4.xsd
```diff
@@ -249,7 +249,7 @@
 				<xs:documentation>significance threshold as calculated by the search engine</xs:documentation>
 			</xs:annotation>
 		</xs:attribute>
-		<xs:attribute name="spectrum_reference" type="xs:unsignedInt">
+                <xs:attribute name="spectrum_reference" type="xs:unsignedInt">
```
timosachsenberg: spaces used while tabs were used before

src/tests/class_tests/openms/data/FalseDiscoveryRate_OMSSA.idXML
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-stylesheet type="text/xsl" href="http://open-ms.sourceforge.net/XSL/IdXML.xsl" ?>
-<IdXML version="1.4" xsi:noNamespaceSchemaLocation="http://open-ms.sourceforge.net/SCHEMAS/IdXML_1_4.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<IdXML version="1.5" xsi:noNamespaceSchemaLocation="http://open-ms.sourceforge.net/SCHEMAS/IdXML_1_4.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
timosachsenberg: still point to old schema

mwalzer: darn, I thought I had them all repla"sed" :see_no_evil:

timosachsenberg:
looks good to me. In this case, I squash and merge so it's easier revertable.

PR #2492, opened 2017-03-30
huge fixes for Visual Studio 2015 & Qt 5.8

cbielow:
- introduces autoMOC & autoUIC
- uses proper inheritance and encapsulation for ui_*.h in DIALOGS (dont expose GUI objects directly)
- removes WebEngine dependency
- fixes Qt5 OpenGL issues & access to widgets (active vs. currentWindow)
- fix CXX_FLAGS forwarding


hroest:
@cbielow wow this is really great work. A quick question though: are you using features from 5.8 and more generally: what should be the minimal Qt version that we still run on. I am asking because debian currently supports https://packages.debian.org/search?keywords=qt5-default

- 5.3.2 [jessie/stretch]
- 5.7.1 [stretch/sid]
- 5.8.0

while ubuntu uses http://packages.ubuntu.com/search?keywords=qt5-default&searchon=names

- 5.2.1 [14.04]
- 5.5.1 [16.04]
- 5.6.1 [16.10]
- 5.7.1 [17.04 / zesty]
 
while Fedora seems to be pretty up to date (https://apps.fedoraproject.org/packages/qt5-qtbase ) using 5.6+ . So from an Ubuntu perspective, it would be good to support an API that does not go beyond 5.2 or 5.5 -- I think for biodocker and other container groups 16.04 is the newest image they will use so using features beyond 5.5 may cause problems down the line.

hroest:
PS: it is of course possible to install ppa repos for individual users such as https://launchpad.net/~beineri  but of course for package maintainers it will be a big pain or close to impossible to have openms depend on a not-shipped Qt version. From this point of view I think 5.6.1 may be a good compromise

cbielow:
I picked Qt 5.8 because it had pre-build binaries for Windows readily available. There should not be any issue in downgrading to anythig >= Qt 5.4. Anything prior might(!) cause trouble due to slight API changes, but we'd need to check more thoroughly.

hroest:
great, but that means we will have to change travis to 16.04 as this will be 5.5.1

src/tests/class_tests/openms_gui/CMakeLists.txt
```diff
@@ -100,9 +100,9 @@ if(HAS_XSERVER)
   # Add GUI tests
   foreach(_gui_test ${GUI_executables_list})
     # moc the test file
-    QT5_WRAP_CPP(_mocced_test_file source/GUI/${_gui_test}.cpp OPTIONS ${BOOST_MOC_ARGS})
+    #QT5_WRAP_CPP(_mocced_test_file source/GUI/${_gui_test}.cpp OPTIONS ${BOOST_MOC_ARGS})
     # define source file dependency
-    set_source_files_properties(source/GUI/${_gui_test}.cpp PROPERTIES OBJECT_DEPENDS ${_mocced_test_file})
+    #set_source_files_properties(source/GUI/${_gui_test}.cpp PROPERTIES OBJECT_DEPENDS ${_mocced_test_file})
```
hroest: maybe we would have needed this for the TOPPView_test ? seems that it needs to be mocced as well?

Issue #2465, opened 2017-03-22
review FalseDiscoveryRate tool


timosachsenberg:
As always just a suggestion open for discussion.

What about:
  -in_target <file>  Identification input file containing a search against a target-only database. (valid formats: 'idXML')
  -in_decoy <file>   Identification input file containing a search against a decoy-only database. (valid 

If there is no good reason for them I would remove them. They only complicate the interface and make integration into workflow systems more difficult (e.g., the engine can't check the semantics "Either specify '-in' alone or 'in_target' together with 'in_decoy' as input.")

Make following parameter advanced:
  -proteins_only     If set only the FDR on protein level is calculated
  -peptides_only     If set only the FDR on peptide (PSM) level is calculated

```
Options (mandatory options marked with '*'):
  -in <file>         Identification input file containing a search against a concatenated sequence database. Either specify '-in' alone or 'in_target' together with 'in_decoy' as input.
                     (valid formats: 'idXML')
  -in_target <file>  Identification input file containing a search against a target-only database. (valid formats: 'idXML')
  -in_decoy <file>   Identification input file containing a search against a decoy-only database. (valid formats: 'idXML')
  -out <file>*       Identification output with annotated FDR (valid formats: 'idXML')
  -proteins_only     If set only the FDR on protein level is calculated
  -peptides_only     If set only the FDR on peptide (PSM) level is calculated
```

hendrikweisser:
> Make following parameter advanced:
> -proteins_only If set only the FDR on protein level is calculated
> -peptides_only If set only the FDR on peptide (PSM) level is calculated

I think these parameters are too important to make them advanced. I don't use the FDR tool often, but when I do, I only want to calculate FDRs for one of the levels.

jpfeuffer:
Yes, please leave it. I agree with @hendrikweisser

timosachsenberg:
sure. I just scanned the top used tools and wanted to bring these parameters to discussion

timosachsenberg:
what about the other ones I mentioned?

PR #2495, opened 2017-04-03
FalseDiscoveryRate: removed redundant parameters, simplified tool

timosachsenberg:
performed changes decided in https://github.com/OpenMS/OpenMS/issues/2465

### DIFF
--- a/src/topp/FalseDiscoveryRate.cpp
+++ b/src/topp/FalseDiscoveryRate.cpp
@@ -29,7 +29,7 @@
 //
 // --------------------------------------------------------------------------
 // $Maintainer: Chris Bielow $
-// $Authors: Andreas Bertsch $
+// $Authors: Timo Sachsenberg, Andreas Bertsch $
 // --------------------------------------------------------------------------
 
 #include <OpenMS/APPLICATIONS/TOPPBase.h>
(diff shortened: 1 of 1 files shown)


mwalzer:
:+1:

src/topp/FalseDiscoveryRate.cpp
```diff
@@ -103,14 +103,8 @@ class TOPPFalseDiscoveryRate :
 
   void registerOptionsAndFlags_()
   {
-    registerInputFile_("in", "<file>", "", "Identification input file containing a search against a concatenated sequence database. Either specify '-in' alone or 'in_target' together with 'in_decoy' as input.", false);
+    registerInputFile_("in", "<file>", "", "Identification input file containing a search against a concatenated sequence database. Either specify '-in' alone or 'in_target' together with 'in_decoy' as input.");
```
cbielow: `in_*` was removed, right?! Update docu

src/topp/FalseDiscoveryRate.cpp
```diff
@@ -63,11 +63,11 @@ using namespace std;
     </table>
 </CENTER>
 
-    This TOPP tool calculates the false discovery rate (FDR) for results of target-decoy searches. It can handle separate searches of a target database (e.g. forward sequences) and a decoy database (e.g. reversed sequences), or a combined search of a concatenated target-decoy database. The FDR calculation can be performed for proteins and/or for peptides (more exactly, peptide spectrum matches).
+    This TOPP tool calculates the false discovery rate (FDR) for results of target-decoy searches. The FDR calculation can be performed for proteins and/or for peptides (more exactly, peptide spectrum matches).
```
cbielow: maybe add a sentence like:

`FalseDiscoveryRate only annotates peptides and proteins with their FDR. A subsequent FDR filtering step needs to be conducted downstream via @ref IDFilter or after exporting the data to a text-based format.`

src/topp/FalseDiscoveryRate.cpp
```diff
@@ -29,7 +29,7 @@
 //
 // --------------------------------------------------------------------------
 // $Maintainer: Chris Bielow $
```
cbielow: want to become maintainer? :)

src/topp/FalseDiscoveryRate.cpp
```diff
@@ -167,72 +138,38 @@ class TOPPFalseDiscoveryRate :
...
+      return INCOMPATIBLE_INPUT_DATA;
+    }
 
-      // TODO @all shouldn't ranks be assigned here as well?
-
-      //-------------------------------------------------------------
-      // writing output
-      //-------------------------------------------------------------
-      IdXMLFile().store(out, prot_target, pep_target);
+    for (vector<ProteinIdentification>::iterator it = prot_ids.begin(); it != prot_ids.end(); ++it)
```
cbielow: just a note: this bloat code is one of the reasons we should have a dedicated class for `vector<Pep|Prot>` things

timosachsenberg:
can we together scetch out the (hopefully quite minimal) interface of these classes? this might lower the bar for e.g. new members to implement this?

timosachsenberg:
By the way: If you do the changes please just add commits (no rebase)

cbielow:
Should we do that here? The changes will be rather large, since we’d need to replace every occurrence of vector<Pep|ProtId> in the lib, right?

--
Dr. Chris Bielow

Head of Bioinformatics Solution Center (BSC)
Freie Universität Berlin, Institut für Informatik, Takustr. 9, room K21, 14195 Berlin
Phone: +49 (0)30 838-63084, Email: Chris.Bielow(at)fu-berlin.de

timosachsenberg:
no maybe new issue

src/topp/FalseDiscoveryRate.cpp
timosachsenberg: its not sooo bad if we have auto ;) but I agree that we should have a dedicated, non-template class that aggregates the IDs

timosachsenberg:
I addressed all comments. merging

PR #2496, opened 2017-04-04
Fix/uint compile

timosachsenberg:


timosachsenberg:
should fix http://cdash.openms.de/viewBuildError.php?buildid=217995

jpfeuffer:
Ah nice.. (and bad since I was on it, too). There are actually more uint types in the project and I am not sure why they do not raise errors. That was I was investigating right now.
It is actually surprising that these errors came up out of nowhere. The last change was just an include of another header.

PR #1797, opened 2016-02-03
Feature/spectrum reference

mwalzer:
added spectrum reference annotation for idXML in IDFileConverter, if input is idXML (most likely, these got lost during reading of the original fileformat of the searchengine, if it was even annotated)

also added a function to add do optional ion matching in SpectraMetaDataLookup when spectra are matched anyways  for  spectrum reference lookup

PR #2498, opened 2017-04-04
Feature/general spectrum reference annotation

mwalzer:
This is the split rework of #1797 , closed after scope creep.
PR is for consistent update/complementation  of id result files.
Some existing functionality for that was refactored from IDFileConverter into SpectrumMetaDataLookup.
Now, either adding missing RT or adding missing spectrum references is possible, available through IDFileConverter.

Rest of #1797 comes with another PR which is actually based on this.

### DIFF
--- a/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
+++ b/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
@@ -288,14 +288,14 @@ namespace OpenMS
      * @brief Add missing "spectrum_reference"s to peptide identifications based on raw data
      *
      * @param peptides Peptide IDs with or without spectrum_reference
-     * @param lookup SpectrumMetaDataLookup with loaded spectra, metadata and optional spectra_data value for Proteins
+     * @param filename the name of the mz_file from which to draw spectrum_references
      * @param stop_on_error Stop when an ID could not be matched to a spectrum (or keep going)?
      * @param override_spectra_data if given ProteinIdentifications should be updated with new "spectra_data" values from SpectrumMetaDataLookup
-     * @param proteins
+     * @param proteins Protein IDs corresponding to the Peptide IDs
      *
      * @return True if all peptide IDs could be annotated successfully (including if all already had "spectrum_reference" values), false otherwise.
      *
-     * Look-up works by matching RT of a peptide ID to the native ID of a spectrum. All spectrum_references are updated/added.
+     * Look-up works by matching RT of a peptide identification with the given spectra. Matched spectra 'native ID' will be annotated to the identification. All spectrum_references are updated/added.
      */
     static bool addMissingSpectrumReferences(std::vector<PeptideIdentification>& peptides, const String& filename,
       bool stop_on_error = false, bool override_spectra_data = false, std::vector<ProteinIdentification> proteins = std::vector<ProteinIdentification>());
--- a/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
+++ b/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
@@ -157,7 +157,6 @@ namespace OpenMS
     static const MetaDataFlags MDF_NATIVEID = 64;
     static const MetaDataFlags MDF_ALL = 127;
 
-    String spectra_data;
 
     /// Meta data of a spectrum
     struct SpectrumMetaData
@@ -181,8 +180,7 @@ namespace OpenMS
     };
 
     /// Constructor
-    SpectrumMetaDataLookup(): SpectrumLookup(),
-      spectra_data()
+    SpectrumMetaDataLookup(): SpectrumLookup()
     {}
 
     /// Destructor
@@ -225,15 +223,16 @@ namespace OpenMS
 
     /**
      * @brief encapsulates readSpectra with file loading convenience if only a file path is given
+     *
      * @param spectra_file file to the SpectrumContainer, must be valid path to mz file
      */
-    void readMzFile(const String& spectra_file)
+    void readMzFileMetaData(const String& spectra_file)
     {
       PeakMap exp;
       FileHandler().loadExperiment(spectra_file, exp, FileTypes::UNKNOWN, ProgressLogger::NONE,
                                    false, false);
-      this->readSpectra(exp.getSpectra());
-      this->setSpectraData(File::absolutePath(spectra_file));
+      readSpectra(exp.getSpectra());
+      setSpectraDataRef(File::absolutePath(spectra_file));
     }
 
     /**
@@ -241,9 +240,9 @@ namespace OpenMS
      *
      * @param spectra_data the name (and path) of the origin of the read SpectrumContainer
      */
-    void setSpectraData(const String& spectra_data)
+    void setSpectraDataRef(const String& spectra_data)
     {
-      this->spectra_data = spectra_data;
+      this->spectra_data_ref = spectra_data;
     }
 
 
@@ -295,29 +294,29 @@ namespace OpenMS
 
        Look-up works by matching the "spectrum_reference" (meta value) of a peptide ID to the native ID of a spectrum. Only peptide IDs without RT (where PeptideIdentification::getRT() returns "NaN") are looked up; the RT is set to that of the corresponding spectrum.
     */
-    static bool addMissingRTsToPeptideIDs(
-      std::vector<PeptideIdentification>& peptides, const SpectrumMetaDataLookup& lookup,
+    static bool addMissingRTsToPeptideIDs(std::vector<PeptideIdentification>& peptides, const String &filename,
       bool stop_on_error = false);
 
     /**
-     * @brief Add missing spectrum_references to peptide identifications vased on raw data
+     * @brief Add missing "spectrum_reference"s to peptide identifications based on raw data
      *
      * @param peptides Peptide IDs with or without spectrum_reference
      * @param lookup SpectrumMetaDataLookup with loaded spectra, metadata and optional spectra_data value for Proteins
      * @param stop_on_error Stop when an ID could not be matched to a spectrum (or keep going)?
-     * @param override_spectra_data if given ProteinIdentifications should be updated with new spectra_data values from SpectrumMetaDataLookup
+     * @param override_spectra_data if given ProteinIdentifications should be updated with new "spectra_data" values from SpectrumMetaDataLookup
      * @param proteins
      *
-     * @return True if all peptide IDs could be annotated successfully (including if all already had spectrum_referece values), false otherwise.
+     * @return True if all peptide IDs could be annotated successfully (including if all already had "spectrum_reference" values), false otherwise.
      *
      * Look-up works by matching RT of a peptide ID to the native ID of a spectrum. All spectrum_references are updated/added.
      */
-    static bool addMissingSpectrumReferences(std::vector<PeptideIdentification>& peptides, const SpectrumMetaDataLookup& lookup,
+    static bool addMissingSpectrumReferences(std::vector<PeptideIdentification>& peptides, const String& filename,
       bool stop_on_error = false, bool override_spectra_data = false, std::vector<ProteinIdentification> proteins = std::vector<ProteinIdentification>());
 
   protected:
 
     std::vector<SpectrumMetaData> metadata_; ///< Meta data for spectra
+    String spectra_data_ref;
 
   private:
 
--- a/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
+++ b/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
@@ -39,6 +39,8 @@
 
 #include <OpenMS/FORMAT/FileHandler.h>
 #include <OpenMS/KERNEL/MSSpectrum.h>
+#include <OpenMS/METADATA/ProteinIdentification.h>
+#include <OpenMS/KERNEL/MSExperiment.h>
 
 #include <limits> // for "quiet_NaN"
 
@@ -155,6 +157,8 @@ namespace OpenMS
     static const MetaDataFlags MDF_NATIVEID = 64;
     static const MetaDataFlags MDF_ALL = 127;
 
+    String spectra_data;
+
     /// Meta data of a spectrum
     struct SpectrumMetaData
     {
@@ -177,7 +181,9 @@ namespace OpenMS
     };
 
     /// Constructor
-    SpectrumMetaDataLookup(): SpectrumLookup() {}
+    SpectrumMetaDataLookup(): SpectrumLookup(),
+      spectra_data()
+    {}
 
     /// Destructor
     virtual ~SpectrumMetaDataLookup() {}
@@ -185,7 +191,7 @@ namespace OpenMS
     /**
        @brief Read spectra and store their meta data
 
-       @tparam SpectrumContainer Spectrum container class, must support @p size and @p operator[]
+       @param SpectrumContainer Spectrum container class, must support @p size and @p operator[]
 
        @param spectra Container of spectra
        @param scan_regexp Regular expression for matching scan numbers in spectrum native IDs (must contain the named group "?<SCAN>")
@@ -216,6 +222,30 @@ namespace OpenMS
         metadata_.push_back(meta);
       }
     }
+
+    /**
+     * @brief encapsulates readSpectra with file loading convenience if only a file path is given
+     * @param spectra_file file to the SpectrumContainer, must be valid path to mz file
+     */
+    void readMzFile(const String& spectra_file)
+    {
+      PeakMap exp;
+      FileHandler().loadExperiment(spectra_file, exp, FileTypes::UNKNOWN, ProgressLogger::NONE,
+                                   false, false);
+      this->readSpectra(exp.getSpectra());
+      this->setSpectraData(File::absolutePath(spectra_file));
+    }
+
+    /**
+     * @brief set spectra_data from read SpectrumContainer origin (i.e. filename)
+     *
+     * @param spectra_data the name (and path) of the origin of the read SpectrumContainer
+     */
+    void setSpectraData(const String& spectra_data)
+    {
+      this->spectra_data = spectra_data;
+    }
+
 
     /**
        @brief Look up meta data of a spectrum
@@ -264,12 +294,26 @@ namespace OpenMS
        @return True if all peptide IDs could be annotated successfully (including if all already had RT values), false otherwise.
 
        Look-up works by matching the "spectrum_reference" (meta value) of a peptide ID to the native ID of a spectrum. Only peptide IDs without RT (where PeptideIdentification::getRT() returns "NaN") are looked up; the RT is set to that of the corresponding spectrum.
-
-       The raw data is only loaded from @p filename if necessary, i.e. if there are any peptide IDs with missing RTs.
     */
     static bool addMissingRTsToPeptideIDs(
-      std::vector<PeptideIdentification>& peptides, const String& filename,
+      std::vector<PeptideIdentification>& peptides, const SpectrumMetaDataLookup& lookup,
       bool stop_on_error = false);
+
+    /**
+     * @brief Add missing spectrum_references to peptide identifications vased on raw data
+     *
+     * @param peptides Peptide IDs with or without spectrum_reference
+     * @param lookup SpectrumMetaDataLookup with loaded spectra, metadata and optional spectra_data value for Proteins
+     * @param stop_on_error Stop when an ID could not be matched to a spectrum (or keep going)?
+     * @param override_spectra_data if given ProteinIdentifications should be updated with new spectra_data values from SpectrumMetaDataLookup
+     * @param proteins
+     *
+     * @return True if all peptide IDs could be annotated successfully (including if all already had spectrum_referece values), false otherwise.
+     *
+     * Look-up works by matching RT of a peptide ID to the native ID of a spectrum. All spectrum_references are updated/added.
+     */
+    static bool addMissingSpectrumReferences(std::vector<PeptideIdentification>& peptides, const SpectrumMetaDataLookup& lookup,
+      bool stop_on_error = false, bool override_spectra_data = false, std::vector<ProteinIdentification> proteins = std::vector<ProteinIdentification>());
 
   protected:
 
--- a/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
+++ b/src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
@@ -221,19 +221,6 @@ namespace OpenMS
       }
     }
 
-    /**
-     * @brief encapsulates readSpectra with file loading convenience if only a file path is given
-     *
-     * @param spectra_file file to the SpectrumContainer, must be valid path to mz file
-     */
-    void readMzFileMetaData(const String& spectra_file)
-    {
-      PeakMap exp;
-      FileHandler().loadExperiment(spectra_file, exp, FileTypes::UNKNOWN, ProgressLogger::NONE,
-                                   false, false);
-      readSpectra(exp.getSpectra());
-      setSpectraDataRef(File::absolutePath(spectra_file));
-    }
 
     /**
      * @brief set spectra_data from read SpectrumContainer origin (i.e. filename)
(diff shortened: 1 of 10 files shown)


src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -155,6 +157,8 @@ namespace OpenMS
     static const MetaDataFlags MDF_NATIVEID = 64;
     static const MetaDataFlags MDF_ALL = 127;
 
+    String spectra_data;
```
timosachsenberg: make protected, otherwise setter is redundant

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -177,15 +181,17 @@ namespace OpenMS
     };
 
     /// Constructor
-    SpectrumMetaDataLookup(): SpectrumLookup() {}
+    SpectrumMetaDataLookup(): SpectrumLookup(),
```
timosachsenberg: newline after SpectrumLookup(), spectra_data is default initialized and needs no explicit construction here

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -218,6 +224,30 @@ namespace OpenMS
     }
 
     /**
+     * @brief encapsulates readSpectra with file loading convenience if only a file path is given
+     * @param spectra_file file to the SpectrumContainer, must be valid path to mz file
+     */
+    void readMzFile(const String& spectra_file)
```
timosachsenberg: name suggests that am MzML file is expected but the function is written for generic spectra files

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -218,6 +224,30 @@ namespace OpenMS
...
+    {
+      PeakMap exp;
+      FileHandler().loadExperiment(spectra_file, exp, FileTypes::UNKNOWN, ProgressLogger::NONE,
+                                   false, false);
+      this->readSpectra(exp.getSpectra());
```
timosachsenberg: we usually don't write the this-> for simple member functions

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -218,6 +224,30 @@ namespace OpenMS
...
+      this->setSpectraData(File::absolutePath(spectra_file));
+    }
+
+    /**
+     * @brief set spectra_data from read SpectrumContainer origin (i.e. filename)
+     *
+     * @param spectra_data the name (and path) of the origin of the read SpectrumContainer
+     */
+    void setSpectraData(const String& spectra_data)
```
timosachsenberg: this name is confusing - particularly in the presence of the readSpectra() function. Better one?

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -264,13 +294,27 @@ namespace OpenMS
        @return True if all peptide IDs could be annotated successfully (including if all already had RT values), false otherwise.
 
        Look-up works by matching the "spectrum_reference" (meta value) of a peptide ID to the native ID of a spectrum. Only peptide IDs without RT (where PeptideIdentification::getRT() returns "NaN") are looked up; the RT is set to that of the corresponding spectrum.
-
-       The raw data is only loaded from @p filename if necessary, i.e. if there are any peptide IDs with missing RTs.
     */
     static bool addMissingRTsToPeptideIDs(
-      std::vector<PeptideIdentification>& peptides, const String& filename,
+      std::vector<PeptideIdentification>& peptides, const SpectrumMetaDataLookup& lookup,
```
timosachsenberg: please correct documentation

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -264,13 +294,27 @@ namespace OpenMS
...
       bool stop_on_error = false);
 
+    /**
+     * @brief Add missing spectrum_references to peptide identifications vased on raw data
+     *
+     * @param peptides Peptide IDs with or without spectrum_reference
+     * @param lookup SpectrumMetaDataLookup with loaded spectra, metadata and optional spectra_data value for Proteins
+     * @param stop_on_error Stop when an ID could not be matched to a spectrum (or keep going)?
+     * @param override_spectra_data if given ProteinIdentifications should be updated with new spectra_data values from SpectrumMetaDataLookup
+     * @param proteins
```
timosachsenberg: add proteins docu

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -264,13 +294,27 @@ namespace OpenMS
...
+     *
+     * @return True if all peptide IDs could be annotated successfully (including if all already had spectrum_referece values), false otherwise.
```
timosachsenberg: typo referece

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -264,13 +294,27 @@ namespace OpenMS
...
+     *
+     * Look-up works by matching RT of a peptide ID to the native ID of a spectrum. All spectrum_references are updated/added.
```
timosachsenberg: is this correct? match retention time against native ID?

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -1280,7 +1280,7 @@ namespace OpenMS
               {
                 pep_id_->push_back(PeptideIdentification());
                 pep_id_->back().setHigherScoreBetter(false); //either a q-value or an e-value, only if neither available there will be another
-                pep_id_->back().setMetaValue("spectrum_reference", spectrumID); // TODO @mths consider SpectrumIDFormat to get just a index number here
+                pep_id_->back().setMetaValue("spectrum_reference", spectrumID);  // as spectrumID is taken from the mz_file and varies widely from vendor to vendor, spectrum_reference as string will reference the spectrum internally
```
timosachsenberg: the comment is a bit misleading. is says that spectrumID varies a lot, thus, we take the spectrum_reference. But in the actual code you set the spectrumID.

src/openms/source/METADATA/SpectrumMetaDataLookup.cpp
```diff
@@ -181,38 +180,83 @@ namespace OpenMS
...
-          LOG_ERROR << "Error: Failed to look up retention time for peptide ID with spectrum reference '" + spectrum_id + "' - no spectrum with corresponding native ID found." << endl;
+          LOG_ERROR << "Error: Failed to look up spectrum native ID for peptide ID with retention time '" + String(it->getRT()) + "'." << endl;
           success = false;
           if (stop_on_error) break;
         }
       }
     }
+    else
+    {
+      LOG_ERROR << "Error: No lookup to find spectrum_references from is available. Is mz_file lookup given?" << endl;
```
timosachsenberg: Better error message?

src/topp/IDFileConverter.cpp
```diff
@@ -170,6 +170,7 @@ class TOPPIDFileConverter :
     registerFlag_("ignore_proteins_per_peptide", "[Sequest only] Workaround to deal with .out files that contain e.g. \"+1\" in references column,\n"
                                                  "but do not list extra references in subsequent lines (try -debug 3 or 4)", true);
     registerStringOption_("scan_regex", "<expression>", "", "[Mascot, pepXML, Percolator only] Regular expression used to extract the scan number or retention time. See documentation for details.", false, true);
+    registerFlag_("no_spectra_data_override", "[+mz_file only] Will switch off spectra_data override in ProteinIdentifications if mz_file is given and spectrum_references are added/updated. Switch off only if you are sure it is absolutely the same mz_file as used for identification.", true);
```
timosachsenberg: better description. What does "switch off" refer to ? the flag or the negated option?

src/topp/IDFileConverter.cpp
```diff
@@ -584,6 +569,12 @@ class TOPPIDFileConverter :
     return EXECUTION_OK;
   }
 
+  void add_spectrum_reference_(std::vector<ProteinIdentification>& protein_identifications,
```
timosachsenberg: hmm might lead to style violation: better use addSpectrumReference

src/topp/IDFileConverter.cpp
```diff
@@ -584,6 +569,12 @@ class TOPPIDFileConverter :
...
+                               std::vector<PeptideIdentification>& peptide_identifications,
+                               SpectrumMetaDataLookup& lookup,
+                               bool override_spectra_data)
+  {
```
timosachsenberg: nothing here?

hendrikweisser:
@mwalzer: Can we meet to review this code together? I don't understand some of the changes you made, but don't want to leave an overly negative review because of this.

src/topp/IDFileConverter.cpp
mwalzer: Switch off, because I phrased it just like that the sentence before.

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
mwalzer: just because mz? I disagree here. Also, the input param is named ```registerInputFile_("mz_file"```

src/topp/IDFileConverter.cpp
mwalzer: But I will rewrite the description for this advanced parameter.

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
mwalzer: Yes, but the ProteinIdentification meta param is called spectra_data ... , maybe setProteinIDSpectraData? Or go by the intention, as spectra_data is suboptimal, too: setSpectraDataProvenance?

src/topp/IDFileConverter.cpp
hendrikweisser: @mwalzer: Please make sure you write this documentation as intended for the end user, not for developers. That means use general terms, not class/variable names ("spectra_data", "ProteinIdentification", "spectrum_references"). This documentation must be understandable without knowing any of the code.

mwalzer: spectra_data and spectrum_references are meta params, which means they are inspectable by looking in the file _and_ they are "explicitly" not required for a valid file but might be essential for building a functioning workflow. And I think we have to distinguish between the regular user and the advanced user, who is likely to look into the files and the advanced parameter set of tools.

src/topp/IDFileConverter.cpp
```diff
@@ -159,7 +159,7 @@ class TOPPIDFileConverter :
     setValidStrings_("out_type", ListUtils::create<String>(formats));
 
     addEmptyLine_();
-    registerInputFile_("mz_file", "<file>", "", "[pepXML, Sequest, Mascot, X! Tandem, Percolator only] Retention times will be looked up in this file", false);
+    registerInputFile_("mz_file", "<file>", "", "[pepXML, Sequest, Mascot, X! Tandem, MS-GF+, Percolator only] Retention times and native spectrum ids (spectrum_references) will be looked up in this file", false);
```
hendrikweisser: mzid instead of MS-GF+

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
hendrikweisser: As discussed, roll back changes.

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
hendrikweisser: Pass file path via `readSpectra`. Remove this function?

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -264,13 +294,27 @@ namespace OpenMS
...
-
-       The raw data is only loaded from @p filename if necessary, i.e. if there are any peptide IDs with missing RTs.
     */
     static bool addMissingRTsToPeptideIDs(
-      std::vector<PeptideIdentification>& peptides, const String& filename,
+      std::vector<PeptideIdentification>& peptides, const SpectrumMetaDataLookup& lookup,
       bool stop_on_error = false);
 
+    /**
+     * @brief Add missing spectrum_references to peptide identifications vased on raw data
```
hendrikweisser: "spectrum references"

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
hendrikweisser: The nomenclature is a bit misleading. "spectrumID" (e.g. in mzid) refers to the spectrum reference.

src/openms/source/METADATA/SpectrumMetaDataLookup.cpp
```diff
@@ -181,38 +180,83 @@ namespace OpenMS
   }
 
 
-  bool SpectrumMetaDataLookup::addMissingRTsToPeptideIDs(
-    vector<PeptideIdentification>& peptides, const String& filename,
+  bool SpectrumMetaDataLookup::addMissingRTsToPeptideIDs(vector<PeptideIdentification>& peptides, const SpectrumMetaDataLookup& lookup,
```
hendrikweisser: Roll back changes (see above).

src/topp/IDFileConverter.cpp
```diff
@@ -329,6 +330,20 @@ class TOPPIDFileConverter :
       else if (in_type == FileTypes::IDXML)
       {
         IdXMLFile().load(in, protein_identifications, peptide_identifications);
+        // get spectrum_references from the mz data, if necessary:
+        if (!mz_file.empty())
```
hendrikweisser: Can this be moved into `SpectrumMetaDataLookup::addMissingSpectrumReferences`?

src/topp/IDFileConverter.cpp
```diff
@@ -337,11 +352,19 @@ class TOPPIDFileConverter :
         MzIdentMLFile().load(in, protein_identifications,
                              peptide_identifications);
 
-        // get retention times from the raw data, if necessary:
+        // get retention times from the mz data, if necessary:
         if (!mz_file.empty())
         {
+          if (lookup.empty())
```
hendrikweisser: Roll back with changes in `SpectrumMetaDataLookup::addMissingRTsToPeptideIDs`.

src/topp/MSGFPlusAdapter.cpp
```diff
@@ -731,7 +731,11 @@ class MSGFPlusAdapter :
         {
           switchScores_(*pep_it);
         }
-        SpectrumMetaDataLookup::addMissingRTsToPeptideIDs(peptide_ids, in, false);
+
```
hendrikweisser: Roll back (see above).

hendrikweisser:
@mwalzer: Following from our discussion, I just thought about something: If the function `addMissingSpectrumReferences` will be generally used to add spectrum references and "spectra_data" information, it may be enough to pass the MS data filename only there, and not in `readSpectra`.

src/topp/IDFileConverter.cpp
mwalzer: Only if mz_file is given, spectrum_reference annotation will be undertaken, analogue to how addMissingRTsToPeptideIDs is invoked if FileType is mzid. ok?

src/topp/IDFileConverter.cpp
mwalzer: yep, cleaned it up.

src/topp/IDFileConverter.cpp
mwalzer: yep.

src/topp/IDFileConverter.cpp
mwalzer: Should be clearer now.

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
mwalzer: cleared

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
mwalzer: BTW, does the old one support OnDiscMSExperiment?

src/topp/IDFileConverter.cpp
hendrikweisser: I think so. The spectrum references would need to be read from the `mz_file`, correct?

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -225,25 +223,26 @@ namespace OpenMS
 
     /**
      * @brief encapsulates readSpectra with file loading convenience if only a file path is given
+     *
      * @param spectra_file file to the SpectrumContainer, must be valid path to mz file
      */
-    void readMzFile(const String& spectra_file)
+    void readMzFileMetaData(const String& spectra_file)
```
hendrikweisser: Is this function used anywhere (besides a test)? Functionality is already included in `addMissingSpectrumReferences`, so could the function be removed?

mwalzer: I thought that maybe spectrumreference annotation is not the only usecase where you might want to have the SpectrumLookup object already setup with the metadata and the spectrum reference prior...

src/topp/IDFileConverter.cpp
mwalzer: yeah.

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
timosachsenberg: not sure. anyways please move implementation to cpp file

timosachsenberg:
Some of the reviewer comments were not addressed so far but are now hidden (show outdated).
Please consider going through the comments again and check if they have been addressed

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
hendrikweisser: @mwalzer: I think that is the only use case. Unless I really overlooked something, `readMzFileMetaData`, `setSpectraDataRef` and the `spectra_data_ref` member can all be removed. "spectra_data" is set in `addMissingSpectrumReferences` based on the filename that is passed to that function.

timosachsenberg: So, what is the conclusion regarding this function?

timosachsenberg:
@mwalzer and @hendrikweisser c++ part looks pretty much finished to me. Can you flesh out the last points of discussion and provide tests and updated pxd files?

src/topp/IDFileConverter.cpp
mwalzer: Coming back to this: I think the comment got pinned on the `if (...)` but was meant for the content/implications of that condition, which are now analogous to how `addMissingRTs...` is handled. 
Can we close the review?

hendrikweisser: The case here is settled, yes, but we still had a discussion about `SpectrumMetaDataLookup::readMzFileMetaData` which wasn't fully resolved. See: https://github.com/OpenMS/OpenMS/pull/2498#discussion_r111946179

hendrikweisser:
I think the last open question is this one: https://github.com/OpenMS/OpenMS/pull/2498#discussion_r111946179

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
mwalzer: gone.

mwalzer:
which is now addressed

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -264,16 +280,30 @@ namespace OpenMS
...
-    static bool addMissingRTsToPeptideIDs(
-      std::vector<PeptideIdentification>& peptides, const String& filename,
+    static bool addMissingRTsToPeptideIDs(std::vector<PeptideIdentification>& peptides, const String &filename,
       bool stop_on_error = false);
 
+    /**
+     * @brief Add missing "spectrum_reference"s to peptide identifications based on raw data
+     *
+     * @param peptides Peptide IDs with or without spectrum_reference
+     * @param lookup SpectrumMetaDataLookup with loaded spectra, metadata and optional spectra_data value for Proteins
```
timosachsenberg: for Proteins? " to be annotated in (all I assume???) ProteinIdentification" objects

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
```diff
@@ -264,16 +280,30 @@ namespace OpenMS
...
+     * @param stop_on_error Stop when an ID could not be matched to a spectrum (or keep going)?
+     * @param override_spectra_data if given ProteinIdentifications should be updated with new "spectra_data" values from SpectrumMetaDataLookup
+     * @param proteins
+     *
+     * @return True if all peptide IDs could be annotated successfully (including if all already had "spectrum_reference" values), false otherwise.
+     *
+     * Look-up works by matching RT of a peptide ID to the native ID of a spectrum. All spectrum_references are updated/added.
```
timosachsenberg: why are RTs of an identification matched to the native ID of a spectrum? is matching done between RTs? not clear how it works from the comment. or is RT wrong? also do peptide ID and native ID mean something else here? Identification and Identifier? if so it might be confusing (applies to other places in comments as well)

src/openms/source/METADATA/SpectrumMetaDataLookup.cpp
```diff
@@ -215,4 +213,51 @@ namespace OpenMS
...
+      try
+      {
+        Size index = lookup.findByRT(it->getRT());
+        SpectrumMetaDataLookup::SpectrumMetaData meta;
+        lookup.getSpectrumMetaData(index, meta);
+        it->setMetaValue("spectrum_reference", meta.native_id);
+      }
+      catch (Exception::ElementNotFound&)
+      {
+        LOG_ERROR << "Error: Failed to look up spectrum native ID for peptide ID with retention time '" + String(it->getRT()) + "'." << endl;
```
timosachsenberg: same here ID means two different things

src/tests/class_tests/openms/source/SpectrumMetaDataLookup_test.cpp
```diff
@@ -168,7 +168,7 @@ START_SECTION((void getSpectrumMetaData(const String&, SpectrumMetaData&, MetaDa
 END_SECTION
 
 
-START_SECTION((bool addMissingRTsToPeptideIDs(vector<PeptideIdentification>&, const String&, bool)))
+START_SECTION((bool addMissingRTsToPeptideIDs(vector<PeptideIdentification>& peptides, const String& filename, bool stop_on_error)))
```
timosachsenberg: signature complete?

timosachsenberg:
looks good. only minor comments regarding the documentation

src/openms/source/METADATA/SpectrumMetaDataLookup.cpp
mwalzer: I think I took that over from SpectrumMetaDataLookup::addMissingSpectrumReferences. Clarified in both places.

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
mwalzer: No, it is just simple as you would imagine. RT gets matched, 'native_id' looked up, annotated to identification.

src/openms/include/OpenMS/METADATA/SpectrumMetaDataLookup.h
mwalzer: Removed deprecated documentation (no nore lookup objects in the call)

timosachsenberg:
+1

PR #2500, opened 2017-04-04
Fix/win compile

timosachsenberg:
some more compile fixes and some conversion warnings removed


cbielow:
could you please give a pointer to the CI build log where this occurred. thx.

timosachsenberg:
was on my vs 2013

cbielow:
ah ok. looks good then.

PR #2502, opened 2017-04-05
Fixes for X! Tandem and Percolator (modification handling)

hendrikweisser:
This addresses problems in XTandemAdapter and with the reading of Percolator result files (derived from X! Tandem searches).

In XTandemAdapter, support for X! Tandem's default N-terminal mods (pyro-Glus and acetylation), which are covered by special parameters, was improved. When reading results, these mods are now added to the search parameters when found. When running searches, they should now be handled correctly when set by the user (in particular, there's an annoying interaction with other variable N-terminal mods).

In the Percolator results, the problem was that N-terminal modifications got attributed to the first residue (the notation in the X! Tandem output is ambiguous). The information about where the mod belongs is essentially lost at that point, so it takes a lot of heuristics to (hopefully) recover it - see new function `PercolatorOutfile::resolveMisassignedNTermMods_`.

Also fixed an oversight in `ModificationDefinitionsSet::addMatches_`, which was due to the idiotic use of "C-term" and "N-term" as origin (i.e. modified residue) of terminal modifications. (I hope to address this problem generally in a future PR.)

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ResidueModification.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ResidueModification.h
@@ -80,7 +80,6 @@ public:
 
             This does not describe the amino acids which are valid for a
             specific amino acid!
-
     */
     enum TermSpecificity
     {
@@ -93,7 +92,6 @@ public:
     };
 
     /** @brief Classification of the modification
-
     */
     enum SourceClassification
     {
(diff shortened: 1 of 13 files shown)


timosachsenberg:
nice! I will probably take a closer look tomorrow or the day after tomorrow.

timosachsenberg:
looks good. Thanks for figuring out the strange logic. Do you have a test we could add?

timosachsenberg:
I merge to get this forward but it would be great to have a test that confirms we don't break this logic after supporting e.g. to prot-term

PR #2503, opened 2017-04-06
Feature/fraction support

timosachsenberg:
This PR is not intended for merge but to gather feedback quickly. It is currently not tested and for discussion only. It is based on results from our developer retreat so I will not provide all information here (feel free to ask if unsure).
The idea is to be able to provide a simple tsv file that defines the run-level and assay-level experimental design. By adding additional optional columns in the future most metadata contained in the mzTab standard can be covered.
In this first step, only a run-level design is provided, which allows defining fractions and technical replicates (not implemented).

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
+++ b/src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
@@ -38,6 +38,7 @@
 #include <OpenMS/CONCEPT/Macros.h>
 #include <OpenMS/KERNEL/MRMFeature.h>
 #include <boost/numeric/conversion/cast.hpp>
+#include <OpenMS/KERNEL/MSChromatogram.h>
 
 namespace OpenMS
 {
--- a/src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
+++ b/src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
@@ -1,0 +1,100 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2016.
+//
+// This software is released under a three-clause BSD license:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of any author or any participating institution
+//    may be used to endorse or promote products derived from this software
+//    without specific prior written permission.
+// For a full list of authors, refer to the file AUTHORS.
+// --------------------------------------------------------------------------
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL ANY OF THE AUTHORS OR THE CONTRIBUTING
+// INSTITUTIONS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer:	Timo Sachsenberg $
+// $Authors: Timo Sachsenberg $
+// --------------------------------------------------------------------------
+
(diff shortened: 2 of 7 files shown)


timosachsenberg:
```
/// Run	Spectra File	Fraction	Technical Replicate   
///   1	humanA.mzML	      1	              1          
///   2	humanB.mzML	      1	              1          
///   3	humanC.mzML	      2	              1              
///   4	humanD.mzML	      2	              1
///   5	humanE.mzML	      3	              1
///   6	humanF.mzML	      3	              1          
///   7	humanF.mzML	      3	              2       (<- example how a 2nd technical replicate is stored for fraction 3)
```

hendrikweisser:
> Example of run-level design

Shouldn't the filename (column "Spectra File") for run 7 be different from run 6?

timosachsenberg:
right, thanks! copy & paste error

hroest:
I would add a level of biological group as well: to represent patients / biological replicates etc. So there are some labels I would like to add, e.g. one for "Subject Id" and maybe one for "Group Id" which would also allow us to have multiple biological replicates (multiple subjects) and multiple experimental groups (e.g. randomized interventions for each biological sample). 

Imagine an experiment where we have different patients under different conditions (e.g. drugs or interventions) and then we would need to be able to represent patients and also conditions. One could also imagine different biological plates or flasks as "subjects" (biological replicates) and different conditions/treatments applied. 

I would model some of this according to what MSStats / SRMStats is doing - that would also make it easier to directly integrate with them. Also I think Skyline has an SRMStats integration since recently, maybe we should look at how to specify the UI - basically here we are talking about a UI to enter experimental design.

hroest:
the second question is how do we map to the original mzXML / raw file and what value "Spectra file" actually should contain. E.g. is this the name of the centroided file, the name of the original raw file as it came of the mass spec? I think a permanent identifier would be great here and I think we / pwiz stores the name of the raw file before conversion so linking to that would make a lot of sense as this would be recorded in the mzML file somewhere

timosachsenberg:
@hroest thanks for your input. I think these need to be handled on the assay level and not on the run level. Or did you already refer to the assay level?
see https://docs.google.com/spreadsheets/d/10MLI_Wy2OLjOQXdKd7448z4Vn-ZyYE2nYwokgTXlUEs/edit?usp=sharing
for an example with an assay section. This concept allows representing both multiplexed and non-multiplexed designs. Factors can be used to group assays (e.g., quantification channels from multiplexed experiments)

timosachsenberg:
The .raw file is probably not a stable identifier and often not present anymore. I think it is best to refer to the first mzML file that enters the workflow. I think the current implementation is that this file name is passed through our internal structures.

hroest:
@timosachsenberg yes, the raw file may not be present but it may still be a unique identifier. Generally people tend to put a lot of thought into the naming of raw files (at least from my experience). But in most cases people use name_of_sample.[raw|mzML] anyways ....

hendrikweisser:
> Generally people tend to put a lot of thought into the naming of raw files (at least from my experience).

Yes, but we would expect the mzML file after conversion to have the same basename, no?

timosachsenberg:
@hroest just to clarify. We don't want to tell people how they should name the raw and converted mzML file.
The idea is that OpenMS workflows start with mzML files. These are the ones referenced in the experimental design. This will allow us to, for example, provide a simple TOPP Tool (ExperimentalDesignGenerator) that takes an in-file list of mzMLs and outputs a tsv file with run-level information where the user can assign the fractions or technical replicates. 
Alternatively, we can provide a simple KNIME node that creates a dynamic table based on the in-files and writes it in the tsv.
This can probably be extended so that a template for assay-level information is generated. For now, we try to solve the fractionated data use-case first (but we thought about how to combine this with groups on the assay level).
I think the last details for the assay part needs to be fleshed out and I would love to discuss this at some point.

mwalzer:
@timosachsenberg as you wrote adding some other columns might be necessary. I think a studyvariable column might be good here. Mapping to multiple SV might be good as well, as this is often the case for some studies design.

mwalzer:
Regarding the identifier of the MZ data file, I agree with @hroest most of the way. From a data provenance proof point of view, I think it would be good practice if we built in some measures to keep some link to the vendor file. And this kind of 'data-injection' as in this PR might just be the right place, no? Like an additional column 'vendor file' to straighten any missing links out. Then, I'd suggest that the spectral file column should contain the file names as they are actually read in, preferentially before processing (lol), to make the annotations and keep during further processing.

timosachsenberg:
@mwalzer regarding SV. This is part of the assay-level and not part of the run level. We can talk about this in more detail at the PSI meeting so I can explain why we use factors instead of study variables here. The differences are subtle but relate to at which point you know what and how to group.

timosachsenberg:
Regarding the comment about the vendor file. Maybe I misunderstood your comment but it sounds like you would like to have a mandatory column called "vendor file". I think we can include something like this but I would strongly suggest making it optional. As @hroest points out, this is intended as a "GUI" style definition of the experimental design. Additional (optional) columns will help make it compatible to mzTab / ISA-tab or similar but all metadata associated with them is not what we want our users to enter manually. For complete evidence tracing / data provenance issues I suggest to use a LIMS that supports exporting this file from experiment metadata.

hroest:
my point was simply on how to make it easiest for experimentalist to (i) keep track of their data and mapping file names to study variables and (ii) to add this to OpenMS. From what I have seen is that most people either keep a file around where they map the raw file names (either with or without the `.raw` ending) to some study variables. Maybe a more flexible solution would be to accept `unique-name[.][mzML|raw|wiff]` so that people have options but with the default that you proposed being `unique-name.mzML`

again -- maybe this discussion is a bit too advanced as it tries to solve the problem more generally?

mwalzer:
In some cases it will not be possible to supply a vendor file, so making it mandatory will not work. I think the majority of labs do _not_ have a LIMS system nor the (tech.) capability to run one. And they convert their vendor files manually. So I think it would be reasonable to give that a chance here. Regarding your reluctance about study variables, please do explain to me what you guys discussed at dev retreat (best probably at PSI meeting).

timosachsenberg:
Great! Looking forward to it. And I wouldn't call it reluctance and we might also consider renaming the factors to study variable. But at this point I am mainly concerned with the fractions.

cbielow:
are you aware of https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4411450/
? This scheme might help?

timosachsenberg:
Sure. I was actually in close contact with Andreas when they created it for QBIC. It's very generic and in that regards tries to address solve similar problems as ISA-Tab. One downside is, that its XML-based and representing multiplexed data in the format is not that intuitive. We actually figured that out when we tried to model a TMT6Plex experiment in the format. For them, it's not a real problem as they use a GUI to construct it but I would love to make these designs easily accessible in a tabular separated file.

PR #2478, opened 2017-03-24
Feature/comet adapter

Leon-Bichmann:
First version - open for changes

PR #2504, opened 2017-04-06
ResidueModification: clean-up of "origin"

hendrikweisser:
This resolves a long-standing annoyance in our class representing PTMs, ResidueModification: The "origin" (modified residue) of terminal modifications was set to "C-term" or "N-term" (like in Unimod) - but only if the modification wasn't also residue-specific. This led to a number of oversights in checks and queries.

The origin of a modification is now stored as a `char`, and only valid one-letter codes and "X" are allowed. For modifications with purely terminal specificity, "X" is used to represent "any amino acid". (Like in the PSI-MOD ontology.)

Going forward, `ResidueModification::getOrigin` should be used strictly to query the affected residue and `ResidueModification::getTermSpecificity` to check for terminal specificity. This should hopefully avoid some headaches in the future.

There is still an unresolved issue in `ModificationsDB::readFromOBOFile`: Modifications with purely terminal specificity were previously not added to the database (due to a bug). However, adding them now leads to undesired behaviour and failing tests. The culprit (or one of them?) is "MOD:01154" (pyruvic acid), which is N-terminal and not residue-specific, but doesn't have mass difference information in "PSI-MOD.obo". To avoid problems I'm still excluding all modifications without residue specificity in that function (but not in the equivalent `CrossLinksDB::readFromOBOFile`!).


hendrikweisser:
@enetz: I've found (and removed) some code in `ModificationsDB::readFromOBOFile` that was intended for reading cross-links from "XLMOD.obo", but not applicable to "PSI-MOD.obo". Since there is now a separate function in CrossLinksDB, can you check if there is any more code in ModificationsDB that could be removed?

src/openms/source/CHEMISTRY/ModificationsDB.cpp
```diff
@@ -344,26 +323,13 @@ namespace OpenMS
 
           for (vector<String>::iterator orig_it = origins.begin(); orig_it != origins.end(); ++orig_it)
           {
-            if (orig_it->size() == 1)
+            if ((orig_it->size() == 1) && (*orig_it != "B") && (*orig_it != "J") && (*orig_it != "Z"))
```
timosachsenberg: do we handle these cases properly? (also above and below?)

src/openms/source/CHEMISTRY/ModificationsDB.cpp
```diff
@@ -571,8 +524,11 @@ namespace OpenMS
       {
         // the mod has so far not been mapped to a unimod mod
         // first check whether the mod is specific
-        if ( (it->second.getOrigin().size() == 1 && it->second.getOrigin() != "X") ||
-              (it->second.getOrigin() == "N-term") || (it->second.getOrigin() == "C-term"))
+        // @TODO: allow terminal specificity (currently breaks various tests,
```
timosachsenberg: so is this a regression?

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -2497,8 +2497,9 @@ namespace OpenMS
           DOMElement* current_cv = current_pep->getOwnerDocument()->createElement(XMLString::transcode("cvParam"));
           current_mod->setAttribute(XMLString::transcode("location"), XMLString::transcode("0"));
           current_mod->setAttribute(XMLString::transcode("monoisotopicMassDelta"), XMLString::transcode(String(mod.getDiffMonoMass()).c_str()));
-          current_mod->setAttribute(XMLString::transcode("residues"), XMLString::transcode(mod.getOrigin().c_str()));
-
+          String origin = mod.getOrigin();
```
timosachsenberg: Maybe add an explicit cast to String: String(mod.getOrigin()) ?

timosachsenberg: also below

src/openms/source/FORMAT/HANDLERS/UnimodXMLHandler.cpp
```diff
@@ -102,44 +102,33 @@ namespace OpenMS
...
-          }
+          warning(LOAD, String("Don't know allowed position called: '") + pos  + "' - setting to anywhere");
         }
 
         if (!pos.hasSubstring("Protein"))
         {
           was_valid_peptide_modification_ = true;
           term_specs_.push_back(position);
-          sites_.push_back(site);
+          if (site.size() > 1) site = "X"; // C-term/N-term
```
timosachsenberg: if it is not performance critical I would check for the substring "-term"

src/tests/class_tests/openms/source/CrossLinksDB_test.cpp
```diff
@@ -47,6 +47,10 @@ struct ResidueModificationOriginCmp
 {
   bool operator() (const ResidueModification* a, const ResidueModification* b)
```
timosachsenberg: I don't know if this operator is correct. Shouldn't it first test if pointers are equal and then if this is the same modification?

src/tests/topp/CMakeLists.txt
```diff
@@ -727,18 +727,14 @@ set_tests_properties("TOPP_IDConflictResolver_2_out1" PROPERTIES DEPENDS "TOPP_I
...
 add_test("TOPP_IDFileConverter_2" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/PepXMLFile_test.pepxml -mz_file ${DATA_DIR_TOPP}/PepXMLFile_test.mzML -mz_name PepXMLFile_test -out IDFileConverter_2_output.tmp -out_type idXML)
 add_test("TOPP_IDFileConverter_2_out1" ${DIFF} -in1 IDFileConverter_2_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_2_output.idXML )
 set_tests_properties("TOPP_IDFileConverter_2_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_2")
 # protXML to idXML:
 add_test("TOPP_IDFileConverter_3" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/IDFileConverter_3_input.protXML -out IDFileConverter_3_output.tmp -out_type idXML)
 add_test("TOPP_IDFileConverter_3_out1" ${DIFF} -in1 IDFileConverter_3_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_3_output.idXML )
 set_tests_properties("TOPP_IDFileConverter_3_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_3")
-# pepXML to idXML (using MS2 RTs):
-add_test("TOPP_IDFileConverter_4" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/PepXMLFile_test.pepxml -mz_file ${DATA_DIR_TOPP}/PepXMLFile_test.mzML -mz_name PepXMLFile_test -out IDFileConverter_4_output.tmp -out_type idXML)
```
timosachsenberg: what happened to him?

enetz:
"XLMOD.obo" contains two types of modifications, the Cross-Links, that link two peptides, and the corresponding Mono-Links, that happen when the second functional group is not bound to a peptide.
Since the Mono-Links can be treated as normal modifications, I included them into the ModificationsDB.
Which type it is can be determined by the attribute "reactionSites" in the "XLMOD.obo", "1" for Mono-Links and "2" for Cross-Links. I checked for that and stored that info in the variable "reading_cross_link" defined as "reactionSites" == "2".
The main difference between ModificationsDB::readFromOBOFile and CrossLinksDB::readFromOBOFile,
is that CrossLinksDB includes only cases with "reading_cross_link" = true and ModsDB only cases with "reading_cross_link" = false.

There may be a more elegant way to do this, but the "PSI-MOD.obo" and the "XLMOD.obo" have a slightly different format, e.g. the terminal modification strings are different, so some of the Cross-Link specific code is needed for the Mono-Links.
I would like to keep the Mono-Links in the ModificationsDB, if they do not make any problems, so could you revert those changes, or rather correct the .setOrigin() lines in my code? As it looks right now, I think the Mono-Links are read, but the terminal specificities are ignored if present. Seems like I forgot to write tests for the Mono-Links in the ModsDB tests. I will do that soon.

src/openms/source/CHEMISTRY/ModificationsDB.cpp
hendrikweisser: There is only one case where it applies: "MOD:00306" which has origin "J" (represents "I or L", so isn't really a modification anyway). We used to read it, but I assume it never got used (why look for modifications of residue J?). Now it's skipped.

src/openms/source/CHEMISTRY/ModificationsDB.cpp
hendrikweisser: No, it was broken before. There were tests for "origin = C-term" and "origin = N-term" in the condition, but those values were never set (except for Eugen's handling of cross-links).

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
hendrikweisser: Why? There's a definition for `string::operator=(char)`. Does it give a warning on your system?

src/tests/class_tests/openms/source/CrossLinksDB_test.cpp
hendrikweisser: That's just for the test (you want to make sure the order is reproducible between systems).

src/tests/topp/CMakeLists.txt
hendrikweisser: The "IDFileConverter_2" and "IDFileConverter_4" tests used to test conversion with and without a flag ("use_precursor_rt"?). That flag has since been removed, so the tests were identical.

src/openms/source/CHEMISTRY/ModificationsDB.cpp
timosachsenberg: I see, might be hard to track down the reason for others so maybe add a comment ?

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
timosachsenberg: Was just an idea to make the conversion explicit.

timosachsenberg:
I think everything is settled from my side.

src/openms/source/FORMAT/HANDLERS/UnimodXMLHandler.cpp
hendrikweisser: I was about to agree, but since I use `site[0]` in the next line, I think it's better to treat anything unexpected as "X". Otherwise I'd have to check for things that are neither residues nor "C/N-term" and throw an error.

hendrikweisser:
@timosachsenberg: Thanks for the review!

@enetz: Thanks for the explanation. There are two issues that I have with this: 1. There's a huge amount of duplicated code between the two `readFromOBOFile` functions, which makes it hard to see where the differences are and to make consistent changes. 2. Your code for handling mono-links in XLMOD.obo creates terminal modifications without residue specificity, which I'm currently excluding:
> Modifications with purely terminal specificity were previously not added to the database (due to a bug). However, adding them now leads to undesired behaviour and failing tests. The culprit (or one of them?) is "MOD:01154" (pyruvic acid), which is N-terminal and not residue-specific, but doesn't have mass difference information in "PSI-MOD.obo". To avoid problems I'm still excluding all modifications without residue specificity in that function (but not in the equivalent CrossLinksDB::readFromOBOFile!).

One possible solution would be to find out which of those cases are actually causing problems (those without mass difference information?) and exclude only those. Before I do that, I would however like to understand the alternatives.

Is there a particular reason why you would like to include the mono-links in ModificationsDB? If you're searching cross-linking data, wouldn't you anyway search against CrossLinksDB? Would it perhaps lead to cleaner implementations if all the cross-linking modifications were handled by one function (in CrossLinksDB)?

enetz:
I want to have the Mono-Links in the ModificationsDB, because chemically they are no different from normal residue modifications and can therefore also be searched by normal ID algorithms. I think besides the slightly different format of the XLMOD.obo file, there is no need to treat them differently from the other modifications. 

The Cross-Links are a completely different type of modification and although most XL-MS ID algorithms as of now search for both at the same time, It is not at all obvious whether it makes sense to search for Mono-Links using an algorithm optimized specifically for Cross-Links.
Handling all these modifications in one function would not be cleaner, I would prefer to only have real Cross-Links in the CrossLinksDB instead of having to check whether a mod has 1 or 2 reactive sites each time.

enetz:
I can probably remove some of the duplicated code in CrossLinksDB, that is PSI-MOD.obo specific, since I only read the XLMOD.obo there. And I can write a separate function for the XLMOD.obo and clean up the code a bit this way, too.

hendrikweisser:
@enetz: Are there any use cases for searching cross-linking data with a "normal" search engine and trying to identify peptides with mono-links? What is your workflow for searching cross-linking data?

enetz:
The term specificity problem is something we need to solve another way, since mods that can appear on terminals without restrictions to specific residues are very common anyway.

It looks like MOD:01154 is something like a template for more specific mods, like MOD:00136 and MOD:00807, both of which have residue specificity and have the attribute
is_a: MOD:01154 ! pyruvic acid
This means there are already at least two mods with the same mass in the ModsDB.
Looks like this is true for most or all of the "Origin: "X"" mods in the PSI-MOD.obo, so maybe they really are not needed.

But do the MonoLinks with for example "(Protein N-Term)" as a specificity cause any problems in the tests,
if you add them as Origin = "X", TermSpec = PROTEIN_N_TERM?

hendrikweisser:
> The term specificity problem is something we need to solve another way, since mods that can appear on terminals without restrictions to specific residues are very common anyway.

Yes, they are common, but most  of them (all the relevant ones?) are already read from Unimod, so it's not a problem in practice.

> But do the MonoLinks with for example "(Protein N-Term)" as a specificity cause any problems in the tests, if you add them as Origin = "X", TermSpec = PROTEIN_N_TERM?

OpenMS generally doesn't support *protein* terminal specificities. I don't know what would happen if we added the mono-links with those specificities.

enetz:
As of now I search for the MonoLinks in my XL-MS ID tool, but that is mostly because it started out as a reimplementation of another XL-MS specific standalone tool and because we did not have an official database containing their masses in the beginning.
They have to be treated as a special case all along the way but in the end are handled as normal modifications in the output files and e.g. by TOPPView anyway.
I would like to outsource this to another search engine, maybe the SimpleSearchEngine, in the future, which would be much easier if they are included in the ModificationsDB.

enetz:
Then we could start supporting _protein_ terminal specificities right now, with the MonoLinks as a test case.
Since no one else is searching for them it should not break anything unrelated, and I can adapt my code to use Origin="X" and TermSpec=PROTEIN_N/C_TERM instead of just Origin="N/C-Term" as I have been doing previously.

hendrikweisser:
@enetz: Okay, I've made the changes that should enable the reading of mono-links again. The tests pass on my system, but since there is no specific test for this functionality yet, can you please check if it works for you?

By the way, using "PROTEIN_C/N_TERM" instead of "C/N_TERM" broke A LOT of tests, so that doesn't seem to be a good idea at this point.

timosachsenberg:
looks finished to me.
@enetz any final remarks. Otherwise feel free to merge it after conflicts are resolved and tests pass

enetz:
I checked and the mono-links are read correctly.
This works for me!

timosachsenberg:
@hendrikweisser maybe you can give advice to @Leon-Bichmann on how to handle the  if (residues=="C-term") and  if (residues=="N-term") 
after the merge (if required)
around:
https://github.com/OpenMS/OpenMS/pull/2478/files/869118d76163307d4886ecbc36414a753cd46498..76c632b3d3bd91099542bb2d3f6dd93abfaa62df#diff-be8f5c01aa5f30430bdbfca29811afe2R286

hendrikweisser:
Okay, I've resolved the merge conflicts. Strangely, I don't see the usual "CI tests are running" bit. Should I wait for that?

timosachsenberg:
I think we might have hit a limit today :) I'll check it on my computer and merge if it passes.

PR #2506, opened 2017-04-07
[HOTFIX] solves compiler errors in MRMTransitionGroup.

jpfeuffer:


jpfeuffer:
See CDash. Probably because of new instantiation in the cpp to overcome Windows export errors.

jpfeuffer:
I'll merge. Travis finally passes again and maybe this will make it into the nightlies.

PR #2507, opened 2017-04-08
XFDR util for the calculation of False Discovery Rates for cross-link identifications

lukaszimmermann:
### Abstract

This pull request contains an util to calculate FDR on cross-link identification as they are provided by OpenProXL or xQuest (2008, Rinner et al.).  The FDR computation is based on the target-decoy approach like it is implemented in xProphet (2012, Walzthoeni et al.).  The FDR is calculated on peptide level.

###  Input/Output

The tool consumes xquest.xml files or idXML files. The xquest.xml files can be either written by xQuest or OpenProXL/OpenProXLLF.  The idXML file might come from OpenProXL/OpenProXLLF.  As output, an idXML file can be specified. Writing and reading MzIdentML is not supported yet (but is about to be added soon). Currently there seem to be some smaller issues which I need to track down. 
In the resulting idXML file, each cross-linked peptide-pair is stored within an PeptideIdentification, containing one or two PeptideHits for the cross-linked peptides.
PeptideIdentifications are annoted with the the following additional meta values:
- OpenXQuest:xprophet_f: This is 1 if the Identification was used during FDR calculation, 0 otherwise. Name was chosen to be consistent with xProphet.
- OpenXQuest:fdr_type:  This is `qfdr` if q-value transformation was applied (default), otherwise `fdr`.
- OpenXQuest:fdr: The actual fdr value of the identification (example: 0.060475)

### Algorithm 
The algorithm works in roughly the following steps:
1.  For xQuest, delta scores and the minimum number of matched ions are computed. This can  be used as a filter for the IDs to be considered for FDR calculations. Currently not supported for idXML input, but filtering by delta scores for idXML will be added soon. 
2. For each spectrum, the ID with the highest scores are referenced in a vector (rank_one_ids)  and sorted in descending order according to the score.
3. Several filters are applied to the rank_one_ids (precursor_mass_error range, delta_score, min_ions_matched, min_score, unique_ids). Only the min_score is available for idXML input, the 
other filters currently require xquest.xml input.
4. For each remaining ID, the type of the cross-link is identified (target/decoy) X (inter/intra/mono/loop)
5. For each class, a complementary cumulative (tail)  histogram of the scores of the respective IDs is computed. That means, for each score cut-off, the number of IDs with higher scores is recorded; for all classes.
6. The FDRs are calculated separately for inter/intra/monolinks. The FP estimate is taken from xProphet. 
7. q-value transformation is applied if requested.
8. The FDRs are assigned back to the IDS.
9. Results are written.

### Further additions
Beside the XFDR util, this pull request also contains the following:

1. XML Schema for xQuest result files: xQuest_1_0.xsd
2. XQuestResultXMLHandler for handling xquest.xml files
3. XQuestResultXMLFile class for representing XQuest result files (Currently only loading is possible)
4. CumulativeHistogram class, which extends  the Histogram class. Normal and complementary mode are supported. Class tests were added accordingly. The Histogram class was extended by a constructor which populates the historgram based on the provided Data iterator
5. One example test for XFDR. 
6. XQuestResultMeta, for storing MetaInformation on an particular xQuest run. I am not sure whether this is useful, therefore these class is currently not explicitly used.

### TO-DO
1. Add support for reading and writing mzIdentML.
2. Enable all filters for all input file types.
3. Add more tests
4. Set qvalue/FDR as score for PeptideHit and set the score_type accordingly.

And of course, this util  will be the container for testing out other FDR calculation methods for xlink IDs.

Any feedback highly appreciated!


src/openms/include/OpenMS/MATH/STATISTICS/CumulativeHistogram.h
```diff
@@ -0,0 +1,179 @@
...
+        @brief Representation of a histogram, which counts the cumulative number of observations.
+        This can be either in normal or complementary mode.
+
+        The first template argument gives the Type of the
+        values that are stored in the bins. The second argument
+        gives the type for the bin size and range.
+
+        @ingroup Math
+    */
+    template <typename ValueType = UInt, typename BinSizeType = double>
```
timosachsenberg: I would actually just add a member getCummulativeHistogram to the orignal histogram class and maybe a member to increase all bin up to a value.

timosachsenberg: or maybe one could even reduce more code by using boost accumulators (see docu or http://programmingexamples.net/wiki/CPP/Boost/Histogram )

src/openms/include/OpenMS/FORMAT/HANDLERS/XQuestResultXMLHandler.h
```diff
@@ -0,0 +1,169 @@
...
+      // Docu in base class
+      //void characters(const XMLCh * const chars, const XMLSize_t /*length*/);
+
+      //Docu in base class
+      //virtual void writeTo(std::ostream & os);
+
+    private:
+   
+      // Main data structures that are populated during loading the file
+      std::vector< XQuestResultMeta > & metas_;
```
timosachsenberg: Just a side note: I think for a file handler it is fine to use references in members as it doesn't provide many possibilities for refactorings anyways. Otherwise, the use of & for member variables is discouraged as this introduces a (potentially unneeded) dependency inside a class.

src/openms/include/OpenMS/FORMAT/HANDLERS/XQuestResultXMLHandler.h
```diff
@@ -0,0 +1,169 @@
...
+      double rt_light_;
+
+      // The masses of the Monolinks
+      std::set< double > monolinks_masses_;
+      
+      // The current spectrum search
+      std::vector< PeptideIdentification > current_spectrum_search_;
+      XQuestResultMeta current_meta_; // The current meta value
+      int & n_hits_; // Total no. of hits found in the result XML file
+      std::vector< int > * cum_hits_;
```
timosachsenberg: why use a pointer here?

src/openms/include/OpenMS/FORMAT/HANDLERS/XQuestResultXMLHandler.h
```diff
@@ -0,0 +1,169 @@
...
+      // Current Retention time of light spectrum
+      double rt_light_;
+
+      // The masses of the Monolinks
+      std::set< double > monolinks_masses_;
+      
+      // The current spectrum search
+      std::vector< PeptideIdentification > current_spectrum_search_;
+      XQuestResultMeta current_meta_; // The current meta value
+      int & n_hits_; // Total no. of hits found in the result XML file
```
timosachsenberg: & required?

src/openms/include/OpenMS/FORMAT/XQuestResultXMLFile.h
```diff
@@ -0,0 +1,101 @@
...
+              bool = false); // Whether the OpenXQuest data will be loaded as Meta Values also into the peptide hits, instead just into the PeptideIdentification
+    
+    // Currently not implemented
+    //void store(const String &, std::vector< std::vector< PeptideIdentification > > & );
+
+    /**
+     * @brief Returns the total number of hits in the file
+     * @return total number of hits in the file
+     */
+    int get_n_hits() const;
```
timosachsenberg: style: use getNHits() or better getNumberOfHits()

src/openms/source/METADATA/XQuestResultMeta.cpp
```diff
@@ -0,0 +1,72 @@
...
+    MetaInfoInterface(source)
+  {
+  }
+
+  XQuestResultMeta::~XQuestResultMeta()
+  {
+  }
+
+/*
+  XQuestResultMeta &  XQuestResultMeta::operator=(const XQuestResultMeta & source)
```
timosachsenberg: required

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    {
+      //Indizes of the rank one elements within the all_ids vector
+      const vector < PeptideIdentification > & all_ids;
+      const vector< UInt >  & rank_one_ids;
+      const String & meta_value;  // Meta value to sort the peptideIdentifications
+
+      inline bool operator()(const UInt & index1, const UInt & index2)
+      {
+        return (  TOPPXFDR::getXLMetaValue<double>(meta_value, all_ids[rank_one_ids[index1]], true)
+                > TOPPXFDR::getXLMetaValue<double>(meta_value, all_ids[rank_one_ids[index2]], true));
```
timosachsenberg: <

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    /**
+     * @brief Tests for descending ordering of a vector of rank one ids based on the score. Only used for assertions.
+     * @param order vector of ranks
+     * @param all_ids Vector with the actual PeptideIdentifictions which are queried for the score.
+     * @param rank_one_ids Vector of indices pointing to the rank one IDs in all_ids.
+     * @return Whether rank vector order reflects descending ordering of rank_one_ids by score.
+     */
+    static bool isSortedDescending(vector< UInt > & order, vector< PeptideIdentification > & all_ids, vector< UInt > & rank_one_ids )
+    {
+      for (Size i = 0; i < order.size() - 1; ++i)
```
timosachsenberg: check for order.empty()  == true, otherwise you will get a nasty underflow here

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    };
+
+    /**
+     * @brief Tests for descending ordering of a vector of rank one ids based on the score. Only used for assertions.
+     * @param order vector of ranks
+     * @param all_ids Vector with the actual PeptideIdentifictions which are queried for the score.
+     * @param rank_one_ids Vector of indices pointing to the rank one IDs in all_ids.
+     * @return Whether rank vector order reflects descending ordering of rank_one_ids by score.
+     */
+    static bool isSortedDescending(vector< UInt > & order, vector< PeptideIdentification > & all_ids, vector< UInt > & rank_one_ids )
```
timosachsenberg: use http://www.cplusplus.com/reference/algorithm/is_sorted/ and appropriate predicates?

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+        }
+      }
+      return true;
+    }
+
+    /**
+     * Searches for a meta value in the cross-link peptide ID. First in PeptideIdentification itself, then
+     * in the associated Peptide hits.
+     */
+    template<typename T>
```
timosachsenberg: this method looks a bit fishy ;) is the template required? the logic is also a bit inexpected (e.g., if size == 2)
Consider reworking it (e.g., split it into 3 members)

timosachsenberg: member functions

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+      }
+      if (pep_hits.size() == 2 && pep_hits[1].metaValueExists(name))
+      {
+        return static_cast<T>(pep_hits[1].getMetaValue(name));
+      }
+      return DataValue::EMPTY;
+    }
+
+    /**
+     * @brief Prepares read vector of PeptideIdentification such that it can be further processed downstream
```
timosachsenberg: maybe a bit more information what happens inside. the comment does not contain much additional information compared to the function name ;)

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+              Internal::XQuestResultXMLHandler::removeSubstring(beta_prot, Internal::XQuestResultXMLHandler::decoy_string);
+              pep_id.setMetaValue( alpha_prot == beta_prot ? "OpenXQuest:is_intraprotein" : "OpenXQuest:is_interprotein" , DataValue());
+            }
+          }
+        }
+      }
+    }
+
+    /**
+     * @brief Add class to the scores Map, of not already present
```
timosachsenberg: of -> if

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+     * @param scores Map where the empty xl class should be added to
+     * @param name name of the xl class to be added
+     */
+    inline static void addEmptyClass(std::map< String, vector< double> > & scores, const String & name)
```
timosachsenberg: it should throw an error if name already present! Otherwise, you may end up with very nasty stuff to debug

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+                        ||     (alpha_is_decoy
+                        &&   ! beta_is_decoy)))
+        {
+          types.push_back(TOPPXFDR::xlclass_hybriddecoysinterlinks);
+        }
+      }
+    }
+
+
+    /** False positve counting as performed by xProphet software package.
```
timosachsenberg: please remove big bulks of commented out code (e.g., you can add a commit just for removing the comments. That way you easily find it again if you need to)

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+
+    /**
+     * @brief Calculates the qFDR values for the provided FDR values, assuming that the FDRs are sorted by score in the input vector
+     * @param fdr Vector with FDR values which should be used for qFDR calculation
+     * @param qfdr Result qFDR values
+     */
+    void calc_qfdr(const vector< double > & fdr, vector< double > & qfdr)
+    {
+      qfdr.resize(fdr.size());
+      for (int i = fdr.size() - 1; i > -1; --i)
```
timosachsenberg: I would write i >= 0 instead of i > -1
but I guess this is just a matter of taste

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+            smallest_fdr = fdr_to_check;
+          }
+        }
+        qfdr[i] = smallest_fdr < current_fdr ? smallest_fdr : current_fdr;
+      }
+    }
+
+    // this function will be used to register the tool parameters
+    // it gets automatically called on tool execution
+    void registerOptionsAndFlags_()
```
timosachsenberg: consider moving this to the top so one can easily check if, for example, the tool documentation reflects the tool options

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    {
+      // Verbose Flag
+      registerFlag_(TOPPXFDR::param_verbose, "Whether the log of information will be loud and noisy");
```
timosachsenberg: consider using the debug paramter (part of every TOPP tool)

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+
+      // File input
+      registerInputFile_(TOPPXFDR::param_in, "<file>", "", "Results in the original xquest.xml format", false);
+      setValidFormats_(TOPPXFDR::param_in, ListUtils::create<String>("xml,mzid,idXML"));
+      
+      // idXML output
+      registerOutputFile_(TOPPXFDR::param_out_idXML, "<idXML_file>", "", "Output as idXML file", true, false);
```
timosachsenberg: remove param_

timosachsenberg: and all occurrences below

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+
+    // this function will be used to register the tool parameters
+    // it gets automatically called on tool execution
+    void registerOptionsAndFlags_()
+    {
+      // Verbose Flag
+      registerFlag_(TOPPXFDR::param_verbose, "Whether the log of information will be loud and noisy");
+
+      // File input
+      registerInputFile_(TOPPXFDR::param_in, "<file>", "", "Results in the original xquest.xml format", false);
```
timosachsenberg: we don't use the toolname again as a subsection:  TOPPXFDR::param_in -> in

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+            if( getXLMetaValue<int>("xl_rank", pep_id, false) == 1)
+            {
+              rank_one_ids.push_back(rank_counter);
+            }
+            rank_counter++;
+          }
+        }
+      }
+      // TODO Currently not supported
+      else if (arg_in.hasSuffix("mzid"))
```
timosachsenberg: please check other tools how they determine the filetype

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+        MzIdentMLFile().load(arg_in, prot_ids, all_ids);
+        this->prepareInput(all_ids);
+
+        Size rank_counter = 0;
+        for (vector< PeptideIdentification >::const_iterator all_ids_it = all_ids.begin();
+             all_ids_it != all_ids.end(); ++all_ids_it)
+        {
+          PeptideIdentification pep_id = *all_ids_it;
+
+          if (getXLMetaValue<int>("xl_rank", pep_id, false) == 1)
```
timosachsenberg: hard to understand in isolation - add comment

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+      }
+      else if (arg_verbose)
+      {
+        LOG_INFO << "INFO: Input is not xQuest. Omit computing delta score and min. number of matched ions." << endl;
+      }
+
+      /*
+       * Sort rank one hits in descending order according to the score
+       */
+      vector< UInt > order_score ( boost::counting_iterator<Size>(0),
```
timosachsenberg: please add a comment for the counting iterator initialization

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+                                   boost::counting_iterator<Size>(rank_one_ids.size()));
+
+      // Configure the sorting of the Peptide Identifications
+      less_than_by_key order_conf = {
```
timosachsenberg: smells like C11

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+        {
+          id = pep_id.getMetaValue("OpenXQuest:id").toString();
+          error_rel = static_cast<double>(pep_id.getMetaValue("OpenXQuest:error_rel"));
+          delta_score = (*(delta_scores[order_score[i]]))[pep_id_index];
+          ions_matched = n_min_ions_matched[order_score[i]];
+        }
+        num_flagged++;
+        double score = getXLMetaValue<double>("OpenXQuest:score", pep_id, true);
+        
+        // Only consider peptide identifications which  fullfill all filter criteria specified by the user
```
timosachsenberg: this line looks rather complex. can you untangle it a bit and add more comments

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+          {
+            scores[*xl_types_it].push_back(score);
+          }
+        }
+      }
+      if (arg_verbose)
+      {
+        LOG_INFO << "INFO: " << num_flagged << " hits have been used to calculate FDR" << std::endl;
+      }
+      // Push empty vector for all remaining empty classes
```
timosachsenberg: why needs this to be done? extend comment

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+        vector< double > current_scores = scores_it->second;
+        String classname = scores_it->first;
+        cum_histograms[classname] = new Math::CumulativeHistogram<>(current_scores.begin(), current_scores.end(),
+                                                                    this->min_score,
+                                                                    this->max_score,
+                                                                    TOPPXFDR::fpnum_score_step, true, true);
+      }
+
+      
+      // This is currently not needed for the FDR calculation
```
timosachsenberg: consider removing unused comments

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+          }
+        }
+        if ( ! assigned)
+        {
+          LOG_WARN << "WARNING: Cross-link could not be identified as either interlink, intralink, or monolink, so no FDR will be available." << endl;
+        }
+      }
+
+      // Delete Delta Scores
+      for (std::vector< std::vector< double >* >::const_iterator it = delta_scores.begin(); it != delta_scores.end(); ++it)
```
timosachsenberg: can we go without the pointer stuff?

src/utils/XFDR.cpp
```diff
@@ -0,0 +1,1095 @@
...
+      {
+        delete *it;
+      }
+
+      // Delete cumulative_histograms
+      for (std::map< String, Math::CumulativeHistogram<> * >::iterator cum_histograms_it = cum_histograms.begin();
```
timosachsenberg: same here. I think we can go without pointers

src/utils/executables.cmake
```diff
@@ -53,6 +53,7 @@ TICCalculator
 TransformationEvaluation
 TopPerc
 XMLValidator
+XFDR
```
timosachsenberg: F before M 
could you also change the order of TopPerc in the file so it is sorted

src/openms/include/OpenMS/FORMAT/HANDLERS/XQuestResultXMLHandler.h
```diff
@@ -0,0 +1,169 @@
...
+      static const String decoy_string;
+
+      /**
+       * @brief Removes a substring from a larger string
+       * @param large String from which `small` should be removed
+       * @param small Substring to be removed from `large`
+       */
+      static void removeSubstring(String & large, const String & small);
+
+      XQuestResultXMLHandler(const String & /* filename */,
```
hroest: no need to comment out the name in the declaration

src/openms/include/OpenMS/FORMAT/HANDLERS/XQuestResultXMLHandler.h
hroest: I dont fully understand the point of doing this, why not simply use a member and have class deal with creation/destruction?

src/utils/XFDR.cpp
lukaszimmermann: This seems to be c++11 only.

src/utils/XFDR.cpp
lukaszimmermann: The prefix is just there such that the static members of the tool are a bit grouped together. So the prefix param_ here means that the String values of the static members denote parameter names.
So I'd actually prefer leaving it (given that static String members are ok for defining constant String values). Another group would be xlclass.

src/utils/XFDR.cpp
lukaszimmermann: This is not a subsection. I have just defined the name of the parameter as a static member of the tool.

src/openms/include/OpenMS/FORMAT/XQuestResultXMLFile.h
```diff
@@ -48,11 +48,11 @@ namespace OpenMS
     XQuestResultXMLFile();
     ~XQuestResultXMLFile();
 
-    void load(const String &,  // filename
-              std::vector< std::vector< PeptideIdentification > > &, // Vector of encompassed spectra, each containing associated PeptideIdentifications
-              std::vector< ProteinIdentification > &,  // The protein identification of the xQuest result file
-              Size = 0,  // Minimum number of PeptideIdentification a spectrum must contain to be loaded
-              bool = false); // Whether the OpenXQuest data will be loaded as Meta Values also into the peptide hits, instead just into the PeptideIdentification
+    void load(const String & filename,  // filename
```
hroest: the documentation should be in doxygen format, not inside the declaration please! use
```
/* @brief
 *
 * @param filename ...
 * @param csms ...
 */
```

src/utils/XFDR.cpp
lukaszimmermann: Seems to work also in C++98: cpp.sh/9wdsd

lukaszimmermann:
I should have addressed everything. mzIdentML files can also now be read in and written.
The FDR / q-value is also now set as score for the peptide hits. Next, there will be some test cases.

lukaszimmermann:
Seems like FalseDiscoveryRate test fails here. I have no clue why.

enetz:
http://cdash.openms.de/testDetails.php?test=40217437&build=223875

the order of protein_refs in idXML is switched,
is it the same as this issue? #2596

timosachsenberg:
yes please merge against develop... fixed test ist included

hroest:
looks good from my side :+1:

timosachsenberg:
lukas can you resolve the conflicts and push an updated version. I will then give it a last look before a merge.

PR #2510, opened 2017-04-10
[FIX] handles mzML files with integer data type

hroest:
- when using pic (positive integer compression), proteowizard uses the
  1000519 tag (32-bit integer)
- since the result of a ms-numpress decompression is always a 64 bit
  floating point number, this seems to be the safest course of action at
  the moment and makes converted files work in OpenMS
- amends 22fc1cf6ff17498c86f148e6cd66494a63bd9492


timosachsenberg:
+1

PR #2512, opened 2017-04-11
Feature/pyopenms

hroest:
fix and adopt pyOpenMS


timosachsenberg:
thanks a ton again for taking care of the pyOpenMS bindings

PR #2514, opened 2017-04-12
Feature/pyopenms

hroest:
- test fixes
- a few new wrappers


hroest:
since this is pyopenms only, I will merge

Issue #2459, opened 2017-03-21
PeakPickerHiRes signal to noise estimator


timosachsenberg:
Hi,
1. I made the experience that the default signal to noise setting of PeakPickerHiRes often looses good features or mass traces on orbitrap and SCIEX instruments.
I would vote for disabling it by default but would like to hear about your experience.
Hi,
2. I made the experience that the most common configuration are profile MS1 peaks and centroided MS2.
I would vote for making this the default.
3. Also, if a spectrum is annotated as centroided, we could just issue a warning and omit picking for that spectrum. Note that we should still stop processing if it is not annotated but has to be estimated from the spectrum.

@hendrikweisser @hannesveit @hroest @cbielow @lars20070 @mwalzer

lars20070:
+1

I always switch the noise estimation off, `signal_to_noise = 0`.

hroest:
I think it depends on the kind of data you have - for TOF data, low res data and for chromatographic data it does make sense to have a S/N cutoff. In my workflow for chromatographic data I set it to a low value, generally to `0.1` for chromatographic data

> Also, if a spectrum is annotated as centroided, we could just issue a warning and omit picking for that spectrum. Note that we should still stop processing if it is not annotated but has to be estimated from the spectrum.

I would include a `force` flag then that still allows you to pick -- nothing more annoying than a software that thinks it knows better what you want and makes the wrong decision!

hannesveit:
1. I agree with changing the default setting to S/N = 0. Of course, as @hroest pointed out, there are various use cases where an S/N cutoff makes sense. But when it comes to choosing a default, at least in my experience, S/N = 0 is probably the safer option.

2. I agree with picking only MS1 by default.

3. I agree with @hroest. Better safe than sorry.

mwalzer:
1. I usually switched S/N off, so I can only support the notion of default ```signal_to_noise = 0```.
2. That's what I experienced as well.
3. I like Hannes' suggestion of a force flag

timosachsenberg:
Note: I did not implement the special handling of already picked spectra.

PR #2516, opened 2017-04-12
[PARAM] disable SN estimation by default, pick only MS1 by default

timosachsenberg:
addresses the parameter issues in https://github.com/OpenMS/OpenMS/issues/2459

### DIFF
--- a/src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
+++ b/src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
@@ -46,7 +46,7 @@ namespace OpenMS
     ProgressLogger()
   {
     // set default parameter values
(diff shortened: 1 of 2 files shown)


mwalzer:
:+1: 
The corresponding tool's documentation seems to be double on its own page ([even before your changes](http://ftp.math.fu-berlin.de/pub/OpenMS/release-documentation/html/TOPP_PeakPickerHiRes.html)). Can you have a quick look?

timosachsenberg:
good point. Maybe I am blind but I don't see any double include. @jpfeuffer any idea?

jpfeuffer:
It's because the Lowmem version has the same name I guess.
https://github.com/OpenMS/OpenMS/blob/36ccb5c6d443b4ec59b18a6c1cc315e29b516b9c/src/utils/LowMemPeakPickerHiRes.cpp

timosachsenberg:
Ahh I see. I'll but it on a separate entry

timosachsenberg:
ok I added the fix. Let's merge and see if it solves the problem

PR #2059, opened 2016-07-19
Fast TraML reading

hroest:
- substantially improves the TraML Handler in memory efficiency and speed
- introduces new CVTermListInterface that is less memory intensive as it simply stores a pointer to its data structure (similar to MetaInfoInterface)
- explicitly stores `<Interpretation>` in TraML and maps the OBO terms to ion types
- explicitly stores product ion m/z
- explicitly stores unimod identifiers for modifications in TraML

on a blood plasma library by Liu et al (http://proteomecentral.proteomexchange.org/cgi/GetDataset?ID=PXD001064 ) of 898590 KB, the `ConvertTSVToTraML` tool now only takes 2:11 minutes to convert and utilizes 940.35 MB of memory. The same tool of Release 2.0 uses 4:01 minutes and utilizes 5760.74 MB of memory. This amounts to a 6-fold reduction in memory footprint and 2-fold reduction in execution time.

PR #2518, opened 2017-04-13
Use UniMod id as integer

hroest:
- consistently use unimod identifier as integer and only convert to "UniMod:XX" when necessary

- some stuff that was forgotten in #2059 
- some code paths still relied on CV terms

### DIFF
--- a/src/openms/source/CHEMISTRY/CrossLinksDB.cpp
+++ b/src/openms/source/CHEMISTRY/CrossLinksDB.cpp
@@ -167,9 +167,8 @@ namespace OpenMS
           if (split[i].hasPrefix("UniMod:"))
           {
             // Parse UniMod identifier to int
-            std::vector<String> unimod_id;
-            split[i].split(":", unimod_id);
-            mod.setUniModRecordId(unimod_id[1].toInt());
+            String identifier = split[i].substr(7, split[i].size());
+            mod.setUniModRecordId(identifier.toInt());
           }
         }
       }
--- a/src/openms/source/CHEMISTRY/CrossLinksDB.cpp
+++ b/src/openms/source/CHEMISTRY/CrossLinksDB.cpp
@@ -316,7 +316,7 @@ namespace OpenMS
     {
 
       // check whether a unimod definition already exists, then simply add synonyms to it
-      if (it->second.getUniModAccession() != "")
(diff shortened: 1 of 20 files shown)


timosachsenberg:
pxd

src/openms/source/CHEMISTRY/CrossLinksDB.cpp
```diff
@@ -166,7 +166,10 @@ namespace OpenMS
         {
           if (split[i].hasPrefix("UniMod:"))
           {
-            mod.setUniModAccession(split[i].trim());
+            // Parse UniMod identifier to int
+            std::vector<String> unimod_id;
```
cbielow: how about: `split[i].substring(...).toInt()` instead of splitting brute force?

hroest:
thanks, I have fixed the issues

src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -769,37 +769,49 @@ START_SECTION([EXTRA] Test integer vs float tags)
...
+  TEST_EQUAL(seq15[3].getModification()->getUniModRecordId(), 4)
+  TEST_EQUAL(seq15[3].getModification()->getUniModAccession(), "UniMod:4")
   }
 
   // Test a few modifications with the accurate mass slightly off to match some other modification
   {
   AASequence seq11 = AASequence::fromString("PEPM[147.035399]TIDEK"); // PSI-MOD oxMet is 147.035399
   TEST_EQUAL(seq11.isModified(), true);
   TEST_STRING_EQUAL(seq11[3].getModificationName(), "MOD:00719")
+  TEST_EQUAL(seq11[3].getModification()->getUniModRecordId(), -1) // no UniMod id gets assigned
```
hroest: I guess I am still a bit confused on how we do modifications, why are there MOD and UniMod both in OpenMS? using sometimes one and other times the others sounds a bit confusing...

timosachsenberg: I think we support both UniMOd as well as PSI-Mod

hroest: is that a good idea? I guess it is easy to get conflicting information with incomplete mapping...

PR #2520, opened 2017-04-13
[DEPRECATION] removed classes deprectated long time ago

timosachsenberg:


hroest:
I really like the tests in this one:

```
// TODO
```

I have never used this part of the code, from my point of view it can go. Maybe lets wait for some other people to weigh in with their opinion.

hroest:
seems like @hendrikweisser marked this as deprecated in 2010

@timosachsenberg can you also remove src/pyOpenMS/pxds/FeatureGroupingAlgorithmIdentification.pxd ?

cbielow:
this is very xcorr specific code, not tested, probably not used by anyone and deprecated. I say burn it.

Issue #2466, opened 2017-03-22
review FileConverter


timosachsenberg:
make advanced:
- in_type
- UID_postprocessing <method>
- out_type
- lossy_compression
- lossy_mass_accuracy

rethink default:
- write_mzML_index is currently no index

```
Options (mandatory options marked with '*'):
  -in <file>*                   Input file to convert. (valid formats: 'mzData', 'mzXML', 'mzML', 'cachedMzML', 'dta', 'dta2d', 'mgf', 'featureXML', 'consensusXML', 'ms2', 'fid', 'tsv',
                                'peplist', 'kroenik', 'edta')
  -in_type <type>               Input file type -- default: determined from file extension or content
                                (valid: 'mzData', 'mzXML', 'mzML', 'cachedMzML', 'dta', 'dta2d', 'mgf', 'featureXML', 'consensusXML', 'ms2', 'fid', 'tsv', 'peplist', 'kroenik', 'edta'
                                )
  -UID_postprocessing <method>  Unique ID post-processing for output data.
                                'none' keeps current IDs even if invalid.
                                'ensure' keeps current IDs but reassigns invalid ones.
                                'reassign' assigns new unique IDs. (default: 'ensure' valid: 'none', 'ensure', 'reassign')
  -out <file>*                  Output file (valid formats: 'mzData', 'mzXML', 'mzML', 'cachedMzML', 'dta2d', 'mgf', 'featureXML', 'consensusXML', 'edta', 'csv')
  -out_type <type>              Output file type -- default: determined from file extension or content
                                Note: that not all conversion paths work or make sense. (valid: 'mzData', 'mzXML', 'mzML', 'cachedMzML', 'dta2d', 'mgf', 'featureXML', 'consensusXML', '
                                edta', 'csv')
  -write_mzML_index             Add an index to the file when writing mzML files (default: no index)
  -lossy_compression            Use numpress compression to achieve optimally small file size (attention: may cause small loss of precision; only for mzML data)
  -lossy_mass_accuracy <error>  Desired (absolute) m/z accuracy for lossy compression (e.g. use 0.0001 for a mass accuracy of 0.2 ppm at 500 m/z, default uses -1.0 for maximal accuracy)
                                (default: '-1')
```

hroest:
the issue  with the index is that the flag is a true flag instead of what we should have done `write_mzML_index true/false`

so its hard to change that but I suggest to  basically switch to a new parameter that has true/false
[...]

PR #2521, opened 2017-04-14
[PARAM] simplified FileConverter parameters

timosachsenberg:


timosachsenberg:
resolves https://github.com/OpenMS/OpenMS/issues/2466

hroest:
actually I dont like the omit index flag - I would prefer a flag where the two options are "true/false" which makes it more descriptive and also clear in a param.xml

timosachsenberg:
k will rework it

timosachsenberg:
@hroest Do you know why the data format and file name has been changed?
```
			<sourceFile id="sf_ru_0" name="FileConverter_4_input.mzXML" location="<TEST_DATA_PATH>">
 -				<cvParam cvRef="MS" accession="MS:1000569" name="SHA-1" value="bfb52b34f2482567ddf4474117ca70c09d96bea1" />
 -				<cvParam cvRef="MS" accession="MS:1000566" name="ISB mzXML format" />
 +			<sourceFile id="sf_ru_0" name="File_test_1.raw" location="">
 +				<cvParam cvRef="MS" accession="MS:1000569" name="SHA-1" value="0000000000000000000000000000000000000000" />
 +				<cvParam cvRef="MS" accession="MS:1000564" name="PSI mzData format" />
  				<cvParam cvRef="MS" accession="MS:1000777" name="spectrum identifier nativeID format" />
  ```

hroest:
@timosachsenberg no, not really, can you try before applying the patch and figure out which code-change causes the change?

cbielow:
as far as I can see, the respective lines where whitelisted, so the change could be from quite a while ago.
Seems also unrelated to this PR, right?
I'd be happy if we could merge this, since #2525 somehow depends on it (loosely).

cbielow:
FileConverter test is failing
http://cdash.openms.de/testDetails.php?test=39331026&build=220084

cbielow:
`offset` and `indexListOffset` might need to go in the whitelist

cbielow:
ok, will merge as discussed. fix is coming the next pr by me

timosachsenberg:
By the way, you can always add such small changes manually to a PR via the web interface.

hroest:
@cbielow @timosachsenberg maybe we should agree to not merge stuff that breaks the build but rather amend the PR

timosachsenberg:
Sure. Was kind of an exception in as I had time monitoring PRs and test results today.

PR #2522, opened 2017-04-15
[BUILD] enable C11 support in cmake

timosachsenberg:
currently, we get some Qt related errors:
http://stackoverflow.com/questions/30530881/facing-unable-to-find-string-literal-operator-error-when-compiling-ui-code-wit

so it might be best to wait after the transition to Qt5 is done.


timosachsenberg:
````/nfs/wsi/abi/old-data/sachsenb/OpenMS_IDE/src/openms/source/SYSTEM/FileWatcher.cpp:48:1: error: unable to find string literal operator 'operator"" __FILE__'
     connect(this, SIGNAL(fileChanged(const QString &)), this, SLOT(monitorFileChanged_(const QString &)));
 ^
```

hroest:
seems like it runs through on travis at least...

did you do a complete clean rebuild?

timosachsenberg:
hmm no but maybe the qt version is more recent? I need to check this tomorrow.

timosachsenberg:
travis is qt 4.8.1
local is qt 4.7.4

timosachsenberg:
I upgraded local to 4.8.6 and it now builds without error.

hroest:
ok, great news that we can do C++11 and Qt5 separately

timosachsenberg:
@jpfeuffer do you know which builds in jenkins would break by this PR?

jpfeuffer:
You mean because of a too old QT version, or a compiler that does not support C++11? Or any of these two reasons?
I think for both questions I would need to look it up or even better, just test.
The thing is, I would rather finalize a build matrix first, before I invest the time in testing all combinations and in the end they are not even used.

timosachsenberg:
Sounds reasonable. So we just merge the c11 feature it and see what breaks? If it is too severe we can revert...

jpfeuffer:
Hmm.. thinking about it: We do not have nightlies for VS2015 yet. Basically all Win builds would break then.

hroest:
why not make it a CMake option for now, setting it to false and we can enable it step by step? by changing the cmake command line and see what breaks? Once we have actual C++11 code in OpenMS of course we would remove the option...

cbielow:
could you check the dev-build docs and update the minimal required Qt version there?!

timosachsenberg:
I can do so but depends on whether we merge qt5 branch first

hroest:
@jpfeuffer why would win builds break? as long as we dont introduce any C++11 features and just turn on the flag, why would it break?

jpfeuffer:
Hmm yes you might be right. I was just thinking because Visual Studio 2010 to 2013 do not even have "standard" flags from what I heard. They seem to just support some features of it and some not. But Probably Cmake will figure that out. We might give it a try.

hroest:
maybe a trial run first to see what cmake does? but even then its not that useful if our build infrastructure does not support C++11, right?

but it seems there is some minimal support for `auto` and others in MSVS https://msdn.microsoft.com/en-us/library/hh567368.aspx even early versions but keeping track of what is allowed and what not is no fun ...

timosachsenberg:
I think we could enable this early and learn what breaks ;). If someone is jumping right into using advanced c11 features we might just warn about it during code review.I think this would be a practical approach and would surcumvent very complex cmake checks / scripts that try to extract and check compiler capabilities.

hroest:
I also feel that this would make @cbielow and @jpfeuffer lives very difficult as they are the ones that have to deal with the fallout ... 

on the other hand, gcc 4.6 which is currently on travis also does not support the full set: https://gcc.gnu.org/gcc-4.6/cxx0x_status.html

cbielow:
after 2.2 release we should make a final decision on which compilers are supported (e.g. GCC 4.7+, VS2013+), then set up out CI/nightly accordingly. And everything which runs there is allowed.
IMHO.

hroest:
just a questions, isnt VS12 on cdash the same as VS2013 or did I misinterpret this? Because VS2013 already has many C++11 features...

jpfeuffer:
No someone named it after the year. Which is confusing I know.

jpfeuffer:
I'll try to change it soon.

PR #2525, opened 2017-04-17
[Feature] mzXML writer compatible with MaxQuant

cbielow:
This PR mainly add support for writing mzXML which is MaxQuant compatible. This allows for a variety of interesting applications :)
The cpp code contains comments where necessary to indicate special requirements (e.g. vendor specifications, required newlines in XML (no joke!) etc).

In particular:
 - bump from mzXML 2.1 to 3.1
 - support of indexed mzXML (be default)
 - support for a new more tags and attributes (optional)
 - support for a new more mass analyzers / ionization methods

All in all, this PR should not have any side effects on other pipelines (which rely on mzML)


src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
```diff
@@ -279,7 +279,7 @@ namespace OpenMS
         }
 
         // Add a new spectrum, initialize and set MS level and RT
-        spectrum_data_.resize(spectrum_data_.size() + 1); // TODO !!
+        spectrum_data_.resize(spectrum_data_.size() + 1);
```
hroest: so whats happening here? why is there a `+1`, why remove the TODO ?

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
```diff
@@ -660,9 +663,19 @@ namespace OpenMS
       if (cexp_->getInstrument() != Instrument() || cexp_->getContacts().size() != 0)
       {
         const Instrument& inst = cexp_->getInstrument();
+        // the Instrument Manufacturer is paramount for some downstream tools
+        // Since the .getVendor() is usually empty, we infer this via the Aquisiton Software, which is unique to Thermo
+        String manufacturer = inst.getVendor();
+        if (String(inst.getSoftware().getName()).toLower().hasSubstring("xcalibur"))
```
hroest: can you check here `&& manufacturer.empty()` ?

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
```diff
@@ -688,20 +701,16 @@ namespace OpenMS
...
         }
         os << "\t\t\t<software type=\"acquisition\" name=\"" << inst.getSoftware().getName() << "\" version=\"" << inst.getSoftware().getVersion() << "\"/>\n";
-        if (analyzers.empty() || !analyzers[0].getResolutionMethod())
-        {
-          os << "\t\t\t<msResolution category=\"msResolution\" value=\"\"/>\n";
-        }
-        else
-        {
+        if (!(analyzers.empty() || !analyzers[0].getResolutionMethod() || cv_terms_[5][analyzers[0].getResolutionMethod()].empty()))
+        { // must not be empty, otherwise MaxQuant crashes upon loading mzXML
```
hroest: nice...

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
```diff
@@ -864,57 +886,42 @@ namespace OpenMS
...
+        if (spec.metaValueExists("filter string"))
         {
-          os << "\" filterLine=\"";
-          os << writeXMLEscape ( (String)spec.getMetaValue("filter string") );
+          os << " filterLine=\"";
+          os << writeXMLEscape((String)spec.getMetaValue("filter string"));
+          os << "\"";
         }
 
-        // base peak mz (used by some programs like MAVEN), according to xsd:
```
hroest: whats happening to the base peak m/z, can MQ not deal with it? I have added this recently and it seems that other software crashes when its not there e8a964c3dbf191d9cbbb80b3d456127bc9b3a41d

hroest: ok, I see its further down -- ignore the comment

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
```diff
@@ -988,9 +1039,19 @@ namespace OpenMS
         }
       }
 
-      os << "\t</msRun>\n"
-         << "\t<indexOffset>0</indexOffset>\n"
-         << "</mzXML>\n";
+      os << "\t</msRun>\n";
+
+      // create scan index (does not take a lot of space and is required for MaxQuant)
```
hroest: can you make the index optional? I dont think its a problem but 

- we do have the `options_.getWriteIndex()` flag
- there may be software out there that cant deal with the index, so it would be frustrating to not be able to remove it

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
```diff
@@ -1012,7 +1073,7 @@ namespace OpenMS
         s_completiontime_ = xercesc::XMLString::transcode("completionTime");
         s_precision_ = xercesc::XMLString::transcode("precision");
         s_byteorder_ = xercesc::XMLString::transcode("byteOrder");
-        s_pairorder_ = xercesc::XMLString::transcode("pairOrder");
+        s_contentType_ = xercesc::XMLString::transcode("contentType");
```
hroest: does this mean  that this breaks backwards compatibility? e.g. back to mzXML 2.1 files?

src/tests/topp/FileConverter_6_output.mzXML
```diff
@@ -28,8 +29,12 @@
...
 			<processingOperation name="parameter: mode" value="test_mode"/>
 		</dataProcessing>
-		<scan num="1" msLevel="1" peaksCount="15" polarity="+" scanType="Full" filterLine="+ c NSI Full ms [ 400.00-1800.00]" basePeakMz="0" retentionTime="PT5.1S" startMz="400" endMz="1800">
-			<peaks precision="32" byteOrder="network" pairOrder="m/z-int">AAAAAEFwAAA/gAAAQWAAAEAAAABBUAAAQEAAAEFAAABAgAAAQTAAAECgAABBIAAAQMAAAEEQAABA4AAAQQAAAEEAAABA4AAAQRAAAEDAAABBIAAAQKAAAEEwAABAgAAAQUAAAEBAAABBUAAAQAAAAEFgAAA/gAAA</peaks>
+		<scan num="1" msLevel="1" peaksCount="15" polarity="+" scanType="Full" filterLine="+ c NSI Full ms [ 400.00-1800.00]" retentionTime="PT5.1S" startMz="400" endMz="1800" lowMz="400.39" highMz="1795.56" basePeakMz="445.347" basePeakIntensity="120054" totIonCurrent="16675500">
+			<peaks precision="32"
+ byteOrder="network"
+ contentType="m/z-int"
+ compressionType="none"
+ compressedLen="0" >AAAAAEFwAAA/gAAAQWAAAEAAAABBUAAAQEAAAEFAAABAgAAAQTAAAECgAABBIAAAQMAAAEEQAABA4AAAQQAAAEEAAABA4AAAQRAAAEDAAABBIAAAQKAAAEEwAABAgAAAQUAAAEBAAABBUAAAQAAAAEFgAAA/gAAA</peaks>
```
hroest: looks great! :-)

src/topp/FileConverter.cpp
```diff
@@ -446,7 +446,7 @@ class TOPPFileConverter :
                                                  CONVERSION_MZXML));
       MzXMLFile f;
       f.setLogType(log_type_);
-      ChromatogramTools().convertChromatogramsToSpectra<MSExperimentType>(exp);
+      //ChromatogramTools().convertChromatogramsToSpectra<MSExperimentType>(exp);
```
hroest: whats happening here?

hroest:
looks good to me, there is still a test failing on travis. Also, it seems that the code is backward compatible to 2.1, is that correct? Finally, it would be great to have the index optional (but on per default) just in case...

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
cbielow: this code is really old. I think what this basically does is avoiding the hassle to specify the type of the spectrum and just use the default ctor. Its doable I guess, but not the most explicit solution.

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
cbielow: I'm a bit hesitant here, since we'd basically want exactly this manufacturer. Even it already says 'Thermo', it should say 'Thermo Scientific'.
But there is other things which must fit to be MQ compatible, so I might just introduce a 'forceMQcompatibility' flag.

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
cbielow: I'd rename the FileConverter flag to `-write_index` then, since it applies to both mzML and mzXML now (after #2521 is merged)
Will adapt the docs accordingly.

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
cbielow: this attribute is utterly useless in all mzXML schemata, since only one value is allowed anyway.
So every good parser should just ignore it -- or throw an error if it does not show the expected value (this is what OpenMS does).

src/topp/FileConverter.cpp
cbielow: this some very old legacy change (by Andreas if I remember correctly).
It basically converts a chromatogram into n separate ms/ms spectra and attaches them to the experiment. It (rightfully) confuses MaxQuant and I cannot see a good reason why we should do this.

hroest: hmm this is not so much legacy but (if I remember correctly) still what proteowizard does with SRM/MRM data and somehow - for better or worse - the "standard" way of doing it. 

>  It (rightfully) confuses MaxQuant and I cannot see a good reason why we should do this.

Do you see any other way to store SRM data in MzXML?

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
hroest: ok, so its just wonderful that this changed from mzXML 2.x to 3.0 - changing a useless attribute that is generally ignored to break backwards compatibility? well so be it, 

- do we have tests of reading 2.x and 3.x mzXML files?

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
hroest: hmm ok, sounds fine to me if MQ needs this. I just hope it wont create problems in the future...

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
hroest: I see its basically a `push_back` of an empty spectrum done in a "smart" way. But it seems modern compilers are smart enough that it doesnt make a difference: http://stackoverflow.com/questions/1902832/resize-versus-push-back-in-stdvector-does-it-avoid-an-unnecessary-copy-assig

still, is fine with me

hroest:
PS: also there seems to be a compilation error: http://cdash.openms.de/viewBuildError.php?buildid=219789

src/openms/source/FORMAT/HANDLERS/MzXMLHandler.cpp
cbielow: our test input-files are still 2.x, we are writing (and reading) 3.x during the tests. No problem there.

src/topp/FileConverter.cpp
cbielow: basically, no :)
If we want to keep this functionality, we should make it XOR, i.e. either spectra or chromatograms, but not both. Would that work for your use cases?

hroest:
:+1: looks good

timosachsenberg:
+1

timosachsenberg:
build timed out

jpfeuffer:
Any explanation for this?
http://cdash.openms.de/testDetails.php?test=39599123&build=220874
```
checking [EXTRA] static bool isValid(const String& filename) ... 
    creating new temporary filename 'MzXMLFile_test_564.tmp' (line 564)
Validation error in file 'MzXMLFile_test_564.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_564.tmp' line 9 column 130: fatal error during schema scan
Validation error in file 'MzXMLFile_test_564.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_564.tmp' line 4 column 155: fatal error during schema scan
 -  line 567:  TEST_EQUAL(f.isValid(tmp_filename, std::cerr),true): got 0, expected 1
: failed

Validation error in file 'MzXMLFile_test_476.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_476.tmp' line 9 column 130: fatal error during schema scan
Validation error in file 'MzXMLFile_test_476.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_476.tmp' line 4 column 155: fatal error during schema scan
 - Error: Invalid mzXML file 'MzXMLFile_test_476.tmp'
 -  invalid file 'MzXMLFile_test_476.tmp' (type: mzXML)
Validation error in file 'MzXMLFile_test_547.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_547.tmp' line 9 column 130: fatal error during schema scan
Validation error in file 'MzXMLFile_test_547.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_547.tmp' line 4 column 155: fatal error during schema scan
 - Error: Invalid mzXML file 'MzXMLFile_test_547.tmp'
 -  invalid file 'MzXMLFile_test_547.tmp' (type: mzXML)
Validation error in file 'MzXMLFile_test_564.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_564.tmp' line 9 column 130: fatal error during schema scan
Validation error in file 'MzXMLFile_test_564.tmp' line 0 column 0: unsupported protocol in URL
Validation error in file 'MzXMLFile_test_564.tmp' line 4 column 155: fatal error during schema scan
 - Error: Invalid mzXML file 'MzXMLFile_test_564.tmp'
 -  invalid file 'MzXMLFile_test_564.tmp' (type: mzXML)
: failed
```
Did I miss to update sth? Some old files/schemas lying around?

hroest:
it seems that this breaks some of the TPP tools, especially the extra newlines in mzXML

PR #2526, opened 2017-04-18
[FEATURE] expose spectral library file

timosachsenberg:


src/utils/MetaboliteSpectralMatcher.cpp
```diff
@@ -53,7 +53,7 @@ using namespace std;
 /**
         @page UTILS_MetaboliteSpectralMatcher MetaboliteSpectralMatcher
 
-        @brief MetaboliteSpectralMatcher assembles metabolite features from singleton mass traces.
+        @brief MetaboliteSpectralMatcher identify small molecules from trandem MS spectra.
```
hroest: tandem?

src/utils/MetaboliteSpectralMatcher.cpp
```diff
@@ -91,106 +85,104 @@ class TOPPMetaboliteSpectralMatcher :
...
-        // addText_("Parameters for the accurate mass search can be given in the 'algorithm' part of INI file.");
-        registerSubsection_("algorithm", "Algorithm parameters section");
-    }
-
-    Param getSubsectionDefaults_(const String& /*section*/) const
+  void registerOptionsAndFlags_()
+  {
+    registerInputFile_("in", "<file>", "", "mzML file");
+    setValidFormats_("in", ListUtils::create<String>("mzML"));
+    registerInputFile_("database", "<file>", "CHEMISTRY/MetaboliteSpectralDB.mzML", "mzML file", false);
```
hroest: maybe add a comment that this file is generally found in OpenMS/share - otherwise it is confusing where this "magic" file comes from. Maybe even add an URL to the comment section where one can download current and previous versions of the file... ?

src/openms/include/OpenMS/ANALYSIS/ID/MetaboliteSpectralMatching.h
```diff
@@ -54,140 +54,132 @@ namespace OpenMS
...
+    /// Default destructor
+    virtual ~MetaboliteSpectralMatching();
 
-      /// main method of MetaboliteSpectralMatching
-      void run(PeakMap&, MzTab&);
+    /// hyperscore computation
+    double computeHyperScore(MSSpectrum<Peak1D>, MSSpectrum<Peak1D>, const double&, const double&);
 
+    /// main method of MetaboliteSpectralMatching
+    void run(PeakMap &, PeakMap &, MzTab &);
```
hroest: please adjust .pxd file

timosachsenberg: I always wondered if it would be possible to "tag" classes and members with pxd code to allow automatic extraction of a pxd file (at least for most classes). Would simplify refactoring and is easier spotted in the review.

src/utils/MetaboliteSpectralMatcher.cpp
timosachsenberg: Yes, that is true. Currently, we don't have a place where we collect this information so its hard to give an URL here. I think I will leave it as is right now until we found a proper place to put the spectral library (and also have someone taking care of updating it)

timosachsenberg:
Comments addressed and builds pass. Will merge.

PR #2527, opened 2017-04-18
[FIX] Removed OPENMS_DLLAPI from class templates

jpfeuffer:
... that have implementations already in header files.

Follow-up on https://github.com/OpenMS/OpenMS/pull/2489 together with @cbielow .
Fine for you @hroest  ?


jpfeuffer:
BTW: The changes were done on the build machines. So we do not expect any errors in the nightlies, too.

hroest:
:+1:

Issue #2320, opened 2016-11-30
Issues in parsing/writing modifications


jpfeuffer:
Issues:
- Segfault with unknown N-Term mods
- Writing unknown amino acids: results for DFPANGERX[113.0840643509]
    - toString: DFPANGER[131.0946294147]   (looks like modified R and is actually interpreted as a  CTerminal modification; also adds H2O)
    - toBracketString: DFPANGERX  (no notion about mass anymore)
    - toUnmodifiedString: DFPANGER (completely missing)

jpfeuffer:
Related to #1792

PR #2530, opened 2017-04-19
Feature/fix square mods

hroest:
fixes a major issue when using square modifications syntax, we came across this using pyOpenMS:

```
>>> import pyopenms
>>> str(pyopenms.AASequence.fromString("N[2457.877]VSVK", True) )
'[2475.8875650638]VSVK'
>>> str(pyopenms.AASequence.fromString(str(pyopenms.AASequence.fromString("N[2457.877]VSVK", True) ), True) )
Warning: unknown N-terminal modification '2475.8875650638' - adding it to the database
Segmentation fault (core dumped)
```

it turns out that when parsing a sequence like N[2457.877]VSVK our `AASequence` "eats" the AA before the square brackets and when calling `toString` it outputs a sequence which is one AA shorter. Firstly, that is incorrect and secondly it means that repeated parsing of strings by OpenMS shorten it by 1 each time until we end up with a String like '[2475.8875650638]VSVK' which then leads to a segfault.

I looked at the code for quite some time and decided that it needed a major overhaul. This PR now substantially enhances our support for the bracket notation syntax which is also evident in the expanded AASequence test:

- `AASequence::fromString().toString()` not produces reproducibly the same result, no more "eating" of AA
- masses are calculated as expected
- `PEPT[+160.230654]IDE` gets correctly parsed as having a T with a modification of the correct mass (for residue as well as for full peptide)
- no more "funny" residues with arbitrary masses, now all modifications are true modifications with the underlying AA residue
- AASequence created from the same string compare as equal
- This allows `b = AASequence::fromString(a.toString())` such that `a == b`
- `.[1600.230654]IDE` gets parsed as N-terminal modification
- `IDE.[1600.230654]` gets parsed as C-terminal modification


src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -714,9 +714,19 @@ namespace OpenMS
 
   std::ostream& operator<<(std::ostream& os, const AASequence& peptide)
   {
+    // this is basically the implementation of toString
+
+    // deal with N-terminal modifications first
     if (peptide.n_term_mod_ != 0)
     {
-      os << ".(" << peptide.n_term_mod_->getId() << ")";
+      if (peptide.n_term_mod_->getId() != "")
```
timosachsenberg: some lint software complaint that .empty() should be preferred

timosachsenberg:
looks good. maybe a second opinion from @hendrikweisser or @enetz ? both have quite a good understanding of mod internals

jpfeuffer:
Can you have a look if it handles unknown AAs correctly now?
https://github.com/OpenMS/OpenMS/issues/2320

jpfeuffer:
Can you add e.g. in ResidueModification.h documentation on how you defined the new rules for using the (full)ID members? It seems like you use the rule that an unknown modification has no ID but a fullID with all the brackets and potentially Termspecifier? Currently there is no documentation on how to handle unknown ones, I think.

Regarding my last comment: Looking at the extensive tests, unknown AAs seem to be handled fine in toString.
Can you only additionally check toBracketString and toUnmodifiedString?

jpfeuffer:
Awesome! This looks great now.

hroest:
I also added the test from #2320 - there is only a single issue that I came across:

the mass of `DFPANGER` differs from `DFPANGERX` and `DFPANXGER` -> the latter are 18 Da *lighter* than the initial sequence. That is strange, it must be some weird interaction with the Internal/Full calculation of monoisotopic weight. I am not really sure what the expected behavior is but it is definitely unexpected. I will look into this some more ...

hroest:
maybe you can argue that you should never add an "X" amino acid without any mass -- but still...

src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -896,6 +898,41 @@ START_SECTION([EXTRA] Arbitrary tag in peptides using square brackets)
...
 
+  // test from #2320
+  {
+    AASequence aa_original = AASequence::fromString("DFPANGER");
+    TEST_REAL_SIMILAR(aa_original.getMonoWeight(), 904.4038997864)
+
+    AASequence aa_withX_original = AASequence::fromString("DFPANGERX");
+    TEST_REAL_SIMILAR(aa_withX_original.getMonoWeight(), 904.4038997864  - 18.0105650638)
+    AASequence aa_withX2_original = AASequence::fromString("DFPANXGER");
+    TEST_REAL_SIMILAR(aa_withX2_original.getMonoWeight(), 904.4038997864  - 18.0105650638)
```
hroest: I dont really think this is what *should* happen

hroest: I guess this comes from 

```
463   double Residue::getMonoWeight(ResidueType res_type) const
[...]
470     case Internal:
471       return mono_weight_ - getInternalToFull().getMonoWeight();

```

jpfeuffer: I agree. This is not optimal since it is somewhat unexpected.

jpfeuffer: Hmm yes.. and X has mass 0. So it results in -H2O

jpfeuffer: How is it compensated in the case of adding an actual mass?

hroest:
lets go ahead with this PR and put the discussion about `X` into #2532 where it belongs

PR #2536, opened 2017-04-20
[FIX] display and accept '--flag' as commandline flags 

cbielow:
Allows users to better distinguish flags from other arguments
i.e.
`FileInfo -in bla.idxml -p`
becomes
`FileInfo -in bla.idxml --p`

The legacy behaviour '-flag' is still valid; old scripts still run and only the usage dialog changes.

Full example:
```
FileInfo -- Shows basic information about the file, such as data ranges and file type.
Version: 2.1.0 Apr 20 2017, 15:27:42, Revision: disabled

Usage:
  FileInfo <options>

Options (mandatory options marked with '*'):
  -in <file>*        Input file  (valid formats: 'mzData', 'mzXML', 'mzML', 'dta', 'dta2d', 'mgf', 'featureXML', 'consensusXML', 'idXML', 'pepXML', 'fid', 'mzid'
                     , 'trafoXML')
  -in_type <type>    Input file type -- default: determined from file extension or content (valid: 'mzData', 'mzXML', 'mzML', 'dta', 'dta2d', 'mgf', 'featureXML'
                     , 'consensusXML', 'idXML', 'pepXML', 'fid', 'mzid', 'trafoXML')
  -out <file>        Optional output file. If left out, the output is written to the command line. (valid formats: 'txt')
  --m                Show meta information about the whole experiment
  --p                Shows data processing information
  --s                Computes a five-number statistics of intensities, qualities, and widths
  --d                Show detailed listing of all spectra and chromatograms (peak files only)
  --c                Check for corrupt data in the file (peak files only)
  --v                Validate the file only (for mzML, mzData, mzXML, featureXML, idXML, consensusXML, pepXML)
  --i                Check whether a given mzML file contains valid indices (conforming to the indexedmzML standard)

Common TOPP options:
  -ini <file>        Use the given TOPP INI file
  -threads <n>       Sets the number of threads allowed to be used by the TOPP tool (default: '1')
  -write_ini <file>  Writes the default configuration file
  --help             Shows options
  --helphelp         Shows all options (including advanced)
```

### DIFF
--- a/src/openms/source/APPLICATIONS/TOPPBase.cpp
+++ b/src/openms/source/APPLICATIONS/TOPPBase.cpp
@@ -184,8 +184,8 @@ namespace OpenMS
       registerStringOption_("id_pool", "<file>", "", String("ID pool file to DocumentID's for all generated output files. Disabled by default. (Set to 'main' to use ") + String() + id_tagger_.getPoolFile() + ")", false);
     }
     registerFlag_("test", "Enables the test mode (needed for internal use only)", true);
-    registerFlag_("-help", "Shows options");
-    registerFlag_("-helphelp", "Shows all options (including advanced)", false);
+    registerFlag_("help", "Shows options");
+    registerFlag_("helphelp", "Shows all options (including advanced)", false);
 
     // parse command line parameters:
     try
@@ -233,7 +233,7 @@ namespace OpenMS
     }
 
     // '--help' given
-    if (param_cmdline_.exists("-help") || param_cmdline_.exists("-helphelp"))
+    if (param_cmdline_.exists("help") || param_cmdline_.exists("helphelp"))
     {
       printUsage_();
       return EXECUTION_OK;
@@ -574,7 +574,7 @@ namespace OpenMS
   void TOPPBase::printUsage_()
   {
     // show advanced options?
-    bool verbose = getFlag_("-helphelp");
+    bool verbose = getFlag_("helphelp");
 
     //common output
     cerr << "\n"
@@ -653,6 +653,7 @@ namespace OpenMS
 
       //NAME + ARGUMENT
       String str_tmp = "  -";
+      if (it->type == ParameterInformation::FLAG) str_tmp += "-"; // use "--" for flags (even though "-" would suffice)
       str_tmp += it->name + " " + it->argument;
       if (it->required)
         str_tmp += '*';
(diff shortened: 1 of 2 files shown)


src/openms/source/APPLICATIONS/TOPPBase.cpp
```diff
@@ -2586,6 +2587,7 @@ namespace OpenMS
     for (int i = argc - 1; i > 0; --i)
     {
       String arg = argv[i];
+      if (arg.hasPrefix("--")) arg = arg.substr(1); // replace '--' by '-' such that it's found in param_map
```
hroest: so basically we now simply accept `--m` and `-m` both as correct?

cbielow: yep. I was afraid to break backwards compatibility, since it could be quite annoying for people to change all their scripts etc...

hroest: but doesnt this now basically mean that for *every* parameter we support `-` and `--` at the same time? basically you could now write `--help` and `-help` with the exact same effect?

cbielow: well, help and helphelp are the (unfortunate) exception here, since they were registered before as "-help" (the "-" was part of the parameter name) -- which was not the smartest solution.
So yes, -help on the commandline would also give you the help text. Do you think thats a problem?

cbielow:
I'm inclined to add a help text somewhere which describes the backwards compatibility. Because people might wonder why their old flags still work while the usage text only speaks of --flag.
probably on this page: http://ftp.mi.fu-berlin.de/pub/OpenMS/release2.1.0-documentation/html/TOPP_general.html

hendrikweisser:
@cbielow: I'm not sure I like this change. I'm used to the UNIX convention of having "-" and "--" for short and long parameter names, respectively (e.g. "-a, --all"). OpenMS has never adhered to this convention, but I fear that attaching a different semantic to "-"/"--" will cause confusion.

> Allows users to better distinguish flags from other arguments

Has that ever been a problem? Are there other (widely used) tools that highlight flags in that way?

cbielow:
what we have right now is violating the UNIX standard (since we use single dash for compound flags), and the proposed change in this PR will still be a violation as well (since we use double-dash for single letter flags).
However you can argue, that 
 - single-letter flags are much more rare (gut feeling -- have not checked)
 - single-letter flags should be replaced by word/compound flags anyway (I'd volunteer if need be)
 - we have arguments which look like flags, but still expect true/false as value, which is confusing and a double-dash would make it clear that you do not need a true/false;
 - double-dash is much better to recognize as flag (both in help text and in the actual command)

I'm not pushing this hard, I just think its a much better interface. Other opinions?

hendrikweisser:
I think it's unconventional and not obvious that "--" would be used to designate flags.

cbielow:
I'll retract the PR, since this does not seem to be such a fancy idea afterall :)

PR #2537, opened 2017-04-21
fix median

topsenw:


src/utils/QCCalculator.cpp
```diff
@@ -156,9 +156,11 @@ class TOPPQCCalculator :
 
   float calculateSNmedian (MSSpectrum<Peak1D>& spec, bool norm = true)
   {
+	if (spec.size() == 0) return 0;
```
mwalzer: Looks good. Leads to the question though, why there would be any empty spectrum.

cbielow: Guess how the bug was found (real data) :-)

cbielow:
C++ style check failed (see http://cdash.openms.de/testDetails.php?test=39516145&build=220397)
i.e. please replace tab with spaces.

cbielow:
thanks Timo.
Will merge.

PR #2539, opened 2017-04-21
fix display of memory in FileInfo

cbielow:
difference can be negative (estimate via Windows API seems inaccurate) and since we are only using unsigned types, the results could look like
`Memory usage while loading:  18822392323423 MB`

but should have been (rounded to MB, i.e. no change)
`Memory usage while loading: 0 MB`

This PR fixes it.


hroest:
nice :+1:

PR #2543, opened 2017-04-21
[FIX] Added support for unidentified features (no PepHits) to IDConflictResolver

jpfeuffer:
Resulted in segfault before.


hroest:
could you add a test case? 

:+1: for fixing a segfault

jpfeuffer:
Ah yes sure. Wait for it.

hroest:
seems like there is an issue:

```
The following tests FAILED:
	1322 - TOPP_IDConflictResolver_3 (Failed)
	1323 - TOPP_IDConflictResolver_3_out1 (Failed)
```

PR #2545, opened 2017-04-21
[FIX] fix X residue mass

### LINKED ISSUE #2532: mass of AA "X"
currently the "special" amino acid "X" has a mass of zero which leads to problems when used as an internal residue since OpenMS assumes that amino acids get hydrolized and loose the mass of H2O when being incorporated into an Amino Acid sequence. This leads to the following counterintuitive behavior:

```
+    AASequence aa_original = AASequence::fromString("DFPANGER");
+    TEST_REAL_SIMILAR(aa_original.getMonoWeight(), 904.4038997864)
+
+    AASequence aa_withX_original = AASequence::fromString("DFPANGERX");
+    TEST_REAL_SIMILAR(aa_withX_original.getMonoWeight(), 904.4038997864  - 18.0105650638)
+    AASequence aa_withX2_original = AASequence::fromString("DFPANXGER");
+    TEST_REAL_SIMILAR(aa_withX2_original.getMonoWeight(), 904.4038997864  - 18.0105650638)
```

where adding an X to a peptide string reduces its mass by 18. This is probably surprising to many people and not what should happen.

One solution is to change the mass of "X" to have the mass of 18, basically giving it the sum formula "H2O". So when this "AA" gets "hydrolyzed", the water gets removed and it contributes a net 0 to the mass of the peptide such that DFPANGER and DFPANGERX have the same mass. It still does not really make sense to ask for the mass of a peptide with an unknown amino acid but at least the answer is less surprising.

The second solution is to basically throw an exception when trying to get the mass of DFPANGERX (note however that DFPANGERX[180] should work as it is well defined).

I would be in favor of the first solution and just give X a mass of 18. I had a discussion with @jpfeuffer which boiled down to this.

hroest:
- fixes #2532 
- I decided to add a few PRECONDITION checks for developers so that we are aware if we do something potentially wrong and try to calculate illegal masses etc.

the downside of this PR is that we dont really have a good way for testing the behavior

### DIFF
--- a/src/openms/source/CHEMISTRY/AASequence.cpp
+++ b/src/openms/source/CHEMISTRY/AASequence.cpp
@@ -248,6 +248,7 @@ namespace OpenMS
 
   EmpiricalFormula AASequence::getFormula(Residue::ResidueType type, Int charge) const
   {
+    OPENMS_PRECONDITION( !this->has( *ResidueDB::getInstance()->getResidue("X") ), "Cannot get formula of sequence with unknown AA with unknown mass.");
 
     if (peptide_.size() >= 1)
     {
@@ -326,6 +327,8 @@ namespace OpenMS
 
   double AASequence::getAverageWeight(Residue::ResidueType type, Int charge) const
   {
+    OPENMS_PRECONDITION( !this->has( *ResidueDB::getInstance()->getResidue("X") ), "Cannot get weight of sequence with unknown AA with unknown mass.");
+
     // check whether tags are present
     double tag_offset(0);
     for (ConstIterator it = this->begin(); it != this->end(); ++it)
@@ -341,6 +344,7 @@ namespace OpenMS
 
   double AASequence::getMonoWeight(Residue::ResidueType type, Int charge) const
   {
+    OPENMS_PRECONDITION( !this->has( *ResidueDB::getInstance()->getResidue("X") ), "Cannot get weight of sequence with unknown AA with unknown mass.");
 
     if (peptide_.size() >= 1)
     {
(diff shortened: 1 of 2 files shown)


jpfeuffer:
We need to fix the tests, too.
http://cdash.openms.de/testDetails.php?test=39575617&build=220765

```Progress of 'Scoring peptide models against spectra...':
Error while parsing sequence SDFQSVAXXSTFDFYQR: found an unknown AA without an estimated mass. Please use PEPTIX[123] syntax to indicate an unknown amino acid with a known mass.

FATAL: uncaught exception!
exception of type Precondition failed occured in line 347, function double OpenMS::AASequence::getMonoWeight(OpenMS::Residue::ResidueType, OpenMS::Int) const of /group/ag_abi/OpenMS/nightly-builds/HEAD/src/openms/source/CHEMISTRY/AASequence.cpp
error message: !this->has( *ResidueDB::getInstance()->getResidue("X") )
---------------------------------------------------```

hroest:
@timosachsenberg why does the fasta file have X in it, can we simply remove the X from the file?

src/tests/topp/SimpleSearchEngine_1.fasta

timosachsenberg:
yes but it should be able to deal with such real life fastas

hroest:
@timosachsenberg this is what we discussed in #2532 and I added a PRECONDITION that would prevent people to call certain functions such as `getMonoWeight` on a sequence with `X` in it as we agreed that it does not make any sense to get the weight of such an amino acid sequence.

Of course we can remove the PRECONDITION statements again, maybe @jpfeuffer @hendrikweisser @mwalzer can comment as well?

timosachsenberg:
no that's fine. I meant that SSE needs to deal with it. Internet is very slow here so I can't promise I will fix it while in Beijing.

hroest:
until then, should we revert this commit? maybe then @timosachsenberg can build on this tree and make a new PR?

timosachsenberg:
no I think it is just a one-liner... let me try it first

PR #2547, opened 2017-04-23
[FEATURE] Added functionality for statistics of Fasta file  

### LINKED ISSUE #2411: FileInfo: print protein database statistics from fasta files
e.g.,: 
- number of proteins
- number of amino acids
- amino acid counts (including the special ones 'X', 'B', ...)
- warn if duplicate sequences
- warn if duplicate headers

amnamasood:
Closes #2411


src/openms/include/OpenMS/ANALYSIS/ID/AccurateMassSearchEngine.h
```diff
@@ -52,7 +52,8 @@
 #include <vector>
 
 namespace OpenMS
-{
+{ 
+  typedef unsigned int uint;
```
cbielow: please dont globally define new types which might already exist in this or other libs.
Try `UInt` if you need this type.

For the code itself: why is this needed here? Its seems not to be used. So please remove.

cbielow: looking at the commits more closely, you probably did not intend to include
26868d9822d333383b211e1778e664cb5a7
please remove.

src/openms/include/OpenMS/FORMAT/FASTAFile.h
```diff
@@ -85,6 +85,17 @@ namespace OpenMS
                && sequence == rhs.sequence;
       }
 
+	  bool HeaderMatches(const FASTAEntry rhs)
```
cbielow: functions start with lower-case
see https://github.com/OpenMS/OpenMS/wiki/Coding-conventions#function-namesmethod-names

src/topp/FileInfo.cpp
```diff
@@ -376,19 +404,73 @@ class TOPPFileInfo :
...
+		os << "Number of sequences: " << entries.size() << "\n"
+			<< "\n";
+	
+		for (loopiter = entries.begin(); loopiter != entries.end(); loopiter=std::next(loopiter))
+		{
+			iter = find_if(entries.begin(), loopiter, bind1st(mem_fun(&FASTAFile::FASTAEntry::HeaderMatches), &(*loopiter)));
+
+			if (iter != loopiter)
+			{
+				os << "WARNING: DUPLICATE SEQUENCE, NUMBER: " << std::distance(entries.begin(), loopiter) << ", ID: " << loopiter->identifier << " SAME AS NUMBER: " << std::distance(entries.begin(), iter) << ", ID: " << iter->identifier << "\n";
```
cbielow: Can you make the warnings (also below) in lower case please.
No reason to scream at the user.

amnamasood:
@cbielow Kindly check it now. I have implemented the required changes and run the test.

cbielow:
Nice!
just cleaning up: I think the test files are missing. Please add them and merge develop to resolve branch conflicts.

src/topp/FileInfo.cpp
```diff
@@ -376,19 +404,73 @@ class TOPPFileInfo :
...
+			number_of_aacids += loopiter->sequence.size();
+		}
+
+		
+		os << "Total amino acids: " << number_of_aacids << "\n\n";
+		os << "Amino acid counts: " << "\n";
+
+		for (it = aacids.begin(); it != aacids.end(); it=std::next(it))
+		{
+			os << it->first << '\t' << it->second << "\n\n";
```
cbielow: looking at the test output: would one \n be better here?

amnamasood: Yes I think so too, so changing it.
Do you think the test file should have more than one sequence, and with repetition?

cbielow: if you're at it anway, why not.
Maybe even a random '*' at the end of a sequence (some FASTA files denote transcription end this way).
Also '-' seems to be valid (https://en.wikipedia.org/wiki/FASTA_format) -- although ive never seen it.

amnamasood: Can you check it now?

src/tests/topp/CMakeLists.txt
```diff
@@ -614,6 +614,10 @@ set_tests_properties("TOPP_FileInfo_15_out1" PROPERTIES DEPENDS "TOPP_FileInfo_1
 add_test("TOPP_FileInfo_16" ${TOPP_BIN_PATH}/FileInfo -test -in ${DATA_DIR_TOPP}/FileInfo_16_input.trafoXML -out FileInfo_16_output.tmp)
 add_test("TOPP_FileInfo_16_out1" ${DIFF} -whitelist "File name" -in1 FileInfo_16_output.tmp -in2 ${DATA_DIR_TOPP}/FileInfo_16_output.txt )
 set_tests_properties("TOPP_FileInfo_16_out1" PROPERTIES DEPENDS "TOPP_FileInfo_16")
+add_test("TOPP_FileInfo_17" ${TOPP_BIN_PATH}/FileInfo -test -in ${DATA_DIR_TOPP}/FileInfo_17_input.fasta -out
+FileInfo_17_output.tmp)
+add_test("TOPP_FileInfo_17_out1" ${DIFF} -whitelist "File name" -in1 FileInfo_17_output.tmp -in2 ${DATA_DIR_TOPP}/FileInfo_17_output.txt )
+set_tests_properties("TOPP_FileInfo_17_out1" PROPERTIES DEPENDS "TOPP_FileInfo_17")
```
hroest: great, thanks for the test! I like it.

hroest:
please resolve conflicts and change all tabs to whitespace (2 spaces indent)

PR #2548, opened 2017-04-24
[FIX] Add empty sections with description right when encountered in paramXMLs.

### LINKED ISSUE #2544: Handling of optional parameters in ini files
I am in the process of fixing a bug in KNIME where we do not write e.g. an optional input file to the ini file, when it is not given.
This results in inis like the following:
https://gist.github.com/jpfeuffer/7cee69eb5c269ccf91e0ee6cded5f97b

Unfortunately the tool (IDMapper) crashes because the spectra category is now basically empty.
Whose fault is it and which rules should we enforce from now on?
Potentially there is a difference in adding the parameter with e.g. an empty string and I am not sure if all our tools take an empty string as a "missing" parameter.
I mean on the CLI you wouldn't add the parameter neither.
@timosachsenberg @cbielow

jpfeuffer:
Should fix #2544 and the IDMapper problem that we currently have in KNIME.

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/Param.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/Param.h
@@ -384,6 +384,13 @@ protected:
       If the section does not exist an empty string is returned.
     */
     const String& getSectionDescription(const String& key) const;
+
+    /**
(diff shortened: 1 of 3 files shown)


jpfeuffer:
Clang times out. QCCalculator style fails (I did not change anything there).

PR #2552, opened 2017-04-25
[FIX]

hroest:
- fix `stdd::stringstream` error
- small fix after #2480


jpfeuffer:
Maybe you want to use the insert_str variable that is already present.
But already +1 from me if it passes.

PR #2559, opened 2017-04-26
[TEST] fix pyopenms test

hroest:
test fix

not sure why it now selects a different mod, but it does

### DIFF
--- a/src/pyOpenMS/tests/unittests/test000.py
+++ b/src/pyOpenMS/tests/unittests/test000.py
@@ -4609,6 +4609,7 @@ def testModificationsDB():
     m = mdb.getModification( s("Phosphorylation"), s("S"), pyopenms.ResidueModification.TermSpecificity.ANYWHERE)
     assert m.getId() == b"Phospho"
 
+    # get out all mods (there should be many, some known ones as well!)
     mods = []
     m = mdb.getAllSearchModifications(mods)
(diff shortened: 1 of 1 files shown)


src/pyOpenMS/tests/unittests/test000.py
```diff
@@ -4629,14 +4631,31 @@ def testModificationsDB():
...
+    assert m.getId() == b"MOD:00719", m.getId()
+    assert m.getFullName() == b"oxidation to L-methionine sulfoxide", m.getFullName()
 
     m = mdb.getBestModificationByMonoMass( 96, 20, b"T", pyopenms.ResidueModification.TermSpecificity.ANYWHERE)
     assert m is not None 
-    assert m.getId() == b"Thr->Pro"
-    assert m.getUniModAccession() == b"UniMod:662"
+    assert m.getId() == b"MOD:00252", m.getId()
+    assert m.getFullName() == b"keratan sulfate D-glucuronosyl-D-galactosyl-D-galactosyl-D-xylosyl-L-threonine", m.getFullName() # something crazy
+    assert m.getUniModAccession() == b"", m.getUniModAccession() # no unimod for crazyness ...
```
hroest: I guess that is the downside of taking the "best" modification - the answer will change when the order of modifications changes ...

PR #2560, opened 2017-04-27
[FEATURE] cleanup of unused code

hroest:
remove some dead code

basically we could remove `getBestModificationByMonoMass` from the code base as we dont really use it and its clunky - we only use `getBestModificationByDiffMonoMass` by simply calculating the mass difference first


enetz:
I agree. If you know one of them you can calculate the other, so allowing both is redundant.
And it looks like getBestModificationByMonoMass is not used anywhere outside of AASequence anyway.

hroest:
we could also just remove `getBestModificationByMonoMass` completely ?

timosachsenberg:
would be fine with me if it is really not used anywhere else.

timosachsenberg:
tests pass now. should we go on with the merge?

PR #2561, opened 2017-04-27
Feature/mzidentml mods

hroest:
fixes for unknown modifications in mzIdentML which often occur in certain workflows, such as Glycan search. We ran into this by parsing a file produced by Byonic which we used for Glycan search - their mzIdentML looked pretty good but we could not parse it. 

from the standard:
```
If the modification is not present in the CV (and this will be checked by the semantic validator
within a given tolerance window), there is a â€œunknown modificationâ€• CV term that MUST
be used instead.
```

the current solution tries to parse the file and if it encounters an unknown mod, it basically throws and we try to parse the "name" field if present. Same for writing mzIdentML if there is no accession number we have to assume its a "custom" modification that we dont know and we just write unknown but try to give as much information as possible for other parsers so that they can do something useful with it (e.g. write an "universal" and "human-readable" bracket string and write down the monoisotopic mass).

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
@@ -816,7 +816,9 @@ namespace OpenMS
           {
             String p;
             //~ TODO simplify mod cv param write
-            p += String("\t<Peptide id=\"") + pepid + String("\">\n\t\t<PeptideSequence>") + jt->getSequence().toUnmodifiedString() + String("</PeptideSequence>\n");
+            // write peptide id with conversion to universal, "human-readable" bracket string notation
+            p += String("\t<Peptide id=\"") + pepid + String("\" name=\"") + 
+                  jt->getSequence().toBracketString(false) + String("\">\n\t\t<PeptideSequence>") + jt->getSequence().toUnmodifiedString() + String("</PeptideSequence>\n");
             if (jt->getSequence().isModified() || jt->metaValueExists("xl_chain"))
             {
               const ResidueModification* n_term_mod = jt->getSequence().getNTerminalModification();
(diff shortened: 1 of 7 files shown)


hroest:
some more wisdom from the standard:

```
7.10
Unknown modifications
In version 1.1.0 onwards of mzIdentML there has been a change with respect to how “unknown modifications” (i.e.
those not present in an allowed CV) are reported on peptides. In version 1.0, <userParam> elements were allowed
on <Peptide> to capture these modifications. In version 1.1.0 onwards, only <cvParam> elements can be given
on <Peptide> and a term “unknown modification” has been added to the PSI-MS CV. This term MUST only be
used if the identified modification is not present in Unimod (or other allowed CV such as PSI-MOD, although PSI-
MOD is now deprecated), according to the identity of the residue modified and the delta mass, within the parent
tolerance specified in the search. The semantic validator will check any uses of the “unknown modification” term
(MS:1001460) and reject files if the modification is present in Unimod.
```

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
```diff
@@ -892,10 +894,30 @@ namespace OpenMS
...
+                      p += "\">\n\t\t\t<cvParam accession=\"XLMOD:" + acc.suffix(':');
+                      p += "\" name=\"" +  mod->getId();
+                      p += "\" cvRef=\"XLMOD\"/>";
+                      p += "\n\t\t</Modification>\n";
+                    }
+                    else
+                    {
+                      // We have an unknown modification, so lets write unknown
+                      // and at least try to write down the delta mass.
+                      if (mod->getDiffMonoMass() > 0.0)
```
timosachsenberg: could this be negative? Unimod mods also have negative delta mass

timosachsenberg:
looks good. only one comment regarding a potential negative delta mass

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
hroest: yes, good catch. should we test `!= 0` or should we do `fabs() < 1e-5` where I think the first one *should* be safe, right?

timosachsenberg: Yes I think in this case it is totally safe.

hroest:
ok, great I changed the diff mono mass - I guess the absolute mass cannot ever be negative...

hroest:
PS: a task for a future parser should be to actually try and add the unknown modifications from the XML instead of trying to parse the optional "name" attribute

hroest:
@mwalzer is the DOM parser just a legacy parser or will we continue with `MzIdentMLDOMHandler.cpp` and `MzIdentMLHandler.cpp` in parallel?

PR #2565, opened 2017-04-27
Fix mz ml autodetect

### LINKED ISSUE #2563: IDFileConverter detects the input type according to the file extension
Therefore, the tool is not easy to integrate into frameworks like Galaxy. Integration would be improved by an additional param where the input file format can be determined. Or is it also possible to read the file format from metadata somehow?

cbielow:
fixes #2563

FileHandler -- use capital 'M' in MzIdentML as required by the schema


jpfeuffer:
Great. +1 if passes

timosachsenberg:
+1 one timeout

src/openms/source/FORMAT/FileHandler.cpp
```diff
@@ -238,7 +238,7 @@ namespace OpenMS
       return FileTypes::MZML;
 
     //"analysisXML" aka. mzid (all lines)
-    if (all_simple.hasSubstring("<mzIdentML"))
+    if (all_simple.hasSubstring("<MzIdentML"))
```
hroest: why not just use `all_simple.toLower()` and then compare?

cbielow: I had that in mind as well, but technically, XML is case sensitive and any other term is not valid XML -- technically.

PR #1642, opened 2015-10-05
Feature/address sanitizer

hroest:
- add build parameter to build with Address Sanitizer
- fix a few issues that caused crashes with Address Sanitizer turned on

PR #2567, opened 2017-04-27
[BUILD] remove address sanitizer

hroest:
- remove from travis builds as they tend to time out
- running the tests takes too long when using the AS

- honestly, I am getting tired of time-outs on travis and the main reason is that building OpenMS is now pretty fast (20 min) but running the tests with AS is pretty slow since it adds extra code that makes the runtime go up a lot

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,17 +46,19 @@ matrix:
     - env: ENABLE_STYLE_TESTING=ON WITH_GUI=ON 
       os: linux
       compiler: gcc
+
 ## gcc tests class and TOPP tests (if we run into a time limit, we can split it)
 ## PP Wavelet is turned off due to unstable behavior ... 
-#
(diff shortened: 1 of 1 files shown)


hroest:
PS: do we still have an AS build in the nightlies? @jpfeuffer

hroest:
looks great now:

```
6 min 27 sec
7 min 22 sec
25 min 23 sec
24 min 22 sec
21 min 10 sec
```

cbielow:
+1 
just out of curiosity: did we ever get something useful from AS?

hroest:
not so much since we use OO C++ mostly but a few things we did pick up

- quite a few `delete` vs `delete []` issues https://github.com/OpenMS/OpenMS/pull/1642/files?w=1
- f7f399845ffc4c75ad1b1aed597f7f4ededd1f90

hroest:
so overall it seems reasonable to run it every night or so since we dont get so much from it....

timosachsenberg:
fine with me

hroest:
@jpfeuffer can you comment on whether any of the nightly build servers run with AS enabled?

jpfeuffer:
No I do not think so. I will look it up. In case we don't. Where should I include it?
How do I report the results?

hroest:
Just add it to the cmake of a debug build.  It is self reporting since the
execution will fail if there is a memory error :-)

sent from my smartphone -- you may keep any typos you find.

Issue #2564, opened 2017-04-27
SQLite not found


cbielow:
CMake errors when configuring OpenMS-dev
```
CMake Error at C:/Program Files/CMake/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find SQLITE (missing: SQLITE_INCLUDE_DIR) (Required is at least
  version "3.15.0")
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindSQLITE.cmake:54 (find_package_handle_standard_args)
  src/openms/cmake_findExternalLibs.cmake:140 (find_package)
  src/openms/CMakeLists.txt:63 (include)
  ```

only when I explicitly use
 
`cmake -DSQLITE_INCLUDE_DIR=c:\dev\contrib_build\include\sqlite .`

it configures successfully.
This should not be required (all the other contrib libs are found when simply giving the contrib include by CMAKE_PREFIX_PATH (which is what I did).

**Comment by jpfeuffer** [comment]
It might be the subfolder. FindSQLITE seems to only search directly in the CMAKE_PREFIXES (e.g. contrib-build/include).
Either try not installing it into the subfolder or add another path in the Find module to search for. (If this was the reason)

**Comment by cbielow** [comment]
during linking i get:

```Error	LNK2001	unresolved external symbol sqlite3_close	OpenMS	C:\dev\OpenMS_build\src\openms\TransitionPQPReader.obj	1	
Error	LNK2001	unresolved external symbol sqlite3_close	OpenMS	C:\dev\OpenMS_build\src\openms\OpenSwathOSWWriter.obj	1	```

**Comment by hroest** [comment]
maybe @grosenberger can comment

@cbielow can you see whether the variable ${SQLITE_LIBRARY} is set in cmake and whether it is set correctly?

**Comment by cbielow** [comment]
//Path to a library.
SQLITE_LIBRARY:FILEPATH=C:/dev/contrib_build/lib/sqlite3.lib

seems correct and is also showing up in OpenMS_LIBRARIES:INTERNAL

However
```
dumpbin /EXPORTS sqlite.lib
Microsoft (R) COFF/PE Dumper Version 14.00.23918.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file sqlite.lib

[...]

PR #2568, opened 2017-04-28
[FIX] search in contrib includes for sqlite (adresses #2564)

cbielow:
[FIX] search in contrib includes for sqlite (addresses #2564)


jpfeuffer:
Has to be tested in the nightlies. MErge.

PR #2570, opened 2017-04-28
[FIX] adopt to CV terms

### LINKED ISSUE #2569: Getting rid of mzML parsing warnings by "fixing" the file
Hi there,

I have some mzML files that generate a lot of warnings of the following types when processed with the command line tools, I observed this in e.g. the OpenSWATHWorkflow tool, the XTandemAdapter as well as in the FileConverter.

> <While loading 'xxx.mzML.gz': Name of CV term not correct: 'MS:1000422 - high-energy collision-induced dissociation' should be 'beam-type collision-induced dissociation'> occurred 97296 times
> <While loading '/tmp/openswath_tmpfile_29.mzML': Name of CV term not correct: 'MS:1000422 - high-energy collision-induced dissociation' should be 'beam-type collision-induced dissociation'> occurred 1520 times

I was wondering if there is a way of fixing this by applying the suggested changes and I tried to use the FileConverter with -in xxx.mzML.gz and -out yyy.mzML.gz but the "converted" file still has the same warnings about the CV term. Before the conversion there were also warnings about en encoding being unspecified and how 64bit was going to be used by default, they looked like this:

> <While loading 'xxx.mzML.gz': Invalid mzML format: Numpress-compressed binary data array 'intensity array' has no child term of MS:1000518 (binary data type) set. Assuming 64 bit float data type.> occurred 96 times

Those have been fixed by the conversion however. So essentially I am wondering why one type of warning is removed by converting but the other is not and if it is possible to make that happen?

If this is already possible I would be thankful if someone could give me an example command to achieve this?

Best,
Paul

hroest:
- fixes #2569
- CV terms obo and the source code were out of sync

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -3546,7 +3546,7 @@ namespace OpenMS
       }
       if (precursor.getActivationMethods().count(Precursor::HCID) != 0)
       {
-        os << "\t\t\t\t\t\t\t<cvParam cvRef=\"MS\" accession=\"MS:1000422\" name=\"high-energy collision-induced dissociation\" />\n";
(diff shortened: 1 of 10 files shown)


timosachsenberg:
+1

PR #2571, opened 2017-04-28
Feature/debug mode

hroest:
- helps to avoid #2569 in the future
- only print the not-so-helpful warnings for developers


timosachsenberg:
+1

hroest:
so if you want the warnings, either compile in debug mode or use "-debug"

PR #2574, opened 2017-05-01
[FIX] [Win] Missing static

jpfeuffer:


hroest:
I thought you cannot specify static in the cpp file and it seems this is also the error we get when compiling: 

```
Error 	

/.../OpenMS/src/openms/source/ANALYSIS/OPENSWATH/ChromatogramExtractor.cpp:185:3: error: 'static' can only be specified inside the class definition

  static String ChromatogramExtractor::extract_id_<OpenSwath::LightTargetedExperiment>(OpenSwath::LightTargetedExperiment& transition_exp_used, String id)
  ^~~~~~~
```

hroest:
hmm maybe we need to move these two functions to the header file?

https://www.gamedev.net/topic/491241-c-mysterious-linker-error--explicit-template-specialization-involved/

>  Alright, I've done some reading on the web, and the solution is to make my template specialization inline and move its implementation from the cpp to the header. That fixes the problem but it doesn't really make me happy because it also means I have to move a whole bunch of #includes from my cpp to my header as well, which doesn't exactly benefit compile times. :-(

jpfeuffer:
Yes that's actually what I did on the build machine. I thought it would work out in the cpp, too.
:(

jpfeuffer:
I'll do a PR from the build machine.

PR #2575, opened 2017-05-01
[FIX][Win] static template fxn win

jpfeuffer:


jpfeuffer:
-.- but it runs on Windows

jpfeuffer:
Now moving it outside of the class but in the header file.

hroest:
lets try without `static`

jpfeuffer:
Yes. works on win, too. Last combination. It HAS to work :)

hroest:
great but now we get multiple definitions on gcc/clang:

```
/.../OpenMS/src/openms/source/ANALYSIS/OPENSWATH/ChromatogramExtractor.cpp:(.text+0x0): multiple definition of `OpenMS::String OpenMS::ChromatogramExtractor::extract_id_<OpenSwath::LightTargetedExperiment>(OpenSwath::LightTargetedExperiment&, OpenMS::String)'
```

which makes sense again since the same function is defined multiple times now :-( 

http://stackoverflow.com/questions/11773960/do-template-specialisations-belong-into-the-header-or-source-file
> As such, it can't be defined in multiple translation units unless it's declared inline.

again, so why is MSVS unhappy if its in the cpp file? Its just a regular function at that point, how come it complains...

hroest:
see also here: http://stackoverflow.com/questions/10535667/does-it-make-any-sense-to-use-inline-keyword-with-templates

anyways, I think I understand the gcc/clang errors but I dont really understand MSVS...

jpfeuffer:
Me neither. This one might also shed some light: http://stackoverflow.com/questions/17195106/template-specialization-different-behaviour-between-msvc-and-gcc-mingw

jpfeuffer:
Seems to run on Win. Should we go with it?

hroest:
yes, lets merge something that works and improve later

it seems according to the article, this would be possible

```
// declare explicit specialization
template<> void MyClass<int>::doSomething();
// declare explicit instantiation
extern template void MyClass<int>::doSomething();
```

and then define it in the cpp and gcc and MSVS should be happy with that. the benefit is small though...I thought its nice to move stuff out of the header files but it also doesnt hurt much

cbielow:
I think the problem is somewhere else:
you declare the whole class as using OPENMS_DLLAPI
which means that the dll should export the symbols -- but since you're looking at a template function the compiler will not do that, but the linker expects its anyway.
Have you tried removing OPENMS_DLLAPI from the class and using it only for the non-template functions?

cbielow:
just for reference:
the error
 C:\Program Files (x86)\Microsoft Visual Studio 11.0\VC\bin\link.exe /ERRORREPORT:QUEUE /OUT:"C:\dev\NIGHTLY\builds\win8-VS12_x86-Debug\src\tests\class_tests\bin\Debug\ChromatogramExtractor_test.exe" /INCREMENTAL /NOLOGO kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib ..\..\..\openms\Debug\OpenMSd.lib ..\..\..\openswathalgo\Debug\OpenSwathAlgod.lib "C:\dev\qt-4.8.4_vs12_32bit\lib\QtNetworkd4.lib" "C:\dev\qt-4.8.4_vs12_32bit\lib\QtCored4.lib" C:\dev\contrib_vs12_32bit\lib\libsvmd.lib "C:\dev\contrib_vs12_32bit\lib\xerces-c_3D.lib" "C:\dev\contrib_vs12_32bit\lib\libboost_iostreams-mt-gd.lib" "C:\dev\contrib_vs12_32bit\lib\libboost_date_time-mt-gd.lib" "C:\dev\contrib_vs12_32bit\lib\libboost_math_c99-mt-gd.lib" "C:\dev\contrib_vs12_32bit\lib\libboost_regex-mt-gd.lib" C:\dev\contrib_vs12_32bit\lib\Wm5Cored.lib C:\dev\contrib_vs12_32bit\lib\Wm5Mathematicsd.lib C:\dev\contrib_vs12_32bit\lib\libbz2d.lib C:\dev\contrib_vs12_32bit\lib\zlibd.lib C:\dev\contrib_vs12_32bit\lib\sqlite3.lib C:\dev\contrib_vs12_32bit\lib\glpkd.lib opengl32.lib /MANIFEST /MANIFESTUAC:"level='asInvoker' uiAccess='false'" /manifest:embed /DEBUG /PDB:"/.../builds/win8-VS12_x86-Debug/src/tests/class_tests/bin/Debug/ChromatogramExtractor_test.pdb" /SUBSYSTEM:CONSOLE /TLBID:1 /DYNAMICBASE /NXCOMPAT /IMPLIB:"/.../builds/win8-VS12_x86-Debug/src/tests/class_tests/openms/Debug/ChromatogramExtractor_test.lib" /MACHINE:X86 /SAFESEH  /machine:X86 /debug ChromatogramExtractor_test.dir\Debug\ChromatogramExtractor_test.obj

ChromatogramExtractor_test.obj : error LNK2019: unresolved external symbol "private: static class OpenMS::String __cdecl OpenMS::ChromatogramExtractor::extract_id_<class OpenMS::TargetedExperiment>(class OpenMS::TargetedExperiment &,class OpenMS::String)" (??$extract_id_@VTargetedExperiment@OpenMS@@@ChromatogramExtractor@OpenMS@@CA?AVString@1@AAVTargetedExperiment@1@V21@@Z) referenced in function "public: static void __cdecl OpenMS::ChromatogramExtractor::return_chromatogram<class OpenMS::TargetedExperiment>(class std::vector<class boost::shared_ptr<struct OpenSwath::OSChromatogram>,class std::allocator<class boost::shared_ptr<struct OpenSwath::OSChromatogram> > > &,class std::vector<struct OpenMS::ChromatogramExtractorAlgorithm::ExtractionCoordinates,class std::allocator<struct OpenMS::ChromatogramExtractorAlgorithm::ExtractionCoordinates> > &,class OpenMS::TargetedExperiment &,class OpenMS::SpectrumSettings,class std::vector<class OpenMS::MSChromatogram<class OpenMS::ChromatogramPeak>,class std::allocator<class OpenMS::MSChromatogram<class OpenMS::ChromatogramPeak> > > &,bool)" (??$return_chromatogram@VTargetedExperiment@OpenMS@@@ChromatogramExtractor@OpenMS@@SAXAAV?$vector@V?$shared_ptr@UOSChromatogram@OpenSwath@@@boost@@V?$allocator@V?$shared_ptr@UOSChromatogram@OpenSwath@@@boost@@@std@@@std@@AAV?$vector@UExtractionCoordinates@ChromatogramExtractorAlgorithm@OpenMS@@V?$allocator@UExtractionCoordinates@ChromatogramExtractorAlgorithm@OpenMS@@@std@@@3@AAVTargetedExperiment@1@VSpectrumSettings@1@AAV?$vector@V?$MSChromatogram@VChromatogramPeak@OpenMS@@@OpenMS@@V?$allocator@V?$MSChromatogram@VChromatogramPeak@OpenMS@@@OpenMS@@@std@@@3@_N@Z) [C:\dev\NIGHTLY\builds\win8-VS12_x86-Debug\src\tests\class_tests\openms\ChromatogramExtractor_test.vcxproj]

PR #2577, opened 2017-05-02
[FIX] error message was not properly propagated

hroest:


hroest:
Basically the exceptions `IllegalArgument` and `MissingInformation` always said `"unknown error"` since there was no call to `GlobalExceptionHandler::getInstance().setMessage`. This is now fixed and the exceptions produce a meaningful output

PR #2578, opened 2017-05-02
[FEATURE] add force flag to OpenSwath

hroest:
allows to skip a few invalid peptides and still read in the whole
library - maybe a good solution for some cases.


hroest:
will help to solve issues like https://github.com/OpenMS/OpenMS/issues/2569#issuecomment-298708274 in the future (or at least circumvent them)

timosachsenberg:
HI @hroest did you consider using the force flag? It's part of every tool and intended to overwrite sanity checks on the data

hroest:
yes, but there is already a -force flag for OpenSwathWorkflow and it does something different so we can separate the two cases

grosenberger:
👍

hroest:
this should allow people to use

```
-Library:force_invalid_mods
```

to read in old libraries

PR #2579, opened 2017-05-03
[DOC] fix documentation

hroest:
correct links to our utils and topp tools


timosachsenberg:
thanks. appreciated

hroest:
we should probably also do src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp

PR #2528, opened 2017-04-18
Minor fixes for validity of mzIdentML

hendrikweisser:
Converted idXML files from the DeMix pipeline to mzIdentML. The results had some validation errors (`FileInfo -v`):
- the date cannot be empty (error: "illegal year value '0000-00-00T00:00:00'")
- "ModificationParams" must not be empty (error: "empty content is not valid for content model '(SearchModification+)'")

These problems are fixed here.

A remaining issue is that modifications that were used in the search - and do occur on the peptides - are not listed in the search parameters. However, the mzid files still validate.
There is code to pick up modifications from the peptides and add them to the search parameters in `PercolatorOutfile::getSearchModifications_`. Is there a class/header for "ID data convenience functions" where this could be moved to and reused?

PR #2581, opened 2017-05-03
Infer fixed/variable modification definitions from data when necessary

hendrikweisser:
As discussed in #2528:
When writing mzIdentML, infer fixed/variable modifications used from search results (peptide IDs), if none are defined in the search parameters.

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationDefinition.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationDefinition.h
@@ -63,8 +63,11 @@ public:
     /// copy constructor
     ModificationDefinition(const ModificationDefinition& rhs);
 
-    /// detailed constructor specifying the modifications name
-    explicit ModificationDefinition(const String& mod);
+    /// detailed constructor specifying the modification by name
+    explicit ModificationDefinition(const String& mod, bool fixed = true, UInt max_occur = 0);
+
+    /// detailed constructor specifying the modification by name
+    ModificationDefinition(const ResidueModification& mod, bool fixed = true, UInt max_occur = 0);
 
     /// destructor
     virtual ~ModificationDefinition();
@@ -79,11 +82,11 @@ public:
     /// returns if the modification if fixed true, else false
     bool isFixedModification() const;
 
-    /// set the maximal number of occurences per peptide, unbound if 0
-    void setMaxOccurences(UInt num);
+    /// set the maximal number of occurrences per peptide (unbounded if 0)
+    void setMaxOccurrences(UInt num);
 
-    /// returns the maximal number of occurences per peptide
-    UInt getMaxOccurences() const;
+    /// returns the maximal number of occurrences per peptide
+    UInt getMaxOccurrences() const;
 
     /// returns the name of the modification
     String getModificationName() const;
@@ -127,8 +130,8 @@ protected:
     /// fixed (true) or variable (false)
     bool fixed_modification_;
 
-    /// maximal number of occurences per peptide
-    UInt max_occurences_;
+    /// maximal number of occurrences per peptide
+    UInt max_occurrences_;
   };
 
 } // namespace OpenMS
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationDefinition.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationDefinition.h
@@ -66,7 +66,7 @@ public:
     /// detailed constructor specifying the modification by name
     explicit ModificationDefinition(const String& mod, bool fixed = true, UInt max_occur = 0);
 
-    /// detailed constructor specifying the modification by name
+    /// direct constructor from a residue modification
     explicit ModificationDefinition(const ResidueModification& mod, bool fixed = true, UInt max_occur = 0);
 
     /// destructor
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationDefinition.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationDefinition.h
@@ -67,7 +67,7 @@ public:
     explicit ModificationDefinition(const String& mod, bool fixed = true, UInt max_occur = 0);
 
     /// detailed constructor specifying the modification by name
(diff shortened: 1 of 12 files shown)


hendrikweisser:
More generally, we should consider replacing `vector<String> ProteinIdentification::SearchParameters::fixed_modifications` and `...::variable_modifications` with a ModificationDefinitionsSet object. That would avoid back-and-forth conversions between strings and pointers to ResidueModification in a number of places.

Opinions?

src/openms/include/OpenMS/CHEMISTRY/ModificationDefinition.h
```diff
@@ -63,8 +63,11 @@ namespace OpenMS
     /// copy constructor
     ModificationDefinition(const ModificationDefinition& rhs);
 
-    /// detailed constructor specifying the modifications name
-    explicit ModificationDefinition(const String& mod);
+    /// detailed constructor specifying the modification by name
+    explicit ModificationDefinition(const String& mod, bool fixed = true, UInt max_occur = 0);
+
+    /// detailed constructor specifying the modification by name
+    ModificationDefinition(const ResidueModification& mod, bool fixed = true, UInt max_occur = 0);
```
timosachsenberg: should be "explicit"

hendrikweisser: Why?

timosachsenberg: To prevent an implicit conversion from ResidueModification to ModificationDefinition. Happens for all single argument constructors and imho also for those with one non-default ones

hendrikweisser: Hm. What's the disadvantage of allowing this conversion? Both classes represent pretty much the same thing.

timosachsenberg: Implicit conversion is considered one of the worst sides of c++ as it produces very hard to debug errors. Actually I think cpplint will also put out a warning about it.

hendrikweisser: Okay, I've changed it.

timosachsenberg:
by the way nice PR.

hendrikweisser:
> by the way nice PR.

Thanks!

src/tests/topp/IDFileConverter_9_output.mzid
```diff
@@ -36,22 +36,22 @@
 	</DBSequence>
 	<Peptide id="PEP_14811341817612382122" name="PEPTIDERR">
 		<PeptideSequence>PEPTIDERR</PeptideSequence>
-	</Peptide> 
+	</Peptide>
  	<Peptide id="PEP_15050004366391181853" name="N[3899.3500281914]PEPTIDEK">
 		<PeptideSequence>NPEPTIDEK</PeptideSequence>
 		<Modification location="1" residues="N" monoisotopicMassDelta="3785.3071">
-			<cvParam cvRef="MS" accession="MS:1001460" name="unknown modification" value="N-Glycan"/>
```
timosachsenberg: is the loss of the value intended?

hendrikweisser: Yes. Our mzid implementation would add this for every "unknown modification", which doesn't make sense to me. See here: https://github.com/OpenMS/OpenMS/pull/2581/files#diff-2cd91736b9c4a6caf19b1ff80df95be3L924

timosachsenberg: Ok just wanted to make sure because I think @hroest recently added it

hendrikweisser: Ah, I see: #2561 
I still think it's wrong to add "N-Glycan" for every unknown modification, but maybe @hroest can comment on whether that was intentional.

hendrikweisser:
Apropos @hroest: I just noticed that I forgot to update the pyOpenMS pxds. (Every time!)

hendrikweisser:
> I just noticed that I forgot to update the pyOpenMS pxds.

How do I encode default parameters in Cython/autowrap? (E.g. for the ModificationDefinition constructors.)

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
```diff
@@ -921,7 +932,7 @@ namespace OpenMS
                         double diffmass = mod->getMonoMass() - jt->getSequence()[i].getMonoWeight();
                         p += "\" monoisotopicMassDelta=\"" + String(diffmass); 
                       }
-                      p += "\">\n\t\t\t<cvParam cvRef=\"MS\" accession=\"MS:1001460\" name=\"unknown modification\" value=\"N-Glycan\"/>";
+                      p += "\">\n\t\t\t<cvParam cvRef=\"MS\" accession=\"MS:1001460\" name=\"unknown modification\"/>";
```
hroest: :+1: yes, N-glycan should be added to every mod :-)

hroest:
> How do I encode default parameters in Cython/autowrap? (E.g. for the ModificationDefinition constructors.)

you cant right now, either you leave out the parameter altogether or you have to give them explictely. Another way to do it would be to declare multiple functions, one with and without the default parameter....

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
hendrikweisser: Did you mean "shouldn't"? Otherwise, why?

hendrikweisser:
> you cant right now, either you leave out the parameter altogether or you have to give them explictely.

What about this: http://cython.readthedocs.io/en/latest/src/reference/language_basics.html#optional-arguments
Is it not supported for pyOpenMS?

hendrikweisser:
I've updated the pxd files, but can't test them on this machine.

hroest:
https://github.com/hendrikweisser/OpenMS/pull/4

hendrikweisser:
Thanks for the pyOpenMS fixes, Hannes!

timosachsenberg:
+1

Issue #2534, opened 2017-04-20
TOPPAS/ExecutePipeline: problems with long paths


hendrikweisser:
TOPPAS and ExecutePipeline have a limit to the lengths of paths that they can handle (255 characters) and will quit with an error if a path gets too long. There are several problems with this:
1. The limitation to 260 characters comes from Windows and does not exist on Mac/Linux. (There, the length of a file name is limited to 256 characters, but paths can get much longer.) Therefore, it would be helpful if the length limit would only be enforced on Windows.
2. The limit applies to files created by TOPPAS/ExecutePipeline, but no attempt is made to create shorter directory or file names if the limit would be reached.
3. Most annoyingly, the file names created during the initial "dry run" - which can then lead to abortion of the workflow - can be much longer than the file names used during actual processing:
> Error: Unexpected internal error (the file '/hps/nobackup/research/proteomics/sanger-data/results/CHPP_TESTIS_Tricine_1006/TOPPAS_out/043-PeptideIndexer-out/CHPP_TESTIS_Tricine_1006.mzML.idXML_to_CHPP_TESTIS_Tricine_1006.mzML.idXML_merged.idXML.idXML.unknown.idXML.idXML.idXML.idXML.idXML.idXML.idXML' is too long (256 chars) and exceeds the allowed limit of 255. Use shorter filenames and/or less sub-directories.)

The actual name of the output file, if it could be created, would be simply "CHPP_TESTIS_Tricine_1006.idXML", not "CHPP_TESTIS_Tricine_1006.mzML.idXML_to_CHPP_TESTIS_Tricine_1006.mzML.idXML_merged.idXML.idXML.unknown.idXML.idXML.idXML.idXML.idXML.idXML.idXML".

hendrikweisser:
See also #2513.

hendrikweisser:
I've started working on this in a branch:
https://github.com/hendrikweisser/OpenMS/tree/fix-toppas-path-limit

So far I've addressed point 1 (by a compile-time check for the "_WIN32" macro).

Does anyone know where the code for the generation of the file names is (@cbielow)? It should be easy to address point 3 at least partially by not appending the same file extension repeatedly.

cbielow:
https://github.com/hendrikweisser/OpenMS/blob/0dadb2fb66f4453ca5adae135b777a4b1506cdb3/src/openms_gui/source/VISUAL/TOPPASToolVertex.cpp#L1027

cbielow:
[...]

PR #2582, opened 2017-05-03
Fixes for TOPPAS path length limit

hendrikweisser:
Addresses #2534, more specifically points 1 (limit can be relaxed on Linux/Mac) and 3 (repeated file extensions) in the issue comment.


cbielow:
nice.
remark: Any particular reason not to go with our Macro `OPENMS_WINDOWSPLATFORM`?
Its basically only an alias for `WIN32`, but it would be more consistent.

hendrikweisser:
> Any particular reason not to go with our Macro OPENMS_WINDOWSPLATFORM?

No, I just wasn't aware of it. It wasn't mentioned on StackOverflow. ;-)
I can replace it.

PR #2587, opened 2017-05-09
[FIX] Posterior Probability in PepXML for LuciphorAdapter

oliveralka:
LuciPhorAdapter (LuXor) needs the Posterior Probability / Phosphoprophet Probability for scoring. 
The score was not written in the PepXMLFile. This commit should fix the issue.


oliveralka:
tests missing - will work on it

hroest:
looks good :+1:

timosachsenberg:
@grosenberger does the extra analysis results make sense to you? I think you also used it before.

grosenberger:
Looks good. As far as I know LuciPHOr 2 only uses the ``probability`` attribute of the pepXML: https://github.com/dfermin/lucXor/blob/master/src/lucxor/PepXML.java

Issue #2531, opened 2017-04-19
FalseDiscoveryRate tool - add filter


timosachsenberg:
I know this might be a bit controversial but we repeatably observe users that apply the FalseDiscoveryRate tool but struggle with the IDFilter step afterward. 
Here, the user has to set score:pep to match a q-value threshold. This is poorly documented in the tool parameters alone because the score filter can be so generically applied to any type of (main) score.

One simple solution would be to add a filter option (maybe subsection "FDR control" and name them "peptide" and "protein") to FalseDiscoveryRate.

I'm not a fan of duplicating functionality in TOPP tools but in this case, I think this might be a good idea.

hroest:
sounds reasonable to me and filtering is not such a difficult task that there will be lots of duplicated code

:+1:

jpfeuffer:
see #2953

PR #2588, opened 2017-05-09
[PARAM,TOOL] FalseDiscoveryRate filter in tool

timosachsenberg:
- reworked parameters (changed flag to true / false string option with better logic)
- allow FDR control in FDR tool

### DIFF
--- a/src/topp/FalseDiscoveryRate.cpp
+++ b/src/topp/FalseDiscoveryRate.cpp
@@ -34,6 +34,7 @@
 
 #include <OpenMS/APPLICATIONS/TOPPBase.h>
 #include <OpenMS/ANALYSIS/ID/FalseDiscoveryRate.h>
+#include <OpenMS/FILTERING/ID/IDFilter.h>
 #include <OpenMS/KERNEL/StandardTypes.h>
 #include <OpenMS/FORMAT/IdXMLFile.h>
 #include <OpenMS/FORMAT/FileTypes.h>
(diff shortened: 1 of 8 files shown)


timosachsenberg:
adresses https://github.com/OpenMS/OpenMS/issues/2531

src/topp/FalseDiscoveryRate.cpp
```diff
@@ -171,6 +177,20 @@ class TOPPFalseDiscoveryRate :
...
+    if (psm_fdr >= 0)
+    {
+      LOG_INFO << "FDR control: Filtering PSMs..." << endl;
+      IDFilter::filterHitsByScore(pep_ids, psm_fdr / 100.0);
+    }
+
+    double protein_fdr = getDoubleOption_("FDR:protein");
+    if (protein_fdr >= 0)
+    {
+      LOG_INFO << "FDR control: Filtering proteins..." << endl;
```
timosachsenberg: I think correct would be to check if the score type is q-value? FDR filtering should work without recalculation if it already contains a q-value. Will add this.

timosachsenberg: On the other hand, I think I just filter if FDR has been calculated by the tool.

hroest:
```
The following tests FAILED:

	1781 - TOPP_FalseDiscoveryRate_4_out1 (Failed)

Submit files (using http)
```

timosachsenberg:
call for votes:
1. let the user specify FDR in percent or fraction?
@hroest @jpfeuffer @cbielow @hendrikweisser

jpfeuffer:
I think I would leave it as fraction -> No change and all the statistic tools (e.g. in R) do it, too

timosachsenberg:
agreed. I changed it to fraction and added range constraints.

timosachsenberg:
merge before or after source release?

mwalzer:
I'd say before source release. My rationale would be to, where CLI change is immediate anyway, include as much as possible.

hroest:
also :+1: for fractions, so should we merge?

PR #2553, opened 2017-04-26
Feature/sqlite fixid

hroest:
fixes an issue that came up when using pqp input files:

- transitions and peptide precursors had the same id
- when writing them out to mzML, we had chromatograms with the same id which could then not be mapped back any more to the pqp / traml file

the solution for this is to label precursor XIC with the tag "_Precursor_i0" at the end, which will also allow for XIC of other isotopes and makes 1:1 mappings possible

PR #2091, opened 2016-08-05
Feature/ms1 openswath

hroest:
allow OpenSwathWorkflow to perform MS1-only extraction and scoring (e.g. if no SWATH-MS data is present)

PR #2589, opened 2017-05-09
[FIX] OpenSWATH/IPF Precursor Fix

grosenberger:
This PR restores compatibility for OpenSWATH MS1 scoring when in parallel transition-level scoring is enabled. #2091 #2553


src/openms/source/ANALYSIS/OPENSWATH/MRMAssay.cpp
```diff
@@ -499,6 +499,7 @@ namespace OpenMS
     startProgress(0, TargetPeptideMap.size(), "Generation of target UIS assays");
 
     // Iterate over all target peptides
+    int transition_index = 0;
```
grosenberger: This fix resolves issues with redundant transitions.

hroest:
so what exactly happened, the identifiers were not unique and SQLite crashed?

src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
```diff
@@ -335,6 +335,12 @@ namespace OpenMS
         }
       }
 
+      for (typename std::vector<SpectrumType>::iterator pr_it = precursor_chromatograms_.begin(); pr_it != precursor_chromatograms_.end(); ++pr_it)
```
grosenberger: We need to copy the precursor chromatograms to the subset MRMTransitionGroup if transition-level scoring is enabled.

grosenberger:
There are two issues:

1. If transition-level scoring is enabled, the splitting of detection and identification transitions did not copy the precursor chromatograms.
2. In some cases, the generated IDs of transitions used for identification are not unique. By adding a serial number, they are guaranteed to be unique. This only affected TraML, as the SQLite PQP writer reindexes all transitions.

The failed tests are due to my Boost version, I will fix this.

src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
hroest: hmm basically you are controlling with tr_ids whether to take certain subgroups of transitions or not. Would it make sense to have a second parameter that controls whether to add the precursor (and if so, which precursors?)

grosenberger: This is a good suggestion and might be very useful for future extensions. I'm wondering how we should best incorporate this. Would we need specific precursor IDs apart from GroupID_Precursor_i0? At the moment these are always derived from the group name, right? Or is there a different function for MS1-only mode?

hroest: I think it goes beyond this PR but my idea was that we could store multiple isotopic traces and charge states as precursor signals ... maybe with their relative intensity? I am not sure about the naming yet ...

hroest:
regarding MS2_Precursor_i0 -> i thought these traces are too noisy and we usually filter them out? are the left in for UIS scoring but kept out for peak picking?

grosenberger:
This is correct, however for identification they can provide additional evidence, which is often even better than the precursor signals extracted from the MS1 maps.

src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
grosenberger: This would definitely be very useful and I guess that extension of the current transition list data structures should be feasible.

timosachsenberg:
ready for merge?

grosenberger:
Yes, from my side. We will implement the more thorough changes at a later stage, for now the patch only restores the previous functionality.

PR #2592, opened 2017-05-10
[FIX] throw exception if start and end position out of protein

timosachsenberg:
boundaries in coverage calculation

### DIFF
--- a/src/openms/source/FORMAT/FeatureXMLFile.cpp
+++ b/src/openms/source/FORMAT/FeatureXMLFile.cpp
@@ -1061,13 +1061,13 @@ namespace OpenMS
     // write peptide hits
     for (Size j = 0; j < id.getHits().size(); ++j)
     {
-      const PeptideHit & h = id.getHits()[j];
+      const PeptideHit& h = id.getHits()[j];
       os << indent << "\t<PeptideHit";
       os << " score=\"" << h.getScore() << "\"";
       os << " sequence=\"" << h.getSequence() << "\"";
       os << " charge=\"" << h.getCharge() << "\"";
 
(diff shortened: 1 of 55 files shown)


mwalzer:
nice

src/openms/source/FORMAT/FeatureXMLFile.cpp
```diff
@@ -1061,33 +1061,42 @@ namespace OpenMS
     // write peptide hits
     for (Size j = 0; j < id.getHits().size(); ++j)
     {
+      const PeptideHit & h = id.getHits()[j];
```
hendrikweisser: No space before "&".

src/openms/source/FORMAT/FeatureXMLFile.cpp
```diff
@@ -1061,33 +1061,42 @@ namespace OpenMS
...
       os << indent << "\t<PeptideHit";
-      os << " score=\"" << id.getHits()[j].getScore() << "\"";
-      os << " sequence=\"" << id.getHits()[j].getSequence() << "\"";
-      os << " charge=\"" << id.getHits()[j].getCharge() << "\"";
+      os << " score=\"" << h.getScore() << "\"";
+      os << " sequence=\"" << h.getSequence() << "\"";
+      os << " charge=\"" << h.getCharge() << "\"";
 
-      vector<PeptideEvidence> pes = id.getHits()[j].getPeptideEvidences();
+      const vector<PeptideEvidence> & pes = id.getHits()[j].getPeptideEvidences();
```
hendrikweisser: No space before "&".

timosachsenberg: ok can do.  Might be good to add this once and for all the cpplint script.

Issue #2343, opened 2017-01-06
Wrong annotated position in proteins lead to crash in MzTabExporter


lars20070:
The `MzTabExporter` crashes when exporting the following `featureXML` [1] with a segmentation fault.
```
MzTabExporter -in BM3415.featureXML -out BM3415.mzTab
```
[1] https://gist.github.com/lars20070/a6f4c0de17d20e1e6f97bada9f908fb5

timosachsenberg:
Hi @lars20070 can you provide a stack trace?

timosachsenberg:
```#0  0x00007ffff6a72a08 in std::pair<unsigned int const, OpenMS::DataValue>::pair (this=0x8e60b0)
    at /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/bits/stl_pair.h:68
#1  0x00007ffff6a72a77 in __gnu_cxx::new_allocator<std::pair<unsigned int const, OpenMS::DataValue> >::construct (this=0x7fffffff77b7, __p=0x8e60b0, __val=...)
    at /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/ext/new_allocator.h:105
#2  0x00007ffff6a72805 in std::_Rb_tree<unsigned int, std::pair<unsigned int const, OpenMS::DataValue>, std::_Select1st<std::pair<unsigned int const, OpenMS::DataValue> >, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, OpenMS::DataValue> > >::_M_create_node (this=0xde7310, __x=...)
    at /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/bits/stl_tree.h:371
#3  0x00007ffff6a722ad in std::_Rb_tree<unsigned int, std::pair<unsigned int const, OpenMS::DataValue>, std::_Select1st<std::pair<unsigned int const, OpenMS::DataValue> >, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, OpenMS::DataValue> > >::_M_clone_node (this=0xde7310, __x=0x3fc0000fe8ff0)
    at /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/bits/stl_tree.h:416
[...]

PR #2597, opened 2017-05-11
Fix handling of protein references when writing idXML

### LINKED ISSUE #2596: idXML: protein references written in wrong order or missing
When writing idXML, protein accessions for peptide hits are collected via `PeptideHit::extractProteinAccessions` (https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/IdXMLFile.cpp#L301). This function collects the accessions from all "peptide evidences" associated with the hit into a `set<String>`. Problematically, this operation:
1. potentially reorders the accessions relative to the order of the evidences;
2. removes duplicates (a peptide can match in different locations on the same protein).

This means that the references in the "protein_refs" attribute in the idXML may no longer correspond to the "aa_before"/"aa_after" and "start"/"end" attributes.

For peptides that match multiple times in the same protein, this can lead to crashes when converting the idXML to mzIdentML using IDFileConverter (because protein accessions are "missing" due to the de-duplication).

hendrikweisser:
Fixes #2596.

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
@@ -1081,7 +1081,17 @@ namespace OpenMS
             for (std::vector<PeptideEvidence>::const_iterator pe = peptide_evidences.begin(); pe != peptide_evidences.end(); ++pe)
             {
               String pevid =  "PEV_" + String(UniqueIdGenerator::getUniqueId());
-              String dBSequence_ref = String(sen_ids.find(pe->getProteinAccession())->second);
+              String dBSequence_ref;
+              map<String, String>::const_iterator pos = sen_ids.find(pe->getProteinAccession());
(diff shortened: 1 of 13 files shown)


timosachsenberg:
hi @hendrikweisser I have a PR open https://github.com/OpenMS/OpenMS/pull/2592

timosachsenberg:
also referenced from https://github.com/OpenMS/OpenMS/issues/2343

timosachsenberg:
may I take the fixed Fido files ;) I did also fix consensusXML and featureXML

timosachsenberg:
if its fine with you I make a merge of our two PRs?

hendrikweisser:
> if its fine with you I make a merge of our two PRs?

Yes, sure. This one can be closed then, right?

timosachsenberg:
Yes, you can close it now! Thanks, for taking care of this critical issue.

mwalzer:
nice

PR #2602, opened 2017-05-12
[ignore] ...

### LINKED ISSUE #2590: AccurateMassSearch defaults findfile mapping struct
I ran into an issue using AccurateMassSearch. It seems that the default values for the positive and negative adduct files are working fine CEMISTRY/... . But it is not able to find the files for the mapping and struct.

```
AccurateMassSearch of node #3 started. Processing ...
Error: File not found (the file 'CHEMISTRY/HMDBMappingFile.tsv' could not be found)
Cannot read input file given from parameter '-db:mapping'!
```

If the static path is used for mapping & struct - it works like a charm. 

May you can have a look on it.

cbielow:
[FIX] TOPPBase: honour 'skipexists' flag also for list of inputs; fixes #2590
[TEST] add TOPP test for AccurateMassSearch

other fixes which popped up after writing the test:
[FIX] fix segfault when number of handles is smalles than number of maps
[FIX] fix segfault in MzTabExport: when first CF gave no hit, the result is empty; using number of maps instead


cbielow:
something went wrong during push after rebase (I guess capitalization ams vs. AMS in branch name -- it does not update the PR...) will open a new one..

PR #2604, opened 2017-05-15
fixes to build.bat on Windows

cbielow:
[FIX] remove subprocess with low prio from build.bat since it caused trouble:
  - it does not allow to cancel the job on the console (it will keep running in the background)
  - it might trick MSBuild.exe into assuming only single-core CPU, even if /maxcpucount is specified
fixed now by just not spawning a child process

[FEATURE] convenience feature for Windows builds (to enable similar behaviour as in 'make ...') for MSVS generators.
before (in root binary dir): `$ ..\openms\tools\build.bat ....`
now: `$ build.bat ..`
This goes along nicely with other commands like `ctest ...` in the same directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,10 +269,14 @@ foreach(p LIB BIN INCLUDE CMAKE DOC SHARE)
 endforeach()
 
 #------------------------------------------------------------------------------
-# lib naming when building with msvc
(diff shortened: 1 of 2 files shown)


cbielow:
tests pass. merge?

jpfeuffer:
+1

Issue #2599, opened 2017-05-11
ConvertTraMLToTSV not in development branch?


PaulPyl:
Hi guys,

I have a quick question, I was looking into issues I had with ConvertTSVToTraML and the reverse and realised that it is in the master branch but not in the development branch. Have these tools been removed or did something go wrong? I can still find some tests for them in the development branch:

```
paul@Cthulhu:/deadpool/software/sources/OpenMS-GitHub$ grep -r "ConvertTSV" src/
src/tests/topp/CMakeLists.txt:  add_test("TOPP_ConvertTSVToTraML_test_1" ${TOPP_BIN_PATH}/TargetedFileConverter -test -in ${DATA_DIR_TOPP}/ConvertTSVToTraML_1_input.tsv -out ConvertTSVToTraML_output.TraML.tmp -out_type TraML)
src/tests/topp/CMakeLists.txt:  add_test("TOPP_ConvertTSVToTraML_test_1_out1" ${DIFF} -in1 ConvertTSVToTraML_output.TraML.tmp -in2 ${DATA_DIR_TOPP}/ConvertTSVToTraML_output.TraML)
[...]
```
Best,
Paul

grosenberger:
Hi Paul,

``ConvertTraMLToTSV`` and ``ConvertTSVToTraML`` have been unified in the tool ``TargetedFileConverter`` in OpenMS/develop.

Best regards,
George

PaulPyl:
I see, thanks for the quick reply. I have a related question now that I am using the TargetedFileconverter. I see the following error when trying to convert this ([iRTs.traml.zip](https://github.com/OpenMS/OpenMS/files/993476/iRTs.traml.zip)) traml file:

```
paul@Cthulhu:/deadpool/projects/paul/M3_OpenMS/MSFragger$ /deadpool/software/sources/OpenMS-Devel/bin/TargetedFileConverter -in iRTs.traml -out iRTs.tsv -debug 40
Debug level: 40
 >> /deadpool/software/sources/OpenMS-Devel/bin/TargetedFileConverter -in iRTs.traml -out iRTs.tsv -debug 40
Value of string option 'test': 0
Log file: 
Writing to ''
Debug level (after ini file): 40
Value of string option 'no_progress': 0
Value of string option 'in': iRTs.traml
Checking input file 'iRTs.traml'
Value of string option 'in_type': 
<Checking input file 'iRTs.traml'> occurred 2 times
Input file type: traML
[...]

PR #2607, opened 2017-05-17
[FIX] check whether CV term exists

hroest:


hroest:
prevents error and partly addresses #2599

PR #2615, opened 2017-05-19
[Feature] Chromatograms for FeatureFinderMetabo (mzML)

oliveralka:
Adds Chromatograms to FFM. 
Optional output as mzML. 
Visualisation with KNIME possible as far as I know.


timosachsenberg:
@hroest or @grosenberger any comments on how these chromatograms are stored. Would be nice to achieve some consistency (e.g., annotation of isolation windows) between OpenSWATH, FFMetabo, and maybe FFIdentification so the potential future code for visualization might be reused.

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -147,6 +150,26 @@ namespace OpenMS
     return tmp_hulls;
   }
 
+  std::vector< OpenMS::MSChromatogram <> > FeatureHypothesis::getChromatograms(UInt64 feature_id) const
+  {
+    std::vector< OpenMS::MSChromatogram <> > tmp_chromatograms;
+    for (Size mt_idx = 0; mt_idx < iso_pattern_.size(); ++mt_idx)
+    {
+      OpenMS::MSChromatogram<> chromatogram;
```
hroest: , I suggest to add some meta-information similar to what we do here: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ChromatogramExtractor.h#L252

e.g. add a precursor, set mz, set native id etc:

```
chromatogram.setNativeID(); // I suggest to match this to the feature id!
chromatogram.setChromatogramType(ChromatogramSettings::BASEPEAK_CHROMATOGRAM);
Precursor prec;
prec.setMZ();
chromatogram.setPrecursor(prec);

// Set the rest of the meta-data
chromatogram.setInstrumentSettings(settings.getInstrumentSettings());
chromatogram.setAcquisitionInfo(settings.getAcquisitionInfo());
chromatogram.setSourceFile(settings.getSourceFile());


for (Size j = 0; j < settings.getDataProcessing().size(); ++j)
{
  settings.getDataProcessing()[j]->setMetaValue("performed_on_spectra", "true");
  chromatogram.getDataProcessing().push_back(settings.getDataProcessing()[j]);
}
```

hroest:
yes, with these changes it should be possible to visualize them in TOPPView @oliveralka have you tried this, does itwork?

hroest:
@oliveralka can ypou add a black box tests that produces a chromatogram and compares it to the expected result?

oliveralka:
@hroest The visualisation in TOPPView works.  Will add a test.

oliveralka:
I would like to have the feature available asap - merge?

timosachsenberg:
@hroest we currently choose the same precursor (monoisotopic feature apex) for all its isotopic traces to bundle and visualize them in TOPPView. Is this the way to go? how would you store the actual m/z?

hroest:
TOPPView will basically group things together if they have the same
precursor, see src/openms_gui/source/VISUAL/SpectraViewWidget.cpp L619
where all things that have equal precursors are grouped together. So they
need to have the same m/z but also the same meta-values (e.g.
peptide_sequence). We could change that and group them based on different
criteria, e.g. based on the peptide_sequence meta value. But thats how it
is currently

Hannes

timosachsenberg:
ok I think we can merge these changes now and adapt the way how we group chromatograms in TOPPView in a subsequent PR.

PR #2617, opened 2017-05-19
[TEST] fix tests, removed PPW test

timosachsenberg:


hroest:
@timosachsenberg what about TwoDOptimization_test.cpp ?

PR #2620, opened 2017-05-21
[TOOL] removed MapAlignmentEvaluation util

timosachsenberg:


hroest:
fine with me, I think it was just used for the that one publication. :+1: for removal

PR #2621, opened 2017-05-21
Feature/metabolite decharger

FabianAicheler:
Adduct grouping/decharging for small molecules. (Currently) branched of clone of the Decharger (with more fitting default settings for metabolomics).

This serves as an initial base utility for small molecule decharging people can publicly be referred to, both for OpenMS as well as KNIME (as the Decharger wasn't well known to users.)

We will keep this as a util for now. End of May, we will travel to Finland to discuss possible adaptions with interested cooperation partners, and work them into the util (both functionality and result representation). 

For that reason, we would like to have the util in KNIME, so if any steps are necessesary besides merge in develop (after successful build tests), please let me know.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.h
+++ b/src/openms/include/OpenMS/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.h
@@ -1,0 +1,163 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2016.
+//
+// This software is released under a three-clause BSD license:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of any author or any participating institution
+//    may be used to endorse or promote products derived from this software
+//    without specific prior written permission.
+// For a full list of authors, refer to the file AUTHORS.
+// --------------------------------------------------------------------------
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL ANY OF THE AUTHORS OR THE CONTRIBUTING
+// INSTITUTIONS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Fabian Aicheler $
+// $Authors: $
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_ANALYSIS_DECHARGING_METABOLITEFEATUREDECONVOLUTION_H
+#define OPENMS_ANALYSIS_DECHARGING_METABOLITEFEATUREDECONVOLUTION_H
+
+// OpenMS
+#include <OpenMS/KERNEL/FeatureMap.h>
+#include <OpenMS/KERNEL/ConsensusMap.h>
+#include <OpenMS/ANALYSIS/DECHARGING/ILPDCWrapper.h>
+#include <OpenMS/DATASTRUCTURES/DPosition.h>
+#include <OpenMS/DATASTRUCTURES/MassExplainer.h>
+
+#include <OpenMS/DATASTRUCTURES/DefaultParamHandler.h>
+
+namespace OpenMS
+{
+
+  class Compomer;
+
+  /**
+    @brief An algorithm to decharge features (i.e. as found by FeatureFinder).
+
+    @htmlinclude OpenMS_FeatureDeconvolution.parameters
+
+    @ingroup Analysis
+  */
+
+  class OPENMS_DLLAPI MetaboliteFeatureDeconvolution :
+    public DefaultParamHandler
+  {
+public:
+
+    enum CHARGEMODE {QFROMFEATURE = 1, QHEURISTIC, QALL};
+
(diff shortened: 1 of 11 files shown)


src/utils/MetaboliteAdductDecharger.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    //-------------------------------------------------------------
+    // calculations
+    //-------------------------------------------------------------
+    ConsensusMap cm, cm2;
+    StopWatch a;
+    a.start();
+    fdc.compute(map_in, map_out, cm, cm2);
+    a.stop();
+    //std::cerr << "took: " << a.getClockTime() << " seconds\n\n\n";
```
hroest: can you remove that timing / replace with ProgressLogger ?

src/openms/include/OpenMS/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.h
```diff
@@ -0,0 +1,163 @@
...
+
+    @ingroup Analysis
+  */
+
+  class OPENMS_DLLAPI MetaboliteFeatureDeconvolution :
+    public DefaultParamHandler
+  {
+public:
+
+    enum CHARGEMODE {QFROMFEATURE = 1, QHEURISTIC, QALL};
```
hroest: why do start at 1 (i think default is start at zero?)

src/openms/source/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.cpp
```diff
@@ -0,0 +1,1209 @@
...
+    defaults_.setValue("retention_max_diff_local", 1.0, "Maximum allowed RT difference between between two co-features, after adduct shifts have been accounted for (if you do not have any adduct shifts, this value should be equal to 'retention_max_diff', otherwise it should be smaller!)");
+    /// TODO should be m/z ppm?!
+    defaults_.setValue("mass_max_diff", 0.05, "Maximum allowed mass difference [in Th] for a single feature.");
+    // Na+:0.1 , (2)H4H-4:0.1:-2:heavy
+    defaults_.setValue("potential_adducts", ListUtils::create<String>("H+:0.9"), "Adducts used to explain mass differences in format: 'Element(+)*:Probability[:RTShift[:Label]]', i.e. the number of '+' indicate the charge, e.g. 'Ca++:0.5' indicates +2. Probabilites have to be in (0,1]. RTShift param is optional and indicates the expected RT shift caused by this adduct, e.g. '(2)H4H-4:1:-3' indicates a 4 deuterium label, which causes early elution by 3 seconds. As a fourth parameter you can add a label which is tagged on every feature which has this adduct. This also determines the map number in the consensus file.");
```
hroest: wouldnt it make sense to have a few good adducts as defaults here (in addition to H+) ?

src/openms/source/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.cpp
```diff
@@ -0,0 +1,1209 @@
...
+      StringList adduct;
+      it->split(':', adduct);
+      if (adduct.size() != 2 && adduct.size() != 3 && adduct.size() != 4)
+      {
+        String error = "MetaboliteFeatureDeconvolution::potential_adducts (" + (*it) + ") does not have two, three or four entries ('Element:Probability' or 'Element:Probability:RTShift' or 'Element:Probability:RTShift:Label'), but " + String(adduct.size()) + " entries!";
+        throw Exception::InvalidParameter(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, error);
+      }
+      // determine charge of adduct (by # of '+')
+      Size l_charge = adduct[0].size();
+      l_charge -= adduct[0].remove('+').size();
```
hroest: what if we have negative mode data and the adduct is negative?

src/openms/source/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.cpp
```diff
@@ -0,0 +1,1209 @@
...
+      for (Size i_RT_window = i_RT + 1
+           ; (i_RT_window < fm_out.size())
+          && ((fm_out[i_RT_window].getRT() - fm_out[i_RT].getRT()) <= rt_diff_max)
+           ; ++i_RT_window)
+      { // ** RT-window
+
+        // knock-out criterion first: RT overlap
+        // use sorted structure and use 2nd start--1stend / 1st start--2ndend
+        const Feature& f1 = fm_out[i_RT];
+        const Feature& f2 = fm_out[i_RT_window];
```
hroest: how would this work with a negative RT shift, e.g. if the adduct causes it to shift back in RT ?

src/openms/source/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.cpp
```diff
@@ -0,0 +1,1209 @@
...
+        {
+          double f_start1 = std::min(f1.getConvexHull().getBoundingBox().minX(), f2.getConvexHull().getBoundingBox().minX());
+          double f_start2 = std::max(f1.getConvexHull().getBoundingBox().minX(), f2.getConvexHull().getBoundingBox().minX());
+          double f_end1 = std::min(f1.getConvexHull().getBoundingBox().maxX(), f2.getConvexHull().getBoundingBox().maxX());
+          double f_end2 = std::max(f1.getConvexHull().getBoundingBox().maxX(), f2.getConvexHull().getBoundingBox().maxX());
+
+          double union_length = f_end2 - f_start1;
+          double intersect_length = std::max(0., f_end1 - f_start2);
+
+          if (intersect_length / union_length < rt_min_overlap)
```
hroest: even better would be to require a minimum  pearson correlation of the chromatographic peak XIC to ensure that they are the same peak (not just at the same RT)

src/openms/source/ANALYSIS/DECHARGING/MetaboliteFeatureDeconvolution.cpp
```diff
@@ -0,0 +1,1209 @@
...
+                }
+            }
+            if (i_RT==429 && i_RT_window==432)
+            {
+                std::cout << "DEBUG reached\n hits: " << hits << " with delta_m: " << naive_mass_diff << " and thres: " << thresh_logp << "\n";
+            }
+*/
+
+            overallHits += hits;
+            // choose most probable hit (TODO think of something clever here)
```
hroest: one idea could be to have weighted edges, the weight could be delta-RT or a pearson or cross-correlation score of the XIC

src/tests/class_tests/openms/source/MetaboliteFeatureDeconvolution_test.cpp
```diff
@@ -0,0 +1,245 @@
...
+      /// labeling table
+      Map<Size, String> getMapLabels()
+      { return map_label_;}
+
+      /// labeling table inverse
+      Map<String, Size> getMapLabelInverse()
+      { return map_label_inverse_;}
+
+			/// status of intensity filter for edges
+			bool isIntensityFilterEnabled()
```
hroest: what happened to the indentation?

hroest:
looks like a good first stab at the problem, thanks a lot. Can you add a few black box tests for the util itself to `./src/tests/topp/CMakeLists.txt` just to make sure that it runs and produces reasonable output with different parameters?

also the algo itself looks a bit rough around the edges, e.g. having lots of debug and `#if 1` stuff in there is not perfect but it would work for an initial try. Also I have some questions regarding negative mode data and potential negative RT shifts which I am not sure whether they are accounted for in the algo. I also have some ideas for improvements, e.g. adding edge weights but that is open for discussion.

FabianAicheler:
Thanks for the feedback/referal to important parts of the code.

Yes, negative mode is currently not implemented in Decharger and this clone of it. It is one important thing we want to try to include over this week before the trip to Finland.

The other stuff is also part of our agenda and we will think about it.

As of this moment, the algorithm should correspond exactly to the Decharger (which is part of the TOPP tools). We'll try of course to extend the number of tests, but at least for the initial inclusion it is further down on our list (as it has been provided/used in this form for years).

FabianAicheler:
Negative mode will follow as deriving pull request.
Same for cosmetic changes, will be a separate feature/pull request.
Adaptions for edge weights/ RT criteria will be discussed in Finland over the next few days.

Issue #2613, opened 2017-05-18
TheoreticalSpectrumGenerator charge parameter is not handled by ParamHandler


mwalzer:
`getSpectrum` parameters `min_charge` and `max_charge` are not handled by ParamHandler like all other parameters.

enetz:
You mean the charges should just be handled by the ParamHandler, like e.g. ion types?
max_charge was already there before and was never handled by the ParamHandler,
@timosachsenberg was there a reason for that?
I agree, I do not see any reason for special treatment of these parameters.

timosachsenberg:
hmm I guess the performance overhead via parameters is too high

timosachsenberg:
note that the other stuff is set only once but charges depend on actual precursor charge

timosachsenberg:
also note that this is hard to get these changes in parameter thread safe so I would be against using param

timosachsenberg:
So to answer your question: I think it is by design to allow using TSG in a multi-threaded context without locks. I would keep it as it is.

timosachsenberg:
close?

mwalzer:
Hadn't had any chance to respond yet

mwalzer:
So, while I agree that it's a hassle to get changes needed often in, I think setting the defaults like that would be good.

mwalzer:
Or have no defaults for min and max charge.

timosachsenberg:
I think it would then be better to have no defaults.

timosachsenberg:
fixed with https://github.com/OpenMS/OpenMS/pull/2622

PR #2622, opened 2017-05-22
TSG without charge defaults

enetz:
as discussed in #2613 

Removed charge default values for getSpectrum() and set mincharge = 1 and maxcharge = 1 everywhere getSpectrum() was called without specifying the charge.


timosachsenberg:
+1

hroest:
@enetz shouldnt we also test the function for other values of min/max charge?

enetz:
@hroest  I only set the charge to 1 in those cases, where none was specified before, meaning where the previous default values were used. Of course we already had tests for other charges and charge ranges. Since the charge ranges were already explicitly set in those cases, I did not change them in this PR.

hroest:
ok, great!

Issue #2562, opened 2017-04-27
Enforce certain include/linking orders


jpfeuffer:
Assume you have boost and eigen installed in the system paths (let's say /usr/local/include).
Also both in the contrib (/Users/user/contrib-build/include).
If they do not provide subdirectories (which some of them do not), it means that whatever library is included first in CMake:
https://github.com/jpfeuffer/OpenMS/blob/2080ba2f015ab28d0b6e31063fc2f181a8edae75/src/openms/CMakeLists.txt#L118-L118
determines both of their fate because it implicitly includes the other at the same time and comes first in the compiler command when creating the objects.

Not sure if we can do much about it. Maybe we should check if any of the includes directly lies in a system path and put it last.

This might help:
http://stackoverflow.com/questions/7752609/cmake-ordering-of-include-directories-how-to-mix-system-and-user-based-includ

jpfeuffer:
Maybe we also need to force our contrib to lay the libraries/includes into subdirectories.

cbielow:
where is the problem here? As far as I understand, `include_directory()` is always preferred over standard system includes. Hence, if we include the contrib like that, our contrib will be preferred (if the libs were build) -- and that sounds like a reasonable thing to do.
Is there a case where this goes wrong?

jpfeuffer:
Ok, I try again. Not sure if there is really a problem but I will try :)
Assume you have boost only in system and boost is first in the EXTERNAL_INCLUDES of the OpenMS library.
But you also have eigen in system. AND in contrib. You want to use eigen from contrib.
The problem is, if boost and eigen are not in subdirectories under /usr/local/include,
then /usr/local/include is included when boost is included and automatically includes the eigen headers with it.
Yes CMake will add /contrib-build/include as another directory, but the compiler will probably choose the headers from the first include, which is from the system.
-> CMake says it found the contrib Eigen but the system Eigen is actually included

hroest:
even worse, from what I have seen is that the compiler will include the *headers* from /usr/local/include but try to link against the files in your contrib. At least this is what happened in #2558 and was no fun

hroest:
I wrote a lengthy analysis over here https://github.com/OpenMS/OpenMS/issues/2558#issuecomment-298067569 it really looks like a CMake issue on OSX only.

[...]

Issue #2558, opened 2017-04-26
Build problems on OSX 10.12 


christineyyeh:
I just ran into a suite of problems compiling the developer version of OpenMS on OS X 10.12. 

**1. The first problem was finding the right version of qt.** 
I found out that OS X 10.12 has retired qt4 due to security issues and you cannot access it with:

`$ brew install qt`

You can now only access qt4 in cartr/qt4/qt@4 through homebrew with the --force flag. Since I then had both versions of qt, I also had to make sure that the /usr/local/bin/qt linked to this version of qt before calling cmake. 

After fixing the version issue, I found that even with the force download, the cmake could not find a field to go in QT_QTWEBKIT_INCLUDE_DIR=PATH. This was because the webkit has been retired and no longer comes with qt4 either. So then, I ran: 

`$ brew install qt-webkit@2.3`

which retrieved the webkit files from the Homebrew boneyard. It came with some warning messages about compromised security and video not working. The cmake command still could not find it initially - so then I specified the QT_QTWEBKIT_INCLUDE_DIR=PATH flag in the cmake command. 

**2. Problems concerning other dependencies**

It seemed that during the configuration, the cmake found paths to newer versions of: seqan, glpk, eigen3 that I installed with homebrew rather than in the ../contrib_build

I uninistalled these one by one. Not really sure why cmake is looking for these dependencies in /usr/local/include

**3. Linker problem with Boost** 

I then got some linker error with boost when linking the OpenMS library itself: 

```
Undefined symbols for architecture x86_64:
  "boost::re_detail_106400::perl_matcher<char const*, std::__1::allocator<boost::sub_match<char const*> >, boost::regex_traits<char, boost::cpp_regex_traits<char> > >::construct_init(boost::basic_regex<char, boost::regex_traits<char, boost::cpp_regex_traits<char> > > const&, boost::regex_constants::_match_flags)", referenced from:
      bool boost::regex_search<char const*, std::__1::allocator<boost::sub_match<char const*> >, char, boost::regex_traits<char, boost::cpp_regex_traits<char> > >(char const*, char const*, boost::match_results<char const*, std::__1::allocator<boost::sub_match<char const*> > >&, boost::basic_regex<char, boost::regex_traits<char, boost::cpp_regex_traits<char> > > const&, boost::regex_constants::_match_flags, char const*) in IndexedMzMLDecoder.cpp.o
[...]

Issue #2549, opened 2017-04-24
FFSuperHirn segfaults on build machines


jpfeuffer:
I wanted to fix the ...XMLFile_tests from e.g. http://cdash.openms.de/testDetails.php?test=39575898&build=220777
 where I wanted to do a "make install" to overwrite the global outdated OpenMS/share.

However, FFSuperhirn segfaults when you execute the binary (e.g. when it should print the parameter documentation etc).

MacOSX Lion. gcc 5.2.

gdb StackTrace:
```Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_INVALID_ADDRESS at address: 0x0000000000000000
0x00000001001376a3 in xercesc_3_1::XMLString::replicate ()
(gdb) backtrace
#0  0x00000001001376a3 in xercesc_3_1::XMLString::replicate ()
#1  0x000000010003237b in xercesc_3_1::ENameMap::ENameMap ()
#2  0x00000001000329e9 in xercesc_3_1::ENameMapFor<xercesc_3_1::XMLChTranscoder>::ENameMapFor ()
#3  0x000000010192c1ff in xercesc_3_1::XMLTransService::initTransService ()
#4  0x000000010190d9a3 in xercesc_3_1::XMLPlatformUtils::Initialize ()
#5  0x00000001012a84b4 in OpenMS::Internal::XMLFile::parse_ (this=0x7fff5fbfc940, filename=@0x7fff5fbfca20, handler=0x7fff5fbfc5e0) at /Users/aiche/NIGHTLY/HEAD/src/openms/source/FORMAT/XMLFile.cpp:118
#6  0x00000001012676df in OpenMS::ParamXMLFile::load (this=0x7fff5fbfc940, filename=@0x7fff5fbfca20, param=@0x7fff5fbfc9b0) at /Users/aiche/NIGHTLY/HEAD/src/openms/source/FORMAT/ParamXMLFile.cpp:366
#7  0x0000000100fbaa2a in OpenMS::ResidueDB::readResiduesFromFile_ (this=0x10603cc00, file_name=@0x7fff5fbfcd00) at /Users/aiche/NIGHTLY/HEAD/src/openms/source/CHEMISTRY/ResidueDB.cpp:196
[...]

PR #2623, opened 2017-05-22
[FIX] Added CMAKE param to explicitly set path to contrib 

jpfeuffer:
which always comes first. #2562 #2558
It also starts to use CMake3.0 features to get rid of some of our CMake export magic and eventually solve #2549 and the general problem of interfacing included libraries.
The thing is that we specify a minimum CMake version for every script. But how often does one use the child CMakes separately from the parent one?


cmake/setup_lib_find_paths.cmake
```diff
@@ -38,16 +38,19 @@
...
-endif()
 
 #------------------------------------------------------------------------------
-# add contrib location of source package <source>/contrib
-list(APPEND CMAKE_PREFIX_PATH ${OPENMS_HOST_DIRECTORY}/contrib/)
+# Specify the path to the contrib-build. Will be added to CMAKE_PREFIX_PATH
+# for searching and as first entry in the includes/libraries to avoid
+# mismatches with installed system libraries
+if(NOT OPENMS_CONTRIB_LIBS)
+  message(FATAL_ERROR "Please specify an explicit path to the built contrib libraries via
```
hroest: I dont think making this a fatal error is a good idea - distros for example will not want this at all as libOpenMS should link only to system libraries! So I suggest to change this to a warning ...

src/openms/CMakeLists.txt
```diff
@@ -108,15 +108,17 @@ if (WITH_CRAWDAD)
...
                    HEADER_FILES  ${OpenMS_sources_h}
                                  ${OpenMS_configured_headers}
-                   INTERNAL_INCLUDES ${PROJECT_SOURCE_DIR}/include
-                                     ${PROJECT_BINARY_DIR}/include
+                   INTERNAL_INCLUDES ${CMAKE_CURRENT_SOURCE_DIR}/include
+                                     ${CMAKE_CURRENT_BINARY_DIR}/include
 									 PRIVATE_INCLUDES ${EOL_BSPLINE_INCLUDE_DIRECTORY}
-                   EXTERNAL_INCLUDES ${QT_INCLUDES}
-                                     ${GSL_INCLUDE_DIRS}
+                   EXTERNAL_INCLUDES ${OPENMS_CONTRIB_LIBS}/include
```
hroest: I guess this may not work if the parameter is optional

tools/travis/cibuild.cmake
```diff
@@ -13,7 +13,8 @@ message(STATUS "CTEST_SOURCE_DIRECTORY: ${CTEST_SOURCE_DIRECTORY}")
 message(STATUS "CTEST_BINARY_DIRECTORY: ${CTEST_BINARY_DIRECTORY}")
 
 set(INITIAL_CACHE
-"CMAKE_PREFIX_PATH=$ENV{SOURCE_DIRECTORY}/contrib\;$ENV{OS_PREFIX_PATH}
+"CMAKE_PREFIX_PATH=$ENV{OS_PREFIX_PATH}
+OPENMS_CONTRIB_LIBS=$ENV{SOURCE_DIRECTORY}/contrib
```
hroest: this is not what we tell people to do  in our install instructions!

tools/travis/lnx-cibuild.before.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+mkdir cmake_tmp
+cd cmake_tmp
+wget --no-check-certificate https://cmake.org/files/v3.0/cmake-3.0.2-Linux-i386.tar.gz
+tar xf cmake-3.0.2-Linux-i386.tar.gz
+sudo cp -r cmake-3.0.2-Linux-i386/* /usr
```
hroest: somehow that did not work on travis

tools/travis/cibuild.cmake
jpfeuffer: Now in the docs.

cmake/add_library_macros.cmake
```diff
@@ -167,9 +172,9 @@ function(openms_add_library)
   #------------------------------------------------------------------------------
   # Link library against other libraries
   if(openms_add_library_LINK_LIBRARIES)
-    ## check for consistent lib arch (e.g. all 64bit)?
+    ## check for consistent lib arch (e.g. all 64bit?)
     check_lib_architecture(openms_add_library_LINK_LIBRARIES)
-    target_link_libraries(${openms_add_library_TARGET_NAME} ${openms_add_library_LINK_LIBRARIES})
+    target_link_libraries(${openms_add_library_TARGET_NAME} PRIVATE ${openms_add_library_LINK_LIBRARIES})
```
hroest: is the PRIVATE causing the compilation issues on cdash?

hroest:
great, I like it. Seems there are some linking issues on cdash, possibly because of the private keyword?

hroest:
PS: does this require cmake 3 or is it just a convenient time to make the switch?

cmake/add_library_macros.cmake
jpfeuffer: It could be, yes. I will investigate. In the end I want to make as much PRIVATE as possible, such that libraries that are only used and not interfaces by us are not passed to libraries linking against us.

jpfeuffer:
Both IMHO. The finer grained library dependency specifications (e.g. PRIVATE) and associations to targets (e.g. target_link_libraries) actually require CMake 3.0. But it is also a good time because it will make the QT5 inclusion and C++11 handling more convenient.

timosachsenberg:
win and macOS are now failing after the C11 PR merge. Before I undo it I think it might be worth checking if this is just because the cmake version is too old. http://cdash.openms.de/index.php?project=OpenMS

hroest:
@jpfeuffer  can we also merge part of this into master? at least the option to set the path? because many people will try to build from source and it would be annoying to miss out on this little piece of good stuff...

jpfeuffer:
Oh.. why was this merged at all? But it seems like the error was resolved in Travis, as newer PRs do succeed? I mean, as I recall, it went through on my Mac (including the Superhirn stuff). Maybe it was a hiccup or a problem from another PR? Anyways. I will cherry-pick the commits related to the new variable.

jpfeuffer:
Ah it was CDash that was failing. Too many merges currently. We need to get CDash back on track.

Issue #2608, opened 2017-05-17
& not properly escaped in featureXML


timosachsenberg:
& is not properly escaped in featureXML
10:54:34 NOTICE: TextExporter of node #10 started. Processing ...
..\..\..\OpenMS\src\openms\source\FORMAT\HANDLERS\XMLHandler.cpp(103): While loading 'X.featureXML': unterminated entity reference 'coxG:'( in line 3397 column 70)

```
			<ProteinHit id="PH_670" accession="ABV95450.1|Dshi_3719|coxL&coxG:" score="0.943420648574829" sequence="">
				<UserParam type="string" name="isDecoy" value="false"/>
				<UserParam type="string" name="target_decoy" value="target"/>
				<UserParam type="float" name="_score" value="0"/>
			</ProteinHit>
```

hroest:
related to #24 , try to wrap it in `writeXMLEscape()`

timosachsenberg:
fixed in  #2626

PR #2626, opened 2017-05-23
[FIX] fix errors caused by & in protein accession

timosachsenberg:


timosachsenberg:
will add some additional escaping

mwalzer:
#2608

src/tests/topp/IDFileConverter_16_output.idXML
```diff
@@ -39,7 +39,7 @@
 			</PeptideHit>
 		</PeptideIdentification>
 		<PeptideIdentification score_type="InterProphet probability" higher_score_better="true" significance_threshold="0" MZ="566.77033723893" RT="0" >
-			<PeptideHit score="0.0505509" sequence=".(Glu->pyro-Glu)EAGIVDELAYA" charge="2" aa_before="K" aa_after="V" protein_refs="PH_4" >
+			<PeptideHit score="0.0505509" sequence=".(Glu-&gt;pyro-Glu)EAGIVDELAYA" charge="2" aa_before="K" aa_after="V" protein_refs="PH_4" >
```
hroest: that was probably not a good idea to have these characters as part of the  official openms names ...

timosachsenberg: they come from Unimod so we probably should blame them ;)

timosachsenberg:
everyone fine with merge to master?

jpfeuffer:
+1

PR #2627, opened 2017-05-23
[FIX] reading of old CSV/TraML files

hroest:
- fixes error when reading old CSV files
- when peptide sequences did not match, the modification was placed at
  the wrong residue
- allows override with flag


hroest:
using the force flag, we can now read and work with libraries that were generated with OpenMS < 2.1

timosachsenberg:
should we merge this to master?

hroest:
yes, please

timosachsenberg:
ok I will cherry-pick this commit and apply to master

hroest:
great, thanks!

On Sun, May 28, 2017 at 12:08 PM, Timo Sachsenberg <notifications@github.com

PR #2637, opened 2017-05-29
[FIX] reading of old CSV/TraML files

timosachsenberg:
- fixes error when reading old CSV files
- when peptide sequences did not match, the modification was placed at
  the wrong residue
- allows override with flag


timosachsenberg:
cherry-pick of https://github.com/OpenMS/OpenMS/pull/2627

PR #2638, opened 2017-05-29
[FIX] fix error with CMake 3.8.1

cbielow:
-- Compiler checks for conversion: OFF
CMake Error at CMakeLists.txt:278 (file):
  file COPY cannot copy file "C:/dev/OpenMS/tools/build.bat" to
  "C:/dev/OpenMS_build/build.bat/build.bat".
-- Boost version: 1.61.0


jpfeuffer:
So they changed the file COPY command? Do you know if it is backwards compatible?

cbielow:
the cmake docu implies that it should work for all cmake version (they all support <dir> as DESTINATION, see https://cmake.org/cmake/help/v3.8/command/file.html). The fact that it worked with a file is actually more baffling....

cbielow:
also works for older cmake, e.g. 2.8 (https://cmake.org/cmake/help/v2.8.1/cmake.html --> `file(<COPY|INSTALL> files... DESTINATION <dir>...)`

timosachsenberg:
applied to master as well

PR #2499, opened 2017-04-04
Feature/spectrum hit annotation/branchof/general spectrum reference annotation

mwalzer:
This is the part of split rework of #1797 , closed after scope creep. 
It is based on #249 (suggested merge after). Following the introduced SpectrumMetaDataLookup changes, this introduces SpectrumAnnotator, using TheoreticalSpectrumGenerator and SpectrumAlignment to match and annotate given identifications with given peak data at a given tolerance. Which ion types to mach is controlled via the Param extension. Also some match statistics are added.
Annotation looks something like:
```
<UserParam type="string" name="matched_ions" value="y6+,b13+,y9+,b2+,a11+,y5+,y4+,y3+,a12-C1H2N2+,y10-C1H4O1S1+,c5-H2O1+,b9++,c10-H3N1++,b11+,b12+,b15+,y10+,b9+,c10-H3N1+,..."/>
<UserParam type="float" name="matched_intensity" value="100662.793632507"/>
<UserParam type="int" name="matched_ion_number" value="229"/>
<UserParam type="int" name="peak_number" value="440"/>
<UserParam type="float" name="sum_intensity" value="159149.309085846"/>
<UserParam type="float" name="NTermIonCurrentRatio" value="0.300109904676478"/>
<UserParam type="float" name="CTermIonCurrentRatio" value="0.202220592085132"/>                                <UserParam type="float" name="median_fragment_error" value="0.0183461949475543"/>
<UserParam type="float" name="IQR_fragment_error" value="0.159449305250064"/>
<UserParam type="float" name="topN_meanfragmenterror" value="0.00127883276817491"/>
...
```

PR #2642, opened 2017-05-29
Feature/spectrum hit annotation/branchof/general spectrum reference annotation

timosachsenberg:


timosachsenberg:
smaller changes to https://github.com/OpenMS/OpenMS/pull/2499

timosachsenberg:
@enetz can you take a quick look at the failing test
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=225363
triggers:
```
Error: Unexpected internal error (!theoretical_spec.isSorted() TheoreticalSpectrumGenerator::getSpectrum did not yield a sorted spectrum!)
```

enetz:
hmm, in MSSpectrum the sorting is done based on position values from ContainerType::operator[](i).getPosition(), while isSorted checks if (this->operator[](i - 1).getMZ() > this->operator[](i).getMZ())

is it possible to get different values from ContainerType::operator[](i).getPosition() and this->operator[](i).getMZ() ?
In Peak1D, getMZ() and getPos() return position_[0], but getPosition() returns "the whole" position_

The only reason I can think of, why the spectrum may actually not be sorted, is if the spectrum size is not equal to one of the DataArrays, but that should throw an exception in the MSSpectrum::select() function.
Are there any "IntegerDataArray[0] size (X) does not match spectrum size (Y)" or similar exceptions or warnings?
Does this only happen to the Travis build or also on your local build?

timosachsenberg:
I paste some code here as a notepad ;)
sorting without vs. with data array:
```
  std::sort(ContainerType::begin(), ContainerType::end(), typename PeakType::PositionLess());
vs.
       std::sort(sorted_indices.begin(), sorted_indices.end(), PairComparatorFirstElement<std::pair<typename PeakType::PositionType, Size> >());
```

timosachsenberg:
ok I think we just need to negate the PRECONDITION :P

mwalzer:
:tada:  :+1:

hroest:
> ok I think we just need to negate the PRECONDITION :P

haha I like your approach....

Issue #2644, opened 2017-05-29
Error in XQuestScores on Windows


jpfeuffer:
Shadowed in the nightlies by the general C++11 failure.

```
3>..\..\..\OpenMS\src\openms\source\ANALYSIS\XLMS\XQuestScores.cpp(109): error C2668: 'pow' : ambiguous call to overloaded function
3>          c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\include\math.h(497): could be 'double pow(double,int)'
3>          c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\include\math.h(122): or       'double pow(double,double)'
3>          while trying to match the argument list '(double, OpenMS::Size)'
3>..\..\..\OpenMS\src\openms\source\ANALYSIS\XLMS\XQuestScores.cpp(113): error C2668: 'pow' : ambiguous call to overloaded function
3>          c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\include\math.h(497): could be 'double pow(double,int)'
3>          c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\include\math.h(122): or       'double pow(double,double)'
3>          while trying to match the argument list '(double, OpenMS::Size)'
```

jpfeuffer:
So this is solved right? Please reopen if not.

enetz:
This should be solved by #2650 , but I could not reproduce the error locally, so I could not test it directly.

PR #2443, opened 2017-03-03
Protein-Protein XLMS Part II: OpenProXL, XL-MS search engine

enetz:
Includes two utils, OpenProXL and OpenProXLLF (for labeled and label-free cross-linkers),
and TheoreticalSpectrumGenerators for XL-MS.

A minor error in IDMapper is also fixed.

PR #2650, opened 2017-05-30
[FIX] fixed ambiguous function calls with casts and removed C11 code

enetz:
adresses #2644 and a remaining issue in #2443

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
+++ b/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
@@ -133,7 +133,7 @@ namespace OpenMS
        * @param scan_index The index of the current spectrum
        * @param scan_index_heavy The index of the heavy spectrum in a spectrum pair with labeled linkers
        */
-      static void buildPeptideIDs(std::vector<PeptideIdentification> & peptide_ids, const std::vector< OPXLDataStructs::CrossLinkSpectrumMatch > & top_csms_spectrum, std::vector< std::vector< OPXLDataStructs::CrossLinkSpectrumMatch > > & all_top_csms, Size all_top_csms_current_index, const PeakMap & spectra, Size scan_index, Size scan_index_heavy = NULL);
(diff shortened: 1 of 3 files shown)


timosachsenberg:
maybe also fix compile warning:
```
In file included from /nfs/wsi/abi/old-data/sachsenb/OpenMS_IDE/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp:35:0:
/nfs/wsi/abi/old-data/sachsenb/OpenMS_IDE/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h:136:352: warning: converting to non-pointer type ‘OpenMS::Size {aka long unsigned int}’ from NULL [-Wconversion-null]
       static void buildPeptideIDs(std::vector<PeptideIdentification> & peptide_ids, const std::vector< OPXLDataStructs::CrossLinkSpectrumMatch > & top_csms_spectrum, std::vector< std::vector< OPXLDataStructs::CrossLinkSpectrumMatch > > & all_top_csms, Size all_top_csms_current_index, const PeakMap & spectra, Size scan_index, Size scan_index_heavy = NULL);
```

jpfeuffer:
Do you know if the original PR is already in master? If so we have to cherry-pick this one,too.
master builds should start showing up on CDash tomorrow.

timosachsenberg:
cool, will cherry-pick

timosachsenberg:
I think these files are not in master.

enetz:
I can confirm, that these files are not in master.

timosachsenberg:
can you take a look at the failing tests in http://cdash.openms.de/viewTest.php?onlyfailed&buildid=225727
Maybe there are still some bugs hidden in the cross-linking code

enetz:
I had a similar issue with different compilers. The UTIL tests for OpenPepXL read in small examples of real experimental data. It seems there is a peak very close to being matched to a theoretical peak and because of different precisions of the double type of different compiler versions or something similar, sometimes it is matched and sometimes not (the result is always reproducible with the same compiler).
From what I saw clang 3.4 and gcc 4.6 do match that peak (e.g. the travis test builds), but  clang 3.8 and gcc 4.8 do not, so it makes sense that gcc 5.2 also does not match that peak.
Is there a simple way to remove one specific peak from an mzML file?
Although it would take quite some time just to find it, I think.
Or is there a way to make such a comparison more consistent across multiple compilers?

timosachsenberg:
ok its strange because doubles should be the same size...
what you could do is try to apply the window mower filter (e.g. with a small window of 0.001 Da) and write out a new test file. that way you could make sure that no close neighboring peak is in the data

PR #2653, opened 2017-05-30
[FEATURE] allow different models for RT normalization

hroest:
also allow non linear models to be used for the RT normalization
employed in OpenSWATH. This may be useful if many anchor points are
used, such as in the CiRT approach by Parker et al (2015). Using
non-linear iRTs may be more accurate, especially when chromatographic
conditions are not fully linear. However, enough anchor points need to
be available for this to work.

- cherry-picked from 89abb37b637850a57d743d07833e1ab540a646c8


timosachsenberg:
looks good to me. @grosenberger fine with you, too?

grosenberger:
Looks great and will be very useful! 👍

PR #2654, opened 2017-05-30
[FIX] update license

hroest:
seems like we missed a few instances ... 

PS: can somebody explain to me whether changing the date has any legal implication whatsoever? it just seems a nuisance when doing git analysis on single files ...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 #                   OpenMS -- Open-Source Mass Spectrometry
 # --------------------------------------------------------------------------
 # Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
-# ETH Zurich, and Freie Universitaet Berlin 2002-2016.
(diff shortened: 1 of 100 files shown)


hroest:
PS: why do we have LICENSE, License.txt and cmake/knime/COPYRIGHT  ?

jpfeuffer:
haha, word! I talked to @timosachsenberg and I think there is no reason. License.txt is shown in the Windows installer and includes references to the Thirdparty licenses. That is the only difference. Feel free to change this here. It should be possible to replace all occurrences of the different versions by the one in the root folder. Preferably the one with the Thirdparties?

timosachsenberg:
I think we probably should update:
https://github.com/OpenMS/OpenMS/blob/develop/tools/update_header_dates.sh

Regarding the legal implications, I think it might be required in some countries to have a date combined with a copyright statement.

PR #2657, opened 2017-05-31
Feature: full SRM/MRM workflow for proteomics and metabolomics

hroest:
Implement a set of fixes that allow implementation of a full targeted MRM/SRM workflow

- add pyTOPP/MRMTransitionGroupPicker.py tool for peak picking
- add pyTOPP/MRMTransitionGroupScorer.py tool for peak scoring
- background estimation for peak apex and peak area
- report S/N values for each ion trace
- better error reporting when some expected value is missing / something goes wrong


timosachsenberg:
cool. can we use https://github.com/WorkflowConversion/CTDopts as option handler? It will allow us to automatically generate KNIME nodes later.

jpfeuffer:
Great idea I would like that, too.

hroest:
Lets merge this first and then think about CTD - for these tools I am not sure whether its worth it as it would simply lead to a 1:1 translation of the TOPP tool and your abstraction layer in KNIME does not benefit from using pyOpenMS vs the original TOPP tool. That would not really be a step forward but simply more work.

The main benefit of these "example" tools is for people who want to tool in Python and want to use Python as a workflow system instead of KNIME/TOPPAS - so they would have some idea what the tools look like.

timosachsenberg:
Ok I see the reasoning for this particular tools and we can leave them as is. In general, I think providing a CTD interface to pyOpenMS tools should be preferred.

hroest:
yes, but if we "import" in Python it should be a tool that is also in PyPI which I think this one is not?

jpfeuffer:
I checked with the developers:

It's already available in conda in the workflowconversion channel. You can import it like so:
conda install -y -c workflowconversion ctdopts

hroest:
yes, conda is one way to do it but on a regular linux install you would
need PyPI packages as well

Issue #2649, opened 2017-05-30
Error with windows header from CoinOR inclusion?


jpfeuffer:
Unfortunately I think it happened what I saw when testing the CoinOR inclusion branch of @FabianAicheler.
With unity builds we can not include CoinOR because it somehow includes the windows headers.
Not 100% sure that it was this pull request, but maybe 80%.

```
Repository	https://github.com/OpenMS/OpenMS/blob/master/src/openms/source/KERNEL/ConsensusFeature.cpp
Build Log Line	116
Error	
 /.../OpenMS/src/openms/source/KERNEL/ConsensusFeature.cpp(224): error C2059: syntax error : '::' [C:\dev\NIGHTLY\builds\win8-VS10_x64-Debug\src\openms\OpenMS.vcxproj]
```

jpfeuffer:
I am not sure if we can give up on Unity builds. I need to compare the speed at some point. Windows is taking ages anyway and we need every second here if we want to add CI soon.

FabianAicheler:
So from eg https://stackoverflow.com/questions/27442885/syntax-error-with-stdnumeric-limitsmax my understanding is that we could try to #undef max whereever we find a problem or wrap the max call in parentheses.

Would that suffice? How would I test it/what is the difference between the travis builds (which the pull request passed) and the unity build?

jpfeuffer:
Travis is without windows unfortunately. Please add the label Windows-CI to your PR when you try to fix it. You will be the first tester.

jpfeuffer:
I found the problem... The build machine had an old contrib installed. The old CoinOR version defined its own min and max, while the new one actually uses the ones from the stdlib.
Good that we upgraded Coin. I will reopen until I installed new contribs.

jpfeuffer:
Worked.

PR #2664, opened 2017-05-31
Test global definition of NOMINMAX

jpfeuffer:


cbielow:
you can try it, but if the config.h header is included AFTER the problematic windows header, this will do nothing...

jpfeuffer:
how can you include it from the beginning?

cbielow:
new idea:
https://cmake.org/cmake/help/latest/command/add_definitions.html

this should work globally and be independent of include order

jpfeuffer:
win-ci

jpfeuffer:
labels do not work (builds after every comment) since the Jenkins plugin is complete crap

jpfeuffer:
win-ci

jpfeuffer:
win-ci

jpfeuffer:
win-ci

jpfeuffer:
Seems to work but we get a warning because of the macro redefinition in File.h/cpp. Should I remove the definitions in the files? What if the global compiler definition leads to problems someday?

hroest:
it seems we are not the only ones that have this issue, so we are in good company: https://bugzilla.mozilla.org/show_bug.cgi?id=830801

> What if the global compiler definition leads to problems someday?

lets deal with it when happens?

jpfeuffer:
win-ci

jpfeuffer:
Ok seems to work. But there are a lot of failing tests.

jpfeuffer:
Let's merge anyway. Seems to be another problem.

jpfeuffer:
Solves #2649

PR #2667, opened 2017-06-01
Fix/mzid XL-MS mods

enetz:
fixed an error with terminal modifications, that appeared when a cross-link modification would be possible at both the peptide terminal and the side chain of the terminal residue


timosachsenberg:
can you add a test for it?

enetz:
alright, will do

PR #2668, opened 2017-06-01
[FIX] caught an overflow error in AScore

enetz:
the binomial coefficient can overflow the `double` type,
if the scored spectrum has too many peaks and matches.

Fixed it by catching the exception and setting the coefficient to the maximal possible `double`.

### DIFF
--- a/src/openms/source/ANALYSIS/ID/AScore.cpp
+++ b/src/openms/source/ANALYSIS/ID/AScore.cpp
@@ -195,7 +195,20 @@ namespace OpenMS
     // score = sum_{k=n..N}(\choose{N}{k}p^k(1-p)^{N-k})
     for (Size k = n; k <= N; ++k)
     {
-      double coeff = boost::math::binomial_coefficient<double>((unsigned int)N, (unsigned int)k);
+      double coeff = 0;
+
+      try
+      {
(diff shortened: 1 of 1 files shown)


timosachsenberg:
didn't we have similar issue before? would it be possible to calculate the log of it instead?

enetz:
Yes we had discussed this, but I had only solved it on my local branch.
This number is multiplied by other factors < 1 and the results summed up in a loop,
so a few overflows here do not necessarily result in a maximal score, the highest numbers I have seen even with a lot of overflows were much less than 500.

The function, that produces the overflow is the `boost::math::binomial_coefficient`
We could use a log by coding our own `binomial_coefficient_log` function, but that would also turn it into a different score.

cbielow:
not sure I understand it fully, but since pow1 and pow2 are very small you could
`score += exp( bin_coeff_log + pow1_log + pow2_log)`?
On another note: the computation of pow here is somewhat overly expensive, since it could be done iteratively (prone to numeric instability?) or at least as `exp(base, exponent)` (where exp() would nicely cancel out in the log case) see https://stackoverflow.com/questions/17891595/pow-vs-exp-performance

hroest:
in the meantime, should this fix go into `master` ?

timosachsenberg:
I will apply it to master. Using consexpr lookup tables might be the fastest option when we get c11.

PR #2669, opened 2017-06-01
Reduced memory allocated by MSGFPlusAdapter_test JVM

jpfeuffer:
Might fail on 32bit versions of Java or if not enough memory is available.


timosachsenberg:
+1

PR #2670, opened 2017-06-01
[FEATURE] Stream reading and writing of FASTA file entries

cbielow:
drastically reducing memory footprint of DecoyDatabase:

for a 4GB bacterial genome file:
 old: 14 GB RAM
 new: 1 GB (due to book-keeping for duplicate identifiers) - otherwise 11 Megabyte

Time decrease by 33% as well (old: 3 min; new 2 min)

Resulting combined FASTA file is interleaved now (but that should not be a problem).

Code of Decoydatabase gets much nicer this way as well.


timosachsenberg:
with interleaved  you mean:
target1
decoy1
target2
decoy2
?

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -49,21 +47,21 @@
 namespace OpenMS
 {
   using namespace std;
+  typedef seqan::RecordReader<std::fstream, seqan::SinglePass<> > FASTARecordReader;
 
   FASTAFile::FASTAFile()
+    : reader_(std::nullptr_t()), // point to nothing
```
timosachsenberg: c11

timosachsenberg: should we wait with the merge until we have everything in?

cbielow:
yes

hroest:
:+1: I like it

timosachsenberg:
I am undecided if we need to provide the old semantics of appending at the end, too. What do @hendrikweisser and @jpfeuffer think about it?

src/utils/DecoyDatabase.cpp
```diff
@@ -59,6 +60,11 @@ using namespace std;
   This allows you to specify your target database plus a contaminant file and obtain a concatenated
   target-decoy database using a single call, e.g., DecoyDatabase -in human.fasta crap.fasta -out human_TD.fasta
 
+  If a combined database is requested (default), i.e. @em only_decoy is not used), then Target and Decoy sequences
```
hendrikweisser: Don't capitalize "Target" and "Decoy".

src/utils/DecoyDatabase.cpp
```diff
@@ -59,6 +60,11 @@ using namespace std;
...
+  are written interleaved. You need all Targets before the Decoy for some reason, use @em only_decoy and concatenate the files
```
hendrikweisser: Should be: "If you need ..."

src/utils/DecoyDatabase.cpp
hendrikweisser: Remove extra ")" after "used".
I tend to prefer "@p" to highlight parameter names, but just keep it consistent.

src/utils/DecoyDatabase.cpp
hendrikweisser: Again, unusual capitalization of "Targets" and "Decoy" (should be "decoys", plural).

hendrikweisser:
> I am undecided if we need to provide the old semantics of appending at the end, too.

As Chris writes in the documentation, you can always generate a file with only decoys and concatenate that to the targets file.

timosachsenberg:
yes I can ;) I just imagine users that later decide they want to split of decoys in notepad

jpfeuffer:
I know what you mean but we can show them easy workarounds when they ask.

cbielow:
the 'issue' with providing backwards compatibility with non-interleaved combined databases is that it would either:
 - require the old memory footprint (high memory)
 - require reading the input data twice (high I/O and time)
Both do not seem very appealing. If the user really needs to split the combined file, he can either write a script or just call DecoyDatabase with `-only_decoy`

cbielow:
included changes as suggested by Hendrik. Thanks.

timosachsenberg:
looks good. I think we can merge after we enabled c11 support. Right now travis is failing.

timosachsenberg:
c11 support is merged

timosachsenberg:
one thing we could consider for the future is providing forward iterators for stream reading the fasta entries. This is imho the c++ way (maybe not pyOpenMS way ;) of writing generators

hroest:
> This is imho the c++ way (maybe not pyOpenMS way ;) of writing generators

iterators are a pain with pyOpenMS automated builds, at least we should also keep the old interface

PR #2685, opened 2017-06-07
Fix enzymatic digestion

### LINKED ISSUE #2594: PeptideIndexer: "unknown AA" error due to selenocysteine
Since #2545, PeptideIndexer may write error messages like this:
> Error while parsing sequence MKKVV...: found an unknown AA without an estimated mass. Please use PEPTIX[123] syntax to indicate an unknown amino acid with a known mass.

While getting this information now is probably a step forward, there are some problems:
- The consequences of the error are unclear, since the tool finishes normally. Are the affected sequences ignored? Is it just a warning?
- The protein sequence that is written may be very long, in which case it isn't easy to see where the problematic character is.
- The sequence that is written may not occur exactly like that in the FASTA file: If "IL_equivalent" is set, "L" will be replaced by "I"; "invalid" letters (e.g. "U") are replaced by "X". (So if I search for "X" in the FASTA file, I may not find anything.)
- In my case, the problematic letter in the original sequence was a "U", which stands for selenocysteine and is actually correct  for the protein in question (E. coli formate dehydrogenase). The fact that OpenMS (or just PeptideIndexer?) doesn't support this points to a more general problem.

cbielow:
[FIX] honor missed cleavage filter for all site_specificities in `isValidProduct()` 
[FIX] fix bug in missed cleavage filter when peptide starts at 0, but does not cover full protein (`return (pep_positions.size() >= getMissedCleavages() + 1);` did the opposite of what it should)
[FEATURE] allow isValidProduct() to operate also on String (not only AASequence) -- useful for PeptideIndexer to avoid 'unknown AA without an estimated mass...' errors, see #2594 
[FIX] fix warnings in PeptideIndexer when AASequence contains unknown amino acids (fixes #2594)
[TEST] extended tests, also covering newly supported SEMI and NONE specificity cases and borderline cases which were faulty before


timosachsenberg:
@pati-ni just FYI some minor issues have been fixed

Issue #2684, opened 2017-06-07
ADDRESS_SANITIZER does not work on mac


jpfeuffer:
Can't find the Asan library during linking.
https://stackoverflow.com/questions/44320465/whats-the-proper-way-to-enable-addresssanitizer-in-cmake-that-works-in-xcode

hroest:
well you have to turn it on specifically on mac OSX so there is not really
a chance somebody would do that by mistake?

jpfeuffer:
Sorry for being brief yesterday. Well I mean with the current implementation in CMake and Address_sanitizer=On, it does not work. It correctly compiles with the flags but during linking it is missing symbols from the Asan library like _Asan_init_. I think the SO thread linked above defines our options. The thing is, we usually do not use XCode, so not sure if setting XCode flags has an effect.

PR #2687, opened 2017-06-07
[FIX] Add ASan linker flags also to libraries.

jpfeuffer:


jpfeuffer:
Fixes #2684  I think. Wait with merge for my go please. Travis will not catch it.

jpfeuffer:
I spent the whole day looking for generic ways to find the library until I saw the flags are only passed to the executables.. At least I have a backup plan if this Easy solution does not work.

cmake/modules/CodeCoverage.cmake
```diff
@@ -120,53 +122,59 @@ ENDIF() # NOT CMAKE_BUILD_TYPE STREQUAL "Debug"
...
+
+    SET(coverage_info "${CMAKE_BINARY_DIR}/${_outputname}.info")
+    SET(coverage_cleaned "${coverage_info}.cleaned")
+
+    SET(ignorelibpaths \"tests/*\" \"/usr/*\" \"/Applications/*\")
+    foreach(libpath ${_addignorelibpaths})
+        list(APPEND ignorelibpaths \"${libpath}/*\")
+    endforeach()
+
+    ## Workaroud that CMake does not complain during configure, that the log is missing.
```
hroest: typo

jpfeuffer: ah thx

jpfeuffer:
Ok it built through but took more than a day with both enabled at the same time. Coverage was using up all 8GB of RAM. AddressSanitizer is relatively fast.
I googled a bit but nobody else reports such a high RAM usage for Coverage builds with those flags.

jpfeuffer:
We could probably use the flags on less targets and check that no dependencies are tracked but if there are no other ideas, I would leave it that way for now. I will check once more with ONLY coverage.

jpfeuffer:
Ok. Everything works. You have to separate ASan and Coverage, then the runtimes are ok.

PR #2695, opened 2017-06-09
[Fix] Python bindings FFMetabo

jpfeuffer:
Missing new argument


jpfeuffer:
@hroest Could this work? Or do I need to add more specializations for MSChromatogram?

hroest:
this should work, but in order to actually get the results you need a call by reference `&`

PR #2697, opened 2017-06-10
[FIX] fix parallel test issue

hroest:
fix issue with parallel tests

- maybe to master?
- also, why do we not check the output of the TOF calibration? seems it was removed in 152959be40705be6a0b2429d0241661ee5fe6161


timosachsenberg:
I can put it on master,too.
Hmm don't recall why this has been removed. Seems it was quite a while ago...

hroest:
>  Hmm don't recall why this has been removed. Seems it was quite a while ago...

this is why we (should) write commit messages ...

PR #2702, opened 2017-06-14
Do not remove duplicates in dep. lib. string

jpfeuffer:
Also removes duplicate but needed occurrences of debug/optimized annotations
To tell the CMake which library to use in which case.
Also needs to go to develop.


jpfeuffer:
Only timeouts.

PR #2703, opened 2017-06-14
[FIX] OPXL: redo of sorting for the top ranked results

timosachsenberg:


jpfeuffer:
Not fixed. Something seems to happen during writing.
```
Warn: No peptides were matched to the decoy portion of the database! Did you provide the correct concatenated database? Are your &apos;decoy_string&apos; (=decoy) and &apos;decoy_string_position&apos; (=suffix) settings correct?
Set &apos;missing_decoy_action&apos; to &apos;error&apos; if you want to elevate this to an error!
Progress of &apos;Writing output...&apos;:
Writing xquest.xml to OpenPepXL_output.xquest.xml.tmp
Writing spec.xml to OpenPepXL_output.spec.xml.tmp
=================================================================
==42350==ERROR: AddressSanitizer: container-overflow on address 0x6170000b3fc0 at pc 0x00010debc794 bp 0x7fff53f93850 sp 0x7fff53f93848
READ of size 8 at 0x6170000b3fc0 thread T0
```

jpfeuffer:
https://abibuilder.informatik.uni-tuebingen.de/jenkins/view/OpenMS/job/openms_nightly_supplement/compiler=appleclang-7.3.0,os_label=elcapitan/ws/build/General_Testing/Testing/20170614-1835/Test.xml/*view*/

jpfeuffer:
And maybe there are not enough checks for empty containers?
```
AASequence::getMonoWeight: Mass for ResidueType 0 not defined for sequences of length 0.
=================================================================
==42351==ERROR: AddressSanitizer: container-overflow on address 0x6140000e1ee8 at pc 0x00010111fd41 bp 0x7fff5eaf4c30 sp 0x7fff5eaf4c28
READ of size 8 at 0x6140000e1ee8 thread T0
```

jpfeuffer:
But I am not sure if the output of ASan is not interspersed due to parallelism. And also: The failing tests all seem to reach an execution time of 600s. Is that a hard limit. Do the tests just get stuck? But why?

jpfeuffer:
How long do the tests take in optimized release mode? Are they among the longest ones?

timosachsenberg:
@jpfeuffer @cbielow @hroest what is your opinion on keeping the min/max check in codacy?

jpfeuffer:
Would have saved a lot of time in the CoinOR inclusion..
We have a global NOMINMAX definition now. None of our dependencies defines min/max macros anymore.

jpfeuffer:
But that it really makes sense, we have to change all of them in the codebase first.

timosachsenberg:
hmm yeah the question is if we should do this after we have NOMINMAX defined.
std::(max)(a,b) looks pretty ugly.
I think I vote for removing the check

enetz:
I did the same fix for OpenPepXLLF as I did earlier for OpenPepXL, since it seems to have helped.
I also tried to make the output writing more robust, but the new error message is much less informative about the faulty position in the code.

The duration of these tests is on the longer side, but the UTILS_OpenPepXL_1 takes about half a second, while there are several others that take more than a whole second.

Also there seems to be a similarly minimal error message for the AScore_test and I also use the AScore in OpenPepXL. Could it be the same error?

enetz:
Actually no, I do not use the AScore or PScore yet in this version of OpenPepXL, so they must be unrelated issues.

enetz:
Alright, using valgrind I could pinpoint the problem (or "a" problem, not sure it is the same issue ASan complained about, but should be) and fix it. However with my current fix one of the output files has changed and I have to check, whether all the necessary info is still in there. It might just be less redundant. Will push the changes once I have verified or fixed that.

jpfeuffer:
Great. I'd be glad to test your fix.

enetz:
Ok, the last commit should take care of the OpenPepXL errors. @jpfeuffer Please test it.
I will try to find out what is wrong with the AScore now.

enetz:
@timosachsenberg  @jpfeuffer For the AScore this is what valgrind found, two separate errors, but both ultimately caused by an uninitialized value created by `DTAFile::load` and then accessed in `AScore::peakPickingPerWindowsInSpectrum_ `:
1.
```
==11331== Conditional jump or move depends on uninitialised value(s)
==11331==    at 0x5FD0546: OpenMS::AScore::peakPickingPerWindowsInSpectrum_(OpenMS::MSSpectrum<OpenMS::Peak1D>&) const (AScore.cpp:531)
==11331==    by 0x5FCD2E2: OpenMS::AScore::compute(OpenMS::PeptideHit const&, OpenMS::MSSpectrum<OpenMS::Peak1D>&, double, bool, unsigned long, unsigned long) (AScore.cpp:110)
==11331==    by 0x4311D1: main (AScore_test.cpp:836)
==11331==  Uninitialised value was created by a heap allocation
==11331==    at 0x4C2B0E0: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11331==    by 0x442F53: __gnu_cxx::new_allocator<OpenMS::Peak1D>::allocate(unsigned long, void const*) (new_allocator.h:92)
==11331==    by 0x4414A8: std::_Vector_base<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> >::_M_allocate(unsigned long) (in /home/eugen/Development/OpenMS-gcc/src/tests/class_tests/bin/AScore_test)
==11331==    by 0x43DCA4: std::vector<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> >::_M_insert_aux(__gnu_cxx::__normal_iterator<OpenMS::Peak1D*, std::vector<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> > >, OpenMS::Peak1D const&) (vector.tcc:327)
==11331==    by 0x43AD95: std::vector<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> >::push_back(OpenMS::Peak1D const&) (stl_vector.h:834)
==11331==    by 0x4371E6: void OpenMS::DTAFile::load<OpenMS::MSSpectrum<OpenMS::Peak1D> >(OpenMS::String const&, OpenMS::MSSpectrum<OpenMS::Peak1D>&) (DTAFile.h:177)
==11331==    by 0x43025D: main (AScore_test.cpp:785)
```
2.
```
==11331== Conditional jump or move depends on uninitialised value(s)
==11331==    at 0x5FD0546: OpenMS::AScore::peakPickingPerWindowsInSpectrum_(OpenMS::MSSpectrum<OpenMS::Peak1D>&) const (AScore.cpp:531)
==11331==    by 0x435BAA: AScoreTest::peakPickingPerWindowsInSpectrumTest_(OpenMS::MSSpectrum<OpenMS::Peak1D>&) const (AScore_test.cpp:95)
==11331==    by 0x42F539: main (AScore_test.cpp:729)
==11331==  Uninitialised value was created by a heap allocation
==11331==    at 0x4C2B0E0: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11331==    by 0x442F53: __gnu_cxx::new_allocator<OpenMS::Peak1D>::allocate(unsigned long, void const*) (new_allocator.h:92)
==11331==    by 0x4414A8: std::_Vector_base<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> >::_M_allocate(unsigned long) (in /home/eugen/Development/OpenMS-gcc/src/tests/class_tests/bin/AScore_test)
==11331==    by 0x43DCA4: std::vector<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> >::_M_insert_aux(__gnu_cxx::__normal_iterator<OpenMS::Peak1D*, std::vector<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> > >, OpenMS::Peak1D const&) (vector.tcc:327)
==11331==    by 0x43AD95: std::vector<OpenMS::Peak1D, std::allocator<OpenMS::Peak1D> >::push_back(OpenMS::Peak1D const&) (stl_vector.h:834)
==11331==    by 0x4371E6: void OpenMS::DTAFile::load<OpenMS::MSSpectrum<OpenMS::Peak1D> >(OpenMS::String const&, OpenMS::MSSpectrum<OpenMS::Peak1D>&) (DTAFile.h:177)
==11331==    by 0x425A6A: main (AScore_test.cpp:115)
```

I do not know enough about DTAFile or the AScore to fix this quickly, but I will look into it until @timosachsenberg is back. Maybe it is just an issue with the test files, that DTAFile reads in?

enetz:
Ok, I think I found the issue. The error was a wrong order of conditions in a while-loop, where a value was read from an iterator in the first condition, and then only the second condition tested whether that iterator has reached the end of the vector. I switched these two and valgrind at least does not complain about this type of error anymore.
Address Sanitizer does not find any critical errors either, but then again I could not reproduce those on my local build before either.
@jpfeuffer please test the mac ASan build with these latest changes, there should not be any container-overflows left.

jpfeuffer:
Good find. Makes sense!

jpfeuffer:
yeah great. That solved those two ASan errors.

jpfeuffer:
I restarted Travis.. it hung up

jpfeuffer:
puh... develop? Can both changes go to master? I know the second AScore fix can. Not sure about the first part(s).

enetz:
Right, the AScore fix should also apply to master, but OpenPepXL is not in master yet.

PR #2704, opened 2017-06-14
Add a Codacy badge to README.md

codacy-badger:


hroest:
should it be the first badge? I would have put it after the licence

timosachsenberg:
no real preference but can move it there

Issue #2630, opened 2017-05-24
IsobaricAnalyzer TMT10plex IsobaricIsotopeCorrector fails


oliveralka:
It seems that the values of the matrix inversion and the NNLS differ - see ERROR.

```
Selecting scans with activation mode: High-energy collision-induced dissociation
Filtering by MS/MS(/MS) and activation mode:
  level 2: 7577 scans
Using MS-level 2 for quantification.
Calibration stats: Median distance of observed reporter ions m/z to expected position (up to 0.5 Th):
  ch 126  (~126.127): -0.000135023 Th
  ch 127N (~127.124): -0.000132188 Th
  ch 127C (~127.131): -8.35508e-05 Th
  ch 128N (~128.128): -0.000118863 Th
  ch 128C (~128.134): -8.54844e-05 Th
  ch 129N (~129.131): -0.000120797 Th
  ch 129C (~129.137): -5.79004e-05 Th
  ch 130N (~130.134): -0.000108472 Th
  ch 130C (~130.141): -2.93164e-05 Th
  ch 131  (~131.138): -6.46289e-05 Th

Error: Unexpected internal error (IsobaricIsotopeCorrector: Isotope correction values of alternative method differ!)
```

I tried to solve the issue last time by altering the absolute to relative deviation between the matrix conversion and NNLS:

```
IsobaricIsotopeCorrector.cpp
line 168: else if ((((std::fabs(m_x(index, 0) - x(index)))/m_x(index, 0))*100) > 1)
```

After the update to MSn  I'm not able to run my dataset anymore - Any ideas why?

cbielow:
at leas one of these seems numerically unstable... 
the test which throws the exception is not of utmost importance. We could simply remove it.
But it would be interesting to see what the numbers are. You are probably seeing a range of very high and very low intensities in one spectrum.

oliveralka:
Added:

```
      {
        ++s_different_count;
        s_different_intensity += std::fabs(m_x(index, 0) - x(index));
        std::cout << "matrix: " << x(index) << " NNLS: " << m_x(index, 0) << std::endl;
      }
```

Example: 

```
Calibration stats: Median distance of observed reporter ions m/z to expected position (up to 0.5 Th):
  ch 126  (~126.127): -0.000135023 Th
  ch 127N (~127.124): -0.000132188 Th
  ch 127C (~127.131): -8.35508e-05 Th
  ch 128N (~128.128): -0.000118863 Th
[...]

PR #2707, opened 2017-06-22
[FIX] Changed NNLS vs. inverse Matrix comparison to warning instead of exception 

oliveralka:
Changed NNLS vs. inverse Matrix comparison to warning instead of exception.

see #2630.


hroest:
this looks good but for the user it is still unclear what s/he is supposed to do about it...

@cbielow  can we also come to a consensus regarding that

cbielow:
First, it would be best to check which of these two methods is actually better (i.e. more stable), and use that to compute the result (the data only gives the values which are different, so not the whole input array) 
@oliveralka Can you provide the input values for the spectrum which fails and the correction matrix used?

oliveralka:
@cbielow What do you mean by input values? An example workflow with data attached?

cbielow:
a vector of 10 reporter ions, which yields different results :)

hroest:
maybe we could simulate a few cases with different noise levels and then
see which method is closest to the truth?

also, which method do other people use?

timosachsenberg:
I will merge it because it makes the analysis possible but I agree that we should dig into this a bit more.

PR #2708, opened 2017-06-23
[FIX] Removed comments not compatible with new autowrap.

jpfeuffer:


jpfeuffer:
Why are there conflicts?

jpfeuffer:
Let me rebase

PR #2712, opened 2017-06-26
[FIX] check file extensions of main file formats

timosachsenberg:


cbielow:
would it make sense to write a function for this kind of repetitive code?!
And maybe add "While storing file 'XZ': invalid filename....", so the user knows at which point it happened.

While this check is nice and useful, we do check this on TOPP tool level already somewhat, right? (if file format restrictions are given), e.g. Any TOPP tool will output
`Invalid parameter: Invalid output file extension 'bla.idxml'. Valid file extensions are: 'txt'.`

hroest:
well it may help if the developer forget the check :-)

timosachsenberg:
I can move it in a separate function

hroest:
:+1: looks good

PR #2715, opened 2017-06-27
[FIX] fix failing XML schema validation on MacOSX 

cbielow:
when XercesC does not have URL support for online schemata. XSD files are now local in /share/SCHEMATA.

Additional benefit: mzXML_test now takes 0.3 seconds, instead of 2.4 seconds, since the online lookup apparently takes a lot of time.


timosachsenberg:
cool. Thanks! Do you know if the online lookup also affecting e.g. the mzML test speed?

cbielow:
mzML does not have dependend schemata which are linked online, so should be no problem

cbielow:
I did not touch any code, thats why :)

PR #2717, opened 2017-06-27
Feature/chrom dataprocessing

hroest:
fixes an issue that for chromatogram files it would always write out a dummy data processing since it was testing for the presence of spectra instead of the presence of data processing vectors


src/utils/OpenSwathWorkflow.cpp
```diff
@@ -808,17 +811,20 @@ class TOPPOpenSwathWorkflow
...
-
+      progresslogger.endProgress();
     }
     else
     {
       LOG_ERROR << "Provide valid TraML, TSV or PQP transition file." << std::endl;
       return PARSE_ERROR;
     }
-    progresslogger.endProgress();
+    std::cout << "Loaded " << transition_exp.getProteins().size() << " proteins, " <<
```
timosachsenberg: LOG_INFO <<

hroest:
merge?

PR #2721, opened 2017-06-28
[PYOPENMS] improve pyTOPP tool

hroest:
minor fix and improvement of pyTOPP tool


timosachsenberg:
would be nice to have some tests for these tools at some point

Issue #1985, opened 2016-05-12
Decharger tool, used solver lib


FabianAicheler:
Hi, as far as I can see, the solver used for Decharging uses glpk. Does anyone have experience how much effort it would be to switch to coinor/coin-mp? Reasons:
- A collegue told me that glpk was slow compared to the coinor cbc :D
- For Windows, I want to use the Decharger tool for a community node in Thermo Fisher Compound Discoverer. It would be nice to be able to say that all the stuff people install is under BSD 3-clause

Concerning the current status I found https://github.com/OpenMS/contrib/issues/9
Is that still the current state or were there further discussions I am not aware of? As far as I can see, we currently do not ship coinor/coin-mp with Windows OpenMS, is that correct?

Thanks

cbielow:
1) your colleage is right, CoinMP is (a lot) faster than GLPK and we should kick GLPK -- if you are at it anyways, any volunteers? :)

2) (citing myself, dont quote me on it though) COIN-MP is EPL [1], which is compatible with BSD [2]

3) We have a an LPWrapper[3], which makes switching between GLPK and COINOR a nobrainer
It should be possible to use COINOR by just switching it on in the build system, i.e. -DUSE_COINOR=ON, see src/openms/cmake_findExternalLibs.cmake:if (${USE_COINOR})
Shipping for the Windows installer should not be required, since the libs are statically linked (there's libClp.lib, libOsi.lib, etc... in the contrib/lib directory)

[1] https://projects.coin-or.org/CoinMP/browser/releases/1.8.0/CoinMP/LICENSE
[2] https://eclipse.org/legal/eplfaq.php#COMPILEWOMOD
[3] http://ftp.mi.fu-berlin.de/pub/OpenMS/develop-documentation/html/classOpenMS_1_1LPWrapper.html

timosachsenberg:
When I use CoinOR I get an error in LPWrapper_test.
Anyone can confirm this?
```
50: checking (void readProblem(String filename, String format)) ...
50:  +  line 311:  TEST_EXCEPTION(Exception::NotImplemented,lp.readProblem("/bla/bluff/blblb/sdfhsdjf/test.txt","LP")): OK
50: : passed
50:
50: checking (void writeProblem(const String &filename, const WriteFormat format) const ) ...
[...]

PR #2723, opened 2017-06-28
[TRAVIS] get coin through package management

hroest:
- should help with timeouts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,14 @@ addons:
          - subversion
          - doxygen 
          - graphviz
+         - coinor-libcbc0
+         - coinor-libcgl0
+         - coinor-libcoinutils-dev
+         - coinor-libcgl-dev
+         - coinor-libcbc-dev
+         - coinor-libclp-dev
+         - coinor-libosi-dev
+         - coinor-libvol-dev
          - libpcre3-dev
          
 before_script:
--- a/tools/travis/lnx-cibuild.before.sh
+++ b/tools/travis/lnx-cibuild.before.sh
@@ -31,9 +31,6 @@ build_contrib EIGEN
 # we build Sqlite as the versions shipped in Ubuntu are not recent enough
 build_contrib SQLITE
 
-# necessary for succesfull build?
-build_contrib COINOR
-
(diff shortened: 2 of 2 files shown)


timosachsenberg:
we now get the
The following tests FAILED:
	 50 - LPWrapper_test (OTHER_FAULT)

timosachsenberg:
I am not sure but I guess the word "failed" might trigger it.
```
checking (Int addRow(std::vector< Int > &row_indices, std::vector< double > &row_values, const String &name, double lower_bound, double upper_bound, Type type)) ... 
LPWrapper_test: OsiClpSolverInterface.cpp:3606: virtual int OsiClpSolverInterface::loadFromCoinModel(CoinModel&, bool): Assertion `integerType' failed.
 +  line 105:  TEST_EQUAL(lp.getNumberOfRows(),3): got 3, expected 3
 +  line 106:  TEST_EQUAL(lp.getRowName(2),"row3"): got row3, expected row3
: passed
```

timosachsenberg:
@FabianAicheler do you know how to handle this assertion?

jpfeuffer:
Maybe you have to go to the CTestConfig and whitelist that one.

jpfeuffer:
https://cmake.org/Wiki/CMake/Testing_With_CTest
see customizing CTest

hroest:
what does that even mean: LPWrapper_test: OsiClpSolverInterface.cpp:3606: virtual int OsiClpSolverInterface::loadFromCoinModel(CoinModel&, bool): Assertion `integerType' failed. ?

It seems like we had that issue before:

- https://github.com/OpenMS/OpenMS/issues/1985
- also other ppl https://groups.google.com/forum/#!msg/cyclus-dev/mcZBvyqa4xw/xwxN2qHpGBYJ

hroest:
in the last few PR, most build failures were due to CoinOR timing out -- we could add a `travis_wait` or we can just install it through apt-get which is the sensible option anyways.

timosachsenberg:
ok so should we merge this PR?

timosachsenberg:
So now we need to fix the LPWrapper_test

PR #2724, opened 2017-06-28
[FIX] include what you use applied to openms core

timosachsenberg:
just a test - works on linux
created with clion 2017.2 EAP


cbielow:
did I misread it or does this actually increase compile time on travis? bad luck?

timosachsenberg:
hmm strange. Ok at least it seems to work to a better degree now than one year ago :P I only had to revert 3 or 4 files to get it to compile on linux

hroest:
I like it, what went wrong?

timosachsenberg:
I am currently not really satisfied with the includes it keeps. And as it did not make a difference in build time I revert it for now. Probably get back to that in a while.

PR #2719, opened 2017-06-27
merge develop into qt5 branch

hroest:
merge develop into qt5 branch

@cbielow and @timosachsenberg can you please look at 797cdfd and 22e3f3d - I think here something got screwed up in the last merge and I tried to fix it but I was not too familiar with the code

Issue #2241, opened 2016-10-17
Upgrade to QT5 and replace QTWebKit (or patch)


jpfeuffer:
As mentioned by @lars20070 and also experienced by me:
QTWebkit does not build with gcc5+ anymore because it optimizes out some code. There is a fix for QT5 ([here](https://codereview.qt-project.org/#/c/107921/)) but QT4 does not seem to be updated anymore. Since we do not include QT in our contrib we can not patch it easily ourselves but would need to give the users detailed instructions on how to patch and build this monster on their own.
This would be a good opportunity to upgrade to QT5 (since also the QTWebkit was deprecated in favor of the QTWebEngine some time ago).

jpfeuffer:
Btw: gcc5 is the standard compiler on LTS Ubuntu 16.04.2 (thx @lars20070 )

timosachsenberg:
I would suggest we create a qt5 branch both for contrib and from develop. Maybe we can distribute the workload (and use githubs project tracking system to see who is working on what)

timosachsenberg:
I created one from the current develop: https://github.com/OpenMS/OpenMS/tree/feature/qt5

timosachsenberg:
here is a nice list of qt5 modules that we might need to include in cmake: http://doc.qt.io/qt-5/qtmodules.html

timosachsenberg:
cmake runs through with some warnings / resources not included:
as Qt5 is now cmake based its sufficient to specify:
```
cmake -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.6.0/ -DQT_QMAKE_EXECUTABLE='/usr/local/Cellar/qt5/5.6.0/bin/qmake' .
```

timosachsenberg:
here is a great reference for integrating Qt5. I think we can copy a lot of the code from https://github.com/BALL-Project/ball/blob/master/CMakeLists.txt

jpfeuffer:
For migration concerning CMake: https://www.kdab.com/using-cmake-with-qt-5/

jpfeuffer:
We should take the chance and upgrade minimum required CMake to at least 2.8.12. It is available in the major Linux distro package managers since the second last stable release (e.g. CentOS6, Ubuntu 14.04)

timosachsenberg:
[...]

PR #2728, opened 2017-06-28
[BUILD] fix travis build for Qt5

hroest:
this builds on #2719 and is a fix for issue #2241

this should fix the warnings we seen when configuring the debug build with clang for Qt5


hroest:
success! the main things failing are these:

```
The following tests FAILED:

	 50 - LPWrapper_test (OTHER_FAULT)

	574 - TOPPView_test (Not Run)

	1931 - UTILS_ImageCreator_1_out1 (Failed)

	1933 - UTILS_ImageCreator_2_out1 (Failed)
```

where the LP wrapper fail is due to the fact that I incorporated #2723 already here to get rid of timeouts. The ImageCreator is probably an issue with the migration as well as the TOPPView test

if we merge this I think we are on track for a travis-qt5 setup which puts us on track for a merge of the qt5 branch!

hroest:
also the images from the ImageCreator are almost identical, there is a single line differnce

Image1:
```
vbindiff /home/hr/openmsall/builds/openms_qt5/src/tests/topp/ImageCreator1.bmp /home/hr/openmsall/source/OpenMS_qt5/src/tests/topp/ImageCreator_1_output.bmp
0000 0020: 00 00 84 03 00 00 61 0F  00 00 61 0F 00 00 00 00  ......a. ..a.....                                                                                            
vs
0000 0020: 00 00 84 03 00 00 89 0B  00 00 89 0B 00 00 00 00  ........ ........                                                                                            
```

Image2:
```
vbindiff /home/hr/openmsall/builds/openms_qt5/src/tests/topp/ImageCreator2.png /home/hr/openmsall/source/OpenMS_qt5/src/tests/topp/ImageCreator_2_output.bmp

0000 0020: 00 00 84 03 00 00 61 0F  00 00 61 0F 00 00 00 00  ......a. ..a.....                                                                                            
vs
0000 0020: 00 00 84 03 00 00 89 0B  00 00 89 0B 00 00 00 00  ........ ........                                                                                            
```


from the header I gather that Qt writes a 14 byte bitmap file header and then a 40 byte DIB header which would make the header stop at position 36 (hex). If we look at Image2 in greater detail we can see that the changed position is within the header at position 27-28 and 2A-2B


```
0000 0000: 42 4D BA 03 00 00 00 00  00 00 36 00 00 00 28 00  BM...... ..6...(.                                                                                            0000 0010: 00 00 14 00 00 00 0F 00  00 00 01 00 18 00 00 00  ........ ........                                                                                            
0000 0020: 00 00 84 03 00 00 61 0F  00 00 61 0F 00 00 00 00  ......a. ..a.....    

vs

0000 0000: 42 4D BA 03 00 00 00 00  00 00 36 00 00 00 28 00  BM...... ..6...(.                                                                                            
0000 0010: 00 00 14 00 00 00 0F 00  00 00 01 00 18 00 00 00  ........ ........                                                                                            0000 0020: 00 00 84 03 00 00 89 0B  00 00 89 0B 00 00 00 00  ........ ........     
```


looking at the Wikipedia table [here](https://en.wikipedia.org/wiki/BMP_file_format#DIB_header_.28bitmap_information_header.29), the change is in the following values:

- [offset 26] the horizontal resolution of the image. (pixel per meter, signed integer)
- [offset 2A] the vertical resolution of the image. (pixel per meter, signed integer)

which does not really make sense in our setting, e.g. we only care about absolute pixels and dont have a resolution set. So overall, it seems like that is a change within Qt that does not affect the image at all and we can just use the new image as the expected output.

hroest:
PS: also its interesting that the second test creates a bmp image called "ImageCreator2.png"

hroest:
so after this, once we fix `TOPPView_test` I think we would be ready from the Linux side. Looking forward to reports on other platforms.

hroest:
I could not really figure out the issue with TOPPView_test yet and I dont really understand whats going on, maybe @cbielow or @timosachsenberg could have a look at it. I commented out the line 
```
246 // #include <source/GUI/moc_TOPPView_test.cpp> // for Qt's introspection
```
but then I get strange errors like 

```
100%] Linking CXX executable ../bin/TOPPView_test
CMakeFiles/TOPPView_test.dir/source/GUI/TOPPView_test.cpp.o: In function `OpenMS::TestTOPPView::~TestTOPPView()':
TOPPView_test.cpp:(.text._ZN6OpenMS12TestTOPPViewD2Ev[_ZN6OpenMS12TestTOPPViewD5Ev]+0x10): undefined reference to `vtable for OpenMS::TestTOPPView'
CMakeFiles/TOPPView_test.dir/source/GUI/TOPPView_test.cpp.o: In function `OpenMS::TestTOPPView::~TestTOPPView()':
TOPPView_test.cpp:(.text._ZN6OpenMS12TestTOPPViewD0Ev[_ZN6OpenMS12TestTOPPViewD5Ev]+0x10): undefined reference to `vtable for OpenMS::TestTOPPView'
CMakeFiles/TOPPView_test.dir/source/GUI/TOPPView_test.cpp.o: In function `main':
TOPPView_test.cpp:(.text.startup+0x1ab): undefined reference to `vtable for OpenMS::TestTOPPView'
collect2: error: ld returned 1 exit status
```

when searching the internet, people recommend to [move the class definition to a header](https://lizardo.wordpress.com/2009/04/24/undefined-reference-to-vtable-for-errors-in-qt-derived-classes/) -- maybe somebody has the time to do that and properly integrate it with MOC ?

PR #2729, opened 2017-06-29
[FIX] attempt to fix LPWrapper_test

timosachsenberg:


timosachsenberg:
@FabianAicheler now I don't get the assertion but several failed tests. Can you check if these changes make sense?

FabianAicheler:
@timosachsenberg I will look into it.

FabianAicheler:
I will have to read up on it further, but my understanding is, that the BOUNDS name should be left unchanged as it was previously:

Afaik, the names serve to group the bounds into vectors, and normally only the first bound vector (eg all bounds labeled "BOUNDX") are considered.

Now to the failure (readProblem     LPWrapper_test.mps)
The error we get says that the second problem column was expected to be integer but turned out continuous (the first column tested before turned out fine) -> because the different name disabled the second bound ("BV" sets to binary (but for coinor LPWrapper only knows continuous and integer problems, so we call it integer, dont ask me why))

My recommendation would be to at least change the names back in LPWrapper_test. I will have to think about it more for LPWrapper_test_integer

FabianAicheler:
Ok, yes from my understanding, for both problem test files, the bounds in a file should have the same name to be considered together. Btw, the problem is

max y 
for
y -x <= 0
2y + 3x <= 12
3y + 2x <= 12
constraint (for integer)
x,y >= 0
x,y <= 5

If we remove the bounds in the integer problem altogether, the problem then SHOULD automatically have the constraints
x,y >= 0
I think the integer test should pass then. But not sure if we can call that fixing the problem or just masking it.

FabianAicheler:
Concerning: LPWrapper_test: OsiClpSolverInterface.cpp:3606: virtual int OsiClpSolverInterface::loadFromCoinModel(CoinModel&, bool): Assertion `integerType' failed. ?

From my understanding, void LPWrapper::setColumnType correctly tells the CoinModel that the column of interest (integer or bool, as coinor does no real bools) is an integer column.

But https://github.com/coin-or/Clp/blob/master/Clp/src/OsiClp/OsiClpSolverInterface.cpp (here line 4686 contains the offending assert, but should match) somehow the assertion fails when reading in a model BUT after reading (and loading the coinmodel) we successfully check that the columns have integer type.

I'm a bit confused, I guess we tell CoinModel that our columns are integer, but not the whole matrix? My recommendation would be to mask the failing assertion, our tests work and their solutions are the correct _integer_ solutions.

timosachsenberg:
hmm I don't know if it is possible to mask the failing assertion. If there is an assertion the test breaks with a non-zero return value.

FabianAicheler:
Hmm, can you also revert the bounds for the integer problem to the original? And then I would like to re-check where the assertion pops up, because it did so at different spots for https://github.com/OpenMS/OpenMS/issues/1985 and https://github.com/OpenMS/OpenMS/pull/2723

timosachsenberg:
ok lets wait and see

FabianAicheler:
So for travis-ci-OpenMS_OpenMS-develop-2729-g++-testClass-testTOPP (travis-ci-build-server) 
http://cdash.openms.de/testDetails.php?test=41416393&build=228910 it failed at 105 (second addRow test)

For travis-ci-OpenMS_OpenMS-develop-2729-clang++-testClass-testTOPP-noGUI (travis-ci-build-server)  http://cdash.openms.de/testDetails.php?test=41414963&build=228908 near 394 (solve test after the LPWrapper_test_integer.mps problem?)

and for  travis-ci-OpenMS_OpenMS-develop-2729-clang++-testClass-testTOPP (travis-ci-build-server) http://cdash.openms.de/testDetails.php?test=41415638&build=228909 again at 105 addRow test.

Always the same message "LPWrapper_test: OsiClpSolverInterface.cpp:3606: virtual int OsiClpSolverInterface::loadFromCoinModel(CoinModel&, bool): Assertion `integerType' failed."

Which sucks, as it's not even deterministic/reproducible (e.g., no problems when I build on abinode or windows).

FabianAicheler:
Maybe we could patch #define NDEBUG into the OsiClpSolverInterface, that disables asserts, yes? Or remove the assert itself?

hroest:
> Maybe we could patch #define NDEBUG into the OsiClpSolverInterface, that disables asserts, yes? Or remove the assert itself?

we install coinor through the package management, so we cant really do that

hroest:
> we install coinor through the package management, so we cant really do that

maybe we could check whether there is a ppa repo for coinor that contains a newer version and that may fix the issue?

timosachsenberg:
Ok seems to work with new coinor: https://github.com/OpenMS/OpenMS/pull/2731

PR #2731, opened 2017-06-30
check for trusty compatibility

timosachsenberg:


timosachsenberg:
@jpfeuffer fixes lp wrapper test on travis

hroest:
I also have a branch for qt5 that (mostly) builds: feature/qt5-travis-trusty

timosachsenberg:
For your configuration we loose the container build infrastructure and use the fixed instances (because it requires sudo).
Currently, there seems to be no real big benefit from the container infrastructure but this might change. For example, the fixed infrastructure has limited resources that travis afaik prioritizes to paying members. The container infrastructure is on amazon and currently not limited for open source projects.

hroest:
I think using the same trick as in precise should also work in trusty, using `- sourceline: ` for a ppa package

PR #2732, opened 2017-06-30
add binutils for gold linker

timosachsenberg:


hroest:
any benefits? seems like this also fixes the LP_Wrapper issues? #2729

timosachsenberg:
the fix for LP_wrapper is already included in this build so this is most likely not related. Gold linker seems to be not really faster for OpenMS but maybe I did something wrong. Was just a quick experiment anyways.

Issue #2693, opened 2017-06-08
Experience building OpenMS with Visual Studio 2017


hendrikweisser:
Somewhat naively, I installed the newest version of Microsoft Visual Studio, VS 2017, on my new Windows 10 laptop and tried to build OpenMS (64 bit) with it. I succeeded in the end, but some tinkering was necessary, since OpenMS only supports VS up to 2015 out of the box. I thought it might be useful to share my experiences here.

1. Building Qt 4.8.7:
Follow the process described in the OpenMS documentation (http://ftp.mi.fu-berlin.de/pub/OpenMS/develop-documentation/html/install_win.html) under "QT installation". Importantly, apply the patch from the "contrib" package and set `-platform win32-msvc2015` in the "configure" call. (Qt 4.8.7 doesn't know about MSVC 2017, but using the 2015 platform seems to have worked.)
Once installed, Qt cannot be moved (without breaking path references in "qmake"), so make sure to install it in the desired place using `-prefix` with an absolute (!) path in the "configure" call.

2. Building the "contrib":
My CMake version (3.8.2) has support for the "Visual Studio 15 2017 Win64" generator, so I used that. (I haven't tried the "Visual Studio 14 2015 Win64" generator; maybe that would have been easier.) I had to adapt the "CMakeLists.txt" in the "contrib" to make the build system aware of that version. Around line 275, I added:
```
  elseif (CMAKE_GENERATOR MATCHES ".*Visual Studio 15.*")
    set(TMP_MSVC_VERSION "15")
```
This already made it work for some "contrib" libraries, but not all:
- Boost: The version currently shipped with OpenMS (1.61) doesn't support VS 2017. I replaced it with version 1.64 (which does). I manually downloaded the "tar.gz" file from the Boost website and put it in my "contrib-build\archives" folder. I also replaced the occurrences of "boost_1_61_0" with "boost_1_64_0" in the "CMakeLists.txt" and adapted the SHA1 hash (run CMake once, it will complain about the SHA1 mismatch, giving you the expected one that you can copy over).
[...]

PR #2734, opened 2017-07-03
enable c11 support

timosachsenberg:


jpfeuffer:
Anyone knows why MSGF+ would fail? Was there a change in CVs?
```
in1:  /home/travis/build/OpenMS/OpenMS/_build/src/tests/topp/MSGFPlusAdapter_1_out2.tmp.mzid   (line: 106, position/column: 61/62)
        <cvParam cvRef="PSI-MS" accession="MS:1000777" name="!
                                                             spectrum identifier nativeID format"/>

in2:  /home/travis/build/OpenMS/OpenMS/src/tests/topp/THIRDPARTY/MSGFPlusAdapter_1_out.mzid   (line: 106, position/column: 61/62)
        <cvParam cvRef="PSI-MS" accession="MS:1000768" name="!
                                                             Thermo nativeID format"/>
```

jpfeuffer:
Ah sorry wrong PR.

jpfeuffer:
Win CI will start once there is a slot ready.

timosachsenberg:
@jpfeuffer nearly passes on all platforms but fails with http://cdash.openms.de/viewBuildError.php?buildid=229450

timosachsenberg:
can we disable the -Wc++98-compat warning?

jpfeuffer:
Is it built with warning=error or why does it trigger an error?
But yes we could. In theory we do not want to write c98 compatible code anyway anymore right?

timosachsenberg:
I mean its not bad to write c98 code but it should definitely not trigger an error if c98 incompatible code is written ;)

jpfeuffer:
Yes sure. IMO you can add it to compile definitions cmake to the clang compiler flags.
Do not forget to add the pedantic versions.

timosachsenberg:
ok looks good. so let's see if windows ci chokes on this

timosachsenberg:
windows seems to pass

jpfeuffer:
Yep.. well I have to adapt the criteria for success a bit. But it builds and the same tests fail as without the flags :) And I need to link to CDash or increase the Jenkins backlog.

enetz:
@timosachsenberg 
I am trying to compile OpenMS on a windows machine, but I am getting thousands of instances of the same error:
`'noexcept' : unknown override specifier`
with VS2013 x64, which does not support `noexcept`.
So do we no longer support VS2013? Or would it be possible to make workaround (if `noexcept` is the only issue) like the ones discussed here:
https://stackoverflow.com/questions/18387640/how-to-deal-with-noexcept-in-visual-studio

timosachsenberg:
Sorry but I think we probably need to drop support for 2013. Introducing a workaround would only work for noexcept but we already have two lambdas in there.

enetz:
Alright, no problem, just wanted to make sure I actually have to upgrade to VSE2015 (which should work, right?)

jpfeuffer:
I guess 2013 is missing a bit more C++11 features, yes. VS2015 will work. Nightlies and CI is running on it. You could try 2017 ;)

jpfeuffer:
#2693

PR #2736, opened 2017-07-03
[FIX] EnzymaticDigestion for unspecfic cleavage

Leon-Bichmann:
fixed wrong generation of peptides if unspecific cleavage was selected

### DIFF
--- a/src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
+++ b/src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
@@ -277,15 +277,33 @@ namespace OpenMS
     // initialization
     output.clear();
 
-    // naive cleavage sites
-    std::vector<Size> pep_positions = tokenize_(sequence.getString());
-    Size count = pep_positions.size();
-
     // disable max length filter by setting to maximum length
     if (max_length == 0)
     {
       max_length = sequence.size();
     }
+
+    // Unspecific cleavage:
+    // For unspecific cleavages every amino acid is a cutting position.
+    // And all substrings of legnth minx_size to max_size are generated.
+    if (enzyme_.getName() == "unspecific cleavage")
+    {
+      for (Size i = 0; i <= sequence.size() - min_length; ++i)
+      {
+        for (Size j = i + min_length; j <= i + max_length; ++j)
+        {
+          if (j > sequence.size()) { continue; }
+          output.push_back(sequence.substr(i, j - 1));
+        }
+      }
+      return;
+    }
+
+    // Specific cleavage
+
+    // naive cleavage sites
+    std::vector<Size> pep_positions = tokenize_(sequence.getString());
+    Size count = pep_positions.size();
 
     // no cleavage sites? return full string
     if (count == 0) 
--- a/src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
+++ b/src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
@@ -39,6 +39,7 @@
 #include <boost/regex.hpp>
 
 #include <iostream>
+#include <limits>
 
 using namespace std;
 
@@ -86,6 +87,7 @@ namespace OpenMS
   void EnzymaticDigestion::setEnzyme(const String enzyme_name)
   {
     enzyme_ = *EnzymesDB::getInstance()->getEnzyme(enzyme_name);
+    if (enzyme_name == "unspecific_cleavage) { missed_cleavages_ = std::numeric_limits<int>::max(); }
   }
 
   String EnzymaticDigestion::getEnzymeName() const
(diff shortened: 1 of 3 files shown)


jpfeuffer:
Great catch. But do you think you could check if it is possible to make it work with the actual Enzyme regex https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/CHEMISTRY/Enzymes.xml#L198
and the tokenize function?
It would save us this special case. Or did I overlook some problems with this that you already figured out?

timosachsenberg:
I think it is a special feature of the unspecific cleavage. 
It is basically a special case of digest that does not care about specificity and missed cleavages.

One could make it work with the normal tokenizer function if you set missed cleavages to sequence.size(). But considering that unspecific cleavage is a special case that is quite computational expensive a special treatment would be IMHO appropriate. The downside is that this functionality needs also be replicated (as a TODO) in:

- void EnzymaticDigestion::digest(const AASequence& protein, vector<AASequence>& output) const
- peptideCount (= (sequence.size() + 1) \choose 2)
- isValidProduct (= always true)

where the latter two are trivial

timosachsenberg:
peptideCount is not thaaat trivial as I figured because the min and max length needs to be considered... s

timosachsenberg:
maybe better to prefer the compatible version for now so I changed it how you suggested.

jpfeuffer:
So it is slower now because it does not start and stop at min/max length but checks all possible candidates right?

timosachsenberg:
jup. maybe we can make an issue out of it and link to the code so it can be reused for a complete fix.

Leon-Bichmann:
thx Timo I guess I wont be of much help from here on ^^

Leon-Bichmann:
Just getting strange results:
python script
```
for i in range(7,26):
    MassCalc = openms_path + 'Digestor -in {f} -out {o} -min_length {mins} -max_length {maxs} enzyme '.format(f=db_path+'sp-uniprot-human.fasta', o=out_path+'sp_peptides_'+str(i)+'.fasta', m='', mins=i, maxs=i+1)
    subprocess.call(MassCalc.split() + ['unspecific cleavage'])
```

> Statistics:
>   total #peptides after digestion:         99978179
>   removed #peptides (length restrictions): 99978179
>   remaining #peptides:                     0
>

timosachsenberg:
can you check out the last version an try again?

Leon-Bichmann:
Alright, its still not working unfortunately. Same output..

timosachsenberg:
hmm also on the command line?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -86,6 +87,7 @@ namespace OpenMS
   void EnzymaticDigestion::setEnzyme(const String enzyme_name)
   {
     enzyme_ = *EnzymesDB::getInstance()->getEnzyme(enzyme_name);
+    if (enzyme_name == "unspecific cleavage") { missed_cleavages_ = std::numeric_limits<Size>::max(); }
```
cbielow: this is a bad place to put this.
what if the user calls 'setMissedCleavages()' afterwards? 
We need to come up with a better design here. Probably inside 'digest()'.

timosachsenberg:
maybe the missed cleavage parameter gets internally set. hmm I guess I need to add another change to this PR.

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
timosachsenberg: yep I am on it.

timosachsenberg:
@cbielow would it make sense to also use SPEC_SEMI cleavage for unspecific cleavage?
That way one could generate both:
all (n+1)\choose(2) sequences (SPEC_NONE) or
only the N- and C-terminal ones (SPEC_SEMI)
```
TFGQVVAR
 FGQVVAR
  GQVVAR
   QVVAR
    VVAR
     VAR
      AR
        R
TFGQVVA
TFGQVV
TFGQV
TFGQ
TFG
TF
T
```

timosachsenberg:
@cbielow better?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -277,16 +287,32 @@ namespace OpenMS
...
-
     // disable max length filter by setting to maximum length
     if (max_length == 0)
     {
       max_length = sequence.size();
     }
 
+    // Unspecific cleavage:
+    // For unspecific cleavages every amino acid is a cutting position.
+    // And all substrings of legnth minx_size to max_size are generated.
```
cbielow: typos

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -277,16 +287,32 @@ namespace OpenMS
...
+    if (enzyme_.getName() == "unspecific cleavage")
```
cbielow: can we replace this with an enum or some sort of expression which is not as error-prone as this name?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -277,16 +287,32 @@ namespace OpenMS
...
+    {
+      for (Size i = 0; i <= sequence.size() - min_length; ++i)
+      {
+        for (Size j = i + min_length; j <= i + max_length; ++j)
+        {
+          if (j > sequence.size()) { continue; }
```
cbielow: why not compute std::min(i + max_length, sequence.size()-1) and use that in the for loop?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -277,16 +287,32 @@ namespace OpenMS
...
+          output.push_back(sequence.substr(i, j - 1));
```
cbielow: output.reserve() could help, since number of entries is known?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
timosachsenberg: I can move it to a constant.

timosachsenberg: probably the PSI ID should be an integer so we could quickly compare that

PR #2738, opened 2017-07-04
[Addendum] Package rework. Cleanup. Used vars for years.

jpfeuffer:


jpfeuffer:
Also removed the official debian stuff. Was not used yet anyway.
If it goes to apt-get we should rely on dyn libs only and add the version number.

hroest:
>  If it goes to apt-get we should rely on dyn libs only and add the version number.

the debian people anyways have their own patches and use a version system that corresponds to their standards.

PR #2740, opened 2017-07-04
Fix/consensus map id map index order

timosachsenberg:
potentially fixes some of the occasionally occurring

```
in1:  C:\dev\NIGHTLY\builds\masterwin8-VS12_x86-Debug\src\tests\topp\FeatureLinkerUnlabeledQT_6_output.tmp   (line: 92, position/column: 50/79)
				<UserParam type="int" name="map_index" value="!
				                                              2"/>

in2:  C:\dev\NIGHTLY\OpenMS-master\src\tests\topp\FeatureLinkerUnlabeledQT_6_output.consensusXML   (line: 92, position/column: 50/79)
				<UserParam type="int" name="map_index" value="!
				                                              0"/>
```

PR #2744, opened 2017-07-05
Fix/rng fix

### LINKED ISSUE #1584: failed tests with boost >= 1.58
I get a couple of failed tests when compiling against boost 1.58

```
99% tests passed, 5 tests failed out of 1735

Total Test time (real) = 194.53 sec

The following tests FAILED:
        326 - SvmTheoreticalSpectrumGenerator_test (Failed)
        328 - SvmTheoreticalSpectrumGeneratorSet_test (Failed)
        490 - IonizationSimulation_test (Failed)
        491 - MSSim_test (Failed)
        494 - RawTandemMSSignalSimulation_test (Failed)
Errors while running CTest
```

These are the 8 tests that fail with the current develop (Feb 2017) branch and boost 1.61:

```
99% tests passed, 8 tests failed out of 1923

Total Test time (real) =  85.79 sec

The following tests FAILED:
        336 - SvmTheoreticalSpectrumGenerator_test (Failed)
        338 - SvmTheoreticalSpectrumGeneratorSet_test (Failed)
        505 - IonizationSimulation_test (Failed)
        506 - MSSim_test (Failed)
        509 - RawTandemMSSignalSimulation_test (Failed)
        516 - MRMAssay_test (Failed)
        1676 - TOPP_FeatureLinkerUnlabeledQT_5_out1 (Failed)
        1678 - TOPP_FeatureLinkerUnlabeledQT_6_out1 (Failed)
Errors while running CTest

```

hroest:
- fixes #1584 
- after merging this and #2740 we should have working tests

this is only a "black box" approach to make the tests work, I did not investigate why the differences occur and whether they may indicate a larger underlying problem. I could verify that all instances use the RNG and the differences are only occurring when switching to boost > 1.58 so the RNG is the most likely cause for the differences.


hroest:
output: 

```
99% tests passed, 2 tests failed out of 2017

Total Test time (real) =  93.20 sec

The following tests FAILED:
        1742 - TOPP_FeatureLinkerUnlabeledQT_5_out1 (Failed)
        1744 - TOPP_FeatureLinkerUnlabeledQT_6_out1 (Failed)
Errors while running CTest

```

src/tests/class_tests/openms/source/MSSim_test.cpp
```diff
@@ -248,10 +247,19 @@ START_SECTION((SimTypes::MSSimExperiment const& getExperiment() const ))
...
                              InMSLevelRange<SimTypes::MSSimExperiment::SpectrumType>(ListUtils::create<Int>("2")));
 
+#if OPENMS_BOOST_VERSION_MINOR < 56
+  TEST_EQUAL(mssim.getExperiment().getNrSpectra(), 230)
   TEST_EQUAL(nr_ms1, 127)
   TEST_EQUAL(nr_ms2, 103)
 
   TEST_EQUAL(nr_ms2 + nr_ms1, mssim.getExperiment().getNrSpectra())
+#else
+  TEST_EQUAL(mssim.getExperiment().getNrSpectra(), 234)
```
hroest: @cbielow can you comment on why this could happen? is that expected?

src/tests/class_tests/openms/source/RawTandemMSSignalSimulation_test.cpp
```diff
@@ -122,7 +122,11 @@ START_SECTION((void generateRawTandemSignals(const SimTypes::FeatureMapSim &, Si
 
     TEST_EQUAL(exp_with_ms2.size(), exp_no_ms2.size());
     TEST_EQUAL(exp_with_ms2[0].size(), exp_no_ms2[0].size());
+#if OPENMS_BOOST_VERSION_MINOR < 56
     TEST_EQUAL(exp_with_ms2[1].size(), exp_no_ms2[1].size());
+#else
+    TEST_EQUAL(exp_with_ms2[1].size(), exp_no_ms2[1].size()-1);
```
hroest: @cbielow can you comment on why this could happen? is that expected?

src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGenerator_test.cpp
```diff
@@ -95,7 +95,13 @@ START_SECTION(void simulate(PeakSpectrum &spectrum, const AASequence &peptide, b
   PeakMap exp;
   MzMLFile mz_file;
 
+#if OPENMS_BOOST_VERSION_MINOR < 56
   mz_file.load(OPENMS_GET_TEST_DATA_PATH("SvmTheoreticalSpectrumGenerator_test.mzML"),exp);
+  TEST_EQUAL(spec.size(), 7);
+#else
+  mz_file.load(OPENMS_GET_TEST_DATA_PATH("SvmTheoreticalSpectrumGenerator_test_boost58.mzML"),exp);
+  TEST_EQUAL(spec.size(), 8);
+#endif
```
hroest: it seems there is an additional peak with boost > 1.58, @timosachsenberg maybe you can comment on this why that could be the case?

hroest: we are simulating the peptide `IFSQVGK` which should have 7 fragment ions, right? We now get an additional ion at 
216.134 which is a y4++ ion that does not appear with the previous boost...

src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGenerator_test.cpp
```diff
@@ -95,7 +96,16 @@ START_SECTION(void simulate(PeakSpectrum &spectrum, const AASequence &peptide, b
...
+#if OPENMS_BOOST_VERSION_MINOR < 56
   mz_file.load(OPENMS_GET_TEST_DATA_PATH("SvmTheoreticalSpectrumGenerator_test.mzML"),exp);
+  TEST_EQUAL(spec.size(), 7);
+#else
+  mz_file.load(OPENMS_GET_TEST_DATA_PATH("SvmTheoreticalSpectrumGenerator_test_boost58.mzML"),exp);
+  TEST_EQUAL(spec.size(), 8);
+  // the extra peak:
+  TEST_EQUAL(spec.getStringDataArrays()[0][2], "YIon  0++") // TODO: ion_nr is always zero, its actually y4++
+  TEST_EQUAL(spec.getIntegerDataArrays()[0][2], 2)
+#endif
```
hroest: why does changing the RNG produce additional peak in this model? @timosachsenberg maybe you can comment on this why that could be the case?

we are simulating the peptide IFSQVGK? We now get an additional ion at 216.134 which is a y4++ ion that does not appear with the previous boost...seems a bit scary if the result not just numerically but qualitatively (e.g. number of peaks) depends so much on the RNG

hroest:
changed files:

```
$ git diff --stat develop 
 [...]
 src/tests/class_tests/openms/source/MRMAssay_test.cpp                               |    21 +-
 src/tests/class_tests/openms/source/MSSim_test.cpp                                  |    18 +-
 src/tests/class_tests/openms/source/RawTandemMSSignalSimulation_test.cpp            |     4 +
 src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGeneratorSet_test.cpp     |     7 +-
 src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGenerator_test.cpp        |     6 +

```

overall I observed the following:

- the MS Simulator produces simulated files with different numbers of spectra using the new RNG
- the SvmTheoreticalSpectrumGenerator produces different spectra as it uses the RNG and the probabilities of the SVM/SVR to produce spectra
- the MRMAssay class shuffles peptides in a different way due to the new RNG

of these I am not really concerned about the last two but the MS simulation may be worthwhile to check

src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGenerator_test.cpp
timosachsenberg: ok so the difference is only related to boost and not because a different LibSVM is used? I am asking because I think I remember a failing SVMWrapper test in the past. For a different version of libSVM, a prediction value of 0 got assigned to either class -1 or class 1 depending on the version.

hroest: hmm that is also possible as I am building against Ubuntu 16.04 -- however the function call clearly uses a RNG while it uses a pre-trained model so my bet would still be on the boost RNG

timosachsenberg: For different boost versions I currently have no explanaton but if there would be indeed different libSVM version  one explanation would be that a different libSVM version assigns the y4++ peak to "not observed" while the other assigns it to "observed".

hroest:
it clearly uses libSVM and the RNG both:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGenerator.cpp#L934

src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGenerator_test.cpp
hroest: it clearly uses libSVM and the RNG both: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGenerator.cpp#L934 currently its not clear which one causes the issue

timosachsenberg:
ok and it's the same libSVM version I guess?

src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGenerator_test.cpp
timosachsenberg: ok that's strange. So the y4++ is duplicated?

jpfeuffer:
This is great news.

timosachsenberg:
+1 for merge but it would be really great if someone familiar with the code could dig into it why peaks and spectra magically appear.

jpfeuffer:
I asked Sandro and the short answer was: For loss peaks and charge variants (like this doubly charged one) they are only added based on a conditional probability which should involve another Bernoulli dice roll somewhere. He is looking for the exact place.

andreott:
In fact, binned intensities are generated for secondary peaks and presumably the rng change for one peak just hit the boundary between the lowest bin (missing) and the first non-missing intensity bin.

https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGenerator.cpp#L934-L939

timosachsenberg:
but why should a charge 2 ion be created from precursor charge 1?

andreott:
It is just based on conditional probabilities to observe a certain ion (e.g. y++) given that you observed some others (e.g. y+) learned from the training data (similar to the scoring models in PepNovo or . If I remember correctly we chose this instead of training SVMs for all types for performance issues since the SVM prediction is quite slow.

timosachsenberg:
ok I see. Conditioning on the precursor charge could IMHO improve the prediction as this charge determines to a great extent what kind of ions can be observed.

andreott:
Yes, there are individual models (SVMs for primary ions and conditional probs for secondary) trained for different precursor charges.

timosachsenberg:
example for an additional feature in getSimulatedFeatures test

```
               <feature id="f_11708330324181821974">
                        <position dim="0">461.423573060337</position>
                        <position dim="1">437.221198307571</position>
                        <intensity>49003.8</intensity>
                        <quality dim="0">0</quality>
                        <quality dim="1">0</quality>
                        <overallquality>0</overallquality>
                        <charge>3</charge>
...
                        <PeptideIdentification identification_run_ref="PI_0" score_type="" higher_score_better="true" significance_threshold="0" MZ="437.221198307571" RT="461.423573060337" >
                                <PeptideHit score="1" sequence="MTMDKSELVQK" charge="3" protein_refs="PH_0">
                                </PeptideHit>
                                <UserParam type="int" name="RT_index" value="232"/>
                        </PeptideIdentification>
                        <UserParam type="float" name="intensity" value="365.999990701675"/>
                        <UserParam type="string" name="description" value="test-protein-1"/>
                        <UserParam type="int" name="map_index" value="0"/>
                        <UserParam type="float" name="RT_egh_variance" value="10.0678660693975"/>
                        <UserParam type="float" name="RT_egh_tau" value="0.151783252982777"/>
                        <UserParam type="float" name="detectability" value="1"/>
                        <UserParam type="float" name="charge_adduct_mass" value="3.021829400313"/>
                        <UserParam type="string" name="charge_adducts" value="H3"/>
                        <UserParam type="int" name="parent_feature_number" value="8"/>
                        <UserParam type="string" name="parent_feature" value="0"/>
                        <UserParam type="floatList" name="elution_profile_intensities" value="[9.12644384416129e-05, 0.00137220202985904, 0.0122025336956594, 0.0675573158145354, 0.239333029283966, 0.556991889745717, 0.872560725119693, 0.940180510858525]"/>
                        <UserParam type="floatList" name="elution_profile_bounds" value="[119, 448, 126, 462]"/>
                        <UserParam type="floatList" name="isotope_intensities" value="[0.452238497259752, 0.30302253033153, 0.15855069599818, 0.060564459242734, 0.0192313484402927, 0.00517617103591027, 0.00121629769160173]"/>
                </feature>                          
```


```

                <feature id="f_5441031069051211757">
                        <position dim="0">454.165094320208</position>
                        <position dim="1">437.221198307571</position>
                        <intensity>182349</intensity>
                        <quality dim="0">0</quality>
                        <quality dim="1">0</quality>
                        <overallquality>0</overallquality>
                        <charge>3</charge>
...
                        <PeptideIdentification identification_run_ref="PI_0" score_type="" higher_score_better="true" significance_threshold="0" MZ="437.221198307571" RT="454.165094320208" >
                                <PeptideHit score="1" sequence="MTMDKSEVLQK" charge="3" protein_refs="PH_1">
                                </PeptideHit>
                                <UserParam type="int" name="RT_index" value="224"/>
                        </PeptideIdentification>
                        <UserParam type="float" name="intensity" value="735.999989509583"/>
                        <UserParam type="string" name="description" value="test-protein-2"/>
                        <UserParam type="int" name="map_index" value="0"/>
                        <UserParam type="float" name="RT_egh_variance" value="16.0321917036418"/>
                        <UserParam type="float" name="RT_egh_tau" value="-0.373401666948493"/>
                        <UserParam type="float" name="detectability" value="1"/>
                        <UserParam type="float" name="charge_adduct_mass" value="3.021829400313"/>
                        <UserParam type="string" name="charge_adducts" value="H3"/>
                        <UserParam type="int" name="parent_feature_number" value="7"/>
                        <UserParam type="string" name="parent_feature" value="0"/>
                        <UserParam type="floatList" name="elution_profile_intensities" value="[3.6728225991149e-05, 0.000217645022456754, 0.00110761343341531, 0.00485663137995866, 0.0181831008907247, 0.0575642163219618, 0.152480788733753, 0.334128295596028, 0.598284963808948, 0.863849387026463, 0.991512146247152, 0.890900938841021, 0.616492181786776, 0.322906993483853, 0.12572621553934]"/>
                        <UserParam type="floatList" name="elution_profile_bounds" value="[112, 434, 126, 462]"/>
                        <UserParam type="floatList" name="isotope_intensities" value="[0.452238497259752, 0.30302253033153, 0.15855069599818, 0.060564459242734, 0.0192313484402927, 0.00517617103591027, 0.00121629769160173]"/>
                </feature>
```

timosachsenberg:
I merge to see how this affects our nightlies

jpfeuffer:
Not at all yet if it is not merged into master :)

jpfeuffer:
And the release tag would need to be updated.

jpfeuffer:
Or we just move on and I switch the nightlies to develop again.

hroest:
overall we have passing tests on boost 1.54 and we switch to 1.58 without changing the OpenMS code. All failed tests make sense as they are involved with the RNG somehow and it does not seem that any other bugs are introduced that are not involved in the RNG.

PR #2746, opened 2017-07-06
Feature/tml dataweights

dmccloskey:
[FEATURE] added support for basic x and y data weighting (i.e., scaling) prior to fitting  a linear model


hroest:
maybe I miss the point but how does this differ from `computeRegressionWeighted` in ./src/openms/include/OpenMS/MATH/STATISTICS/LinearRegression.h ? Is this more generic as you can use the weight for any transformation, not just linear?

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +63,295 @@ namespace OpenMS
...
+    else
+    {
+      std::cout << "weight " + weight + " is not supported" << std::endl;
+    }
+    return valid;
+  }
+  
+  std::vector<std::string> TransformationModel::getValidXWeights() const
+  {
+    //std::vector<std::string> valid_weights{"1/x","1/x2","ln(x)",""}; C++ 11
```
hroest: we now have C++11 support, maybe you can make this also `static const` ?

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +63,295 @@ namespace OpenMS
...
+        datum_weighted = std::log(10e-5);
+      }
+      else
+      {
+        datum_weighted = std::log(datum);
+      }
+    }
+    else if (weight == "ln(y)")
+    {
+      if (datum < 10e-8)
```
hroest: why do x and y differ in when you take the default weight? e.g. 10e-5 vs 10e-8 ?

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
dmccloskey: OK great.  It was a bit of a pain working around some of the missing C++ features.  Are you refering to the function for the variable declaration when you say "static const"?

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
dmccloskey: I am not completely sure to be honest.  They are the defaults that are used in the proprietary peak integration software called MultiQuant from ABSCIEX that I weaning myself off of.

dmccloskey:
It is supposed to be more generic.  I will eventually like to add support for the other TransformationModel classes as well as a couple other TransformationModels (i.e., Quadratic, Hill, Cubic) that are often used for calibration curves.

timosachsenberg:
Hi David. Nice to see that you are making great progress. One thing I noted is that the weighting function is currently not annotated to the trafoXML. Is this intended?

dmccloskey:
If I forgot to add something it was definitely not intended ;).  Let me know how to go about fixing it or where I should look for direction on fixing and I will do that.

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +63,295 @@ namespace OpenMS
...
   }
+  
+  void TransformationModel::weightData(TransformationModel::DataPoints& data, const Param& params)
+  {
+    // weight x values 
+    std::vector<std::string> valid_weights;
+    valid_weights = getValidXWeights();
+    if (params.exists("x_weight") && checkValidWeight(params.getValue("x_weight"), valid_weights) && !data.empty())
+    {
+      std::string x_weight = params.getValue("x_weight");
```
timosachsenberg: I think you could move this line out of the if block. 
This would also make the if clause simpler.
e.g.:
const std::string x_weight = params.exists("x_weight") ? params.getValue("x_weight") : "";
if (!x_weight.empty() && checkValidWeight(x_weight, valid_weights) && !data.empty())

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +63,295 @@ namespace OpenMS
...
+      x_weight_ = x_weight;
```
timosachsenberg: or directly assign x_weight_ instead of x_weight above. that way you could remove this line

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
timosachsenberg: this would also make the next else and if blocks shorter or even obsolete

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +63,295 @@ namespace OpenMS
...
+  bool TransformationModel::checkValidWeight(const std::string& weight, const std::vector<std::string>& valid_weights) const
+  {    
+    bool valid = false;
+    if (std::find(valid_weights.begin(), valid_weights.end(), weight) != valid_weights.end())
+    {
+      valid=true;
+    }
+    else
+    {
+      std::cout << "weight " + weight + " is not supported" << std::endl;
```
timosachsenberg: consider using the macro LOG_INFO or LOG_WARN

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +63,295 @@ namespace OpenMS
...
+    }
+    return valid;
+  }
+  
+  std::vector<std::string> TransformationModel::getValidXWeights() const
+  {
+    //std::vector<std::string> valid_weights{"1/x","1/x2","ln(x)",""}; C++ 11
+    std::vector<std::string> valid_weights;
```
timosachsenberg: could also use  c11 list initilizers
std::vector<std::string> mem = {"1/x", ..............};

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +63,295 @@ namespace OpenMS
...
+      }
+    }
+    else if (weight == "")
+    {
+      datum_weighted = datum;
+    }
+    else
+    {
+      datum_weighted = datum;
+      std::cout << "weight " + weight + " not supported." << std::endl;
```
timosachsenberg: consider using LOG_INFO and LOG_WARN

timosachsenberg:
Hi Douglas.
It looks quite good. I made some minor comments how the code could be made a bit shorter and maybe easier to maintain.

Other points:

- Either the transformation is not written out in the param tag of the trafoXML or you did not provide an example file:
```
<Param  type="string" name="x_weight" value=""/>
<Param  type="string" name="y_weight" value=""/>
```

As @hroest pointed out, truncating weights by "magic constants" might surprise other users and potentially limit the general applicability of you code. What you could do is introducing, for example, a "x_weight_trunc_min" and "y_weight_trunc_min" parameter. That way, the user can configure how weights get truncated based on the x/y value and this parameter could also be written out in the trafoXML file.

timosachsenberg:
Truncation seems mainly to prevent undefined values (e.g., division by zeros, log(0)). @dmccloskey would it make sense to just throw an error if these values are used? The caller of the function could probably handle these corner cases (e.g. change 0 to a small value) more appropriately.

dmccloskey:
Thank you for your comments Timo.  I apologize for the delay.  I was on vacation last week.  I will implement those comments and fix up the code as you suggested.

dmccloskey:
Having some sort of user parameter to truncate the weights would make things much easier on the user.  Having to handle individual cases where a data point falls below a threshold is no fun (speaking from experience).

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
dmccloskey: Is there documentation or an example of how to use LOG_INFO or LOG_WARN?

timosachsenberg:
e.g., @jpfeuffer or @hroest what is your take on this?

jpfeuffer:
Cutoff parameters sound good, I think. They do not really complicate the interface too much but are more convenient and general. If a user needs special treatment they can put the limits in these parameters and filter beforehand on their own.

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -163,10 +109,28 @@ namespace OpenMS
...
     }
     return valid;
   }
+
+  double TransformationModel::checkDatumRange(const double& datum, const double& datum_min, const double& datum_max)
+  {    
+    double datum_checked = datum;
+    if (datum >= datum_max)
+    {
+      LOG_INFO << "datum " << datum << " is not out of range.";
```
hroest: -- it is *not* out of range or it *is* out of range?

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +64,201 @@ namespace OpenMS
   {
     params.clear();
   }
+  
+  void TransformationModel::weightData(TransformationModel::DataPoints& data, const Param& params)
+  {
+    // get x datum ranges
+    x_datum_min_ = params.exists("x_datum_min") ? (double)params.getValue("x_datum_min") : -1e15;
```
hroest: should this be `-1e15` or `1e-15` here?

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModel.cpp
```diff
@@ -60,5 +64,201 @@ namespace OpenMS
   {
     params.clear();
   }
+  
+  void TransformationModel::weightData(TransformationModel::DataPoints& data, const Param& params)
```
hroest: I think the general interface would not pass the parameters into the function but rather the idea is to do this through `setParameters` and you will have to update the member variables in `updateMembers_`

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
```diff
@@ -44,6 +44,9 @@ namespace OpenMS
   {
     params_ = params;
     data_given_ = !data.empty();
+    TransformationModel::DataPoints data_weighted = data;
+    weightData(data_weighted, params); // weight the data
```
hroest: I dont think this should be the default, lets not apply the transformation and spent all the time doing this when there is no work to be done.

hroest:
In general it looks good but I would prefer if the code flow would stay the same in case there is nothing to do. So basically there should be no overhead if you dont want weighted data.

dmccloskey:
Ok sounds good.  I will change that.  I also need to improve the test
coverage for linear model as well as the adding in the new method interface
for the python wrapper to checkDatumRange

dmccloskey:
I finished addressing the comments so far.  Please let me know if there are any other issues that I should address.

timosachsenberg:
Hi @dmccloskey . Nice progress.

There is an open question regarding the tests: The value entries for:
'''
	<Param  type="string" name="x_weight" value=""/>
 	<Param  type="string" name="y_weight" value=""/>
'''
are empty strings . Shouldn't these indicate the weight function?
How do you handle cases when no weighting is used?

dmccloskey:
By default, the weight function is set to "", indicating no weighting of the data.  When no weighting is used, the methods that apply weighting functions are simply skipped.  The code now adds in default paramaters to specify that no weighting was used.

timosachsenberg:
sounds good. can you add some test files for the weighted cases, too?

dmccloskey:
What test files are you referring to?  I added tests for the weighted cases into the "class_tests".  Are you referring to tests for the topp tools?

timosachsenberg:
Yeah, the TOPP ones.

timosachsenberg:
I think you sufficiently covered the case of using weighting and transforming.
In OpenMS, we support inverting a transformation (see e.g. MapRTTransformer -invert option).
I think this should now also incorporate unweighting?

dmccloskey:
Sorry for the delay in reply.  Yes, since it calls LinearTransform, it should unweight the data automatically.

timosachsenberg:
Can you point me to the test that coverst this? Sorry I can't find it.

dmccloskey:
I test out "weight"/"unweight" functions individually in transformationModel_test.cpp.  I also test linear transformation in the class initialization method in transformationModelLinear_test.cpp.  However, I do not test the "evaluate"/"invert" functions directly.  There were no tests to start with, so I did not add in any for the weight/unweight cases.  Would you like me to add in tests for the evaluate/invert methods as well along with cases for the weighted/unweighted data?

timosachsenberg:
If you find time it would be great to have one round of weight, transform, inverse, unweight to check if we still get the same values

dmccloskey:
OK I will add that in.

dmccloskey:
First off, I apologize for taking so long to get the tests made.  Second, that was a good suggestion on your part.  I forgot to reverse the weights when calling the inverse() function.

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
```diff
@@ -135,9 +135,32 @@ namespace OpenMS
     }
     intercept_ = -intercept_ / slope_;
     slope_ = 1.0 / slope_;
+    
+    // invert the weights:
+    double x_datum_min, x_datum_max, y_datum_min, y_datum_max;
```
timosachsenberg: I think this part could be nicely rewritten using std::swap

timosachsenberg:
no problem at all - take your time.

dmccloskey:
Please let me know if there are any other issues.  The last commit added in the std::swap.

timosachsenberg:
looks good from my side. I will restart the travis build once again.

timosachsenberg:
some tests are failing in travis:
	206 - TransformationXMLFile_test (Failed)
	435 - PoseClusteringAffineSuperimposer_test (Failed)
	436 - PoseClusteringShiftSuperimposer_test (Failed)
	453 - TransformationDescription_test (Failed)
	454 - TransformationModel_test (Failed)
	1315 - TOPP_FileMerger_10_out3 (Failed)
	1427 - TOPP_MapAlignerPoseClustering_1_out5 (Failed)
	1428 - TOPP_MapAlignerPoseClustering_1_out6 (Failed)
	1437 - TOPP_MapAlignerPoseClustering_4_out1 (Failed)
	1461 - TOPP_MapRTTransformer_3 (Failed)
	1462 - TOPP_MapRTTransformer_3_out1 (Failed)
	1530 - TOPP_OpenSwathRTNormalizer_test_1_out1 (Failed)
	1533 - TOPP_OpenSwathRTNormalizer_test_3_out1 (Failed)
	1535 - TOPP_OpenSwathRTNormalizer_test_4_out1 (Failed)
can you check if these might need updated trafoXML files or similar?

dmccloskey:
Tests fixed:
206 - TransformationXMLFile_test (Failed)
435 - PoseClusteringAffineSuperimposer_test (Failed)
436 - PoseClusteringShiftSuperimposer_test (Failed)
453 - TransformationDescription_test (Failed)
454 - TransformationModel_test (Failed)

Tests still failing that use trafoXML:
1315 - TOPP_FileMerger_10_out3 (Failed)
1427 - TOPP_MapAlignerPoseClustering_1_out5 (Failed)
1428 - TOPP_MapAlignerPoseClustering_1_out6 (Failed)
1437 - TOPP_MapAlignerPoseClustering_4_out1 (Failed)
1461 - TOPP_MapRTTransformer_3 (Failed)
1462 - TOPP_MapRTTransformer_3_out1 (Failed)
1530 - TOPP_OpenSwathRTNormalizer_test_1_out1 (Failed)
1533 - TOPP_OpenSwathRTNormalizer_test_3_out1 (Failed)
1535 - TOPP_OpenSwathRTNormalizer_test_4_out1 (Failed)

I am not sure what is wrong for the remaining tests.  I have attached an example output.  If you don't mind interpreting what the problem is, I can fix it.  All of the failed tests are like that.
[failedTestOuput.txt](https://github.com/OpenMS/OpenMS/files/1249077/failedTestOuput.txt)

src/tests/class_tests/openms/source/TransformationModel_test.cpp
```diff
@@ -136,22 +136,22 @@ START_SECTION((double weightDatum(double& datum, const string& weight) const))
   TEST_REAL_SIMILAR(dw.weightDatum(0.0,test), 0.0);
   TEST_REAL_SIMILAR(dw.weightDatum(2.0,test), 2.0);
   test = "ln(x)";
-  TEST_REAL_SIMILAR(dw.weightDatum(0.0,test), std::log(10e-5));
+  TEST_REAL_SIMILAR(dw.weightDatum(0.0,test), std::log(1e-55));
```
timosachsenberg: 1e-55?

src/tests/topp/MapAlignerPoseClustering_1_trafo2.trafoXML
```diff
@@ -2,12 +2,12 @@
 <TrafoXML version="1.0" xsi:noNamespaceSchemaLocation="http://open-ms.sourceforge.net/schemas/TrafoXML_1_0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 	<Transformation name="linear">
 		<Param  type="string" name="symmetric_regression" value="false"/>
-		<Param  type="string" name="x_weight" value=""/>
+		<!-- <Param  type="string" name="x_weight" value=""/>
```
timosachsenberg: hmm I am not sure if this is the correct way to handle this. Wouldn't it make sense to omit these parameter if no weighting functions are specified?

timosachsenberg:
Hi Douglas. I think the reason for the failing tests is, that the parameters x_weight etc. are now written to the trafoXML.

This is in the new file:
1414: in1:  /usr/local/openms-build/src/tests/topp/MapAlignerPoseClustering_1_trafo2.tmp   (line: 4, position/column: 30/45)
1414: 		<Param  type="string" name="!
1414: 		                            x_weight" value=""/>
1414: 

And this is in the test file:
1414: in2:  /usr/local/OpenMS/src/tests/topp/MapAlignerPoseClustering_1_trafo2.trafoXML   (line: 4, position/column: 30/45)
1414: 		<Param  type="string" name="!
1414: 		                            symmetric_regression" value="false"/>

So one possibility is to update the test files.

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
```diff
@@ -44,6 +44,21 @@ namespace OpenMS
   {
     params_ = params;
     data_given_ = !data.empty();
+    // set x_weight
+    params_.setValue("x_weight",params.exists("x_weight") ? (std::string)params.getValue("x_weight") : "");
```
timosachsenberg: Ok I think this part can be improved and might result in the failing tests.
You should check if x_weight exists and is not the empty string.
Only then store the meta values in param_
same for y_weight and the other models.
That way you are backwards compatible to the old test files.

timosachsenberg:
See my comment above. I think the problem is that you write out the x_weight/y_weight and associated min/max parameter even if no weighting is done. So one solution is to update all these files. The better solution would be to omit the weighting information if no weighting is performed. This would mean that you don't need to update the old ones.

timosachsenberg:
If there is no hurry we can finish these last details on the user meeting.

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
dmccloskey: That makes complete sense.

dmccloskey:
I would like to have this feature available before the end of the month (along with another pull request that will be coming hopefully in the next few days).  I will make the modifications that you have mentioned.  If it becomes a huge pain, then I will wait until the user meeting.

dmccloskey:
All tests are passing now (at least on my end).  I ended up just re-making all of the test files.

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/TransformationModel.h
```diff
@@ -68,16 +68,57 @@ namespace OpenMS
...
+    1 / y2	If |y| < 10-8 then w = 10e16; otherwise w = 1 / y2.
+    ln x	If x < 0 an error is generated; otherwise if x < 10-5 then w = ln 10e-5,
+    otherwise w = |ln x|.
+    */
+    virtual void weightData(DataPoints& data, const Param& params);
+    
+    /// Unweight the data by the given weight function
+    virtual void unWeightData(DataPoints& data, const Param& params);
+    
+    ///
```
timosachsenberg: please check that all newfunctions have proper comments

timosachsenberg:
@dmccloskey I would like to have another review by @hendrikweisser as he has recently been working with trafoXML

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/TransformationModel.h
```diff
@@ -72,37 +72,50 @@ namespace OpenMS
...
-    /// 
+    /**
+    @brief Check for a valid wighting function string
+    */
     bool checkValidWeight(const std::string& weight, const std::vector<std::string>& valid_weights) const;
 
-    ///
+    /**
+    @brief Check that the data is within the valid min and max bounds
+    */
```
timosachsenberg: should be clear what the output is in case the datum is /is not in range

dmccloskey:
Let me know if there is an ETA for when this can be merged.

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
```diff
@@ -92,7 +116,12 @@ namespace OpenMS
 
   double TransformationModelLinear::evaluate(double value) const
   {
-    return slope_ * value + intercept_;
+    double weighted_value;
+    weighted_value = weightDatum(value, x_weight_);
+    double eval;
+    eval = unWeightDatum(slope_ * weighted_value + intercept_, y_weight_);
+    return eval;
+    // return slope_ * value + intercept_;
```
hroest: this seems likely to impose a performance penalty even if no weighting is done

hroest:
generally looks good, the only question from a architecture point of view is whether we should have a separate class TransformationModelWeighted.h that implements the weighting. As I commented, this may be conceptually nicer and may not lead to a performance penalty if no weighting is done.

dmccloskey:
I didn't make a new classs, but I modified the method to not weight/unweight the data if no weighting is specified.  Let me know if this would work.

hroest:
that sound good, I did some refactoring here: https://github.com/dmccloskey/OpenMS/pull/1

I think that we should only supply parameters once, and I was able to remove a lot of code duplication between parent / child class. This reduces errors and makes maintenance easier. Also I added another member variable that keeps track of whether any weighting is done or not and this should make the regular (non-weighted) case faster. Please accept my PR and then we should be ready to go -- and remove the code tags from MSVS

hroest:
can you fix these errors:

- http://cdash.openms.de/viewTest.php?onlyfailed&buildid=244236

it seems you missed a few places where the number of parameters got changed
```
PoseClusteringAffineSuperimposer_test | Failed | 130ms | Completed (Failed)
PoseClusteringShiftSuperimposer_test | Failed | 120ms | Completed (Failed)
TOPP_MapRTTransformer_3_out1 | Failed | 40ms | Completed (Failed)


 -  line 119:  TEST_EQUAL(params.size(),8): got 2, expected 8
```

also 
```
723>C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\source\src\tests\class_tests\openms\source\TransformationModel_test.cpp(139): error C2124: divide or mod by zero [C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\build\src\tests\class_tests\openms\TransformationModel_test.vcxproj]
```

hroest:
I dont think this really works:

```
double inf = 1.0/0.0;
```

dmccloskey:
All tests are passing, and inf has been changed.

hroest:
great, lets see what windows says. if it all passes, I think we can merge -- maybe we can squash the commits first? to clean up and not commit the MSVS tags files into the codebase

dmccloskey:
Am I able to merge the PR and squash the commits?  If so I can go ahead and do that.

timosachsenberg:
+1 I squashed and merged

PR #2751, opened 2017-07-06
Master: sync

hroest:


timosachsenberg:
Question is what should we do with master. Would a new tag 2.2.1 be sufficient once we integrated all hotfixes? Or do you think we need a hotfix release with changelog, version number changes in the code, tagging of contrib and all other stuff too?

hroest:
given that the compiled code does not change at all, I dont think thats necessary and a new tag is sufficient. However, a new tarball would be a good idea.

PR #2752, opened 2017-07-06
Fix/map deterministic

hroest:
use `std::map` instead of `boost::unordered_map` for deterministic output


hroest:
this is the last missing piece to get all tests to pass on 17.04

PR #2757, opened 2017-07-07
Fix/gcc7 warnings

hroest:
fixes some issues

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/MASSDECOMPOSITION/IMS/RealMassDecomposer.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/MASSDECOMPOSITION/IMS/RealMassDecomposer.h
@@ -126,7 +126,7 @@ private:
         Decomposer to be used for exact decomposing using
         integer arithmetic.
       */
-      std::auto_ptr<integer_decomposer_type> decomposer_;
+      std::shared_ptr<integer_decomposer_type> decomposer_;
     };
 
   } // namespace ims
--- a/src/openms/include/OpenMS/CONCEPT/GlobalExceptionHandler.h
+++ b/src/openms/include/OpenMS/CONCEPT/GlobalExceptionHandler.h
@@ -139,11 +139,7 @@ protected:
       throw();
 
       /// The OPENMS new handler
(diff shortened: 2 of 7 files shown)


src/openms/include/OpenMS/CHEMISTRY/MASSDECOMPOSITION/IMS/RealMassDecomposer.h
```diff
@@ -126,7 +126,7 @@ namespace OpenMS
         Decomposer to be used for exact decomposing using
         integer arithmetic.
       */
-      std::auto_ptr<integer_decomposer_type> decomposer_;
+      std::shared_ptr<integer_decomposer_type> decomposer_;
```
timosachsenberg: would unique_ptr also work?

hroest: not sure about the copy constructor here, it may fail when building a copy constructor as `unique_ptr` cannot be copied.

timosachsenberg: where? I just saw one position where make_unique should work

hroest: well basically I was worried that `unique_ptr` may not work out of the box. See https://stackoverflow.com/questions/16030081/copy-constructor-for-a-class-with-unique-ptr#16030118

- Since the unique_ptr can not be shared, you need to either deep-copy its content or convert the unique_ptr to a shared_ptr.

so no real reason but I felt shared_ptr is less work and works as-is

timosachsenberg: Ok I see. Fine with me. I was just asking because I saw a video on C11 smart pointers (https://www.youtube.com/watch?v=JfmTagWcqoE). As I understood it, it is better to use unique_ptrs in this case and perform deep copies.

PR #2758, opened 2017-07-09
[BUILD] quickbuild script

hroest:
- allows quick and easy build
- should work on most Linux (and probably OSX)
- provides the one-step build experience


hroest:
basically creates three directory

- contrib
- contrib-build
- openms-build

and builds the contrib and openms both out-of-source in a one-step process.

hroest:
@jpfeuffer @timosachsenberg should we merge?

jpfeuffer:
Sure! why not. Merged.

PR #2482, opened 2017-03-26
Feature/TopPerc

mwalzer:
Wrapped up TopPerc after refactoring the Feature into core class and PSMFeatureExtractor and Percolator adapter. Finally added more documentation and tests.
Multi-search engine feature is still under development as is the mascot feature set on which testing support is very welcome. Marked as experimental in the meantime.

PR #2761, opened 2017-07-10
Fix/top perc

timosachsenberg:
replaces https://github.com/OpenMS/OpenMS/pull/2482


timosachsenberg:
I think its fine to merge as experimental util now

Leon-Bichmann:
Im getting an error using PSMFeatureExtractor:
Internal error: Request for unregistered parameter 'mzid_out'

seems like the flat mzid_out was removed but the variable is still there

timosachsenberg:
potentially fixed with https://github.com/OpenMS/OpenMS/pull/2776

PR #2767, opened 2017-07-12
[FIX] reduced memory consumption

timosachsenberg:
Post-filtering by length led to out of memory problems for unspecific cleavage.
This small change filters after every digest of a protein.


jpfeuffer:
+1

Issue #2747, opened 2017-07-06
Cleanup WinCI Tests


jpfeuffer:
See for example http://cdash.openms.de/viewTest.php?onlyfailed&buildid=229889.
With #2744 and associated ones we are left with:

@enetz @lukaszimmermann 
```
UTILS_OpenPepXL_1_out_2	Failed	80ms	Completed (Failed)
UTILS_OpenPepXLLF_1_out_2	Failed	100ms	Completed (Failed)
```

@hendrikweisser 
`TOPP_FeatureFinderIdentification_2_out1	Failed	150ms	Completed (Failed)`

And probably some multithreading/configuration/windows API thing:
`StopWatch_test	Failed	18s 220ms	Completed (Failed)`

Could you have a look into what might be the issues there? Did I tag the right people?

jpfeuffer:
Yes, can confirm that this solved the first two issues thanks.

jpfeuffer:
StopWatch fixed in #2797

jpfeuffer:
Last error on Windows is the FeatureFinderIdentification_test.
```
Offending lines:			(tab_width = 8, first_column = 1)

in1:  C:\jenkins\ws\openms_nightly_packaging\c2e226b2\build\src\tests\topp\FeatureFinderIdentification_2.tmp   (line: 637, position/column: 19/41)
			<overallquality>!
			                0.455317</overallquality>

in2:  C:\jenkins\ws\openms_nightly_packaging\c2e226b2\source\src\tests\topp\FeatureFinderIdentification_2_output.featureXML   (line: 637, position/column: 19/41)
			<overallquality>!
			                0.405836</overallquality>
```
@hendrikweisser Any thoughts? Could this be a problem with different random number generators? Or a sorting difference?

hroest:
@jpfeuffer can we close this now?

jpfeuffer:
No unfortunately the FFID test error still persists.

hroest:
@jpfeuffer we also have these issues: http://cdash.openms.de/viewBuildError.php?buildid=246937 which look a bit like it does not understand C++11:

```
  Building Custom Rule /.../OpenMS-master/src/openms/CMakeLists.txt
  CMake does not need to re-run because C:\dev\NIGHTLY\builds\masterwin8-VS12_x86-Release\src\openms\CMakeFiles\generate.stamp is up-to-date.
  Generating include/OpenMS/SYSTEM/moc_FileWatcher.cxx
[...]

PR #2768, opened 2017-07-13
Fix/xquest spec xml rounding error

enetz:
fixes OpenPepXL from #2747 

Spectra in the Base64 encoded xquest spec.xml format were written slightly differently on Linux and Windows.
Two issues were identified and fixed:

1. somehow the double type mz values had a different least significant digit for some peaks after encoding and decoding, strangely always an increase from ...2 to ...3 from Linux to Windows
e.g. Linux: 464.282836914062, Windows: 464.282836914063.
I did not check the double values before encoding and am not sure where exactly this happened,
but these values were not otherwise manipulated besides being read from an mzML, converted into Strings, encoded from Strings using `Base64::encodeStrings()` and finally decoded for testing. I solved this by rounding the numbers. This output format is purely for visualization and we do not need a very high precision for that.

2. Some peaks appeared twice in the spectra with different charges assigned in an `IntegerDataArray`.
When sorting such spectra by position, only the mz values are considered. Unfortunately it seems our
`MSSpectrum::sortByPosition()` function is not consistent in the order of equal values between compilers, so the peaks with different charges appeared at different positions on Windows and Linux.
Since assigning multiple charges to the same peak was a bug anyway, I decided to only use the smallest charge for each peak.


timosachsenberg:
"Unfortunately it seems our
MSSpectrum::sortByPosition() function is not consistent in the order of equal values between compilers"
hmm sounds like we should use http://www.cplusplus.com/reference/algorithm/stable_sort/

timosachsenberg:
I also opened a small fix for the non-stable sorting https://github.com/OpenMS/OpenMS/pull/2769
(edited online so please wait till it passes)

PR #2770, opened 2017-07-13
make sorting deterministic in case of equivalent values

timosachsenberg:


enetz:
+1
`std::stable_sort` is a bit slower than `std::sort`, but it is better than checking every `DataArray` for differences in annotation.

Issue #766, opened 2014-03-27
Performance of Eigen splines


hroest:
Due to the switch from GSL splines to Eigen splines #651 the performance of the spline implementation seems to have suffered. Compare the FeatureFinderHiRes of 1.11 vs the current HEAD

```
# Release 1.11
$ ./bin/PeakPickerHiRes -in test.mzML -algorithm:signal_to_noise 0.0
Progress of 'loading mzML file':
-- done [took 40.79 s (CPU), 41.56 s (Wall)] --
Progress of 'smoothing data':
-- done [took 12.92 s (CPU), 12.99 s (Wall)] --
Progress of 'storing mzML file':
-- done [took 0.17 s (CPU), 0.23 s (Wall)] --
PeakPickerHiRes took 56.36 s (wall), 54.45 s (CPU), 0.00 s (system), 54.45 s (user).

# Current HEAD
$ ./bin/PeakPickerHiRes -in test.mzML -algorithm:signal_to_noise 0.0
File '/home/hr/.OpenMS/OpenMS.ini' is deprecated.
Progress of 'loading spectra list':
-- done [took 25.48 s (CPU), 26.38 s (Wall)] --
Progress of 'picking peaks':
-- done [took 01:07 m (CPU), 01:08 m (Wall)] -- 
Progress of 'storing mzML file':
-- done [took 0.19 s (CPU), 0.25 s (Wall)] -- 
PeakPickerHiRes took 01:35 m (wall), 01:34 m (CPU), 0.00 s (system), 01:34 m (user).

```

when checking the profiler, one can see that most of the time is spent computing and evaluating the spline. This is a severe difference from 13s to 67s in the peak picking part of the algorithm

PR #2773, opened 2017-07-14
[FIX] Remove spline2d class, and update remaining files with references to it

timosachsenberg:


timosachsenberg:
completes https://github.com/OpenMS/OpenMS/pull/1033

timosachsenberg:
fixed https://github.com/OpenMS/OpenMS/issues/766

Merging and fixing the original branch was a bit more trouble than adding the few changed lines by hand. So I opened a new PR.

jpfeuffer:
cool looks good.

jpfeuffer:
Sorry but I think we forgot about pyOpenMS.

jpfeuffer:
Is there still the class test left, too?

jpfeuffer:
Hmm actually there is no occurrence in the pxds. But we have failures in the pyOpenMS builds related to Spline2d. Maybe pyOpenMS tries to wrap the class_test?
The class test was commented out and was therefore forgotten I think.
https://github.com/OpenMS/OpenMS/blob/ec89b2f6406d19ba41d0137e64f6e7396736fffc/src/tests/class_tests/openms/executables.cmake#L270

jpfeuffer:
I'll create a PR.

PR #2775, opened 2017-07-14
Feature/tbb

hroest:
This adds experimental support for TBB in the OpenMS build system

- basically it reactivates code that we had previously 7a75774ee7f4655d667623061bc1a7e5f5bee7a6
- TBB complements OpenMP in that it provides thread-safe containers and parallel algorithms
- if we want to make OpenMS truly parallel we need TBB at some point

the reason for doing this now is that we have a student working on the OpenSWATH code and he used TBB to improve some locking mutexes and work around the limitations that we have in our code (e.g. global singletons).


timosachsenberg:
Probably a good idea to add it back.

cbielow:
how would installation work in terms of contrib? For Windows its probably some effort. Other platforms can use package managers?

jpfeuffer:
C++17 STL also offers parallel algorithms, we should keep that in mind

timosachsenberg:
I read a bit into it and I think it's possible to combine TBB with (and also later switch completely) to the upcoming C++ features. Still, I think TBB is beyond what C17 offers so just bumping up the requirements would not help.

hroest:
I also think that we should use TBB as we cannot switch to C17 for everybody (compiling this on older machine is going to be a nightmare). I think we can add this PR to the build system now and then add it to the contrib (right now there is no TBB code in OpenMS) for all three platforms. It would be great if we could get some support from Windows/OSX people on that second step. Regarding OSX, I dont know whether `brew install` is sufficient but I think it should be, for Windows I guess we will have to compile it (or maybe there are binaries? - at least here it says so https://www.threadingbuildingblocks.org/download that the downloads include the binaries, so maybe no compilation necessary?).

cmake/multithreading.cmake
```diff
@@ -37,6 +37,24 @@
...
+message(STATUS "Intel TBB: ${MT_ENABLE_TBB}")
+if (MT_ENABLE_TBB)
+  find_package(TBB)
+  if (NOT TBB_FOUND)
+    message(FATAL_ERROR "TBB not found but requested.")
+  endif()
+endif()
+
+if (TBB_FOUND)
+  INCLUDE_DIRECTORIES(${MT_TBB_INCLUDE_DIR})
```
jpfeuffer: Did you test this? Where will this variable be set to a reasonable value?

cbielow:
ok. for Windows this looks easy:
In Cmake: download zip, extract binary subfolder matching the VS version. Extract ./include. Done.

cmake/multithreading.cmake
hroest: no, I did not check, I just copied it from 7a75774

hroest:
ok, I checked the variables and they contain the libraries and the header dirs - so I believe this should work in principle. 

this PR does not change anything at the moment but prepares future PR.

Can we merge ?

timosachsenberg:
Ok fine with me.

timosachsenberg:
I squas and merge so it would be easy to revert.

PR #2782, opened 2017-07-19
Removes templates for MSSpectrum and MSChromatogram

timosachsenberg:


timosachsenberg:
Simplifies code base significantly and improves interaction with pyOpenMS.

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/LinearResamplerAlign.h
```diff
@@ -164,13 +213,59 @@ namespace OpenMS
         @param resample_it Iterator pointing to end of the output spectrum range (m/z need to be populated, intensities should be zero)
 
     */
-    template <typename PeakTypeIterator, typename ConstPeakTypeIterator>
-    void raster(ConstPeakTypeIterator raw_it, ConstPeakTypeIterator raw_end, PeakTypeIterator resample_it, PeakTypeIterator resample_end)
+    void raster(MSChromatogram::ConstIterator raw_it,
```
hroest: why not just leave the template here? seems like a copy-paste issue here that is perfectly suited for a template?

src/openms/include/OpenMS/KERNEL/MSChromatogram.h
```diff
@@ -51,30 +51,26 @@ namespace OpenMS
     @brief The representation of a chromatogram.
     @ingroup Kernel
   */
-  template <typename PeakT>
-  class MSChromatogram :
-    private std::vector<PeakT>,
+
+  class OPENMS_DLLAPI MSChromatogram :
```
hroest: I wonder whether some of the small utility functions should not stay in the header for performance reasons? Or does LTO take care of that?

hroest:
A few comments:

- I dont like the de-templating of things too much - we can still put templates in the .cpp files and there will just be 2 instances created, one for MSSpectrum and one for MSChromatogram. There is no need to duplicate code
- Moving critical small functions from headers to cpp may have performance impact unless we enable link time optimization: https://stackoverflow.com/questions/7223993/will-c-linker-automatically-inline-functions-without-inline-keyword-withou
  * So I guess if you want link time optimization, use the -flto (for GCC) or /GL (for the MSVC compiler) and /LTCG (for the MSVC linker) options.
- I currently dont see any such flags in `cmake/compiler_flags.cmake` and we should probably add them for general performance as well. I think for clang we would need the gold linker?

hroest:
PS; also pyOpenMS is not done yet, did you try and compile it?

src/openms/include/OpenMS/KERNEL/MSChromatogram.h
timosachsenberg: With LTO it's certainly no problem. Without LTO I am pretty sure that it makes (if at all) a neglectable difference. Most time will be most likely be spent on moving the spectrum into the cache.

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/LinearResamplerAlign.h
timosachsenberg: Ok I can change that back

timosachsenberg:
Let me address some of your comments.

"I dont like the de-templating of things too much - we can still put templates in the .cpp files and there will just be 2 instances created, one for MSSpectrum and one for MSChromatogram. There is no need to duplicate code"

I removed the template for MSSpectrum and MSChromatogram because:
- nobody ever used it (except misuse aka MSSpectrum<ChromatogramPeak>)
- it is in my view the wrong level of abstraction/generalization. Given that these classes are not mere containers but carry meta information etc. (note: we might introduce a RawPeakDataArrays base that replaces the inheritance from vector<>. If these RawPeakDataArrays are templates might be subject for discussion)
- algorithms and filters should use Position and Intensity iterators and provide overloads for MSSpectrum and MSChromatogram for easy wrapping in python 

"Moving critical small functions from headers to cpp may have performance impact unless we enable link time optimization: https://stackoverflow.com/questions/7223993/will-c-linker-automatically-inline-functions-without-inline-keyword-withou
So I guess if you want link time optimization, use the -flto (for GCC) or /GL (for the MSVC compiler) and /LTCG (for the MSVC linker) options.
I currently dont see any such flags in cmake/compiler_flags.cmake and we should probably add them for general performance as well. I think for clang we would need the gold linker?"

I think this is no problem in practice even without LTO  but to be sure we might need to benchmark this. Do you have some candidates in mind? Those in MSSpectrum and MSChromatogram are in the same compile unit so expensive stuff like sorting is optimized,

hroest:
I am not argueing against removing the template parameters of MSSpectrum / MSChromatogram, but I dont like the change you made in `LinearResamplerAlign.h ` since I think we should *move* towards using algo-classes that are templated and can use both, MSSpectrum and MSChromatogram. The conceptual idea is the same as you had, it should just be data and there should be no distinction between sticking a spectrum or a chromatogram into the algo. In think in the future we should use an interface that is either based on `std::vector` or at least use a data-interface class from which both MSSpectrum and MSChromatogram inherit.

> it is in my view the wrong level of abstraction/generalization. Given that these classes are not mere containers but carry meta information etc. (note: we might introduce a RawPeakDataArrays base that replaces the inheritance from vector<>. If these RawPeakDataArrays are templates might be subject for discussion)

yes, but from an algorithms point of view, they should be just data containers. So again, either we change the interface to `std::vector` or we write some nice interface. I have done both in the past for OpenSwath and we should discuss this further - we even have `OpenMS/INTERFACES/`. I am in favour of pure `std::vector` interfaces just because then it will be easier for other people to use our algorithms.

> algorithms and filters should use Position and Intensity iterators and provide overloads for MSSpectrum and MSChromatogram for easy wrapping in python

we will have to look at an example, in general wrapping objects tends to be easier than wrapping iterators in Python

> I think this is no problem in practice even without LTO but to be sure we might need to benchmark this. Do you have some candidates in mind? Those in MSSpectrum and MSChromatogram are in the same compile unit so expensive stuff like sorting is optimized,

maybe I misunderstood something here, but I am talking about the linking step of linking all the .o files together to the OpenMS.so file. I thought this is what the stackoverflow post describes, if you just link together object files, these jumps are by defauly *not* optimized. This is not such an issue when everything was in the header but now it may be an issue

PS: did compile time decrease now?

timosachsenberg:
T: algorithms and filters should use Position and Intensity iterators and provide overloads for MSSpectrum and MSChromatogram for easy wrapping in python
H: we will have to look at an example, in general wrapping objects tends to be easier than wrapping iterators in Python

Ok I think this was a misunderstanding. I would provide functions filter(MSSpectrum & ) and filter(MSChromatogram & c)
as the interface pyOpenMS uses. Both would just call the templated (on the iterator type) filter(InputPosItBegin, InputPosItEnd, ...) generic function.

timosachsenberg:
Compile times did not decrease.

hroest:
>  as the interface pyOpenMS uses. Both would just call the templated (on the iterator type) filter(InputPosItBegin, InputPosItEnd, ...) generic function.

:+1: this would also allow, potentially, to apply algorithms only on a part of a spectrum / chromatogram

timosachsenberg:
Note:  providing a position and intensity iterator seems pretty convenient using boost
https://stackoverflow.com/questions/36824062/how-can-i-make-an-iterator-over-fields-of-containers-items

src/pyOpenMS/CMakeLists.txt
```diff
@@ -62,25 +62,35 @@ execute_process(
 message(STATUS "Python found at ${PYTHON_EXECUTABLE} with version ${PY_VERSION} (if this is wrong, configure with -DPYTHON_EXECUTABLE:FILEPATH=/path/to/python)")
 
 #------------------------------------------------------------------------------
-# Windows support restricted to Python 2.7 at the moment!
-#  * we also require VC 2008 and the corresponding vcredist to be installed
+# Windows support restricted to Python 2.[67] and Python 3.[34] at the moment!
+#  * we also require VC 2008 / 2010 and the corresponding vcredist to be installed
 if(WIN32)
-    if(NOT MSVC90)
+    if(MSVC90 AND (${PY_VERSION} STREQUAL "2.6" OR ${PY_VERSION} STREQUAL "2.7" ) )
```
hroest: hmm not sure why this shows up here, I think this was a PR against master and the sync is just proposed?

timosachsenberg:
build pyOpenMS

src/openms/include/OpenMS/KERNEL/MSChromatogram.h
hroest: I agree as only complicated functions are moved to the cpp. But lets not move simple functions there as this will incur an overhead of a function call without LTO

timosachsenberg:
I made some performance measurements to check the impact of removing the inline.
It's in the range of 10% for access heavy tasks like feature detection.

hroest:
>  It's in the range of 10% for access heavy tasks like feature detection.

actually, to me 10% seems a lot just for overhead of calling functions. do you know which functions these are? maybe it makes sense to not move all of them to the .cpp -  or maybe we should first do the de-templatization and in a second step move stuff to the .cpp

cbielow:
I tried LTO some years ago, and the major drawback was that any change in source code would result in a full library link, which takes ages. So its definitely something one would want to avoid when developing code. This should be done only in binary releases (or we'd need to becnhmark how bad it really is nowadays).
Then again, if you're writing a fancy algorithm, you'd want to have performance changes reflected immediately and not rely on LTO everytime to see the full picture.
So, without actually knowing for sure, my gut feeling says LTO is a hassle since it slows development/compile and adds an extra dimension when benchmarking.

timosachsenberg:
is link time a problem if the gold linker is used?

cbielow:
it would be on Windows :)

timosachsenberg:
I am personally for a merge as it is kind of tedious to synchronize this branch with develop.

timosachsenberg:
build pyopenms

timosachsenberg:
travis has one timeout https://travis-ci.org/OpenMS/OpenMS/builds/264655438?utm_source=github_status&utm_medium=notification all others passd

jpfeuffer:
I am also fine with merging and then thinking about inlining.

Issue #2783, opened 2017-07-19
AccurateMassSearch cannot write output file (regression)


hendrikweisser:
I've just updated OpenMS in KNIME and found that the AccurateMassSearch node in an existing workflow doesn't work anymore. The output in the KNIME log was:
> ERROR AccurateMassSearch   0:16       Failing process stdout: [Read 54 entries from mapping file!, Read 1 entries from adduct file 'C:\Users\hendrik.weisser\Data\Nucleotide-MS\pos_adducts.tsv'., Read 15 entries from adduct file 'C:\Users\hendrik.weisser\Software\OpenMS\share\OpenMS\CHEMISTRY\NegativeAdducts.tsv'.]
> ERROR AccurateMassSearch   0:16       Failing process stderr: []
> ERROR AccurateMassSearch   0:16       Execute failed: Failed to execute node AccurateMassSearch

So no useful information.

I then updated my local OpenMS installation and tried AccurateMassSearch there. I got the following (weirdly worded) error:
> Error: Unable to write file (the file 'While storing 'Search\results.csv'. Invalid file extension. Should be: 'tsv'' could not be created)

If I try to write to an output file with ending ".tsv", I get this instead:
> Invalid parameter: Invalid output file extension 'Search\results.tsv'. Valid file extensions are: 'csv'.

So it seems that two different parts of the code can't agree on what the file extension should be.
(Using ".tmp" as the extension makes it work on the command line - which is also why the test still works! However, that doesn't help with KNIME.)

I'm going to prepare a fix on the OpenMS side. For KNIME, it would be good to know why the error message from the tool doesn't appear anywhere. (I'm assuming the failure in KNIME is caused by the same bug - any way I can test that?)

timosachsenberg:
maybe related to my change that forces checks of file extensions?

hendrikweisser:
> maybe related to my change that forces checks of file extensions?

Yep! See PR #2784.

Issue #1462, opened 2015-06-11
change MzTabExporter output file extension from *.tsv to *.mzTab


lars20070:
Currently, `MzTabExporter` writes output files with the extension `*.tsv`. It would be clearer to use the extension `*.mzTab` instead [1]. Any post-processing scripts that transform the `mzTab` to a better readable format could then use the `*.tsv` extension (`mzTab2tsv`).

[1] https://code.google.com/p/mztab/wiki/ExampleFiles

timosachsenberg:
good idea but we need to be careful to not brake anything on the KNIME side

lars20070:
@jpfeuffer Where in *Knime* do we need to introduce the new `mzTab` file type?

lars20070:
Fixed in #3403.

PR #2784, opened 2017-07-19
Fix for #2783 (AccurateMassSearch, output file types)

hendrikweisser:
Small fix in AccurateMassSearch, and improvement of the error messages for invalid file types.

### DIFF
--- a/src/openms/include/OpenMS/CONCEPT/Exception.h
+++ b/src/openms/include/OpenMS/CONCEPT/Exception.h
@@ -638,7 +638,7 @@ public:
       public BaseException
     {
 public:
-      UnableToCreateFile(const char* file, int line, const char* function, const std::string& filename) noexcept;
+      UnableToCreateFile(const char* file, int line, const char* function, const std::string& filename, const std::string& message = "") noexcept;
     };
 
     /**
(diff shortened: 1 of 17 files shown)


timosachsenberg:
looks good +1. Would it make sense to accept both csv and tsv?

hroest:
> looks good +1. Would it make sense to accept both csv and tsv?

sounds good but I think TSV is reserved for something, e.g. msInspect file, see https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/FORMAT/FileTypes.h#L88 and L94 

I think we should change that, it does not make sense to have TSV only be one file format and honestly I think we should accept TSV and CSV always as people dont distinguish the two (myself included)

hendrikweisser:
> I think we should change that, it does not make sense to have TSV only be one file format and honestly I think we should accept TSV and CSV always as people dont distinguish the two (myself included)

I think the previous guideline (defined by Oliver?) was to use .csv consistently for every character-delimited, tabular text format. One advantage of this is that on Windows, files with .csv extension can be opened in Excel by double-clicking on them (.tsv not, by default).

hendrikweisser:
By the way, why don't we use .mzTab for mzTab files?

lars20070:
@hendrikweisser https://github.com/OpenMS/OpenMS/issues/1462

timosachsenberg:
feel free to change to mzTab

hendrikweisser:
> feel free to change to mzTab

Tried it, but when I then run AccurateMassSearch, I get this error:

```
FATAL: uncaught exception!
exception of type InvalidParameter occured in line 1015, function void __cdecl OpenMS::TOPPBase::setValidFormats_(const class OpenMS::String &,const class std::vector<class OpenMS::String,class std::allocator<class OpenMS::String> > &,const bool) of C:\Users\hendrik.weisser\Software\OpenMS\src\openms\source\APPLICATIONS\TOPPBase.cpp
error message: The file format 'mzTab' is invalid!
```

Is mzTab not defined as a file type currently?

timosachsenberg:
nope https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/FORMAT/FileTypes.h

jpfeuffer:
Seems like you also need to change the tests in DataFilter. They check the exact Exception message (which is actually a bit weird).

hendrikweisser:
@jpfeuffer: Okay, I'm taking care of it.

hendrikweisser:
@jpfeuffer: Can we get a new version of the KNIME plug-in that includes this fix, please? Without it, AccurateMassSearch is unusable in the current version.
(I've tried to revert to the previous working version in KNIME, but it errors out saying: "No repository found containing: ...")

PR #2789, opened 2017-07-22
[BUILD] use 2 parallel jobs for build

hroest:
According to this [1] we have 2 CPUs for our build, so it makes sense to use both of them but not use more than 2. 

Ref

1. https://docs.travis-ci.com/user/reference/precise/#Virtualization-environments

### DIFF
--- a/tools/travis/cibuild.cmake
+++ b/tools/travis/cibuild.cmake
@@ -45,7 +45,7 @@ set (CTEST_CUSTOM_WARNING_EXCEPTION
     )
 
 # try to speed up the builds so we don't get killed
-set(CTEST_BUILD_FLAGS -j5)
+set(CTEST_BUILD_FLAGS -j2)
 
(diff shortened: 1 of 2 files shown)


jpfeuffer:
As far as I know, ctest_build does not have a PARALLEL option (https://cmake.org/cmake/help/v3.3/command/ctest_build.html)
Only ctest_test.
For building I think you need to use compiler flags (which might be different across platforms.)

jpfeuffer:
And I also experienced what is currently happening. Even though they claim that you only have 2 cores, there is still some speed up if you go higher :)
We could try to find some optimum though

jpfeuffer:
Build pyopenms

hroest:
> And I also experienced what is currently happening. Even though they claim that you only have 2 cores, there is still some speed up if you go higher :)
:-)

PR #2792, opened 2017-07-25
Build pyOpenMS under Windows

hroest:
contains all fixes necessary to build pyOpenMS eggs/wheels under Windows

- replaces the assumption that we are using MSVS 2008
- correctly renames all boost libs
- now also copies sqlite3.dll


hroest:
build failures are unrelated (due to migration of travis to trusty and missing glpk package)

hroest:
@jpfeuffer maybe we can even think about adding pyOpenMS to one of the build servers on Windows/ It seems that they work more reliably now and adding pyOpenMS is now not such a huge pain any more

jpfeuffer:
Okay. I could also add a pyOpenMS build to CI whenever a PR has a certain title or something. Also doing pyOpenMS on all platforms should be no problem to do Nightly.

hroest:
> Okay. I could also add a pyOpenMS build to CI whenever a PR has a certain title or something. Also doing pyOpenMS on all platforms should be no problem to do Nightly.

ok, that would be great! Do you think you could set that up as a trial? I dont know whether the title is best, since people may forgt to include that in their title - is it possible to check for a specific tag?

jpfeuffer:
Would it suffice to build pyopenms on linux for every PR? We have much more resources there.

hroest:
> Would it suffice to build pyopenms on linux for every PR (with a label or a comment)? We have much more resources there.

sure, that would work - its just that the API stays in sync so any would work: OSX, Linux or Windows

also using the docker containers would work on Linux, whatever works better. I can help with docker stuff

I think eventually it would be nice to have nightly wheels from windows but I suggest to first do the Qt5 merge and then set it up (otherwise we would have to do some work twice)

PR #2794, opened 2017-07-25
Feature/fa idxml

timosachsenberg:
add mzIdentML functionality of reading/writing out fragment annotations to idXML


enetz:
The failing OpenPepXL tests are due to the new fragments annotations being written out in idXML,
so it should be fine to just replace the test files with the generated temps.

Do you have any tests to test the correct parsing of these annotations? For a manual test you could try read the OpenPepXL_input.mzML and the generated OpenPepXL_output.idXML.tmp from the topp tests in TOPPView. It should show the same annotations as after reading in the OpenPepXL_output.mzid.tmp.
But checking the FragmentAnnotation vectors after writing and reading a file in the idXMLFile class test would be better, like I did with mzid here:
https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/MzIdentMLFile_test.cpp#L445

timosachsenberg:
Thanks. Will add.

enetz:
The failing test says:
`Message: the file '/home/travis/build/OpenMS/OpenMS/src/tests/class_tests/openms/data/IdXML_XLMS_labelled.idXML' could not be found`
Do you have a file like that, or should I create one? 
Should be easy by calling `MzIdentMLFile().load(...)` and `IdXMLFile().store(...)` on the file `MzIdentML_XLMS_labelled.mzid`.

timosachsenberg:
thakns I added it

PR #2765, opened 2017-07-11
[FEATURE] Support for both channels and MS runs in cXML -> mzTab

jpfeuffer:
Please wait for merge. I have to test it here, since my system is messed up because of two QT versions.

PR #2796, opened 2017-07-25
Update pxds for recent changes to getPrimaryRunPath

jpfeuffer:
follow up for #2765 and pyopenms


jpfeuffer:
Well we seem to get gcc timeouts with the new trusty environment. Otherwise seems fine. won't be catched in CI anyway.

PR #2801, opened 2017-07-26
[FIX, PARAM] fixed comets and OMSSAs default tolerance.

timosachsenberg:
unified several precursor mass tolerance parameters of tools and fixed those that are pretty off

### DIFF
--- a/src/topp/MyriMatchAdapter.cpp
+++ b/src/topp/MyriMatchAdapter.cpp
@@ -223,9 +223,9 @@ protected:
     setValidFormats_("in", ListUtils::create<String>("mzML"));
     registerOutputFile_("out", "<file>", "", "Output file");
     setValidFormats_("out", ListUtils::create<String>("idXML"));
-    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 1.5, "Precursor mono mass tolerance.", false);
+    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 10.0, "Precursor mono mass tolerance.", false);
 
-    registerStringOption_("precursor_mass_tolerance_unit", "<unit>", "Da", "Unit to be used for precursor mass tolerance.", false);
+    registerStringOption_("precursor_mass_tolerance_unit", "<unit>", "ppm", "Unit to be used for precursor mass tolerance.", false);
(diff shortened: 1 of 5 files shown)


jpfeuffer:
I like the change! It will break some workflows though.

src/topp/OMSSAAdapter.cpp
```diff
@@ -201,8 +201,9 @@ class TOPPOMSSAAdapter :
     registerOutputFile_("out", "<file>", "", "Output file ");
     setValidFormats_("out", ListUtils::create<String>("idXML"));
 
-    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 1.5, "Precursor mass tolerance (Default: Dalton)", false);
-    registerFlag_("precursor_mass_tolerance_unit_ppm", "If this flag is set, ppm is used as precursor mass tolerance unit");
+    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 10.0, "Precursor mass tolerance (Default: Dalton)", false);
```
jpfeuffer: mono mass?

src/topp/MyriMatchAdapter.cpp
```diff
@@ -223,9 +223,9 @@ class MyriMatchAdapter :
     setValidFormats_("in", ListUtils::create<String>("mzML"));
     registerOutputFile_("out", "<file>", "", "Output file");
     setValidFormats_("out", ListUtils::create<String>("idXML"));
-    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 1.5, "Precursor mono mass tolerance.", false);
+    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 10.0, "Precursor mono mass tolerance.", false);
```
jpfeuffer: Unify expressions in all the parameters?

src/topp/OMSSAAdapter.cpp
timosachsenberg: ups default is now ppm. what do you mean with mono mass?

jpfeuffer: In all the other descriptions you specifically mention that it is the monoisotopic mass that is specified here.

timosachsenberg: ah I just saw it

timosachsenberg: Ok can do

PR #2802, opened 2017-07-27
[FIX] Channel reference in mass map for tmt10plex for channel 128N in…

lukaszimmermann:
I think there was a wrong entry in the channel map for tmt10 plex.

In the original, minus_1 in channel 128N referenced to 2, which is channel 127C. I think this should rather be 1, so referencing to channel 127N , which is then also consistent to the mass map [1]

[1] https://tools.thermofisher.com/content/sfs/COAPDFs/2015/SPECS_90111.pdf


jpfeuffer:
@oliveralka Does that impact current analyses?

hroest:
seems pretty clear to me especially as the comment does not reflect the code - I assume the comment is correct in this case. @cbielow any comment?

oliveralka:
yes, this is right - thanks Lukas.

@jpfeuffer It may be better to run the analysis again after merge - but I would like to proceed with the data analysis and we can give both version to the guys from MSstats - it may be interesting how much the datasets differ.

PR #2804, opened 2017-07-27
Feature/tmt11plex TMT11 plex support

lukaszimmermann:
Adds TMT 11plex support by splitting 131 reporter into 131N and 131C ions.


src/openms/source/ANALYSIS/QUANTITATION/TMTElevenPlexQuantitationMethod.cpp
```diff
@@ -0,0 +1,197 @@
...
+    //    "126", 126.127726, x, x, 127C, 128C
+    //    "127N", 127.124761, x, x, 128N, 129N
+    //    "127C", 127.131081, x, 126, 128C, 129C
+    //    "128N", 128.128116, x, 127N, 129N, 130N
+    //    "128C", 128.134436, 126, 127C, 129C, 130C
+    //    "129N", 129.131471, 127N, 128N, 130N, 131
+    //    "129C", 129.137790, 127C, 128C, 130C, x
+    //    "130N", 130.134825, 128N, 129N, 131, x
+    //    "130C", 130.141145, 128C, 129C, x, x
+    //    "131", 131.138180, 129N, 130N, x, x
```
hroest: can you please update the comment and split 131N and 131C?

hroest:
:+1: I like it, please update the comment

@timosachsenberg @jpfeuffer I dont really like how we need 300+ LOC to go from 10plex to 11plex ... there must be an easier/more generic way of doing this!

src/openms/source/ANALYSIS/QUANTITATION/IsobaricChannelExtractor.cpp
```diff
@@ -52,6 +53,7 @@ namespace OpenMS
 
   // Maximum allowed search window for TMT-10 reporter ions. The channels are only 0.006 Th apart.
   // Allowing anything larger will result in wrong quantifications for empty channels.
+  // Also used for TMT_11PLEX
   double TMT_10PLEX_CHANNEL_TOLERANCE = 0.003;
```
timosachsenberg: maybe rename to _10AND11PLEX_ ...

src/openms/source/ANALYSIS/QUANTITATION/IsobaricChannelExtractor.cpp
```diff
@@ -213,9 +215,10 @@ namespace OpenMS
     interpolate_precursor_purity_ = getParameters().getValue("purity_interpolation") == "true";
 
     /* check for sensible parameters */
-    if (dynamic_cast<const TMTTenPlexQuantitationMethod*>(quant_method_) != NULL && reporter_mass_shift_ > TMT_10PLEX_CHANNEL_TOLERANCE)
+    if (((dynamic_cast<const TMTTenPlexQuantitationMethod*>(quant_method_) != NULL) || (dynamic_cast<const TMTElevenPlexQuantitationMethod*>(quant_method_) != NULL))
```
timosachsenberg: if dynamic casts for type checks are needed it is usually a sign of bad design (at least if you have control over the all types as we do)

timosachsenberg: slightly better might be to check the name

lukaszimmermann:
There is a large deal of redundancy between the implementations of  `IsobaricQuantitationMethod`.
The header files are essentially identical and the cpp files only differ in some initialization and the values that the implemented members return. There is no real behavior in the classes.

@hroest  In Java, I'd probably go for an `enum` in this case. Since all of the implementations of
`IsobaricQuantitationMethod` are singletons (aren't they?), I'd probably go for something like this:

https://stackoverflow.com/questions/1965249/how-to-write-a-java-enum-like-class-with-multiple-data-fields-in-c#

using C++11 `constexpr`. 

Shall I make a proposal for refactoring?

hroest:
build pyopenms

hroest:
@jpfeuffer not sure whether its the PR or the test, but it seems the pyOpenMS check fails: 

```
  File "/home/jenkins/ws/openms_pr_pyopenms/fd57e9df/source/tools/PythonExtensionChecker.py", line 1163, in <module>
    main(options)
  File "/home/jenkins/ws/openms_pr_pyopenms/fd57e9df/source/tools/PythonExtensionChecker.py", line 1143, in main
    options.generate_pxd)
  File "/home/jenkins/ws/openms_pr_pyopenms/fd57e9df/source/tools/PythonExtensionChecker.py", line 973, in checkPythonPxdHeader
    "and that you specified the correct directory." )
Exception: No doxygen files found in directory:
```

seems more like an install issue than an issue with this PR, right?

jpfeuffer:
Yes. It doxygen is missing for the PyChecker. Will fix it.

jpfeuffer:
Yeah problem is we don't do our test suite when building pyopenms. The checker script has no Test.xml to append to.

jpfeuffer:
Can't we use the Build.xml which should always be there? We just need the header for CDash right? We could check for both.

jpfeuffer:
What is the status here? I restarted travis. It will hopefully also check pyopenms now.

lukaszimmermann:
I merged develop

hroest:
Shall I make a proposal for refactoring?

@lukaszimmermann  sure if you have the time to do it, me (and everybody else) would be in favor of making the code easier

PR #2805, opened 2017-07-27
[FIX] Also annotate first file when merging cXML

jpfeuffer:


jpfeuffer:
build pyopenms

jpfeuffer:
build pyopenms

jpfeuffer:
Pyopenms Trigger works, too.

timosachsenberg:
so the magic words to trigger are biudl ypopenms?

src/topp/FileMerger.cpp
```diff
@@ -245,8 +245,18 @@ class TOPPFileMerger :
     {
       ConsensusMap out;
       ConsensusXMLFile fh;
+      // load the metadata from the first file
       fh.load(file_list[0], out);
-      // skip first file
+      // but annotate the origins
+      if (annotate_file_origin)
+      {
+        for (ConsensusMap::iterator it = out.begin(); it != out.end(); ++it)
```
timosachsenberg: We could think about using for_each and a lambda in the future.
in my view its even easier to read

jpfeuffer: true.. was copy pasta

jpfeuffer:
Haha yes :) up for discussion. can be a regex, too. Have to fix the builds a little bit though. Does not run through yet.

hroest:
very nice, I like it a lot! Thanks @jpfeuffer seems like there is an issue with the extension checker, maybe a missing argument?

PR #2806, opened 2017-07-28
[FIX] TOPP-FileInfo

cbielow:
[FIX]: percentage of modified peptides for idXML input had too many digits (e.g. 45.434242342 %) --> now fixed to 1 digit, e.g 45.4 %

[FEATURE] TOPP-FileInfo: compute average length of peptides in idXML


cbielow:
2 timeouts -- otherwise fine

PR #2807, opened 2017-07-28
add/remove fragment annotations in TOPPView

timosachsenberg:
builds on #2794


timosachsenberg:
@enetz I renamed FragmentAnnotation to PeakAnnotation as it might be used in a larger context now. E.g. MS level 1

timosachsenberg:
current peptide_id_index and hit_index are now tracked in LayerData. This should make a lot of stuff easier.
PeakAnnotations added and removed are now synchronized with the PeakAnnotation of the current PeptideHit in the IdentificationView. This allows storing manual annotations back on disc.

src/openms/include/OpenMS/METADATA/PeptideHit.h
```diff
@@ -71,17 +71,17 @@ namespace OpenMS
       In the example "[alpha|ci$y3-H2O-NH3]" "alpha" or "beta" determines on which of the two peptides the fragmentation occured,
       "ci" or "xi" determines whether the cross-link and with it the other peptide is contained in the fragment,
       and the last part is the ion type with the fragmentation position (index) and losses.
-      The separators "|" and "$" are used to separate the parts easily when parsing the annotation.
+      The separators "|" and "$" are used to separate the parts easily when parsing the annotationa
```
enetz: spelling error

enetz:
In general this is a cool new functionality and works.

I am currently testing the annotations in TOPPView and preparing a fix for XL-MS data.
Current issues:

- Adding additional peak annotations that do not fit the XL-MS peak annotation schema to spectra annotated by XL-MS peak annotations causes a few problems. I am working on this.

- Does not work for mzIdentML files yet. That was made specifically for XL-MS data and the format is quite rigid compared to idXML. I am not sure if we can write out completely random strings as annotations.
I will look it up.

- Annotations are saved to PeptideHits only when a spectrum is closed, meaning when a different spectrum or ID is selected. Currently if you add annotations and immediately save the IDs to a file without selecting another spectrum first, the new annotations are not stored.
We need another trigger for updating the PeakAnnotations in the PeptideHit.
Maybe do it for the current spectrum when clicking on a save button?
Or every time an annotation is added? I don't think anyone would notice the slight delay after they click on OK after typing the annotation text.

- (Cosmetic, less important) The colour of a new annotation is blue, but the standard for annotations from a file is red. Currently if you add an annotation it is created as blue and when you change the ID or the spectrum and then back it is redrawn as red. I do not know what the expected behaviour would be. Would it be possible to set the initial colour to red as well to avoid this change of colour after changing the spectrum? Letting the user decide on a colour would mean we also have to save the colour of each peak to the file, so to avoid that we should agree on a general annotation colour (blue?) and use the annotation content to define colour schemes for special cases like green and red for b- and y-ions or linear and cross-linked ions.

timosachsenberg:
thanks @enetz for additions

PR #2808, opened 2017-07-29
Feature/sq mass open swath

hroest:
Use sqMass input files for OpenSWATH

- create new Swath-SqMass helper class
- create new spectrum access class for Swath-SqMass files
- rewrite most of SqMass access code for faster / parallel access 
- add capability to store multiple runs in single file
- add capability to store full mzML data structure in sqlite file for faithful conversion
- substantial speedup of OpenSWATH
  - moving PeakPicker / Smoothers into class members
  - improving efficiency of cross correlation calculation


timosachsenberg:
Sounds good. What are the prospects of supporting sqMass in parallel to mzML throughout OpenMS?

hroest:
since we have an abstraction layer to access spectra and a way to read them into a MSExperiment, I think it should not be too hard. But of course the individual developers need to implement that, but I think ATM the OpenSWATH code is the only one that struggles with large file sizes. So I dont think other areas need those kind of optimizations but I think some code that also uses profile data such as Smoothers, PeakPickers and FFMultiplex could benefit.

timosachsenberg:
do you want to have this integrated into the next OpenMS version? If so please sync.
@grosenberger can you review it?

grosenberger:
This will be a great addition and speed-up the whole workflow. How can we convert .RAW/.WIFF or mzML/mzXML files to sqMass? Are there plans to support conversion in OpenMS or ProteoWizard?

hroest:
this is close to completion, any more comments?

grosenberger:
Great, let's merge. Is there already a tool that can convert RAW/WIFF or mzML to sqMass?

hroest:
@grosenberger currently we dont have any direct converters to sqMass and we probably should continue using mzML as archival storage

PR #2811, opened 2017-07-30
[FEATURE] parallel python builds

hroest:
please only merge after https://github.com/uweschmitt/autowrap/pull/64


hroest:
see #2813

PR #2812, opened 2017-07-30
[FEATURE] FeatureLinkerUnlabeledKD separate tolerances

hannesveit:
Introduces separate RT and m/z tolerances for alignment and linking.


timosachsenberg:
Any chance that sensible parameter can somehow be automatically determined?

hannesveit:
I was considering it. I think it would require an additional full pass over the data. We could rebuild the alignment graph after RT transformation using the user-specified alignment tolerance and then estimate an appropriate linking tolerance based on the conflict-free connected components of sufficient size (i.e., potential consensus features that have sub-features from a sufficient number of maps and at most one sub-feature per map). For each of these, we could determine the maximum distance to the cluster center and use the maximum of all those maxima as the linking tolerance, maybe plus an additional safety margin. Would be nice to have. Any volunteers? :)

I think for now it's definitely an improvement to at least be able to specify separate tolerances.

timosachsenberg:
I think nightlies were unrelated

PR #2813, opened 2017-07-31
Feature/pyopenms split

hroest:
would allow to split up pyopenms into individual compilation units

- will decrease RAM usage of pyopenms compile drastically
- automatically splits up the CPP code in N different .cpp files which can be compiled individually
- does not help beyond a certain point as we still need to pull in all of the OpenMS headers and the STL / boost headers into the single CPP file
- currently does not take any dependency into account, basically splits the files up randomly. This is a bit of a pain as all classes still need to know about all other classes - in the future we may group them logically so that each module only needs to know about part of the library but this also needs to happen on the OpenMS side first...
- ~~depends on https://github.com/uweschmitt/autowrap/pull/65 - please *dont* merge yet~~
- uses autowrap 13.1, so this can be merged now
- not sure whether it will make travis builds viable but its a possibility...

basically, the resulting pyopenms folder looks like this

```
$ ls -ltrh pyOpenMS/build/lib.linux-x86_64-2.7/pyopenms/
total 67M
-rw-rw-r-- 1 hr hr 1.5K Jul 29 20:31 sysinfo.py
-rw-rw-r-- 1 hr hr  589 Jul 29 20:31 python_extras.py
-rw-rw-r-- 1 hr hr 2.4K Jul 29 20:31 License.txt
-rw-rw-r-- 1 hr hr 1.3K Jul 29 20:31 docompile.py
-rw-rw-r-- 1 hr hr 1.6K Jul 30 14:23 __init__.py
-rw-rw-r-- 1 hr hr   83 Jul 30 23:25 qt_version_info.py
-rwxrwxr-x 1 hr hr 563K Jul 30 23:25 libSuperHirn.so
-rwxrwxr-x 1 hr hr 113K Jul 30 23:25 libOpenSwathAlgo.so
-rwxrwxr-x 1 hr hr  46M Jul 30 23:25 libOpenMS.so
-rw-rw-r-- 1 hr hr  200 Jul 30 23:25 all_modules.py
-rwxrwxr-x 1 hr hr 2.2M Jul 30 23:26 pyopenms_1.so
-rwxrwxr-x 1 hr hr 2.6M Jul 30 23:28 pyopenms_2.so
-rwxrwxr-x 1 hr hr 2.0M Jul 30 23:29 pyopenms_3.so
-rwxrwxr-x 1 hr hr 3.1M Jul 30 23:30 pyopenms_4.so
-rwxrwxr-x 1 hr hr 2.4M Jul 30 23:31 pyopenms_5.so
-rwxrwxr-x 1 hr hr 2.3M Jul 30 23:32 pyopenms_6.so
-rwxrwxr-x 1 hr hr 2.2M Jul 30 23:33 pyopenms_7.so
-rwxrwxr-x 1 hr hr 4.2M Jul 30 23:35 pyopenms_8.so
```

with 

```
$ cat pyOpenMS/build/lib.linux-x86_64-2.7/pyopenms/all_modules.py 
from .pyopenms_1 import *
from .pyopenms_2 import *
from .pyopenms_3 import *
from .pyopenms_4 import *
from .pyopenms_5 import *
from .pyopenms_6 import *
from .pyopenms_7 import *
from .pyopenms_8 import *
```

which is auto-generated...while the most recent pyopenms had about 20 MB size and took 8+GB in ram, this compilation took around 2-3GB per module. We can adjust the number of modules we want with the cmake parameter `-DPY_NUM_MODULES`


hroest:
looks like it makes peak RAM usage a bit better and takes longer to compile:

```
924.34user 25.39system 17:40.75elapsed 89%CPU (0avgtext+0avgdata 3165684maxresident)k
719.63user 18.28system 12:16.93elapsed 100%CPU (0avgtext+0avgdata 7358716maxresident)k
```

hroest:
Now pyopenms is also integrated with travis...

hroest:
seems like its working ... do we want to merge @jpfeuffer @timosachsenberg ?

hroest:
- `49b` took 26 minutes

hroest:
@jpfeuffer @timosachsenberg @cbielow this adds another entry into our build matrix, so it would be good to get some feedback on this. Also it would be great to know whether it builds on Windows.

timosachsenberg:
cool. +1 from my side.

cmake/compiler_flags.cmake
```diff
@@ -135,6 +135,11 @@ elseif ("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
                   -Wno-switch-enum
                   -Wno-missing-prototypes
                   -Wno-missing-variable-declarations
+                  -Wno-deprecated
+                  -Wno-deprecated-register
+                  -Wno-covered-switch-default
+                  -Wno-date-time
+                  -Wno-missing-noreturn
```
hroest: any objections?

hroest:
now also works on Python3, also tested under Windows and it works

hroest:
if nobody objects, I suggest to merge...

timosachsenberg:
Fine with me. Are some of these warnings caused by autowrap? Or are these just the ones from our code

hroest:
> Fine with me. Are some of these warnings caused by autowrap? Or are these just the ones from our code

the removed warnings in the cmake file are due to our switch to C++11 (partially) and they only affect the building of libOpenMS. The ones in `setup.py` are also due to Cython - of course they overlap with ours since they include the same header files but then Cythons auto-generated code also introduces a few warnings on top of that. We should not worry about these but mainly about reducing those in the library build process.

I think we need to go through these warnings again and do some code cleanup. However, at the moment we have 1000+ warnings so its anyways not really manageable and reducing some of the less severe ones is a good place to start

Issue #2815, opened 2017-07-31
Travis timeouts


hroest:
We could address the travis timeouts by using the caching system smartly:

- they offer directory cache
- they offer `ccache`

https://docs.travis-ci.com/user/caching/

hroest:
currently with ccache I can get the gcc builds down to 10 minutes: https://travis-ci.org/OpenMS/OpenMS/builds/259547633

hroest:
@timosachsenberg @jpfeuffer we are now getting regular travis timeouts for clang builds, so there is a real question how useful these builds really are for us

jpfeuffer:
We have pyopenms builds from a docker file now. Maybe we can substitute one travis gcc run with a clang run. And maybe we can add a debug build in travis? or do you expect timeouts?
see #3246

hroest:
I looked at debug builds a few times but in my hands they create timeouts. You can try again, but already now we have trouble with the release builds if we have to start from scratch and cant use ccache

PR #2817, opened 2017-08-01
Feature/travis ccache

hroest:
addresses  #2815

- when no operation is performed, then the travis build time can be as low as 10-15 minutes


timosachsenberg:
cool but what about clang_

timosachsenberg:
maybe need to add CXX_FLAGS
https://github.com/travis-ci/travis-ci/issues/5383

hroest:
seems like logstream is failing, maybe the ccache does not properly distinguish between debug/release mode? 

```
100% tests passed, 0 tests failed out of 2021

Total Test time (real) = 592.17 sec

Submit files (using http)

   Using HTTP submit method
---------------------------
99% tests passed, 1 tests failed out of 2021

Total Test time (real) = 565.65 sec

The following tests FAILED:

	 15 - LogStream_test (SEGFAULT)

Submit files (using http)
```

timosachsenberg:
hmm it should download the correct ccache:
where is this failed test from?


0.02s
2.45 sattempting to download cache archive
fetching featuretravis_ccache/cache-linux-trusty-569ca2db85986fdbb3df6891f17dec9b55646a80e28ad31ebdded1f646402bd8--compiler-clang-debug.tgz
fetching featuretravis_ccache/cache--compiler-clang-debug.tgz
fetching develop/cache-linux-trusty-569ca2db85986fdbb3df6891f17dec9b55646a80e28ad31ebdded1f646402bd8--compiler-clang-debug.tgz
fetching develop/cache--compiler-clang-debug.tgz
could not download cache
0.03s
2.06sadding /home/travis/build/OpenMS/OpenMS/contrib to cache
creating directory /home/travis/build/OpenMS/OpenMS/contrib

hroest:
also it seems that clang-debug does not get accelerated at all?

hroest:
well we can just remove the different caches and see what happens ...

hroest:
also seems that it doesnt actually use clang...
```
--
-- The C compiler identification is GNU 4.8.5 -- The CXX compiler identification is GNU 4.8.5 -- Check for working C compiler: /usr/lib/ccache/gcc -- Check for working C compiler: /usr/lib/ccache/gcc -- works
```

timosachsenberg:
hmm it only got accelerated the second time it ran through... with could make sense afterall. First successful build warms up the ccache. Second benefits.

hroest:
hmm even on a hot cache it does not work well with clang: 

https://travis-ci.org/OpenMS/OpenMS/builds/259980848

it clearly copies the ccache data:

```
$ Installing caching utilities

0.01s

1.14sattempting to download cache archive

fetching featuretravis_ccache/cache-linux-trusty-ddf8f921679b8db6eed108970e2db61773203be704861cab9b1a2557fa2c9f7e--compiler-clang.tgz

found cache

cache.ccache

0.02s

1.84sadding /home/travis/.ccache to cache
```

hroest:
seems like ccache only really works for gcc but its better than nothing.

merge?

hroest:
ccache is a compiler cache that transparently sits on top of the compiler and first checks its cache before falling back on actually compiling. It has a direct and a preprocessor mode and there are a few gotchas since Clang support was only really added in version 3.2, and the current version is 3.2.3. It’s a project with a long history and it’s main focus is to be correct before being fast.
https://pspdfkit.com/blog/2015/ccache-for-fun-and-profit/

> ccache version 3.1.9

timosachsenberg:
are there cases when we need to force delete a cache? and if yes. how is this done?

hroest:
>are there cases when we need to force delete a cache?

I guess the cache can get corrupted (due to race condition during upload etc) and then nothing will compile? seems not like an unreasonable scenario given Murphys law

>  and if yes. how is this done?

you can always do it here: https://travis-ci.org/OpenMS/OpenMS/caches

timosachsenberg:
cool. let's try this and squash and merge

PR #2821, opened 2017-08-02
[FIX] pyOpenMS macOS fixes

grosenberger:
This PR intends to fix the current problems when building pyOpenMS on macOS. Attached is a file [pyopenms-2.2.0-cp27-cp27m-macosx_10_12_intel.whl.zip](https://github.com/OpenMS/OpenMS/files/1194121/pyopenms-2.2.0-cp27-cp27m-macosx_10_12_intel.whl.zip) for the standard Python 2.7.10 shipped with macOS 10.12 that was created using the patch and it would be great if people with the same or different systems could check whether it works for them.

### DIFF
--- a/src/pyOpenMS/MANIFEST.in
+++ b/src/pyOpenMS/MANIFEST.in
@@ -7,6 +7,7 @@ include pyopenms/libOpenSwathAlgo.*
 include pyopenms/libSuperHirn.*
 include pyopenms/QtCore*
 include pyopenms/QtNetwork*
+include pyopenms/libz.*
 exclude pyopenms/*.cpp
 global-exclude pyopenms.cpp
 global-exclude HMDB2StructMapping.tsv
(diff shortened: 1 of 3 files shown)


hroest:
what about libsqlite?

grosenberger:
Hmm, this does not seem to be a dependency of pyopenms.so, libOpenMS.dylib or the others, right? Is it still a requirement? I see that it is contained in the Windows builds though.

hroest:
maybe its statically linked?

grosenberger:
It seems that they are statically linked for the macOS builds.

jpfeuffer:
Installation and simple loading of idXML works on my macOS11.12

jpfeuffer:
Btw: Where can I find building instructions for the wheel? I can try to add them to the nightlies then. Maybe after QT5 but sometime soon.

hroest:
> Btw: Where can I find building instructions for the wheel? I can try to add them to the nightlies then. Maybe after QT5 but sometime soon.

@jpfeuffer the wheel is automatically built with `make pyopenms`, see https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/CMakeLists.txt#L433 which calls `setup.py bdist_wheel`

jpfeuffer:
@grosenberger Worked for me. Good to merge?

grosenberger:
Great, thanks for testing! We will also need to port the changes back to the master branch so we can provide the official 2.2 builds on PyPI.

grosenberger:
Can we merge these changes before the release? Then we can provide pyOpenMS builds for 2.3 on PyPI again.

hroest:
:+1: for merge into master

@grosenberger can you rebase and resolve conflicts and then create a PR against master?

PR #2822, opened 2017-08-02
Using a constant MetaValue name for precomputed precursor errors in TOPPView

enetz:
Added a constant string for a MetaValue or user parameter name for precursor error to the `Constants` class.
TOPPView reads out this parameter from an `idXML` or `mzIdentML` file and shows this value if it is available.
Otherwise the ppm error is computed from the `AASequence` of the `PeptideHit` and the spectrum precurser mass (which should work for most applications).
The MetaValue is meant to be used by searches, where not all the necessary masses are included in the `AASequence` of one `PeptideHit`, e.g. RNPxl and OpenPepXL.

RNPxl and OpenPepXL precompute the precursor error in ppm and now write it out under the same userParam.


timosachsenberg:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=233433

enetz:
Ups, new values are being written to the files. Have to update the test files.

enetz:
The last failing test is due to the order of annotations on my computer vs Travis for `[M+H]-H2O++ U-HPO3` and `[M+H]++ U-H3PO4`, which have the same mass.
@timosachsenberg should this still happen after using stable_sort in spectrum sorting? Is there another place where we need a stable_sort?

jpfeuffer:
But it seems to be already in the nightlies #2828

timosachsenberg:
Will take a look at this. Might be in the postannotation step.

hroest:
@timosachsenberg @enetz seems like this breaks the nightlies.

timosachsenberg:
should be fixed with https://github.com/OpenMS/OpenMS/pull/2824

PR #2823, opened 2017-08-02
[FEATURE] add option to search also isotopic peak before the annotated

timosachsenberg:
Compensates for monoisotopic peak misassignments

### DIFF
--- a/src/utils/RNPxlSearch.cpp
+++ b/src/utils/RNPxlSearch.cpp
@@ -1999,7 +1999,7 @@ protected:
         double precursor_mz = precursor[0].getMZ();
 
         // map (corrected) precursor mass to spectra
-        for (auto i : precursor_isotopes)
+        for (int i : precursor_isotopes)
         {
           double precursor_mass = (double) precursor_charge * precursor_mz - (double) precursor_charge * Constants::PROTON_MASS_U;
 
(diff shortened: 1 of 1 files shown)


src/utils/RNPxlSearch.cpp
```diff
@@ -1999,7 +1999,7 @@ class RNPxlSearch :
         double precursor_mz = precursor[0].getMZ();
 
         // map (corrected) precursor mass to spectra
-        for (auto i : precursor_isotopes)
+        for (int i : precursor_isotopes)
```
hroest: so why does this fail? is this C++14? I thought that should work in C++11 ?

timosachsenberg: I changed it to make it easier readable. Auto is great for long typenames but not really needed for ints

enetz:
Does the precursor error calculation consider this? And should this correction be noted somewhere in the output? May be nice to know for some post-processing analyses.

timosachsenberg:
Good point. Currently it behaves like when other search engines do the same. E.g. Msgf will also have a large precursor error.

timosachsenberg:
We can merge this and think about representing isotopic misassignments later. That discussion will probably affect more tools

src/utils/RNPxlSearch.cpp
hroest: ah ok I thought it caused travis to fail...my bad

PR #2824, opened 2017-08-02
[FEATURE] compensate for mono-isotopic peak misassignment

timosachsenberg:


timosachsenberg:
same as in RNPxlSearch

enetz:
Nice addition, but is it a good idea to set {-1, 0} as the standard? I would consider it unexpected behaviour for a normal peptide search. This also breaks the precursor error calculation in TOPPView, which only considers the sequence mass and the MS2 precursor mass. This compensation should be documented in the output, in a way that allows for correct precursor mass error calculations.

timosachsenberg:
-1 0 is standard for MSGFPlus so I guess its probably empirically backed up

timosachsenberg:
for comet the parameter file has even: -1/0/1/2/3 , which I think is -3 -2 -1 0 1 in our notation

timosachsenberg:
Any comments about the notation. It differs from msgfplus and comet.

enetz:
If most other tools use the inverse notation, we also should. Less confusion with the adapters etc.
And the other notation would have mostly positive numbers and be easier to read in my opinion.
The interpretation of the misassignment error would also be obvious from the parameter itself (e.g. a hit with +2 would mean the 3rd peak in an isotopic pattern was misassigned as the mono-isotopic peak)

timosachsenberg:
Ok will update

timosachsenberg:
changed notation.
should also fix https://github.com/OpenMS/OpenMS/issues/2828

hroest:
doesnt fix pyOpenMS yet

PR #2829, opened 2017-08-04
unittest for  BilinearInterpolation python wrapper

zaidurrehman:
I have added unit tests for the python wrapper for BilinearInterpolation class. However, at the moment 3 out of 16 implemented tests fail.

Here are the details of the failed tests:

1. test_empty()
```
>>> import pyopenms
>>> bilip = pyopenms.BilinearInterpolation()
>>> print bilip.empty()
True
>>> bilip.getData().resize(1, 2, float())
>>> print bilip.empty()
True
```
The second `print` statement should return `False` as in the [cpp test](https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/BilinearInterpolation_test.cpp#L472). In fact, all calls to `empty()` return `True`.

2. test_supportMax_0()
```
>>> bilip = pyopenms.BilinearInterpolation()
>>> bilip.setMapping_0(3.0, 1.0, 2.0)
>>> bilip.setMapping_1(5.0, 3.0, 4.0)
>>> bilip.getData().resize(2, 3, float())

>>> bilip.supportMin_0()
-1.0
>>> bilip.supportMax_0()
-1.0
>>> bilip.supportMin_1()
-11.0
>>> bilip.supportMax_1()
-11.0
```
The values returned should be -4, 5, -16 and 4 respectively, as in [cpp test](https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/BilinearInterpolation_test.cpp#L327). However, `index2key` methods return correct values.

3.test_ supportMax_1()
```
>>> bilip = pyopenms.BilinearInterpolation()
>>> bilip.setMapping_1(3.0, 1.0, 2.0)
>>> bilip.setMapping_0(5.0, 3.0, 4.0)
>>> bilip.getData().resize(3, 2, float())

>>> bilip.supportMin_1()
-1.0
>>> bilip.supportMax_1()
-1.0
>>> bilip.supportMin_0()
-11.0
>>> bilip.supportMax_0()
-11.0
```

The values returned should have been -4, 5, 16 and 4 respectively, as in [cpp test](https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/BilinearInterpolation_test.cpp#L450)


hroest:
please replace all `bilip.getData().xxx` with 
```
tmp = bilip.getData()
// manipulate tmp
bilip.setData(tmp)
```
then it should work. The reason is that `bilip.getData()` produces a temporary object and if you manipulate that you do not change the state of `bilip`

hroest:
apart from that I like it, thanks a lot for your contribution

zaidurrehman:
- all `bilip.getData().xxx` replaced with the suggestion above.
- additional tests for `getData()`, `setData()`, `addValue()` and `value()`
- all tests pass

lars20070:
@hroest Can we merge? Failing tests unrelated.

PR #2793, opened 2017-07-25
[FIX] fix for boost 1.64

hroest:
- fixes #2790

PR #2834, opened 2017-08-07
[FIX] Amend to boost164 fix

jpfeuffer:
see #2793


timosachsenberg:
@oliveralka might be related to your problem

jpfeuffer:
Only RNPXL test is failing. Please merge before nightlies if you agree that it will be fine.

PR #2840, opened 2017-08-08
[NOP] prevent some copies by making stuff const

timosachsenberg:
small speedup of hyperscore (10%)


cbielow:
+1 if tests pass.

If you want this even faster, one could replace the findNearest() (which I guess is expensive) within the for-loop with something smarter, e.g. use a while-loop which starts at the last match and stops when the distance is minimized. This way, you traverse the spectrum at most once during the whole for-loop (assuming both experimental and theoretical are sorted).

PR #2841, opened 2017-08-08
[GUI, FIX] more robust PeakAnnotations in TOPPView

enetz:
PeakAnnotations in PeptideHits are updated on pressing the save IDs button or when changing the selected spectrum or ID.
Charges are now only contained in the charge field and should not be part
of the annotation string.


timosachsenberg:
+1

PR #2842, opened 2017-08-09
[FEATURE] implementation of Morpheus PSM score

timosachsenberg:
@enetz implemented the morpheus scoring. Did not work well when compared to HyperScore but I only used b and y ions. Original implementation also generates various losses so it might perform better with larger number of peaks in the theoretical spectrum.


src/openms/source/ANALYSIS/RNPXL/MorpheusScore.cpp
```diff
@@ -0,0 +1,109 @@
...
+      }
+      else if (d > 0) // theo peak is right of exp. peak (outside of tolerance window)
+      {
+        ++t;
+      }
+    }
+
+    while (e < n_e) { total_intensity += exp_spectrum[e].getIntensity(); ++e; }
+
+    // similar to above but we now make sure that every experimental peak intensity is summed up
```
enetz: This comment should be the other way around, right? You summed up the total intensity above and only sum the intensity of the matching peaks here.

enetz:
This is nice, but I think there should be a way to do this in one while loop.
Is there a reason you can not just put line 91 `match_intensity += exp_spectrum[e].getIntensity();;`  directly under line 62 `++matches;` and remove the whole second loop?

src/openms/source/ANALYSIS/RNPXL/MorpheusScore.cpp
timosachsenberg: ah I see this is kind of misleading. what I wanted to stress is that every **matched** experimental peak is summed up. The code above doesn't care about that but only considers every **matched** theoretical peak

src/openms/source/ANALYSIS/RNPXL/MorpheusScore.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    while (t < n_t && e < n_e)
+    {
+      const double theo_mz = theo_spectrum[t].getMZ();
+      const double exp_mz = exp_spectrum[e].getMZ();
+      const double d = exp_mz - theo_mz;
+      const double max_dist_dalton = fragment_mass_tolerance_unit_ppm ? theo_mz * fragment_mass_tolerance * 1e-6 : fragment_mass_tolerance;
+      if (abs(d) <= max_dist_dalton) // match in tolerance window? 
+      {
+        match_intensity += exp_spectrum[e].getIntensity();;
```
enetz: two `;`s

timosachsenberg:
This is nice, but I think there should be a way to do this in one while loop.
Is there a reason you can not just put line 91 match_intensity += exp_spectrum[e].getIntensity();; directly under line 62 ++matches; and remove the whole second loop?

I don't think so without extensive (and probably slower) logic. If you look at how indices are incremented you see the difference. Above, it makes sure that every theoretical peak is matched at most once. Below it makes sure that every experimental peak that matches one or more theoretical peaks is summed up exactly once.

enetz:
Yes, now I see it. You increment a different index after each `if (abs(d) <= max_dist_dalton)` line.

timosachsenberg:
if you find time. Can you also check that I did not make an obvious error in translating the code from https://github.com/cwenger/Morpheus/blob/master/Morpheus/Morpheus/PeptideSpectrumMatch.cs#L61
It should be pretty straight forward but its better to have another set of eyes looking at it.

src/openms/source/ANALYSIS/RNPXL/MorpheusScore.cpp
```diff
@@ -0,0 +1,110 @@
...
+// $Authors: Timo Sachsenberg $
+// --------------------------------------------------------------------------
+
+#include <OpenMS/ANALYSIS/RNPXL/MorpheusScore.h>
+#include <OpenMS/KERNEL/MSSpectrum.h>
+#include <cmath>
+
+namespace OpenMS
+{
+  std::tuple<double, Size, Size, Size, Size> MorpheusScore::compute(double fragment_mass_tolerance,
```
timosachsenberg: probably better to return a struct than an (unnamed) tuple. I will probably change this

timosachsenberg:
I added tests. unrelated failed...

PR #2843, opened 2017-08-09
Adds the util MSstatsConverter, which converts consensusXML and experimental design to input for the R package MSstats

lukaszimmermann:
The following strategy is currently used to determine the required values for the MSstats input CSV:

- **ProteinName**  Accession from the  Peptide Evidence of the Peptide Hit
- **PeptideSequence** Unmodified sequence from the peptide hit
-  **PrecursorCharge** Charge of PeptideHit
-  **FragmentIon** This is tried to be parsed from the annotation string of the PeakAnnotations
-  **ProductCharge** Also taken from the PeakAnnotation if available
-  **IsotopeLabelType** Has to be set as util parameter
- **Condition, BioReplicate, Run**  Specified via an additional CSV file with four columns: `FileName, Condition BioReplicate, Run`. 
- **Intensity** Intensity of the consensus feature

The input files must fulfill the following criteria:
- Each filename that appears in the consensusXML also must appear in the experimental design
- The experimental design CSV is only allowed to contain the four aforementioned columns. The column names must match, but the order of the columns is irrelevant. Missing or additional columns are not permitted. 
- The consensusXML is currently assumed to be ID-mapped.

Most interestingly, the `FileName` on the experimental design input is matched to the PeptideIdentification via the `spectra_data` meta value of the ProteinIdentification. I do not know whether there is a more sensible way of doing that association.

To do:
- Unassigned Peptides are currently ignored. These should also be added (or perhaps optionally be added)
- Tests, especially with PeakAnnotations.

All comments and suggestions welcome.


src/utils/MSstatsConverter.cpp
```diff
@@ -0,0 +1,419 @@
...
+    {
+      String const & identifier = protein_identification.getIdentifier();
+
+      if (protein_identification.metaValueExists(TOPPMSstatsConverter::meta_value_exp_design_key) == false)
+      {
+        LOG_ERROR << "ERROR: ProteinIdentification does not have meta value for original file. Cannot continue!" << std::endl;
+        return ILLEGAL_PARAMETERS;
+      }
+
+      for (auto const & meta_value : protein_identification.getMetaValue(TOPPMSstatsConverter::meta_value_exp_design_key).toStringList())
```
timosachsenberg: this line is a bit long and it's, thus, significantly harder to deduce the type of auto

src/utils/MSstatsConverter.cpp
```diff
@@ -0,0 +1,419 @@
...
+      {
+        // Split the identifier by path separator character
+        std::vector< std::string > strs;
+        boost::split(strs, meta_value, boost::is_any_of("/\\"));
```
timosachsenberg: maybe check the File and related classes in OpenMS. I think we might have more robust ways to get the filename

src/utils/MSstatsConverter.cpp
```diff
@@ -0,0 +1,419 @@
...
+            LOG_ERROR << "ERROR: Experimental design does not contain information on file " << filename << ". Cannot continue!" << std::endl;
+            return ILLEGAL_PARAMETERS;
+          }
+          std::set< Size > const & indices = edkey_to_rowindex[filename];
+          runs.insert(indices.begin(), indices.end());
+        }
+
+        for (auto const & pep_hit : pep_id.getHits())
+        {
+          std::vector< PeptideHit::PeakAnnotation > const & original_fragment_annotations = pep_hit.getPeakAnnotations();
```
timosachsenberg: we use "const TPYENAME &" all over OpenMS so while it is not wrong it might be more consistent

timosachsenberg: Also, the examples in https://github.com/OpenMS/OpenMS/wiki/Coding-conventions follow this rule.

lukaszimmermann:
This has just been tested with LFQ. I'd be happy if you could provide the data for iTraq, then this PR could be tested and updated for this kind of data, and I could also write some meaningful tests.

If I got that right, unassigned peptides do not have quantitative values associated with it, so I'd also not include them for MSstats. Perhaps we could issue or warning or something.

Yeah, the options you mentioned are important. I think I can make an updated PR tomorrow with the changes.

lars20070:
@lukaszimmermann [Here](http://www.ebi.ac.uk/pride/archive/projects/PXD001265) some iTRAQ data which were re-analysed with [this workflow](https://github.com/OpenMS/OpenMS/wiki/Workflow-Collection#thermo-orbitrap-velos). For details see chapter 12.3 in

Nilse, L. (2016). OpenMS: A Modular, Open-Source Workflow System for the Analysis of Quantitative Proteomics Data. In C. Bessant (Ed.), Proteome Informatics (pp. 259–288). Cambridge: Royal Society of Chemistry. http://doi.org/10.1039/9781782626732-00259

lukaszimmermann:
Thanks, I am going to take a look.

lukaszimmermann:
Some documentation and a sketch of the algorithm currently being used for writing the input for MSstats:

https://docs.google.com/document/d/1Iq63x7X_5t-dMCXdiX2vh4oeK6MttGfkgQtTj8f5dWk/edit?usp=sharing

lukaszimmermann:
The MSstatsConverter is now consistent with the OpenMS Experimental Design TSV. The Column Names in MSstats are treated the following way:

- **Run**: For the MSstats input, this is the same as the `Run` column in the OpenMS TSV.  Matching is done by `Spectra File`.
- **Fraction**: This is the same as the `Fraction` column in the OpenMS TSV. Again, matching is done by `Spectra File`.
- **BioReplicate**: The user can choose, which column from the Condition Table (the second one in the OpenMS design) should be used here. This defaults to `Biological Replicate`
- **Condition**: The user can choose, which column from the Condition Table should be used here. There is no default. It is very important to note that a `Condition` in MSstats input is a different thing than a `Condition` in the OpenMS TSV. That is why the user can choose the column from the condition table that should be used as MSstats `Condition`. Most likely the user does not want the `Condition` column in MSstats to be the same as the OpenMS `Condition`.

There is no support for iTRAQ, as MSstats does not seem to support it in the current version, quote README:
>  MSstats does not support for metabolic labeling or iTRAQ experiments

Thus, any Quantification column in the OpenMS TSV is ignored. Consequently, the file names in the `Spectra File` column in the Run Table (the first one) must be unique (which is the case for a label-free specification).

The experimental design is given in two TSV file parameters (because we currently have two tables for the experimental design spec). It feels a bit awkward to have two tables in one TSV file. But I can change that if requested.

hroest:
there are also some comments from cpplint: http://cdash.openms.de/testDetails.php?test=45135029&build=243934

lars20070:
Thank you @lukaszimmermann. We will test next week.

@kernursula We were looking for consensus features with multiple (conflicting) IDs. Guess `R` and `MSstats` will make it easy to find them.

lukaszimmermann:
I will double check the code of this again. Shall we include this with a kind of 'untested' disclaimer for the 2.3 release?

lukaszimmermann:
Currently I test the `MSstatsConverter` on iPRG-2015 data. Will update once the tests have been completed.

timosachsenberg:
Thanks. I think, given that this might become a central tool in OpenMS, it's probably too early to include it in 2.3. I would be happy to merge it to develop once the tests are there and we have a good understanding how robust the adapter works.

timosachsenberg:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=250929

also

_ as member prefix is discouraged by our coding conventions

timosachsenberg:
http://cdash.openms.de/testDetails.php?test=46435928&build=251382

lukaszimmermann:
I made a workflow to demonstrate how the converter works:
https://www.dropbox.com/s/1zhacb2dr6ndbiw/msstats_lfq_and_peptide_identification.knwf?dl=0

Essentially,  this is LFQ on the iPRG2015 benchmark set with a workflow embedding of the functionality of the MSstatsConverter appended.  The input of the converter is highlighted with a workflow annotation, so it currently takes consensusXML and two experimental design files. The last CSV output produced in the workflow is identical to the output of the MSstatsConverter (this is what I tested). The output is available here:
https://www.dropbox.com/s/o5p720qo8yp08v6/result_converter2_sorted.csv?dl=0

I consider this workflow as kind of a baseline what the MSstatsConverter is supposed to do. We can find more applications (e.g. with fractions or protein inference) and I can extend the workflow accordingly.

lukaszimmermann:
I skyped with Meena and discussed the algorithm with her. The thing missing is the handling of multiple rows in the resulting CSV file (due to different retention times). I will also make one more example workflow handling fractions.

lukaszimmermann:
MSstatsConverter does not allow the same feature to appear multiple times with different intensities (and different retention times). Thus, the intensity for one feature must be calculated from a set of intensities, where each element is associated with a different
retention time.

MSstatsConverter now offers this as another parameter `retention_time_resolution_method`, which has the following options:
- **min** (The minimal intensity over all retention times of the feature is picked)
- **max** (The maximum intensity over all retention times of the feature is picked)
- **mean** (The mean intensity over all retention times of the feature is picked)
- **manual** (The feature is repeated for each retention time value in the output CSV and
 the output CSV file will contain the retention time as an additional column.) This output is
then invalid for MSstats and features have to be combined over their retention times as a
consecutive step in the workflow)

tjeerdijk:
HI Lukas,

Would you have time to discuss your work on the MSstatsConverter soon. Always welcome at the Mii!

Take care from Tjeerd

lukaszimmermann:
In this branch, Experimental Design now also supports the Sample table. Run table and Sample table are separated by one line only consisting of '$'.

timosachsenberg:
nice but maybe '$' is not that common.

can you replace that with a check for an empty line?

hroest:
can you merge with develop and look into the failed tests?

src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
```diff
@@ -53,51 +61,393 @@ namespace OpenMS
   class OPENMS_DLLAPI ExperimentalDesign
   {
   public:
+    ExperimentalDesign() {};
```
hroest: can you move all the implementations from the `.h` file to the `.cpp` file?

src/tests/topp/CMakeLists.txt
```diff
@@ -1919,13 +1919,6 @@ set_tests_properties("TOPP_ProteinQuantifier_8_out1" PROPERTIES DEPENDS "TOPP_Pr
...
-# mzTab output:
-#add_test("TOPP_ProteinQuantifier_10" ${TOPP_BIN_PATH}/ProteinQuantifier -test -in ${DATA_DIR_TOPP}/ProteinQuantifier_input.consensusXML -mzTab_out ProteinQuantifier_10_output.tmp -top 0 -average sum)
-#add_test("TOPP_ProteinQuantifier_10_out1" ${DIFF} -in1 ProteinQuantifier_10_output.tmp -in2 ${DATA_DIR_TOPP}/ProteinQuantifier_10_output.mzTab)
-#set_tests_properties("TOPP_ProteinQuantifier_10_out1" PROPERTIES DEPENDS "TOPP_ProteinQuantifier_10")
-#add_test("TOPP_ProteinQuantifier_11" ${TOPP_BIN_PATH}/ProteinQuantifier -test -in ${DATA_DIR_TOPP}/ProteinQuantifier_input.featureXML -mzTab_out ProteinQuantifier_11_output.tmp -top 0 -average sum -filter_charge)
-#add_test("TOPP_ProteinQuantifier_11_out1" ${DIFF} -in1 ProteinQuantifier_11_output.tmp -in2 ${DATA_DIR_TOPP}/ProteinQuantifier_11_output.mzTab)
-#set_tests_properties("TOPP_ProteinQuantifier_11_out1" PROPERTIES DEPENDS "TOPP_ProteinQuantifier_11")
 #ratios
 add_test("TOPP_ProteinQuantifier_12" ${TOPP_BIN_PATH}/ProteinQuantifier -test -in ${DATA_DIR_TOPP}/ProteinQuantifier_input.consensusXML -out ProteinQuantifier_12_output.tmp -top 3 -include_all -consensus:fix_peptides -average sum -ratios)
```
hroest: we probably need tests of the new TOPP tools

src/utils/ProteomicLFQ.cpp
```diff
@@ -0,0 +1,323 @@
...
+        // TODO: overwrite primaryMSRun with picked mzML name (for submission)
+        // mzML_file.store(OUTPUTFILENAME, ms_centroided);
+        // TODO: free all MS2 spectra (to release memory!)
+
+        //-------------------------------------------------------------
+        // Feature detection
+        //-------------------------------------------------------------
+        // TODO: call FeatureFinderIdentification (and think about external ids ;)
+        // TODO: free memory of centroided PeakMap
+        // TODO: free parts of feature map not needed for further processing (e.g., subfeatures...)
```
hroest: seems like there is a lot of TODO left here

src/utils/ProteomicLFQ.cpp
```diff
@@ -0,0 +1,323 @@
...
+          return INCOMPATIBLE_INPUT_DATA;
+        }
+
+        // check if spectra are sorted
+        for (Size i = 0; i < ms_raw.size(); ++i)
+        {
+          if (!ms_raw[i].isSorted())
+          {
+            ms_raw[i].sortByPosition();
+            writeLog_("Info: Sorte peaks by m/z.");
```
cbielow: typo: sorte

src/utils/ProteomicLFQ.cpp
```diff
@@ -0,0 +1,323 @@
...
+      aligner.align(feature_maps, transformations, reference_index);
+
+      // find model parameters:
+      Param model_params = aligner.getDefaults().copy("model:", true);
+      String model_type = model_params.getValue("type");
+
+      if (model_type != "none")
+      {
+        model_params = model_params.copy(model_type + ":", true);
+        for (TransformationDescription t : transformations)
```
cbielow: how does this work?
`t` is copied, i.e. all computations on it are lost

lukaszimmermann:
The point here is that this feature branch was merged with:
https://github.com/OpenMS/OpenMS/tree/feature/proteomic_lfq

to acquire the ExperimentalDesign implementation which is required for MSstats. This branch also extends the interface of ExperimentalDesign.
I can not comment on ProteomicLFQ stuff. So I am not sure how this PR can be resolved. To resolve conflicts with develop, I could merge again with proteomic_lfq, but this PR would need to wait until proteomic_lfq is merged.

Other solution would be to split this PR into two:
* One with cherry-picked commits for Experimental Design
* One with cherry-picked commits for MSstats.

Then the MSstats PR would need to wait for the Experimental Design PR.

@timosachsenberg what do you think?

timosachsenberg:
I would cherry pick the changes for the experimental design and MSstats. Probably not that difficult if done one a file basis

hroest:
I agree, please split up this PR into manageable chunks with the experimental design separate from MSStats and separate from ProteomeLFQ

lukaszimmermann:
Experimental Design stuff moved to #3332. I guess this PR can be closed, will open new one
with MSstats extension based on #3332.

PR #2844, opened 2017-08-10
[FIX] also sort PD frag. annotations

timosachsenberg:


timosachsenberg:
tests pass, should fix the RNPxl test

PR #2846, opened 2017-08-11
[REFACTOR] Use boost::regex instead of regex in SiriusMzTabWriter

lukaszimmermann:
<regex> should perhaps be avoided since it does not work with GCC < 4.9.


oliveralka:
Looks good. 
Regex is working now as expected on linux, gcc 4.8.5.

oliveralka:
Random build failure or problem?
https://travis-ci.org/OpenMS/OpenMS/jobs/263432664
24 - pyopenms_test_memoryleaktests (Failed)

jpfeuffer:
I guess unrelated. But not 100% sure.

lukaszimmermann:
Could be related to this issue:

https://stackoverflow.com/questions/20133961/memory-leak-in-boostregex-library

jpfeuffer:
If that's true, I guess our options are:
- Ignore
- Try to whitelist this during the check
- Switch to the std:: one and require recent compilers
- Write our own

I would go for the second if that is possible. What do you think @hroest ?

cbielow:
do you really need a regex for this?
`std::string scan_index = path.substr(path.find_last_not_of("0123456789"));`
(untested)

lukaszimmermann:
Very likely. Thanks! Gonna change that.

lukaszimmermann:
Ok, works well.

PR #2847, opened 2017-08-11
[Feature] get data array by name

enetz:
New getters for meta data arrays in MSSpectrum.

I am not sure, what to return when no array with the given name exists,
for now an empty DataArray of the respective type is returned.


timosachsenberg:
can you also add the const versions?

timosachsenberg:
and pyOpenMS pxd file :)

Issue #2828, opened 2017-08-03
RNPXL Tests failing in nightlies on macOS


jpfeuffer:
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_nightly_packaging/compiler=appleclang-7.3.0,os_label=elcapitan/lastCompletedBuild/testReport/

jpfeuffer:
Because of fragment annotation changes.

jpfeuffer:
Could be set ordering issues.

jpfeuffer:
Still not fixed on Linux gcc4.8
nightly branch, sha 52ce45bd334514bb17144032b0b5009a6cd5d148

jpfeuffer:
Labelled it blocker because I want to enforce test error free nightly packages before the release.

timosachsenberg:
passes on my gcc 4.8.5

timosachsenberg:
when I cklick on https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_nightly_packaging/compiler=appleclang-7.3.0,os_label=elcapitan/lastCompletedBuild/testReport/ I don't see it failing. Am I looking in the wrong spot?

jpfeuffer:
Yes now it only fails on linux. https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_nightly_packaging/compiler=g++-4.8,os_label=ubuntu_1404/lastCompletedBuild/testReport/

PR #2849, opened 2017-08-11
[FIX] fix sorting of ion annotations (need probably also sort unshifted

timosachsenberg:
ions)

lets see if this is enough


jpfeuffer:
Yes. solves #2828

PR #2854, opened 2017-08-14
Fix long filename issue in TOPPAS

cbielow:
[FIX] augment #2582 which introduced a long-filename problem, by adding a new suffix in every TOPPAS TOPP node resulting in chained suffices, e.g. mzml.idxml.tsv....
This commit removes old file extensions before adding new ones, thus keeping file length small.

Also shortens the path further by shorter unique names and not using the redundant ./toppas_tmp/ subdir anymore.

Successfully tested with all TOPPAS example pipelines and 'LabelFree_Fusion_20170802.toppas' from @lars20070.

### DIFF
--- a/src/openms/include/OpenMS/SYSTEM/File.h
+++ b/src/openms/include/OpenMS/SYSTEM/File.h
@@ -146,8 +146,13 @@ public:
     */
     static String findDoc(const String& filename);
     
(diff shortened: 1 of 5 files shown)


cbielow:
tests look good. 1 timeout.

lars20070:
Good solution. Thanks @cbielow.

PR #2855, opened 2017-08-14
[FIX] fix return a reference to local variable.

timosachsenberg:
this fixes the build issue so I would like to merge this quickly. @enetz it currently doesn't check for the presence of the name and if it is missing we probably get a segfault. Best way would probably be to return the iterator but then the function is not easily wrapped with pyOpenMS.

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/MSSpectrum.h
+++ b/src/openms/include/OpenMS/KERNEL/MSSpectrum.h
@@ -366,99 +366,45 @@ public:
     }
 
     /// Returns a mutable reference to the first integer meta data array with the given name
-    /// if none with the given name exists, an empty array is returned
     inline IntegerDataArray& getIntegerDataArrayByName(String name)
     {
-      Size n_arrays = integer_data_arrays_.size();
-      for (Size i = 0; i < n_arrays; i++)
-      {
-        if (integer_data_arrays_[i].getName() == name)
-        {
-          return integer_data_arrays_[i];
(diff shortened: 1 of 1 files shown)


cbielow:
mhh.. both solutions are illegal C++ if the name is not found :/ (segfault as you mentioned)

Why not use the iterator and provide another (less efficient) interface for pyOpenMS? Or would Raw pointers be any better?
Alternatively, throw an exception if the array is not found -- but I guess sometimes its expected that it does not exist, so exception is not warranted.

timosachsenberg:
@hroest is there any way we could use the iterator solution? would this mean that autowrap needs to be extended or only some addon file for pyOpenMS

timosachsenberg:
another solution would be to return a reference to a static instance of an empty vector

cbielow:
since the return value is not const, how does the user tell if he got the real array or the 'does not exist' dummy version? People might even start to fill in data... this gets messy quickly...

timosachsenberg:
Yep. These are no real solutions. Iterator would be best or we need a has/get combination

hroest:
there are ways to get an iterator, but its quite ugly and also there is a chance that it will point to an invalid `end()` element which cannot be tested in Python. See https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/pxds/MSChromatogram.pxd#L46 for examples, so `libcpp_vector[Klass].iterator` works as return value

Issue #2791, opened 2017-07-24
Test and AddressSanitizer error in SiriusAdapter


jpfeuffer:
```
Progress of 'loading spectra list':
-- done [took 0.02 s (CPU), 0.03 s (Wall)] -- 
=================================================================
==24149==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000b5020 at pc 0x40d68a bp 0x7ffe75ad7c00 sp 0x7ffe75ad7bf0
READ of size 8 at 0x6020000b5020 thread T0
    #0 0x40d689 in OpenMS::DPosition<1u, double>::operator[](unsigned long) const /home/jenkins/ws/openms_nightly_supplement/fd57e9df/source/src/openms/include/OpenMS/DATASTRUCTURES/DPosition.h:140
    #1 0x408162 in OpenMS::Peak1D::getMZ() const /home/jenkins/ws/openms_nightly_supplement/fd57e9df/source/src/openms/include/OpenMS/KERNEL/Peak1D.h:116
    #2 0x7f425fd5ce5a in OpenMS::SiriusMSFile::store(OpenMS::MSExperiment const&, OpenMS::String&) /home/jenkins/ws/openms_nightly_supplement/fd57e9df/source/src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp:237
    #3 0x40bed0 in TOPPSiriusAdapter::main_(int, char const**) /home/jenkins/ws/openms_nightly_supplement/fd57e9df/source/src/utils/SiriusAdapter.cpp:220
    #4 0x7f4260277b50 in OpenMS::TOPPBase::main(int, char const**) /home/jenkins/ws/openms_nightly_supplement/fd57e9df/source/src/openms/source/APPLICATIONS/TOPPBase.cpp:491
    #5 0x40764a in main /home/jenkins/ws/openms_nightly_supplement/fd57e9df/source/src/utils/SiriusAdapter.cpp:314
    #6 0x7f425d57d82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #7 0x4074a8 in _start (/home/jenkins/ws/openms_nightly_supplement/fd57e9df/build/bin/SiriusAdapter+0x4074a8)

0x6020000b5020 is located 0 bytes to the right of 16-byte region [0x6020000b5010,0x6020000b5020)
allocated by thread T0 here:
[...]

PR #2856, opened 2017-08-14
Fix/siriusadapternightlies Tries to solve the error in SiriusAdapter test

lukaszimmermann:
I tested on El Capitan with the same compiler (Apple clang++ 7.3.0) as on abibuilder.
The files in the tests are identical, so I could not reproduce the error. Also `FuzzyDIff` passes. I tested the branches `develop` and `nightly`. Currently I lack ideas on how to deal further with the issue. Here, I just regenerated the test file on a Linux machine. Perhaps it helps.

Does anyone have another idea?


jpfeuffer:
Doesn't the failing test suggest that there is either an additional newline after line 4 or the rest of the MTD lines missing completely?

jpfeuffer:
Have you seen, that the file is basically completely empty?

https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_nightly_packaging/compiler=appleclang-7.3.0,os_label=elcapitan/ws/build/src/tests/topp/SiriusAdapter_1_output.tmp/*view*/

lukaszimmermann:
No, haven't seen the file.  But this is helpful, because the description should definitely not be null.

jpfeuffer:
Test output:
```
2017-08-15 16:22:14.890 SiriusAdapter[47248:48952591] CFPasteboardRef CFPasteboardCreate(CFAllocatorRef, CFStringRef) : failed to create global data
2017-08-15 16:22:15.167 SiriusAdapter[47248:48952591] CFPasteboardRef CFPasteboardCreate(CFAllocatorRef, CFStringRef) : failed to create global data
Progress of 'loading spectra list':
No MS1 spectrum for this precursor. Occurred 0 times.
0 spectra were skipped due to precursor charge below -1 and above +1.
Charge of 0 was set to +1 due to positive polarity 5 times.
Charge of 0 was set to -1 due to negative polarity 0 times.
Fatal error: Running SiriusAdapter returned an error code or could no compute the input
Deleting temporary directory '/tmp/2017-08-15_162215_abibuilder1.Informatik.Uni-Tuebingen.De_47248_1'. Set debug level to 2 or higher to keep it.
Deleting temporary msfile '/tmp/2017-08-15_162215_abibuilder1.Informatik.Uni-Tuebingen.De_47248_2.ms'. Set debug level to 2 or higher to keep it.
SiriusAdapter took 0.16 s (wall), 0.14 s (CPU), 0.01 s (system), 0.13 s (user).
```

jpfeuffer:
I guess they are doing some weird stuff in their shell scripts.

jpfeuffer:
No sorry, the CFPasteboardRef error come from running it as admin.

lukaszimmermann:
I think that the sirius executable misbehaves for some reason on the build machine. I have access, so I a going to directly take a look.

oliveralka:
May be related to the regex error on linux? There an empty file may be created - was the file totally empty or did it have the "MTD" if it?

lukaszimmermann:
The output says:
Fatal error: Running SiriusAdapter returned an error code or could no compute the input

So I think that sirius cannot be executed on the El Capitan machine (but I do not know yet, why)

jpfeuffer:
Yes, we should start by changing the error code.
And by the way: The test also fails on linux.

jpfeuffer:
Why is the following in the startup script?
CLASSPATH="$APP_HOME/../../../All/Sirius/lib/*:${GUROBI_HOME}/lib/gurobi.jar"

lukaszimmermann:
Ok this explains it. Thanks @jpfeuffer  I am going to prepare a fix.

jpfeuffer:
Should be just ../lib. We only install/clone the binaries specific for a certain platform.

jpfeuffer:
Thanks! and dont forget linux. Windows seems to be ok.

jpfeuffer:
see #2791

lukaszimmermann:
PR with fixed classpath open.

jpfeuffer:
https://github.com/OpenMS/THIRDPARTY/pull/16

jpfeuffer:
Still crashes on Linux. I can have a look why but we should definitely think about grabbing the stderr of the child process if possible.

lukaszimmermann:
Yes, we could do that. I won't have time for this issue before Monday, though.

jpfeuffer:
The issue on Linux should be solved with https://github.com/OpenMS/THIRDPARTY/pull/18.
Still would be great if you could improve the error output and fix the executable search in this PR.

lukaszimmermann:
Yes will do that. Thanks for your work.

lukaszimmermann:
Made the requested changes. 

I think that the `-fingerid` parameter of `SiriusAdapter` is redundant. Can't we just call CSIFingerID if the user has provided the `-out_CSIFingerID` parameter and leave the call out otherwise? 
In fact, it does not really make sense to specify `-fingerid` without `-out_CSIFingerID` and vice versa.

@oliveralka  what do you think?

lukaszimmermann:
pyopenms test fail.

timosachsenberg:
ok this PR is finished. consider removing the parameter -fingerid in other PR

PR #2857, opened 2017-08-15
[FIX] BSA example files converted to more recent mzML format 

cbielow:
(using our own FileConverter)

MSGF+ will read these happily now (before it complained that spectrum reference is not set)


jpfeuffer:
thx

PR #2858, opened 2017-08-16
fix mzTab2tsv_PRT script

lars20070:
Fixes R script for `mzTab` post-processing of the `PRT` section.


timosachsenberg:
can you give a visual output how this looks like

lars20070:
@timosachsenberg Is `opt_global_protein_group_type` in `mzTab` somewhere documented?
I have got Fido output without any `opt_global_protein_group_type = protein_group`.

timosachsenberg:
it is not documented because it is an optional column.

lars20070:
Do you have an example `mzTab` with `single_protein`, `protein_group` and `indistinguishable_group`? Then I could figure out how groups are reported.

jpfeuffer:
I think the standard behaviour of Fido is to just group indistinguishable groups. Connected components (bigger groups) are currently only annotated if you do the group resolution with the flag.

PR #2859, opened 2017-08-16
[FIX] Make PercolatorAdapter an official tool

jpfeuffer:
It is registered as a TOPP tool in the ToolHandler.
Probably the reason for the Adapter to be wrongly categorized in KNIME.


jpfeuffer:
Only timeout. Any objections?

cbielow:
there seem to be a few TODO's in the code, but if it works well in practice, ok.

One thing we definitely need before promoting to TOPP is:
http://ftp.mi.fu-berlin.de/pub/OpenMS/develop-documentation/html/TOPP_documentation.html
It needs to show up there.

And some TOPP tests (if percolator.exe is found) would be nice.

jpfeuffer:
Hmm okay I see. I agree. It was mostly about consistency here. I think it is listed as a TOPP tool in the ToolHandler and that might be why the category is messed up.
@mwalzer @Leon-Bichmann Is there a chance that you can add percolator specific tests, too?

timosachsenberg:
I checked. It is correctly listed as TOPP tool in the tool handler.

timosachsenberg:
test files are missing but stub code is there

PR #2861, opened 2017-08-17
[FEATURE] faster idXML store; memory consumption report

cbielow:
this PR introduces of some of new features which are required for the new Aho-Corasick PeptideIndexer.
Its a smaller, more digestable chunk, to make the review a bit easier. Another PR is coming up soon.

The changes introduced here are not dramatic:
[FEATURE] 10% faster writing of idXML 
[FEATURE] idXML can now report timings when loading/storing (via ProgressLogger)
[NOOP] minor formatting of Progresslogger output
[FEATURE] move semantics for FASTAEntries (potentially faster) when loading FASTA files
[FEATURE] TOPP tools report their peak memory usage when using -debug

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/FASTAFile.h
+++ b/src/openms/include/OpenMS/FORMAT/FASTAFile.h
@@ -41,6 +41,7 @@
 #include <functional>
 #include <fstream>
 #include <memory>
+#include <utility>
 #include <vector>
 
 namespace OpenMS
@@ -77,9 +78,9 @@ public:
       String sequence;
 
       FASTAEntry() :
-        identifier(""),
-        description(""),
-        sequence("")
+        identifier(),
+        description(),
+        sequence()
       {
       }
 
@@ -89,6 +90,31 @@ public:
         sequence(seq)
       {
       }
+      
+      FASTAEntry(const FASTAEntry& rhs)
+        :
+        identifier(rhs.identifier),
+        description(rhs.description),
+        sequence(rhs.sequence)
+      {
+      }
+
+      FASTAEntry(FASTAEntry&& rhs) noexcept
+       :
+        identifier(::std::move(rhs.identifier)),
+        description(::std::move(rhs.description)),
+        sequence(::std::move(rhs.sequence)) 
+      {
+      }
+
+      FASTAEntry& operator=(const FASTAEntry& rhs)
+      {
+        if (*this == rhs) return *this;
+        identifier = rhs.identifier;
+        description = rhs.description;
+        sequence = rhs.sequence;
+        return *this;
+      }
 
       bool operator==(const FASTAEntry& rhs) const
       {
@@ -97,19 +123,19 @@ public:
                && sequence == rhs.sequence;
       }
     
-      bool headerMatches(const FASTAEntry rhs)
+      bool headerMatches(const FASTAEntry& rhs) const
       {
         return identifier == rhs.identifier && 
   	     description == rhs.description;
       }
  
-      bool sequenceMatches(const FASTAEntry rhs)
+      bool sequenceMatches(const FASTAEntry& rhs) const
       {
         return sequence == rhs.sequence;
       }
     };
 
-    /// Copy constructor
+    /// Default constructor
     FASTAFile();
 
     /// Destructor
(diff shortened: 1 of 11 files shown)


src/openms/include/OpenMS/FORMAT/FASTAFile.h
```diff
@@ -77,9 +78,9 @@ namespace OpenMS
       String sequence;
 
       FASTAEntry() :
```
timosachsenberg: is FASTAEntry a simple POD? Then we can get rid of most code for construction, destruction and move.
https://stackoverflow.com/questions/146452/what-are-pod-types-in-c?noredirect=1&lq=1
https://stackoverflow.com/questions/12979529/default-constructors-and-pod

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -132,7 +132,7 @@ namespace OpenMS
     f.readStart(filename);
     while (f.readNext(p))
     {
-      data.push_back(p);
+      data.push_back(std::move(p));
```
timosachsenberg: personal interst: any (e.g., compiler) reason why emplace_back() is not used?

src/topp/FileInfo.cpp
```diff
@@ -427,34 +397,27 @@ class TOPPFileInfo : public TOPPBase
...
 
       os << "Number of sequences: " << entries.size() << "\n"
          << "\n";
 
-      for (loopiter = entries.begin(); loopiter != entries.end(); loopiter = std::next(loopiter))
+      for (auto loopiter = entries.begin(); loopiter != entries.end(); ++loopiter)
       {
-        iter = find_if(entries.begin(), loopiter, bind1st(mem_fun(&FASTAFile::FASTAEntry::headerMatches), &(*loopiter)));
-
+        auto iter = find_if(entries.begin(), loopiter, [&loopiter](auto& entry) { return entry.headerMatches(*loopiter); });
```
timosachsenberg: is auto& entry in the lambda c14?

src/topp/FileInfo.cpp
```diff
@@ -427,34 +397,27 @@ class TOPPFileInfo : public TOPPBase
...
-
-      mu.after();
-      std::cout << "\n\nMemory usage while loading: " << mu << std::endl;
+      std::cout << "\n\n" << mu.delta("loading FASTA") << std::endl;
 
       os << "Number of sequences: " << entries.size() << "\n"
          << "\n";
 
-      for (loopiter = entries.begin(); loopiter != entries.end(); loopiter = std::next(loopiter))
+      for (auto loopiter = entries.begin(); loopiter != entries.end(); ++loopiter)
```
timosachsenberg: for (const auto& entriy : entries) ? (I did not see *loopiter beeing changed in the small code snipped I see...)

src/topp/FileInfo.cpp
```diff
@@ -471,7 +434,7 @@ class TOPPFileInfo : public TOPPBase
       os << "Amino acid counts: "
          << "\n";
 
-      for (it = aacids.begin(); it != aacids.end(); it = std::next(it))
+      for (auto it = aacids.begin(); it != aacids.end(); ++it)
```
timosachsenberg: for (const auto& aa : aacids) looks imho nicer

timosachsenberg:
looks good. most comments are just me trying to learn c11 :P

src/openms/source/FORMAT/FASTAFile.cpp
cbielow: no difference IMHO, since we're moving already constructed objects (as opposed to forwarding c'tor arguments)

src/openms/include/OpenMS/FORMAT/FASTAFile.h
cbielow: not a POD, since it holds string (which is not a POD)

src/topp/FileInfo.cpp
cbielow: ups..

cbielow:
looks good. CI errors are related to SiriusAdapter and old Python errors.

jpfeuffer:
This might have broken nightly Mac builds.

jpfeuffer:
https://abibuilder.informatik.uni-tuebingen.de/jenkins/view/OpenMS/job/openms_nightly_packaging/compiler=appleclang-7.3.0,os_label=elcapitan/80/artifact/build/General_Testing/Testing/Temporary/LastBuild_20170818-2300.log/*view*/

cbielow:
`std::fabs(long long)` is valid. So it might be defined in another header.
should be `<cstdlib>` or `<cmath>`. Before we start guessing, can someone please try?!

PR #2862, opened 2017-08-17
Extension of trafoXML to store annotations on data points

hendrikweisser:
This turns `TransformationModel::DataPoint` from a `pair<double, double>` to a `struct {double, double, String}`, and adds a corresponding optional attribute to `<Pair>` in the trafoXML file format.

The current use is to annotate data points with the peptide sequences that generated them in MapAlignerIdentification.

@hrost, @grosenberger: pyOpenMS isn't updated yet. On the C++ side, the `DataPoint` is originally defined in `TransformationModel`. However, in pyOpenMS, it's only referenced in `TransformationDescription`. I think I'll have to add a `cdef cppclass` definition for `DataPoint` - where should it go?


hendrikweisser:
Travis CI gives one failed build (clang, not sure why) and failures of "TOPP_SiriusAdapter_1", which seem unrelated to my changes.

oliveralka:
Fail seems unrelated and "TOPP_SiriusAdapter_1"-fails should hopefully be fixed with the recent merge of lukas in "OpenMS/THIRDPARTY".

timosachsenberg:
Hi @hendrikweisser can we include this in the release as-is? or should we rather wait after the release?

hendrikweisser:
@timosachsenberg: Which release? Wasn't there just a (source) release?

pyOpenMS still isn't updated here (haven't heard from Hannes or George), so maybe that would be a reason to wait? Other than that, this shouldn't break anything.

timosachsenberg:
We would like to create a full release by the end of September. So we need to finish PRs that should be integrated ASAP.

hroest:
probably the best place to look would be `src/pyOpenMS/pxds/MassTrace.pxd` and write something like

```
cdef extern from "<OpenMS/ANALYSIS/MAPMATCHING/TransformationModel.h>" namespace "OpenMS":
  cdef cppclass TM_DataPoint "OpenMS::TransformationModel":
    TM_DataPoint()
    TM_DataPoint(TM_DataPoint())
    ...
```

in order to not pollute the common namespace. you can put it either in src/pyOpenMS/pxds/TransformationModel.pxd or in a new file src/pyOpenMS/pxds/TM_DataPoint.pxd

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/TransformationDescription.h
```diff
@@ -105,6 +105,13 @@ namespace OpenMS
...
 
+    /**
+      @brief Sets the data points (backwards-compatible overload)
+
+      Removes the model that was previously fitted to the data (if any).
+    */
+    void setDataPoints(const std::vector<std::pair<double, double> >& data);
+
     /// Returns the data points
     const DataPoints& getDataPoints() const;
```
hroest: i guess `getDataPoints` in pyOpenMS will break b/c of this

hendrikweisser:
I've updated the pyOpenMS pxds and got pyOpenMS to compile on my system.

hroest:
@hendrikweisser did you also adjust the unit tests? it seems that it compiles but the unit tests dont all pass: https://travis-ci.org/OpenMS/OpenMS/jobs/270400990 (unfortunately cdash is down so we cant see the actual error messages but I guess you need to update the unit tests too)

hendrikweisser:
@hroest: Thanks for the reminder. Should be done now.

PR #2863, opened 2017-08-18
[FIX] fix

hroest:


hroest:
fixes pyOpenMS compiles and test itself

PR #2865, opened 2017-08-22
[FEATURE] Improved OpenSWATH SQLite export

grosenberger:
This PR extends the OpenSWATH OSW SQLite format by adding support to store multiple runs in the same database. The tables are linked via unique identifiers. Unique identifiers are internally of type OpenMS::UInt64, however, they are stored as TEXT in the database to prevent problems for the downstream tools, because SQLite does not support UInt64. If anybody has a better idea how to solve this issue, I would be happy to implement the suggestions.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
@@ -110,7 +110,7 @@ namespace OpenMS
       const char * create_sql =
         "CREATE TABLE FEATURE(" \
         "ID TEXT PRIMARY KEY NOT NULL," \
-        "PRECURSOR_ID TEXT NOT NULL," \
+        "PRECURSOR_ID INT NOT NULL," \
         "RT REAL NOT NULL," \
         "DELTA_RT REAL NOT NULL," \
         "LEFT_WIDTH REAL NOT NULL," \
@@ -161,7 +161,7 @@ namespace OpenMS
 
         "CREATE TABLE FEATURE_TRANSITION(" \
         "FEATURE_ID TEXT NOT NULL," \
-        "TRANSITION_ID TEXT NOT NULL," \
+        "TRANSITION_ID INT NOT NULL," \
         "AREA_INTENSITY REAL NOT NULL," \
         "APEX_INTENSITY REAL NOT NULL," \
         "VAR_LOG_INTENSITY REAL NULL," \
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
@@ -194,8 +194,8 @@ namespace OpenMS
 
       // Insert run_id information
       std::stringstream sql_run;
-      sql_run << "INSERT INTO RUN (ID, FILENAME) VALUES ("
-              << (String)run_id_ << ", '"
+      sql_run << "INSERT INTO RUN (ID, FILENAME) VALUES ('"
+              << run_id_ << "', '"
               << input_filename_ << "'); ";
 
       // Execute SQL insert statement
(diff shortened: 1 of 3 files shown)


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
```diff
@@ -181,6 +191,22 @@ namespace OpenMS
...
+              << run_id_ << "', '"
+              << input_filename_ << "'); ";
+
+      // Execute SQL insert statement
+      rc = sqlite3_exec(db, sql_run.str().c_str(), callback, 0, &zErrMsg);
+      if( rc != SQLITE_OK )
+      {
+        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
+            zErrMsg);
+        sqlite3_free(zErrMsg);
```
cbielow: this free() is never executed --> memory leak
you'd have to rescue the content of zErrMsg to a local string (since it gets unwinded), then call free() and throw the exception using the local string

timosachsenberg:
Do you want this PR integrated into the next OpenMS release? Then please sync. We probably also need another reviewer @hroest would you volunteer?

hroest:
are you sure that sqlite does not support 64bit integers? This seems to indicate otherwise: 

http://www.sqlite.org/datatype3.html

`INTEGER. The value is a signed integer, stored in 1, 2, 3, 4, 6, or 8 bytes depending on the magnitude of the value.`

however, a UInt64 would overflow this (potentially). But instead of storing it as text (maybe very slow) why not ensure that the value is smaller than the max value (e.g. 2**63-1, see also http://jakegoulding.com/blog/2011/02/06/sqlite-64-bit-integers/ ) as the value itself does not matter much, it matters more that it is unique. So we can just right-shift the value and cast to a signed 64bit integer, right? We would just decrease the potential space of unique values by a factor of 2 which should not matter much.

timosachsenberg:
probably a bit level conversion via *(int*)&x would also work

hroest:
> probably a bit level conversion via *(int*)&xwould also work if we restrict ourself to 64bit

hmm not sure what that would do in case of overflow, maybe produce a negative id? well that would not be too bad either, because who cares whether the id is positive or negative as long as we have a large space?

timosachsenberg:
We should be able to use int64_t. It is now available in c11.
http://en.cppreference.com/w/cpp/types/integer

grosenberger:
Ok, I implemented the proposed conversion to int64_t. This seems to work and will produce negative IDs for large numbers.

src/utils/OpenSwathWorkflow.cpp
```diff
@@ -499,7 +499,7 @@ class TOPPOpenSwathWorkflow
       feature_finder_param.setValue("TransitionGroupPicker:PeakPickerMRM:method", "corrected");
       feature_finder_param.setValue("TransitionGroupPicker:PeakPickerMRM:signal_to_noise", 0.1);
       feature_finder_param.setValue("TransitionGroupPicker:PeakPickerMRM:gauss_width", 30.0);
-      feature_finder_param.setValue("uis_threshold_sn",-1);
+      feature_finder_param.setValue("uis_threshold_sn",0);
```
grosenberger: This updated parameter will reduce the output file size substantially by omitting transitions below the threshold.

timosachsenberg:
looks good from my side though a comment why the conversion is done would be nice. @hroest anything from your side?

grosenberger:
I added comments to make this more clear.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
```diff
@@ -108,10 +112,16 @@ namespace OpenMS
 
       // Create SQL structure
       const char * create_sql =
+        "CREATE TABLE RUN(" \
+        "ID INT PRIMARY KEY NOT NULL," \
+        "FILENAME TEXT NOT NULL); " \
```
hroest: maybe we could have another field here, like a comment but (i) I am not sure we need it and (ii) we can always add it later

PR #2870, opened 2017-08-27
Feature/peak annotation

dmccloskey:
[FEATURE] calculate of peak shape metrics for downstream QC/QA


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -292,25 +292,8 @@ namespace OpenMS
...
-            if (std::fabs(it->getMZ() - peak_apex) <= peak_apex_dist)
-            {
-              peak_apex_int = p[1];
-              peak_apex_dist = std::fabs(it->getMZ() - peak_apex);
-            }
-            rt_sum += it->getMZ();
-            intensity_sum += it->getIntensity();
-          }
-        }
+        calculatePeakApexInt_(used_chromatogram,best_left,best_right,hull_points,intensity_sum,rt_sum,peak_apex_int,peak_apex);
```
timosachsenberg: minor stuff: Please add space after comma

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -368,6 +351,61 @@ namespace OpenMS
...
           total_peak_apices += peak_apex_int;
         }
+    
+        if (compute_peak_shape_metrics_)
+        { //for backwards compatibility with TOPP tests
+          // Calculate peak shape metrics that will be used for later QC
+          double width_at_5 = 0;
+          double width_at_10 = 0;
+          double width_at_50 = 0;
+          // double area_2_height = 0;
```
timosachsenberg: please remove unused comments

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -368,6 +351,61 @@ namespace OpenMS
...
+          double end_time_at_5 = 0;
+          double total_width = 0;
+          double tailing_factor = 0;
+          double asymmetry_factor = 0;
+          double baseline_delta_2_height = 0;
+          double slope_of_baseline = 0;
+          int points_across_baseline = 0;
+          int points_across_half_height = 0;
+          
+          calculatePeakShapeMetrics_(used_chromatogram,
```
timosachsenberg: the parameter list is rather long.
consider adding a (class internal) struct
ChromatogramPeakShapeMetrics that contains these as member

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
+    double startIntensity, endIntensity;
+    double intensity, intensity_prev, retention_time, retention_time_prev;
+    double start_time_at_50, end_time_at_50;
+    double base, height, height_5, height_10, height_50;
+    
+    // pass 1
+    for (MSChromatogram::const_iterator it = chromatogram.begin(); it != chromatogram.end(); it++)
+    {
+      MSChromatogram::const_iterator it_prev = it;
+      it_prev--; //previous point
```
timosachsenberg: This will not work for the first point

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
+      intensity = std::max(it->getIntensity()-avg_noise_level,0.0); //background-subtracted intensity
+      intensity_prev = std::max(it_prev->getIntensity()-avg_noise_level,0.0); //background-subtracted intensity of the previous point
+      retention_time = it->getMZ();
```
timosachsenberg: getRT()? also below

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
+      retention_time_prev = it_prev->getMZ();
+
+      // start and end intensities
+      if (retention_time_prev == best_left)
+      {
+        startIntensity = intensity;
```
timosachsenberg: minor style issue: use start_intensity instead (also other variables below)

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
+          {
+            base = retention_time - retention_time_prev;
+            height = intensity_prev - intensity;
+            height_50 = 0.5*peak_height - intensity;
+            end_time_at_50 = retention_time - height*base/height_50;
+          }
+        } 
+
+        // points across the peak
+        points_across_baseline += 1;
```
timosachsenberg: shorter: ++points_across_baseline

timosachsenberg: also below

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
+    ConvexHull2D::PointArrayType & hull_points,
+    double & intensity_sum, 
+    double & rt_sum,
+    double & peak_apex_int,
+    double & peak_apex)
+  {  
+    intensity_sum = 0.0;
+    rt_sum = 0.0;
+    peak_apex_int = -1;
+    double peak_apex_dist = std::fabs(chromatogram.begin()->getMZ() - peak_apex);
```
timosachsenberg: getRT() instead of getMZ() would be clearer.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
     }
   }
+  
+  void MRMTransitionGroupPicker::calculatePeakApexInt_(const MSChromatogram& chromatogram,
+    double best_left, double best_right, 
+    ConvexHull2D::PointArrayType & hull_points,
+    double & intensity_sum, 
+    double & rt_sum,
+    double & peak_apex_int,
+    double & peak_apex)
```
timosachsenberg: maybe call this argument peak_apex_rt instead of peak_apex.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
+        endIntensity = intensity_prev;
+      }
+
+      if (retention_time > best_left && retention_time < best_right)
+      {
+        //start and end retention times
+        if (retention_time < peak_apex)
+        {
+          // start_time_at_5
+          if (intensity >= 0.05*peak_height && \
```
timosachsenberg: \ is not wrong but not required (and not really used in the code base)

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,171 @@ namespace OpenMS
...
+            intensity_prev < 0.05*peak_height && \
+            points_across_baseline > 1)
+          {
+            base = retention_time - retention_time_prev;
+            height = intensity - intensity_prev;
```
timosachsenberg: when I first read this it was not clear what height is. I think calling it delta_int and delta_rt would be clearer? what do you think?

src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
```diff
@@ -454,6 +454,211 @@ START_SECTION((template < typename SpectrumT > void remove_overlapping_features(
...
+  double & end_time_at_5,
+  double & total_width,
+  double & tailing_factor,
+  double & asymmetry_factor,
+  double & baseline_delta_2_height,
+  double & slope_of_baseline,
+  int & points_across_baseline,
+  int & points_across_half_height) ))
+{
+  // Toy chromatogram
```
timosachsenberg: can you give some information on this data? is it from real data or generated in R or python?
It is very important to know for tests if this is really ground truth data or only to detect regressions.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: I don't think there is a getRT operator.  At least from what I can see from the documentation.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: I was thinking in terms of a triangle for some reason when I decided on those names.  I agree, not the most clear.  Your suggestion is much better.

src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
dmccloskey: I added some information: "data is taken from raw LC-MS/MS data points acquired for L-Glutamate in RBCs".  Let me know if more information is needed.

dmccloskey:
I incorporated the changes you had left in the comments.  Let me know how it looks.

dmccloskey:
Let me know if there is an ETA for when this can be merged.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -292,25 +292,8 @@ namespace OpenMS
...
-            {
-              peak_apex_int = p[1];
-              peak_apex_dist = std::fabs(it->getMZ() - peak_apex);
-            }
-            rt_sum += it->getMZ();
-            intensity_sum += it->getIntensity();
-          }
-        }
+        calculatePeakApexInt_(used_chromatogram, best_left, best_right, hull_points, intensity_sum, rt_sum, peak_apex_int, peak_apex);
+        double peak_apex_int_copy = peak_apex_int; //copy of the max peak intensity before background subtraction
```
grosenberger: Can this line be deleted?

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,158 @@ namespace OpenMS
...
+  
+  void MRMTransitionGroupPicker::calculatePeakShapeMetrics_(const MSChromatogram& chromatogram, 
+    double best_left, double best_right, 
+    double peak_height, double peak_apex_rt, double avg_noise_level,
+    PeakShapeMetrics_ & peakShapeMetrics)
+  {
+    peakShapeMetrics.points_across_baseline = 0;
+    double start_intensity, end_intensity;
+    double intensity, intensity_prev, retention_time, retention_time_prev;
+    double start_time_at_50, end_time_at_50;
```
grosenberger: Are these variables used downstream somewhere or can they be removed?

hroest:
generally looks good but please fix these issues:

https://www.codacy.com/app/hroest/OpenMS/pullRequest?prid=869979

dmccloskey:
Fixed the issues.  The "intensity" issue looks like that was a false positive.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
timosachsenberg: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/KERNEL/ChromatogramPeak.h#L111

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,157 @@ namespace OpenMS
...
+        p[0] = it->getMZ();
+        p[1] = it->getIntensity();
+        hull_points.push_back(p);
+        if (std::fabs(it->getMZ() - peak_apex_rt) <= peak_apex_dist)
+        {
+          peak_apex_int = p[1];
+          peak_apex_dist = std::fabs(it->getMZ() - peak_apex_rt);
+        }
+        rt_sum += it->getMZ();
+        intensity_sum += it->getIntensity();
```
hroest: @dmccloskey just to highlight again, this is how I "integrate" the peak: there is no integration, its a simple sum. Note that therefore this only works if we use the same sampling speed across multiple runs, otherwise we run into problems here. The correct way to do this would be to indeed perform a trapezoidal peak integration here.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,157 @@ namespace OpenMS
...
+    {
+      MSChromatogram::const_iterator it_prev = it;
+      --it_prev; //previous point
+      intensity = std::max(it->getIntensity()-avg_noise_level, 0.0); //background-subtracted intensity
+      intensity_prev = std::max(it_prev->getIntensity()-avg_noise_level, 0.0); //background-subtracted intensity of the previous point
+      retention_time = it->getMZ();
+      retention_time_prev = it_prev->getMZ();
+
+      // start and end intensities
+      if (retention_time_prev == best_left)
```
hroest: I am a bit uncomfortable with comparing floats, this has a tendency to go wrong ...

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,157 @@ namespace OpenMS
...
+      {
+        start_intensity = intensity;
+      }
+      else if (retention_time == best_right)
```
hroest: same here

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,157 @@ namespace OpenMS
...
+      {
+        end_intensity = intensity_prev;
+      }
+
+      if (retention_time > best_left && retention_time < best_right)
```
hroest: better use `>=` and `<=`

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +161,157 @@ namespace OpenMS
...
+    double start_intensity, end_intensity;
+    double intensity, intensity_prev, retention_time, retention_time_prev;
+    double delta_rt, delta_int, height_5, height_10, height_50;
+    
+    // pass 1
+    for (MSChromatogram::const_iterator it = chromatogram.begin() + 1; it != chromatogram.end(); ++it)
+    {
+      MSChromatogram::const_iterator it_prev = it;
+      --it_prev; //previous point
+      intensity = std::max(it->getIntensity()-avg_noise_level, 0.0); //background-subtracted intensity
```
hroest: `double intensity`

hroest:
looks good apart from the few points mentioned above. The major point is that I feel the code may break for `if (retention_time_prev == best_left)` as in any transition *group* the left/right start point may not actually be a valid RT in any chromatogram.

> Fixed the issues. The "intensity" issue looks like that was a false positive.

no the tool is actually correct

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: I see.  Ok this would explain the discrepency that I was finding between the areas reported.  I was using peak height, so it was not a big deal, but using peak area is much more common than peak height.

dmccloskey:
Thanks for the review.  I will clean those fixes up in the next commits.  

On the point about trapezoidal integration:  I think trapezoidal integration will be fine for now.  I believe that is how many of the instrument vendor softwares currently determine the peak area.  Fitting to a Gaussian model could be better at some point, but I am not sure if that would be needed.  

On the point about sampling rate: that will be a major discrepency in SRM.  I can tell you right now that the sampling rate is not uniform across all of the SRMs.

On the point about background subtract: the instrument vendors currently use the "exact" method, but there are limitations because it ignores the fact that the peak shown is only a part of the actual peak.  It would be better to generate some sort of gaussian model for the peak and then calculate the peak area from that (including the part of the peak that extends to e.g., other peaks and into the background).  However, this probably outside of our scope for now.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
hroest: I think also for peak height, a model that interpolates between the two most intense peaks may be beneficial, e.g. a spline interpolation -- as the "actual" peak height is usually higher than the one observed due to undersampling

dmccloskey: This is very true.  What might be even better that you discussed previously was fitting the data points to a Gaussian or some other model.  I think this would be a great approach.  I have a lot of manually integrated peaks (~100,000) generated from standards as well as an assortment of biological matrices that could be used for fitting such a model as well as validation.

hroest: sounds like a great resource to test and validate these models, so these standards have known concentration ratios?

hroest:
this seems on track @dmccloskey can you look at the still failing tests

dmccloskey:
I can update the tests.  I will let you know when they pass.

dmccloskey:
Tests are passing now.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +169,185 @@ namespace OpenMS
...
+  void MRMTransitionGroupPicker::calculatePeakShapeMetrics_(const MSChromatogram& chromatogram, 
+    double best_left, double best_right, 
+    double peak_height, double peak_apex_rt, double avg_noise_level,
+    PeakShapeMetrics_ & peakShapeMetrics)
+  {
+    peakShapeMetrics.points_across_baseline = 0;
+    double start_intensity(0), end_intensity(0);
+    double delta_rt, delta_int, height_5, height_10, height_50;
+    
+    // pass 1
```
hroest: does this comment have any significance? is there a pass2 ?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -494,6 +502,52 @@ namespace OpenMS
...
+      double baseline_delta_2_height = 0.0;
+      double slope_of_baseline = 0.0;
+      int points_across_baseline = 0;
+      int points_across_half_height = 0;
+    };
+
+    /**
+    @brief Calculates standard peak shape quality metrics
+
+    Standard peak shape quality metrics are calculated for down stream QC/QA.
```
hroest: most of these are self-explanatory. can you add some comments especially for the non-standard metrics such as tailing / asymmetry factor / slope of baseline

- what is their meaning
- how are they calculated?

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -157,6 +169,185 @@ namespace OpenMS
...
+      double intensity_prev = std::max(it_prev->getIntensity()-avg_noise_level, 0.0); //background-subtracted intensity of the previous point
+      double retention_time = it->getMZ();
+      double retention_time_prev = it_prev->getMZ();
+
+      // start and end intensities
+      if (retention_time_prev <= best_left && retention_time >= best_left)
+      {
+        start_intensity = intensity;
+      }
+      else if (retention_time_prev <= best_right && retention_time >= best_right)
```
hroest: hmm now thinking about this, it feels that its still not quite correct. I think it should be correct like this:

```
retention_time_prev < best_right && retention_time >= best_right
```

otherwise, this condition will trigger twice in the case where `best_right` is exactly at a RT position, once when it is equal to the previous one and once when it is equal to the current one. Same for above.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: Nope.  That was there from a previous version I had.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: I see what you mean.  That is probably why the tests changed also.  I will fix that.

dmccloskey:
Docs and tests have been updated per the comments.

PR #2871, opened 2017-08-29
[FEATURE] deleting selections of peak annotations in TOPPView

enetz:
A selection of peak annotations deleted from the TOPPView spectrum canvas
using the DEL key
will now also be removed from the peak annotations of
the current PeptideHit.

### DIFF
--- a/src/openms_gui/include/OpenMS/VISUAL/LayerData.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/LayerData.h
@@ -276,7 +276,7 @@ public:
     void synchronizePeakAnnotations();
 
     /// remove peak annotations in the given list from the currently active PeptideHit
-    void removePeakAnnotationsFromPeptideHit(std::vector<Annotation1DItem*>);
+    void removePeakAnnotationsFromPeptideHit(const std::vector<Annotation1DItem*>& selected_annotations);
 
     /// if this layer is visible
     bool visible;
--- a/src/openms_gui/include/OpenMS/VISUAL/LayerData.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/LayerData.h
@@ -275,6 +275,9 @@ public:
     /// if no PeptideIdentification or PeptideHit for the spectrum exist, it is generated
     void synchronizePeakAnnotations();
 
+    /// remove peak annotations in the given list from the currently active PeptideHit
+    void removePeakAnnotationsFromPeptideHit(std::vector<Annotation1DItem*>);
+
     /// if this layer is visible
     bool visible;
 
--- a/src/openms_gui/source/VISUAL/ANNOTATION/Annotations1DContainer.cpp
+++ b/src/openms_gui/source/VISUAL/ANNOTATION/Annotations1DContainer.cpp
@@ -195,4 +195,17 @@ namespace OpenMS
     }
   }
 
+  std::vector<Annotation1DItem*> Annotations1DContainer::getSelectedItems()
+  {
+    std::vector<Annotation1DItem*> annotation_items;
+    for (Iterator it = begin(); it != end(); ++it)
+    {
+      if ((*it)->isSelected())
+      {
+        annotation_items.push_back(*it);
+      }
+    }
+    return annotation_items;
+  }
+
 } //Namespace
--- a/src/openms_gui/source/VISUAL/ANNOTATION/Annotations1DContainer.cpp
+++ b/src/openms_gui/source/VISUAL/ANNOTATION/Annotations1DContainer.cpp
@@ -197,14 +197,13 @@ namespace OpenMS
 
   std::vector<Annotation1DItem*> Annotations1DContainer::getSelectedItems()
   {
-    std::vector<Annotation1DItem*> annotation_items;
-    for (Iterator it = begin(); it != end(); ++it)
-    {
-      if ((*it)->isSelected())
-      {
-        annotation_items.push_back(*it);
-      }
-    }
+    // initialize with maximal possible size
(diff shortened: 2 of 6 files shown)


enetz:
@timosachsenberg I did not change the code for RightClick->"Delete" on one annotation, as that already seems to work.

src/openms_gui/source/VISUAL/LayerData.cpp
```diff
@@ -217,4 +217,80 @@ namespace OpenMS
...
+              {
+                if (String(pa->getText()).hasPrefix(tmp_a.annotation))
+                {
+                  to_remove.push_back(tmp_a);
+                  annotations_changed = true;
+                }
+              }
+            }
+          }
+          for (auto tmp_a : to_remove)
```
timosachsenberg: const auto& ?

timosachsenberg: please also check above

src/openms_gui/source/VISUAL/LayerData.cpp
```diff
@@ -217,4 +217,80 @@ namespace OpenMS
...
+      // store user fragment annotations
+      vector<PeptideIdentification>& pep_ids = spectrum.getPeptideIdentifications();
+
+      // no ID selected
+      if (peptide_id_index == -1 || peptide_hit_index == -1)
+      {
+        return;
+      }
+
+      if (pep_ids.empty())
```
timosachsenberg: maybe just
if (...) { return; }
and get rid of the else block (might be easier to read)
Same in line 254.

enetz:
Ok, I tested a few more things, this works and is quite robust.
Since the problem with mzid is only writing custom annotations to file, removing annotations is no problem and works for both idXML and mzid equally well.
The only additional issue I found is with editing annotations using RightClick->"Edit".
Edits are not saved yet.
So to make sure users are not confused we should either remove the Edit function,
or update annotations after each edit.

src/openms_gui/source/VISUAL/ANNOTATION/Annotations1DContainer.cpp
```diff
@@ -195,4 +195,17 @@ namespace OpenMS
     }
   }
 
+  std::vector<Annotation1DItem*> Annotations1DContainer::getSelectedItems()
+  {
+    std::vector<Annotation1DItem*> annotation_items;
+    for (Iterator it = begin(); it != end(); ++it)
```
timosachsenberg: consider using http://www.cplusplus.com/reference/algorithm/copy_if/

src/openms_gui/include/OpenMS/VISUAL/LayerData.h
```diff
@@ -275,6 +275,9 @@ namespace OpenMS
     /// if no PeptideIdentification or PeptideHit for the spectrum exist, it is generated
     void synchronizePeakAnnotations();
 
+    /// remove peak annotations in the given list from the currently active PeptideHit
+    void removePeakAnnotationsFromPeptideHit(std::vector<Annotation1DItem*>);
```
timosachsenberg: can this be passed as const reference?

timosachsenberg:
thanks for checking - ok I think it is quite important that edit updates the annotations. Can you take a look at that too?

enetz:
Ok, it should only be another call to the synchronize function.

timosachsenberg:
Annotated m/z geht auch? Von drop down?

enetz:
do you mean the measurements between peaks (shift+LeftDrag)?
yes, with the last commit they can also be deleted without crashing again :) 
with both the DEL key and using the Delete dropdown entry.

timosachsenberg:
And drop down annotate Mz.?

enetz:
Yes, annotations from `Add peak annotation mz` behave the same way as normal peak annotations.

Only labels from `Add label` can still not be saved to file, because they are not anchored to a peak.
Otherwise those work fine as well, and can be removed in the same selection with peak annotations without problems.

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
```diff
@@ -495,7 +495,11 @@ namespace OpenMS
 
     for (Size i = 0; i < frag_annotations.size(); ++i)
     {
-      if (frag_annotations[i].annotation.hasSubstring(String("[alpha|")) || frag_annotations[i].annotation.hasSubstring(String("[beta|")))
+      bool has_alpha = frag_annotations[i].annotation.hasSubstring(String("alpha|"));
+      bool has_beta = frag_annotations[i].annotation.hasSubstring(String("beta|"));
+      // if it has both, it is a complex fragment and more difficult to parse
+      // those are ignored for the coverage indicator for now
+      if ( has_alpha ^ has_beta )
```
timosachsenberg: please use || or "or" instead of bitwise operator^

timosachsenberg: note: the first ones are logical operators the last one is bitwise operator.

enetz: I needed an XOR in this case. Should I write it as
`(has_alpha || has_beta ) && !(has_alpha && has_beta)`?

timosachsenberg: Just use A != B ;)

enetz: right, thanks :)

PR #2873, opened 2017-08-29
SiriusAdapter: checks if csifingerid_result.csv has more than 1 line (headers)

oliveralka:
Seems to be related to segfault error in SiriusAdapter cloud prototype. 

Checks file length - needs more than one line.

### DIFF
--- a/src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
+++ b/src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
@@ -63,7 +63,18 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & paths, Size number
 
     ifstream file(pathtocsicsv);
 
-    if (file)
+    int number_of_lines = 0;
+
+    if (file) 
+    {
+        std::string line;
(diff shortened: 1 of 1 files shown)


oliveralka:
test failed due to "1 - test_pyopenms_unittests (Failed)" unrelated?

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -63,7 +63,18 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & paths, Size number
 
     ifstream file(pathtocsicsv);
 
-    if (file)
+    int number_of_lines = 0;
+
+    if (file) 
+    {
+        std::string line;
+        while (std::getline(file, line))
```
timosachsenberg: will this walk to the end of th efile ?

oliveralka: yes should - is not needed hm? Only need to check if there are more than one file.

oliveralka:
Will refactor a few things on the way and will open a new pull request when changes are ready.

PR #2874, opened 2017-08-30
New tool: FeatureFinderMetaboIdent

hendrikweisser:
Variant of FeatureFinderIdentification that provides targeted feature detection for small molecules. In some ways, a more sophisticated (better?) version of EICExtractor.

Relies on #2862, which I still need to update.

Would be nice if this could make it into the upcoming release.
Tests to follow.


timosachsenberg:
Hi @hendrikweisser . Cool addition.
For the next release it is a bit too late - sorry.

hroest:
pretty cool, I am very much in favor. There is one point that we need to discuss at some point and that is the input list of targets. I think we now have TraML, CSV (output from TargetedFileConverter) that corresponds to the TraML and then your format that is used as input here. It may be nice to think about how to consolidate those file formats.

hendrikweisser:
@hroest: Good point about the input format. I've had a quick look at the TargetedFileConverter documentation regarding the TSV format specification. Unfortunately it looks very proteomics-specific, so I don't think it would be a good idea to reuse the same format here.

hendrikweisser:
@hroest: I've thought about the input format again. TraML and the TargetedExperiment class have explicit support for metabolomics (compounds), but TargetedFileConverter and its TSV format don't. Perhaps we can design a metabolomics-specific TSV format that would work for both TargetedFileConverter and FFMetId.

hendrikweisser:
Only took a bit over a month, but I've added a test now.

timosachsenberg:
@dmccloskey might also be interesting for you

dmccloskey:
Having better support for metabolomics in TargetedFileConverter would be great.  It has been on my wish list of items to try and do.  

Does the new feature integrate with databases for compound ID?

hendrikweisser:
> Does the new feature integrate with databases for compound ID?

If by "new feature" you mean the new tool (FFMetId), the answer is no, at least for now. The input file is a simple tab-separated text file (TSV, see here for an example: https://github.com/hendrikweisser/OpenMS/blob/5e9ced43a599280badde634b130ee4cbd906d24a/src/tests/topp/FeatureFinderMetaboIdent_1_input.tsv). The output is featureXML, which you can then convert to TSV using TextExporter (with option "feature:add_metavalues 0" to get all the annotations).

hroest:
> @hroest: I've thought about the input format again. TraML and the TargetedExperiment class have explicit support for metabolomics (compounds), but TargetedFileConverter and its TSV format don't. Perhaps we can design a metabolomics-specific TSV format that would work for both TargetedFileConverter and FFMetId.

@hendrikweisser we now do have support for metabolomics in the TSV format, see https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/ConvertTSVToTraML_5_input.tsv  it would be great if you could rename your headers as follows:

- Name -> CompoundName
- Formula -> SumFormula

The first one is very important as I use the presence/absence of "CompoundName" to determine whether a specific file contains proteomic or metabolomic data.

hroest:
PS: this is the TraML that is generated from the tsv above, as you  can see it directly supports metabolomics and all metabolic compounds are stored not as "Peptide" but as "Compound" and  it follows the suggested CV terms. 

https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/ConvertTSVToTraML_5_output.TraML

hendrikweisser:
@hroest:
> it would be great if you could rename your headers as follows:

Unfortunately there's a slight problem with that: If I rename those fields, I end up with a mix of "CamelCase" and "Snake_case" in the header (like in your TSV exampe - what's up with that?). I'd have to rename the other fields accordingly, but unfortunately "RTRange" looks a bit weird.
So I'd like to make sure that we actually gain anything from this renaming. Do we? Given that the other columns in the formats don't match, there isn't really any interoperability, or am I wrong?

hroest:
I guess the main thing to gain is that the conversion to traml would work. Alternatively you can also modify the tsv reader to recognize your header names.

hendrikweisser:
> I guess the main thing to gain is that the conversion to traml would work.

But would that actually work if the RT, m/z and charge information is missing?

hroest:
> But would that actually work if the RT, m/z and charge information is missing?

I dont think I understand, that information is present in the input file? The information should be present, right? Or is your argument aesthetic regarding camel / snake case?

hendrikweisser:
> The information should be present, right?

Charge is there, which would correspond to PrecursorCharge. However, m/z is defined indirectly through mass and charge, and RT is not calibrated to be between 0 and 1.

hroest:
> So I'd like to make sure that we actually gain anything from this renaming. Do we? Given that the other columns in the formats don't match, there isn't really any interoperability, or am I wrong?

I would say we gain if we can convert the tsv to TraML. If you have a 1:1 mapping of your file to a TraML file I think we do gain as TraML is a standard format. It just seems like an easy fix now whereas if we want to do it later its a pain to do it and we need to check for different possible CSV headers and write out different headers as well when converting TraML back to TSV, that is my reasoning. 


> However, m/z is defined indirectly through mass and charge, 

I guess we can work with that, however when converting TraML back to tsv I guess we would just write out m/z 

> and RT is not calibrated to be between 0 and 1.

that is not a big problem, there is a switch in TargetedFileConverter:

```
defaults_.setValue("retentionTimeInterpretation", "iRT", "How to interpret the provided retention time (the retention time column can either be interpreted to be      in iRT, minutes or seconds)", ListUtils::create<String>("advanced"));
defaults_.setValidStrings("retentionTimeInterpretation", ListUtils::create<String>("iRT,seconds,minutes"));
```

hroest:
PS: my point is that if the shared columns match then we can work from there and incorporate your other fields into the TraML format. If they dont match it will make everything harder down the line. 

If you rename RT to either `RetentionTime` or `Tr_recalibrated` and add 
- `PrecursorMz`
- `ProductMz`
- `LibraryIntensity`

it should work. We can also make LibraryIntensity optional for the conversion.

hroest:
PS: can you fix
```

src/utils/FeatureFinderMetaboIdent.cpp
Member variable 'TOPPFeatureFinderMetaboIdent::keep_chromatograms_' is not initialized in the constructor.
136

TOPPFeatureFinderMetaboIdent() :

```

hroest:
@hendrikweisser whats the status of this?

hendrikweisser:
@hroest: I haven't had time for this yet, but yes, let's work towards compatibility. Questions for you:
1. Do you want to take this opportunity to fix inconsistencies in the column names of the OpenSWATH TSV format? (Between "ProductMz 	Tr_recalibrated 	transition_name" there are three difference styles.)
2. Can we make columns that don't make sense for FFMetId optional (e.g. "ProductMZ")?

Another point that just occurred to me is that the semantics of a "row" in the two formats differ: In OpenSWATH, one row corresponds to one transition, whereas for FFMetId, one row will generate multiple transitions. Is that going to be a problem?

hroest:
> Do you want to take this opportunity to fix inconsistencies in the column names of the OpenSWATH TSV format? (Between "ProductMz Tr_recalibrated transition_name" there are three difference styles.)

yes, I would be very much in favor, even though for now we just need to *accept* different input data

> Can we make columns that don't make sense for FFMetId optional (e.g. "ProductMZ")?

sure

> Another point that just occurred to me is that the semantics of a "row" in the two formats differ: In OpenSWATH, one row corresponds to one transition, whereas for FFMetId, one row will generate multiple transitions. Is that going to be a problem?

well it is not optimal as it is easier if each chromatogram can be referenced to a single row with a single row, but it wont really create problems as long as you process the data downstream in your tool correctly.

hroest:
@hendrikweisser we have now improved the consistency of the header columns, lets hope we can also move forward here!

hendrikweisser:
@hroest: Thanks for making the changes in OpenSWATH (to George as well).
I've renamed the columns as follows:
`CompoundName    SumFormula    Mass    Charge    RetentionTime    RetentionTimeRange    IsoDistribution`

Is that good or did I overlook something?

By the way, I currently support multiple charges and RT values per row in the input. How would that be handled during a conversion?

hendrikweisser:
@hroest: Hannes, I'm trying to revive this tool (should be useful for RNA quantification). I'm struggling to resolve the merge conflicts in the OpenSWATH `ChromatogramExtractor` - see https://github.com/OpenMS/OpenMS/pull/2874/conflicts.
What I did was refactor out a function `make_coordinates_` that deals with the different cases for RT start/end, then add branches in `prepare_coordinates` to handle compounds instead of peptides (see original changeset: https://github.com/OpenMS/OpenMS/pull/2874/files#diff-21966afb4ba9653d60dbe2a4f2968a0b).

How would that go together with the changes you've made since then?

hendrikweisser:
@hroest:
> How would that go together with the changes you've made since then?

Since it looks like you added support for compounds, I've tried it with your version of `ChromatogramExtractor`, without any of my changes. However, when I run the FFMetId test, I get this error (from `ChromatogramExtractorAlgorithm`):
`Error: Unexpected internal error (Output and extraction coordinates need to have the same size: 24 != 12)` 
(That is after adapting FFMetId to the new way of generating isotope distributions.)

hendrikweisser:
> However, when I run the FFMetId test, I get this error

Seems like this was due to a mistake I made during merging. Never mind, it looks like it's working now.

hendrikweisser:
I've added documentation/comments. Moving code to an algorithm class isn't a priority at the moment. From my side this should be good to merge.

(Not sure what's up with Travis. Can the builds be restarted?)

timosachsenberg:
comments / documentation is still a bit sparse but well... just complaining because I sometimes got surprised in the proteomics counterpart ;)
Travis is currently struggling with build times after they once again changed their build infrastructure.
If we find the time we should test CircleCI or similar projects.

PR #2878, opened 2017-09-01
[FIX] Reduce warning, fix error

oliveralka:
fix some warnings 
errors 

ctest passes 99%
```
449 - SVMWrapper_test (Failed)
 -  line 1055:  TEST_EQUAL((predicted_labels[i] < 0 && probabilities[i] < 0.5) || (predicted_labels[i] > 0 && probabilities[i] >= 0.5),true): got 0, expected 1
```


src/openms/source/CHEMISTRY/EnzymaticDigestionLogModel.cpp
```diff
@@ -161,7 +161,7 @@ namespace OpenMS
   {
     SignedSize count = 1;
     AASequence::ConstIterator iterator = protein.begin();
-    while (nextCleavageSite_(protein, iterator), iterator != protein.end())
+    while (static_cast<void>(nextCleavageSite_(protein, iterator)), iterator != protein.end())
```
timosachsenberg: hmm what did he complain about here?

oliveralka: possible misuse of comma operator here [-Wcomma]

timosachsenberg: ok what you could also do to make this more readable is to write:
```
nextCleavageSite_(protein, iterator);
while (iterator != protein.end())
{
// as before...
nextCleavageSite_(protein, iterator);
}
```
that is the doing the same thing and don't uses the comma operator in while loops

oliveralka: ok, thanks!

oliveralka: How would this function look then? 
```
Size EnzymaticDigestionLogModel::peptideCount(const AASequence& protein)
  {
    SignedSize count = 1;
    AASequence::ConstIterator iterator = protein.begin();
    while (static_cast<void>(nextCleavageSite_(protein, iterator)), iterator != protein.end())
    {
      ++count;
    }
    Size sum = count;
    return sum;
  }
```
Like this?
```
Size EnzymaticDigestionLogModel::peptideCount(const AASequence& protein)
  {
    SignedSize count = 1;
    AASequence::ConstIterator iterator = protein.begin();
    while (iterator != protein.end())
    {
      nextCleavageSite_(protein, iterator)
      ++count;
    }
    Size sum = count;
    return sum;
  }
```

timosachsenberg: nextCleavageSite_(protein, iterator);
while (iterator != protein.end())
{
++count;
nextCleavageSite_(protein, iterator);
}

src/openms/source/CHEMISTRY/EnzymaticDigestionLogModel.cpp
```diff
@@ -161,9 +161,10 @@ namespace OpenMS
   {
     SignedSize count = 1;
     AASequence::ConstIterator iterator = protein.begin();
-    while (static_cast<void>(nextCleavageSite_(protein, iterator)), iterator != protein.end())
+    while (iterator != protein.end())
```
timosachsenberg: I think this now (first count, then advance) produces different counts as before (iterator was first advanced then tested for end()).
Though I don't see for sure if it is now correct or was before.
same below.

oliveralka: What du you think is best just try it out or revert back and live with the warning?

oliveralka: What du you think is best just try it out or revert back and live with the warning?

timosachsenberg: it should produce the same result with:
nextCleavageSite_(protein, iterator);
while (iterator != protein.end())
{
++count;
nextCleavageSite_(protein, iterator);
}

src/openms/source/ANALYSIS/QUANTITATION/IsobaricChannelExtractor.cpp
```diff
@@ -531,7 +531,7 @@ namespace OpenMS
         // remember potential precursor and continue
         pState.precursorScan = it;
         // reset last MS2 -- we expect to see a new one soon and the old one should not be used for the following MS3 (if any)
-        it_last_MS2 == ms_exp_data.end();
+        it_last_MS2 = ms_exp_data.end();
```
hroest: thanks! I guess its not a very serious bug that I can tell but thanks for catching that

src/openms/source/CHEMISTRY/EnzymaticDigestionLogModel.cpp
```diff
@@ -161,9 +161,10 @@ namespace OpenMS
   {
     SignedSize count = 1;
     AASequence::ConstIterator iterator = protein.begin();
-    while (nextCleavageSite_(protein, iterator), iterator != protein.end())
+    while (iterator != protein.end())
```
hroest: I think this changes the code flow for an empty protein, makes it correct now. may also be good to test

src/openms/source/CHEMISTRY/EnzymaticDigestionLogModel.cpp
```diff
@@ -175,8 +176,9 @@ namespace OpenMS
     output.clear();
     AASequence::ConstIterator begin = protein.begin();
     AASequence::ConstIterator end = protein.begin();
-    while (nextCleavageSite_(protein, end), end != protein.end())
+    while (end != protein.end())
     {
+      nextCleavageSite_(protein, end)
```
oliveralka: Is this correct?

oliveralka:
May someone has a look at the while loops in 
`src/openms/source/CHEMISTRY/EnzymaticDigestionLogModel.cpp`
not sure if they are correct.
Are any more changes for master warnings required before merge?

src/openms/source/CHEMISTRY/EnzymaticDigestionLogModel.cpp
timosachsenberg: flow is probably correct now but count is 1 instead of zero for empty proteins

src/openms/source/CHEMISTRY/EnzymaticDigestionLogModel.cpp
timosachsenberg: if it was correct before then it now produces a different result.
Note that 
while (nextCleavageSite_(protein, end), end != protein.end()) {}
is exactly the same as
nextCleavageSite_(protein, end);
while (end != protein.end()) {}

lukaszimmermann:
ok, looks good. I merge.

PR #2879, opened 2017-09-01
AccurateMassSearch: write out m/z deviation in Da

hendrikweisser:
AccurateMassSearch(Engine) used to write the relative m/z deviation (in ppm) between a feature and a compound into a meta value "ppm_mz_error". It now writes both relative (ppm) and absolute (Da) deviations into "mz_error_ppm" and "mz_error_Da", respectively.


timosachsenberg:
looks good. I hope the change in parameter name does not affect anyone.

PR #2880, opened 2017-09-01
[FIX, REFACTOR] SiriusAdapter 

oliveralka:
fix: 
Add number parameter (how many candidates should be written) > there was the issue that the number parameter was overwritten and then not set back to the original number. 
TOPP_Test

refactor:
-fingerid parameter is now used if -out_fingerid  (output folder) is set.


timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=237368

lukaszimmermann:
ups, my bad

oliveralka:
Thanks Timo & Lukas!

lukaszimmermann:
looks good, gonna merge

PR #2886, opened 2017-09-03
Release_2_3/code uncrustify

lukaszimmermann:
These are the code changes that the application of `uncrustify` to `tools/uncrustify.cfg` would introduce. Not sure if we really want to apply those or if we want to cherry-pick some of these changes. It mostly adds curly braces for blocks, parentheses for expressions, and changes indentation. Basically all files are changed.  I open a PR to see if the changes break any tests.


timosachsenberg:
Probably better to not apply it to the nsis folder

timosachsenberg:
maybe we should discuss this and do this at some other point. We have quite some PRs open and it might complicate things.

PR #2891, opened 2017-09-05
[FIX] Change code in Python stuff that is not gonna work in Python 3

lukaszimmermann:
http://python-future.org/compatible_idioms.html#octal-constants


hroest:
+1 it doesnt hurt even though Python3 tests have already been working on my machine here for a while

lukaszimmermann:
yes, I could not compile pyOpenMS due to the octal literal notation and the tests in the `collections_.py` file could not be executed with a Python 3 interpreter.

PR #2892, opened 2017-09-05
Peptide digestion filter

pati-ni:
Add **missed_cleavages** feature to the **IDFilter**. 

Peptide hits can be filtered on a `[min,max]` range of missed_cleavages based on a specified enzyme.

@lars20070


lars20070:
Just tested the functionality. Works as expected.

+1

lars20070:
Could you please use ranges for the missed cleavages https://github.com/pati-ni/OpenMS/blob/2a3b846855d910eaf23ec87eb8c7c0fe4e76b307/src/topp/IDFilter.cpp#L253 instead of `min` and `max`?

lars20070:
@cbielow Could this still be included in OpenMS 2.3?

src/openms/include/OpenMS/CHEMISTRY/EnzymaticDigestion.h
```diff
@@ -122,6 +123,9 @@ namespace OpenMS
     /// @note: the protein sequence must be unmodified (i.e. only amino acid characters, no modifications or masses in brackets)
     bool isValidProduct(const String& protein, Size pep_pos, Size pep_length, bool methionine_cleavage = false, bool ignore_missed_cleavages = true) const;
 
+    ///
+    bool filterByMissingCleavages(const String& sequence, std::function<bool(const Int)> filter) const;
```
cbielow: missing doc. The interface is pretty contrived, so I think its important to have it.

sequence: is an unmodified(!) peptide sequence
filter: ....

Also, the variable name in the definition for sequence is different

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -336,6 +336,62 @@ namespace OpenMS
...
+       @brief Filter Peptide Hit by its digestion product
+
+    */
+
+    struct PeptideDigestionFilter
+    {
+
+      typedef PeptideHit argument_type;
+
+      EnzymaticDigestion& digestion_;
```
cbielow: this section is public here, but should be private

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -336,6 +336,62 @@ namespace OpenMS
...
+      Int min_cleavages_;
+      Int max_cleavages_;
+      
+      PeptideDigestionFilter(
+        EnzymaticDigestion& digestion) :  
+      digestion_(digestion)
+      {}
+
+      inline void setMinCleavagesAllowed(const Int min_cleavages){ min_cleavages_ = min_cleavages; }
```
cbielow: The functions are used exactly once during the duration of the object. So we might as well set the value in the C'tor. Allow defaults (-1) if you want.
--> remove getter/setter

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -336,6 +336,62 @@ namespace OpenMS
...
+      inline void disableCleavagesLowerBound(){ min_cleavages_ = disabledValue(); }
+      inline void disableCleavagesUpperBound(){ max_cleavages_ = disabledValue(); }
+
+      inline Int disabledValue(){ return -1; }
+      
+      /// Filter function on min max cutoff values 
+      bool operator()(PeptideHit& p)
+      {
+        return digestion_.filterByMissingCleavages(
+          p.getSequence().toString(),
```
cbielow: I believe this must be `p.getSequence().toUnmodifiedString()`!! Otherwise there will be issues with modifications.
Please add a test-case.

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -336,6 +336,62 @@ namespace OpenMS
...
+
+      inline Int disabledValue(){ return -1; }
+      
+      /// Filter function on min max cutoff values 
+      bool operator()(PeptideHit& p)
+      {
+        return digestion_.filterByMissingCleavages(
+          p.getSequence().toString(), 
+          [&](const Int missed_cleavages)
+          {
```
cbielow: [DOC] return true if peptide should be removed (does not pass filter)

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -336,6 +336,62 @@ namespace OpenMS
...
+            bool max_filter = max_cleavages_ != disabledValue() ? 
+                              missed_cleavages > this->max_cleavages_ : false;
```
cbielow: remove `this` (for consistency with min-case)

src/tests/topp/CMakeLists.txt
```diff
@@ -891,6 +891,25 @@ add_test("TOPP_IDFilter_16" ${TOPP_BIN_PATH}/IDFilter -test -in ${DATA_DIR_TOPP}
 add_test("TOPP_IDFilter_16_out1" ${DIFF} -in1 IDFilter_16_output.tmp -in2 ${DATA_DIR_TOPP}/IDFilter_16_output.idXML)
 set_tests_properties("TOPP_IDFilter_16_out1" PROPERTIES DEPENDS "TOPP_IDFilter_16")
 
+#filters by peptide missed cleavages
+add_test("TOPP_IDFilter_17" ${TOPP_BIN_PATH}/IDFilter -test -in ${DATA_DIR_TOPP}/IDFilter_missed_cleavages_input.idXML -out IDFilter_17_output.tmp -missed_cleavages:enzyme Lys-N -missed_cleavages:max 2)
```
cbielow: change to `-missed_cleavages :2` as Lars suggested.. same below

src/topp/IDFilter.cpp
```diff
@@ -197,7 +197,34 @@ class TOPPIDFilter :
     setMinInt_("digest:missed_cleavages", -1);
     registerFlag_("digest:methionine_cleavage", "Allow methionine cleavage at the protein start", false);
 
-
+    registerTOPPSubsection_(
```
cbielow: please make these one liners. its really hard to read.

src/topp/IDFilter.cpp
```diff
@@ -479,6 +506,39 @@ class TOPPIDFilter :
...
+    Int min_cleavages = getIntOption_("missed_cleavages:min");
+    Int max_cleavages = getIntOption_("missed_cleavages:max");
+    
+    if (not (min_cleavages == -1 && max_cleavages == -1))
+    {
+      LOG_INFO << "Filtering peptide hits by their missed cleavages count..." << endl;
+
+      // Configure Enzymatic digestion
+      EnzymaticDigestion digestion;
+      String enzyme = getStringOption_("missed_cleavages:enzyme").trim();
```
cbielow: no need to trim(), since the values are a controlled set anyway.

src/topp/IDFilter.cpp
```diff
@@ -479,6 +506,39 @@ class TOPPIDFilter :
...
+      if (!enzyme.empty())
+      {
+        digestion.setEnzyme(enzyme);
+      }
+
+      Int min_cleavages = getIntOption_("missed_cleavages:min");
```
cbielow: parse range here

src/topp/IDFilter.cpp
```diff
@@ -479,6 +506,39 @@ class TOPPIDFilter :
...
+      Int max_cleavages = getIntOption_("missed_cleavages:max");
+      
+      // Build the digest filter function
+      IDFilter::PeptideDigestionFilter filter(digestion);
```
cbielow: init filter with all params in Ctor

src/topp/IDFilter.cpp
```diff
@@ -479,6 +506,39 @@ class TOPPIDFilter :
...
+      filter.setMaxCleavagesAllowed(max_cleavages);
+      filter.setMinCleavagesAllowed(min_cleavages);
+
+      // Filter peptide hits      
+      for (auto& peptide : peptides)
+      {
+        auto hits = peptide.getHits();
+        filter.filterPeptideSequences(hits);
```
cbielow: `filter.filterPeptideSequences(peptide.getHits())` should do (since it will modify in-place)?

pati-ni: It violates its [const property](https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/METADATA/PeptideIdentification.h#L126).

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
pati-ni: Good catch! I added some tests for that case though I can not quite see why this would not work.

src/openms/include/OpenMS/CHEMISTRY/EnzymaticDigestion.h
```diff
@@ -122,6 +123,13 @@ namespace OpenMS
     /// @note: the protein sequence must be unmodified (i.e. only amino acid characters, no modifications or masses in brackets)
     bool isValidProduct(const String& protein, Size pep_pos, Size pep_length, bool methionine_cleavage = false, bool ignore_missed_cleavages = true) const;
 
+    /**
+       @param sequence amino-acid sequence to be digested
+       @param filter a function that takes as a parameter the missed cleavages number
+       @return a bool value based on the missed cleavages number with respect to the filter function
```
cbielow: I know its hard to describe what the return value is since the filter function is unknown, but the ways its used here, it will return true if .... 
currently, the user cannot tell what filterByMissingCleavages is actually doing.

src/tests/class_tests/openms/source/IDFilter_test.cpp
```diff
@@ -180,6 +182,52 @@ START_SECTION((static void extractPeptideSequences(const vector<PeptideIdentific
...
+  digestion.setEnzyme("Trypsin");
+  
+  IDFilter::PeptideDigestionFilter filter(digestion, 0, 1);
+  vector<PeptideHit>hits, test_hits;
+
+  
+  // No cleavage
+  hits.push_back(PeptideHit(0, 0, 0, AASequence::fromString("(MOD:00051)DFPIANGER")));
+  hits.push_back(PeptideHit(0, 0, 0, AASequence::fromString("DFPIANGER")));
+  hits.push_back(PeptideHit(0, 0, 0, AASequence::fromString("DFPIAN(Deamidated)GER")));
```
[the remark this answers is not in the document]
pati-ni: I find difficult to be certain of such example.
If I use Asp-N_ambic `Asp-N Ammonium bicarbonate cleaves before D or E.`
and `F(MOD:00051)DFPIANGR` as the sequence will that satisfy the case you re describing (zero missed cleavages)?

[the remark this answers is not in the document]
cbielow: yes. perfect. thanks.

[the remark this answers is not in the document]
pati-ni: It bit me for the sequence `"F(MOD:00051)DFPIANGER"` 

`Message: the value 'MOD:00051' was used but is not valid; Retrieving the modification failed. It is not available for the residue 'F' and term specificity 0.`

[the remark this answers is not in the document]
lars20070: `MOD:00051` = `A protein modification that effectively converts an L-aspartic acid residue to N-acetyl-L-aspartic acid.` As I understand this modification can only occur at aspartic acid  `D` and not `F`.

Where is the  sequence `F(MOD:00051)DFPIANGER` from?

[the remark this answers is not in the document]
kernursula: The isue of cleavages after modifiied amino acids is very complicated. It depends on the enzyme and the modification, for example: Trypsin can cleave after monomethylated Lysine, but not dimethylated or acetylated lysines.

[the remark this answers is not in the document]
pati-ni: Currently digestion works naively with UnmodifiedStrings. This seems to be a functionality missing in digestion, that is ignoring modifications that may nullify the rule for a certain cleavage.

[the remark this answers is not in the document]
lars20070: @cbielow Are you fine with the sequence `K(Dimethyl)FPIAUGR` for the test?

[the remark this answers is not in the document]
cbielow: yes. In combination with enyzme `Asp-N_ambic` (Asp-N Ammonium bicarbonate cleaves before D or E.) this would test exactly what was intended, i.e.
a cleavage before `D` should not occur (since its part of the modification name, which should not be passed into the regex).

timosachsenberg:
Does it handle unspecific and no cleavage?

pati-ni:
@timosachsenberg Yes, I think it does for both.

Given that [this regex](https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/CHEMISTRY/Enzymes.xml#L256) is correct it should function correctly for unspecified cleavages. 

There is a [special case](https://github.com/pati-ni/OpenMS/blob/76afac79199d6a8b871961233eb63dba753c752d/src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp#L120) where it is the no cleavage case that returns no cleavages.

lars20070:
@pati-ni I think this should read
```
registerStringOption_("missed_cleavages:missed_cleavages", "[min]:[max]", ":", "[min,max] range of allowed missed cleavages\n" "Empty value ignores bound", false);
```
here https://github.com/OpenMS/OpenMS/blob/develop/src/topp/IDFilter.cpp#L201
i.e. parameter `missed_cleavages` in section `missed_cleavages`. At the moment the new parameter is not available in `TOPPAS`.

pati-ni:
Should it be something like `missed_cleavages:range` ?

lars20070:
@kernursula suggests `missed_cleavages:number_of_missed_cleavages`.

timosachsenberg:
I just saw that there seems to be some kind of redundancy regarding the paramter

   registerIntOption_("digest:missed_cleavages", "<integer>", -1, 
                       "filter peptide evidences that have more than the specified missed_cleavages\n"
                       "By default missed cleavages are ignored", false);
    setMinInt_("digest:missed_cleavages", -1);
    registerFlag_("digest:methionine_cleavage", "Allow methionine cleavage at the protein start", false);

    registerTOPPSubsection_("missed_cleavages", "Filter peptide sequences by their missed cleavages number");
    registerStringOption_("missed_cleavages", "[min]:[max]", ":", "[min,max] range of allowed missed cleavages\n" "Empty value ignores bound", false);
    registerStringOption_("missed_cleavages:enzyme", "<enzyme>", "Trypsin", "Specify the digestion enzyme", false);
    setValidStrings_("missed_cleavages:enzyme", all_enzymes);

timosachsenberg:
I just saw that there seems to be some kind of redundancy regarding the paramter

   registerIntOption_("digest:missed_cleavages", "<integer>", -1, 
                       "filter peptide evidences that have more than the specified missed_cleavages\n"
                       "By default missed cleavages are ignored", false);
    setMinInt_("digest:missed_cleavages", -1);
    registerFlag_("digest:methionine_cleavage", "Allow methionine cleavage at the protein start", false);

    registerTOPPSubsection_("missed_cleavages", "Filter peptide sequences by their missed cleavages number");
    registerStringOption_("missed_cleavages", "[min]:[max]", ":", "[min,max] range of allowed missed cleavages\n" "Empty value ignores bound", false);
    registerStringOption_("missed_cleavages:enzyme", "<enzyme>", "Trypsin", "Specify the digestion enzyme", false);
    setValidStrings_("missed_cleavages:enzyme", all_enzymes);

timosachsenberg:
maybe we should just (re-)use "digest:missed_cleavages" ?

pati-ni:
@timosachsenberg  No that is a different subsystem actually, the digestion subsection requires a fasta file with proteins to digest. This instead just digests the peptide sequences.

lars20070:
I am also for `missed_cleavages` to appear twice i.e. once in each section.
But they should be consistent i.e. same name and both ranges `min:max`.

kernursula:
Maybe a bit off-topic but now is the chance to mention it. Personally I don`t understand the description of "digestion". It says " Perform Protein digest and filter peptides based on digestion products". But what the filter actually does is to remove proteins and not peptides from your files. Reading the description we (Lars, Oliver, Timo) thought that the "digestion" part could do what the "missed cleavages" part can do now. Maybe the descriptions could be more detailed?

lars20070:
As I understand, the `digestion` section removes proteins from the `idXML` while the `miscleavage` section removes PSMs from the `idXML`. Maybe that should be  made clearer with the names/docu of the sections.

pati-ni:
Ok, I have made a fix for the issue, should I attempt to fix the documentation as well?

pati-ni:
Both use the digestion system to perform filtering. If anyone has some can come up with a naming  scheme that would make it clearer, let me know.

Maybe name the sections protein_digestion, peptide_sequence_digestion respectively.

pati-ni:
@lars20070 fixed namespaces. @kernursula can you check the current iteration of the documentation and give me some feedback?

kernursula:
Here are my suggestions for the docu of digestion_peptide_level:
+++ digestion_peptide_level +++
This filter removes PSMs from the input file which do not confirm with the rules specified below
++ enzyme ++
Enzyme used for digestion
++ number_of_missed_cleavages ++
Peptides within the specified range will not be removed e.g.
number_of_missed_cleavages: 0:1 -> peptides with two or more missed cleavages will be removed or
number_of_missed_cleavages: 0:0 -> peptides with one or more missed cleavages will be removed

lars20070:
Great. Thank you @pati-ni.

Regarding the documentation. How about the following two sections:
```
digestion_peptide_level
(docu: This filter option removes PSMs which do not confirm with the rules specified below.)
enzyme
(docu: enzyme used for the digestion of the sample)
number_of_missed_cleavages
(docu: range of allowed missed cleavages in the peptide sequences)
```
And your previous `digestion` section.
```
digestion_protein_level
(docu: This filter option removes protein IDs which do not confirm with the rules specified below.)
fasta
(docu: fasta protein sequence database.)
enzyme
(docu: enzyme used for the digestion of the sample)
specificity
(docu: enzyme specificity)
number_of_missed_cleavages
(docu: range of allowed missed cleavages in the peptide sequences)
methionine_cleavage
(docu: Allow methionine cleavage at the N-terminus of the protein.)
```

kernursula:
For the docu of digestion_protein_level: I can not really help, because I don`t fully understand what the filter does, sorry. maybe something like: Perform in silico Protein digest of fasta.file, compare PSMs to digestion products and removes proteins that  ??

timosachsenberg:
ok I took a (very) quick look and the semantics seem to be as follows:
- You can either filter PeptideHits  if they are a valid digestion product of a given Protein (fasta) database and enzyme settings
- Or you can filter PeptideHits only according to the number of missed cleavages and a given enzyme

For me, these seem rather similar and one could just make the fasta database filename optional.
If it is given, it will use the database. If not, only resort to counting missed cleavages.

Would this make sense?

kernursula:
But when I use the "digestion" option in ID filter only the number of Proteins drops, the number of peptides does not change, so to my understanding not peptides, but proteins are filtered. 
Using the new "missed cleavages" option, here actually the number of peptides drops, meaning peptides are filtered. Aren`t those two different things?

lars20070:
FYI, here the additional changes. https://github.com/OpenMS/OpenMS/pull/2928

pati-ni:
@timosachsenberg you are almost right. 

The `'digestion'` filters on the `PeptideEvidence` level. So @kernursula  if you see the protein number changing this is just a side effect of not keeping unreferenced proteins.

The `'missed_cleavages'` option filters on `PeptideHit` level. 

So @timosachsenberg I do not know if it is exactly a good idea of keeping two basically different type of filters with redundant parameters together. For sure though it can be done on the CLI level but under the hood their differences are fundamental.

Hope that this helps..

lars20070:
@pati-ni Thanks for the clarification. I am for two separate sections `missed_cleavages` and `in_silico_digestion` as outlined [here](https://github.com/OpenMS/OpenMS/pull/2892#issuecomment-333085132). After that I am for merging #2928.

timosachsenberg:
ok that is fine with me

PR #2895, opened 2017-09-06
Fix/qt5 fix3d

hroest:
fixes some issues on Linux with 3D rendering

- removes flicker
- fixes the positioning of the 3D object

### DIFF
--- a/src/openms_gui/include/OpenMS/VISUAL/Spectrum3DOpenGLCanvas.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/Spectrum3DOpenGLCanvas.h
@@ -82,7 +82,7 @@ public:
     virtual ~Spectrum3DOpenGLCanvas();
 
     ///helper function to replicate old behaviour of QGLWidget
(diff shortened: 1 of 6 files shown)


hroest:
for Linux, this is a marked improvement :-) 

before:

3dview_1

after:

paint3d_after

hroest:
In my testing, `QPainter::beginNativePainting()` was responsible for the behavior above where nothing would get painted // all painting would be in the upper right corner

Regarding begin/end I dont think we need them and they only create warning output on the CLI. see also the docu: http://doc.qt.io/qt-5/qpainter.html#begin

* Note that most of the time, you can use one of the constructors instead of begin(), and that end() is automatically done at destruction.

hroest:
@jpfeuffer @timosachsenberg can you also have a look here?

hroest:
@jpfeuffer @timosachsenberg can you also have a look here?

hroest:
@jpfeuffer @timosachsenberg can you also have a look here? merge?

timosachsenberg:
I think this is good to merge. Can't confirm on mac right now.

PR #2896, opened 2017-09-07
[DOC] updated required Visual Studio version for building on windows

enetz:
The doc still said:
have a Visual Studio 2005/2008/2010[SP1]/2012/2013/2015 installed
while we do not support anything older than 2015 anymore

### DIFF
--- a/doc/doxygen/install/install-win.doxygen
+++ b/doc/doxygen/install/install-win.doxygen
@@ -32,15 +32,11 @@
 
     <b>Pre-Requisites (Software)</b>
     <ul>
-      <li>have a Visual Studio 2005/2008/2010[SP1]/2012/2013/2015 installed (The C++ Express editions should work as well, we tested VSE2008, VSE2012 and VSE2013).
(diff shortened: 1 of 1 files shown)


enetz:
@jpfeuffer do we officially support VS 2017 at this point, or is that still at the experimental stage?

cbielow:
VS2013 might actually work, since it supports a major subset of C++11.
Either we test this or you write it vaguely, i.e. VS2013 might still work compile.

enetz:
I tested it a few months ago, see #2734 
Did not go well

cbielow:
ok, so vs2013 is out.

timosachsenberg:
Any news on VS2017?

timosachsenberg:
applied to master

jpfeuffer:
VS2017 should be fully supported by now. Windows CI runs with it (although it currently has hiccups because of ssh problems to the machine).

timosachsenberg:
ok I can add it

PR #2898, opened 2017-09-07
Re-Write of PeptideIndexer with efficient memory and runtime

cbielow:
major changes are:
 - using a single Aho-Corasick-based matching algorithm which can handle ambiguous AA's (dropping support for general mismatches)
 - this greatly simplifies parameters (e.g. removes -full_tolerant_search)
 - large FASTA files (>200 MB) are automatically read and streamed chunkwise, for minimal memory footprint (usually <3 GB)
 - almost linear speedup with threads (for searching and reading of FASTA file; not reading/writing of idXML)

minor changes:
 - report more statistics for matching at the end
 - proteinHit attributes (rank, score) are not preserved from input idXML, since they are invalid (and thus misleading) anyway
 - un-matched peptides (which should not be present, but can be enforced by user as before), do not have a target_decoy metavalue anymore (because it would be empty anyway).
  - duplicate FASTA accessions within the FASTA file do not create a warning anymore (since the FASTA file might be read piecewise and checking the FASTA should be done once prior even to search-engine invokation); this might lead to multiple peptide hits (making them non-unique), one for each sequence


cbielow:
while I'm rather confident in correctness of this algorithm, this should not go into 2.3 since there could be issues with the slight modifications to the way meta-values are written out (e.g. how downstream TOPP tools react) -- a good point for pipeline tests btw :)

timosachsenberg:
Does this affect the export of common meta values in textexporter?

cbielow:
indeed it does when using '100', but having unmatched peptides.
If one were to use '0', then it will still work.
Might be a matter of taste. The new behaviour will 'break' old workflows (if they use 100 and have unmatched peptides). Going back to the old behaviour introduces empty meta-values for unmatched peptides (and increases filesize a bit).

timosachsenberg:
What are your plans for the dropped features? Move them in another tool? Thought these might be useful for proteogenomic

cbielow:
> What are your plans for the dropped features? Move them in another tool? Thought these might be useful for proteogenomic

you refer to allowing real mismatches? This would indeed require a new tool or the use of a previous release of OpenMS.

timosachsenberg:
@hendrikweisser I think you have most experience with proteogenomics. What is your take on this?

cbielow:
it is possible to extend this algorithm to allow for mismatches, but it take some time (weeks) before I get around implementing it.
suggestion: to allow for a long-enough testing period of the new implementation (especially meta-values as mentioned above), lets merge this now. I'll add support for MM later. Also, this saves potential merge conflicts later and allows other people to try the FASTA streaming which comes with this PR.

timosachsenberg:
fine with me. can you resolve the merge conflict?

cbielow:
done

jpfeuffer:
4>C:\Program Files (x86)\Windows Kits\10\Include\10.0.14393.0\ucrt\stdio.h(1925): warning C4005: 'snprintf': macro redefinition (compiling source file C:\jenkins\ws\openms_nightly_packaging\c2e226b2\source\src\openms\source\ANALYSIS\ID\AhoCorasickAmbiguous.cpp) 

Windows builds fail.

jpfeuffer:
Either it is because of VS14 in the nightlies or the unity builds. In any case, would be good to fix it.

cbielow:
you mean http://cdash.openms.de/viewBuildError.php?buildid=240107 ?
I tried unity build with VS 14 (2015) and it works nicely on my machine.
However, I have `C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0` .. so yours is more recent.
That might be the difference.
Any chance to access the build machine directly?

PR #2900, opened 2017-09-08
[FIX] pyOpenMS macOS fixes

grosenberger:
Updated PR against master for https://github.com/OpenMS/OpenMS/pull/2821.


src/pyOpenMS/create_cpp_extension.py
```diff
@@ -283,8 +283,9 @@ def doCythonCompile(arg):
     # copy and set correct permissions (needs to be writeable for next copy)
     shutil.copy(j(QT_LIBRARY_DIR, "QtCore.framework", "QtCore"), "pyopenms")
     shutil.copy(j(QT_LIBRARY_DIR, "QtNetwork.framework", "QtNetwork"), "pyopenms")
-    os.chmod("pyopenms/QtCore", 0o744)
-    os.chmod("pyopenms/QtNetwork", 0o744)
+    os.chmod("pyopenms/QtCore", 0744)
+    os.chmod("pyopenms/QtNetwork", 0744)
```
hroest: can you compare with #2891 -- apparently Python3 likes `0o744` better?

grosenberger: Good point, I correct this back to `0o744 `.

Issue #2899, opened 2017-09-08
XMLHandler's StringManager needs to go!


cbielow:
just documenting this here; implementation will probably take 1-2 days, but could be done piece-by-piece.

while looking at the read/write performance of our XML files, I stumbled upon our long-forgotten `StringManager` class which is part of the XMLHandler and therefore affects all our Xerces-based xml reading. In short, it takes ownership of every string which is passed from xerces to our internal datastructures (usually from XMLCh to char*).

For a common idXML this means the following number of pointers (to heap allocated characters) are created
xml_strings: #14086589  c_strings: #21363930
this alone amounts to about 280 MB, not counting the actual data being pointed to (which is probably about 5x as big). Note: only the last few entries in these arrays are used, and only once!
This wasteful style also explains why we need 2300 MB Peak Memory to read in a 200 MB idXML file.

Not storing the array pointers (effectively creating a memory leak for the moment), saves 10% of loading time and the mentioned 280 MB RAM.

The proper solution IMHO would be to remove the StringManager completely and use unique_ptr with a custom deleter (which calls `xerces::XMLString::release(ptr)`). In in the mentioned example, this would mean 10% faster, and approx 1/2 the RAM for loading.

cbielow:
another workaround: using the vectors as before but as a ring-buffer of fixed size (here 100), this is what happens to the Mem & runtime:
```
FileInfo took 31.06 s (wall), 28.19 s (CPU), 0.86 s (system), 27.33 s (user).
Peak Memory Usage: 2412 MB
```
after:
```
FileInfo took 24.28 s (wall), 24.18 s (CPU), 0.56 s (system), 23.62 s (user).
Peak Memory Usage: 1244 MB
```

That's **30% speed increase and 50% memory decrease**. We can expect similar improvements for many XML formats (when reading). PR coming up.

timosachsenberg:
Nice find!
I would probably prefer automated memory management with unique_ptr instead of the ring buffer. 
Something in the line of:
```
    auto xerces_deleter = [](XMLCh * a) { xerces::XMLString::release(a); };
    std::unique_ptr<XMLCh, decltype(lambda)> f(XMLString::transcode(str), xerces_deleter);
```

cbielow:
Agreed. But this change was doable in 10min. The other one takes more time to implement but should also be a bit faster due to hot caches.

-- 
Sent via mobile phone.

[...]

PR #2901, opened 2017-09-09
[FIX]!!! huge speed and memory improvement by not storing all xerces …

cbielow:
…transcoded strings but using a ringbuffer instead.

For speed & RAM improvements, see #2899.

Not thouroughly tested yet.
Assumes that during loading, the format-specific xml-handlers to not store pointers to transcoded strings while at the same time keep on calling convert() more than 99x (the ringbuffer will free the data pointed to).

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
+++ b/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
@@ -102,6 +102,8 @@ public:
 private:
       mutable std::vector<XMLCh *> xml_strings_;
       mutable std::vector<char *> c_strings_;
+      mutable size_t idx_xml_, idx_c_;
+
     };
 
     /**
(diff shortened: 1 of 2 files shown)


hroest:
very nice find performance analysis, but this seems pretty dangerous to merge. I think this is a nice proof of concept to show how much speed/memory one can gain but should never be merged.

I think the ring buffer is one of the worst interfaces possible - especially in a call-back situation as in xercesc. How would you possibly keep track of how many strings you have converted since the last time you accessed a string - it sounds like a nightmare to not make errors and has the worst of two worlds regarding ownership. Basically with this approach, ownership stays with the StringManager but only for limited time over which we have almost no control, so now there are basically no more guarantees... IMHO there are only two ways to do this

- tie ownership to lifetime of `StringManager` (and reset the StringManager from the caller side from time to time when you know you wont need any strings any more)
- transfer ownership to caller using `unique_ptr` and `std::move` operations

So in my opinion the correct way to go about this is to add another fxn to StringManger that transfers ownership as outlined here: https://github.com/OpenMS/OpenMS/issues/2899#issuecomment-328363047

hroest:
@cbielow I also see that `sm_.clear();` gets never called in `IdXMLFile.cpp` or `MzIdentMLHandler.cpp`, can you check what happens when you add that call after parsing an element completely? That would go towards my first proposed solution and may already solve some of the memory issues...

PR #2902, opened 2017-09-11
Feature/sm memory management

hroest:
change StringManager memory management

- take ownership away from StringManager
- only use safe containers, either `std::string` or `std::basic_string<XMLChar>`

see also C++ cookbook https://resources.oreilly.com/examples/9780596007614/blob/master/14-4.h

replaces #2901

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
+++ b/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
@@ -35,8 +35,6 @@
 #ifndef OPENMS_FORMAT_HANDLERS_XMLHANDLER_H
 #define OPENMS_FORMAT_HANDLERS_XMLHANDLER_H
 
-#include <iosfwd>
-
 #include <OpenMS/CONCEPT/Types.h>
 #include <OpenMS/CONCEPT/Macros.h>
 
@@ -45,11 +43,16 @@
 #include <OpenMS/DATASTRUCTURES/ListUtils.h>
 #include <OpenMS/METADATA/MetaInfoInterface.h>
 
+#include <boost/scoped_array.hpp>
+
+#include <xercesc/util/XMLString.hpp>
 #include <xercesc/sax2/DefaultHandler.hpp>
 #include <xercesc/sax/Locator.hpp>
 #include <xercesc/sax2/Attributes.hpp>
 
 #include <algorithm>
+#include <iosfwd>
+#include <string>
 
 namespace OpenMS
 {
@@ -69,6 +72,36 @@ namespace OpenMS
     */
     class OPENMS_DLLAPI StringManager
     {
+
+      typedef std::basic_string<XMLCh> XercesString;
+
+      // Converts from a narrow-character string to a wide-character string.
+      inline XercesString fromNative(const char* str) const
+      {
+          boost::scoped_array<XMLCh> ptr(xercesc::XMLString::transcode(str));
+          return XercesString(ptr.get( ));
+      }
+
+      // Converts from a narrow-character string to a wide-charactr string.
+      inline XercesString fromNative(const std::string& str) const
+      {
+          return fromNative(str.c_str( ));
+      }
+
+      // Converts from a wide-character string to a narrow-character string.
+      inline std::string toNative(const XMLCh* str) const
+      {
+          boost::scoped_array<char> ptr(xercesc::XMLString::transcode(str));
+          return std::string(ptr.get( ));
+      }
+
+      // Converts from a wide-character string to a narrow-character string.
+      inline std::string toNative(const XercesString& str) const
+      {
+          return toNative(str.c_str( ));
+      }
+
+
 public:
       /// Constructor
       StringManager();
@@ -79,17 +112,29 @@ public:
       /// Frees memory of all owned strings
       void clear();
 
-      /// Transcode the supplied C string to XMLCh* and take ownership of the XMLCh*
-      XMLCh * convert(const char * str) const;
+      /// Transcode the supplied C string to XMLCh*
+      XercesString convert(const char * str) const
+      {
+        return fromNative(str);
+      }
 
-      /// Transcode the supplied C++ string to XMLCh* and take ownership of the XMLCh*
-      XMLCh * convert(const std::string & str) const;
+      /// Transcode the supplied C++ string to XMLCh*
+      XercesString convert(const std::string & str) const
+      {
+        return fromNative(str.c_str( ));
+      }
 
-      /// Transcode the supplied OpenMS string to XMLCh* and take ownership of the XMLCh*
-      XMLCh * convert(const String & str) const;
+      /// Transcode the supplied OpenMS string to XMLCh*
+      XercesString convert(const String & str) const
+      {
+        return fromNative(str.c_str( ));
+      }
 
-      /// Transcode the supplied XMLCh* to a C string and take ownership of the C string
-      char * convert(const XMLCh * str) const;
+      /// Transcode the supplied XMLCh* to a String
+      String convert(const XMLCh * str) const
+      {
+        return toNative(str);
+      }
 
       /**
        * @brief Transcodes the supplied XMLCh* and appends it to the OpenMS String
@@ -100,6 +145,7 @@ public:
       static void appendASCII(const XMLCh * str, const XMLSize_t length, String & result);
 
 private:
+
       mutable std::vector<XMLCh *> xml_strings_;
       mutable std::vector<char *> c_strings_;
     };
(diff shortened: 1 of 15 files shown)


src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
```diff
@@ -69,27 +72,69 @@ namespace OpenMS
...
     class OPENMS_DLLAPI StringManager
     {
+
+      typedef std::basic_string<XMLCh> XercesString;
+
+      // Converts from a narrow-character string to a wide-character string.
+      inline XercesString fromNative(const char* str) const
+      {
+        boost::scoped_array<XMLCh> ptr(xercesc::XMLString::transcode(str));
+        return XercesString(ptr.get( ));
```
hroest: while this is taken from the cookbook, I am not sure whether this is really 100% safe. xerces is quite specific about releasing the memory and by using a `boost::scoped_array<XMLCh>`, we basically do our own memory management here and never tell xerces that we released this string. What if the xerces memory manager tries to do something "smart" like allocating continous memory blocks and we never tell it that  a certain memory area has been released? I think it would be safer to use

```
XMLCh* ptr(xercesc::XMLString::transcode(str));
XercesString result(ptr.get( ));
XMLString::release(ptr);
return result;
```

any thoughts? I am just a bit uncomfortable with this...

timosachsenberg:
test needs fixing

/.../OpenMS/src/tests/class_tests/openms/source/CompressedInputSource_test.cpp:73:13: error: no matching constructor for initialization of 'OpenMS::CompressedInputSource'
   ptr = new CompressedInputSource(Internal::StringManager().convert(filename.c_str()),bz);
            ^                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/.../OpenMS/src/openms/include/OpenMS/FORMAT/CompressedInputSource.h:52:5: note: candidate constructor not viable: no known conversion from 'XercesString' (aka 'basic_string<unsigned short>') to 'const OpenMS::String' for 1st argument
    CompressedInputSource(const   String & file_path, const String & header, xercesc::MemoryManager * const manager = xercesc::XMLPlatformUtils::fgMemoryManager);
    ^
/.../OpenMS/src/openms/include/OpenMS/FORMAT/CompressedInputSource.h:54:5: note: candidate constructor not viable: no known conversion from 'XercesString' (aka 'basic_string<unsigned short>') to 'const XMLCh *const' (aka 'const unsigned short *const') for 1st argument
    CompressedInputSource(const   XMLCh * const file_path, const String & header, xercesc::MemoryManager * const manager = xercesc::XMLPlatformUtils::fgMemoryManager);
    ^
/.../OpenMS/src/openms/include/OpenMS/FORMAT/CompressedInputSource.h:68:5: note: candidate constructor not viable: requires single argument 'source', but 2 arguments were provided
    CompressedInputSource(const CompressedInputSource & source);

timosachsenberg:
should be:
ptr = new CompressedInputSource(Internal::StringManager().convert(filename.c_str()).c_str(),bz);
will fix

src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
```diff
@@ -69,27 +72,73 @@ namespace OpenMS
...
       /// Constructor
       StringManager();
 
-      /// Destructor. Destroys the strings in the various lists
+      /// Destructor
       ~StringManager();
 
-      /// Frees memory of all owned strings
+      /// NOP
       void clear();
```
timosachsenberg: ok this would then need to go I guess

timosachsenberg:
we also need to update the documentation of StringManager

timosachsenberg:
and finally, it would be great to rename the StringManager to XMLStringConversion

cbielow:
nice PR!
Its almost as fast as #2901, with:
```FileInfo took 25.89 s (wall), 25.86 s (CPU), 0.50 s (system), 25.37 s (user).```
on that test file.

The reason I did not go for conversion to std::string in my 'dirty' PR was that it puts additional stress on the allocator (due to another copy when using transcode as Hannes mentioned already) -- I had a modified version of the idXML parser which uses strcmp() on char* directly in a few places and is another 10-15% faster than what you see here.
But the removing the push_back() from SM is already a big step forward, I think we all agree. So lets go ahead with this PR and we can optimize even more later on, e.g. using @hroest's approach of 'in-place' conversion for ASCII-strings.

File: src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h [hunk 3]
```diff
@@ -45,11 +43,16 @@
 #include <OpenMS/DATASTRUCTURES/ListUtils.h>
 #include <OpenMS/METADATA/MetaInfoInterface.h>
 
+#include <boost/scoped_array.hpp>
```
cbielow: do we need these includes?

hroest:
@cbielow can you test how fast the following is:

```
```

hroest:
OpenMS::String convertStringASCII(const XMLCh * str) const
{
  String result;
  // safer than strlen
  size_t xml_len = 0;
  for (const XMLCh* it = str; (char)*it != '\0'; it++, xml_len++) {}
  result.resize(xml_len); 
  std::string::iterator str_it = result.begin();
  for (const XMLCh* it = str; (char)*it != '\0'; it++)
  {
    *str_it = (char)*it;
    ++str_it;
  }

  return result;
}

cbielow:
you mean when as a replacement in
```
inline std::string toNative(const XMLCh* str) const
      {
        std::string result;
        // safer than strlen
        size_t xml_len = 0;
        for (const XMLCh* it = str; (char)*it != '\0'; ++it, ++xml_len) {}
        result.resize(xml_len); 
        std::string::iterator str_it = result.begin();
        for (const XMLCh* it = str; (char)*it != '\0'; ++it)
        {
          *str_it = (char)*it;
          ++str_it;
        }

        return result;
      }
```?

well, it's faster.
```
FileInfo took 23.24 s (wall), 23.24 s (CPU), 0.53 s (system), 22.71 s (user).
```

I guess its safe to do this, since our formats do not contain fancy unicode? And even if they did, we'd store them in narrow std::string anyway, thus not supporting wide chars. But what about filenames references inside parameter sections etc?
Is `startElement()` and `endElement()` then safe to use the code above. We'd just to be careful with attributes?

**Comment by hroest** [comment]
well I would have two functions

```
OpenMS::String convertStringASCII(const XMLCh * str) const;
OpenMS::String convert(const XMLCh * str) const;
```
one for when you know its ASCII (from your xsd) and one for all other,  user-generated cases (e.g. where the user could write anything, such as free attributes, filenames etc). The problem with the ASCII fxn is that it will terminate early if it finds an early null terminator and will only consider the lower 8 bits of any 16bit code point (which is fine for ASCII) but still expanding two 16bit code points into two characters. Basically if there is unicode in there, the information is lost and the content garbled. So you need to be *sure* that there is only pure ASCII and we only be sure for XML attributes we control. Even the obo terms may not be ASCII (maybe @mwalzer knows?). So unfortunately we cannot use this trick everywhere

**Comment by timosachsenberg** [comment]
tests passed - ready for merge?

**Comment by cbielow** [comment]
IMHO yes

File: src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h [hunk 4]
```diff
@@ -45,51 +43,95 @@
...
+     * std::basic_string<XMLCh> to the caller who takes ownership of the data.
      *
     */
     class OPENMS_DLLAPI StringManager
     {
+
+      typedef std::basic_string<XMLCh> XercesString;
+
+      // Converts from a narrow-character string to a wide-character string.
+      inline XercesString fromNative(const char* str) const
```
cbielow: well... technically, these should be fromNative_, to adhere to OpenMS coding conventions...
/nitpick

src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
```diff
@@ -45,51 +43,95 @@
...
+      }
+
+      // Converts from a narrow-character string to a wide-charactr string.
+      inline XercesString fromNative(const std::string& str) const
+      {
+        return fromNative(str.c_str());
+      }
+
+      // Converts from a wide-character string to a narrow-character string.
+      inline std::string toNative(const XMLCh* str) const
```
cbielow: should we use OpenMS::String for consistency here?
Might RVO have trouble if its not the same type (std::string vs. OpenMS::String)?

timosachsenberg: Yes, I think you are right. They are different types so copy elision does not work.

timosachsenberg:
Ok I think everything is addressed from this PR and it gives already a nice speedup - so I merge

timosachsenberg:
fyi: on a large mzML (1e6 MS2 spectra) loading times are reduced from 7:30 to 7:02 minutes

cbielow:
dominated by base64-data I guess?
idXML has a lot of tags, and not much data in it.. that could be the reason for the disparity

hroest:
I agree, in my experience the base64 data conversion does dominate the CPU time. @timosachsenberg what happens when you skip that step and use `setFillData(false)` on the peak option file to not actually load any data? I guess there the improvement should be much more visible and it would tell you how much is dominated by base64 processing.

PR #2906, opened 2017-09-11
[BUILD] fix travis

hroest:
seems like we were missing omp after the travis update: https://blog.travis-ci.com/2017-08-29-trusty-image-updates


jpfeuffer:
Saw this, too, today. The reason is the upgrade from clang 3.5 to 3.9 which now supports OpenMP (but only with an additional installed library). Since OpenMP support defaults to ON, our CMake finds the OpenMP support but cannot link it later (it seems to compile though).
It might be that you have to install libomp (LLVM clang) instead of libgomp (gcc) though.

hroest:
well it looks like libomp is only available from 16.04 onwards: https://packages.ubuntu.com/search?keywords=libomp5

hroest:
PS: also annoying that travis seems to randomly select the old and new images, so be careful when testing that really the new image was used: 

```
Build image provisioning date and time
Tue Aug 29 02:50:02 UTC 2017
```

hroest:
according to this: https://github.com/ehein6/stinger/commit/17721a819f798001dd28407b82fefc0fe390e44e one idea is to pull in `debian-sid` to get the libomp-dev package but that seems to cause a whole bunch of other problems. Not really sure what to do here

jpfeuffer:
Since we did not use OpenMP for a long time with clang, I would be okay with disabling OpenMP support (using our CMake command line option) for clang until travis supports 16.04 or we find a better solution. If we disable support in the beginning, it should not try to link it later.

hroest:
and  I also just found out that debian-side for sure does not work: https://travis-ci.org/OpenMS/OpenMS/jobs/274325552

hroest:
we should probably ask over at https://github.com/travis-ci/travis-ci/issues/8315 how to address the issue with clang / OpenMP

hroest:
seems like commit 724aae7 now works
https://travis-ci.org/OpenMS/OpenMS/builds/274328408

hroest:
> Since we did not use OpenMP for a long time with clang, I would be okay with disabling OpenMP support (using our CMake command line option) for clang until travis supports 16.04 or we find a better solution. If we disable support in the beginning, it should not try to link it later.

seems to work and it seems to be the only working solution for now

hroest:
travis works now for the new image

https://travis-ci.org/OpenMS/OpenMS/builds/274337977?utm_source=github_status&utm_medium=notification

```
Build image provisioning date and time
Tue Aug 29 02:50:02 UTC 2017
```

I am in favor of merge @jpfeuffer @timosachsenberg

hroest:
@jpfeuffer maybe we can get clang to work with the intel openmp library? e.g. install `libiomp-dev` and then link against that?

jpfeuffer:
Yes it might work with libiomp from what I have read. I will try a small PR.

hroest:
@jpfeuffer @timosachsenberg maybe we should add this to master as well?

timosachsenberg:
ok will add

PR #2909, opened 2017-09-12
Test OpenMP with Intel's lib

jpfeuffer:


jpfeuffer:
Cmake needs to be updated to automatically detect the correct "-fopenmp=libiomp5" flag.
We probably need to set it manually to enable it.

hroest:
its not too hard to update cmake, see https://github.com/hroest/docker_files/blob/master/openms/dependencies/manylinux/Dockerfile#L3 but it will add time to the build. but why not try that and see what it does? On the other hand, we are not really testing OpenMP anyways with our tests, so maybe having it disabled does not really matter at this stage?

hroest:
do we still need this PR?

PR #2914, opened 2017-09-13
Update lnx-cibuild.before.sh

timosachsenberg:


hroest:
:+1: seems like master is building and working again, see https://travis-ci.org/OpenMS/OpenMS/jobs/274971068#L6 (except TOPP_PercolatorAdapter_1_out1)

PR #2915, opened 2017-09-13
[FIX] omit copies and score calculations

timosachsenberg:


cbielow:
I guess even the copying via getHits() can be avoided in many places here -- working with a reference will do -- but it might not lead to significant speedup.

so, +1

timosachsenberg:
hmm there seem to be some failing tests so I will probably need to look into this more closely...

timosachsenberg:
ok in order to get same results as before I kept the same logic. Just replaced the O(n*m) part. No its O(n+m)

PR #2917, opened 2017-09-13
[FIX, TEST] update Percolator test file

timosachsenberg:


timosachsenberg:
@Leon-Bichmann hmm still not correct. Will close.

PR #2919, opened 2017-09-14
Fix/hyper score

timosachsenberg:
fix logfactorial calculation of HyperScore


timosachsenberg:
@enetz now its in very close range to the "wrong" hyperscore (that, interestingly, performed a bit better but failed for small number of matches)

timosachsenberg:
my current explanation for the extra 1% of ids with the "wrong" implementation is that log scores of the dot product got negative. The reason why this happens is, that after normalization with Normalizer, all MS peaks are scaled relative to the largest MS2 peak (which is 1.0). If the dot product (basically the sum of intensities) does not exceed a sum of 1.0, then the log(dot_product) is negative. This seems to penalize bad spectra somehow and lead to a slightly better ID rate. I confirmed this 1% difference using more than 1e6 spectra.

PR #2924, opened 2017-09-22
[FIX] Find executables in renamed bin folder on Mac

jpfeuffer:

PR #2925, opened 2017-09-23
[Fix] find executables for toppas bundle macOS

jpfeuffer:
includes #2924 into master.


timosachsenberg:
+1

timosachsenberg:
can this be merged or is there a risk regarding our KNIME update site and the UM?

jpfeuffer:
I don't expect any problems since I only added a path to be looked up but I can't guarantee. We can wait, but until then, TOPPAS is broken in the Mac installer.

PR #2929, opened 2017-09-29
[FIX] SiriusAdapterPATH cmake knime binaries

oliveralka:
Could not find the executables and failed.
`FATAL: External invocation of Sirius failed`


jpfeuffer:
yes should be fine.

PR #2930, opened 2017-09-29
[FIX] In SiriusAdapter,  SIRIUS_PATH is now being used when set and executable is empty 

lukaszimmermann:
Fixes issue in SiriusAdapter that the SIRIUS_PATH variable has not been used in KNIME.


oliveralka:
looks good!

PR #2932, opened 2017-10-04
fix compile error on Win10/VS2015 (duplicate definition of snprintf)

cbielow:
augments #2898


jpfeuffer:
Interesting. But looks good except for timeouts on clang. Will merge.

PR #1334, opened 2015-04-02
Feature/pep xml prophets

hroest:
- improve pepXML parsing
- add support for the `search_score_summary` tag
- add support to store interProphet *and* peptideProphet results
- support for additional attributes in `spectrum_query`
- support to write out pepxml attributes

PR #2935, opened 2017-10-05
General-purpose ID data structures

hendrikweisser:
This is finally a concrete attempt to introduce general-purpose data structures for identification results. The goal is to support identified proteins/peptides and small molecules (and, soon, RNAs/oligonucleotides) in the same structure. Other goals include:
- having one self-contained data structure for all ID results,
- avoiding redundancy (and the discrepancies that result from it),
- preferring "flat" structures over "nested" ones,
- supporting different kinds of identification (e.g. database search of MS2 spectra, accurate mass search of features),
- allowing different scores side-by-side,
- keeping track of data processing steps.

This is work in progress, still far from the point of mergeability, so this PR is primarily a request for feedback.

Some general remarks:
- Data items that should be stored non-redundantly (e.g. peptides, proteins, search queries, score types, ...) are associated with unique keys in bidirectional maps (`boost::bimap`). The keys are used to reference these items in other data structures. (The design is thus somewhat similar to a relational database.)
- The idea would be to include a full `IdentificationData` instance in feature/consensus maps and have references to `DataQuery` instances (via their unique keys) in the features.
- Conversion to/from `ProteinIdentification` and `PeptideIdentification` is not straight-forward, mainly because those classes store information redundantly (e.g. in protein and peptide hits), but cannot guarantee consistency. (For example, the same peptide may be associated with different proteins in different entries.)

What obvious things are still missing at the moment?
- convenience functions,
- support for protein inference,
- integration with the rest of OpenMS.


src/openms/include/OpenMS/METADATA/IdentificationData.h
```diff
@@ -0,0 +1,320 @@
...
+
+
+    /*!
+      Meta data for a search hit (e.g. peptide-spectrum match).
+    */
+    struct MatchMetaData: public MetaInfoInterface
+    {
+      Int charge;
+
+      std::unordered_map<ScoreTypeKey, double> scores;
```
timosachsenberg: probably need a way to define primary score

src/openms/include/OpenMS/METADATA/IdentificationData.h
```diff
@@ -0,0 +1,320 @@
...
+
+      // is it useful to store this, as different processing steps/score types
+      // may give different rankings?
+      Size rank; // rank among matches for the same spectrum/feature
```
timosachsenberg: I would love to get rid of it but this might be some work. a rank could always be modelled as score type

src/openms/source/METADATA/IdentificationData.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+using namespace std;
+
+namespace OpenMS
+{
+  void IdentificationData::importIDs(
+    const vector<ProteinIdentification>& proteins,
+    const vector<PeptideIdentification>& peptides)
+  {
+    map<String, ProcessingStepKey> id_to_step;
```
timosachsenberg: is this id or accession?

timosachsenberg:
Hi and thanks for taking a look at this long standing issue. Do you have a graphical representation (would save me drawing my own ;)

Are the pepXML subscores also supported? I recall that they added a bit of complexity

timosachsenberg:
- storing fragment annotations of hits (see PeptideHit)

src/openms/source/METADATA/IdentificationData.cpp
hendrikweisser: The "id" is the identifier that is used to link PeptideIdentifications to a ProteinIdentification.

hendrikweisser:
> Do you have a graphical representation (would save me drawing my own ;)

No, but I like the idea. Do you have experience with any tool that automatically generate a UML diagram from C++ code? (Apparently Doxygen can do it - see [here](https://www.stack.nl/~dimitri/doxygen/manual/diagrams.html) and the example [here](https://www.stack.nl/~dimitri/doxygen/manual/examples/diagrams/html/class_c.html) - but I haven't tried it.)

> Are the pepXML subscores also supported? I recall that they added a bit of complexity

I haven't looked into it, but in principle any kind of score should be supported now.

src/openms/include/OpenMS/METADATA/IdentificationData.h
hendrikweisser: This is an important point, but unfortunately it's - probably? - not enough to just designate one score type as the primary one (since you may have different types of molecules, e.g. peptides vs. compounds, and different levels, e.g. PSMs, peptides, proteins). You could store a different primary score for each molecule/level/..., but maybe there's a better solution? I haven't thought it through yet.

hendrikweisser:
> storing fragment annotations of hits (see PeptideHit)

This should be included now.

src/openms/include/OpenMS/METADATA/IdentificationData.h
hendrikweisser: I think I've found a good (and quite simple) solution: Replace all the `unordered_map<ScoreTypeKey, double>` with `vector<pair<ScoreTypeKey, double>>`. Then the order in the vector would give a precedence, with the last entry being the "primary score". This would work quite naturally as sequential processing steps append their scores to the list, so that new scores would take precedence over old ones.
Trade-offs:
- less efficient look-up of scores by score type (but the list shouldn't get very long anyway),
- no guarantee that entries are unique by score type (but that doesn't seem very important and we'd still have a clearly defined order of precedence).

src/openms/source/METADATA/IdentificationData.cpp
```diff
@@ -0,0 +1,341 @@
...
+            prot_data[*step_it].first.push_back(hit);
+            prot_data[*step_it].second = score_it->first;
+            steps.insert(*step_it);
+            break;
+          }
+        }
+      }
+    }
+
+    for (set<ProcessingStepKey>::const_iterator step_it = steps.begin();
```
timosachsenberg: consider using range iteration:
 for (auto const & s : steps) 
it makes refactoring easier (and is also easier to read)

timosachsenberg:
regarding the PepXML subscores, I was referring to:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/METADATA/PeptideHit.h#L189

src/openms/source/METADATA/IdentificationData.cpp
hendrikweisser: I'm not used to that yet. ;-)

hendrikweisser:
> regarding the PepXML subscores, I was referring to:

Do you have an example file where those scores are used?

I think they could be represented as follows in the current design:
- Create a `DataProcessingStep` (with corresponding `DataProcessingParameters`) for the PeptideProphet analysis.
- Create a `ScoreType` (referencing the `DataProcessingParameters`) for each type of score.
- Add the scores to the list in each `MatchMetaData` (PSM) instance, making sure the "main score" goes last.

timosachsenberg:
I think this is the relevant PR : https://github.com/OpenMS/OpenMS/pull/1334

enetz:
```
   // for protein groups / compound adduct groups
    struct ParentMoleculeGroupMetaData: public MetaInfoInterface
    {
      enum MoleculeType molecule_type;

      ScoreList scores;

      CvTerm group_type;

      // ordered list of references to data processing steps:
      std::vector<ProcessingStepKey> processing_steps;
};

// TODO: is there a multibimap? We want to map groups to e.g. proteins. But also know in which groups a protein is present.
bimap<GroupMetaDataKey, ParentMetaDataKey> grouping;


```

enetz:
how to model cross-links (protein-protein, protein-RNA / DNA or other small-molecules?
```
   // for cross-links
    enum MoleculeType
    {
      MT_PROTEIN,
      MT_COMPOUND,
      // MT_RNA,
      MT_CROSSLINK ?????????
      SIZE_OF_MOLECULETYPES
    };

    struct CrossLinkMetaData: public MetaInfoInterface
    {
      std::pair<UniqueKey> cross_link_elements; // of MoleculeType
      std::pair<size_t> cross_link_position; // of MoleculeType

      CvTerm cross_link_type;
      
      //  3 cases - can we model this already?
      // uncleavable label-free: 2 peptides -> 1 spectrum
      // uncleavable labeled: 2 peptides -> 2 spectra (where peptides and spectra don't have a 1 to 1 matching. (heavy / light))
      // cleavable label-free: 2 peptide -> 1 MS2 -> 2 MS3 (both peptides in the MS/MS and only one for each of the MS3)

      ScoreList scores;

      // ordered list of references to data processing steps:
      std::vector<ProcessingStepKey> processing_steps;
};

    std::unordered_map<IdentifiedMoleculeKey,
                       CrossLinkMetaData> crosslink_meta_data;

```

hroest:
I have some general questions

- what is the memory footprint of the new new structures, are they more efficient than the old ones?
- it seems that you are using a lot singletons (maps to map one species to another), how does this work with multi-threading? I see a potential issue here, especially due to lack of encapsulation.

hendrikweisser:
@enetz: For protein groups I was thinking something like this:
```
struct ParentMoleculeGroup: public MetaInfoInterface
{
  std::unordered_set<ParentMoleculeKey> members;

  ScoreList scores;

  std::vector<ProcessingStepKey> processing_steps;
};

std::vector<ParentMoleculeGroup> parent_molecule_groups;
```

Since those groups aren't referenced anywhere else, they don't need their own key.
`molecule_type` shouldn't be necessary because that can be looked up via the members.
Are there CV terms for different protein group types?

Open questions that impinge on whether a different structure may be better:
- Is there a need to support multiple groupings (e.g. from different protein inference engines) at the same time and easily tell them apart? (In the proposed structure above, that would only be possible by looking at the processing steps.)
- Should we have a constraint that each protein can only be in one group (per inference set)?
- Are look-ups more likely to be by protein (to find the group) rather than group (to find the proteins)?

@jpfeuffer: You're working on protein inference. Any thoughts on what data structure(s) would be convenient?

hendrikweisser:
@netz: Regarding cross-links, I like your proposal. Just to make sure that I understood it correctly: The cross-link ID would be stored as an "identified molecule", so it can be linked to a spectrum ("data query") via a "molecule-query match". And in the meta data of that cross-link ID (`CrossLinkMetaData`), you would reference the two associated peptides (or whatever) by their keys. Those peptides would be stored in the normal way as "identified peptides".

To answer your questions from the comments:
- Yes, for consistency I would allow multiple scores and processing steps also for cross-links.
- The "uncleavable label-free" case is the basic one that would be represented in the way I described above.
- The "uncleavable labeled" case would work in the same way - you'd get two independent "molecule-query matches", one for light, one for heavy. (Should `CrossLinkMetaData` get a member to store heavy/light information or is that represented elsewhere, e.g. on the peptides as modifications?)
- The "cleavable label-free" case is interesting and I don't think I fully understand it yet. The MS3 spectra and their identified peptides probably don't need any special treatment - just normal "molecule-query matches" to the peptides. Could the MS2 be treated in the same way as "uncleavable label-free"?

This last case actually makes me think of multiplexed peptide identifications, where two (or more?) peptides are identified from the same (mixed) spectrum. This is something that I haven't considered in the design so far. Perhaps cross-links can be treated as a special case of this?
If so, should we generalize `CrossLinksMetaData` to `MultiplexedMetaData` and allow an arbitrary number of molecules (e.g. peptides) to be identified from one spectrum at the same time?

timosachsenberg:
- Should we have a constraint that each protein can only be in one group (per inference set)?
no they can be part of multiple sets (e.g., FIDO outputs these if no greedy resolution is enabled)

hendrikweisser:
@hroest:
> what is the memory footprint of the new new structures, are they more efficient than the old ones?

That's something we need to measure. How do I do that - just "sizeof"?

Some more thoughts on this topic:
- There are areas where we could easily optimize the class for space if we're okay with the trade-offs, e.g. replace the "unordered_..." data structures with ordered ones, or remove the references to data processing steps in individual data items.
- It's an apples/oranges comparison, because the new structures store things the old ones couldn't (e.g. compounds). On the other hand, the old structures can be highly redundant. So an empty or almost-empty `vector<PeptideIdentification>`/`vector<ProteinIdentification>` will surely be smaller than a corresponding `IdentificationData`. With multiple search runs, and depending on whether you include e.g. protein sequences, it'll be the other way around.
- Efficiency is not the reason why we have to replace the old structures. We have to make new structures to support the extended capabilities we need. We should of course aim to make these new structures as efficient as possible, but in my opinion the memory footprint of the old ones shouldn't be a constraint.

> it seems that you are using a lot singletons (maps to map one species to another), how does this work with multi-threading? I see a potential issue here, especially due to lack of encapsulation.

I haven't considered any of that yet, so there are most probably issues, but I'm not sure which one you're referring to. Can you explain a bit more?

hendrikweisser:
@timosachsenberg:
> - Should we generalize CrossLinkMetaData to MultiplexedMetaData
>
> I think these might be different concepts. In cross-linking you also want to know the cross-linked positions. 

If the only difference is needing to store the positions, maybe using meta values would be enough?

> The method determines in how many spectra these are measured

I don't understand that.

enetz:
@hendrikweisser the main complication with some of the cross-linking methods, is that several spectra are pooled together / merged in some way to produce one ID for a cross-link with one score.
In the "uncleavable labeled" case this is still quite simple, two MS2 spectra (from light and heavy precursors) are compared and sets of peaks are extracted to essentially generate one new temporary spectrum that is used for the ID.
Mapping that result back to the input spectra means that one ID with one score is associated with two spectra.
We could represent that by duplicating the ID and attaching it to the two spectra separately, but we still need to encode the association between the two spectra somehow and it adds a lot of redundancy.

For "cleavable label-free" there are some methods that use 3 or even 5 spectra for one ID and one score
(One CID MS2 to cleave the linker instead of the peptides, then use peptide precursors from this MS2 spectrum to sequence the peptides separately in MS3, sometimes using multiple fragmentation methods, e.g. HCD and ETD).
In this case we could treat the MS3 as separate peptide IDs, but what we care about is the combination of this info into a cross-link with two sequences and positions and the whole cross-link gets one score in the end. Also the MS2 spectrum contains additional fragments and can be included in the final score as well and here both peptides are contained in one spectrum again.

One way to cover all these cases would be a way to associate multiple spectra to each ID
(the opposite of your proposed idea of several IDs per spectrum)
and to have a way to attach ID dependent MetaInfo to each of those spectra
(e.g. light / heavy, peptide precursors / alpha peptide / beta peptide)
This would also be easy to extend to additional methodologies that combine spectra for single IDs in some way.

I don't know how to do this smartly. If we just use a list of spectrum references for each ID by default, then in by far the most cases we would have a list of one.

hendrikweisser:
@enetz: Thanks for the explanations.

Currently, there's a one-to-one relationship between queries (spectra) and IDs (peptides) in `query_matches` (although there can be several alternative IDs from the same query). We could of course change that to a many-to-many relationship to cover all possibilities. However, that would introduce overhead that probably isn't needed in >90% of cases. Like you said:

> If we just use a list of spectrum references for each ID by default, then in by far the most cases we would have a list of one.

So I'm wondering if there isn't another solution as well.

> have a way to attach ID dependent MetaInfo to each of those spectra (e.g. light / heavy, peptide precursors / alpha peptide / beta peptide)

This sounds to me like you *are* actually assigning different molecules to individual spectra - even if you needed to consider several spectra together to make the assignments. And it does make sense, from a physical point of view, in that the instrument tries to isolate a precursor of one analyte species and fragment that (at least in DDA). So after you've done the identification, you should be able to assign the molecule(s) that generated the precursor.

Since you understand this topic better than I do, can you check whether for all the cross-linking scenarios, you would be able to make assignments to individual spectra in the end? Because in that case, I think it would be better to keep the current structure and just add a way of grouping related spectra or PSMs together (which could be ignored in >90% of cases).

hroest:
@hendrikweisser @enetz this sounds like something that should be hidden behind an interface, e.g. there should be interface to query the peptides associated with spectra and one could use a more "expensive" mapping if necessary but it should be hidden from the user. 

We also have to think about that there is a mapping of precursor -> spectra, a mapping of id -> precursor and id -> spectra.

Also in DDA there are search engines that identify multiple peptides per spectrum, of course in DIA that is the usual case.

hendrikweisser:
@hroest:
> there should be interface to query the peptides associated with spectra and one could use a more "expensive" mapping if necessary but it should be hidden from the user

Fair point, but we can provide convenient interfaces independent of the internal data structures. But we still want internal structures that make natural/common operations easy to perform.

> We also have to think about that there is a mapping of precursor -> spectra, a mapping of id -> precursor and id -> spectra.

I'm not sure I understand that. You only have a precursor if you have a (fragment ion) spectrum, so how is "id -> precursor" not equivalent to "id -> spectrum"?
I wasn't planning to include the "precursor -> spectrum" mapping in the ID data, since that's already in the raw data (referenced via file key and identifier, e.g. spectrum native ID, in `DataQuery`).

> Also in DDA there are search engines that identify multiple peptides per spectrum, of course in DIA that is the usual case.

Yes, but that would be relatively easy to cover by adding a "Multiplex" molecule type (since we already have different molecule types) that contains references to multiple molecules. In that way, multiplexed identifications would be treated as a special case.
The important question, both for multiple spectra and multiple IDs, is whether it's better to treat them as special cases separate from the norm (one spectrum/one ID), or as the general case which includes the norm.

hendrikweisser:
I've started to work on exporting the contents of `IdentificationData` to mzTab. Given that we eventually want to replace idXML with mzTab (our own variant that includes support for RNA), it would be good to have internal structures that are compatible with mzTab wherever possible. In many cases, mzTab supports a lot more meta data than `IdentificationData` currently does. So I'm thinking the easiest solution to capture this information may be to straight-up include relevant structures from the `MzTab` class. For example, have an `MzTabMetaData` at the top level, have an `MzTabProteinSectionRow` for every protein (problem: we don't have proteins, we have general "parent molecules"), etc.

Thoughts?

timosachsenberg:
Not sure if I understand correctly. The ProteinSectionRow captures a lot more e.g. the quantitative values. I don't know if we want to include that here.

timosachsenberg:
also the MzTab meta data section might not fit too well (e.g. type / flavor of MzTab file is included) and I would rather cherry-pick the meta data entries

timosachsenberg:
but yes, in general I think we can use most meta data entries

hendrikweisser:
@timosachsenberg: Thanks for sharing your perspective. I guess I don't like the prospect of having to code up all of the meta data items from mzTab again in `IdentificationData`. That is a lot of meta data!

There are some other problems that lead to incompatibilities between the two structures:
- Fixed and variable modifications are stored once in the meta data section in mzTab. In `IdentificationData` different sets of fixed/variable modifcations can be stored in multiple entries for "DB search parameters". There's no restriction in `IdentificationData` that multiple searches have to use the same modification parameters, but it wouldn't be possible to represent this situation in mzTab.
- Database and taxonomy information is stored individually for every protein in mzTab, but there's no field for the protein sequence (although an optional column can be added, of course).
- mzTab has a peptide and a PSM section, but it only has one small molecule section. It would be possible to represent "small molecule-spectrum matches" there, but a lot of data would have to be repeated if the same molecule was identified multiple times.
- Similarly, the PSM section contains redundant data from the peptide section ("accession", "unique", "database", "database_version") and some fields that should be in the peptide section instead ("pre", "post", "start", "end").

All of this makes me doubt whether using mzTab to replace idXML is really feasible. Perhaps we should be thinking about an SQLite-based format that captures an object-relational mapping of `IdentificationData` again.

hendrikweisser:
@enetz: I'd still be very interested in your answer to this question (from [here](https://github.com/OpenMS/OpenMS/pull/2935#issuecomment-336831356)):
> Since you understand this topic better than I do, can you check whether for all the cross-linking scenarios, you would be able to make assignments to individual spectra in the end? Because in that case, I think it would be better to keep the current structure and just add a way of grouping related spectra or PSMs together (which could be ignored in >90% of cases).

By the way, I've implemented basic mzTab export for `IdentificationData` now, but it's in a different branch (together with RNA-related stuff) - see here:
https://github.com/hendrikweisser/OpenMS/blob/rna-ms2-id/src/openms/include/OpenMS/METADATA/IdentificationData.h
https://github.com/hendrikweisser/OpenMS/blob/rna-ms2-id/src/openms/source/METADATA/IdentificationData.cpp

enetz:
@hendrikweisser Yes, grouping PSMs should also work. So it's either grouping spectra under one PSM, or defining several PSMs and grouping these, although I think the latter would be more redundant, as those PSMs would have the same score in many cases and some other info would be duplicated as well. Depending on the way the grouping is implemented, maybe only a "master PSM" needs to have this info, but that would just add unnecessary complexity.

hendrikweisser:
After a bit of a hiatus I've had some time to look at this again. The "piece de resistance" has been how to represent PSMs in a way that works for the standard case (one spectrum -> one peptide), for multiplexed spectra (one spectrum -> multiple peptides) and for cross-linking (multiple spectrum-peptide pairs). One could imagine another use case of "merged spectra" (multiple spectra -> one peptide).

The current solution, which works only for the standard case, is roughly this structure:
`map<pair<PeptideKey, SpectrumKey>, PSMInfo>`
Where `PSMInfo` contains scores and meta data about the match.

As most of you have suggested, adding a separate structure on top of this to define groups may have more disadvantages than advantages. So my proposal now would be this:
`map<set<pair<SpectrumKey, PeptideKey>>, PSMInfo>`

This would keep one-to-one correspondences between spectra and peptides, but allow to have a set of those associated with each "identification event". This should work well for the standard case (where there would be only one element in the set) and cross-linking. It would require repeating keys (either for spectra or peptides) for the other two cases, but I don't see that as a big problem.

We should make sure that for the standard case, all PSMs from the same spectrum occur in sequence in the map, to help with the common operation "get the best PSM per spectrum". If that's not true by default, we should define a corresponding comparison operator.

Comments?

timosachsenberg:
Hi Hendrik. Thanks for taking another look at this.

```
After a bit of a hiatus I've had some time to look at this again. The "piece de resistance" has been how to represent PSMs in a way that works for the standard case (one spectrum -> one peptide), for multiplexed spectra (one spectrum -> multiple peptides) and for cross-linking (multiple spectrum-peptide pairs). One could imagine another use case of "merged spectra" (multiple spectra -> one peptide).
```
Nice summary, so basically its a n:m relationship that we want to model.

```
The current solution, which works only for the standard case, is roughly this structure:
map<pair<PeptideKey, SpectrumKey>, PSMInfo>
Where PSMInfo contains scores and meta data about the match.
As most of you have suggested, adding a separate structure on top of this to define groups may have more disadvantages than advantages.
```

Regarding the last sentence, I am not decided, yet. Need to think this through a bit.

```
So my proposal now would be this:
map<set<pair<SpectrumKey, PeptideKey>>, PSMInfo>
This would keep one-to-one correspondences between spectra and peptides, but allow to have a set of those associated with each "identification event". This should work well for the standard case (where there would be only one element in the set) and cross-linking. It would require repeating keys (either for spectra or peptides) for the other two cases, but I don't see that as a big problem.
```
Let me brainstorm a bit: Given this relation db like design, this is equivalent to a class design, where a PSMInfo object has a reference to a set of spectrum+peptide pairs. I guess the PSMInfo object would probably then need to store the type of relation (e.g., an enum SPECTRUM_PEPTIDE, SPECTRA_PEPTIDE, SPECTRUM_PEPTIDES, SPECTRA_PEPTIDES) as this is otherwise hard to retrieve from the set.

```
We should make sure that for the standard case, all PSMs from the same spectrum occur in sequence in the map, to help with the common operation "get the best PSM per spectrum". If that's not true by default, we should define a corresponding comparison operator.
```
I think this is possible.
My biggest concern is, that having that many maps and sets is rather inefficient in memory and performance-wise. Also, I have some concerns that updates might get difficult if this low-level representation is chosen. How would we e.g., remove a PeptideKey and update the set of pairs?

In that regards, it might also make sense to look at http://www.boost.org/doc/libs/1_62_0/libs/multi_index/doc/index.html
and some benchmarks on http://david-grs.github.io/why_boost_multi_index_container-part1/

hendrikweisser:
> Nice summary, so basically its a n:m relationship that we want to model.

It's a detail, but I don't think we would allow "general" n:m, i.e. `pair<set<SpectrumKey>, set<PeptideKey>>`, because that would lose the individual spectrum-peptide assignments in the cross-linking case (i.e. you wouldn't be able to tell which spectrum corresponds to heavy/light). The use cases that we have so far are 1:1, 1:n, n*(1:1), and n:1.

> Given this relation db like design, this is equivalent to a class design, where a PSMInfo object has a reference to a set of spectrum+peptide pairs.

There's the additional constraint that each "key combination" (e.g. `set<pair<SpectrumKey, PeptideKey>>` can only occur once, because we want to accumulate all related information in one `PSMInfo` object. E.g. if we have two search engines that identify the same PSM, I don't want two PSMs, but one PSM with two scores.

> I guess the PSMInfo object would probably then need to store the type of relation (e.g., an enum SPECTRUM_PEPTIDE, SPECTRA_PEPTIDE, SPECTRUM_PEPTIDES, SPECTRA_PEPTIDES) as this is otherwise hard to retrieve from the set.

This is an interesting point. I was hoping that an explicit annotation wouldn't be necessary - that if we handle the general case, it would work for all the specific (use) cases.
Ideally, algorithms that expect one-to-one relationships could still be applied gracefully. My mental example is visualization of peptide hits in TOPPView: This clearly works for the standard case. For multiplexed spectra, we get multiple annotations of the same spectrum, which is okay (depending on how we plot them). For merged spectra, we get the same annotation on multiple spectra, which is also okay (albeit we can't distinguish them from independent IDs). For labeled cross-linking data, we get the right light/heavy annotations on the corresponding spectra (but don't see that they belong together).
In other words, a "complex" PSM of the form `((spec1, pep1), (spec2, pep2), ...) -> info` could be treated as multiple PSMs with redundant meta data `(spec1, pep1) -> info`, `(spec2, pep2) -> info`, ... by algorithms that don't support the "complex" case.

> My biggest concern is, that having that many maps and sets is rather inefficient in memory and performance-wise.

One map with many sets, but it's a valid concern. I haven't come up with an alternative yet that avoids the `map<set<...>>` and also isn't much more complex.

> How would we e.g., remove a PeptideKey and update the set of pairs?

Very carefully. ;-)

hendrikweisser:
> So my proposal now would be this:
> `map<set<pair<SpectrumKey, PeptideKey>>, PSMInfo>`

This has some problems, actually, because `PSMInfo` includes charge and peak annotations (of fragment ions), both of which are specific to one `pair<SpectrumKey, PeptideKey>`.
Back to the drawing board...

hendrikweisser:
I'm closing this as it's hopelessly outdated. Development of the IdentificationData classes continued together with the work on an RNA search engine.

PR #2936, opened 2017-10-05
Refactoring of enzymatic digestion to support RNA digestion

hendrikweisser:
Added a new UTIL RNADigestor.

Turned `Enzyme` into an abstract base class, now called `DigestionEnzyme`, with new subclasses `DigestionEnzymeProtein` and `DigestionEnzymeRNA`.
Similarly, turned `EnzymesDB` into a templated base class `DigestionEnzymeDB`, with new subclasses `ProteaseDB` and `RNaseDB`. (Inheritance and the singleton pattern are somewhat tricky to combine!)
Also turned `EnzymaticDigestion` into a base class, with new subclasses `ProteaseDigestion` and `RNaseDigestion`.

pyOpenMS still to do.


src/openms/include/OpenMS/CHEMISTRY/DigestionEnzymeDB.h
```diff
@@ -0,0 +1,288 @@
...
+      {
+        // @TODO: why does this use a different exception than "getEnzyme"?
+        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
+                                         String("Enzyme with regex " + cleavage_regex + " was not registered in Enzyme DB, register first!").c_str());
+      }
+      return enzyme_regex_[cleavage_regex];
+    }
+
+    /// load enzymes from given file
+    void setEnzymes(const String& filename)
```
timosachsenberg: shouldn't we make the DB read only? Otherwise we will need a lot of looking

src/openms/include/OpenMS/CHEMISTRY/DigestionEnzymeDB.h
```diff
@@ -0,0 +1,288 @@
...
+    {
+      clear();
+      readEnzymesFromFile_(filename);
+    }
+
+    /// deletes all enzymes, resulting in an empty database
+    void clear()
```
timosachsenberg: same here

src/openms/include/OpenMS/CHEMISTRY/DigestionEnzyme.h
```diff
@@ -0,0 +1,175 @@
...
+     @ingroup Chemistry
+
+     @brief Abstract base class for digestion enzymes
+  */
+  class OPENMS_DLLAPI DigestionEnzyme
+  {
+  public:
+
+    /// possible substrates for digestion enzymes
+    enum Substrate
```
timosachsenberg: do we need this enum? base classes should not know about derived ones (and this enum is basically a typeID)

src/openms/include/OpenMS/CHEMISTRY/DigestionEnzymeProtein.h
```diff
@@ -33,90 +33,62 @@
...
     /** @name Assignment
      */
     //@{
     /// assignment operator
-    Enzyme& operator=(const Enzyme& enzyme);
+    DigestionEnzymeProtein& operator=(const DigestionEnzymeProtein& enzyme);
     //@}
 
+    /// returns the substrate (here: protein)
+    enum DigestionEnzyme::Substrate getSubstrate() const;
```
timosachsenberg: is this needed? sounds like a backdoor to dynamic type id. 
So either you know at the caller what type you have and don't need this function.
Or you use a DigestionENzyme* and don't know the type but need to deduce it using some dynamic_cast or type id. In both cases this usually indicates that there is something wrong with the design.

timosachsenberg:
In general, it looks very good. My main point of critics that we need to discuss are:
- that we should make the DBs read-only, otherwise we need to add tons of locks
- get rid of any dynamic type ID query functionality in the base class (aka getSubstrate())

src/openms/include/OpenMS/CHEMISTRY/DigestionEnzymeDB.h
hendrikweisser: That function was already there, but you make a valid point. It's not actually used anywhere in OpenMS and I'm happy to remove it.

src/openms/include/OpenMS/CHEMISTRY/DigestionEnzymeProtein.h
hendrikweisser: I think you're right. I'll remove it.

timosachsenberg:
looks good from my side. there is still a conflict and it would be cool if you could add pyOpenMS bindings too at some point

hendrikweisser:
Thanks for the review, Timo! I'm working on the Python bindings.

hendrikweisser:
I've fixed the merge conflict and updated pyOpenMS. Unfortunately I'm getting compile errors from pyOpenMS that I can't figure out:
```
pyopenms/pyopenms_2.cpp:4469:31: error: 'RNaseDB' is not a member of 'OpenMS'
   autowrap::AutowrapPtrHolder<OpenMS::RNaseDB>  inst;
                               ^
pyopenms/pyopenms_2.cpp:4469:31: error: 'RNaseDB' is not a member of 'OpenMS'
pyopenms/pyopenms_2.cpp:4469:46: error: template argument 1 is invalid
   autowrap::AutowrapPtrHolder<OpenMS::RNaseDB>  inst;
                                              ^
```
The weird thing is that the wrapper files for RNaseDB, "pxds/RNaseDB.pxd" and "addons/RNaseDB.pyx" are copy-and-replace versions of the ones for ProteaseDB, which seems to be working.

timosachsenberg:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rt-using

hroest:
> copy-and-replace versions of the ones for ProteaseDB, which seems to be working.

well that was the problem!

src/openms/include/OpenMS/CHEMISTRY/RNaseDB.h
```diff
@@ -0,0 +1,64 @@
...
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Hendrik Weisser $
+// $Authors: Hendrik Weisser $
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_CHEMISTRY_PROTEASEDB_H
+#define OPENMS_CHEMISTRY_PROTEASEDB_H
```
hroest: copy-paste-error: change to `OPENMS_CHEMISTRY_RNASEDB_H` and it works

hendrikweisser: Argh. Thanks, Hannes. Well spotted.

timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=243444

PR #2944, opened 2017-10-10
Fix/calculate bg estimation

dmccloskey:
I changed the calculateBgEstimation_ method in MRMTransitionGroupPicker.  This version uses the intensity at the start and end of the peak to estimate the background.  I found that neighboring peaks would often inflate the background estimation when using the average intensity of the region surounding the start and end of the peak.

@hroest might want to take a look at this implementation to see if it misses the intention of the original.


src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -103,41 +103,53 @@ namespace OpenMS
...
+    double intensity_left = 0.0;
+    double rt_apex = 0.0;
+    double intensity_right = 0.0;
+    for (MSChromatogram::const_iterator it = chromatogram.begin() + 1; it != chromatogram.end(); it++)
     {
-      if (it->getMZ() > best_left)
+      MSChromatogram::const_iterator it_prev = it;
+      it_prev--; //previous point
+
+      if (it->getMZ() == best_left)
```
hroest: I dont think thats a good idea, floating point numbers should not be compared with equal. what if there is a numerical error somewhere?

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -103,41 +103,53 @@ namespace OpenMS
...
+    double delta_int_apex = delta_int*delta_rt_apex/delta_rt;
 
-    // decrease the iterator and the nr_points by one (because we went one too far)
-    double intensity_right = (it--)->getIntensity();
-    nr_points--;
-
-    avg_noise_level = (intensity_right + intensity_left) / 2;
-    background = avg_noise_level * nr_points;
+    avg_noise_level = intensity_max - delta_int_apex;
+    background = intensity_min*delta_rt + 0.5*delta_int*delta_rt;
```
hroest: it seems that here you are integrating using the real RT. This however wont work with the rest of the code which basically uses "measurement points" and does not use RT. Basically the rest of the code sums up intensity, so we cannot start using `delta_rt` here or the background estimate will be completely off.

hroest:
I am not quite sure how this is conceptually different, I think my method uses the edge points on the right and left and than uses the average (instead of the minimum) to get a background estimate. Also there are some issues with the code - maybe we can add it as an optional parameter and people can choose to use your bg estimate?

dmccloskey:
Thanks for the feed back Hannes.  Can you clarify what you mean by the following:

"it seems that here you are integrating using the real RT. This however wont work with the rest of the code which basically uses "measurement points" and does not use RT. Basically the rest of the code sums up intensity, so we cannot start using delta_rt here or the background estimate will be completely off."

I am not sure what you mean by "measurement points".  At least for the background intensity, it seems to give the right results on the test data that I am using.  The peak area background seems to make sense as well.  Does the implementation I added throw off the intensity background estimation or the area estimate (or both).  If it is just the area, then I think I am missing how you are actually integrating the peak area.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -114,15 +152,15 @@ namespace OpenMS
       MSChromatogram::const_iterator it_prev = it;
       it_prev--; //previous point
 
-      if (it->getMZ() == best_left)
+      if (it->getMZ() >= best_left && it_prev->getIntensity() <= best_left)
```
hroest: hmm you probably meant `it_prev->getMZ()` ?

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -114,15 +152,15 @@ namespace OpenMS
...
       {
         intensity_left = it->getIntensity();
       }
       else if (it->getIntensity() >= peak_height && it_prev->getIntensity() <= peak_height)
       {
         rt_apex = (it->getMZ() + it_prev->getMZ())/2;
       }
-      else if (it->getMZ() == best_right)
+      else if (it->getMZ() >= best_right && it_prev->getIntensity() <= best_right)
```
hroest: same here

hroest:
can you please remove the file `.vscode/tags` ?

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -140,6 +140,56 @@ namespace OpenMS
...
+      rt_max = best_left;
+      intensity_min = intensity_right;
+      rt_min = best_right;
+    }
+    else 
+    {
+      intensity_max = intensity_right;
+      rt_max = best_right;
+      intensity_min = intensity_left;
+      rt_min = best_left;
```
hroest: can you remove this variable? it breaks the build: 

/.../OpenMS/src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp:169:50: error: variable ‘rt_min’ set but not used [-Werror=unused-but-set-variable]

     double intensity_max, intensity_min, rt_max, rt_min;
                                                  ^
cc1plus: all warnings being treated as errors
Linking CXX shared library ../../lib/libOpenMS.so

dmccloskey:
This should fix the rt_min and .vscode folder issues that you commented on.  The bad comparison was fixed in an earlier commit.

dmccloskey:
There is something not quite right still with the calculateBgEstimationExact_.  Hold off on pulling anything until I can confirm things are looking good on a much larger sample set.

dmccloskey:
The background calculations are working now.

dmccloskey:
#2944 This should fix it.

hroest:
looks like there is one more error:

```
    Showing 1 file with new issues

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
Same expression on both sides of '||'.
321

if ((background_subtraction_ == "smoothed_average" || background_subtraction_ == "smoothed_average") && smoothed_chroms.size() <= k)

```

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -140,6 +140,54 @@ namespace OpenMS
...
   }
 
+  void MRMTransitionGroupPicker::calculateBgEstimationExact_(const MSChromatogram& chromatogram,
+      double best_left, double best_right, double peak_height, double & background, double & avg_noise_level)
+  {
+    // determine (in the chromatogram) the intensity at the left / right border
+    double intensity_left = 0.0;
+    double rt_apex = 0.0;
+    double intensity_right = 0.0;
+    for (MSChromatogram::const_iterator it = chromatogram.begin() + 1; it != chromatogram.end(); ++it)
```
hroest: maybe just use `std::next( chromatogram.begin() )` instead of `+1` ?

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -140,6 +140,54 @@ namespace OpenMS
...
+    else 
+    {
+      intensity_max = intensity_right;
+      intensity_min = intensity_left;
+      rt_min = best_left;
+    }
+    double delta_int = intensity_max - intensity_min;
+    double delta_rt = best_right - best_left;
+    double delta_rt_apex = std::fabs(rt_min-rt_apex);
+    double delta_int_apex = delta_int*delta_rt_apex/delta_rt;
```
hroest: as I said before, this will fail badly unless your spacing is exactly 1 second. You should use `double delta_int_apex = delta_int*delta_rt_apex/nr_points;`

see also the code above calculateBgEstimationAverage_

src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
```diff
@@ -454,6 +493,56 @@ START_SECTION((template < typename SpectrumT > void remove_overlapping_features(
...
+  // Correct the background
+  MRMTransitionGroupPicker picker;
+
+  picker.calculateBgEstimationExact_(chromatogram, 
+    best_left, best_right, 
+    peak_height, background,
+    noise_level);
+
+  TEST_REAL_SIMILAR(background, 1140.392866);
+  TEST_REAL_SIMILAR(noise_level, 1908.596906);
```
hroest: If you compare the two results, it seems also clear here that when you apply this in the code

```
           intensity_sum -= background;
           peak_apex_int -= avg_noise_level;
```

that with this background estimate, you are not getting the right / expected result

hroest:
>  If it is just the area, then I think I am missing how you are actually integrating the peak area.

there is no integration at the moment, the area is simply calculated by summing up the intensity. I think this is where our confusion comes from - however, for your method to work we would have to change the way the intensity is calculated as well.

hroest:
@dmccloskey see this branch for a first implementation of a proper peak integration https://github.com/dmccloskey/OpenMS/pull/2

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: If the trapezoid integration introduced in feature/peakAnnotation is now the default, should this not be changed?  I.e., would it be best to change the calculateBgEstimationAverage_ to match instead?

dmccloskey:
Let me know how best to proceed on this one concerning the integration of the trapezoid rule.  If feature/peakAnnotation is merged into develop in the near future, I can just merge that into this branch and then update the codes and tests to accomodate that.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
hroest: trapezoid integration is currently just an experimental feature and not the default

hroest:
> Let me know how best to proceed on this one concerning the integration of the trapezoid rule

I think we probably want to do some testing first before we change the quantification strategy. It may also make sense to give users a choice on this

dmccloskey:
> I think we probably want to do some testing first before we change the quantification strategy. It may also make sense to give users a choice on this

I definitely agree.  What is your timeline for adding in an option for trapezoid integration?  I can fix the code how it is now to match integration by intensity summation for this PR, and then I guess in the future issue another fix and PR to accomodate different integration strategies once they are in place?

hroest:
> I can fix the code how it is now to match integration by intensity summation for this PR, and then I guess in the future issue another fix and PR to accomodate different integration strategies once they are in place?

sounds like the best idea, so that we can attack this problem piece by piece. Please do so.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -185,7 +191,9 @@ namespace OpenMS
     double delta_int_apex = delta_int*delta_rt_apex/delta_rt;
 
     avg_noise_level = intensity_min + delta_int_apex;
```
hroest: so basically what you are calculating here is the noise level *at the peak apex* - so this calculation is probably best for correcting peak apex calculations. I am not so sure how good it is for calculating peak area corrections, especially if the peak apex is not centered. Practically I dont think it makes a big difference but to me it seems that what I am doing currently is better for the area correction while what you are doing is better for the peak apex intensity correction

dmccloskey:
I screwed up on how I was calculating the area.  The latest push is actually the correct way of doing it for the exact method and gives comparable results.

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: Yes exactly.  For the area version I am walking the line from best left to best right and summing the intensities.

dmccloskey:
All tests are now passing.  Not sure what the conflicting files are about.  This was not a problem in the second to last push.

hroest:
@dmccloskey I made the merge here, please accept this PR https://github.com/dmccloskey/OpenMS/pull/4 or merge yourself...

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -498,6 +502,28 @@ namespace OpenMS
...
+        The background is estimated by averaging the noise on either side of the
+        peak and then subtracting that from the total intensity.
+    */
+    void calculateBgEstimationAverage_(const MSChromatogram& chromatogram,
+                                  double best_left, double best_right, double & background, double & avg_noise_level);
+    
+    /**
+      @brief Will use the chromatogram to estimate the background noise 
+        using the exact intensity of the left and right peak borders,
+        and then subtract it
```
hroest: it seems like the description of the two functions is now exactly the same, but they dont do the same. can we rectify that, I would suggest you update yours

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -498,6 +502,28 @@ namespace OpenMS
     void findLargestPeak(std::vector<MSChromatogram >& picked_chroms, int& chr_idx, int& peak_idx);
     
     /**
+      @brief Will use the chromatogram to estimate the background noise 
+        using the average intensity to the left and right of the peak borders,
+        and then subtract it
```
hroest:
```
@brief Will use the chromatogram to estimate the background noise 

Background estimation is computed as the average intensity of the left and right peak borders to produce the average noise level. The background is then computed by integrating (multiplying by number of points) over all data points, yielding a trapezoidal area under the curve.
```

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -154,6 +154,76 @@ namespace OpenMS
...
+      --it_prev; //previous point
+
+      if (it->getMZ() >= best_left && it_prev->getMZ() < best_right)
+      {
+        // calculate the background using the formula
+        // y = mx + b where x = retention time, m = slope, b = left intensity
+        double delta_int = intensity_right - intensity_left; // sign will determine line direction
+        double delta_rt_current = (it->getMZ() - best_left);
+        double background_int_current = delta_int/delta_rt*delta_rt_current + intensity_left;
+        background = background + background_int_current;
```
hroest: I guess this also works for non-uniformly spaced data while mine assumes the data is uniformly spaced? I think apart from that our code is using the same idea, right?

hroest:
looks good, maybe comment the code and explain the differences between the approaches and then we are good to go

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
dmccloskey: The integration strategy for both methods is using the method of summing the intensities at each point.  Once the trapezoid integration is implemented as an alternative integration strategy, the code will need to be updated as we spoke about previously.

dmccloskey: Both methods have their pros and cons:

average: probably optimal for operating on the raw chromatogram as it will not be susceptible to spurious points right next to the peak boundaries.

exact: probably optimal for operating on the smoothed chromatogram as it will give a more precise calculating of the noise at the peak apex and the background under the peak.

dmccloskey:
If any other comments are needed let me know, but I think they should be much better explained now.

PR #2946, opened 2017-10-11
[FEATURE] OpenSWATH Support for PercolatorAdapter

grosenberger:
This PR introduces support for OpenSWATH results in PercolatorAdapter. The motivation for this feature was to provide a replacement / alternative for mProphet/PyProphet that can be easily deployed in OpenMS / KNIME workflows. Currently, Percolator support is limited to semi-supervised learning and scoring. Future extensions might make use of other features like FIDO for protein inference, etc.

The implementation is different to idXML / mzid support. OSW SQLite files are directly parsed and read on MS1, MS2 and transition-level. Percolator is then executed individually for the different levels (parameter: ``-osw_level``). The results can be stored in the same OSW file or in new files. Importantly, transition-level scoring requires MS2 and MS1 scores to be present in the input file.

The resulting OSW file can then be further processed by PyProphet to export the legacy TSV format or to run IPF, etc.

```
# Score on MS2-level
PercolatorAdapter --in_osw openswath_results.osw --out openswath_results.osw --osw_level ms2

# Score on MS1-level
PercolatorAdapter --in_osw openswath_results.osw --out openswath_results.osw --osw_level ms1

# Score on transition-level
PercolatorAdapter --in_osw openswath_results.osw --out openswath_results.osw --osw_level transition
```

We have assessed and compared the results using two ground-truth data sets: The SWATH-MS Gold Standard data set (SGS) of the OpenSWATH paper and the phosphopeptide reference data set of the IPF paper. The results on both data sets are qualitatively similar to PyProphet, however this might be different on other data sets.


src/topp/PercolatorAdapter.cpp
```diff
@@ -638,6 +649,219 @@ class PercolatorAdapter :
...
+    sqlite3_step( stmt );
+
+    int cols = sqlite3_column_count(stmt);
+
+    // Generate features
+    int k = 0;
+    std::vector<std::string> group_id_index;
+    std::vector<std::string> features_names;
+
+    while (sqlite3_column_type( stmt, 0 ) != SQLITE_NULL)
```
hroest: I wonder whether we could put the sqlite specific stuff into the main library somewhere so that we dont have the abstractions leaking so much

src/topp/PercolatorAdapter.cpp
```diff
@@ -638,6 +649,219 @@ class PercolatorAdapter :
...
+                    "PEP DOUBLE NOT NULL);";
+
+    } else {
+      table = "SCORE_MS2";
+      create_sql =  "DROP TABLE IF EXISTS " + table + "; " \
+                    "CREATE TABLE " + table + "(" \
+                    "FEATURE_ID INT NOT NULL," \
+                    "SCORE DOUBLE NOT NULL," \
+                    "QVALUE DOUBLE NOT NULL," \
+                    "PEP DOUBLE NOT NULL);";
```
hroest: do you add these tables? again, it seems like a good idea to keep all of the code in the `FORMAT` folder so that we have some overview over which tables (may) exist and how to read these tables

grosenberger: Good point. I will move the functions to ``FORMAT``.

hroest:
sounds great and very useful, thanks a lot.

src/topp/PercolatorAdapter.cpp
```diff
@@ -638,6 +649,219 @@ class PercolatorAdapter :
     }
     return EXECUTION_OK;
   }
+
+  ExitCodes readOSWFile_(std::string in_osw, std::string osw_level, std::stringstream& pin_output, double ipf_max_peakgroup_pep, double ipf_max_transition_isotope_overlap, double ipf_min_transition_sn) {
```
hroest: can you pass by const ref if possible?
``` 
ExitCodes readOSWFile_(const std::string& in_osw, const std::string& osw_level, 
std::stringstream& pin_output, const double& ipf_max_peakgroup_pep, 
const double& ipf_max_transition_isotope_overlap, const double& ipf_min_transition_sn) {
```

src/openms/include/OpenMS/FORMAT/OSWFile.h
```diff
@@ -0,0 +1,86 @@
...
+    /// Destructor
+    virtual ~OSWFile();
+
+    /**
+      @brief Reads an OSW SQLite file and returns the data on MS1-, MS2- or transition-level
+      as stringstream TXT input for PercolatorAdapter.
+    */
+    void read(const std::string& in_osw, const std::string& osw_level, std::stringstream& pin_output, const double& ipf_max_peakgroup_pep, const double& ipf_max_transition_isotope_overlap, const double& ipf_min_transition_sn);
+
+    static int callback(void * /* NotUsed */, int argc, char **argv, char **azColName);
```
hroest: I guess that can be private?

grosenberger:
Thank you for the review, @hroest. I have implemented your feedback and added TOPP-level tests. I believe that two Travis-CI checks failed due to job running time limitations.

Issue #2948, opened 2017-10-12
MSSimulator and PTMs


lopippo:
Greetings,

I turned to MSSimulator to simulate mass spectra in the context of a mass spec teaching course. I was willing to model spectra for modified or unmodified proteins (acetylation on N-term and phosphorylations). Looking at the manual, I did not find a way to add this information. Is there a way to do it by somehow documenting the fasta file? Maybe, one option would be to define pseudo-amino acids, like B for phosphoserine, for example? Were would it be best to add these definitions?

Thank you for that useful program.
Sincerely,
Filippo

cbielow:
Hi,

you can just add modifications to the FASTA file,

e.g.
```
>seq1 optional comment [# intensity=567.4 #]
VPTPNVSVVDLTC(Carbamidomethyl)R
```

This is not explicitly mentioned in the docs. A fix is coming up.
Note, the mods are fixed. If you require variable mods, you need to do the combinatorics 'manually'.

cbielow:
Let me know if you run into any trouble.

cbielow:
fixed in #2949 

closing for now. Re-open if something does not work

lopippo:
Fine, thank you. I'll test that next week.

PR #2949, opened 2017-10-13
[DOC] add documention for PTM simulation

cbielow:
as requested in #2948


cbielow:
2 travis timeouts. fine otherwise

hroest:
@cbielow

> For amino acid modifications, insert their name at the respective amino acid residues. The modifications are fixed. 

makes sense, that is what you show in the docu. But for the variable ones I still dont understand how one would do that

> If you need variable modifications, you have to add the desired combinatorial variants (presence/absence of one or all modifications) to the FASTA file.

what do you mean, isnt that what you also said for the fixed ones, insert them at the respective AA residue?

Issue #2051, opened 2016-07-14
PhosphoScoring: improve default score for unambiguous cases


hendrikweisser:
The PhosphoScoring tool (our implementation of Ascore) currently assigns a score of -1 to all peptides that do not enter the actual scoring algorithm. This includes peptides without phosphorylation as well as peptides with as many phosphorylations as phospho-sites. However, it would be good to be able to distinguish between these two classes easily based on the score.
A simple solution would be to assign a very high score value (e.g. 1000) to phospho-peptides with unambiguous localization. That way, filtering according to a min. score of 0 would retain all phospho-peptides and remove all non-phosphorylated peptides.
(In OpenMS it is already possible to filter by modification in IDFilter, but improving the scoring scheme would still be helpful for analyses e.g. in R or Excel.)

PR #2951, opened 2017-10-14
PhosphoScoring: improve default score for unambiguous cases (solves #2051)

hendrikweisser:
Apart from some coding style adaptations, the main change is to use DefaultParamHandler for general AScore parameters (rather than passing them in every function) and the addition of the new "unambiguous_score" (advanced) parameter.

### DIFF
--- a/doc/doxygen/parameters/DefaultParamHandlerDocumenter.cpp
+++ b/doc/doxygen/parameters/DefaultParamHandlerDocumenter.cpp
@@ -32,6 +32,7 @@
 // $Authors: Marc Sturm $
 // --------------------------------------------------------------------------
 
+#include <OpenMS/ANALYSIS/ID/AScore.h>
 #include <OpenMS/ANALYSIS/ID/IDMapper.h>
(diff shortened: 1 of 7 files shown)


hendrikweisser:
The two CI test failures are due to time-outs.

hroest:
can you look at this:

```

408 std::vector<PeakSpectrum> windows_top10 = ptr_test->peakPickingPerWindowsInSpectrumTest_(real_spectrum);
```

timosachsenberg:
I think it is fine to introduce the default param handler but I would be interested if this affects performance in a way. Note that before the algorithm was basically a static function that can be easier inlined by the compiler. Now he certainly can't because he doesn't know if there is e.g. concurrent access to the member variable.

hendrikweisser:
@hroest: I don't know what you're referring to.

@timosachsenberg: I don't think this is performance-critical code. I think the parameter handling is cleaner, and certainly better conforming to OpenMS standards now.

hroest:
@hendrikweisser I am talking about the codacy results: https://www.codacy.com/app/OpenMS/OpenMS/pullRequest?prid=991428

PR #2952, opened 2017-10-15
Adjust 'StringView::substr' API to conventions (fixes #2920)

### LINKED ISSUE #2920: StringView::substr API doesn't agree with String::substr and std::string::substr
The second argument to `String::substr` and `std::string::substr` gives the length of the desired substring. In `StringView::substr`, the second argument is `end_index`, which gives the end position of the substring (undocumented whether it's at-the-end or past-the-end). If the starting position (first argument) isn't zero, then the semantics are different, which can lead to confusion.
`StringView::substr` should be changed to support the expected interface.

hendrikweisser:


hendrikweisser:
CI test failures are due to time-outs.

timosachsenberg:
+1

Issue #2942, opened 2017-10-10
FalseDiscoveryRate filter option does not work for proteins


Stortebecker:
Hi.

I like the [filter option](https://github.com/OpenMS/OpenMS/issues/2531) added to FalseDiscoveryRate. However, I recognized that the output when filtering for 0.01 peptide FDR is not exactly the same as when filtering with IDFilter. 

Also, in my hands the FDR filtering on protein level does not work at all inside the FalseDiscoveryRate node. When I use the option `-FDR:protein 0.01`, I get the following error: `Error: File not found (the file '' could not be found)`.

Using OpenMS release v2.2 on Linux 64bit.

timosachsenberg:
thanks for reporting.
Looking at the code again, it seems to directly call the IDFilter code if -protein "true" is passed and FDR:protein is set to a value < 1
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/FalseDiscoveryRate.cpp#L168
Can you provide an example file and command line that triggers the error message?

Stortebecker:
Hi.
Here is a [test file](https://gist.github.com/Stortebecker/66a19d6d614b82896f4875adddf55c60). Sorry for the size. 
I ran `FalseDiscoveryRate -in FidoAdapter_out.idxml -out output.idxml -protein true -FDR:protein 0.01`

Stortebecker:
Interestingly, the FDR filtering on protein seems to start, but then it stops:
```
FDR control: Filtering proteins...
Error: File not found (the file '' could not be found)
```

timosachsenberg:
ok tks I can reproduce the problem. will look into it

timosachsenberg:
I also get the Segmentation Fault if I use -score:pep so it is most likely something going wrong during parsing...

Stortebecker:
I do not get errors running FileInfo or IDFilter on that file:

````
IDFilter -in FidoAdapter_out.idxml -out output_IDFilter
Removing unreferenced protein hits...
Warning: While updating protein groups, some proteins were removed from groups that are still present. The new grouping (especially the group probabilities) may not be completely valid any more.
Before filtering:
1 protein identification(s) with 1357 protein hit(s),
3565 peptide identification(s) with 3565 peptides hit(s).
After filtering:
1 protein identification(s) with 1152 protein hit(s),
3565 peptide identification(s) with 3565 peptides hit(s).
[...]

PR #2953, opened 2017-10-16
Fix/fdr

timosachsenberg:
fixes https://github.com/OpenMS/OpenMS/issues/2942

@jpfeuffer tests are currently failing because some references are filtered in the current settings. Can you give me a hint how this is done properly for both files with / without groups?


jpfeuffer:
Can you elaborate? As your log info correctly states it should be fine to just remove the proteins that are filtered and either remove or set the group information to outdated. A user would need to run Fido again to update it.

jpfeuffer:
The relation between peptides and proteins depends heavily on the settings like keep_unreferenced proteins etc.

timosachsenberg:
so currently I apply these functions if something has been filtered.
```
+    if (filter_applied)
      {
 +      IDFilter::removeUnreferencedProteins(prot_ids, pep_ids);
 +      IDFilter::updateProteinReferences(pep_ids, prot_ids, true);
 +      IDFilter::updateHitRanks(prot_ids);
 +      IDFilter::updateHitRanks(pep_ids);
 +      IDFilter::removeEmptyIdentifications(pep_ids);
 +      // we want to keep "empty" protein IDs because they contain search meta data
      }
```
The critical points are:
   IDFilter::removeUnreferencedProteins(prot_ids, pep_ids);     I 
   IDFilter::updateProteinReferences(pep_ids, prot_ids, true); // true or false?

which are enabled/disabled by flags in the IDFilter tool.
Ideally I would go without these flags and be compatible with normal idXML but also with those that contain groups.
Do you know what the correct combination is?

jpfeuffer:
With "normal IdXML" you mean the current commited test files?

timosachsenberg:
If people use this tool, they probably mainly use it with or without protein inference applied.
My guess was, that the stuff I wrote should result in what the user most likely expects. Though, I am not sure if this is correct for the case of applying protein inference algorithms with / without greedy resolution..

jpfeuffer:
Okay I see.
Filtering by a cutoff in general (if performed on the same scores as inference, i.e. PPs) should not invalidate greedy resolution of the remaining associations.

Protein filtering will invalidate saved groups (ambiguity groups need to be completely rearranged, indistinguishable ones only lose members). I think it makes sense to remove unreferenced peptides afterwards although not necessary if they lose their references to proteins correctly.

Peptide filtering will invalidate all grouping equally badly. I currently can't think of a case where peptide filtering makes sense after protein inference (except for maybe reporting, although there I  would report all and filter them externally)

timosachsenberg:
Cool thanks. So I can adapt the tests  to fit the new output I guess

timosachsenberg:
Many of the test files changed. To put this in context:

 1: default param, -PSM true -protein false
 2: with runs separated, -algorithm:treat_runs_separately -PSM true -protein false)
 3: with charge variants individual, -algorithm:split_charge_variants -PSM true -protein false 
 4: with charge variants and no charge 1 decoys -> q_value 0)
 5: proteins only (input is derived from "THIRDPARTY/FidoAdapter_1_output.idXML"), default param, -PSM true -protein false
 6: FDR filtering, -PSM true -protein false -FDR:PSM 0.05)
 7: FDR filtering, -PSM false -protein trie -FDR:protein 0.05)

timosachsenberg:
4: seems to be correct. All unreferenced proteins are removed. All peptide hits are kept as is.

timosachsenberg:
@jpfeuffer can you take a look at test 5 and 7. I uploaded the test files that make the test pass. No proteins are contained in both cases. Is this expected?

timosachsenberg:
I think the filtering worked as expected but the tests are not ideal.

src/tests/topp/FalseDiscoveryRate_5_output.idXML
```diff
@@ -5,7586 +5,6 @@
 	</SearchParameters>
 	<IdentificationRun date="2015-06-29T15:12:34" search_engine="Fido" search_engine_version="" search_parameters_ref="SP_0" >
 		<ProteinIdentification score_type="q-value" higher_score_better="false" significance_threshold="0" >
```
hroest: why does this remove all Protein ids? Is this intended?

jpfeuffer: I think it is because in the input (FDR_output_1.idXML), all proteins have a score of 0. So the FDR for all will be #decoys / #all which is probably higher than the default cutoff. I think that was the reason for Timo's comment.

timosachsenberg:
Ok I fixed the last test. Question is how to proceed. As I understood @jpfeuffer the behavior is correct but the test files not ideal. What do you think?

jpfeuffer:
I'm would be more comfortable if we had a mini test. With e.g. 10 proteins. Should be enough to test correct functionality.

src/tests/topp/FalseDiscoveryRate_5_output.idXML
timosachsenberg: does this make sense? I mean the q-value should be the minimal FDR so these should not be removed

jpfeuffer: I think the FDR calculation does not care about the type of score that is in there. It uses what is in there and calculates the FDR anew.

jpfeuffer: It is weird that a dataset with already annotated q-values is used for FDR calculation AT ALL. That is what we want to compute here.

timosachsenberg:
@jpfeuffer do you have a small idXML at hand to address you last request?

timosachsenberg:
I will create the test files ...

jpfeuffer:
Sorry but I don't have any. You could maybe look at some FidoAdapter tests?

timosachsenberg:
I created the test files. Number 5 is from a normal search. Number 7 is manually created to check protein FDR filtering.

timosachsenberg:
Still wondering if we can get an idXML in, kind of a canonical state by applying:
```
       IDFilter::removeUnreferencedProteins(prot_ids, pep_ids);
       IDFilter::updateProteinReferences(pep_ids, prot_ids, true);
       IDFilter::updateHitRanks(prot_ids);
       IDFilter::updateHitRanks(pep_ids);
       IDFilter::removeEmptyIdentifications(pep_ids);
```
before storing the idXML file.

timosachsenberg:
@jpfeuffer  windows and mac fails are unrelated

jpfeuffer:
Seems to work correctly. It would be nice though if you could test two more points:
- If a peptide belongs to two proteins (one filtered, one kept)
- Increase FDR so that you get a resulting decoy protein and a protein with q-val > 0
Up to you. Minor suggestion.

timosachsenberg:
@jpfeuffer regarding the second point. I think there should now be a target protein contained that has a lower score than a decoy protein (which is not reported).
Strangely it doesn't appear in the list. Can you double check?

jpfeuffer:
Hm at first glance I would say so, too.

timosachsenberg:
thanks for checking... my bad. Will fix

timosachsenberg:
ok seems to work. strangely it keeps the decoy proteins in the filtered results. This is different for PSMs. Is this intended?

jpfeuffer:
Cool, looks much better now.
Well I think this is debatable. I would say usually you do not need decoys anymore after FDR calculation. And I think same defaults would be nice.

timosachsenberg:
ok I opened an issue: https://github.com/OpenMS/OpenMS/issues/3192

timosachsenberg:
latest addition fixes https://github.com/OpenMS/OpenMS/issues/3192
I think I also addressed the other request (proteins get removed (in this case because it is a decoy) and the peptide hits are updated.

timosachsenberg:
failed tests unrelated... will merge

PR #2959, opened 2017-10-17
Feature/masstrace correlation

hroest:
this code needs some cleanup but it would be nice to have in OpenMS eventually


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MasstraceCorrelator.h
```diff
@@ -0,0 +1,160 @@
...
+  {
+
+  public:
+
+    MasstraceCorrelator();
+
+    ~MasstraceCorrelator();
+
+    // a mass trace is a vector of pairs in (RT, Intensity)
+    typedef std::vector<std::pair<double, double> > masstracePointsType;
```
timosachsenberg: minor: typedefs should be capitalized

PR #2960, opened 2017-10-19
Moved Py File to src/pyOpenMS

cbielow:
to cleanup root directory


cbielow:
the file is run manually, so there are no dependencies. i.e. CI-tests are irrelevant.

jpfeuffer:
+1

PR #2964, opened 2017-10-20
[FIX] TOPPView, simplified synchronizing of peak annotations (fixes #2956)

### LINKED ISSUE #2956: TOPPView peak annotation merges user annotations and mz annotations in a strange way.
When I add the m/z value to an assigned peak and I click on another spectrum or save the annotation as a new idxml. The peak label and the m/z value are merged. Unfortunately in a way that only the m/z and the charge is left. Same is true for adding a peak annotation manually and a m/z. Seems like only using the label function works kind of but this doesn’t help for the already annotated peaks. 

e.g. two labels with “y1 +1” and “m/z 147.1127” becomes “147.1127 +1” 

BTW, I liked it better in the older versions when it read “2+” instead of “+2”. And it is a bit annoying that ToppView doesn’t automatically zoom into the appropriate m/z range for each MS2 but instead keeps the zoomed range of the previous spectrum.

enetz:
Multiple annotation per peak are now possible.
Adding and editing peaks works as intended.

### DIFF
--- a/src/openms_gui/source/VISUAL/LayerData.cpp
+++ b/src/openms_gui/source/VISUAL/LayerData.cpp
@@ -140,10 +140,12 @@ namespace OpenMS
   void LayerData::updatePeptideHitAnnotations_(PeptideHit& hit)
   {
     // copy user annotations to fragment annotation vector
-    Annotations1DContainer & las = getAnnotations(current_spectrum_);
+    Annotations1DContainer & las = getCurrentAnnotations();
 
-    vector<PeptideHit::PeakAnnotation> fas = hit.getPeakAnnotations();
+    // initialize with an empty vector
+    vector<PeptideHit::PeakAnnotation> fas;
 
+    // do not change PeptideHit annotations, if there are no annotations on the spectrum
     bool annotations_changed(false);
 
     // regular expression for a charge at the end of the annotation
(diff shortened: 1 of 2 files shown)


enetz:
fixes #2956

timosachsenberg:
cool but does everything work as before? I remember that we needed to track which peaks have already been annotated for some reason...

enetz:
Maybe I was overthinking things before, or didn't want to use the brute force way. Everything seems to work now, I tested:
Adding manual peak annotations (with and without charges), adding Peak annotation mz,
removing all those, saving to idXML and loading again,
In all orders I could think of and with multiple different annotations on the same peak.

This is simpler than before and seems to work fine.

What I am curious about are the `temporary_annotations_` and the function `removeTemporaryAnnotations()` in `TOPPViewIdentificationViewBehaviour`.
How are those different than the normal annotations from `LayerData.getAnnotations()` ?
I had to move the call to `removeTemporaryAnnotations()` after the synchronization, this way it seems to work.

timosachsenberg:
Maybe from the theoretical spectrum generator? If no fragment annotations are present in the metadata

enetz:
might be, but `removeTemporaryAnnotations()` removes the temporary ones AND everything that matches these in the `LayerData` annotations. That also removed all annotations, that were generated previously in any way.

To clarify that: All annotations that were available either from an ID file or manually added BEFORE a spectrum was selected, were removed after it was deselected. So only the newest annotations, added after the spectrum was selected for the last time (if there were any changes), were kept.
(That is if  `removeTemporaryAnnotations()` is called before synchronizing).

Now I remove the `LayerData` annotations when leaving an MS2 spectrum anyway, so the matching and removing in `removeTemporaryAnnotations()` seems redundant.

timosachsenberg:
Should then probably also go to master

PR #2965, opened 2017-10-20
[Feature] Added support for writing citations to CTD

jpfeuffer:
for use in e.g. KNIME.
Unfortunately CTD 1.7 only supports the use of a doi or url.
I am in touch with Luis to do a schema change, soon.


hroest:
well a doi is a start -- then people will know that they (i) should cite us and (ii) where to look for the full citation

timosachsenberg:
looks good

PR #2966, opened 2017-10-20
[FEATURE] add library init routine

### LINKED ISSUE #352: SEGFAULT in xerces libraries when removing default instance of SvmTheoreticalSpectrumGenerator [352]
Submitted by aiche on 2011-09-21 13:48:56

If you remove the default instance of SvmTheoreticalSpectrumGenerator in SvmTheoreticalSpectrumGenerator.C some tests fail with segfaults caused by xerces methods like 

xercesc_3_0::XMLString::operator new
xercesc_3_0::XMLString::transcode

until we know how to fix this we will leave the default instance in the class.

hroest:
fixes #352

### DIFF
--- a/src/openms/include/OpenMS/CONCEPT/Init.h
+++ b/src/openms/include/OpenMS/CONCEPT/Init.h
@@ -1,0 +1,50 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2017.
+//
+// This software is released under a three-clause BSD license:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of any author or any participating institution
+//    may be used to endorse or promote products derived from this software
+//    without specific prior written permission.
+// For a full list of authors, refer to the file AUTHORS.
+// --------------------------------------------------------------------------
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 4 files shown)


src/openms/include/OpenMS/CONCEPT/Init.h
```diff
@@ -0,0 +1,50 @@
...
+#include <OpenMS/CONCEPT/Exception.h>
+
+/**
+    @brief Initialization procedures
+
+    Part of the library needs to be initialized to be used properly, these
+    procedures need to be run at startup (and preferably only once).
+
+*/
+
```
timosachsenberg: don't we need an: extern xerces_init xinit; ?

src/openms/source/CONCEPT/Init.cpp
```diff
@@ -0,0 +1,63 @@
...
+        xercesc::XMLPlatformUtils::Initialize();
+      }
+
+      ~xerces_init() 
+      {
+        xercesc::XMLPlatformUtils::Terminate();
+      }
+
+    };
+    xerces_init xinit;
```
hroest: maybe add const?

src/openms/include/OpenMS/CONCEPT/Init.h
hroest: only if we want to expose it to external callers, right? which we dont want to?

Global variables have global scope, and can be used anywhere in the program. Like functions, you must use a forward declaration (via keyword extern) to use a global variable defined in another file. By default, non-const global variables have external linkage.Jun 19, 2007

so maybe we should make it const

timosachsenberg: "By default, non-const variables declared outside of a block are assumed to be external. However, const variables declared outside of a block are assumed to be internal."
Yes I guess we need const or static to declare internal linkage.

hroest: I have just added const on github, lets see whether the tests work

hroest:
what happened to the windows build?

jpfeuffer:
Hmm unfortunately it did not upload to CDash. I increased the retention time of the build logs for CI on Jenkins (which was the reason for the deletion).

jpfeuffer:
rebuild win

jpfeuffer:
probably does not work after it was merged

timosachsenberg:
hmm was the header file added to the cmake?

PR #2967, opened 2017-10-20
[DOC] add docu

hroest:


timosachsenberg:
thanks for adding the documentation. For users, it might be a bit easier to provide the table in an either/or  format listing proteomics / metabolomics specific fields or instead of using ** provide the context in brackets

PR #2968, opened 2017-10-20
Fix/mzxml

hroest:
after #2525 fixes mzXML files for the TPP (specifically SpectraST)


timosachsenberg:
looks good and seems to preserve MQ compatibility

timosachsenberg:
should this also go into master?

Issue #2779, opened 2017-07-18
Storing and exporting non-Unimod modifications (e.g. from MSFragger)


Matthias313:
Dear all, 
I tried to use the MzTabExporter tool after the IDfilter which resulted in an error message, see below:
```
14:18:26 NOTICE: MzTabExporter of node #12 started. Processing ...
exporting identifications: "C:/Users/MATTHI~1/AppData/Local/Temp/6/2017-07-18_141154_ukmz1005immz_18988_1/TOPPAS_tmp/msfragger/007_IDFilter/out/MFA174.idXML" to mzTab: 
exporting identifications: "C:/Users/MATTHI~1/AppData/Local/Temp/6/2017-07-18_141154_ukmz1005immz_18988_1/TOPPAS_tmp/msfragger/007_IDFilter/out/MFA175.idXML" to mzTab: 
exporting identifications: "C:/Users/MATTHI~1/AppData/Local/Temp/6/2017-07-18_141154_ukmz1005immz_18988_1/TOPPAS_tmp/msfragger/007_IDFilter/out/MFA173.idXML" to mzTab: 
Error: Unexpected internal error (Modification or Substitution identifier MUST NOT be null or empty in MzTabModification)

==============================================================================
14:18:30 ERROR: MzTabExporter failed!
```

Currently, we use OpenMS2.1 and @lars20070 told me that this might have been fixed in the latest version (OpenMS2.2). If this is the case, we would start to update our version, otherwise it would be helpful if this issue could be fixed.

Furthermore, lars suggested that the problem lies within this code:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/MzTab.cpp#L275

lars20070:
@timosachsenberg Does [this exception](https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/MzTab.cpp#L275) look familiar? Any idea why the `mod_identifier_` could be `null`? I will check tomorrow on `develop`.

lars20070:
@Matthias313 Here the probable reason for the crash.
```
lars@SchillingLinux:~/Desktop/test$ IDFileConverter -in MFA175.pepXML -out MFA175_NEW.idXML
Progress of 'Loading...':
Non-fatal error while loading 'MFA175.pepXML': No modification description given. Trying to define by modification mass.
[...]

PR #2969, opened 2017-10-21
Feature/msfragger Adds MSFraggerAdapter as util

lukaszimmermann:
Adds the MSFraggerAdapter as util. The util essentially boils down to parameter validation and writing the `fragger.params` file  

Some considerations are:

- There is no support for directly providing the `fragger.params` file, as this would be redundant to the ini file and also error-prone. 
- The default values for the parameters are taken from the official documentation (PDF). The flag` allow_multiple_variable_mods_on_residue` is set to true per default, hence the logic was inverted to the parameter flag `not_allow_multiple_variable_mods_on_residue`.
- In MSFragger Variable modifications can be specified arbitrarily by providing the mass and some special syntax for residues and termini. This is unchanged here.
- For the QProcess, a temp directory is created and the input spectra / database are linked there. This is required, because MSFragger writes indices to the same directory where the database resides.
- Non-Unimod modifications can be present in the result pepXML file. This will become problematic when exported to mzTab ( #2779 ). 
- The temp working dir is deleted upon tool destruction. Can be prevented with -debug 2 to manually inspect the fragger.params file.
- All parameters are non-advanced, except for the static modifications per residue (where cysteine has an added mass of 57.021464  per default)
- The MSFragger.jar  can probably not be added to THIRDPARTY because of licensing. 

Tests pending. I am going to do some searches with and without adapter and compare the results.


src/utils/MSFraggerAdapter.cpp
```diff
@@ -0,0 +1,859 @@
...
+    registerTOPPSubsection_("digest", "In-Silico Digestion Parameters");
+
+    // Enzyme
+    StringList enzyme_names;
+    EnzymesDB::getInstance()->getAllNames(enzyme_names);
+    registerStringOption_(TOPPMSFraggerAdapter::param_search_enzyme_name, "<search_enzyme_name>", "Trypsin", "Name of the enzyme to be written to the pepXML file", false, false);
+    setValidStrings_(TOPPMSFraggerAdapter::param_search_enzyme_name, enzyme_names);
+
+    // Cut after
+    registerStringOption_(TOPPMSFraggerAdapter::param_search_enzyme_cutafter, "<search_enzyme_cutafter>", "KR", "Residues after which the enzyme cuts", false , false);
```
Leon-Bichmann: you could use valid strings for this option and add the MSFragger Enzyme codes into share/Chemistry/Enzymes.xml

I think XTandem Adapter uses a getAllXTandemNames function

src/utils/MSFraggerAdapter.cpp
```diff
@@ -0,0 +1,859 @@
...
+
+    // No cut before
+    registerStringOption_(TOPPMSFraggerAdapter::param_search_enzyme_nocutbefore, "<search_enzyme_nocutbefore>", "P", "Residues that the enzyme will not cut before", false, false);
```
Leon-Bichmann: same here you could infer this option from the enzyme selection

src/utils/MSFraggerAdapter.cpp
Leon-Bichmann: Or otherwise you could add the most important ones to the description, how to specify no enzyme or unspecific cleavage for example

src/utils/MSFraggerAdapter.cpp
```diff
@@ -0,0 +1,859 @@
...
+    // TOPP tolerance
+    registerTOPPSubsection_("tolerance", "Search Tolerances");
+
+    // Precursor mass tolerance and unit
+    _registerNonNegativeDouble(TOPPMSFraggerAdapter::param_precursor_mass_tolerance, "<precursor_mass_tolerance>", 20.0, "Precursor mass tolerance", false, false);
+    registerStringOption_(TOPPMSFraggerAdapter::param_precursor_mass_unit, "<precursor_mass_unit>", "ppm", "Unit of precursor mass tolerance", false, false);
+    setValidStrings_(TOPPMSFraggerAdapter::param_precursor_mass_unit, validUnits);
+
+    // Precursor true tolerance
+    _registerNonNegativeDouble(TOPPMSFraggerAdapter::param_precursor_true_tolerance, "<precursor_true_tolerance>", 0.0, "Precursor true tolerance. It is strongly recommended to set this feature!", false, false);
```
Leon-Bichmann: a better description for this parameter would be nice, i guess.. I am only aware of precursor mass tolerance

src/utils/MSFraggerAdapter.cpp
```diff
@@ -0,0 +1,859 @@
...
+    setValidStrings_(TOPPMSFraggerAdapter::param_search_enzyme_name, enzyme_names);
+
+    // Cut after
+    registerStringOption_(TOPPMSFraggerAdapter::param_search_enzyme_cutafter, "<search_enzyme_cutafter>", "KR", "Residues after which the enzyme cuts", false , false);
+
+    // No cut before
+    registerStringOption_(TOPPMSFraggerAdapter::param_search_enzyme_nocutbefore, "<search_enzyme_nocutbefore>", "P", "Residues that the enzyme will not cut before", false, false);
+
+    // Number of enzyme termini
+    registerStringOption_(TOPPMSFraggerAdapter::param_num_enzyme_termini, "<num_enzyme_termini>", "2", "Number of enzyme termini (0, 1, or 2 for non-enzymatic, semi-enzymatic, fully-enzymatic)", false, false);
```
Leon-Bichmann: In the Comet and MSGFPlus valid strings are used and in Comet mapped to integers later:
```
setValidStrings_("num_enzyme_termini", ListUtils::create<String>("semi,fully,C-term unspecific,N-term unspecific"));
    map<string,int> num_enzyme_termini;
    num_enzyme_termini["semi"] = 1;
    num_enzyme_termini["fully"] = 2;

```

src/utils/MSFraggerAdapter.cpp
```diff
@@ -0,0 +1,859 @@
...
+    @verbinclude UTILS_MSFraggerAdapter.cli
+    <B>INI file documentation of this tool:</B>
+    @htmlinclude UTILS_MSFraggerAdapter.html
+ */
+
+// We do not want this class to show up in the docu:
+/// @cond TOPPCLASSES
+
+
+class TOPPMSFraggerAdapter final :
```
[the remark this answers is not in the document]
lukaszimmermann: Declares class as final, so it cannot be inherited from. Introduced with c++11

src/utils/MSFraggerAdapter.cpp
lukaszimmermann: I improved the description of the parameters

src/utils/MSFraggerAdapter.cpp
lukaszimmermann: There doesn't seem to be a MSFragger Enzyme code. One only specifies the cleavage site.
I could probably use the XTandemID to calculate the respective parameters for MSFragger:
Advantage: Standardized enzymes
Disadvantages: Error prone and artificially reduces the power of the MSFragger parameters.

lukaszimmermann: If one does not want cut, then the string with all amino acid residues can be specified for `enzyme_butnotbefore`. Unspecific cleavages does not seem to be possible.

src/utils/MSFraggerAdapter.cpp
lukaszimmermann: Is addressed.

src/utils/MSFraggerAdapter.cpp
```diff
@@ -0,0 +1,865 @@
...
+    {
+      return ILLEGAL_PARAMETERS;
+    }
+
+    QStringList process_params; // the actual process is Java, not MSFragger
+    process_params << "-Xmx" + QString::number(this->getIntOption_(TOPPMSFraggerAdapter::param_java_heapmemory)) + "m"
+        << "-jar" << this->executable.toQString()
+        << this->parameter_file_path.toQString();
+
+    for (const QString & input_file : this->input_files)
```
hroest: can you add a debug level output of the command line call? This may allow people to reconstruct what was done and potentially what went wrong?

hroest:
this looks really great, only thing missing are tests in the thirdparty folder. I guess people will have to download the tool separately and tell OpenMS where it is, I think that is fine.

src/utils/MSFraggerAdapter.cpp
lukaszimmermann: Done

hroest:
@lukaszimmermann can you also fix these errors: http://cdash.openms.de/viewTest.php?onlyfailed&buildid=243016

/home/travis/build/OpenMS/OpenMS/src/utils/MSFraggerAdapter.cpp:683:  Tab found; better to use spaces  [whitespace/tab] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/MSFraggerAdapter.cpp:684:  Tab found; better to use spaces  [whitespace/tab] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/MSFraggerAdapter.cpp:685:  Tab found; better to use spaces  [whitespace/tab] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/MSFraggerAdapter.cpp:686:  Tab found; better to use spaces  [whitespace/tab] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/MSFraggerAdapter.cpp:687:  Tab found; better to use spaces  [whitespace/tab] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/MSFraggerAdapter.cpp:688:  Tab found; better to use spaces  [whitespace/tab] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/MSFraggerAdapter.cpp:689:  Tab found; better to use spaces  [whitespace/tab] [5]

timosachsenberg:
Really cool and thanks for taking a look at that task.
By the way - how does the result look like regarding the detected modifications? And can we give a simple statistics or similar to guide users what mass deltas (or unimod modifications) to expect in the data?

lars20070:
Very nice. Thanks @lukaszimmermann. Together with @MFreidank we will test the new adapter on some of our datasets next week. As @timosachsenberg suggests, a simple mass delta distribution will be goal of our test.

src/utils/MSFraggerAdapter.cpp
jpfeuffer: Hmm we could do 2 parameters.
Predefined enzyme: shows a selection of our enzymes in the DB. You can add cutting rules instead of an ID for MSFragger. Plus a selection "User-specified rule".
User-specified rule: Only used, when User-specified rule is selected.

Or omit the user-specified rule selection and override the standard enzyme, when the second param is given.

hroest:
so one of the things I am worried about is how we parese the "unknown" mods and that is that  OpenMS tries to map them to "known" ones. However, if you have a distribution then some mods will be mapped to the "right" one, others will be mapped to no mod at all (user-defined mods) and some will be mapped to a close-by mod that is very unlikely and probably not the correct one but simply off due to a mass error. I think we need to look closely again at how OpenMS parses unknown mods and introduce some parsing logic into this, e.g. tell pepxml parser that it should *not* try to map the mods to unimod as this will have to be done by a separate tool that takes the distributions into account. Alternatively, we will have to store the "experimental" mass defect and store the "most likely explanation" in a residue mod.

src/utils/MSFraggerAdapter.cpp
jpfeuffer: unspecified cleavage might be able to be achieved by just listing all 20 AAs under "cutafter" and setting missed cleavages to max peptide length (-1?).

hroest:
PS: but it seems the current adapter is only copying the output files, not parsing them?

lukaszimmermann:
Exactly, currently the output is only copied, because at the current stage I do not know how to handle the unknown mods. I am not even sure if parsing the output file should be part of the adapter, as this is what the IDFileConverter is for.

hroest:
we should probably move the discussion regarding open mods and how to represent them in OpenMS into a separate issue

hroest:
@lukaszimmermann @timosachsenberg there is still a style test that is not working, can you look into that before we merge?

lukaszimmermann:
I do not understand the issue here.

hroest:
@lukaszimmermann this is clearly a false positives, see this branch where I tested a newer cppcheck:

https://travis-ci.org/OpenMS/OpenMS/jobs/301581914

```
853/1791 Test #853: src/utils/MRMPairFinder.cpp_cppcheck_test ......................................................................   Passed    0.05 sec
        Start 854: src/utils/MRMTransitionGroupPicker.cpp_cppcheck_test
        Start 855: src/utils/MSFraggerAdapter.cpp_cppcheck_test
        Start 856: src/utils/MSSimulator.cpp_cppcheck_test
854/1791 Test #854: src/utils/MRMTransitionGroupPicker.cpp_cppcheck_test ...........................................................   Passed    0.03 sec
855/1791 Test #855: src/utils/MSFraggerAdapter.cpp_cppcheck_test ...................................................................   Passed    0.06 sec
856/1791 Test #856: src/utils/MSSimulator.cpp_cppcheck_test ........................................................................   Passed    0.06 sec
        Start 857: src/utils/MassCalculator.cpp_cppcheck_test
        Start 858: src/utils/MetaProSIP.cpp_cppcheck_test



The following tests FAILED:

	104 - src/openms/source/ANALYSIS/OPENSWATH/PeakPickerMRM.cpp_cppcheck_test (Failed)
	113 - src/openms/source/ANALYSIS/QUANTITATION/IsobaricChannelExtractor.cpp_cppcheck_test (Failed)
	144 - src/openms/source/ANALYSIS/TARGETED/PrecursorIonSelection.cpp_cppcheck_test (Failed)
	155 - src/openms/source/APPLICATIONS/TOPPBase.cpp_cppcheck_test (Failed)
	349 - src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp_cppcheck_test (Failed)
	369 - src/openms/source/FORMAT/InspectOutfile.cpp_cppcheck_test (Failed)
	385 - src/openms/source/FORMAT/MzTabFile.cpp_cppcheck_test (Failed)
	404 - src/openms/source/FORMAT/SequestOutfile.cpp_cppcheck_test (Failed)
	559 - src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderAlgorithmPicked.cpp_cppcheck_test (Failed)
	596 - src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakShape.cpp_cppcheck_test (Failed)
	597 - src/openms/source/TRANSFORMATIONS/RAW2PEAK/TwoDOptimization.cpp_cppcheck_test (Failed)
	611 - src/superhirn/source/TRANSFORMATIONS/FEATUREFINDER/SUPERHIRN/LCMS.cpp_cppcheck_test (Failed)
	662 - src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp_cppcheck_test (Failed)
	679 - src/openms_gui/source/VISUAL/Spectrum2DCanvas.cpp_cppcheck_test (Failed)
	693 - src/openms_gui/source/VISUAL/TOPPASScene.cpp_cppcheck_test (Failed)
	719 - src/openms_gui/source/VISUAL/VISUALIZER/MetaInfoVisualizer.cpp_cppcheck_test (Failed)
	747 - src/topp/FeatureFinderIdentification.cpp_cppcheck_test (Failed)
	784 - src/topp/MapStatistics.cpp_cppcheck_test (Failed)
	801 - src/topp/PTPredict.cpp_cppcheck_test (Failed)
	814 - src/topp/RTPredict.cpp_cppcheck_test (Failed)
	845 - src/utils/IDMassAccuracy.cpp_cppcheck_test (Failed)
	858 - src/utils/MetaProSIP.cpp_cppcheck_test (Failed)
	861 - src/utils/MultiplexResolver.cpp_cppcheck_test (Failed)
	883 - src/utils/RTEvaluation.cpp_cppcheck_test (Failed)
```


but as you can see with the newer version, a lot of other tests fail. So we cant just switch to a newer version.

so how do we proceed? I suggest one of two options

- start fixing the above issues and then merge this PR
- disable cppcheck for MSFraggerAdapter for now, merge this PR and then work on fixing OpenMS until we can switch to a newer version of cppcheck and then re-enable the tests for MSFraggerAdapter

I think we should go the second route as we dont know how long the first one would take. I am therefore all in favor of disabling the test for now

hroest:
@lukaszimmermann this should do the trick:

```
$ git diff
diff --git a/src/tests/coding/cppcheck.cmake b/src/tests/coding/cppcheck.cmake
index 33788f0..58041b6 100644
--- a/src/tests/coding/cppcheck.cmake
+++ b/src/tests/coding/cppcheck.cmake
@@ -40,7 +40,7 @@ set(SOURCE_FILE_REGEX "\\.cpp$")
 # - MSNumpress.cpp since it is an external file
 # - DIAScoring.cpp since it triggers an cppcheck bug (may be fixed in newer versions of cppcheck)
 # - third party sources
-set(DO_NOT_TEST_THESE_FILES_REGEX "(/(moc|ui)_|/MSNumpress.cpp|/DIAScoring.cpp|thirdparty)")
+set(DO_NOT_TEST_THESE_FILES_REGEX "(/(moc|ui)_|/MSNumpress.cpp|/DIAScoring.cpp|MSFraggerAdapter.cpp|thirdparty)")
 
 # --------------------------------------------------------------------------
 # add_cpp_check_tests : This macro generates cppcheck tests for files in the
```

hroest:
also: can you remove this file: DO_NOT_BACKUP_THIS_FOLDER

src/tests/coding/cppcheck.cmake
```diff
@@ -40,7 +40,7 @@ set(SOURCE_FILE_REGEX "\\.cpp$")
 # - MSNumpress.cpp since it is an external file
 # - DIAScoring.cpp since it triggers an cppcheck bug (may be fixed in newer versions of cppcheck)
 # - third party sources
-set(DO_NOT_TEST_THESE_FILES_REGEX "(/(moc|ui)_|/MSNumpress.cpp|/DIAScoring.cpp|thirdparty)")
++set(DO_NOT_TEST_THESE_FILES_REGEX "(/(moc|ui)_|/MSNumpress.cpp|/DIAScoring.cpp|MSFraggerAdapter.cpp|thirdparty)")
```
timosachsenberg: there seems to be an additional +

lukaszimmermann:
Ok, thanks. I merged again with develop. Let's see how it goes.

hroest:
@lukaszimmermann ok now all the travis tests pass

hroest:
@lukaszimmermann can you add some black box tests to `src/tests/topp/THIRDPARTY/third_party_tests.cmake` that are enabled when the executable is found?

lukaszimmermann:
Yes, will do this soon

hroest:
@lukaszimmermann should we merge before or should we wait with the merge?

lukaszimmermann:
Depends on how urgent this is. I can probably do this on the weekend

hroest:
weekend sounds good - its not *that* urgent but I would like your contribution to be merged soon

lukaszimmermann:
Ok, I added some thirdparty tests.

lukaszimmermann:
The MSFragger.jar is of course missing on Travis. Shall it be added? I am not sure if this is compatible with the license.

jpfeuffer:
I guess we might be able to store it on our own CI servers though.

timosachsenberg:
I wrote the technology transfer office to get some official statement

Leon-Bichmann:
Would it be possible to get a "consent of MICHIGAN" as they write in their licence? Maybe if we ask the research group they give us a permission?

hroest:
We probably cannot add it to travis (unless we find a way to add a private version of it there or we get explicit permission of Michigan to use it for this and this purpose only) but we can add it to the nightly tests

lukaszimmermann:
@timosachsenberg  Is there any update from the technology transfer office.

timosachsenberg:
Yes. Currently its not possible. I am in contact with the authors and we will see if this can be changed

hroest:
can we ask them to have the users agree to an additional license clause that would allow us to distribute the software? similar to the vendor dlls that proteowizard distributes and we distribute as well through pwiz? Isnt this situation similar?

timosachsenberg:
This would be nice but according to them is not possible at this point.

hroest:
why are the tests even executed on travis if the binary is not present? I would be fine with merging and only having the nightly tests where we can install MSFragger. @jpfeuffer would that be possible?

jpfeuffer:
Could be a case sensitive string equality check. Let's see. Will install it for the nightlies.

hroest:
> Could be a case sensitive string equality check. Let's see. 

ok, lets try again. If it compiles, I am all for merging, maybe test them once on the nightlies? 

> Will install it for the nightlies.

great, thanks a lot

hroest:
@jpfeuffer I think we are ready to merge here, I would rather go ahead with this now than wait...

PR #2970, opened 2017-10-23
Fix/travis split

hroest:
there isnt much point having the clang travis tests around if they always time out. I guess there are two options

- use jenkins and our own build infrastructure for CI
- (temporarily) remove clang builds

I guess we should try to move away from travis in the medium to long run or just use it for quick evaluations as the gcc release build. We anyways dont really need Debug/Release builds across multiple compilers for each PR. Its enough to see if everything builds (including pyopenms) and the tests work in one configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,15 +83,19 @@ matrix:
       compiler: clang
 
 ## clang tests class and TOPP tests: without gui and without address sanitizer (debug build)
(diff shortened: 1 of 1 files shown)


timosachsenberg:
fine with me

hroest:
@jpfeuffer ? are you  ok  with this too? i tried to use gcc for the release builds but it took too long and c827d91 also timed out.

jpfeuffer:
Yes that is good. We should have enough resources on linux to add some own CI, too.

hroest:
yes, if we if had a clang debug build, that would be great as it is missing in our build matrix right now. Also, I think the chance of an error only in clang or only in debug is pretty low and we would catch it in the nightlies...

jpfeuffer:
Agreed.

PR #2971, opened 2017-10-23
[FIX] Patch for OpenSWATH PQP File Conversion

grosenberger:
This PR fixes two issues with PQP files generated by TargetedFileConverter and used by OpenSwathWorkflow.

1. To do transition-level scoring, OpenSwathWorkflow commonly reads a sorted TraML or TSV file, where "detection transitions" are parsed first and "identification transitions" are added later. When writing an OpenSWATH TSV report, the ``decoy`` flag is set correctly because the first transitions of a peak group are used to set this flag. This is not always the case when a PQP input is used in combination with a OpenSWATH TSV report. The patch below ensures that this is always the case.
2. A bug led to wrong peptide-protein associations when converting TraML to PQP files.

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
@@ -725,10 +725,16 @@ namespace OpenMS
 
       // Go through all transitions, for each transition get chromatogram and
       // the chromatogram and the assay to the MRMTransitionGroup
+      Size detection_assay_it = -1; // store index for the last detection transition
       for (Size i = 0; i < assay_it->second.size(); i++)
       {
         const TransitionType* transition = assay_it->second[i];
         precursor_mz = transition->getPrecursorMZ();
+
+        if (transition->isDetectingTransition())
+        {
+          detection_assay_it = i;
+        }
 
         // continue if we only have MS1 (we wont have any chromatograms for
         // the transitions)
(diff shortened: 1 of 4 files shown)


hroest:
can you look at this error:
```


/.../OpenMS/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp:796:32: error: comparison of unsigned expression < 0 is always false [-Werror=type-limits]

       if (detection_assay_it < 0)
                                ^
cc1plus: all warnings being treated as errors
Linking CXX shared library ../../lib/libOpenMS.so
```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -725,11 +725,17 @@ namespace OpenMS
 
       // Go through all transitions, for each transition get chromatogram and
       // the chromatogram and the assay to the MRMTransitionGroup
+      Size detection_assay_it = -1; // store index for the last detection transition
```
hroest: you cannot assign -1 to a type of `Size`

timosachsenberg:
should this also go into master?

grosenberger:
If still possible that would be great. Should I just point the PR to OpenMS/master to do this?

timosachsenberg:
Via the web interface I think this will probably not work

grosenberger:
Ok, I will make a new PR to master.

PR #2974, opened 2017-10-24
Feature/osw ppm

hroest:
- some cleanup in OpenSWATH
- allow DIA extraction to be performed in Da or ppm


grosenberger:
Nice! The flag ``-ppm`` for ``OpenSwathWorkflow`` remains the same, right?

hroest:
yes, the `-ppm` flag is for the actual extraction of the chromatograms, this is for the re-scoring of the full spectrum

PR #2764, opened 2017-07-11
SiriusAdapter

oliveralka:
SiriusAdapter for integration of de novo identification (#2268).
This is the prototype. It works with Version 3.5 of Sirius CLI (latest). 

Tool consists of:
-SiriusAdapter: Adapter (input, parameter handling, run Sirius/CSI:FingerID, output)
-SiriusMSConverter: Conversion of MzML to internal .ms format
-SiriusMzTabWriter: Conversion of Sirius output to mztab
-CsiFinderIdMzTabWriter: Conversion of CSI:FingerID output to mztab

Missing: 
- Class tests
- Blackbox tests
- Sirius CLI binaries (not sure where they belong to - 3 different versions depending on Os)

PR #2976, opened 2017-10-25
[TEST] CSI:FingerID 

oliveralka:
Add test for CSI:FingerID 
Dependent on CMakeFlag - to reduce the burden on the REST in Jena & to reduce the runtime of ctest (sometimes CSIFingerID-test runs long depending on the REST and the calculation ~22 to 800s). 

CMakeFlag: ENABLE_FINGERID_TEST

Suggested in #2764


CMakeLists.txt
```diff
@@ -185,6 +185,14 @@ if(ADDRESS_SANITIZER)
 endif()
 
 #------------------------------------------------------------------------------
+# Enable fingerid_test
+#------------------------------------------------------------------------------
+option(ENABLE_FINGERID_TEST "Enable test for fingerid (SiriusAdapter)" OFF)
+if(ENABLE_FINGERID_TEST)
```
jpfeuffer: Did you try to just use the option throughout the CMake scripts?
I would guess that there is no need for a second variable.

CMakeLists.txt
```diff
@@ -185,6 +185,14 @@ if(ADDRESS_SANITIZER)
...
+   set(enable_fingerid = true)
```
jpfeuffer: I don't think this does what you want. I think it sets the variable to a list of "=" and "true" (which probably evaluates to true anyway if you check the variable in a boolean way).
See https://cmake.org/cmake/help/v3.0/command/set.html
But maybe this is not needed. See comment above.

src/tests/topp/THIRDPARTY/third_party_tests.cmake
```diff
@@ -193,6 +193,11 @@ if (NOT (${SIRIUS_BINARY} STREQUAL "SIRIUS_BINARY-NOTFOUND"))
   add_test("TOPP_SiriusAdapter_1" ${TOPP_BIN_PATH}/SiriusAdapter -test -executable "${SIRIUS_BINARY}" -in ${DATA_DIR_TOPP}/THIRDPARTY/SiriusAdapter_1_input.mzML -out_sirius SiriusAdapter_1_output.tmp -auto_charge -profile qtof -database all)
   add_test("TOPP_SiriusAdapter_1_out" ${DIFF} -in1 SiriusAdapter_1_output.tmp -in2 ${DATA_DIR_TOPP}/THIRDPARTY/SiriusAdapter_1_output.mzTab -whitelist "MTD")
   set_tests_properties("TOPP_SiriusAdapter_1_out" PROPERTIES DEPENDS "TOPP_SiriusAdapter_1")
+  if (enable_fingerid)
```
jpfeuffer: Consider using the variable set by the Option.

jpfeuffer:
Should we activate it in the nightlies?

oliveralka:
I will look at your suggestions first and change that - afterwards in the nightlies.

oliveralka:
You can activate in the nightlies - sometimes the calculations of CSI:FingerID may be not consistent.  (Happens from time to time)

jpfeuffer:
Hmm what does that mean? That tests randomly fail? In the same version? That would be unfavourable.

oliveralka:
I'm not sure - I had the problem that the ranking in CSIFingerID was different from the one before - so diff said it failed - that happened to me once now (two rows in the output had been switched).

jpfeuffer:
Hm okay. Yes we can try. Let's also wait for @hroest opinion.

hroest:
I dont think we should enable it in the automated tests if it depends on the (unreliable) behavior of an external server. What if the server is not available, what if they update their version, what if returns different results from time to time? I think we should have the test to document what we expect to come out and enable it when we work on this part of the code.

oliveralka:
@jpfeuffer fail of nightlies independent of CSI:FingerID test.
Let's got with the suggestion of Hannes and use it for internal code testing only.

hroest:
I only get 
```
1975: Test command: /home/hr/openmsall/builds/openms_experimental/bin/SiriusAdapter "-test" "-executable" "/home/hr/openmsall/source/THIRDPARTY/Linux/64bit/sirius" "-in" "/home/hr/openmsall/source/OpenMS_experimental/src/tests/topp/THIRDPARTY/SiriusAdapter_1_input.mzML" "-out_sirius" "SiriusAdapter_2_output.tmp" "-out_fingerid" "SiriusAdapter_2_foutput.tmp" "-auto_charge" "-profile" "qtof" "-database" "all"
1975: Test timeout computed to be: 1500
1975: File '/home/hr/.OpenMS/OpenMS.ini' is deprecated.
1975: Updating missing/wrong entries in '/home/hr/.OpenMS/OpenMS.ini' with defaults!
1975: Executable is: /home/hr/openmsall/source/THIRDPARTY/Linux/64bit/Sirius/sirius
1975: Progress of 'loading spectra list':
1975: <File '/home/hr/.OpenMS/OpenMS.ini' is deprecated.> occurred 2 times
1975: No MS1 spectrum for this precursor. Occurred 0 times.
1975: <Updating missing/wrong entries in '/home/hr/.OpenMS/OpenMS.ini' with defaults!> occurred 2 times
1975: 0 spectra were skipped due to precursor charge below -1 and above +1.
1975: Charge of 0 was set to +1 due to positive polarity 5 times.
1975: Charge of 0 was set to -1 due to negative polarity 0 times.
1975: Executing: /home/hr/openmsall/source/THIRDPARTY/Linux/64bit/Sirius/sirius
1975: Working Dir is: /home/hr/openmsall/source/THIRDPARTY/Linux/64bit/Sirius
1975: FATAL: External invocation of Sirius failed. Standard output and error were:
1975:
1975:
1975: 1
1975: SiriusAdapter took 06:23 m (wall), 0.08 s (CPU), 0.00 s (system), 0.08 s (user).
6/7 Test #1975: TOPP_SiriusAdapter_2 ......................***Failed  383.30 sec
```

so it does not seem to work for me for whatever reason

hroest:
ok, now it worked. must really have been their server
```

The following tests passed:
        UTILSWRITEINI_SiriusAdapter
        UTILSWRITEINI_SiriusAdapter_SectionName
        UTILSWRITECTD_SiriusAdapter
        TOPP_SiriusAdapter_1
        TOPP_SiriusAdapter_1_out
        TOPP_SiriusAdapter_2
        TOPP_SiriusAdapter_2_out

100% tests passed, 0 tests failed out of 7

```

PR #2978, opened 2017-10-25
[FIX] reduced memory allocated for MSGF JVM

jpfeuffer:
Tests were sometimes failing on our low resource win builds.
512 MB should be enough for the small file.. I hope. Will also ask for upgrade of the VMs.


jpfeuffer:
Yay full green.

PR #2979, opened 2017-10-25
[doc] added cites to the qc tool set

mwalzer:


src/utils/QCCalculator.cpp
```diff
@@ -116,7 +116,7 @@ class TOPPQCCalculator :
 {
 public:
   TOPPQCCalculator() :
-    TOPPBase("QCCalculator", "Calculates basic quality parameters from MS experiments and subsequent analysis data as identification or feature detection.", false)
+    TOPPBase("QCCalculator", "Calculates basic quality parameters from MS experiments and subsequent analysis data as identification or feature detection.", false, {{ "Walzer M, Pernas LE, Nasso S, Bittremieux W, Nahnsen S, Kelchtermans P,  Martens, L", "qcML: An Exchange Format for Quality Control Metrics from Mass Spectrometry Experiments", "Molecular & Cellular Proteomics 2014; 13(8)" , "10.1074/mcp.M113.035907"}})
```
timosachsenberg: can you add line breaks

PR #2981, opened 2017-10-27
[FEATURE] Add contrib and THIRDPARTY as submodules.

jpfeuffer:
Mostly to keep track of the used commits.
Also helps in simplifying the quick_build scripts (later).
Problem: Since partial clones are not possible, you always have to check out the full THIRDPARTY independent of the platform.
Solution: Make the platform dependent folders separate repositories and add them as submodules of THIRDPARTY.
Will merge this into develop, too, after evaluation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ DartConfiguration.tcl
 /*.cmake
 Testing
 CMakeCache.txt
-contrib
 CMakeFiles
 Makefile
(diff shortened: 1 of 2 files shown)


mwalzer:
Can you add some examples of use and documentation(links?)

jpfeuffer:
Yes I will. Sure!

PR #2986, opened 2017-10-29
Feature/AbsoluteQuantitation

dmccloskey:
Support for absolute quantitation using IDMS.  Classes currently support the ability to apply a pre-generated calibration curve to data processed using the OpenSWATH pipeline.  Classes will eventually support automatically generating calibration curves from Standard samples.

NOTE:  I am currently using a new class  (AbsoluteQuantitationMethod and AbsoluteQuantitationMethodFile) to import the calibration data from .csv.  Eventually, it would be good to merge this with TransitionTSVReader and include the necessary headers in the TraML format.


hroest:
please see the build errors and comments

- http://cdash.openms.de/viewBuildError.php?buildid=244087
- https://www.codacy.com/app/OpenMS/OpenMS/pullRequest?prid=1029020

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitation.h
```diff
@@ -0,0 +1,260 @@
...
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Douglas McCloskey $
+// $Authors: Douglas McCloskey $
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_ANALYSIS_OPENSWATH_ABSOLUTEQUANTITATION_H
+#define OPENMS_ANALYSIS_OPENSWATH_ABSOLUTEQUANTITATION_H
```
hroest: please change to `OPENMS_ANALYSIS_QUANTITATION_ABSOLUTEQUANTITATION_H`

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
```diff
@@ -0,0 +1,206 @@
...
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Douglas McCloskey $
+// $Authors: Douglas McCloskey $
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_ANALYSIS_OPENSWATH_ABSOLUTEQUANTIFICATIONMETHOD_H
+#define OPENMS_ANALYSIS_OPENSWATH_ABSOLUTEQUANTIFICATIONMETHOD_H
```
hroest: same here

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
```diff
@@ -0,0 +1,206 @@
...
+  @param datum datum to evaluate the model at
+  @param transformation_model_params model parameters
+
+  @return evaluated datum.
+  */ 
+  double evaluateTransformationModel(const String & transformation_model,
+    const double& datum,
+    const Param& transformation_model_params);
+                                  
+  // members
```
hroest: can you make these members private?

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitation.h
```diff
@@ -0,0 +1,260 @@
...
+        It is assumed that all duplicate components have been removed.  If not,
+        the function will quantify all components, but the first internal standard found will
+        be used to calculate the ratio for the calculation.
+
+      @param unknowns A FeatureMap to quantify.
+
+    */ 
+    void quantifyComponents(FeatureMap& unknowns);    
+    
+    // members
```
hroest: can yo make these members private?

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationStandards.h
```diff
@@ -0,0 +1,114 @@
...
+
+namespace OpenMS
+{
+
+  /**
+    @brief AbsoluteQuantitationStandards is a class to handle the relationship between
+    runs, components, and actual concentrations.
+
+    TODO:
+    1. Move to FORMAT
```
hroest: please do so and adjust the header guards - actually this probably should go to METADATA

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationStandards.h
```diff
@@ -0,0 +1,114 @@
...
+       @brief Structure to map runs to components to known concentrations
+
+       Note that for the method to work, the features must be annotated with
+         a metaValue for "run_id"
+ 
+     */ 
+     void mapConcentrationsToComponents();
+                                      
+    // members
+    std::map<String,std::vector<featureConcentration>> components_to_concentrations;
```
hroest: can you make the members private?

src/openms/include/OpenMS/FORMAT/AbsoluteQuantitationMethodFile.h
```diff
@@ -0,0 +1,102 @@
...
+    void store(const String & filename, const std::vector<AbsoluteQuantitationMethod> & aqm_list);
+    
+    /**
+        @brief Checks if a file is valid with respect to the mapping file and the controlled vocabulary.
+
+        @param line Header line of the .csv file.
+        @param headers A map of header strings to column positions.
+        @param params_headers A map of transformation model parameter header strings to column positions.
+    */
+    void parseHeader(StringList & line, std::map<String, int> & headers,
```
hroest: can you make parseHeader and parseLine private?

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -0,0 +1,438 @@
...
+    if (component_1.metaValueExists(feature_name) && component_2.metaValueExists(feature_name))
+    {
+      double feature_1 = component_1.getMetaValue(feature_name);
+      double feature_2 = component_2.getMetaValue(feature_name);
+      // std::cout <<  "ratio = " << ratio << "." << std::endl;
+      ratio = feature_1/feature_2;
+    } 
+    else if (component_1.metaValueExists(feature_name))
+    {
+      LOG_INFO << "Warning: no IS found for component " << component_1.getMetaValue("native_id") << ".";
```
hroest: did you mean `"Warning: no IS found for component " << component_2.getMetaValue("native_id")`

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -0,0 +1,438 @@
...
+    size_t IS_component_it, IS_component_group_it;
+
+    // // iterate through the unknowns
+    // for (size_t i = 0; i < unknowns.size(); i++)
+    // {      
+
+    // iterate through each component_group/feature     
+    for (size_t feature_it = 0; feature_it < unknowns.size(); ++feature_it)
+    {
+      component_group_name = (String)unknowns[feature_it].getMetaValue("PeptideRef");
```
hroest: @timosachsenberg @dmccloskey we should find a better way to store this information than in metavalues, e.g. we probably need to change the `Feature` class at some point. E.g. each feature should be able to store a unique reference to its peptide...

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -0,0 +1,438 @@
...
+                    break;
+                  }
+                }
+                if (IS_found)
+                {
+                  break;
+                }
+              }
+            }
+            if (IS_found)
```
hroest: you could factor out the whole method searching for the IS into its own fxn, e.g. lines 210-250

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -0,0 +1,438 @@
...
+    );
+
+    // loop from all points to min_points
+    for (size_t n_points = component_concentrations.size(); n_points >= min_points; --n_points)
+    {
+      n_loops = component_concentrations.size() - n_points;
+      for (size_t  component_it = 0; component_it < n_loops; ++component_it)
+      {
+        // TODO: support for outliers
+        // // loop through max_outliers
```
hroest: @dmccloskey we have some pretty neat support for outlier detection and removal over at the RT alignment code. Please have a look at OpenSwathRTNormalizer and more specifically at OpenMS/ANALYSIS/OPENSWATH/MRMRTNormalizer.h I can work with you to make that code more generic but I think it already is -- e.g. you have a `std::vector<std::pair<double, double> > ` and it will find non-outlier points using either chauvenet, an iterative method based on the residual or jackknife aproach or RANSAC. There is also a publication describing these methods, Parker et al 2015

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
```diff
@@ -0,0 +1,259 @@
...
+  {
+  }
+  
+  void AbsoluteQuantitationMethod::setLOD(const double& llod, const double& ulod)
+  {
+    llod_ = llod;
+    ulod_ = ulod;
+  }
+  
+  void AbsoluteQuantitationMethod::getLOD(double& llod, double& ulod)
```
hroest: I am not so sure about this interface, it wont work for Python bindings.

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <OpenMS/ANALYSIS/MAPMATCHING/TransformationModelLowess.h>
+
+#include <cstddef> // for size_t & ptrdiff_t
+#include <vector>
+#include <string>
+
+namespace OpenMS
+{
+    
+  AbsoluteQuantitationMethod::AbsoluteQuantitationMethod()
```
hroest: why not provide a constructor that sets all these parameters at once ?

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
```diff
@@ -0,0 +1,259 @@
...
+    else if (transformation_model == "TransformationModelInterpolated")
+    {
+      TransformationModelInterpolated tm(data, transformation_model_params);
+      params = tm.getParameters();
+    }
+    else if (transformation_model == "TransformationModelLowess")
+    {
+      TransformationModelLowess tm(data, transformation_model_params);
+      params = tm.getParameters();
+    }
```
hroest: @timosachsenberg dont we have a factory fxn to avoid this kind of code? it seems like there should be one ...

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
```diff
@@ -0,0 +1,259 @@
...
+    if (transformation_model == "TransformationModelLinear")
+    {
+      TransformationModelLinear tm(data, transformation_model_params);
+      tm.invert();
+      result = tm.evaluate(datum);
+    }
+    else if (transformation_model == "TransformationModelBSpline")
+    {
+      TransformationModelBSpline tm(data, transformation_model_params);
+      // tm.invert(); // not supported
```
hroest: I dont think I follow the logic here, how can the result be correct if it requires inversion of the model but you only invert for the linear case?

src/tests/class_tests/openms/source/AbsoluteQuantitation_test.cpp
```diff
@@ -0,0 +1,309 @@
...
+  AbsoluteQuantitation absquant;
+
+  // set-up the unknown FeatureMap
+  FeatureMap unknown_feature_map;
+  // set-up the features and sub-features
+  std::vector<Feature> unknown_feature_subordinates;
+  Feature unknown_feature, component, IS_component;
+  String feature_name = "peak_apex_int";
+  // component 1
+  unknown_feature.setMetaValue("PeptideRef","component_group1");
```
hroest: again, this works for now but definitely is not the way to go forward. One problem is that we also use PeptideRef in the TOPPView GUI app to group together chromatograms.

hroest:
before merging, please also remove  .vscode/tags and squash

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
dmccloskey: Good idea.

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
dmccloskey: Can you send me the citation?  I am not able to find the Parker et al 2015 study.

This is a good idea.  There are a few criteria that would be good to have (e.g., minimum bias/accuracy for each point in the linear model) that I don't see yet in any of the methods.  However, I can see consolidating such an algorithm in the MRMRTNormalizer class as beneficial for future work as well.

Let me know your thoughts on what you mean by making it more general.

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
dmccloskey: At the moment I don't believe that there is a way to invert the other models, unless I am mistaken.

src/tests/class_tests/openms/source/AbsoluteQuantitation_test.cpp
dmccloskey: I am up for suggestions on this.

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
dmccloskey: I will do that.

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
dmccloskey: This is not a huge concern now, but in the future it might be good to add support for more complex models (i.e., Quadratic and Hill) that are also quite common in the pharmaceutical and clinical realm.  A brute force method would always work, but some of the more specialized methods may not.

hroest: @dmccloskey this is the study: https://www.ncbi.nlm.nih.gov/pubmed/26199342 my idea was just that this will give you an easy way to perform outlier detection and removal. Also feel free to implement further methods for outlier detection.

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
hroest: well, the easy way would be to switch  x/y coordinates and then recompute, right? or am I missing something?

src/tests/class_tests/openms/source/AbsoluteQuantitation_test.cpp
hroest: this is not an issue with your PR but with our way of storing component ids in features - its pretty ugly :-(

hroest:
please fix these issues:

- http://cdash.openms.de/viewBuildError.php?buildid=244224
- http://cdash.openms.de/testDetails.php?test=45177284&build=244223

```
Checking /home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp...
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:115: performance: Variable 'bias' is reassigned a value before the old one has been used.
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:154: performance: Variable 'calculated_concentration' is reassigned a value before the old one has been used.
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:315: style: The scope of the variable 'n_loops' can be reduced.
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:390: style: The scope of the variable 'calculated_concentration_ratio' can be reduced.
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:390: style: The scope of the variable 'actual_concentration_ratio' can be reduced.
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:390: style: The scope of the variable 'feature_amount_ratio' can be reduced.
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:390: style: The scope of the variable 'bias' can be reduced.
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp:313: style: Variable 'max_outliers' is assigned a value that is never used.
```

src/tests/class_tests/openms/source/AbsoluteQuantitation_test.cpp
dmccloskey: I agree.

dmccloskey:
I broke a few things.  I will let you know when I have them fixed.

src/openms/include/OpenMS/FORMAT/AbsoluteQuantitationMethodFile.h
dmccloskey: This seems to mess up the tests.  Is there a way of testing private methods?

hroest: well in principle you should not test private methods. sometimes it is useful though, so you can use a trick: make it protected, in the test inherit from your main class and pass it through. With this we maintain a clean interface and API of the library but can still test private methods

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
dmccloskey: I believe you would have to invert the model.  This is what the software vendors do as well.

src/openms/include/OpenMS/FORMAT/AbsoluteQuantitationMethodFile.h
dmccloskey: Is there an example of this in OpenMS?  I don't quite follow.

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
dmccloskey: No, I want to alert the user that the IS for component_1 is missing.

dmccloskey:
I think these latest comments addressed all of your previous comments.

src/openms/include/OpenMS/FORMAT/AbsoluteQuantitationMethodFile.h
hroest: for example: https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/MasstraceCorrelator_test.cpp#L47

here the function `matchMassTraces_` is `protected` in the parent class but the child class is just a facade that passes through the parameters and it is `public` in the child class. I would suggest something similar.

hroest:
@dmccloskey sounds good, this is almost ready to go. I would like to make the functions private though. 

Also, maybe you can consider setting the parameters either through `DefaultParamHandler` or have a single function for setting them. For getting them I would prefer simple  functions

```
double getX() const;
```

as these are easier to wrap in Python. What do you think?

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.cpp
timosachsenberg: yep could be solved using a or Factory<> but also adds some boilerplate code.

dmccloskey:
That sounds good.  I will fix the private/protected functions and add in the "wrapper" class to test them, and make the parameter calls a bit easier for wrapping in python.

dmccloskey:
I believe all of the updates were made.  Let me know if something is off.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitation.h
```diff
@@ -0,0 +1,250 @@
...
+        It is assumed that all duplicate components have been removed.  If not,
+        the function will quantify all components, but the first internal standard found will
+        be used to calculate the ratio for the calculation.
+
+      @param unknowns A FeatureMap to quantify.
+
+    */ 
+    void quantifyComponents(FeatureMap& unknowns);    
+
+    void findIS_();
```
hroest: can you make some clarification about this interface, are all these methods public or are some of them internal? why are only some of these methods in the python bindings?

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
```diff
@@ -0,0 +1,213 @@
...
+#include <cstddef> // for size_t & ptrdiff_t
+#include <vector>
+#include <string>
+
+namespace OpenMS
+{
+
+  /**
+    @brief AbsoluteQuantitationMethod is a class to handle the relationship between
+    runs, components, and actual concentrations.
```
hroest: can you add something like "this class represents one component and its LOD, LOQ ..."

src/openms/include/OpenMS/FORMAT/AbsoluteQuantitationMethodFile.h
```diff
@@ -0,0 +1,102 @@
...
+        @exception Exception::ParseError is thrown if an error occurs during parsing
+    */
+    void load(const String & filename, std::vector<AbsoluteQuantitationMethod> & aqm_list);
+
+    /**
+        @brief Stores an AbsoluteQuantitationMethod file.
+
+        @exception Exception::UnableToCreateFile is thrown if the file could not be created
+    */
+    void store(const String & filename, const std::vector<AbsoluteQuantitationMethod> & aqm_list);
```
hroest: can your emove "store" if its not implemented?

src/openms/include/OpenMS/METADATA/AbsoluteQuantitationStandards.h
```diff
@@ -0,0 +1,111 @@
...
+#include <cstddef> // for size_t & ptrdiff_t
+#include <vector>
+#include <string>
+
+namespace OpenMS
+{
+
+  /**
+    @brief AbsoluteQuantitationStandards is a class to handle the relationship between
+    runs, components, and actual concentrations.
```
hroest: please expand this docu, this is the same docu as in the other class

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
dmccloskey: I completely missed this.  I will update the documentation.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitation.h
dmccloskey: I moved findIS_ to private.  It is an internal class that is not currently used.

src/openms/include/OpenMS/FORMAT/AbsoluteQuantitationMethodFile.h
dmccloskey: Store will be needed.  I am waiting on a fix for csv::store before implementing it.

src/openms/include/OpenMS/METADATA/AbsoluteQuantitationStandards.h
dmccloskey: this was the correct documentation.  I also expanded it.

dmccloskey:
@hroest let me know when this is ready to be merged.

hroest:
yes, travis looks good and the Win build fails because we have some other issues. As you can see from the travis build, its all green: https://travis-ci.org/OpenMS/OpenMS/builds/306649050?utm_source=github_status&utm_medium=notification (you can also see that when clicking on the "red cross" next to 8e40107)

dmccloskey:
OK awesome.  Time for the next iteration of these classes...

PR #2973, opened 2017-10-24
Feature/spec lib update

timosachsenberg:
- some style changes/modernization of code base
- ppm support for queries
- support for isotopic misassignment in queries
- more extensive search parameter annotation in results

PR #2988, opened 2017-10-30
[Fix] Add links for new tools to TOPP/Utils docu

jpfeuffer:


jpfeuffer:
Sorry but I can't get rid of this MSP file introduced in #2973. On Mac they are treated as the same file.

jpfeuffer:
Okay I deleted the msp file on GitHub since I think it is the old one. Let's see if it fixes the error.

jpfeuffer:
Ok Windows and Linux pass. Let's test the nightlies.

jpfeuffer:
Merged docu part to release/2.3.0 too

lars20070:
Thanks @jpfeuffer.

PR #2993, opened 2017-11-01
Fix/shaker scores, fixes #2921

### LINKED ISSUE #2921: IDFileConverter - error with PeptideShaker mzident output
Hello,

it seems like the mzidentML output from PeptideShaker can't be converted with `IDFileConverter` due to strings being used in value tags, I always get an error like the following:

```
Error: Unexpected internal error (Could not convert string 'Confident' to a double value)
```

`Confident` can be something different of course, in this case probably coming from

```
<cvParam cvRef="PSI-MS" accession="MS:1002540" name="PeptideShaker PSM confidence type" value="Confident"/>
```

in the mzid file.

Is this an error in the converter or should a mzid file never contain strings as values?

enetz:
Reading in string type score for PeptideShaker mzid output.
Should fix #2921

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
@@ -2051,7 +2051,14 @@ namespace OpenMS
         {
           for (vector<CVTerm>::const_iterator cv = cvs->second.begin(); cv != cvs->second.end(); ++cv)
           {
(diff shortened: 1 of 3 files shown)


enetz:
I chose to make a specific exception for the PeptideShaker confidence type score for now.
One alternative would be to treat all secondary metadata scores as strings. This would break some tests at this point, since they are written with type="float" in idXML and I think it would be better to keep that information.

hroest:
:+1: seems to work fine

Stortebecker:
Great, thanks @enetz !

PR #2994, opened 2017-11-06
Feature/mrm mapper

hroest:
move code of MRM Mapping into library

- allow for multiple mappings
- add pyOpenMS wrapper
- cleanup code
- note: this means the old parameters wont work any more on the TOPP tool


src/openms/source/ANALYSIS/TARGETED/MRMMapping.cpp
```diff
@@ -0,0 +1,181 @@
...
+        notmapped++;
+        if (error_on_unmapped_)
+        {
+          throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
+              "Did not find a mapping for chromatogram " + String(i) + "! Maybe try to increase your mapping tolerance.");
+        }
+      }
+      else if (map_multiple_assays_)
+      {
+        for (auto & c : mapped_chroms) output.addChromatogram(c);
```
dmccloskey: :+1:

hroest:
there seems to be an issue with the test on travis, I need to investigate

hroest:
seems like it all passes, merge?

PR #2996, opened 2017-11-07
[FEATURE] simplified BinnedSpectrum implementation. Some bug fixes

timosachsenberg:


src/openms/source/COMPARISON/SPECTRA/BinnedSpectrum.cpp
```diff
@@ -40,91 +40,73 @@ using namespace std;
...
 
-  void BinnedSpectrum::setBinning()
+  void BinnedSpectrum::binSpectrum_(const PeakSpectrum& ps)
   {
-    if (raw_spec_.empty())
-    {
-      throw BinnedSpectrum::NoSpectrumIntegrated(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
-    }
-    bins_.clear();
+    // TODO: add assertion to check if spectrum ps is sorted by position or retrieve maximum differently
```
hroest: maybe use` OPENMS_PRECONDITION` ?

src/openms/source/COMPARISON/SPECTRA/BinnedSpectrum.cpp
```diff
@@ -40,91 +40,73 @@ using namespace std;
...
+      // add peak to neighboring bins
       for (Size j = 0; j < bin_spread_; ++j)
       {
-        bins_[bin_number + j + 1] = bins_.at(bin_number + j + 1) + raw_spec_[i].getIntensity();
-        // we are not in one of the first bins (0 to bin_spread)
-        //not working:  if (bin_number-j-1 >= 0)
-        if (bin_number >= j + 1)
+        bins_[idx + j + 1] = bins_.at(idx + j + 1) + p.getIntensity();
+        // prevent spreading over left boundaries
+        if (idx >= j + 1)
```
hroest: why not just write `idx-j-1 >= 0`? 

also, why dont we check to go over the right boundary? what if `idx+j+1` is larger than the vector?

hroest:
looks pretty good, the windows test failure is unrelated

src/openms/source/COMPARISON/SPECTRA/BinnedSpectrum.cpp
timosachsenberg: good point .... its copy & pasted code from the old implementation

timosachsenberg:
To summarize what I did:
Previously BInnedSpectrum stored the full (non-binned) raw spectrum and, thus, was able to recalculate bins on demand. So, one could say it was a binned view on a raw spectrum that had to cache the precalculated view data.
Now it is a simple implementation of a binned spectrum that is created once from a raw spectrum.
I tihnk this implementation is much better and simpler and could, in principle, also be used to reimplement aview (if one ever wanted to have something like that ;) ).

timosachsenberg:
will make some minor renames so please don't merge yet

timosachsenberg:
ok I added operator+ to SparseVectors so in principle we should now be able to do a simple accumulation of binned spectra

hroest:
can you check the style test:

```
The following tests FAILED:

	1750 - src/topp/SpecLibSearcher.cpp_cpplint_test (Failed)
```

src/tests/class_tests/openms/source/BinnedSumAgreeingIntensities_test.cpp
```diff
@@ -93,7 +93,10 @@ START_SECTION((double operator()(const BinnedSpectrum &spec1, const BinnedSpectr
   BinnedSpectrum bs2 (1.5,2,s2);
 
   double score = (*ptr)(bs1, bs2);
-  TEST_REAL_SIMILAR(score, 0.997576)
+  TEST_REAL_SIMILAR(score, 0.99707)
```
hroest: any reason why the numerics change?

hroest:
how come the tests pass before the contrib PR is merged? https://github.com/OpenMS/contrib/pull/55

src/tests/class_tests/openms/source/BinnedSumAgreeingIntensities_test.cpp
timosachsenberg: the reason is that the score normalization is performed by dividing by (sum1 + sum2).
Previously, sum1 and sum2 were calculated only on a subset of the dimensions (the minimum of the highest non-zero indices of both spectra). It is now calculated for the full spectrum.
I looked into the original publication and, as I understood it, the way I am doing it now is correct.

timosachsenberg:
let's wait till travis-ci finishes. I think it will not pass the tests.
@jpfeuffer why is the windows ci passing?

src/tests/class_tests/openms/source/BinnedSumAgreeingIntensities_test.cpp
timosachsenberg: in other words: if s1 had the last non-zero coefficient at, say, index 100 and s2 at index 200, then the sums were previously only calculated up to index 100. Now it is calculated for the full range of each spectrum.

timosachsenberg:
By the way, I also removed the check for matching precursor m/z. In my view it doesn't belong into the scoring class. There are different scenarios like matching without considering precursor m/z (m/z agnostic). Matching while only considering the precursor mass (charge agnostic). Matching with relative / absolute tolerances (normal case), etc.. I would leave this decision to the caller - otherwise, it will restrict how the scoring functions can be used.

jpfeuffer:
There were two potential problems:
- The GitHub API changed and we have to upgrade the plugin.
- The commit status build step was not the last one.

jpfeuffer:
But the build definitely also failed on Windows.

timosachsenberg:
Now supports also ppm bin sizes...

mwalzer:
cool, ppm bins sounds awesome :+1:

timosachsenberg:
windows and mac ci seem to be related to not finding qt4

timosachsenberg:
python Bindings need to be updated so please don't merge

jpfeuffer:
If you reset the submodule to develop we can build again.
I would like maracluster to be introduced with the maracluster PR.
And seems like you tried to update the submodule from an even older state (without crux)

jpfeuffer:
Windows fails. Probably missing dll_export for your functor or something like that.

timosachsenberg:
rebuild win

timosachsenberg:
builds on my computer and now passes on windows. travis timed out

jpfeuffer:
Are you on Linux? Do you think travis will work for other PRs if we merge?

timosachsenberg:
yes I am on linux. I have no idea but I could squash and merge the commits so we can easily revert it.

jpfeuffer:
Win still works. Contrib seems to be at correct HEAD. Since travis is probably just a caching timeout, it would be fine for me to merge.

hroest:
@timosachsenberg @jpfeuffer please do not merge branches when travis is not green. There is usually a reason for this, e.g. the build fails. Better to re-start travis and try on your local machine to be sure before merge (or you can quickly ping me).

PR #2997, opened 2017-11-07
fix #793 (metavalue check for PeptideIdentification)

### LINKED ISSUE #793: Remove metavalue check from PeptideIdentification for Release
cccd4a1 should be removed for release 2.0
since its just a precautionary measure some branch code still relies on getMetaValue("RT") etc, then this would be catched right now (so the developer can fix it).

cbielow:
the fix has been active for a few years now. Not needed any longer.


timosachsenberg:
+1 on pass.

PR #3000, opened 2017-11-08
[FIX] In XFDR, fix extraction of first rank IDs if input id IdXML and MzIndentML

lukaszimmermann:
There was an error in XFDR which resulted in Meta Values not being found when IdXML or MzIdentML was used as input.

- Fix some style issues
- Remove some code redundancy
- Made some code adaptions for C++11


enetz:
@lukaszimmermann I think I found the problem, it should be the error message here: http://cdash.openms.de/viewBuildError.php?buildid=245726

```
/.../OpenMS/src/utils/XFDR.cpp:829:49: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
           && (is_xquest_input ? ions_matched  >= arg_minionsmatched : true)
```

since both are unlikely to be negative, you should use an unsigned type, e.g. Size or UInt for both `ions_matched` and `arg_minionsmatched`

maybe just make line 531 into
`const UInt arg_minionsmatched = this->getIntOption_(TOPPXFDR::param_minionsmatched);`

and then also make sure there are no warnings about signed and unsigned comparisons

timosachsenberg:
just a side note: using ternary expression in long, combined expression is usually considered bad style.
see e.g. https://agiletribe.wordpress.com/2011/11/01/21-avoid-ternary-conditional-operator/ 
I also agree with them that it is fine in simple expressions like: int a = (b>c) ? f(b) : f(c);

timosachsenberg:
@enetz or @lukaszimmermann can you also apply this to master?

timosachsenberg:
will do as I am already on it

PR #3004, opened 2017-11-10
Feature/mrm feature qc

dmccloskey:
This feature allows for either filtering or flagging Features and Subordinates in FeatureMaps containing MRMFeatures.  The class allows the user to filter or flag by transition or transition group specific parameters.  

Use cases:
1. Filtering a FeatureMap generated from the OpenSWATH workflow based on a blank sample.
2. Filtering transitions/transition groups that do not pass user specified parameters for e.g., metadata, RT window, etc.
3. Flagging for further review by the user transitions/transition groups that do not pass user specified parameters for e.g., metadata, RT window, etc.

flag/filter parameters are currently read in by a .csv file analogous to a TraML file.  Eventually, this could be included in the TraML standard.


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -0,0 +1,179 @@
...
+      @return Map of labels/transition types and their corresponding number.
+    */ 
+    std::map<String,int> countLabelsAndTransitionTypes(const Feature & component_group,
+      const TargetedExperiment & transitions);
+
+    template <typename T>
+    bool checkRange(T const& value, T const& value_l, T const& value_u);
+
+    // Members
+
```
hroest: can you make these private?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureQC.h
```diff
@@ -0,0 +1,155 @@
...
+      /// resolution lower bound 
+      double resolution_l_;
+      /// resolution upper bound 
+      double resolution_u_;
+      /// retention time lower bound 
+      double rt_diff_l_;
+      /// retention time upper bound 
+      double rt_diff_u_;
+    };
+
```
hroest: can you make the members private or do you need to access them?

if they are public, please dont use a final underscore, eg use `components_qcs` instead

src/openms/include/OpenMS/FORMAT/MRMFeatureQCFile.h
```diff
@@ -0,0 +1,100 @@
...
+  void store(const String & filename, const MRMFeatureQC & mrmfqc);
+  
+  /**
+    @brief Checks if a file is valid with respect to the mapping file and the controlled vocabulary.
+
+    @param line Header line of the .csv file.
+    @param headers A map of header strings to column positions.
+    @param params_headers A map of transformation model parameter header strings to column positions.
+  */
+  void parseHeader(StringList & line, std::map<String, int> & headers,
```
hroest: same here make parseHeader and parseLine protected

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -0,0 +1,409 @@
...
+            std::cout << "n_transitions" << std::endl; //debugging
+            if (! checkRange(labels_and_transition_types["n_transitions"],
+              filter_criteria.component_group_qcs_[cg_qc_it].n_transitions_l_,
+              filter_criteria.component_group_qcs_[cg_qc_it].n_transitions_u_))
+            {
+              cg_qc_pass = false;
+            }
+
+            // ion ratio QC
+            for (size_t sub_it2 = 0; sub_it2 < features[feature_it].getSubordinates().size(); ++sub_it2)
```
hroest: I am kind of wondering how you encode the heavy/light in OpenMS, it seems that they are subfeatures of the same feature. Is that correct?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureQC.h
```diff
@@ -0,0 +1,155 @@
...
+    MRMFeatureQC();
+
+    /// Destructor
+    ~MRMFeatureQC();
+    //@}
+
+    // Members
+
+    /// QCs within a component group
+    struct ComponentGroupQCs
```
hroest: I think this needs more comments and explanations of what these structs represent and how they relate to one another. If this goes into the public API of OpenMS I think we should add some more information about what a Component is, what a ComponentGroup is etc and how the relate to the TraML entries. This will allow other users to use them correctly.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
dmccloskey: Yes.  They are subfeatures of the same feature.  Basically I am using the "LabelType" tag in the TraML file to encode this.  In the case of metabolomics the "heavy" refers to the internal standard (i.e., there will most often be only 1 "heavy" transition per transition group).

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureQC.h
dmccloskey: I will go ahead and add more documentation.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureQC.h
dmccloskey: They are public.  I will go ahead and change that.

dmccloskey:
The trick for the protected member worked quite well.  Let me know if there is anything else that needs to be addressed here.

dmccloskey:
Pending the pyopenms build, everything should be ready to go.

hroest:
can you remove the  .vscode/tags files?

hroest:
@dmccloskey why was this branch merged before all tests passed? lets please not do that any more!

dmccloskey:
Sorry about that.  I thought everything had built just fine.  I will be more careful next time.  I will also go back and fix whatever is breaking the pyopenms build.

hroest:
just try to make sure that all of travis has built, you can click on the little red cross next to the commit message to check

Issue #3002, opened 2017-11-10
MSSpectrum::get...DataArrayByName segfaults if name not found


hendrikweisser:
See e.g. the implementation here: https://github.com/OpenMS/OpenMS/blob/563e87e481119e58957c78769f49e633f546c7b6/src/openms/include/OpenMS/KERNEL/MSSpectrum.h#L256

The result of `find_if` always gets dereferenced, even if it's the past-the-end iterator (if `name` wasn't found).

enetz:
first some context, this was implemented and discussed here #2847  and then here #2855 .
@timosachsenberg @hendrikweisser  should we implement `has...DataArrayWithName()` functions? would that help?

hroest:
I am not sure how the iterator solution is a good idea as it exposes the internal workings to the user and at this point we might as well give the user raw access to the data structure and ask them to use find themselves. I suggest the interface should be 

```
void hasIntegerDataArray(const String& name);
// may throw InvalidArgumentException, you need to check hasIntegerDataArray first
IntegerDataArray& getIntegerDataArray(const String& name);
```

what would the alternative interface look like?

```
// may return invalid iterator, you need to compare against getIntegerDataArrayEndIterator
IntegerDataArray::iterator getIntegerDataArray(const String& name);
IntegerDataArray::iterator getIntegerDataArrayEndIterator(const String& name);
```

now I think the first interface is clearer and it is easier to check that people adhere to it.

hroest:
I am not sure how the iterator solution is a good idea as it exposes the internal workings to the user and at this point we might as well give the user raw access to the data structure and ask them to use find themselves. I suggest the interface should be 

```
void hasIntegerDataArray(const String& name);
// may throw InvalidArgumentException, you need to check hasIntegerDataArray first
IntegerDataArray& getIntegerDataArray(const String& name);
```

what would the alternative interface look like?

```
// may return invalid iterator, you need to compare against getIntegerDataArrayEndIterator
IntegerDataArray::iterator getIntegerDataArray(const String& name);
IntegerDataArray::iterator getIntegerDataArrayEndIterator(const String& name);
```

now I think the first interface is clearer and it is easier to check that people adhere to it. In either case, the current interface is pretty messy and contains multiple ways to access the data, either by getting a single array or by getting all of them ... 

```
const IntegerDataArrays& getIntegerDataArrays() const;
IntegerDataArrays& getIntegerDataArrays();
[...]

PR #3007, opened 2017-11-13
TOPPView: better support for peak annotations in "Identification View"

hendrikweisser:
In TOPPView's "Identification View", peak annotations stored for peptide hits can be visualized (as an alternative to generating a theoretical spectrum). Previously only labels with the annotations were added to peaks in the measured spectrum. With this change, a second spectrum containing the peaks from the annotations as well as the labels is generated (analogous to a theoretical spectrum). This highlights the actual peaks that are annotated.

(It would be really nice if different peaks could be shown in different colours, but currently that's only possible for the labels.)

@enetz: Can you please check if this still works for showing the special annotations for cross-linked peptides?

### DIFF
--- a/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
+++ b/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
@@ -145,6 +145,68 @@ namespace OpenMS
     // else if (layer.type == LayerData::DT_CHROMATOGRAM)
   }
 
+  void TOPPViewIdentificationViewBehavior::addFragmentAnnotations_(const PeptideHit& ph)
+  {
+    // called anew for every click on a spectrum
+    LayerData& current_layer = tv_->getActive1DWidget()->canvas()->getCurrentLayer();
+
+    if (current_layer.getCurrentSpectrum().empty())
+    {
+      LOG_WARN << "Spectrum is empty! Nothing to annotate!" << std::endl;
+    }
+
+    if (!current_layer.getCurrentSpectrum().isSorted())
+    {
+      QMessageBox::warning(tv_, "Error", "The spectrum is not sorted! Aborting!");
+      return;
+    }
+
+    typedef std::vector<PeptideHit::PeakAnnotation> FragmentAnnotations;
+
+    const FragmentAnnotations& fa = ph.getPeakAnnotations();
+
+    for (FragmentAnnotations::const_iterator it = fa.begin(); it!= fa.end(); ++it)
+    {
+      // query closest peak to expected position
+      int peak_idx = current_layer.getCurrentSpectrum().findNearest(it->mz, 1e-2);
+
+      // check if m/z fits
+      if (peak_idx == -1)
+      {
+        LOG_WARN << "Annotation present for missing peak.  m/z: " << it->mz << std::endl;
+        continue;
+      }
+
+      const double peak_int = current_layer.getCurrentSpectrum()[peak_idx].getIntensity();
+      DPosition<2> position = DPosition<2>(it->mz, peak_int);
+      String annotation = it->annotation;
+      // write out positive and negative charges with the correct sign, at the end of the annotation string
+      if (it->charge != 0)
+      {
+        annotation = it->charge > 0 ? annotation + "+" + String(it->charge): annotation + String(it->charge);
+      }
+
+      Annotation1DItem* item;
+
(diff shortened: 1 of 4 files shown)


enetz:
@hendrikweisser yes, this works well with my data
although in my opinion the bright green peak colour is too bright on a white background

hendrikweisser:
The "cpplint" style test failed with this error:
`Storage class (static, extern, typedef, etc) should be first.  [build/storage_class] [5]`
For this line (https://github.com/OpenMS/OpenMS/pull/3007/files#diff-2004e1cf1a16834bba9edf9018761662R864):
`for (const auto& ann : annotations)`

I'd say that's a false-positive.

hendrikweisser:
@enetz: Thanks for checking.
> although in my opinion the bright green peak colour is too bright on a white background

Do you know how to change that colour?

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
```diff
@@ -894,12 +833,143 @@ namespace OpenMS
...
+      String label = ann.annotation;
+      // write out positive and negative charges with the correct sign at the end of the annotation string
+      switch (ann.charge)
+      {
+      case 0: break;
+      case 1: label += "+"; break;
+      case 2: label += "++"; break;
+      case -1: label += "-"; break;
+      case -2: label += "--"; break;
+      default: label += ((ann.charge > 0) ? "+" : "") + String(ann.charge);
```
hroest: can you change this so that it also works for negative mode spectra?

hroest:
> The "cpplint" style test failed with this error:

I fixed it

timosachsenberg:
can you post a screenshot?

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
hendrikweisser: I think that should work already. Or have I overlooked something? (In the default case, `String(ann.charge)` already contains the minus sign.)

hendrikweisser:
Screenshot:
toppview_idview_screenshot

timosachsenberg:
cool looks pretty neat

timosachsenberg:
The peak color can (currently) be changed on a layer base: see e.g., https://github.com/OpenMS/OpenMS/blob/develop/src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp#L1061

An painting of lines is done here: https://github.com/OpenMS/OpenMS/blob/develop/src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp#L755
Which uses the current color

hendrikweisser:
Thanks, Timo. There's already a "Todo" item that shows up in the Spectrum1DCanvas documentation:
> Use spectrum StringDataArray with name 'label' for peak annotations (Hiwi, Johannes)

We should do that (once #3002 is resolved) and at the same time add support for a "color" IntegerDataArray that would allow setting different colours for individual peaks.

hendrikweisser:
CI test "pyopenms_test_memoryleaktests" failed. Not sure if it's relevant.

hroest:
> CI test "pyopenms_test_memoryleaktests" failed. Not sure if it's relevant.

no, generally not

timosachsenberg:
The question is, if and where we generate these color arrays.
1. I would not generate another vector of colors in the search engine or the external tool as this would mean we need to extend the PeptideHit data structure
2. we could hardcode the coloring to by matching annotations (e.g. is it annotated as b- or y-ions, etc.?) We should be consistent in the use of these strings anyways so it would not hurt to define them
3. we could add a (#CD5C5C html) color to the annotation string of every annotation. If it is there, we use it to color the peak or annotate the color to the spectrum in a stringdataarray named "color"

hendrikweisser:
@timosachsenberg:
> 2. we could hardcode the coloring to by matching annotations (e.g. is it annotated as b- or y-ions, etc.?) 

That would have been my approach. We already do this to get different colours for the ion name labels.

I agree with your argument against option 1, and I don't like option 3 because that would mix different information in one string (which you'd then have to parse out again).

By the way, the colors could be stored as ints, because Qt's QRgb (which you can convert to a QColor) is "equivalent to an unsigned int".

hendrikweisser:
I think this could be merged already. Getting different colours for individual peaks should be a separate set of changes.
CI and Nightlies failures seems unrelated.

PR #3008, opened 2017-11-13
[FIX] control tmpdir through ENV

### LINKED ISSUE #2995: OpenMS only allows limited control over TMPDIR
Hi,
  I built the OpenMS (2.2.0)
```
Installation information:
==================
Data path    : /home/xxx/software/OpenMS-2.2.0/share/OpenMS
Temp path    : /tmp
Userdata path: /home/xxx/
```

But, I can not find the ```OpenMS.ini``` at both ```${HOME}/.OpenMS/``` and ```${HOME}```. So, I failed to changed the ```tmp_dir```, and I wonder how could I change the default tmp path. 
Thanks!

hroest:
- fixes #2995

### DIFF
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -450,6 +450,7 @@ namespace OpenMS
   String File::getTempDirectory()
   {
     Param p = getSystemParameters();
+    String dir;
     if (getenv("OPENMS_TMPDIR") != 0)
     {
       dir = getenv("OPENMS_TMPDIR");
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -450,11 +450,20 @@ namespace OpenMS
   String File::getTempDirectory()
   {
     Param p = getSystemParameters();
-    if (p.exists("temp_dir") && String(p.getValue("temp_dir")).trim() != "")
+    if (getenv("OPENMS_TMPDIR") != 0)
     {
-      return p.getValue("temp_dir");
+      dir = getenv("OPENMS_TMPDIR");
     }
-    return String(QDir::tempPath());
+    else if (p.exists("temp_dir") && String(p.getValue("temp_dir")).trim() != "")
+    {
+      dir = p.getValue("temp_dir");
+    }
+    else
+    {
+      dir = String(QDir::tempPath());
+    }
+    dir.ensureLastChar('/');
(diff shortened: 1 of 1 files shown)


hroest:
@timosachsenberg @jpfeuffer I would also add this to master as it has minimal impact but potentially makes the life easier of many people

timosachsenberg:
sounds good. go ahead! By the way - do we have a place in the doc or wiki where we document this?

src/openms/source/SYSTEM/File.cpp
```diff
@@ -450,11 +450,21 @@ namespace OpenMS
...
+    {
+      dir = p.getValue("temp_dir");
+    }
+    else
     {
-      return p.getValue("temp_dir");
+      dir = String(QDir::tempPath());
     }
-    return String(QDir::tempPath());
+    dir.ensureLastChar('/');
```
cbielow: potentially dangerous on Windows if the path already ends in '\'

cbielow: unfortunately, there are other places in File.cpp where this is used as well

cbielow:
the OPENMS_TMPDIR needs documentation (both the function and user docs; probably in TOPP docu and developer docs as well??)

src/openms/source/SYSTEM/File.cpp
hroest: @cbielow are you sure that is the case here, it seems oddly strange that all functions here use `/` as path separetors. I think it works because we use Qt which expects `/` as path separator and translates this to the appropriate path on the platform. See for example this docu: http://doc.qt.io/qt-4.8/qfileinfo.html and also https://stackoverflow.com/questions/37821181/qprocess-running-process-whose-path-contain-spaces

so I think its actually required to use `/` -- however we should document that in our `File.h`

cbielow: the reason its dangerous is because we do not have control over the value of `dir`. The user could specify it as `c:\temp\` in his environment and we would add `/` giving `c:\temp\/`, which could be problematic. Its simply because `.ensureLastChar()` is a function which is completely unaware of paths and platform separators. The code before was fine, because there was no user input and using `/` on windows in principle does no harm.
Or did I miss something?

hroest:
@cbielow ok I removed that line

timosachsenberg:
rebuild win

jpfeuffer:
Replaced by linked PR.

PR #3011, opened 2017-11-13
Cppcheck

hroest:
multiple fixes suggested by cppcheck / codacy

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/KDTree.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/KDTree.h
@@ -1758,6 +1758,7 @@ protected:
 
 
 
+  // TODO: why not pass node by const-ref?
   std::pair<_Link_type,size_type>
   _M_get_j_min( std::pair<_Link_type,size_type> const node, size_type const level)
   {
@@ -1787,6 +1788,7 @@ protected:
 
 
 
+  // TODO: why not pass node by const-ref?
   std::pair<_Link_type,size_type>
   _M_get_j_max( std::pair<_Link_type,size_type> const node, size_type const level)
   {
--- a/src/openms/source/ANALYSIS/TARGETED/PrecursorIonSelection.cpp
+++ b/src/openms/source/ANALYSIS/TARGETED/PrecursorIonSelection.cpp
@@ -735,12 +735,7 @@ namespace OpenMS
     std::vector<PeptideIdentification> curr_pep_ids, all_pep_ids;
     std::vector<ProteinIdentification> curr_prot_ids, all_prot_ids;
 
-    std::ofstream* precs = 0;
-    if (precursor_path != "")
-    {
-      precs = new std::ofstream(precursor_path.c_str());
-    }
-
+    std::ofstream precs(precursor_path.c_str());
 #ifdef PIS_DEBUG
     std::cout << max_iteration_ << std::endl;
 #endif
@@ -767,9 +762,9 @@ namespace OpenMS
                   <<  new_features[c].getMetaValue("msms_score");
 #endif
 
-        if (precursor_path != "")
+        if (precs.good())
         {
-          (*precs) << new_features[c].getRT() << " " << new_features[c].getMZ() << " " << new_features[c].getIntensity() << std::endl;
+          precs << new_features[c].getRT() << " " << new_features[c].getMZ() << " " << new_features[c].getIntensity() << std::endl;
         }
 
 
@@ -979,12 +974,6 @@ namespace OpenMS
 
     }
 #endif
-
-    // cleanup
-    if (precursor_path != "")
-    {
-      delete precs;
-    }
   }
 
   void PrecursorIonSelection::simulateILPBasedIPSRun_(FeatureMap& features, PeakMap& experiment,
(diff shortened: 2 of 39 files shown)


src/openms/include/OpenMS/DATASTRUCTURES/KDTree.h
```diff
@@ -1787,6 +1788,7 @@ class KDTree : protected _Alloc_base<_Val, _Alloc>
 
 
 
+  // TODO: why not pass node by const-ref?
   std::pair<_Link_type,size_type>
   _M_get_j_max( std::pair<_Link_type,size_type> const node, size_type const level)
```
hroest: @hannesveit any insights?

src/topp/FeatureFinderIdentification.cpp
```diff
@@ -176,6 +176,9 @@ class TOPPFeatureFinderIdentification :
   public TOPPBase
 {
 public:
+
+  // TODO
+  // cppcheck-suppress uninitMemberVar
   TOPPFeatureFinderIdentification() :
```
hroest: @hendrikweisser can you add this? potentially in a separate issue?

src/topp/FeatureFinderIdentification.cpp
```diff
@@ -427,6 +430,8 @@ class TOPPFeatureFinderIdentification :
       {
         RTRegion region;
         region.start = *rt_it - rt_tolerance;
+        // TODO
+        // cppcheck-suppress uninitStructMember
```
hroest: @hendrikweisser can you add this? potentially in a separate issue?

src/utils/IDMassAccuracy.cpp
```diff
@@ -237,6 +237,9 @@ class TOPPIDMassAccuracy :
               md.exp_mz = it->getMZ();
               md.theo_mz = (hit.getSequence().getMonoWeight() + (double)charge * Constants::PROTON_MASS_U) / (double)charge;
               md.charge = charge;
+              // TODO
+              // md.intensity =  ??
+              // cppcheck-suppress uninitStructMember
               precursor_diffs.push_back(md);
```
hroest: @timosachsenberg can you fix this?

src/topp/IDMerger.cpp
```diff
@@ -281,7 +281,7 @@ class TOPPIDMerger :
 
       if (!add_to.empty())
       {
-        remove(file_names.begin(), file_names.end(), add_to);
+        file_names.erase(std::remove(file_names.begin(), file_names.end(), add_to), file_names.end());
```
hroest: @hendrikweisser @timosachsenberg agreed?

src/utils/IDMassAccuracy.cpp
timosachsenberg: struct MassDifference
{
  double exp_mz = 0.0;
  Int charge  = 0;
  double theo_mz = 0.0;
  double intensity = 0.0;
};
should work

src/topp/FeatureFinderIdentification.cpp
hroest: is this potentially a false positive as the end should always be set at line 437?

src/topp/IDMerger.cpp
timosachsenberg: Let's see. Now, add_to it is actually removed.
Before your change, the add_to filename was only moved to the end of file_names.
In the next line add__to is added at the front of file_names.
Need to check the downstream code because both ways could (in principle) make sense.

timosachsenberg: yep your change seems correct

src/openms/source/ANALYSIS/TARGETED/PrecursorIonSelection.cpp
```diff
@@ -737,7 +737,9 @@ namespace OpenMS
 
     std::ofstream* precs = 0;
     if (precursor_path != "")
+    {
       precs = new std::ofstream(precursor_path.c_str());
```
cbielow: this is bad IMHO.
What about
```
std::ofstream os(precursor_path.c_str());
...
if (os.good()) ...
```

src/openms_gui/source/VISUAL/TOPPASScene.cpp
```diff
@@ -467,13 +467,15 @@ namespace OpenMS
     }
 
     TOPPASEdge* edge;
+    // cppcheck-suppress uninitvar
```
cbielow: would `TOPPASEdge* edge(0)` also work? (less code and no dependency on cppcheck syntax)

src/openms/include/OpenMS/DATASTRUCTURES/KDTree.h
cbielow: I don't see why not use &. But should we change the code (since its not ours and it seems cleaner to leave it as is)

src/openms_gui/source/VISUAL/TOPPASScene.cpp
timosachsenberg: or TOPPASEdge* edge = nullptr;

src/openms/include/OpenMS/DATASTRUCTURES/KDTree.h
hroest: since this is external code I am a bit hesitant to change it.

src/openms/source/ANALYSIS/TARGETED/PrecursorIonSelection.cpp
hroest: sounds better

timosachsenberg:
only an unrelated test failed on msvc15. merge?

hroest:
well there are still issues with 

- ` src/openms/include/OpenMS/DATASTRUCTURES/KDTree.h `
- ` src/topp/FeatureFinderIdentification.cpp `

but maybe we should address these in a separate issue

PR #3014, opened 2017-11-16
[FEATURE] TargetedSpectraExtractor

pasqoo:
This class filters, annotates, picks, and scores spectra (e.g., taken from a DDA experiment) based on a target list.

The input experiment is expected to be a standard .mzML file.
The input target list is expected to be a standard TraML formatted .csv file.

The class deals with filtering and annotating only those spectra that reflect the characteristics described by the target list.
The filtering is done based on the transition_name, PrecursorMz and RetentionTime.
The spectra are smoothed and peaks are picked for each spectrum.
The spectra are then scored based on total TIC, FWHM and SNR.
One spectrum is chosen for each of those transitions for which at least one
valid spectrum was found and matched.


src/openms/source/ANALYSIS/OPENSWATH/SpectraExtractor.cpp
```diff
@@ -0,0 +1,583 @@
...
+  {
+    mz_tolerance_ = mz_tolerance;
+  }
+
+  double SpectraExtractor::getMZTolerance() const
+  {
+    return mz_tolerance_;
+  }
+
+  void SpectraExtractor::setMZToleranceUnits(const String& mz_tolerance_units)
```
timosachsenberg: using a string might be prone to errors. consider using a bool if you only want to chagne between absolute and relative (ppm) tolerances

timosachsenberg:
Hi, thanks for this nice contribution.
I have some general comments and some ideas to reduce the code size and make it easier maintainable.

Looking at the PR, it seems that the majority of code is spent to set/get parameters and then to execute a "workflow" that performs picking/scoring/etc.
Would it make sense to just copy the parameters of algorithms? For example, in 
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/FeatureFinderMetabo.cpp#L141
The parameters of the underlying algorithms (like picking, filtering, ...in your case) are queried and exposed to the user in one global parameter object with subsections.
If the algorithm is executed, its parameters can be extracted again from the subsections: https://github.com/OpenMS/OpenMS/blob/develop/src/topp/FeatureFinderMetabo.cpp#L219 
That way you could potentially save 80% of the code.

The other thing is, that I would potentially give the class a less general name or what do you think @hroest .

hroest:
I agree with Timo on both points, I think getting a better name would be useful and also saving some code using the Param system in  OpenMS. However, for me I think the second part is optional as it can also slow down the code a lot and adds dependencies so if you like the code with a lot of get/set options I am not in principle against it.

I suggested to use `TargetedSpectraSelector` or `TargetedSpectraExtractor` as potential names for the class.

pasqoo:
Thanks @timosachsenberg and @hroest for your reviews.
I replaced the "mz units" string parameter with a bool and renamed the class.

If the parameter handling is acceptable as it is, I'd like to keep the getters/setters.

src/tests/class_tests/openms/source/TargetedSpectraExtractor_test.cpp
```diff
@@ -0,0 +1,837 @@
...
+  for (UInt i=0; i<selected_spectra.size(); ++i)
+  {
+    TEST_NOT_EQUAL(selected_spectra[i].getName(), "")
+    TEST_EQUAL(selected_spectra[i].getName(), selected_features[i].getMetaValue("transition_name"))
+    TEST_EQUAL(selected_spectra[i].getFloatDataArrays()[1][0], selected_features[i].getIntensity())
+    TEST_EQUAL(selected_spectra[i].getFloatDataArrays()[1][0] >= ptr->getMinScore(), true)
+  }
+
+  ptr->selectSpectra(scored, selected_spectra);
+  TEST_NOT_EQUAL(selected_spectra.size(), 0)
```
hroest: can you please test for something meaningful, e.g. the actual number of selected spectra?

src/tests/class_tests/openms/source/TargetedSpectraExtractor_test.cpp
```diff
@@ -0,0 +1,837 @@
...
+  for (UInt i=0; i<selected_spectra.size(); ++i)
+  {
+    TEST_NOT_EQUAL(selected_spectra[i].getName(), "")
+    TEST_EQUAL(selected_spectra[i].getFloatDataArrays()[1][0] >= ptr->getMinScore(), true)
+  }
```
hroest: maybe you can test the first spectrum to check that it is correct

src/tests/class_tests/openms/source/TargetedSpectraExtractor_test.cpp
```diff
@@ -0,0 +1,837 @@
...
+  ptr->setTICWeight(1.0);
+  ptr->setFWHMWeight(1.0);
+  ptr->setSNRWeight(1.0);
+  ptr->setMinScore(15.0);
+
+  vector<MSSpectrum> extracted_spectra;
+  FeatureMap extracted_features;
+  ptr->extractSpectra(experiment, targeted_exp, extracted_spectra, extracted_features);
+
+  TEST_EQUAL(extracted_spectra.size(), extracted_features.size())
```
hroest: again, test for equality with a number, to make sure that the algorithm produces the same output also in the future. again, testing the first spectrum may not be a bad idea either

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.h
```diff
@@ -0,0 +1,273 @@
...
+
+    /**
+      @brief Filters and annotates those spectra that could potentially match the
+      transitions of the target list.
+
+      The spectra taken into account are those that fall within the precursor RT
+      window and MZ tolerance set by the user through the setRTWindows() and
+      setMZTolerance setters. Default values are provided for both parameters.
+
+      @warning The picked spectrum could be empty, meaning no peaks were found.
```
hroest: what happens if two spectra match within the m/z and RT tol?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.h
```diff
@@ -0,0 +1,273 @@
...
+    The input experiment is expected to be a standard .mzML file.
+    The input target list is expected to be a standard TraML formatted .csv file.
+
+    The class deals with filtering and annotating only those spectra that reflect
+    the characteristics described by the target list.
+    The filtering is done based on the transition_name, PrecursorMz and RetentionTime.
+    The spectra are smoothed and peaks are picked for each spectrum.
+    The spectra are then scored based on total TIC, FWHM and SNR.
+    One spectrum is chosen for each of those transitions for which at least one
+    valid spectrum was found and matched.
```
hroest: given that this is a whole algorithm, maybe you can comment in which order one would typically call the methods of this class

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -0,0 +1,581 @@
...
+  {
+    snr_weight_ = snr_weight;
+  }
+
+  double TargetedSpectraExtractor::getSNRWeight() const
+  {
+    return snr_weight_;
+  }
+
+  void TargetedSpectraExtractor::updateMembers_()
```
hroest: this just seems like a bad idea as you provide two interfaces with which to change the members of the class:

- through the regular Param class
- through all the get/set methods

I suggest to use only one method

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -0,0 +1,581 @@
...
+    FeatureMap& features
+  )
+  {
+    annotated_spectra.clear();
+    features.clear(true);
+    const std::vector<ReactionMonitoringTransition> transitions = targeted_exp.getTransitions();
+    for (UInt i=0; i<spectra.size(); ++i)
+    {
+      MSSpectrum spectrum = spectra[i];
+      // It is supposed to have RT in minutes in target list file, therefore we divide by 60.0
```
hroest: hmmm ... please note that the user may use seconds or minutes in the TraML file, so I would be careful about that assumption

hroest: especially as the default assumption of the CSV to TraML converter assumes seconds

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.h
pasqoo: they will match the same transition, and then `selectSpectra` would select the one with the highest score for that transition

src/tests/class_tests/openms/source/TargetedSpectraExtractor_test.cpp
```diff
@@ -0,0 +1,837 @@
...
+TargetedSpectraExtractor* ptr = 0;
+TargetedSpectraExtractor* null_ptr = 0;
+const String experiment_path = OPENMS_GET_TEST_DATA_PATH("TargetedSpectraExtractor_13C1_spectra0to100.mzML");
+const String target_list_path = OPENMS_GET_TEST_DATA_PATH("TargetedSpectraExtractor_13CFlux_TraML.csv");
+MzMLFile mzml;
+MSExperiment experiment;
+TransitionTSVReader tsv_reader;
+TargetedExperiment targeted_exp;
+mzml.load(experiment_path, experiment);
+tsv_reader.convertTSVToTargetedExperiment(target_list_path.c_str(), FileTypes::CSV, targeted_exp);
```
hroest: I suggest to specify here `retentionTimeInterpretation` as either minutes or seconds

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -0,0 +1,581 @@
...
+      const double rt_right_lim = spectrum_rt + getRTWindow() / 60.0 / 2.0;
+      const std::vector<Precursor> precursors = spectrum.getPrecursors();
+      if (precursors.size() < 1)
+      {
+        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
+                                         "Spectrum does not contain precursor info.");
+      }
+      const double spectrum_mz = precursors[0].getMZ();
+      const double mz_left_lim = spectrum_mz - getMZTolerance();
+      const double mz_right_lim = spectrum_mz + getMZTolerance();
```
hroest: why is the RT in a window and needs to be divided by 2.0 and the m/z not? I would prefer consistency

hroest: PS: it may also be useful to express m/z tolerance in ppm

hroest:
I have made a few comments and a question regarding public methods: it seems that the main method `extractSpectra` calls all functions in the right order and is basically what should be the public interface. Therefore, would it make sense to make all other methods private as they should not be called from outside the class or do you see a usecase where you want to call the sub-methods and not the `extractSpectra` method?

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
pasqoo: RT is a window (hence divided by two) while m/z is a tolerance (hence +/-).

hroest:
@jpfeuffer can you have a look at why this does not compile on Windows? cdash is very unhelpful

```
                 with
                 [
                     _Ty=std::pair<double,double>
                 ]
         /.../source/src/openms/source/ANALYSIS/OPENSWATH/MRMRTNormalizer.cpp(71): note: see reference to function template instantiation 'std::vector<std::pair<double,double>,std::allocator<_Ty>> OpenMS::Math::RANSAC<OpenMS::Math::RansacModelLinear>::ransac(const std::vector<_Ty,std::allocator<_Ty>> &,::size_t,::size_t,double,::size_t,bool,int (__cdecl *)(int))' being compiled
                 with
                 [
                     _Ty=std::pair<double,double>
                 ]
         /.../source/src/openms/source/FILTERING/CALIBRATION/MZTrafoModel.cpp(209): note: see reference to class template instantiation 'OpenMS::Math::RANSAC<OpenMS::Math::RansacModelLinear>' being compiled

     4>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\include\xhash(28): error C2440: 'type cast': cannot convert from 'const OpenMS::String' to '::size_t' [C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\build\src\openms\OpenMS.vcxproj]

         C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\include\xhash(28): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
         C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\contrib_build\include\boost/functional/hash/extensions.hpp(262): note: see reference to function template instantiation 'size_t stdext::hash_value<T>(const _Kty &)' being compiled
                 with
                 [
                     T=OpenMS::String,
                     _Kty=OpenMS::String
                 ]
         C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\contrib_build\include\boost/functional/hash/extensions.hpp(261): note: while compiling class template member function 'size_t boost::hash<K>::operator ()(const T &) const'
                 with
                 [
```

pasqoo:
@hroest It's desired to keep the methods as they are (concerning public or private) so that the user can call them in different orders for debugging purposes.  For example, calling `selectSpectra()` with or without the `FeatureMap` to compare how the scores are working.

I'll soon implement the `mz_tolerance` ppm support and then I think I have taken care of all the comments.

jpfeuffer:
I think it has something to do with that you can only use unordered_map on objects that can be implicitly cast to a size_t (for the hash I guess) and since here an OpenMS::String is the input it fails to do two implicit conversions. Not sure why gcc can handle it or Windows can not.

pasqoo:
@hroest @timosachsenberg I should be done modifying the code. Thanks for your feedback.
I added a parameter `rt_unit_` which informs the class of which unit (minutes or seconds) is used when loading the data with `TransitionTSVReader`.

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -0,0 +1,429 @@
...
+    peak_height_min_ = (double)param_.getValue("peak_height_min");
+    peak_height_max_ = (double)param_.getValue("peak_height_max");
+    fwhm_threshold_ = (double)param_.getValue("fwhm_threshold");
+
+    tic_weight_ = (double)param_.getValue("tic_weight");
+    fwhm_weight_ = (double)param_.getValue("fwhm_weight");
+    snr_weight_ = (double)param_.getValue("snr_weight");
+  }
+
+  void TargetedSpectraExtractor::getDefaultParameters(Param& params)
```
timosachsenberg: please check this method again.
e.g.: rt_window still has written "seconds" in the description.

I personally still think that it would make sense to "forward" the parameters from the underlying algorithms for better maintainability. But I leave that to you to decide.

timosachsenberg: Also, consider looking at other tools parameter names. It would be nicer to make these consistent throughout OpenMS.

Description is - in parts - a bit short and could be extended (e.g., what does minimum forward match mean? what is the unit? Is it just a score? etc.)

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -0,0 +1,442 @@
...
+  )
+  {
+    annotated_spectra.clear();
+    features.clear(true);
+    const std::vector<ReactionMonitoringTransition> transitions = targeted_exp.getTransitions();
+    for (UInt i=0; i<spectra.size(); ++i)
+    {
+      MSSpectrum spectrum = spectra[i];
+      const double divisor = (String)param_.getValue("rt_unit") == "seconds" ? 1.0 : 60.0;
+      const double spectrum_rt = spectrum.getRT() / divisor;
```
hroest: I am not really sure whether that is correct, I `spectrum.RT()` is always in seconds, see
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/KERNEL/MSSpectrum.h#L172

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -0,0 +1,442 @@
...
+        if (!peptide.rts.empty() &&
+            peptide.rts[0].hasCVTerm("MS:1000895") &&
+            !peptide.rts[0].getCVTerms()["MS:1000895"].empty())
+        {
+          target_rt = peptide.rts[0].getCVTerms()["MS:1000895"][0].getValue().toString().toDouble();
+        }
+        else
+        {
+          target_rt = peptide.getRetentionTime(); // accesses to MS:1000896. TODO: update if "guesstimate" in TargetedExperimentHelper.h is fixed
+        }
```
hroest: this is fixed in #3043

pasqoo: I did not think of that, thanks! I also prefer to keep everything in seconds since OpenMS does that by default.

pasqoo:
@timosachsenberg I modified the code trying to use the "parameter forward" methodology you recommended. Is it the right way to do it?

// TODO to myself: update the comments inside header code, which still refer to getters/setters

pasqoo:
@timosachsenberg @hroest The PR is ready again.

pasqoo:
I will update this PR to reflect the changes merged in https://github.com/OpenMS/OpenMS/pull/3043

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -162,21 +158,9 @@ namespace OpenMS
...
-          target_rt = peptide.rts[0].getCVTerms()["MS:1000895"][0].getValue().toString().toDouble();
-        }
-        else
-        {
-          target_rt = peptide.getRetentionTime(); // accesses to MS:1000896. TODO: update once https://github.com/OpenMS/OpenMS/pull/3043 is merged
-        }
-        target_rt *= (String)param_.getValue("rt_unit") == "seconds" ? 1.0 : 60.0;
-
+        double target_rt = peptide.getRetentionTime();
+        target_rt *= peptide.getRetentionTimeUnit() == TargetedExperimentHelper::RetentionTime::RTUnit::SECOND ? 1.0 : 60.0;
```
timosachsenberg: I find:

if (peptide.getRetentionTimeUnit() == TargetedExperimentHelper::RetentionTime::RTUnit::MINUTES) { target_rt *= 60.0; }; 
a bit easier readable in this specific case

timosachsenberg: but that's just personal taste

pasqoo: That totally makes sense :+1: really no need to multiply something times 1 :sweat_smile:

pasqoo:
@timosachsenberg @hroest The PR is up to date with the proposed changes. What do you think about its current state?

src/tests/class_tests/openms/source/TargetedSpectraExtractor_test.cpp
```diff
@@ -0,0 +1,698 @@
...
+// measured on a QTRAP 5500 corresponding to C-Aconitate
+// taken from E. coli grown on glucose M9 during steady-state
+// for flux analysis.
+
+MSSpectrum spectrum;
+spectrum.resize(73);
+MSSpectrum::Iterator it = spectrum.begin();
+
+it->setMZ(61.92);
+it++->setIntensity(6705.41660838088f);
```
timosachsenberg: looks a bit strange to my eyes. Other tests just load a dta file.
If you want to have these values in a human readable way in the test you could just use intitializer lilsts
and add the peaks in a loop. 
e.g.:
```
  std::vector<int> mz = {1,2,3}; 
  std::vector<int> intensity= {1,2,3}; 
  for (Size i = 0; i != mz.size(); ++i) { spectrum.push_back(Peak1D(mz[i],intensity[i])); }
```

Would be shorter and imho a bit easier to read.

pasqoo: Thanks, I modified it and I also made sure that I use `Size` instead of `UInt` when dealing with `.size()` return values.

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+      LOG_DEBUG << "[" << i << "]\trt: " << spectrum_rt << "\tmz: " << spectrum_mz << std::endl;
+
+      for (Size j=0; j<transitions.size(); ++j)
+      {
+        const TargetedExperimentHelper::Peptide peptide = targeted_exp.getPeptideByRef(transitions[j].getPeptideRef());
+        double target_rt = peptide.getRetentionTime();
+        if (peptide.getRetentionTimeUnit() == TargetedExperimentHelper::RetentionTime::RTUnit::MINUTE)
+        {
+          target_rt *= 60.0;
```
hroest: :+1: this looks good now :-)

hroest:
well it all looks good to me except Windows compile failures. see what @jpfeuffer said:

> I think it has something to do with that you can only use unordered_map on objects that can be implicitly cast to a size_t (for the hash I guess) and since here an OpenMS::String is the input it fails to do two implicit conversions (to std::string and then to size_t). Not sure why gcc can handle it or Windows can not.
> Solution could be an own hash function or explicit casting to std::string.

please address this, it needs to compile on Windows as well before we merge.

hroest:
@jpfeuffer can you look at this test failure again, it is not clear to me how the code location (MRMRTNormalizer) is related to this PR ?

hroest:
build win

hroest:
rebuild win

hroest:
@jpfeuffer can you rebuild win ?

pasqoo:
@timosachsenberg @hroest @jpfeuffer I'm not sure the build fails because of my code. Checking the PR page also shows a lot of red crosses. Could it be caused by something else?
What I did in my last commit is basically use `std::string` instead of `String`, but the build still fails.

hroest:
hmm, it looks like the failurse are related to this PR since other PR build just fine on Windows, e.g. https://github.com/OpenMS/OpenMS/pull/3090 is all green. The reason you see so many failed tests is that the library does not build and therefore all tests cannot be built either and fail. However I think there is something strange going on with the error message reported on cdash, maybe @jpfeuffer can provide some help here?

timosachsenberg:
I checked the build log and there it seems it is complaining about a different position:
```         C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\contrib_build\include\boost/unordered/unordered_map.hpp(68): note: see reference to class template instantiation 'boost::unordered::detail::table_impl<boost::unordered::detail::map<A,K,T,H,P>>' being compiled
                 with
                 [

                     A=std::allocator<std::pair<const OpenMS::String,OpenMS::Residue *>>,

                     K=OpenMS::String,

                     T=OpenMS::Residue *,

                     H=boost::hash<OpenMS::String>,

                     P=std::equal_to<OpenMS::String>

                 ]

         C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\source\src\openms\include\OpenMS/CHEMISTRY/ResidueDB.h(213): note: see reference to class template instantiation 'boost::unordered::unordered_map<OpenMS::String,OpenMS::Residue *,boost::hash<K>,std::equal_to<K>,std::allocator<std::pair<const K,T>>>' being compiled

                 with

                 [

                     K=OpenMS::String,

                     T=OpenMS::Residue *

                 ]
```

hroest:
and on cdash it complains again about another, unrelated position: http://cdash.openms.de/viewBuildError.php?buildid=249518

```
Error 	

                 with
                 [
                     _Ty=std::pair<double,double>
                 ]
         /.../source/src/openms/source/ANALYSIS/OPENSWATH/MRMRTNormalizer.cpp(71): note: see reference to function template instantiation 'std::vector<std::pair<double,double>,std::allocator<_Ty>> OpenMS::Math::RANSAC<OpenMS::Math::RansacModelLinear>::ransac(const std::vector<_Ty,std::allocator<_Ty>> &,::size_t,::size_t,double,::size_t,bool,int (__cdecl *)(int))' being compiled
                 with
                 [
                     _Ty=std::pair<double,double>
                 ]
         /.../source/src/openms/source/FILTERING/CALIBRATION/MZTrafoModel.cpp(209): note: see reference to class template instantiation 'OpenMS::Math::RANSAC<OpenMS::Math::RansacModelLinear>' being compiled

     4>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\include\xhash(28): error C2440: 'type cast': cannot convert from 'const OpenMS::String' to '::size_t' [C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\build\src\openms\OpenMS.vcxproj]

         C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\include\xhash(28): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
         C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\contrib_build\include\boost/functional/hash/extensions.hpp(262): note: see reference to function template instantiation 'size_t stdext::hash_value<T>(const _Kty &)' being compiled
                 with
                 [
                     T=OpenMS::String,
                     _Kty=OpenMS::String
                 ]
         C:\jenkins\ws\openms_pr_windows_ci\979a8a2f\contrib_build\include\boost/functional/hash/extensions.hpp(261): note: while compiling class template member function 'size_t boost::hash<K>::operator ()(const T &) const'
                 with
                 [
```

whats happening here?

timosachsenberg:
I added the full output - seems to be the same line that errors (was just truncated)

timosachsenberg:
can you try to replace std::unordered_map with boost::unordered_map?
or even better: replace all boost::unordered_map in OpenMS by std::unordered_map?

hroest:
lets try  `boost::unordered_map` here, then merge and fix this later on, e.g. in #3092

hroest:
thanks @pasqoo for your patience with this, this is something we need to fix inside OpenMS but it will take some effort to remove all boost::unordered_map. Until then I guess we cannot use both and we are stuck with the boost version.

pasqoo:
Happy to see you found the source of the problem! Thank you all for the support. I'll keep this detail about `unordered_map` in mind for the next features/PRs. I'll also probably ask some questions about it on Gitter :smile:

PR #3021, opened 2017-11-24
Feature/MRMFeatureQC

dmccloskey:
Filtering or Flagging of MRM-based FeatureMaps based on user criteria.

replaces #3004 (see there for in-depth discussion)


dmccloskey:
For the life of me, I am not able to get the pyopenms build to work.  I keep getting the following very unhelpful error: `Exception: no converter for component_name`.  Do you have any idea what is causing this?

hroest:
I will look into it, but are you sure that 2f7f5ec does not build? I had it work on my computer before I created the PR. But lets see what travis says

hroest:
@dmccloskey as you can see here https://travis-ci.org/OpenMS/OpenMS/builds/306784997 the commit 2f7f5ec pretty much runs through, pyOpenMS builds and only the memoryleak test fails. However that is not too bad, so I think with 2f7f5ec we have a working PR. Maybe we can just revert to that version and then we should be good to go?

dmccloskey:
I am fine reverting to that version.  The other pushes were just me trying to figure out why pyopenms would not build.

hroest:
ok, can you make a new commit reverting to that version and then pushing it here to let it build?

dmccloskey:
All checks look to be passing.  I have no idea why the build broke last time then.  Do you know what changed in the pyopenms build?

hroest:
I am not aware that anything changed, but either way 2f7f5ec  works and we can use it (also win looks okay with a few known unrelated failures).

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -0,0 +1,408 @@
...
+      // iterate through each component/sub-feature
+      for (size_t sub_it = 0; sub_it < features[feature_it].getSubordinates().size(); ++sub_it)
+      {
+        String component_name = (String)features[feature_it].getSubordinates()[sub_it].getMetaValue("native_id"); 
+        std::cout << "component_name" << component_name << std::endl; //debugging
+        bool c_qc_pass = true;
+
+        // iterate through multi-feature/multi-sub-feature QCs/filters
+        // iterate through component_groups
+        for (size_t cg_qc_it = 0; cg_qc_it < filter_criteria.component_group_qcs.size(); ++cg_qc_it)
```
hroest: It may make senseto replace this with a lookup, e.g. store the criteria in a map of `std::map<std::string, *ComponentGroupQC>` or similar

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -0,0 +1,408 @@
...
+        {
+          if (filter_criteria.component_group_qcs[cg_qc_it].component_group_name == component_group_name)
+          {
+            // labels and transition counts QC
+              std::cout << "n_heavy" << std::endl; //debugging
```
hroest: maybe replace these calls with `LOG_DEBUG << `

hroest:
seems like it all passes, I have a few comments regarding debug output and making the algo a bit more efficient. But we can also do that in the second iteration

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
dmccloskey: I actually meant to comment out those lines.  Apparently, I missed those two.

hroest: you can do it in the next PR

jpfeuffer:
http://cdash.openms.de/viewBuildError.php?buildid=247249
There seem to be errors related to this in the nightlies.

dmccloskey:
fix is in this pull request: #3032

PR #3024, opened 2017-11-24
[FIX] In XFDR, fix extraction of first rank IDs if input id IdXML and…

timosachsenberg:
… MzIndentML (#3000)

* [FIX] In XFDR, fix extraction of first rank IDs if input id IdXML and MzIdentML

* [REFACTOR] Some more code refinements in XFDR

* [FIX] In XFDR, change arg for nminionsmatched to unsigned int

* Update XFDR.cpp


enetz:
@timosachsenberg only time outs, should we try to restart tests or merge?

enetz:
only XFDR was changed and it did not work anyway, so it can only improve. I will merge.

PR #3025, opened 2017-11-24
[FIX] Fixed SVMWrapper_test

jpfeuffer:
I don't know, why this only shows up on macOS and why only since some time but:
IMHO the test was designed wrongly. According to my short research, at a tie, LibSVM assigns the first encountered
label in the training data, which is a different one in the second test.
Let's see what other platforms say.

### DIFF
--- a/src/tests/class_tests/openms/source/SVMWrapper_test.cpp
+++ b/src/tests/class_tests/openms/source/SVMWrapper_test.cpp
@@ -1041,8 +1041,9 @@ START_SECTION((void getSVCProbabilities(struct svm_problem *problem, std::vector
 							|| (predicted_labels[i] > 0 && probabilities[i] >= 0.5), true)
 	}
 	labels.clear();
-	labels.resize(4, -1);
-	labels.resize(8, 1);
+	// Start with -1 as "first" label
+	labels.resize(count / 2, -1);
+	labels.resize(count, 1);
 	problem = encoder.encodeLibSVMProblem(encoded_vectors, labels);
 	svm.train(problem);
 	svm.predict(problem, predicted_labels);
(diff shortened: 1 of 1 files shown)


hroest:
since we dont have OSX on travis we almost have to merge and see what happens?

jpfeuffer:
Yes. I just wanted to make sure that other platforms work. I tested it on the Mac machine. Pretty sure it passes now, there.

jpfeuffer:
Wait for Win here.

src/tests/class_tests/openms/source/SVMWrapper_test.cpp
```diff
@@ -1051,8 +1052,10 @@ START_SECTION((void getSVCProbabilities(struct svm_problem *problem, std::vector
 	TEST_EQUAL(predicted_labels.size() == probabilities.size(), true)
 	for (Size i = 0; i < predicted_labels.size(); ++i)
 	{
-		TEST_EQUAL((predicted_labels[i] < 0 && probabilities[i] < 0.5)
-							|| (predicted_labels[i] > 0 && probabilities[i] >= 0.5), true)
+		// At probability 0.5, LibSVM will assign the first encountered label in the training data
+		// in this case "-1"
+		TEST_EQUAL((predicted_labels[i] < 0 && probabilities[i] <= 0.5)
```
timosachsenberg: Cool. Thanks for checking. I was always assuming this is related to a specific version!

jpfeuffer:
Win errors unrelated. Merge.

hroest:
seems like two tests failed:

```
projectroot.src.tests.topp.TOPP_PercolatorAdapter_1_out1 | 75 ms | 9
projectroot.src.tests.topp.TOPP_FeatureFinderIdentification_2_out1 | 0.17 sec | 10
```

so thats already an improvement, lets hope we can fix the rest of the errors soon

PR #3026, opened 2017-11-24
[PYOPENMS,BUILD] add option to disable test

hroest:
disable pyOpenMS memory leak test on shared infrastructure


hroest:
since we have seen some failures on travis with that test and it does not make a lot of sense to have it enabled on shared infrastructure where total memory may vary a lot. Therefore lets disable it for travis (and lets first see whether this really does what we want it to do).

hroest:
seems like it worked, now 23 instead of 24 tests:

```
100% tests passed, 0 tests failed out of 23

```

PR #3028, opened 2017-11-25
[TEST] Disable Percolator test until new release on Win

jpfeuffer:
Alternative would be to generate a special output file to compare to on Win.
This here is easier, we just have to remember to remove it.


jpfeuffer:
Ok that was enough. I kept the Executing tests. If you think this is fine until we can update thirdparty, feel free to merge.

PR #3031, opened 2017-11-26
[Add] option precursor to MzMLSplitter 

oliveralka:
Resolves TODO: "precursor" 

The implementation works but i'm not sure about the memory allocation, since the size is reserved beforehand but I extend or reduce the original size dependent on the MSLevel of the last spectrum. 

line 173: `part.reserveSpaceSpectra(n_spec); `
line 176, 193: `part.addSpectrum(spectra[i]);`

I'm open to suggestions.


oliveralka:
Will add test.

src/utils/MzMLSplitter.cpp
```diff
@@ -86,8 +86,7 @@ class TOPPMzMLSplitter :
     setMinInt_("size", 0);
     registerStringOption_("unit", "<choice>", "MB", "Unit for 'size' (base 1024)", false);
     setValidStrings_("unit", ListUtils::create<String>("KB,MB,GB"));
-    // @TODO:
-    // registerFlag_("precursor", "Make sure precursor spectra end up in the same part as their fragment spectra");
+    registerFlag_("precursor", "Make sure precursor spectra end up in the same part as their fragment spectra");
```
timosachsenberg: different name? no a flag but maybe a string

src/utils/MzMLSplitter.cpp
```diff
@@ -155,76 +155,96 @@ class TOPPMzMLSplitter :
...
+      //chromatograms
+      if (n_chrom > 0)
       {
-        int last_spectrum = spec_start + n_spec;   
-        part.reserveSpaceSpectra(n_spec);  
-        for (Size i = spec_start; i < spec_start + n_spec; ++i)
+        parts_chrom_ranges.push_back(std::make_pair(chrom_start, chrom_start + n_chrom - 1));   
+      }
+
+      //spectra
```
timosachsenberg: I know it is pedantic but space after // is easier to read. Also would be nicer to be consistent in the use of lower / upper case for first letter in comments (e.g. Last spectrum vs. chromatograms)

src/utils/MzMLSplitter.cpp
```diff
@@ -86,7 +86,7 @@ class TOPPMzMLSplitter :
     setMinInt_("size", 0);
     registerStringOption_("unit", "<choice>", "MB", "Unit for 'size' (base 1024)", false);
     setValidStrings_("unit", ListUtils::create<String>("KB,MB,GB"));
-    registerFlag_("precursor", "Make sure precursor spectra end up in the same part as their fragment spectra");
+    registerFlag_("precursor", "The precursor spectrum and its corresponding MSn spectra will be saved in the same file");
```
timosachsenberg: I would rename this parameter. Also, it is not clear how setting the flag relates to the other options.

timosachsenberg:
is the option also tested on MS2 only data?
also on MS1 only?

oliveralka:
@timosachsenberg: Will edit the code as you suggested.
 It was tested on MS1, MS1&MS2 but not on MS2 only (will do that).

src/utils/MzMLSplitter.cpp
oliveralka: It is perfectly fine to point that out - thanks.

oliveralka:
Will close and rewrite due to unforeseen difficulties with my algorithm outline. Further approach was discussed with @timosachsenberg.

PR #3034, opened 2017-11-27
[FIX, TEST] attempt to fix ffid test

timosachsenberg:


timosachsenberg:
I removed the identification that caused the difference. Let's see if this resolves the issue.

jpfeuffer:
Great. Fine for me, although it would be cool if there would be a more detailed explanation on why there are differences.

jpfeuffer:
By the way: @timosachsenberg did you check if it was the same feature that was causing macOS nightlies to fail?

timosachsenberg:
Yes, it is the same feature. I have no idea why this feature creates this problem. Maybe it is related to the SVM differences between platforms? It could be that the SVM scored both hits of the same elution profile with the same score and on one platform the first hit is selected for targeted extracted, on the other platform the latter.

timosachsenberg:
failed tests unrelated

jpfeuffer:
Sorry, but this did not solve the issue. Better check CDash for the PRs until I can confirm that Jenkins is stable...

timosachsenberg:
hmm seems to be a different issue?

PR #3040, opened 2017-11-28
[FIX] FeatureFinderIdentification disable test as we get no stable result on the different platforms

timosachsenberg:


timosachsenberg:
let's see if disabling a single test is sufficient

PR #3041, opened 2017-11-28
[FIX] test input conversion errors in MRMFeatureQCFile

dmccloskey:
Implemented testing for empty strings before converting to double or integer values from .csv user input.

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
@@ -167,7 +167,7 @@ namespace OpenMS
 
               // find the ion ratio pair
               if (filter_criteria.component_group_qcs[cg_qc_it].ion_ratio_pair_name_1 != ""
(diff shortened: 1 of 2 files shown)


src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -166,7 +166,9 @@ namespace OpenMS
               String component_name2 = (String)features[feature_it].getSubordinates()[sub_it2].getMetaValue("native_id"); 
 
               // find the ion ratio pair
-              if (filter_criteria.component_group_qcs[cg_qc_it].ion_ratio_pair_name_1 == component_name
+              if (filter_criteria.component_group_qcs[cg_qc_it].ion_ratio_pair_name_1 != ""
+                && filter_criteria.component_group_qcs[cg_qc_it].ion_ratio_pair_name_1 != ""
```
timosachsenberg: typo: _1 should be _2 in the second line I guess

dmccloskey: Typo indeed.  I will correct that now.

PR #3043, opened 2017-11-28
Feature/traml rt

### LINKED ISSUE #3038: TargetedExperimentHelper::getRetentionTime() assumes "MS:1000896" exists
While working on https://github.com/OpenMS/OpenMS/pull/3014 I noticed that `TargetedExperimentHelper::getRetentionTime()`
https://github.com/OpenMS/OpenMS/blob/cfcf9ff09613191aa3502dbf5c6df9a613379376/src/openms/include/OpenMS/ANALYSIS/TARGETED/TargetedExperimentHelper.h#L384
assumes that the RT info is saved in `"MS:1000896"`.
In the comment inside the code it is reported that it is a "guesstimate", but the code actually accesses the key before verifying it exists in the map (for example with a `.hasCVTerm("MS:1000896")` or `getCVTerms().has("MS:1000896")`).

One case where the RT info is saved somewhere else is when one sets `retentionTimeInterpretation` parameter from `TransitionTSVReader` as `"minutes"` or `"seconds"`. In those two cases, the RT info is saved in `"MS:1000895"`.
Here the code: https://github.com/OpenMS/OpenMS/blob/2eb48eac4dbd03b7e7ededb4d6e8dd64f92edb25/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVReader.cpp#L1016

More info:
OpenMS version: the issue is present in `master` and `develop`
Platform: Ubuntu 17.10

hroest:
fixes #3038

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/TARGETED/TargetedExperimentHelper.h
+++ b/src/openms/include/OpenMS/ANALYSIS/TARGETED/TargetedExperimentHelper.h
@@ -144,14 +144,45 @@ namespace OpenMS
     {
 public:
 
+      enum RTUnit
+      {
+        Second = 0,        // seconds
+        Minute,            // minutes
+        unit_Unannotated,  // no stored annotation
+        SizeOfRTUnit
+      };
+
+      enum RTType
+      {
+        Local,            //
+        Normalized,       //
+        Predicted,        //
+        HPINS,            // H-PINS "The de facto standard providing the retention times"
+        iRT,              // iRT retention time standard
+        type_Unannotated, // no stored annotation
+        SizeOfRTType
+      };
+
       RetentionTime() :
-        CVTermListInterface()
+        CVTermListInterface(),
+        software_ref(""),
+        retention_time_unit(RTUnit::SizeOfRTUnit),
+        retention_time_type(RTType::SizeOfRTType),
+        retention_time_set(false),
+        retention_time(0.0)
+        // retention_time_width(0.0),
+        // retention_time_lower(0.0),
+        // retention_time_upper(0.0)
       {
       }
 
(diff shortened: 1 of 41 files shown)


src/openms/include/OpenMS/ANALYSIS/TARGETED/TargetedExperimentHelper.h
```diff
@@ -139,19 +139,60 @@ namespace OpenMS
...
+      actual value is stored in retention_time_ and can be accessed by getRT /
+      setRT. Currently support for RT windows or lower/upper limits is not
+      implemented but is available via CV terms.
+    */
     class OPENMS_DLLAPI RetentionTime :
       public CVTermListInterface
     {
 public:
 
+      enum RTUnit
```
timosachsenberg: consider uppercase enum entries

timosachsenberg: consider enum class RTUnit

src/openms/include/OpenMS/ANALYSIS/TARGETED/TargetedExperimentHelper.h
```diff
@@ -139,19 +139,60 @@ namespace OpenMS
...
+      {
+        Second = 0,        // seconds
+        Minute,            // minutes
+        unit_Unannotated,  // no stored annotation
```
timosachsenberg: we often use UNKNOWN

jpfeuffer:
Sorry, I guess there was a problem downloading the prebuilt contrib.

jpfeuffer:
rebuild win

hroest:
all seems to work - can we merge? @timosachsenberg ?

Issue #905, opened 2014-07-20
ResidueDB not thread safe.


timosachsenberg:
e.g. parsing of AASequences containing modifications with multiple threads causes segmentation faults.
Reason is that modified residues are implicitly added to the ResidueDB singleton if they don't exist.
#5  0x00007ffff6b95586 in OpenMS::Map<OpenMS::String, OpenMS::Map<OpenMS::String, OpenMS::Residue*> >::operator[](this=0x7678e8, key=...)

```
at /abi-data/sachsenb/OpenMS_IDE/src/openms/include/OpenMS/DATASTRUCTURES/Map.h:127
```
#6  0x00007ffff6b92179 in OpenMS::ResidueDB::addResidue_ (this=0x7670b0, r=0x7fffdc004400) at /abi-data/sachsenb/OpenMS_IDE/src/openms/source/CHEMISTRY/ResidueDB.cpp:161
#7  0x00007ffff6b94aae in OpenMS::ResidueDB::getModifiedResidue (this=0x7670b0, residue=0x846070, modification=...)

```
at /abi-data/sachsenb/OpenMS_IDE/src/openms/source/CHEMISTRY/ResidueDB.cpp:483
```
#8  0x00007ffff6b5b00a in OpenMS::AASequence::setModification (this=0x7fffdc004ad0, index=16, modification=...)

```
at /abi-data/sachsenb/OpenMS_IDE/src/openms/source/CHEMISTRY/AASequence.cpp:902
```
#9  0x00007ffff70a8be9 in OpenMS::ModifiedPeptideGenerator::applyFixedModifications (fixed_mods_begin=..., fixed_mods_end=..., peptide=...)

```
at /abi-data/sachsenb/OpenMS_IDE/src/openms/source/ANALYSIS/RNPXL/ModifiedPeptideGenerator.cpp:71
```
#10 0x0000000000497534 in SimpleSearchEngine::main_ (.omp_data_i=0x7fffffffaf40) at /abi-data/sachsenb/OpenMS_IDE/src/utils/SimpleSearchEngine.cpp:582
#11 0x00007ffff400ae3a in gomp_thread_start (xdata=<value optimized out>) at ../.././libgomp/team.c:115
#12 0x00000033854079d1 in start_thread () from /lib64/libpthread.so.0
#13 0x00000030bd2e8b6d in clone () from /lib64/libc.so.6

hroest:
Related to #957

hendrikweisser:
[...]

PR #3046, opened 2017-11-29
minor fixes to PeptideIndexer

### LINKED ISSUE #2594: PeptideIndexer: "unknown AA" error due to selenocysteine

cbielow:
[FIX] fix PeptideIndexer crash (division by 0) if 0 peptides are mapped to the protein DB.

[FEATURE] parse protein sequences which contain modifications as AASequence (since otherwise hits in regions with modfied AA would not be found) and emit a warning to user to fix his Protein Database (fixes #2594)

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
@@ -37,7 +37,6 @@
 
 
 #include <OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h>
-#include <OpenMS/CHEMISTRY/AASequence.h>
 #include <OpenMS/CHEMISTRY/ProteaseDigestion.h>
 #include <OpenMS/CHEMISTRY/ProteaseDB.h>
 #include <OpenMS/CONCEPT/LogStream.h>
@@ -225,6 +224,8 @@ public:
       std::vector<bool> protein_is_decoy; // protein index -> is decoy?
       std::vector<std::string> protein_accessions; // protein index -> accession
 
+      bool invalid_protein_sequence = false; // check for proteins with modifications, i.e. '[' or '(', and throw an exception
+
       { // new scope - forget data after search
       
         /*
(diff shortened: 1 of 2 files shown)


src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -328,6 +329,14 @@ namespace OpenMS
               String prot = proteins.chunkAt(i).sequence;
 
               prot.remove('*');
+
+              // check for invalid sequences with modifications
+              if (prot.has('[') || prot.has('('))
+              { 
+                 invalid_protein_sequence = true; // not omp-critical because its write-only
+                 prot = AASequence::fromString(prot).toUnmodifiedString();
```
timosachsenberg: I think this is not thread-safe (and can't easily be fixed (without huge performance penalties) because ResidueDB is not thread-safe).
I am pretty sure that this might lead to a concurrent read/write to ResidueDB.

timosachsenberg: see https://github.com/OpenMS/OpenMS/issues/905

timosachsenberg: please correct me if I am wrong

cbielow: you're right. So I'm going to elevate to error for now until we have a solution for ResidueDB.

cbielow:
ready to merge

Issue #3042, opened 2017-11-28
TransitionTSVReader::createPeptide_: Problem with E(Cation:Cu[I])


grosenberger:
This issue was originally reported by @bretttully. TargetedFileConverter parses modified peptide sequences with support for various formats.

- TSV/CSV files can contain modifications in TPP, UniMod or PSI-MOD formats.
- TraML files store modifications as UniMod (e.g. S(UniMod:21)) modifications
- PQP files currently employ PSI-MOD modifications (e.g. S(Phospho))

When generating a PQP file from TSV or TraML, OpenMS::AASequence converts all potential formats to PSI-MOD, which is then stored. However, when converting back, TargetedFileConverter runs into problems with complex modifications, e.g. E(Cation:Cu[I]) as it wants to parse the brackets as well in TransitionTSVReader::createPeptide_.

How should we store PTMs in OpenMS in general (and in PQP files)? Is PSI-MOD or UniMod preferred?

timosachsenberg:
I repeatedly heard that PSI-MOD is basically dead/unmaintained.

grosenberger:
Alright, then I guess that storing PTMs in UniMod format would be preferable (and consistent with TraML) and I will prepare a PR accordingly.

hroest:
yes, I would also suggest UniMod only and otherwise we just store the mass, e.g. E[34.45545]

hroest:
PS: it would also be nice if the parser would not break on things like `E(Cation:Cu[I])` - so maybe we need to correctly parse square brackets inside round brackets, right? Basically would a test be something like  `AASequence("PE(Cation:Cu[I])PTIDE` to check this?

PR #3054, opened 2017-11-30
[FIX] OpenSWATH: Store UniMod sequence in PQP files

grosenberger:
This PR addresses issue https://github.com/OpenMS/OpenMS/issues/3042 reported by @bretttully. TargetedFileConverter now always stores modifications in UniMod format in PQP files, which improves compatibility with TraML.

This is implemented by the new function AASequence::toUniModString(), which works analogously to AASequence::toString(), except that known modifications are reported in UniMod format rather than the internal nomenclature.

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/AASequence.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/AASequence.h
@@ -414,6 +414,16 @@ protected:
     String toUnmodifiedString() const;
 
     /**
+        @brief returns the peptide as string with UniMod-style modifications embedded in brackets
+
+        Uses round brackets when possible (id is known) or square brackets for
+        unknown modifications where only the mass is known.
+
+        i.e.: n[43]PEPC(UniMod:4)PEPM[147]PEPRc[16]
+    */
+    String toUniModString() const;
+
+    /**
         @brief create a TPP compatible string of the modified sequence using bracket notation.
 
         Instead of using the modification names, it writes the modification masses in brackets
--- a/src/openms/include/OpenMS/CHEMISTRY/AASequence.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/AASequence.h
@@ -406,7 +406,7 @@ protected:
         Uses round brackets when possible (id is known) or square brackets for
         unknown modifications where only the mass is known.
 
-        i.e.: n[43]PEPC(Carbamidomethyl)PEPM[147]PEPRc[16]
+        i.e.: .n[43]PEPC(Carbamidomethyl)PEPM[147]PEPR.[16]
     */
     String toString() const;
 
@@ -419,7 +419,7 @@ protected:
         Uses round brackets when possible (id is known) or square brackets for
         unknown modifications where only the mass is known.
 
-        i.e.: n[43]PEPC(UniMod:4)PEPM[147]PEPRc[16]
+        i.e.: .n[43]PEPC(UniMod:4)PEPM[147]PEPR.[16]
     */
     String toUniModString() const;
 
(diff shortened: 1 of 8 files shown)


src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -107,6 +107,68 @@ namespace OpenMS
...
+      else
+      {
+        bs += ".[" + String(mod.getDiffMonoMass()) + "]";
+      }
+    }
+
+    for (Size i = 0; i != seq.size(); ++i)
+    {
+      const Residue& r = seq[i];
+      const String aa = r.getOneLetterCode() != "" ? r.getOneLetterCode() : "X";
```
cbielow: just related to this PR: in which cases do we expect this to be empty? This is somewhat poorly documented within Residue class.

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -107,6 +107,68 @@ namespace OpenMS
...
+      else  // amino acid not modified
+      {
+        bs += aa;
+      }
+    }
+
+    if (seq.hasCTerminalModification())
+    {
+      const ResidueModification& mod = *(seq.getCTerminalModification());
+      unimod = mod.getUniModRecordId();
```
cbielow: this code is duplicated somewhat 3x...
IMHO it would make sense to write a private function `string getUnimod(const ResidueModification& mod, const char site) const` which assembles the partial string (`site` would be either `.` or the one-letter code).

hroest:
looks good to me, what about replacing 
https://github.com/OpenMS/OpenMS/blob/790e29ee4af30db3d4ae3ce9910155e3e5595627/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVReader.cpp#L1193

as well? that would make our output more consistent. Maybe we can do this at a later point?

src/openms/source/CHEMISTRY/AASequence.cpp
grosenberger: There are slight differences between the three parts (getDiffMonoMass vs getMonoWeight) and I'm thus not sure whether this would simplify the code.

cbielow: ok, you'll need a third parameter (weight), but this cuts down 50 lines of code into ~5.
Up to you of course :)

timosachsenberg:
https://www.codacy.com/app/hroest/OpenMS/pullRequest?prid=1121344

src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVReader.cpp
```diff
@@ -1228,27 +1228,8 @@ namespace OpenMS
...
-          for (Size modloc = 0; modloc < lightpep.modifications.size(); modloc++)
-          {
-            if (lightpep.modifications[modloc].location == loc)
-            {
-              mytransition.FullPeptideName += "(UniMod:" + String(lightpep.modifications[modloc].unimod_id) + ")";
-            }
-          }
-        }
-      }
+      mytransition.FullPeptideName = TargetedExperimentHelper::getAASequence(pep).toUniModString();;
```
timosachsenberg: double ;;

src/openms/include/OpenMS/CHEMISTRY/AASequence.h
```diff
@@ -413,6 +413,16 @@ namespace OpenMS
     /// returns the peptide as string without any modifications
     String toUnmodifiedString() const;
 
+    /**
+        @brief returns the peptide as string with UniMod-style modifications embedded in brackets
+
+        Uses round brackets when possible (id is known) or square brackets for
+        unknown modifications where only the mass is known.
+
+        i.e.: n[43]PEPC(UniMod:4)PEPM[147]PEPRc[16]
```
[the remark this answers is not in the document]
hroest: yes, I think it should be.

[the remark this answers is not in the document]
grosenberger: Good point, I changed this.

Issue #2497, opened 2017-04-04
X!Tandem reporting of non-tryptic peptides not reported/considered


cbielow:
Running X!Tandem (Vengeane 2015.12.15.2) in KNIME using `Trypsin` as enzyme yields peptides such as
`AKIDELINND`, which is clearly not tryptic on either C or N Terminus.
The N-Terminus I can understand somewhat, but why is 'D' (aspartate) considered a valid cutting site?
The protein is here:
```
>sp|Q12051|GGPPS_YEAST .... GN=BTS1 PE=1 SV=1
ME*AKIDELINND*PVWSSQNE....
```

I've searched for an explanation, and found `[X]|[D]` in the Tandem docs (http://www.thegpm.org/TANDEM/api/pcs.html) as possible cutting rule (aspn), but thats the N-terminal, not C-terminal and even checking the default.xml seems to specifiy Trypsin only.

So why does XTandem report these peptides? This has implications for PeptideIndexing, but if we dont understand the reasons, its hard to model why this should be allowed.

julianu:
I can confirm this behaviour of the recent X!Tandem releases (also see #2505) and would be interested in the reasoning behind it as well.

jpfeuffer:
First thing again: You say KNIME. Is it reproducible outside of KNIME? Plugin version: Stable or trunk?

Another thing is, that during refinement it checks for point mutations (if activated).

We really need to output the actual used parameters of XTandem if possible (see e.g. the output, parameters option). We could do this in debug mode only or add another parameter to the adapter.
I still haven't seen any real proof (e.g. in the source code) of the rumors that XTandem by default considers weird modifications etc. Actually I think it is all about the correct settings.
Maybe @hendrikweisser can comment on this, since he recently modified the adapter

We need to think about this for the upcoming release.

jpfeuffer:
Maybe in KNIME we still read the default_configuration file?
It says there:
```<note type="input" label="refine, unanticipated cleavage">yes</note>```

jpfeuffer:
That might be it. See the referenced issue.

julianu:
I used the stable KNIME releases.
The "refine, unanticipated cleavage" is, at least in my case, in the unused parameters section. Nevertheless, will make a test run with X!Tandem alone and come back here.

julianu:
Ah, reading the manual helps:
http://www.thegpm.org/tandem/release.html

Cyclone  (2011.12.01) release:
[...]

PR #3059, opened 2017-12-04
[Feature] Add xtandem cleavage rules

jpfeuffer:
Extends allow_methionine_cleavage to allow_nterm_protein_cleavage which also considers cleavage of the second AA (which can be X), following the XTandem standards.

Adds default option to allow for non-"missed cleavages affecting" random D|P cleavages, also following the observations from the XTandem author(s).


jpfeuffer:
Related to #2497

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -165,6 +166,10 @@ namespace OpenMS
       { // '0' is included in cleavage_positions, so starting fragments will be found as well
         spec_n = true;
       }
+      else if (allow_random_asp_pro_cleavage && (sequence[pos-1] == 'D') && (sequence[pos] == 'P'))
```
timosachsenberg: can pos be 0?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -174,6 +179,10 @@ namespace OpenMS
       {
         spec_c = true;
       }
+      else if (allow_random_asp_pro_cleavage && (sequence[end-1] == 'D') && (sequence[end] == 'P'))
```
timosachsenberg: can end-1 < 0?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
jpfeuffer: Yes but then it would be found in the first "if case" since 0 is always included in the cleavage_positions. I hope this assumption is not too loose.

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
jpfeuffer: no, see line 139. length >=1 and end = pos + length

jpfeuffer:
@cbielow Can you have a look if this is already correctly used in PeptideIndexer or if it needs changes?

src/openms/include/OpenMS/CHEMISTRY/EnzymaticDigestion.h
```diff
@@ -112,7 +112,7 @@ namespace OpenMS
     void digestUnmodified(const StringView& sequence, std::vector<StringView>& output, Size min_length = 1, Size max_length = 0) const;
 
     /// Returns true if the fragment starting at position @p pos with length @p length within the sequence @p sequence could be generated by the current model
-    bool isValidProduct(const String& sequence, Size pos, Size length, bool ignore_missed_cleavages = true) const;
+    bool isValidProduct(const String& sequence, Size pos, Size length, bool ignore_missed_cleavages = true, bool allow_random_asp_pro_cleavage = false) const;
```
timosachsenberg: Hmm so we have this also in the base class

timosachsenberg: this should not be in the base class of RNases and Proteases

jpfeuffer: Yes unfortunately that is the only place where I had access to the needed data. To avoid this, bigger restructuring would be necessary I guees.

timosachsenberg: @hendrikweisser any idea how we can get this hin without making the current design obsolete.

jpfeuffer: Ok, we discussed some changes for this offline. Wait for update.

jpfeuffer:
Please check again.

jpfeuffer:
Only Percolator fails, fixed in other Pr.

timosachsenberg:
+1

src/openms/include/OpenMS/CHEMISTRY/ProteaseDigestion.h
```diff
@@ -72,11 +72,11 @@ namespace OpenMS
     /// Returns the number of peptides a digestion of @p protein would yield under the current enzyme and missed cleavage settings.
     Size peptideCount(const AASequence& protein);
 
-    /// Variant of EnzymaticDigestion::isValidProduct() with support for methionine cleavage
-    bool isValidProduct(const String& protein, Size pep_pos, Size pep_length, bool ignore_missed_cleavages = true, bool methionine_cleavage = false) const;
+    /// Variant of EnzymaticDigestion::isValidProduct() with support for n-term protein cleavage and random D|P cleavage
+    bool isValidProduct(const String& protein, Size pep_pos, Size pep_length, bool ignore_missed_cleavages = true, bool allow_nterm_protein_cleavage = false, bool allow_random_asp_pro_cleavage = true) const;
```
timosachsenberg: one last question.... is allow_random_asp_pro_cleavage = true a sensible default?

timosachsenberg:
Julianus and I had a quick chat and we came to the conclusion that the allow_random_asp_pro_cleavage option might lead to unexpected results for non-xtandem searches (if enabled by default). We will see if we need another take on that issue and maybe just enable the allow_random_asp_pro_cleavage if a XTandem search is indexed in PeptideIndexer

src/openms/include/OpenMS/CHEMISTRY/ProteaseDigestion.h
hroest: I dont think so, better to make it false by default

jpfeuffer:
Changing defaults right now: What about the second AA at protein n-term. Should this be standard with allow_nterm_cleavage?
Or do we need another flag or even an integer?

jpfeuffer:
In conjunction with enzymes that cut after D it works, too.
The real cutting sites take preference and are counted as missed cleavages if skipped.

jpfeuffer:
Any complaints on the defaults?

timosachsenberg:
looks good and seems to be well covered by tests +1

Issue #2601, opened 2017-05-12
LowMemPeakPickerHiRes_RandomAccess underscore in name?


jpfeuffer:
```
     [java] Invalid ctd-format: LowMemPeakPickerHiRes_RandomAccess.ctd
     [java] Reason: The node name "LowMemPeakPickerHiRes_RandomAccess" in file "/Users/aiche/KNIME_PACKAGING_TRUNK/assembly/final/descriptors/LowMemPeakPickerHiRes_RandomAccess.ctd" is invalid.
```

We should rename the tool to show up in KNIME.

PR #3062, opened 2017-12-05
[FIX] renamed LowMemPeakPickerHiRes_RandomAccess to

timosachsenberg:
LowMemPeakPickerHiResRandomAccess

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,6 +31,9 @@ New tools:
 Deprecated and removed tools:
   - None
 
+Renamed tool:
+  - LowMemPeakPickerHiRes_RandomAccess was renamed to LowMemPeakPickerHiResRandomAccess
+
 Major changes in functionality:
   - Experimental design
     - Add support for fractions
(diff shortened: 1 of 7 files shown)


jpfeuffer:
Ah great. Fixes #2601 if it passes.

jpfeuffer:
Looks good. Should pass. Let's see.

hroest:
well, `./src/utils/LowMemPeakPickerHiRes.cpp` and `./src/utils/LowMemPeakPickerHiRes_RandomAccess.cpp` are not really useful tools and more used for testing whether these algorithms compile and run through properly. 

I dont think that they are useful to anybody, the `LowMemPeakPickerHiRes.cpp` is basically implemented in the `PeakPickerHiRes` anyways, so we could just remove that. We could move the random access part either also into the regular PPHiRes or we could move it to a unit-test file

timosachsenberg:
@hroest sounds good. we are a bit on a tight schedule for the release so I would suggest to remove it in develop for the next release

timosachsenberg:
I would suggest we merge this PR as a quick fix for https://github.com/OpenMS/OpenMS/issues/2601
In 2.4 we proceed with removal

jpfeuffer:
OK sounds good. Reduces confusion for the user.

jpfeuffer:
Only Percolator fails + timeouts. Win is busy with contrib and other PRs.

PR #3068, opened 2017-12-06
[STYLE] added override (c11)

timosachsenberg:


jpfeuffer:
Weird. Once travis timed-out, once it only took 15 mins.
Not sure about win-ci. Could be a random error.

timosachsenberg:
I think this might be related to ccache?

timosachsenberg:
hmm I am wondering why the contrib is updated? did I used the wrong submodule checkout?

jpfeuffer:
I don't know what you did :) Usually one should not touch the submodules unless you actually want to update or use it. Seems like you actually updated to the newest commit which is fine. Didn't you ask me how to do it recently? Maybe you tried the git pull and forgot to revert the changes before commiting?

hroest:
> I think this might be related to ccache?

yes. If you change all the files at once, the cache wont work any more and you will get these errors. Unfortunate but this is how it works - it also means we probably cannot do a full build from scratch any more on travis as it will always time out :-1:

PR #3069, opened 2017-12-06
Modernize2/use nullptr

timosachsenberg:


timosachsenberg:
just a quick experiment with clang-tidy modernize options...

hroest:
a lot of the library seems to build, lets see what travis says. There are some issues however with code style now

hroest:
@jpfeuffer how come the window build for 52b41a66b40ca6c0f140b8fb05c5b17c5cec0449 runs through: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_windows_ci/compiler=msvc-15,os_label=win10_64/886/ 

> Test Result (no failures)

shouldnt the logstream test have failed to build and produce an error? instead it runs through: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_windows_ci/compiler=msvc-15,os_label=win10_64/886/testReport/projectroot.src.tests.class_tests/openms/LogStream_test/ I dont quite understand this?

timosachsenberg:
@hroest 
LogStream (LogStreamBuf *buf=0, bool delete_buf=true, std::ostream *stream=0)
seems multi-default constructors have the same problem as single argument constructors: implicit conversion 
https://stackoverflow.com/questions/187640/default-parameters-with-c-constructors

hroest:
@timosachsenberg my point was that in 52b41a6 there is clearly a syntax error as `l2` is not defined but the Windows test run through and reports no error?

timosachsenberg:
rebuild win

hroest:
looks good to me. again, I was more worried about the win test *not* catching the error but thats independent of this PR

hroest:
ok, happy to merge, just two concerns: why is the contrib updated and since this touches many files we may want to coordinate with other PR so that we dont have too many merge conflicts

timosachsenberg:
the submodule was pulled I guess so this only contains the recent changes already in contrib

PR #3070, opened 2017-12-06
Update LICENSE/AUTHORS files to represent institutions better

hendrikweisser:
This change removes the mentions of institutions (Uni Tübingen, FU Berlin, ETH Zürich) from the licence - leaving just "The OpenMS Team" - and adds institutions to the list of authors in AUTHORS.
The goal is to give a fairer representation of the institutions that have contributed to OpenMS.

(In earlier discussions I suggested adding an INSTITUTIONS file, but I think listing author affiliations is more elegant and shouldn't require approval from the institutions.)

I didn't know everybody's affiliations and may have got some incorrect, so please check and amend.

If there's agreement that this change should be made, I'll replace the licence headers in the source code files (perhaps together with changing 2017 to 2018).


timosachsenberg:
@dmccloskey et al. fell free to add you to this (or the old authors file) once it is included

hroest:
it may actually be an issue that we first have to discuss with a lawyer - I would definitely not change right now without consulting a lawyer. The question is whether the wording is correct and where the copyright lies.

Also it may make sense to start an "OpenMS e.V." at some point instead of having "The OpenMS team" in there which is probably legally a meaningless term.

timosachsenberg:
I read a bit into it and think Hannes is right that we should name a single legal entity that gets the copyright assigned to. This doesn't seem come with any disadvantage as everyone already put the code under the license terms of the project. The main advantage for everyone would be that individuals are not exposed to liability in the unlikely case of a copyright infringement suit.

hroest:
> The main advantage for everyone would be that individuals are not exposed to liability in the unlikely case of a copyright infringement suit.

:+1: 

I also like the idea of Hendrik that we have some sort of record which institutions were involved in development, however more as a list for us than for any legal reasons

hendrikweisser:
I have to make sure that my new employer gets adequate recognition for work that they pay for and then contribute to an open-source project.
There's also no real reason why we added ETH Zurich to the copyright notice, but not any of the institutions that contributed later (e.g. Wellcome Trust Sanger Institute or Uni Freiburg - @lars20070) - especially now that there's no OpenMS development at ETH Zurich any more (as far as I know).

So how do we proceed? I think there are possibly two separate questions:
1. Are we allowed to remove the three universities from the copyright notice? (If yes, I think we should go through with the changes proposed here.)
2. For the future, should the copyright be assigned to a "single legal entity", e.g. an "OpenMS e.V."? (@timosachsenberg, can you talk to Oliver about this? In any case, this doesn't seem possible in the immediate short-term.)

timosachsenberg:
Let's try to call @okohlbacher :) ...

hroest:
> I have to make sure that my new employer gets adequate recognition for work that they pay for and then contribute to an open-source project.

I see that point. Other projects, such as the Linux kernel, do that by looking at the email addresses of commits

okohlbacher:
Well, as we discussed, I am totally fine with having a CONTRIBUTORS file
and changing the headers in each file to reference that. WOuld that work
for everyone?

> On 7. Dec 2017, at 19:18, Hannes Roest <notifications@github.com> wrote:
> 
> I have to make sure that my new employer gets adequate recognition for work that they pay for and then contribute to an open-source project.
> 
> I see that point. Other projects, such as the Linux kernel, do that by looking at the email addresses of commits
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

---
Oliver Kohlbacher
Professor, Center for Bioinformatics, Dept. of Computer Science, & Faculty of Medicine
Director, Quantitative Biology Center, University of Tuebingen, Germany
Fellow, Max Planck Institute for Developmental Biology

email  oliver.kohlbacher@uni-tuebingen.de  /  oliver.kohlbacher@tuebingen.mpg.de
WWW    kohlbacherLab.org   twitter  @OKohlbacher    
phone  +49 7071 29 70457   fax      +49 7071 29 5152

hendrikweisser:
@okohlbacher: Initially I suggested having an INSTITUTIONS file in addition to the current AUTHORS file, but now I think it may be easier to just list author affiliations in the AUTHORS file, and keep only "The OpenMS Team" in the licence/header. Would you be happy with that?
I actually like the notion of renaming it CONTRIBUTORS, to refer to both the individuals and their institutions.

hendrikweisser:
@timosachsenberg: Could you prod @okohlbacher regarding my last question, please?

cbielow:
as a sidenote:
just in case someone is wondering if removing the license text and replacing it with a simple `// please see LICENSE and AUTHORS/CONTRIBUTORS for more information ...` in all our .h and .cpp files makes a difference in compile time: not even a split second (gcc 7.2; make -j20 openms) :-)
Nevertheless: I'd opt to do it anyway, so we avoid
 - having to update the copyright year in every file
 - actually show some code when opening a file in an editor

timosachsenberg:
feel free to reopen

PR #3072, opened 2017-12-06
[FEATURE] Add ChromatogramPeak constructor with position and intensity args

pasqoo:
Allows the user to instantiate a `ChromatogramPeak` by passing a position and an intensity as arguments, as in:
`ChromatogramPeak p(3.0, 5.0);`

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/ChromatogramPeak.h
+++ b/src/openms/include/OpenMS/KERNEL/ChromatogramPeak.h
@@ -85,6 +85,12 @@ public:
       intensity_(p.intensity_)
     {}
 
+    /// Constructor with position and intensity
+    inline ChromatogramPeak(const PositionType position, const IntensityType intensity) :
+      position_(position),
+      intensity_(intensity)
+    {}
+
     /**
       @brief Destructor
 
--- a/src/openms/include/OpenMS/KERNEL/ChromatogramPeak.h
+++ b/src/openms/include/OpenMS/KERNEL/ChromatogramPeak.h
@@ -86,8 +86,8 @@ public:
     {}
 
     /// Constructor with position and intensity
-    inline ChromatogramPeak(const PositionType position, const IntensityType intensity) :
-      position_(position),
+    inline ChromatogramPeak(const PositionType retention_time, const IntensityType intensity) :
+      position_(retention_time),
       intensity_(intensity)
     {}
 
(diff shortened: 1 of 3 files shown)


timosachsenberg:
nice. very useful

src/openms/include/OpenMS/KERNEL/ChromatogramPeak.h
```diff
@@ -85,6 +85,12 @@ namespace OpenMS
       intensity_(p.intensity_)
     {}
 
+    /// Constructor with position and intensity
+    inline ChromatogramPeak(const PositionType position, const IntensityType intensity) :
```
hroest: just for clarity, I would rename the argument from "position" to "retention_time"

pasqoo: Sure!
About the pxd file, is it necessary to give argument names? Because for the rest of the file they are not present @hroest

pasqoo:
@hroest I am confused about when/why I should use `PositionType` and/or/instead of `CoordinateType`...
I see that they are both typedefs from `double`. But what is the deciding factor behind it?
Thanks

src/openms/include/OpenMS/KERNEL/ChromatogramPeak.h
hroest: no its not necessary but it helps people to understand the code :-)

hroest:
@pasqoo PositionType is the type here, but in this context it refers to retention time as we are talking about chromatograms. It just helps with the interpretation if we are clear about these points (I think). *Especially* if both types are just doubles, it will be easy to make a mistake and provide the wrong variable (e.g. switch the two) and therefore its important to be clear which is which.

pasqoo:
@hroest I see, thanks for the details.

So the inconsistency introduced in the .pxd file by replacing `PositionType` with `CoordinateType` is fine? What makes me think is the fact that now, after the last commits, the .pxd and .h files are using two different types for the same argument.

hroest:
well, I agree its a small issue but currently there is no PositionType in pxd - feel free to define it.

pasqoo:
@timosachsenberg @hroest The failed CI test says:

Is there anything I can do about it?

timosachsenberg:
No its related to the maximum build time we get on travis

timosachsenberg:
Thanks for your contribution!

PR #3083, opened 2017-12-08
[DOC] Updated docu to reflect included submodules + general updates

jpfeuffer:


timosachsenberg:
looks good. thanks for also putting all the other stuff (docker etc.) in there.

PR #1342, opened 2015-04-10
[FEATURE] ModificationsDB: Use custom UniMod & PSI-MOD XML

grosenberger:
For some applications, it would be very convenient to use ModificationsDB instances based on custom UniMod XML & PSI-MOD OBO files. This pull requests suggests the function ModificationsDB::getCustomInstance() and changes in File::find() to support loading of these files using absolute and relative pathes.

PR #3086, opened 2017-12-08
[FEATURE] ModificationsDB: Load databases from defined path 

grosenberger:
The TOPP tool OpenSwathAssayGenerator requires a definition of residue modifiability when used as part of the OpenSWATH/IPF workflow. Similar to database search or PTM site-localization algorithms, this could be provided in the form of Phospho (STY) as part of the input parameters. However, we make use of ModificationsDB, which provides great abstraction for different reference databases.

Unimod and PSI-MOD compiled in ModificationsDB however are very inclusive databases. For Phosphorylation, S, T, Y, D, H, C, R & K can be modified in the current Unimod version. Most users are however interested only in the canonical residues S,T & Y. Currently, we thus modify ``unimod.xml`` in ``share/OpenMS/CHEMISTRY`` or we set ``OPENMS_DATA_PATH`` for the OpenSwathAssayGenerator TOPP tool. This works well for developers, however, it is too complicated and error prone for users and is more difficult to implement with workflow managers like KNIME. We thus need to provide a simpler interface that processes a separate ``unimod.xml`` file as part of the TOPP tool and validate the input.

We have previously discussed how to extend ``ModificationsDB`` (https://github.com/OpenMS/OpenMS/pull/1342) and I have now implemented the proposed changes from back then.


hroest:
looks good, from the last discussion I think we also were talking about `modDB->readFromUnimodXMLFile(...)` and the other read functions. The idea was that they should probably be private and only called from ModificationsDB itself and therefore nobody should be able to call these functions just by themselves. Maybe that should be included here too?

hroest:
@cbielow what is your view on making

```
void readFromOBOFile(const String& filename);
void readFromUnimodXMLFile(const String& filename);
```

private? I am in favor

hroest:
looks like your tests are now different

```
Offending lines:			(tab_width = 8, first_column = 1)

in1:  /home/travis/build/OpenMS/OpenMS/_build/src/tests/topp/OpenSwathAssayGenerator_output_2.TraML.tmp   (line: 84293, position/column: 93/94)
        <cvParam cvRef="MS" accession="MS:1000827" name="isolation window target m/z" value="!
                                                                                             903.3877" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>

in2:  /home/travis/build/OpenMS/OpenMS/src/tests/topp/OpenSwathAssayGenerator_output_2.TraML   (line: 84293, position/column: 93/94)
        <cvParam cvRef="MS" accession="MS:1000827" name="isolation window target m/z" value="!
                                                                                             989.3993" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>



```

grosenberger:
It is the usual Boost version problem with the random number generator. I'm preparing a patch that skips the random component.

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -65,22 +65,29 @@ namespace OpenMS
...
+      databases. This can be done by providing a path when initializing
+      ModificationsDB.
   */
   class OPENMS_DLLAPI ModificationsDB
   {
 public:
 
     /// Returns a pointer to the modifications DB (singleton)
-    inline static ModificationsDB* getInstance()
+    inline static ModificationsDB* getInstance(OpenMS::String unimod_file = "CHEMISTRY/unimod.xml", OpenMS::String psimod_file = "CHEMISTRY/PSI-MOD.obo", OpenMS::String xlmod_file = "CHEMISTRY/XLMOD.obo")
```
grosenberger: This seems to be problematic for pyOpenMS. @hroest, do you know what I need to change in the corresponding pxd to make this work?

grosenberger:
These patches seem to have resolved the pyOpenMS issues, thanks!

hroest:
any other comments? if not, I think we are ready to merge

jpfeuffer:
I have a question: How would this work with the restrictions that we impose on modification parameters in e.g. search engine adapters?

grosenberger:
I believe that most search engines except Mascot currently require specific parameters for residue modifiability. To enable more generic settings for different engines, it might thus be possible in the future to translate these using a restricted ModificationsDB as implemented for OpenSwathAssayGenerator.

jpfeuffer:
I think I meant something different:
For our search engine adapters, we currently read the contents of the Modification databases and expect/restrict users to select one of these modifications for the modification parameter on the command line, which we then translate and pass to the underlying tool according to their schemas. This is also helpful because we can display the options that the user has.
But it is bad if the databases can change arbitrarily. Still I think that this is a step in the right direction.

grosenberger:
Yes, that is a good point and most tools built on top of ModificationsDB make this assumption of completeness. In this regard, the default will not change. We only add an option to make this more specific for certain scenarios, as required for example for OpenSwathAssayGenerator. Since the scope is limited to single instances, I hope the isolation from other OpenMS components is strict enough.

PR #3088, opened 2017-12-08
[FEATURE] TargetedFileConverter: Improved 3rd party support

grosenberger:
This PR improves 3rd party support in TargetedFileConverter. The goal is to seamlessly import Skyline and Spectronaut transition lists and to convert them to TraML. the following changes have been implemented for this purpose:

- Header synonyms: If detected in the TSV/CSV, Skyline or Spectronaut transition list headers are translated to according OpenSWATH TSV columns.
- Group IDs: If not available, transition_group_id and transition_name are generated
- SpectraST: Unannotated fragment ions in SpectraST MRM files are omitted.


hroest:
+1 sounds good. Can you make sure that the file format stays consistent with our description here: http://openswath.org/en/latest/docs/openswath_workflow.html#input ? 
- do we have a test of Spectronaut and of the Skyline format? I currently only see a single additional test. I think that would be useful!

grosenberger:
Absolutely, I'm working on it and I will add tests for both. For Skyline, we can also provide a Skyline Report .skyr file, which will directly export the compatible format.

I will also adapt the description on the website.

hroest:
I think I was talking to Hendrik about this and the idea was to change the headers as follows

```
-    "Tr_recalibrated",
-    "transition_name",
+    "RetentionTime",
+    "TransitionName",

-    "transition_group_id",
-    "decoy",
+    "TransitionGroupId",
+    "Decoy",

-    "detecting_transition",
-    "identifying_transition",
-    "quantifying_transition"
+    "DetectingTransition",
+    "IdentifyingTransition",
+    "QuantifyingTransition"
```

that would mean consistent CamelCase and easy readability. I think this PR is a good opportunity to make that change (while keeping backward compatible). I dont know whether any tool consumes the TSVs that our `TargetedFileConverter` produces, so it should not be a problem to change it. Also its a TSV format and we never promised stability in that format. 

see https://github.com/OpenMS/OpenMS/pull/2874#issuecomment-347660864

PR #3090, opened 2017-12-11
[FEATURE] PeakIntegrator

pasqoo:
Compute the area, background and shape metrics of a peak.

The containers supported by the methods are MSChromatogram and MSSpectrum.

Integration techniques implemented are: intesity sum, trapezoidal rule, simpson's rule.


src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
```diff
@@ -0,0 +1,355 @@
...
+  double PeakIntegrator::getEndTimeAt50() const { return end_time_at_50_; }
+  double PeakIntegrator::getTotalWidth() const { return total_width_; }
+  double PeakIntegrator::getTailingFactor() const { return tailing_factor_; }
+  double PeakIntegrator::getAsymmetryFactor() const { return asymmetry_factor_; }
+  double PeakIntegrator::getBaselineDeltaToHeight() const { return baseline_delta_2_height_; }
+  double PeakIntegrator::getSlopeOfBaseline() const { return slope_of_baseline_; }
+  Int PeakIntegrator::getPointsAcrossBaseline() const { return points_across_baseline_; }
+  Int PeakIntegrator::getPointsAcrossHalfHeight() const { return points_across_half_height_; }
+
+  const std::map<String, double> PeakIntegrator::getPeakShapeMetrics() const
```
hroest: I am not a big fan of this approach, why not explicitely encode this in a struct and remove all these methods, basically have

```
struct PeakShapeMetrics
{
double peak_area;
...
}

PeakShapeMetrics PeakIntegrator::getPeakShapeMetrics() const 
```

to me it seems that would get rid of a lot of unnecessary code. I only really see advantages to this approach, honestly.

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
```diff
@@ -0,0 +1,355 @@
...
+          if (area != -1.0)
+          {
+            peak_area_ += area;
+            ++valids;
+          }
+        }
+        peak_area_ /= valids;
+      }
+    }
+    else
```
hroest: I suggest
```
else if (integration_type_ == "intensity_sum")
{
}
else
{
  throw ...
}
```
to ensure that we have 3 distinct ways of integrating and throw an exception if any other way is used. I think otherwise it will be hard to add additional integration techniques as it will not be clear whether they are explicitly supported if we always have a "fallback option" that is used if no other method fits.

hroest:
great that looks really good, I have some comments on how to return the PeakShapeMetrics, I think this could be done more elegantly. Also, did you test this against the current integration / background subtraction methods? 

I would be nice to have the "old" approach (sum integration) return the same result so that we can ditch the old code and switch to the PeakIntegrator completely. Well we can also do that in a separate PR, I like also the approach of having a clean PR and add one thing at a time ...

pasqoo:
@hroest Thanks for your review. I updated the files trying to follow what you recommended.
The output from `calculatePeakShapeMetrics()` now goes inside a `struct`.

One of our goals is to replace that old code you're referring to (a part of `MRMTransitionGroupPicker`), updating the OpenSwath workflow once this PR gets merged. I checked that the results from this class' tests looked legit.

@dmccloskey The getters for the metrics are gone, replaced by the single "return a struct" getter.
The getters for the outputs from `integratePeak()` and `estimateBackground()` were not changed.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
```diff
@@ -0,0 +1,386 @@
...
+      */
+      double width_at_10 = 0.0;
+      /**
+        The width of the peak at 50% the peak's height.
+      */
+      double width_at_50 = 0.0;
+      /**
+        The start position at which the intensity is 5% the peak's height.
+      */
+      double start_time_at_5 = 0.0;
```
timosachsenberg: time might be a misnomer if metrics are calculated for mass peaks (works only for chromatographic peaks)

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
```diff
@@ -0,0 +1,386 @@
...
+      The user should make sure to use the same integration_type between calls of estimateBackground() and
+      integratePeak().
+
+      @note Make sure the chromatogram is sorted with respect to retention time.
+
+      @param[in] chromatogram The chromatogram which contains the peak
+      @param[in] left The left retention time boundary
+      @param[in] right The right retention time boundary
+    */
+    void estimateBackground(const MSChromatogram& chromatogram, const double& left, const double& right);
```
timosachsenberg: would it make sense to also provide these functions for a start and end iterator?

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
```diff
@@ -0,0 +1,320 @@
...
+  {
+    const double int_l = p.PosBegin(left)->getIntensity();
+    const double int_r = (p.PosEnd(right)-1)->getIntensity();
+    const double delta_int = int_r - int_l;
+    const double delta_pos = (p.PosEnd(right)-1)->getPos() - p.PosBegin(left)->getPos();
+    const double min_int_pos = int_r <= int_l ? (p.PosEnd(right)-1)->getPos() : p.PosBegin(left)->getPos();
+    const double delta_int_apex = std::fabs(delta_int) * std::fabs(min_int_pos - peak_apex_pos_) / delta_pos;
+    background_height_ = std::min(int_r, int_l) + delta_int_apex;
+    double background = 0.0;
+    if (baseline_type_ == "base_to_base")
```
timosachsenberg: minor comment: if you want to increase maintainability (and possible hard to find bugs) you could replace the string literals distributed all over the place by constants.
For example, if you define a: ```constexpr const char* INTEGRATION_TYPE_SIMPSON = "simpson"; ```
and use it throughout the code (and parameter), you don't run into the risk of hard to detect typos (aka using "simspon" at one position in the code).

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
```diff
@@ -0,0 +1,320 @@
...
+        background = delta_pos * (std::min(int_r, int_l) + 0.5 * std::fabs(delta_int));
+      }
+      else if (integration_type_ == "intensity_sum")
+      {
+        // calculate the background using the formula
+        // y = mx + b where x = rt or mz, m = slope, b = left intensity
+        // sign of delta_int will determine line direction
+        // background += delta_int / delta_pos * (it->getPos() - left) + int_l;
+        UInt n_points = 0;
+        for (auto it=p.PosBegin(left); it!=p.PosEnd(right); ++it, ++n_points)
```
timosachsenberg: style: add spaces (e.g., around =)

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
```diff
@@ -0,0 +1,320 @@
...
+    else if (baseline_type_ == "vertical_division")
+    {
+      if (integration_type_ == "trapezoid" || integration_type_ == "simpson")
+      {
+        background = delta_pos * std::min(int_r, int_l);
+      }
+      else if (integration_type_ == "intensity_sum")
+      {
+        UInt n_points = 0;
+        for (auto it=p.PosBegin(left); it!=p.PosEnd(right); ++it, ++n_points)
```
timosachsenberg: this can be simpler written with std::distance
also: are you sure you are counting the number of points correctly?

timosachsenberg: if not, please also check the other occurences of this type of coe

timosachsenberg:
In general, it looks very good.
From a design point of view there might be some improvements. You could reduce the amount of "state" the class has to the parameters only. 
So basically, if you remove all internal members (except those set by the parameters) and directly return a result (e.g. a struct) you get a better class.
You may ask why? The reason is that you now can use the same PeakIntegrator object in a multi-threaded context (e.g. one thread doesn't overwrite the result of the other thread - which otherwise can easily happen if you store results in members).

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
hroest: I think its fine, I would suggest however to throw an error if not found, e.g. 

```
if (baseline_type_ == "base_to_base")
{
}
else if (baseline_type_ == "vertical_division") 
{
}
else
  throw Exception
```

hroest:
> From a design point of view there might be some improvements. You could reduce the amount of "state" the class has to the parameters only.

I agree, we already agreed to use a struct to return the peak shape metrics which is a step into that direction. It also reduces a lot of boilerplate `getX` code.

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
pasqoo: I will use that `std::distance`, seems the method I should be using!
Yes I think I've been counting them correctly. I guess `std::distance` will help me finding out the truth :smile:

pasqoo:
@timosachsenberg @hroest At this point I think I've addressed all comments, except the "reduce state".
I will work on that.
I could use some help on those pyopenms tests failing. I think it complains about how I wrap the struct.

pasqoo:
@hroest @timosachsenberg The class should now reflect the changes you proposed. I removed the output members from it and so the state is therefore reduced. The documentation has to be updated accordingly.

While working on this, I noticed that part of the tests were run with unrealistic input values.
For example, the `estimateBackground()` tests were run before `integratePeak()`s and so the value of `peak_apex_pos` would be the default one instead of a realistic value.
With the current version (last commit) the class outputs the same results (all class' tests pass) because I'm providing the same inputs as before the refactor.

I'll work on updating the inputs and the tests values, and on updating the documentation/comments.

pasqoo:
@hroest @timosachsenberg The PR should be up to date now, ready to be reviewed again.

As I mentioned on Gitter, I noticed that test subsections don't get an isolated scope.
To reproduce the behaviour, uncomment line 559 and comment line 560 in the tests file.
The variables will have a value, even though they were just declared. What am I missing?

It also looks like the pyOpenMS tests are still failing.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
```diff
@@ -77,7 +168,10 @@ namespace OpenMS
       @param[in] left The left retention time boundary
       @param[in] right The right retention time boundary
     */
-    void estimateBackground(const MSChromatogram& chromatogram, const double& left, const double& right);
+    void estimateBackground(
+      const MSChromatogram& chromatogram, const double& left, const double& right, const double& peak_apex_pos,
+      double& background_area, double& background_height
```
hroest: note how this will not work with pyOpenMS as Python does not support passing primitive data types (double, int) by reference. You will not be able to get these data back.

hroest:
I fixed the pyOpenMS compilation errors but you still need to adjust the actual calls into the API - I also suggest to not pass primitive types by reference as these cannot be mapped to Python and will therefore be passed by value

hroest:
> As I mentioned on Gitter, I noticed that test subsections don't get an isolated scope.
To reproduce the behaviour, uncomment line 559 and comment line 560 in the tests file.
The variables will have a value, even though they were just declared. What am I missing?

the variables are properly scoped, you can easily test this by removing line 556 which produces a compile error. Also you can rename the variable to be sure. Using uninitialized variables produces undefined behavior. as far as I can see (Again its easy to confirm) due to the structure of the code the values from the previous test are used (e.g. try setting the variables to a different value after line 546 and you will see that value in the output). So it seems an easy enough case to solve, unitialized variable, the compiler does what it wants and re-uses the memory without zeroing it. That makes all sense and is expected.

But what is unexpected and I dont understand at all why there are no warning produced for this case. We have `-Wall -Wextra` turned on, so this should produce a warning. But it does not!. Our warnings arent catching that and this is now an even stranger case: 

if I comment out lines 560 and 561 which uses the `peak_height` variable, then I get 

```
[100%] Building CXX object src/tests/class_tests/openms/CMakeFiles/PeakIntegrator_test.dir/source/PeakIntegrator_test.cpp.o
/home/hr/openmsall/source/OpenMS_experimental/src/tests/class_tests/openms/source/PeakIntegrator_test.cpp: In function ‘int main(int, char**)’:
/home/hr/openmsall/source/OpenMS_experimental/src/tests/class_tests/openms/source/PeakIntegrator_test.cpp:559:45: warning: ‘peak_height’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   std::cout << "IN TESTS: peak_height: " << peak_height << " \t peak_apex_pos: " << peak_apex_pos << std::endl;
                                             ^
/home/hr/openmsall/source/OpenMS_experimental/src/tests/class_tests/openms/source/PeakIntegrator_test.cpp:559:85: warning: ‘peak_apex_pos’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   std::cout << "IN TESTS: peak_height: " << peak_height << " \t peak_apex_pos: " << peak_apex_pos << std::endl;
                                                                                     ^
[100%] Linking CXX executable ../bin/PeakIntegrator_test
[100%] Built target PeakIntegrator_test
```

but if have them in, then I get 

```
make PeakIntegrator_test
[  0%] Built target OpenSwathAlgo
[100%] Built target OpenMS
Scanning dependencies of target PeakIntegrator_test
[100%] Building CXX object src/tests/class_tests/openms/CMakeFiles/PeakIntegrator_test.dir/source/PeakIntegrator_test.cpp.o
[100%] Linking CXX executable ../bin/PeakIntegrator_test
[100%] Built target PeakIntegrator_test
```

so for some reason the compiler does not emit a warning when using the variable uninitialized *iff* we call `integratePeak` afterwards using that variable.

pasqoo:
Thanks for your explanation and fixes. So I was victim of undefined behavior :smiley: 
I will update all tests' variables, giving them an initial value and I'll try to fix the python wrapping. Somehow the wrappers should work.

hroest:
hmm this really seems to be a strange case that happens with gcc, I have been able to reproduce this with a minimal example:

```
#include <iostream>

class klass 
{
public: void test2(double k) {std::cout << k << std::endl; k += 42; }
public: void test3(double& k) {k = 42;}
};

int main(int argc, const char** argv)
{
    double q;
    klass* ptr;
    ptr = new klass();
    ptr->test2(q);
    ptr->test3(q);
    ptr->test2(q);
}
```

This compiles even with `-O3 -Wuninitialized -Werror` on the latest gcc, however it does not on clang, you can easily try this on https://godbolt.org/

hroest:
>  I will update all tests' variables, giving them an initial value and I'll try to fix the python wrapping. Somehow the wrappers should work.

sounds good. Also look into how to pass variables in pyOpenMS as Python does not support passing primitive types (int, double) by reference.

pasqoo:
I looked for more info as well on that missed warning, and I got the feeling it's not something trivial to solve.
Something related could be found here: https://gcc.gnu.org/wiki/Better_Uninitialized_Warnings

The PR is now up to date again, ready for review. I moved some parts of code and later realized it would have been better to do that on a separate "style" commit instead. Hope it's still easy to see the differences.

We realized that `integratePeak()` is currently not returning the hull points. I'll implement that soon.

pasqoo:
@hroest @timosachsenberg The PR is up to date. Unfortunately, one of the tests exceeded the time limit allotted.

hroest:
@pasqoo I created a minimal test case and reported a bug against gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83468 it seems like that it has to do with how they perform optimization and not something easily resolved.

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
```diff
@@ -0,0 +1,359 @@
...
+    }
+    else if (baseline_type_ == BASELINE_TYPE_VERTICALDIVISION)
+    {
+      if (integration_type_ == INTEGRATION_TYPE_TRAPEZOID || integration_type_ == INTEGRATION_TYPE_SIMPSON)
+      {
+        background = delta_pos * std::min(int_r, int_l);
+      }
+      else if (integration_type_ == INTEGRATION_TYPE_INTENSITYSUM)
+      {
+        for (auto it = p.PosBegin(left); it != p.PosEnd(right); ++it)
```
timosachsenberg: what should this line of code do?

pasqoo: That `for` is not useful anymore. It used to count the number of elementes between the iterators (so there was not only `++it`, but also `++count`, which then got removed), before you recommended using `std::distance()`. I'll update the code.

PR #3093, opened 2017-12-12
Fix/style changes

hroest:
Some style changes after #3088

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVReader.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVReader.cpp
@@ -44,71 +44,70 @@
 namespace OpenMS
 {
 
-template<class T>   // primary template
-bool extractName(T& value, const std::string& header_name,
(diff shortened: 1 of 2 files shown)


hroest:
@grosenberger let me know what you think

src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVReader.cpp
```diff
@@ -253,380 +318,107 @@ namespace OpenMS
       mytransition.precursor = String(tmp_line[header_dict["PrecursorMz"]]).toDouble();
 
       // ProductMz
-      if (header_dict.find("ProductMz") != header_dict.end())
-      {
-        mytransition.product = String(tmp_line[header_dict["ProductMz"]]).toDouble();
-      }
-      else if (header_dict.find("FragmentMz") != header_dict.end()) // Spectronaut
-      {
-        mytransition.library_intensity = String(tmp_line[header_dict["FragmentMz"]]).toDouble();
```
grosenberger: Thanks for spotting this bug!

PR #3095, opened 2017-12-12
Feature/file info level based peak type

timosachsenberg:
The peak type is now printed for every individual MS level (and not only of the first spectrum)


hroest:
rebuild win

PR #3096, opened 2017-12-12
[FEATURE] smoothed chromatogram reporting in OpenSWATH

dmccloskey:
Allows MRMTransitionGroupPicker the option to calculate the peak area and height based on the raw or smoothed data.  This allows for the hull points to be reported for retrospective plotting for the smoothed chromatogram.  This also allows enforces the background to be calculated on the same chromatogram as the peak area/height.


src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -57,8 +57,11 @@ namespace OpenMS
     defaults_.setValue("stop_after_intensity_ratio", 0.0001, "Stop after reaching intensity ratio");
     defaults_.setValue("min_peak_width", -1.0, "Minimal peak width (s), discard all peaks below this value (-1 means no action).", ListUtils::create<String>("advanced"));
 
+    defaults_.setValue("peak_integration", "original", "Calculate the peak area and height either the smoothed or the raw chromatogram data.", ListUtils::create<String>("advanced"));
+    defaults_.setValidStrings("peak_integration", ListUtils::create<String>("original,smoothed"));
+
     defaults_.setValue("background_subtraction", "none", "Try to apply a background subtraction to the peak (experimental). The background is estimated at the peak boundaries, either the smoothed or the raw chromatogram data can be used for that.", ListUtils::create<String>("advanced"));
-    defaults_.setValidStrings("background_subtraction", ListUtils::create<String>("none,smoothed_average,smoothed_exact,original_average,original_exact"));
+    defaults_.setValidStrings("background_subtraction", ListUtils::create<String>("none,average,exact"));
```
hroest: can you add "original" and "smoothed" as a alias to be compatible with previous versions of OpenSWATH, see 34a55e4392f406727a3778c7d3ec443012d2c74e

dmccloskey: So adding in the option for the user to calculate the background based on the original data if they specify to calculate the area using the smoothed data and vice versa?  I am a bit hesitant on that because the peak area and peak height calculations should be consistent with how the background area and height are calculated.  

Let me know your thoughts on this, but at least from my perspective I think having that option might create more problems than solving.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -277,9 +277,27 @@ namespace OpenMS
...
+          std::cerr << "Tried to calculate peak area and height without any smoothed chromatograms" << std::endl;
+          used_chromatogram = resampleChromatogram_(chromatogram, master_peak_container, best_left, best_right);
+        }        
+        else if (peak_integration_ == "smoothed")
+        {
+          used_chromatogram = resampleChromatogram_(smoothed_chroms[k], master_peak_container, best_left, best_right);
+        }
+        else
+        {
+          std::cerr << "No peak integration specified.  Defaulting to original." << std::endl;
```
hroest: this should be a hard error, e.g. `throw`

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -277,9 +277,27 @@ namespace OpenMS
...
-        const SpectrumT used_chromatogram = resampleChromatogram_(chromatogram, master_peak_container, best_left, best_right);
-        // const SpectrumT& used_chromatogram = chromatogram; // instead of resampling
+        if (peak_integration_ == "original")
+        {
+          used_chromatogram = resampleChromatogram_(chromatogram, master_peak_container, best_left, best_right);
+          // const SpectrumT& used_chromatogram = chromatogram; // instead of resampling
+        }
+        else if (peak_integration_ == "smoothed" && smoothed_chroms.size() <= k)
+        {
+          std::cerr << "Tried to calculate peak area and height without any smoothed chromatograms" << std::endl;
```
hroest: this should be a hard error, e.g. `throw`

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
hroest: well, I would prefer if the options stayed somewhat backwards compatible and in #2944 we had the following change

```
-    defaults_.setValidStrings("background_subtraction", ListUtils::create<String>("none,smoothed,original"));
+    defaults_.setValidStrings("background_subtraction", ListUtils::create<String>("none,smoothed_average,smoothed_exact,original_average,original_exact"));
```

which caused some issues

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -294,24 +312,17 @@ namespace OpenMS
         double background(0), avg_noise_level(0);
         if (background_subtraction_ != "none")
         {
-          if ((background_subtraction_ == "smoothed_average" || background_subtraction_ == "smoothed_exact") && smoothed_chroms.size() <= k)
+          if ((peak_integration_ == "smoothed") && smoothed_chroms.size() <= k)
           {
             std::cerr << "Tried to calculate background estimation without any smoothed chromatograms" << std::endl;
             background =  0;
+            avg_noise_level = 0;
```
hroest: this should be a hard error, e.g. `throw`

src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
hroest: it is my understanding that `background_subtraction == original` is now the same as
```
background_subtraction == average
peak_integration == original
```

right? so I am saying that some people use OpenMS 2.1 with the option  `background_subtraction == original` and it is non-trivial to understand which options need to be set to achieve the same results now with your patch.

hroest: PS: please also change the description of the `"background_subtraction"` parameter

dmccloskey: The first version of `background_subtract == original` was before we had added the two methods to estimate the background.  What would you recommend changing the name of the background subtract parameters to be so that it is a bit easier to understand?

Also, I should note that once #3090 is implemented, and we get a chance to add this interface to `MRMTransitionGroupPicker`, it will change the options for `background_subtract` again.

hroest: Yes, it would be helpful if the parameters have somewhat the same meaning across OpenMS versions. Basically but just making it an alias we would help that along, using the "principle of least surprise". But I agree that with #3090 we will change the interface again ...

PR #3097, opened 2017-12-14
[FEATURE] OpenSwathAssayGenerator improvements

grosenberger:
Up until now, ``OpenSwathAssayGenerator`` was mainly used for sample-specific spectral libraries in combination with IPF. However, it replicates also most functionality of ``spectrast2tsv.py`` from msproteomicstools and could thus replace it for a complete SWATH-MS library generation workflow in OpenMS (via ``SpectrastAdapter``. For this reason, speed needed to be optimized and I/O support for transition lists and OpenSWATH PQP files was added.

When selecting the ~1.2M optimal transitions from the ~7M total transitions / ~215K peptides of the PHL, ``OpenSwathAssayGenerator`` was too slow in comparison to ``spectrast2tsv.py``.
- spectrast2tsv.py (sptxt -> TSV): ~14min
- OpenSwathAssayGenerator (TraML -> TraML): >5h

Originally, the tool operated on the transition-level, parsing the meta information from the TraML with optional validation against the theoretical reference. By moving this to the peptide-level and by abandoning the generation of new transition group and transition indices, the speed could be improved:
- OpenSwathAssayGenerator (TraML -> TraML): ~15min
- OpenSwathAssayGenerator (MRM -> TraML): ~12min
- OpenSwathAssayGenerator (MRM -> TSV): ~10min
- OpenSwathAssayGenerator (MRM -> PQP): ~17min

The output remains the same, however the running time is competitive with ``spectrast2tsv.py`` and also includes the writing of the TraML file. Export to PQP remains expensive because different indices need to be generated and writing of SQLite is also more expensive.

Further, the IPF parameters of ``OpenSwathAssayGenerator`` have been modified to reflect ideal default parameters that are easier to understand.

I'm working on a parallel PR for ``OpenSwathDecoyGenerator`` that will also introduce the other formats and will provide simplified parameters.


src/tests/class_tests/openms/data/MRMAssay_reannotateTransitions_output_1.TraML
```diff
@@ -25,7 +25,7 @@
     </Protein>
   </ProteinList>
   <CompoundList>
-    <Peptide id="1_sp|O75882|ATRN_HUMAN_AAAAAAVSGSAAAEAK_2_50.01" sequence="AAAAAAVSGSAAAEAK">
```
grosenberger: Peptides and transitions are ordered differently. The peptide and transition indices are taken from the input and not generated from scratch.

hroest:
great, looks really cool. Looks like it is now faster than `spectrast2tsv.py`, that is really neat! Looking forward to the same PR for the decoy assay generator.

src/topp/OpenSwathAssayGenerator.cpp
```diff
@@ -217,14 +284,33 @@ class TOPPOpenSwathAssayGenerator :
...
+      TransitionTSVReader tsv_reader = TransitionTSVReader();
+      tsv_reader.setLogType(log_type_);
+      tsv_reader.convertTargetedExperimentToTSV(tr_file, targeted_exp);
+    }
+    if (out_type == FileTypes::PQP)
+    {
+      const char * tr_file = out.c_str();
+      TransitionPQPReader pqp_reader = TransitionPQPReader();
+      pqp_reader.setLogType(log_type_);
+      pqp_reader.convertTargetedExperimentToPQP(tr_file, targeted_exp);
```
timosachsenberg: this reads a bit strange. we have a reader that is also a PQP writer

grosenberger: Good point. Originally, we provided read-only support via TransitionTSVReader. We could potentially promote both TransitionTSVReader and TransitionPQPReader to a real format in OpenMS. Since other open PR would be affected as well, I suggest we do this in a separate PR.

timosachsenberg: fine with me

timosachsenberg:
Needs conflict resolution

grosenberger:
Ok, this should work again with the other changes. Please merge if you think that it is ready.

src/topp/OpenSwathAssayGenerator.cpp
```diff
@@ -188,7 +230,7 @@ class TOPPOpenSwathAssayGenerator :
     }
 
     // Load Unimod file
-    if (enable_ms1_uis_scoring || enable_ms2_uis_scoring)
+    if (enable_ipf)
```
hroest: I wonder whether it may make sense to control the unimod file even when not using IPF - wouldnt it give you more control over how OpenMS interprets the mods? So only allowing a custom unimod file when IPF is active maybe too restrictive, why not decouple the two options?

hroest:
:+1: I would be ready to merge, I only have one more question regarding the utility of a custom unimode file even when IPF is turned off.

src/topp/OpenSwathAssayGenerator.cpp
grosenberger: Yes, I think that is a good idea, especially when importing data from SpectraST or generic transition lists that use bracket annotation. I will provide a patch to enable this option.

src/topp/OpenSwathAssayGenerator.cpp
```diff
@@ -229,20 +230,19 @@ class TOPPOpenSwathAssayGenerator :
...
 
+    // Require Unimod XML file when running IPF to prevent accidential mistakes
+    if (enable_ipf && unimod_file.empty())
+    {
+      throw Exception::InvalidParameter(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Please provide a valid Unimod XML file for IPF.");
+    }
+
     // Load Unimod file
-    if (enable_ipf)
+    if (!unimod_file.empty())
```
hroest: :+1:

PR #3099, opened 2017-12-14
[FEATURE] OpenSwathDecoyGenerator improvements 

grosenberger:
Over the past years, we have made several adjustments to ``OpenSwathDecoyGenerator`` to deal with more complex spectral libraries. One particular problem was non-canonical PTMs, particularly at the C-/N-term. We introduced several new parameters for the TOPP tool, which were used in combination to exclude some of the difficult peptides when they failed checks. Finding the right combination was sometimes frustrating for users.

This PR attempts to remove most of these parameters and to enable only sensible combinations. Further, some changes to the algorithm ensure that for the vast majority of applications, the default parameters work reliably out of the box, without the need for tweaking.

Further, ``OpenSwathDecoyGenerator`` now supports all other formats for spectral libraries as well. Generating decoys for the PHL takes only 3-5min depending on the format.

Since the algorithmic changes are substantial and the default parameters are modified, we should do further testing to ensure that all challenges are covered. I will test with SWATH-MS Gold Standard, PTM-SWATH-MS Gold Standard and the SWATH-MS interlab study / PHL data sets within the next days and report back.

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
@@ -149,7 +149,7 @@ namespace OpenMS
     {
       // Block tryptic residues and N-/C-terminus from shuffling
       IndexType idx = MRMDecoy::find_all_tryptic_and_term(peptide.sequence);
-      
+
       shuffled = peptide;
       std::vector<Size> peptide_index;
       for (Size i = 0; i < peptide.sequence.size(); i++)
@@ -365,6 +365,8 @@ namespace OpenMS
     startProgress(0, exp.getPeptides().size(), "Generating decoy peptides");
     for (Size pep_idx = 0; pep_idx < exp.getPeptides().size(); ++pep_idx)
     {
+      setProgress(++progress);
+
       OpenMS::TargetedExperiment::Peptide peptide = exp.getPeptides()[pep_idx];
 
       peptide.id = decoy_tag + peptide.id;
@@ -415,7 +417,6 @@ namespace OpenMS
       }
 
       peptides.push_back(peptide);
-      setProgress(++progress);
     }
     endProgress();
     dec.setPeptides(peptides); // temporary set peptides, overwrite later again!
@@ -432,6 +433,8 @@ namespace OpenMS
     for (MRMDecoy::PeptideTransitionMapType::iterator pep_it = peptide_trans_map.begin();
          pep_it != peptide_trans_map.end(); ++pep_it)
     {
+      setProgress(++progress);
+
       String peptide_ref = pep_it->first;
       String decoy_peptide_ref = decoy_tag + pep_it->first; // see above, the decoy peptide id is computed deterministically from the target id
       const TargetedExperiment::Peptide target_peptide = exp.getPeptideByRef(peptide_ref);
@@ -497,7 +500,6 @@ namespace OpenMS
           LOG_DEBUG << "[peptide] Skipping " << decoy_tr.getPeptideRef() << " due to missing annotation" << std::endl;
         }
       } // end loop over transitions
-      setProgress(++progress);
     } // end loop over peptides
     endProgress();
 
--- a/src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
@@ -117,7 +117,7 @@ namespace OpenMS
 
   OpenMS::TargetedExperiment::Peptide MRMDecoy::shufflePeptide(
     OpenMS::TargetedExperiment::Peptide peptide, double identity_threshold, int seed,
-    int max_attempts, bool replace_aa_instead_append)
+    int max_attempts)
   {
 #ifdef DEBUG_MRMDECOY
     std::cout << " shuffle peptide " << peptide.sequence << std::endl;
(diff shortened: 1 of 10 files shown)


src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -117,6 +147,9 @@ namespace OpenMS
     while (MRMDecoy::AASequenceIdentity(peptide.sequence, shuffled.sequence) > identity_threshold &&
            attempts < max_attempts)
     {
+      // Block tryptic residues and N-/C-terminus from shuffling
+      IndexType idx = MRMDecoy::find_all_tryptic_and_term(peptide.sequence);
```
grosenberger: For the ``shuffle`` method, we now also exclude N- and C-terminal residues from being shuffled. This has the benefit that if the terminal residues are modified, we don't have to worry about these modifications being shuffled.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -184,53 +217,41 @@ namespace OpenMS
...
       ++attempts;
 
-      // If our attempts have failed so far, we will append two random AA to
+      // If our attempts have failed so far, we will mutate a random AA of
       // the sequence and see whether we can achieve sufficient shuffling with
-      // these additional AA added to the sequence.
+      // the new sequence.
       if (attempts % 10 == 9)
       {
-        if (replace_aa_instead_append)
```
grosenberger: Previously, the ``shuffle`` method appended two random residues if the sequence similarity did not reach a lower threshold. However, this did not guarantee that the N-term remained constant or was even a tryptic residue, which might potentially introduce bias with the b- and y-ion series score of OpenSWATH.

For IPF, we implemented an approach to swap individual residues and the implementation here follows a similar strategy. Instead of appending, existing (non-modified and non-terminal) residues are randomly changed. A similar strategy (or ``mutated`` approach) [has been suggested](http://www.mcponline.org/content/suppl/2017/10/25/RA117.000314.DC1/4947_1_supp_14434_zy1gl9.pdf) by others as well.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -329,50 +349,60 @@ namespace OpenMS
...
       if (!peptide.getPeptideGroupLabel().empty())
       {
         peptide.setPeptideGroupLabel(decoy_tag + peptide.getPeptideGroupLabel());
       }
 
       if (method == "pseudo-reverse")
       {
-        peptide = MRMDecoy::pseudoreversePeptide(peptide);
+        // exclude peptide if it has C/N terminal modifications because we can't do a (partial) reverse
+        if (MRMDecoy::has_CNterminal_mods(peptide))
```
grosenberger: For the ``pseudo-reverse`` and ``reverse`` methods, any C-/N-terminally modified peptides will be excluded, as they might not be compatible.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -329,50 +349,60 @@ namespace OpenMS
...
       {
         peptide = MRMDecoy::shufflePeptide(peptide, identity_threshold, -1, max_attempts);
       }
+
       for (Size prot_idx = 0; prot_idx < peptide.protein_refs.size(); ++prot_idx)
       {
         peptide.protein_refs[prot_idx] = decoy_tag + peptide.protein_refs[prot_idx];
       }
 
-      if (MRMDecoy::AASequenceIdentity(original_sequence, peptide.sequence) > identity_threshold)
```
grosenberger: The amino acid sequence identity threshold is not a control anymore, but an optimization parameter for the ``shuffle`` method. The difference to the target peptides is directly controlled by comparing the fragment ion m/z.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -442,28 +471,17 @@ namespace OpenMS
           {
             if (std::fabs(tr.getProductMZ() - decoy_tr.getProductMZ()) < similarity_threshold)
             {
-              if (!exclude_similar)
```
grosenberger: If the fragment ions of targets and decoys are too similar, the corresponding decoy peptide will directly be excluded.

src/tests/class_tests/openms/source/MRMDecoy_test.cpp
```diff
@@ -93,6 +93,25 @@ START_SECTION((std::vector<std::pair<std::string::size_type, std::string> > find
 
 END_SECTION
 
+START_SECTION((std::vector<std::pair<std::string::size_type, std::string> > find_all_tryptic_and_term(std::string sequence)))
+{
+  MRMDecoy gen;
+
+  String sequence = "TRESTPEPTIKDE";
+  std::vector<pair<std::string::size_type, std::string> > tryptic_results = gen.find_all_tryptic_and_term(sequence);
+  std::vector<pair<std::string::size_type, std::string> > tryptic_control = boost::assign::list_of(std::make_pair(0, "T")) (std::make_pair(1, "R")) (std::make_pair(5, "P")) (std::make_pair(7, "P")) (std::make_pair(10, "K")) (std::make_pair(12, "E"));
```
hroest: I am not sure this is really doing what it is supposed to, Trypsin cuts at `[K|R]` except when followed by `P` so this should really be

```
(std::make_pair(0, "T")
(std::make_pair(1, "R")
(std::make_pair(10, "K")
(std::make_pair(12, "E")
```

or am I missing something here?

src/topp/OpenSwathDecoyGenerator.cpp
```diff
@@ -162,33 +211,86 @@ class TOPPOpenSwathDecoyGenerator
...
 
     MRMDecoy decoys = MRMDecoy();
+    decoys.setLogType(ProgressLogger::CMD);
 
-    std::cout << "Generate decoys" << std::endl;
-    decoys.generateDecoys(targeted_exp, targeted_decoy, method, decoy_tag, identity_threshold, max_attempts, mz_threshold, mz_shift, exclude_similar, similarity_threshold, remove_CNterm_mods, precursor_mass_shift, allowed_fragment_types, allowed_fragment_charges, enable_detection_specific_losses, enable_detection_unspecific_losses, remove_unannotated);
+    LOG_INFO << "Generate decoys" << std::endl;
+    decoys.generateDecoys(targeted_exp, targeted_decoy, method, decoy_tag, identity_threshold, max_attempts, product_mz_threshold, product_mz_shift, similarity_threshold, precursor_mz_shift, allowed_fragment_types, allowed_fragment_charges, enable_detection_specific_losses, enable_detection_unspecific_losses);
 
-    if (append)
```
grosenberger: We check that there is a minimum fraction of decoys vs targets that survive the filtering. With the default parameters and the PHL, about 95% of the targets are covered.

src/tests/class_tests/openms/source/MRMDecoy_test.cpp
hroest: Due to their strong fragment ion signal, it may make sense to keep Prolines where they are, is this what is intended here?

grosenberger: We adapted this function from the ``SpectraST`` decoy generator: http://pubs.acs.org/doi/full/10.1021/pr900947u

timosachsenberg: gen.find_all_tryptic_and_term(sequence); should be finAllTrypticAndTerm
by the way, does this use the EnzymaticDigestion capabilities in the core library? And if not, are there specific reason?

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
timosachsenberg: In the DecoyDatabase, I think we only use reverse. Would it make sense to reuse (or generalize) MRMDecoy to be applicable to non-MRM tools in OpenMS ?

grosenberger: There also seems to be a ``shuffle`` method in DecoyDatabase on protein sequence-level. It might be worthwhile to consider adapting the methods also there, however there are important differences. Here, we need to already consider PTMs, whereas with DecoyDatabase, the search engines will take care about this. What might be worthwhile to investigate is to also keep the N- and C-terminus fixed in DecoyDatabase, as this will generate more similar decoys. But this of course depends on the assumptions of the search engines.

src/tests/class_tests/openms/source/MRMDecoy_test.cpp
grosenberger: I can rename the function. Here, we don't do tryptic digestion, but only need to recover the indices of all K, R & P residues. We wrote the original function to represent the SpectraST implementation. I'm not sure whether it is worthwhile to port this to the full EnymaticDigestion class.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
timosachsenberg: ok I see.

src/tests/class_tests/openms/source/MRMDecoy_test.cpp
timosachsenberg: Ok it just seemed similar to what tokenize_ etc. does and, thus, might help to reduce some code duplication (without having looked deeper into the code)

timosachsenberg:
feel free to merge

src/tests/class_tests/openms/source/MRMDecoy_test.cpp
hroest: ok, but lets call it something different than tryptic termini as it clearly also finds other AA which are not tryptic termini (such as P)

grosenberger: Good suggestion. The functions are now called findFixedResidues and findFixedAndTermResidues.

grosenberger:
Thanks for the review. I'm still working on the validation and further tweaks might be necessary.

grosenberger:
I now conducted the comparison with some of our test data sets and tweaked the parameters.

On both SGS and PTM-SGS data sets, the results are essentially identical. When applying the PHL to an older 32-windowed SWATH-MS runs, the results are slightly more conservative (which is desired). There does not seem to be a big difference in terms of individual score distributions.

This is ready for merging from my side.

PR #3100, opened 2017-12-16
[FEATURE] Automated calibration curve fitting for Absolute Quantitation

dmccloskey:
Enables the ability to automatically optimize the points chosen for a calibration curve based on a set of user thresholds (e.g., bias, correlation, number of points, etc.).  The algorithm optimizes each curve iteratively using algorithms described in `MRMRTNormalizer`.


dmccloskey:
Does anyone know what caused the Travis failure?

hroest:
see http://cdash.openms.de/viewBuildError.php?buildid=250218

src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
```diff
@@ -64,7 +64,7 @@ namespace OpenMS
         weightData(data_weighted);
       }
 
-      size_t size = data.size();
+      size_t size = data_weighted.size();
```
hroest: nice! any regression test for this bug?

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -274,151 +407,266 @@ namespace OpenMS
...
+          optimized_params);
+      }
+      else
+      {
+        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
+          String("Method ") + outlier_detection_method_ + " is not a valid method for optimizeCalibrationCurveIterative");
+      }
+
+      //DEBUG
+      std::cout << "R = " << std::to_string(correlation_coefficient) << ".  "
```
hroest: comment out debug?

dmccloskey: Yes.  That should be commented out.

src/pyOpenMS/pxds/AbsoluteQuantitation.pxd
```diff
@@ -29,7 +29,7 @@ cdef extern from "<OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitation.h>" namespa
             Param & transformation_model_params,
             Param & optimized_params) nogil except +
 
-        void optimizeCalibrationCurves(
-            libcpp_map[ String, libcpp_vector{ AQS_featureConcentration ]] & components_concentrations) nogil except +
+        # void optimizeCalibrationCurves(
```
dmccloskey: Did it build after the `{` typo was correct?  I would like to be able to call this function from pyopenms.  Should I just return the changed input instead of changing in place?

src/pyOpenMS/pxds/AbsoluteQuantitation.pxd
```diff
@@ -5,6 +5,7 @@ from AbsoluteQuantitationStandards cimport *
 from Feature cimport *
 from FeatureMap cimport *
 from String cimport *
+from DefaultParamHandler cimport *
```
dmccloskey: No I can't believe I missed that!

hroest:
@dmccloskey please merge https://github.com/dmccloskey/OpenMS/pull/7 and then we also have pyOpenMS functionality.

hroest:
@dmccloskey it may make sense to set up your own pyOpenMS builds so that you can check your code yourself and also get newer functionality

dmccloskey:
@hroest That is a good suggestion.  I get a bit lazy with the PyOpenMS builds because they take so long.  I test them usually when I update my dockerized version of OpenMS, but that is usually only on a weekly basis.

hroest:
on my machine they take about 7 minutes after the library is built. You can speed it up using the arguments `-DPY_NUM_THREADS=2 -DPY_NUM_MODULES=4` during cmake

dmccloskey:
OK I will do that then!  Thanks.

hroest:
the issue is generally the memory consumption and that you begin swapping at some point. so you need to adjust the number of distinct compilation modules and the number of threads so that the compilation fits into memory. the values above were working for me, maybe start with a single thread in your case ...

dmccloskey:
@hroest Do you know why the original wrap for `optimizeCalibrationCurves` fails?  Is is because std::map is not mapped to a python type?  If that is the case, I can refactor to use a list instead if you think that would make things easier.

hroest:
@dmccloskey it is because we dont support nested types of `std::map< Type, std::vector<OtherType> >`, we currently can do `std::map< Type, std::vector<double> >` but not any complicated `OtherType`. Doing this automatically is nontrivial in the general case as you can imagine

dmccloskey:
OK sounds good.  If the pyopenms fix that you did does not work for some reason, I can just refactor the class in a following PR to allow for native python support using a `libcpp_vector [ AbsoluteQuantitationStandards::featureConcentrations]`

dmccloskey:
Also let me know when you can when this can be merged.  I would like to update my local version before my flight tomorrow morning :)

PR #3103, opened 2017-12-18
NumberOfJobs in quickbuild script & add "build" folders to gitignore 

oliveralka:
NumberOfJobs in quickbuild script. 
Add "build" folders to gitignore.


timosachsenberg:
@jpfeuffer will querying the number of jobs interfere with CI?

jpfeuffer:
Hmm I guess this breaks builds within scripts. We do not use the quickbuild script for CI though.
Maybe it is better as a command line parameter when calling the script?

oliveralka:
Ok, my thoughts were that it would be more user friendly this way - since it asks how many jobs you want to use.

timosachsenberg:
@hroest you initiated the idea of the quickbuild script. What is your take on this?

timosachsenberg:
I think both makes sense. e.g., QT has kind of a quick build script that shows a license text the user has to accept

hroest:
well but qt also allows you to skip this with specific commandline options. I think it would be more useful to allow both, e.g. to either use an env-variable or the first argument as the number of jobs and when empty then simply use 1 job. I think it is useful to have the quickbuild script being able to build OpenMS automatically, e.g. in docker containers and such.

oliveralka:
Is the new commit more like it? (test ran into time limit)

hroest:
yes that is what I had in mind

:+1: for merge

PR #3104, opened 2017-12-18
[doc] export OPENMS_DATA_PATH for .deb package

oliveralka:
Documentation: export OPENMS_DATA_PATH for .deb package

### DIFF
--- a/doc/doxygen/install/install-linux-bin.doxygen
+++ b/doc/doxygen/install/install-linux-bin.doxygen
@@ -7,10 +7,15 @@
   @section package_managers_linux Install via package managers
   Currently there are packaged versions of %OpenMS provided for Fedora, OpenSUSE, Debian, and Ubuntu <a href="https://pkgs.org/search/openms">available in the corresponding package managers</a> of these distributions (they might require a lookup in thirdparty "science" repositories).
   For other GNU/Linux distributions or to obtain the most recent version of the library you need to build your own %OpenMS.
+  
+  To ensure the tool functionality please add the OPENMS_DATA_PATH variable. 
+  \code
+  export OPENMS_DATA_PATH=/usr/shared/OpenMS 
+  \endcode
 
   @note Please note that these packages are not directly maintained by us and we can not guarantee the same behaviour as when building it on your own.
   Also their availability is subject to change and support might be limited (due to unforeseen or untested behaviour).
-
+  
   @section build_your_own_linux Build it on your own
   Please check <a href="http://www.openms.de/">OpenMS.de</a> for details on how to obtain the source code and @subpage install_linux on how to build %OpenMS.
 
--- a/doc/doxygen/install/install-linux-bin.doxygen
+++ b/doc/doxygen/install/install-linux-bin.doxygen
@@ -12,6 +12,11 @@
   \code
   export OPENMS_DATA_PATH=/usr/share/OpenMS 
   \endcode
+  
+  Executables for THIRDPARTY applications can be found in:
+  \code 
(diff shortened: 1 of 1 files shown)


doc/doxygen/install/install-linux-bin.doxygen
```diff
@@ -7,10 +7,15 @@
   @section package_managers_linux Install via package managers
   Currently there are packaged versions of %OpenMS provided for Fedora, OpenSUSE, Debian, and Ubuntu <a href="https://pkgs.org/search/openms">available in the corresponding package managers</a> of these distributions (they might require a lookup in thirdparty "science" repositories).
   For other GNU/Linux distributions or to obtain the most recent version of the library you need to build your own %OpenMS.
+  
+  To ensure the tool functionality please add the OPENMS_DATA_PATH variable. 
+  \code
+  export OPENMS_DATA_PATH=/usr/shared/OpenMS
```
jpfeuffer: Is it share or shared?

oliveralka: share is right

timosachsenberg:
develop or release branch?

jpfeuffer:
Release

timosachsenberg:
will put on release branch

timosachsenberg:
is /usr/bin/SEARCHENGINES kind of a misnomer - we now not only ship search engines?
Should we update the docu that the user needs to set the path to the individual subfolders?

jpfeuffer:
Yes I agree with both. Is a relic from old times. Should be renamed and documented.

timosachsenberg:
k should this be put into this release?

oliveralka:
release +1

jpfeuffer:
How about adding the issue about Searchengines here?

jpfeuffer:
How about adding the issue about Searchengines here?

jpfeuffer:
Naming issue would be here:
https://github.com/OpenMS/OpenMS/blob/develop/cmake/install_macros.cmake#L143
Maybe we should change it to share/OpenMS/THIRDPARTY like on Windows.

timosachsenberg:
will this make debian guys angry? putting binaries there?

jpfeuffer:
Maybe. Do they even ship the thirdparties?

jpfeuffer:
Also affects macOS. Uses same mechanism.

hroest:
my preference would be to ship OpenMS "batteries included" but from a debian perspective that does not make sense. I am pretty sure they would like to break it all down into individual packages and they already do have some in debian:

- https://packages.debian.org/sid/comet-ms
- https://packages.debian.org/sid/tandem-mass

so if we go through them, we would have to package all the third party tools / get somebody to package it. We then have the issue of (i) slowing it all down and (ii) not having control over the versions but of course that would be "right way to do it" 

> will this make debian guys angry? putting binaries there?

we will just need to tell them and I assume they will delete them. There is anyways a policy in debian that you cannot ship binaries, all binaries have to be compiled from source and have to be under an open source license.

jpfeuffer:
I agree. They will change it if it does not agree with their policies. Completely modularizing indeed will take some time. Anyway, this pr is mainly about the naming. Thirdparties are shipped in any case.

oliveralka:
@jpfeuffer - I'm not entirely sure of the consensus related to this pull request. Any further changes necessary?

jpfeuffer:
This could be changed.

_Sent from my Xperia Z5 Compact using [FastHub-Libre](https://github.com/thermatk/FastHub-Libre/)_

oliveralka:
Closed merge #3122

PR #3105, opened 2017-12-19
[FIX] fix comet adapter test caused by uninitialized value

timosachsenberg:


timosachsenberg:
FYI @Leon-Bichmann

jpfeuffer:
Oh I still would have waited for CI. But this should trigger a RC build now.

timosachsenberg:
ah ok I knew that CI would fail because the relaese branch still contains non working CI targets.
I merged because the first ones passed

jpfeuffer:
Ah yes ok

PR #3109, opened 2017-12-20
update .gitignore

oliveralka:
Adds contrib submodule to gitignore


oliveralka:
I think that was a merge issue in my personal branch - I'm about to check this. Seems that submodules do not have to be added to .gitignore usually.

PR #3112, opened 2017-12-20
[FEATURE] CsvFile::addRow() and CsvFile::clearBuffer()

pasqoo:
This PR adds two methods to the CsvFile class.

`addRow()` adds a row to the `buffer_` inherited from _TextFile_. It takes into account the values of `itemenclosed_` and `itemseperator_`.

`clearBuffer()` simply calls `buffer_.clear()`.


timosachsenberg:
what about renaming fload and fstore to load an store as in the TextFile base class?
instead of clearBuffer() I would also probably just rename it to clear()

pasqoo:
@timosachsenberg Updated. I had to change other files to avoid ambiguities (those other classes had methods with the same name). :smile:

PR #3114, opened 2017-12-20
Feature/crux

hroest:
searches MS data using tide-search and percolator (using crux)


timosachsenberg:
nice. what you could do is to provide a tighter integration to our EnzymeDB.

hroest:
> nice. what you could do is to provide a tighter integration to our EnzymeDB.

yes, but its easier this way and I am not sure that there is a 1:1 mapping

timosachsenberg:
I think there doesn't need to be a 1:1 mapping for all enzymes but if we want to reprocess the data we should have the OpenMS enzymes in the ID files.

See e.g. https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/CHEMISTRY/Enzymes.xml where we map the IDs for the different search engines. 
```
    vector<String> all_enzymes;
    ProteaseDB::getInstance()->getAllXTandemNames(all_enzymes);
    registerStringOption_("enzyme", "<choice>", "Trypsin", "The enzyme used for peptide digestion.", false);
    setValidStrings_("enzyme", all_enzymes);
```
an we later store:
```
search_parameters.digestion_enzyme = *(ProteaseDB::getInstance()->getEnzyme(enzyme_name));
```

timosachsenberg:
so basically I am suggesting to add the names to a method ->getAllCruxNames(...)
and store the enzymes in Enzymes.XML

timosachsenberg:
I would be fine with the merge now but would like to see this minor change at some point

PR #3116, opened 2017-12-21
Feature/mascot online

hroest:
Adopt the MascotAdapterOnline for the new version of mascot (2.6)

- this is in preparation for the qt5 branch where we need a test for this part of OpenMS
- tests are only turned on with `-DWITH_MASCOT_TEST=On` which should not be used on a regular basis (otherwise the Mascot guys will get angry)


hroest:
also: it now compiles with `MASCOTREMOTEQUERY_DEBUG` defined

timosachsenberg:
Very cool! Thanks

timosachsenberg:
looks good. feel free to merge

timosachsenberg:
cool will try to sync the qt5 branch

Issue #2275, opened 2016-11-01
Make LogStream thread-safe


lars20070:
After building OpenMS under VS2015 (debug mode) in Windows Server 2012, the `LogStream` class test fails as seen in the screenshot below.

![screenshot from 2016-10-31 17-01-41](https://cloud.githubusercontent.com/assets/6149805/19883008/f1582f1a-a011-11e6-8d2c-8c6c8fea1e04.png)

timosachsenberg:
@lars20070 is this still an issue? if yes could you debug into it and post the stack trace?

jpfeuffer:
I don't know what exactly changed but we get this now too on our build machines. Seems to be an OpenMP problem.

timosachsenberg:
maybe change
#pragma omp parallel
to
#pragma omp parallel for
?

timosachsenberg:
hmm locking at the code more closely, there seem to be some
critical sections missing.
E.g. LogStreamBuf::overflow doesn't seem to be thread safe etc.

jpfeuffer:
Hm yes maybe (but did you see that it calls sync, which is critical? I'm not very trained in parallel programming though)
But the pragma seems definitely wrong doesn't it? What does just parallel do with the for loop?

hendrikweisser:
Let's assume this got fixed in #3126 and close this ticket until we hear otherwise.

PR #3126, opened 2017-12-22
[FIX?] LogStream_test parallel for

jpfeuffer:
Indeed this test gave random errors in around 1 in 5 executions with OpenMP on Win10 VS2015.
I don't know the exact reason and probably big parts of the class are not really threadsafe (as found by @timosachsenberg ) but changing the pragma to parallel for gets rid of the random errors (at least there were 0 in 500 executions).
See #2275

### DIFF
--- a/src/tests/class_tests/openms/source/LogStream_test.cpp
+++ b/src/tests/class_tests/openms/source/LogStream_test.cpp
@@ -87,11 +87,12 @@ START_SECTION(([EXTRA] OpenMP - test))
   Log_info.insert(stream_by_logger);
   Log_info.remove(cout);
 
-#ifdef _OPENMP
-omp_set_num_threads(8);
-#pragma omp parallel
-#endif
+
   {
+    #ifdef _OPENMP
+	omp_set_num_threads(8);
+    #pragma omp parallel for
+    #endif
     for (int i=0;i<10000;++i)
     {
       LOG_DEBUG << "1\n";


timosachsenberg:
Release or development?

jpfeuffer:
both please

timosachsenberg:
I could put it on release this evening

src/tests/class_tests/openms/source/LogStream_test.cpp
```diff
@@ -87,11 +87,12 @@ START_SECTION(([EXTRA] OpenMP - test))
...
-omp_set_num_threads(8);
-#pragma omp parallel
-#endif
+
   {
+    #ifdef _OPENMP
+	omp_set_num_threads(8);
+    #pragma omp parallel for
+    #endif
     for (int i=0;i<10000;++i)
```
hroest: what about making the `i` an unsigned int? I think otherwise its incompatible with OpenMP 2.x

timosachsenberg: No it needs to be signed for compatibility afaik

timosachsenberg: https://stackoverflow.com/questions/2820621/why-arent-unsigned-openmp-index-variables-allowed

timosachsenberg:
applied to release/2.3.0

PR #3129, opened 2017-12-28
[FEATURE] Simplified parameters for OpenSwathWorkflow

grosenberger:
For most scenarios, the default parameters in OpenSwathWorkflow will produce reasonable results. However, when using DIA data with overlapping precursor isolation windows, the parameter ``min_upper_edge_dist`` needs to be set manually (or the check needs to be disabled).

This PR suggests that we automatically estimate this parameter by default, if all overlaps are identical and if the SONAR scores are not enabled.

Further, this PR suggests to rename the parameters for MS1 and IPF scoring and to display them with the reduced set of parameters.

### DIFF
--- a/src/utils/OpenSwathWorkflow.cpp
+++ b/src/utils/OpenSwathWorkflow.cpp
@@ -146,8 +146,9 @@ static bool SortPairDoubleByFirst(const std::pair<double,double> & left, const s
   -mz_extraction_window to your instrument resolution, which can be in Th or
   ppm (using -ppm).
 
-  Furthermore, if you wish to use MS1 information, use the -use_ms1_traces flag
-  and provide an MS1 map in addition to the SWATH data.
+  Furthermore, if you wish to use MS1 information, use the -enable_ms1 flag
+  and provide an MS1 map in addition to the SWATH data. If IPF transition-
+  level scoring should be conducted, set the -enable_ipf flag.
 
   If you encounter issues with peak picking, try to disable peak filtering by
   setting -Scoring:TransitionGroupPicker:compute_peak_quality false which will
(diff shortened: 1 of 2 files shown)


src/utils/OpenSwathWorkflow.cpp
```diff
@@ -416,8 +417,8 @@ class TOPPOpenSwathWorkflow
     registerInputFile_("swath_windows_file", "<file>", "", "Optional, tab separated file containing the SWATH windows for extraction: lower_offset upper_offset \\newline 400 425 \\newline ... Note that the first line is a header and will be skipped.", false, true);
     registerFlag_("sort_swath_maps", "Sort of input SWATH files when matching to SWATH windows from swath_windows_file", true);
 
-    registerFlag_("use_ms1_traces", "Extract the precursor ion trace(s) and use for scoring", true);
-    registerFlag_("enable_uis_scoring", "Enable additional scoring of identification assays", true);
+    registerFlag_("enable_ms1", "MS1: Set this flag if precursor ion trace(s) should be extracted and used for scoring", false);
```
grosenberger: @hroest Do you think this name change makes sense or do you prefer the old (or a different) name?

src/utils/OpenSwathWorkflow.cpp
```diff
@@ -432,7 +433,7 @@ class TOPPOpenSwathWorkflow
     registerOutputFile_("out_chrom", "<file>", "", "Also output all computed chromatograms output in mzML (chrom.mzML) or sqMass (SQLite format)", false, true);
     setValidFormats_("out_chrom", ListUtils::create<String>("mzML,sqMass"));
 
-    registerDoubleOption_("min_upper_edge_dist", "<double>", 0.0, "Minimal distance to the edge to still consider a precursor, in Thomson", false, true);
+    registerDoubleOption_("min_upper_edge_dist", "<double>", -1, "Minimal distance to the edge to still consider a precursor, in Thomson (-1 means that the parameter is estimated automatically from the input). This parameter is disabled in SONAR mode.", false, true);
```
grosenberger: ``-1`` enables automated estimation

src/utils/OpenSwathWorkflow.cpp
```diff
@@ -901,6 +901,38 @@ class TOPPOpenSwathWorkflow
     }
     std::sort(sw_windows.begin(), sw_windows.end(), SortPairDoubleByFirst);
 
+    // (iv) Estimate min_upper_edge_dist from swathes
+    if (sonar) // We skip this step in SONAR-mode
```
grosenberger: @hroest In SONAR-mode, we don't require min_upper_edge_dist at all, right? Is there any other preferred behavior?

src/utils/OpenSwathWorkflow.cpp
hroest: well why not just make it the default option and  use `registerStringOption_("enable_ms1" ... ` ?

grosenberger: That's a good suggestion. If no MS1 maps are there, I believe this will be skipped anyway, right? I will make the same changes for ``enable_ipf`` as well, because this will run only if identification transitions are present in the library.

PR #3125, opened 2017-12-22
Feature/qt5 update

timosachsenberg:

PR #3130, opened 2017-12-29
Update qt5 branch with develop

hroest:
I am not so happy about 0bdd21f but it seems otherwise the pyOpenMS tests fail


hroest:
replaces #3125 as it actually compiles and all tests run through

ok, we got travis workign while codacy and jenkins complain. I think that is fair as jenkins and the windows builds dont support qt5 at the moment. So from my perspective this seems ready for a merge, there is still an issue of a single pyOpenMS class not working on travis but I think that should not break the whole merge.

PR #3133, opened 2017-12-31
[FIX] changed names in AbsoluteQuantitationStandards to match

dmccloskey:
Changed the names of run_id and component_/IS_id to match the names in the other AbsoluteQuantitation classes (i.e., _id to _name).  This should make things more clear for developers and users alike.


hroest:
please also fix and ajdust the pyOpenMS bindings

dmccloskey:
Good catch.  They are updated now.

Issue #757, opened 2014-03-25
clang compiler warnings


mcmenace:
In the following we collect clang warings that we believe are false-positives or at least warnings that are low priority for removal.

hroest:
update for OpenMS 2.0

errors for the library only:

```
      1 [-Woverloaded-virtual]
      1 [-Wsign-compare]
      1 [-Wunknown-pragmas]
      1 [-Wunused-value]
      3 [-Wc++11-extensions]
      3 [-Wunused-function]
      5 [-Wcovered-switch-default]
      5 [-Wmissing-noreturn]
      6 [-Wconditional-uninitialized]
      6 [-Wunreachable-code]
     22 [-Wunused-macros]
    101 [-Wshadow]
```

for the full build (+tools +tests)

```
      1 [-Wlanguage-extension-token]
      1 [-Woverloaded-virtual]
      1 [-Wsign-compare]
      1 [-Wunknown-pragmas]
      1 [-Wunused-value]
      1 [-Wvariadic-macros]
      2 [-Wheader-hygiene]
      2 [-Wunused-exception-parameter]
      3 [-Wc++11-extensions]
      3 [-Wunused-function]
      5 [-Wcovered-switch-default]
      6 [-Wconditional-uninitialized]
      7 [-Wextra-semi]
      7 [-Wmissing-noreturn]
     24 [-Wunused-macros]
     28 [-Wunreachable-code]
    332 [-Wshadow]
```

and ordered by file (only files with 5 or more warnings):

```
      5 src/tests/class_tests/openms/source/IDFilter_test.cpp
      5 src/tests/class_tests/openms/source/ParamXMLFile_test.cpp
      5 src/topp/FileMerger.cpp
      5 src/topp/SpecLibSearcher.cpp
      6 src/tests/class_tests/openms/source/InclusionExclusionList_test.cpp
      6 src/tests/class_tests/openms/source/IsotopeDistribution_test.cpp
      6 src/tests/class_tests/openms/source/SampleTreatment_test.cpp
      6 src/utils/QCCalculator.cpp
      7 src/openms/source/ANALYSIS/ID/ProtonDistributionModel.cpp
      8 src/openms/source/ANALYSIS/ID/HiddenMarkovModel.cpp
      8 src/tests/class_tests/openms/source/MRMDecoy_test.cpp
     11 src/topp/MzTabExporter.cpp
[...]

PR #3139, opened 2018-01-05
[BUILD] remove double promotion warnings

hroest:
a small fix to reduce clang warnings, see  #757


hroest:
@timosachsenberg @jpfeuffer any comments on this?

jpfeuffer:
I like it. Getting rid of them is currently out of scope and sometimes hardly possible.

timosachsenberg:
fine with me... we might think about re-adding them for vectorizable code at some point

PR #3143, opened 2018-01-06
[FIX] fix compile error on some systems

### LINKED ISSUE #3141: MRMScoring compile error
I get a compile error in method:
https://github.com/OpenMS/OpenMS/blob/3fd677a2d8333a832749b29e32681d45d6cd3d8d/src/openswathalgo/include/OpenMS/ANALYSIS/OPENSWATH/OPENSWATHALGO/ALGO/MRMScoring.h#L148
when a nullptr is returned for the string (before a 0 was returned which is also wrong)

a quick fix would be to return an empty string but the method looks a bit fishy.
@grosenberger can you give this method a quick check? is it really needed? wouldn't returning a double be more appropriate?

timosachsenberg:
fixes #3141


hroest:
looks good to me, any idea why windows build failed?

timosachsenberg:
hmm I don't see the reason why. In Jenkins it says all tests passed

PR #3145, opened 2018-01-08
[FEATURE] AbsoluteQuantitationStandardsFile

pasqoo:
Load files containing runConcentration data, and save it to an `AbsoluteQuantitationStandards::runConcentration` vector. One element for each line.


pasqoo:
@timosachsenberg @hroest Hey guys, sorry to bother you with these tests' results: I can't understand where the problem resides. There are a bunch of warnings, but none of them seem directly related to _AbsoluteQuantitationStandardsFile_. Am I missing something?
Here the warnings: http://cdash.openms.de/viewBuildError.php?type=1&buildid=253105
The build succeeds with msvc-14, but fails with msvc-15

hroest:
looks like just a single test failed: http://cdash.openms.de/viewTest.php?onlyfailed&buildid=253105 but the `LogStream_test` does fail randomly, so I am not really worried.

hroest:
rebuild win

src/tests/class_tests/openms/data/150516_calibration_concentrations.csv
```diff
@@ -0,0 +1,9273 @@
...
+150516_CM1_Level3,AICAr.AICAr_1.Light,AICAr.AICAr_1.Heavy,0.4,1,uM,1
+150516_CM1_Level3,akg.akg_1.Light,akg.akg_1.Heavy,0,1,uM,1
+150516_CM1_Level3,ala-L.ala-L_1.Light,ala-L.ala-L_1.Heavy,0,1,uM,1
+150516_CM1_Level3,amp.amp_1.Light,amp.amp_1.Heavy,0,1,uM,1
+150516_CM1_Level3,arg-L.arg-L_1.Light,arg-L.arg-L_1.Heavy,0,1,uM,1
+150516_CM1_Level3,asn-L.asn-L_1.Light,asn-L.asn-L_1.Heavy,0,1,uM,1
+150516_CM1_Level3,asp-L.asp-L_1.Light,asp-L.asp-L_1.Heavy,0,1,uM,1
+150516_CM1_Level3,atp.atp_1.Light,atp.atp_1.Heavy,8,1,uM,1
+150516_CM1_Level3,btn.btn_1.Light,btn.btn_1.Heavy,0,1,uM,1
+150516_CM1_Level3,camp.camp_1.Light,camp.camp_1.Heavy,0,1
```
hroest: do you need a 10k+ line test file just to test whether the parser works?

src/openms/include/OpenMS/FORMAT/AbsoluteQuantitationStandardsFile.h
```diff
@@ -0,0 +1,85 @@
...
+public:
+    AbsoluteQuantitationStandardsFile();
+    virtual ~AbsoluteQuantitationStandardsFile();
+
+    /**
+      @brief Load runConcentration data from a file and save it in memory.
+
+      @param[in] filename The file path from which the method is reading the data
+      @param[out] run_concentrations Where the runConcentration data is going to be saved
+    */
```
hroest: I think you should document the format of the file more specifically, e.g. what headers are required in the file and what they represent

hroest:
looks good overall, I only have a few comments

src/tests/class_tests/openms/data/150516_calibration_concentrations.csv
pasqoo: It is going to be used also on another class, _AbsoluteQuantitationStandards_.
But I could create dummy data inside the tests files, or at least strip this file down.

pasqoo:
@hroest Thanks for your review. I updated the code.
I removed the old test data file, and I added a stripped down version of it (~350 lines).
I also renamed the file so that it's obvious to which class it pertains.

pasqoo:
Decided to strip the data file even more. In case I'll need more data in other classes, I'll add new files with the same name as the requiring class.

PR #3146, opened 2018-01-08
[BUILD] pyOpenMS build script for Linux

hroest:
docker script to produce the manylinux wheels

### DIFF
--- a/tools/create-manylinux.sh
+++ b/tools/create-manylinux.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # Create manylinux packages from current release using the docker image
 #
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Can you have a quick look on the codacy suggestions? They seem reasonable. I will try to create a job and add it to our release procedure.

hroest:
hmm I would prefer to leave it as is as I have just used it to generate the wheels and I dont want to re-create all the wheels to test (it takes a few hours)

jpfeuffer:
Ah of course, I meant future releases only. ;)

hroest:
merge @jpfeuffer  ?

PR #3148, opened 2018-01-08
[BUILD] make pychecker compatible with Py3

### LINKED ISSUE #3144: Tools scripts not compatible with Python 3
I was trying to build pyOpenMS on Windows and tried to run the PythonChecker.
Since I was using python3 as the PYTHON_EXECUTABLE I just used that for the checker, too. There I noticed that it of course was written for python2.
Now just the question how we should go about this? Especially with a virtualenv e.g. to keep build machines clean, it would be nicer to be able to use the same python interpreter/virtualenv for the scripts. Which probably means that we have to duplicate all the scripts with py3 versions?
What do you think?

hroest:
fixes #3144

### DIFF
--- a/tools/PythonCheckerLib.py
+++ b/tools/PythonCheckerLib.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 
 def parse_pxd_file(path):
     import pdb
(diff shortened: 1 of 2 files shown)


hroest:
rewrote the script so that it works with both Python2 and Python3, so we can use the same script for all versions.

jpfeuffer:
Great thanks a lot! I will merge and try.

PR #3149, opened 2018-01-09
[FEATURE] ChromeleonFile

pasqoo:
Load Chromeleon HPLC text file and save it into a `MSExperiment`.


pasqoo:
Support on which _CVTerm_ s to use to save the metadata informations would be of great help.
As you can see here https://github.com/OpenMS/OpenMS/pull/3149/files#diff-9a5d164ee99681e755b1557e7b4e39d7R76 I went a bit general, using just `metaValue`, and I wonder if there's a better place for all those info.

hroest:
looks like it passes under Windows but not under Linux g++

hroest:
hmm on my system it also seems to work fine:

```
$ ctest -R ChromeleonFile_test
Test project /home/hr/openmsall/builds/openms
    Start 163: ChromeleonFile_test
1/1 Test #163: ChromeleonFile_test ..............   Passed    0.17 sec

100% tests passed, 0 tests failed out of 1
```

hroest:
@jpfeuffer @timosachsenberg can you have a look at this? Maby gcc regex is to blame? see https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions

> <regex> was implemented and released in GCC 4.9.0.  In your (older) version of GCC, it is not implemented.

hroest:
@pasqoo can you try out switching to boost regex as I did here: https://github.com/biosustain/OpenMS/pull/92 and merge my branch?

hroest:
yes, I can report that using boost::regex it works: https://travis-ci.org/Roestlab/OpenMS/jobs/326970795 please merge the branch

hroest:
PS: we need to support gcc 4.8 if we want to have manylinux builds

pasqoo:
@hroest Thanks for taking care of it. I'll do some more checks and then announce when the PR is ready for another review.
What about my first comment here? https://github.com/OpenMS/OpenMS/pull/3149#issuecomment-356274657 Do you have any advice?

pasqoo:
The PR is ready again for a review.

src/openms/include/OpenMS/FORMAT/ChromeleonFile.h
```diff
@@ -0,0 +1,67 @@
...
+public:
+    /// Constructor
+    ChromeleonFile();
+    /// Destructor
+    virtual ~ChromeleonFile();
+
+    /**
+      @brief Load the file's data and metadata, and save it into a `MSExperiment`.
+
+      @param[in] filename The filepath to the file containing the informations
```
timosachsenberg: typo: informations -> information (also below)
but maybe better to just write: filename Path to the Chromeleon input file

src/openms/include/OpenMS/FORMAT/ChromeleonFile.h
```diff
@@ -0,0 +1,67 @@
...
+#include <OpenMS/config.h> // OPENMS_DLLAPI
+#include <OpenMS/DATASTRUCTURES/String.h>
+#include <OpenMS/KERNEL/MSExperiment.h>
+#include <boost/regex.hpp>
+#include <fstream>
+
+namespace OpenMS
+{
+  /**
+    @brief Load Chromeleon HPLC text file and save it into a `MSExperiment`.
```
timosachsenberg: maybe add a few lines how the format looks like (e.g. header + 2 lines)

src/openms/include/OpenMS/FORMAT/ChromeleonFile.h
```diff
@@ -0,0 +1,67 @@
...
+  */
+  class OPENMS_DLLAPI ChromeleonFile
+  {
+public:
+    /// Constructor
+    ChromeleonFile();
```
timosachsenberg: = default; would make the empty constructor in the cpp file obsolete

timosachsenberg:
fine from my side

pasqoo:
@timosachsenberg Thanks Timo! The code is up to date.

hroest:
>  As you can see here https://github.com/OpenMS/OpenMS/pull/3149/files#diff-9a5d164ee99681e755b1557e7b4e39d7R76 I went a bit general (from line 78 to 83), using just setMetaValue(), and I wonder if there's a better place for all those info.

I suggest to look at the `ExperimentalSettings` class https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/METADATA/ExperimentalSettings.h#L59 which allows you to store reference to `Instrument`, a Sample description and store information about the original source file which you could parse from `File` and therefore never lose the information which raw file generated the mzML data. Other information you could store in the `HPLC` object. Also look at https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/CV/psi-ms.obo for specific CV terms which does not seem to have UV terms at the moment though. Also you can use the DataProcessing information to store whichs software was used to process the data

pasqoo:
Thanks @hroest . I have already checked those classes, and at the moment this class is using `ExperimentalSettings`.
It's fine from our side to leave the remaining metadata as generic metavalues.
The PR is up to date at this point.

PR #3150, opened 2018-01-09
[FIX] more correct CompNovo oxonium mass

timosachsenberg:


hroest:
looks good :+1: from me, is the failing Win build related?

PR #3159, opened 2018-01-11
[Fix] pyOpenMS glob for boost lib name fixes too strict

jpfeuffer:
It does not match the current contrib builds.
This change should be backwards compatible.


jpfeuffer:
I canceled jenkins build.

jpfeuffer:
Not catched by CI.

hroest:
:+1:

Issue #2266, opened 2016-10-26
Integrate De Novo Peptide Search


timosachsenberg:
In mass spectrometry, de novo peptide sequencing is the method in which a peptide amino acid sequence is determined from tandem mass spectra.

Knowing the amino acid sequence of peptides from a protein digest is essential to study the biological function of the protein. Today, analysis by a tandem mass spectrometer is the most common method to solve the sequencing of peptides. Generally, there are three approaches: database search, spectral library search and de novo sequencing. Database search and spectral library search finds a match with a known peptide sequence or previously measured spectrum. De novo sequencing is an assignment of fragment ions from a mass spectrum without knowledge on the database.

In addition, one could think about wrapping existing de novo engines like novor.

timosachsenberg:
Publication:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4604512/
Binaries: (free for academic use)
https://www.rapidnovor.com/novor-cmd/

hroest:
Alternatively we could use deepnovo:

- Github: https://github.com/nh2tran/DeepNovo
- Paper: http://www.pnas.org/content/114/31/8247.abstract

it seems they perform equal/better than rapidnovor and substantially better than PEAKS

Leon-Bichmann:
I had a look at Novor to integrate it during the EuBic2018 meeting:

- I guess the advantage of Novor over DeepNovo, is its super fast runtime ( 300 spectra / second)
- However its accuracy is worse than DeepNovo
- The executable is a jar file
- A handful of parameters can be passed in a parameter.txt file:
  - enzyme = Trypsin   (**Disadvantage: Enzyme only trypsin is supported currently**)
  - fragmentation = CID   (CID, HCD)
  - massAnalyzer = Trap   (Trap, TOF, FT)
  - fragmentIonErrorTol = 0.5Da
  - precursorErrorTol = 15ppm
  - variableModifications = Oxidation (M)   (comma-delimited list of IDs can be used here)
  - fixedModifications = Carbamidomethyl (C) (See http://rapidnovor.com/wiki/Built-in_PTMs)
  - forbiddenResidues = I,U
  - novorFile = external.novor (a customized ML param set can be used trained on special data)

timosachsenberg:
DeepNovor seems also allow non commercial use only. According to their github it seems also that they did not reach a state to be useful / robust for general data.

hroest:
[...]

Issue #2265, opened 2016-10-26
Implement De Novo Peptide Search


timosachsenberg:
In mass spectrometry, de novo peptide sequencing is the method in which a peptide amino acid sequence is determined from tandem mass spectra.

Knowing the amino acid sequence of peptides from a protein digest is essential to study the biological function of the protein. Today, analysis by a tandem mass spectrometer is the most common method to solve the sequencing of peptides. Generally, there are three approaches: database search, spectral library search and de novo sequencing. Database search and spectral library search finds a match with a known peptide sequence or previously measured spectrum. De novo sequencing is an assignment of fragment ions from a mass spectrum without knowledge on the database. 

In OpenMS there are some implementations of de novo search engines that might be worth to reevaluate in a student project.

lars20070:
Anyone currently working on de-novo sequencing? Evaluation of our existing algorithms and comparison to other external tools would be interesting for us. Perhaps we could collaborate.

timosachsenberg:
No currently we have nobody assigned to this project. Once an adapter is available We could help with integrating the executables into the packages and knime  builds.

lars20070:
Ok. Thanks @timosachsenberg.
More IDs than `PEAKS` and 10x faster very sounds interesting.

hroest:
seems to make sense to create an adaptor but it also does not seem to be open-source, see https://www.rapidnovor.com/download/ so we cant really distribute it as far as I can tell

hroest:
@lars20070 @timosachsenberg lets move discussion of writing adapters to existing de novo search engines over here: #2266

PR #3160, opened 2018-01-12
[FEATURE] NovorAdapter 

oliveralka:
Adapter for Novor (de novo sequencing tool).

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4604512/

Had to introduce some changes in the MascotGenericFile.

edit:
#2266
#2265


timosachsenberg:
set filename of mzML by:
'''
    protein_identification.setPrimaryMSRunPath(the filename of the mzML);
'''

oliveralka:
Maybe put in the documentation, that novor.jar is not distributed with OpenMS and has to be downloaded (link to website).

src/openms/source/FORMAT/MascotGenericFile.cpp
```diff
@@ -359,7 +361,8 @@ namespace OpenMS
         bool skip_spectrum_charges(param_.getValue("skip_spectrum_charges").toBool());
         if (!skip_spectrum_charges)
         {
-          os << "CHARGE=" << charge << "\n";
+          String cs = charge < 0 ? "-" : "+";
+          os << "CHARGE=" << charge << cs << "\n";
```
hroest: does this break any of the other consumers we have for mgf, specifically the mascot server?

src/utils/NovorAdapter.cpp
```diff
@@ -0,0 +1,389 @@
...
+    This tool can be used for de novo sequencing of peptides from MS/MS data.
+
+    Please use MS2-Spectra only. If filtering is needed please use the @ref TOPP_FileFilter.
+
+
+    <B>The command line parameters of this tool are:</B>
+    @verbinclude UTILS_NovorAdapter.cli
+    <B>INI file documentation of this tool:</B>
+    @htmlinclude UTILS_NovorAdapter.html
+*/
```
hroest: can you improve the documentation here, e.g. something similar to the XTandem / Comet adapter: https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/TOPP_CometAdapter.html ?

src/utils/NovorAdapter.cpp
```diff
@@ -0,0 +1,389 @@
...
+    registerStringList_("fixed_modifications", "<mods>", vector<String>(), "Fixed modifications", false);
+    setValidStrings_("fixed_modifications", ListUtils::create<String>("Acetyl (K),Acetyl (N-term),Amidated (C-term),Ammonia-loss (N-term C),Biotin (K),Biotin (N-term),Carbamidomethyl (C),Carbamyl (K),Carbamyl (N-term),Carboxymethyl (C),Deamidated (NQ),Dehydrated (N-term C),Dioxidation (M),Methyl (C-term),Methyl (DE),Oxidation (M),Oxidation (HW),Phospho (ST),Phospho (Y),Pyro-carbamidomethyl (N-term C),Pyro-Glu (E),Pyro-Glu (Q),Sodium (C-term),Sodium (DE),Sulfo (STY),Trimethyl (RK)"));
+   // forbidden residues
+   registerStringList_("forbiddenResidues", "<mods>", vector<String>(), "Forbidden Resiudes", false);
+   setValidStrings_("forbiddenResidues", ListUtils::create<String>("I,U"));
```
hroest: what about options not available through the adapter? is there a config file you can pass to the tool ? or are these all the available options?

src/utils/NovorAdapter.cpp
```diff
@@ -0,0 +1,389 @@
...
+    {
+      if (debug_level_ == 1) 
+      {
+        writeDebug_("Deleting temporary directory '" + tmp_dir + "'. Set debug level to 2 or higher to keep it.", 1);
+      }
+      File::removeDirRecursively(tmp_dir);
+    }
+  }
+
+  void createParamFile_(ostream& os)
```
hroest: I think it may make sense to allow advanced users to just pass a param file...we also do this for other adapters

src/utils/NovorAdapter.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+       PeptideHit ph;
+       ph.setCharge(sl[4].toInt());
+       ph.setScore(sl[8].toDouble());
+
+       // replace PTM name (see http://wiki.rapidnovor.com/wiki/Built-in_PTMs)
+       String sequence = sl[9];
+       sequence.substitute("(Cam)", "(Carbamidomethyl)");
+       sequence.substitute("(O)","(Oxidation)");
+       sequence.substitute("(PyroCam)", "(Pyro-carbamidomethyl)");
```
hroest: the list on the wiki is substantially longer than the ones here, will this be a problem if there are mods in the output file that are not recognized here?

hroest:
looks really good, thanks for the effort. This could be great to have in OpenMS. Can you comment on the impact on the mascot file, can you maybe build with `-DWITH_MASCOT_TEST=ON` and see whether the mascot server still works?

oliveralka:
@hroest: One test is failing, but it seems related to the whitelisting and not to the changes made to MascotGenericFile.

```
    Start 1518: TOPP_MascotAdapter_1
1/6 Test #1518: TOPP_MascotAdapter_1 ..............   Passed    0.07 sec
    Start 1519: TOPP_MascotAdapter_1_out1
2/6 Test #1519: TOPP_MascotAdapter_1_out1 .........   Passed    0.04 sec
    Start 1520: TOPP_MascotAdapter_2
3/6 Test #1520: TOPP_MascotAdapter_2 ..............   Passed    0.06 sec
    Start 1521: TOPP_MascotAdapter_2_out1
4/6 Test #1521: TOPP_MascotAdapter_2_out1 .........   Passed    0.03 sec
    Start 1968: TOPP_MascotAdapterOnline_1
5/6 Test #1968: TOPP_MascotAdapterOnline_1 ........   Passed    7.37 sec
    Start 1969: TOPP_MascotAdapterOnline_1_out1
6/6 Test #1969: TOPP_MascotAdapterOnline_1_out1 ...***Failed    0.03 sec
```
```
bash-4.2$ diff /home/**/software/openms/source/OpenMS/openms-build/src/tests/topp/MascotAdapterOnline_1_out1.tmp /home/**/software/openms/source/OpenMS/src/tests/topp/THIRDPARTY/MascotAdapterOnline_1_out.idXML
4c4
< 	<SearchParameters id="SP_0" db="SwissProt" db_version="SwissProt_2017_12.fasta" taxonomy="All entries" mass_type="monoisotopic" charges="" enzyme="trypsin" missed_cleavages="1" precursor_peak_tolerance="3" precursor_peak_tolerance_ppm="false" peak_mass_tolerance="0.3" peak_mass_tolerance_ppm="false" >
---
> 	<SearchParameters id="SP_0" db="SwissProt" db_version="SwissProt_2017_11.fasta" taxonomy="All entries" mass_type="monoisotopic" charges="" enzyme="trypsin" missed_cleavages="1" precursor_peak_tolerance="3" precursor_peak_tolerance_ppm="false" peak_mass_tolerance="0.3" peak_mass_tolerance_ppm="false" >
6c6
< 	<IdentificationRun date="2018-01-12T18:55:44" search_engine="Mascot" search_engine_version="2.6.1" search_parameters_ref="SP_0" >
---
> 	<IdentificationRun date="2017-12-20T22:34:20" search_engine="Mascot" search_engine_version="2.6.0.242" search_parameters_ref="SP_0" >
20c20
< 			<UserParam type="string" name="SearchNumber" value="TmAlbswE"/>
---
> 			<UserParam type="string" name="SearchNumber" value="TmAmbHOE"/>
```

src/utils/NovorAdapter.cpp
oliveralka: l think these are the only ones which have a different "short name" compared to unimod, so there should not be any problems with the others. If the list of Novor-modifications is extended we may have to check again and update. Will add a comment in the code - or do you have another suggestion in that regard.

src/utils/NovorAdapter.cpp
oliveralka: These are all the available options, but there seem to be some external options in the standard param.txt of novor. I will add the option to use the param.txt directly if needed (as suggested from you below - optional).

src/openms/source/FORMAT/MascotGenericFile.cpp
oliveralka: see below - does not seem to be the case.

hroest:
yes, looks like they changed the database version in the meantime ... should be fixed in #3165

src/utils/NovorAdapter.cpp
timosachsenberg: One problem with that is that we need to parse the param file manually to properly annotate the search engine settings. 
If the param file format changes, this easily leads to difficult / undetected errors. Note that creating the param file for a wrong search engine version is usually more easily detected if it is done from tools parameters.

hroest: I see the problem. I agree parsing the param file is not what we want to do. Still it is annoying if there is some parameter you would like to use and simply cant because its not supported by OpenMS. Maybe an alternative would be to 

* pass a param file and in that case ignore any options passed to OpenMS?
* pass a file whose content simply gets appended to the param file written by OpenMS

in either case I agree we dont want to parse the file. But I am also fine to leave it as is

hroest:
great, :+1: from me

oliveralka:
Added the last missing parameter (novorFile), which is a custom parameter file in .novor format, which can be used for advances users to set algorithm parameters. I did not add the param.txt file as input, since I would have to parse the whole output file in a different way. If the integration of the param.txt is wished I can do that as well, but I would prefer not to. (This feature was not tested since I don't have a custom .novor file)

src/utils/NovorAdapter.cpp
```diff
@@ -173,12 +179,19 @@ class TOPPNovorAdapter :
...
+     << "fragmentIonErrorTol = " << getDoubleOption_("fragment_mass_tolerance") << "Da" << "\n"
+     << "precursorErrorTol = " << getDoubleOption_("precursor_mass_tolerance") << getStringOption_("precursor_error_units") << "\n"
      << "variableModifications = " << variable_mod << "\n"
      << "fixedModifications = "    << fixed_mod << "\n"
      << "forbiddenResidues = " << forbidden_res << "\n";
   
+   // novorFile for custom alogrithm parameters of nova
+   String cparamfile = getStringOption_("novorFile");
+   ifstream cpfile(cparamfile);
+     if (!cpfile)
```
timosachsenberg: looks like different indentation

timosachsenberg:
thanks! nice addition. we need to add it to the changelog at some point

lars20070:
@oliveralka Is [this](http:// www.rapidnovor.org/novor) the correct `Novo` version tested with the adapter. If yes, I will add it to the [`THIRDPARTY`](https://github.com/OpenMS/THIRDPARTY/tree/master/All) tools.

timosachsenberg:
@lars20070 we are not allowed to distribute Novor so please don't add it.

lars20070:
Ok, I forgot. Maybe just a `README` with version (1.05) and download link.

oliveralka:
@lars20070 This should be the correction version. 
For integration and testing the version v1.06.0634 (stable) was used.
A link to the homepage can be found in the NovorAdapter tool description (https://www.rapidnovor.com).
We change it to "https://www.rapidnovor.com/download/" if that is more straightforward.

jpfeuffer:
You can add it to the general README of the THIRDPARTY repository but please do not add a novoR folder, as it might break our merging procedure in the nightlies.

lars20070:
Ok. No changes needed. The version is well documented in Doxygen.
https://github.com/oliveralka/OpenMS/blob/4978a4049ab4536a7a1223ad8a493e8d6016a1e5/src/utils/NovorAdapter.cpp#L80

The `README` in the download from https://www.rapidnovor.com/download/ states version `v1.05` and not `v1.06.0634`. Hope this `Novor` version is the correct one for the adapter.

PR #3164, opened 2018-01-12
[FIX] pyopenms enum for win

jpfeuffer:
This fixes it on Windows. Does it make sense?

### DIFF
--- a/src/pyOpenMS/pxds/TargetedExperimentHelper.pxd
+++ b/src/pyOpenMS/pxds/TargetedExperimentHelper.pxd
@@ -9,6 +9,8 @@ from CVTermList cimport *
 cdef extern from "<OpenMS/ANALYSIS/TARGETED/TargetedExperimentHelper.h>" namespace "OpenMS::TargetedExperimentHelper::RetentionTime::RTUnit":
 
     ctypedef enum RTUnit "OpenMS::TargetedExperimentHelper::RetentionTime::RTUnit":
+            # wrap-attach:
+            #    RetentionTime
             SECOND,        # RT stored in seconds
             MINUTE,        # RT stored in minutes
             UNKNOWN,       # no stored annotation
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Seems like you have to access it with pyopenms.RTUnit now (without RetentionTime). Even if I add back the wrap_attach.
Don't know if this is wanted.
I don't get why it compiles on Linux and Mac, though.

hroest:
> Seems like you have to access it with pyopenms.RTUnit now (without RetentionTime). Even if I add back the wrap_attach.
> Don't know if this is wanted.

its not that bad, its just an issue of polluting the global pyopenms namespace with these small enums that should be in their own namespace / attached to a class

@jpfeuffer can you check whether adding the "wrap-attach" still works on Windows?

src/pyOpenMS/pxds/TargetedExperimentHelper.pxd
```diff
@@ -6,26 +6,28 @@ from Map cimport *
...
-        PREDICTED,        # predicted by referenced software
-        HPINS,            # H-PINS "The de facto standard providing the retention times"
-        IRT,              # iRT retention time standard
-        UNKNOWN,          # no stored annotation
-        SIZE_OF_RTTYPE
+cdef extern from "<OpenMS/ANALYSIS/TARGETED/TargetedExperimentHelper.h>" namespace "OpenMS::TargetedExperimentHelper::RetentionTime::RTUnit":
+
+    ctypedef enum RTUnit "OpenMS::TargetedExperimentHelper::RetentionTime::RTUnit":
+            # wrap-attach:
+            #    RetentionTime
```
hroest: with this all tests pass again ...

jpfeuffer:
Yes! I tried again today and with the wrap-attach it also works on windows. There must have been some old code cached somewhere. Will merge to see the effect on the nightlies.

PR #3166, opened 2018-01-15
[FIX] AbsoluteQuantitationMethod::getLLOD() getter correctly returns llod_

pasqoo:
I'm working on a class named _AbsoluteQuantitationMethodFile_ and I found this bug.
For now I'm only correcting this line, since I guess it is high priority.
I will also add tests that would catch this kind of bugs once I'm done with my other classes.


hroest:
looks, but definitely you should add tests to the corresponding class https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/AbsoluteQuantitationMethod_test.cpp

pasqoo:
@hroest Since I had to add the tests, I also went for a general cleanup of the class. The PR is up to date and therefore ready for a review.

PR #3167, opened 2018-01-16
[FEATURE] Try submodules for travis.

jpfeuffer:


hroest:
@jpfeuffer please also note that we cache the contrib directory at the moment which speeds up the build significantly - we need to also look at how this will affect build times on travis

hroest:
seems like we sometimes get the following error

> fatal: destination path 'contrib' already exists and is not an empty directory.

jpfeuffer:
Yes I will keep the caching. It's just about where to get the sources from. Would be cool if the cache is automatically cleared if sth changes in the contrib. Travis has serious downtime right now. I fixed the folder problem already but travis doesnt start the new builds.

jpfeuffer:
According to travis docu, travis will actually create a cache for every PR. Is this the case for us, too?
We could think about adding thirdparty to the cache too.

jpfeuffer:
@timosachsenberg can you check if that solves your EIGEN problem?

timosachsenberg:
I manually cleared the cache (at least I hope that's what I did ;) by manually selecting the ci build. In its option menu one can access the caches.)

PR #3169, opened 2018-01-17
[FEATURE] AbsoluteQuantitationStandards::mapComponentsToConcentrations()

pasqoo:
Added this method to the class.

The class implements two structs: `runConcentration` and `featureConcentration`.

Given in input a vector of `runConcentration`s and a vector of `FeatureMap`s, the method creates a `std::map` from component names to `featureConcentration`s.


src/openms/source/METADATA/AbsoluteQuantitationStandards.cpp
```diff
@@ -28,34 +28,77 @@
...
+    for (const AbsoluteQuantitationStandards::runConcentration& run : run_concentrations)
     {
+      if (run.sample_name == "" || run.component_name == "")
+      {
+        continue;
+      }
+      for (const FeatureMap& fmap : feature_maps) // not all elements are necessarily processed (break; is present inside the loop)
+      {
+        AbsoluteQuantitationStandards::featureConcentration fc;
+        if (!fmap.metaValueExists("sample_name") || fmap.getMetaValue("sample_name") != run.sample_name) // if the FeatureMap doesn't have a sample_name, or if it is not the one we're looking for: skip.
```
dmccloskey: @timosachsenberg and @hroest we are currently using a new `metaValue` in `Feature` called "sample_name" to map to different `FeatureMaps`.  Is there a standardized `Feature` attribute that can be used for this?  Alternatively, is there another convention that you or others have used to map to different `FeatureMaps`?

pasqoo:
Not sure about what makes those tests fail

timosachsenberg:
Do you want to store different references to samples in a single featureXML? Or only one sample id per featureXML?

dmccloskey:
It would be 1 sample per featureXML

timosachsenberg:
As featureXML map to a single mzML file, we usually reference the MS run (mzML file). 
If this also works for you I would prefer that over an extra meta value.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/KERNEL/FeatureMap.h#L250

pasqoo:
@timosachsenberg @hroest The PR is up to date. Not sure about what makes travis-ci fail, but I can see that the windows build fails because of CruxAdapter.
The apple build has many fails on TOPP tools.

pasqoo:
Update: I need to change something that I did not notice before, please hold the reviews. I will let you know once I'm done.

pasqoo:
The PR is ready for another review @timosachsenberg @hroest

hroest:
@pasqoo currently complex templates like `libcpp_map[String, libcpp_vector[AQS_featureConcentration]]` are not supported by pyOpenMS automatic wrapping. Either you change the type or you have to  wrap it manually as for example here: https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/addons/AbsoluteQuantitation.pyx

pasqoo:
@hroest We decided to implement an additional method named `getComponentFeatureConcentrations()`. This method works in pyOpenMS because it takes a component_name in input and only outputs a vector instead of a complex map. It internally uses the previous method.
Therefore I won't expose the previous method (which stays there for c++ users) to pyOpenMS.

hroest:
>  Therefore I won't expose the previous method (which stays there for c++ users) to pyOpenMS.

sounds good, so it just wont be part of the public API

pasqoo:
Not on the python side. But it is part of the public api for c++ users (did not move the previous method to protected or private).

PR #3170, opened 2018-01-17
[FEATURE] AbsoluteQuantitationMethodFile::store()

pasqoo:
Implementation proposal.

Given a filename in input and a vector of `AbsoluteQuantitationMethod`, store the data to a .csv file.


src/tests/class_tests/openms/source/AbsoluteQuantitationMethodFile_test.cpp
```diff
@@ -261,6 +261,34 @@ START_SECTION((void load(const String & filename, std::vector<AbsoluteQuantitati
 
 END_SECTION
 
+START_SECTION(void store(const String & filename, const std::vector<AbsoluteQuantitationMethod> & aqm_list) const)
+  AbsoluteQuantitationMethodFile aqmf;
+  vector<AbsoluteQuantitationMethod> aqm_list1, aqm_list2;
+  aqmf.load(OPENMS_GET_TEST_DATA_PATH("AbsoluteQuantitationMethodFile_1.csv"), aqm_list1);
+  aqmf.store(OPENMS_GET_TEST_DATA_PATH("AbsoluteQuantitationMethodFile_2.csv"), aqm_list1);
+  aqmf.load(OPENMS_GET_TEST_DATA_PATH("AbsoluteQuantitationMethodFile_1.csv"), aqm_list2);
```
hroest: shouldnt this load ` "AbsoluteQuantitationMethodFile_2.csv"` in order to really test the store method?

src/tests/class_tests/openms/source/AbsoluteQuantitationMethodFile_test.cpp
```diff
@@ -261,6 +261,34 @@ START_SECTION((void load(const String & filename, std::vector<AbsoluteQuantitati
...
+  TEST_EQUAL(aqm_list1.size(), aqm_list2.size())
+  for (Size i = 0; i < aqm_list1.size(); ++i)
```
hroest: if `AbsoluteQuantitationMethod` had an `operator==` then you could just test for equality. But this is fine too...

src/tests/class_tests/openms/source/AbsoluteQuantitationMethodFile_test.cpp
pasqoo: Yeah, of course it should! Thanks for catching this. During the tests I kept looking at the files outputs and they were identical except for the end of lines (no '\r' in the output files). Didn't notice this bug.

hroest:
looks good (only failed tests are crux adapter)

src/tests/class_tests/openms/source/AbsoluteQuantitationMethodFile_test.cpp
pasqoo: I will implement this in the coming days

PR #3175, opened 2018-01-19
[FIX] avoid 2x memory usage in FileInfo

cbielow:
I have no idea how that got past CR.
but it effectively doubles memory usage when reporting ranges. Rather ugly, especially for large files.

### DIFF
--- a/src/topp/FileInfo.cpp
+++ b/src/topp/FileInfo.cpp
@@ -369,11 +369,11 @@ protected:
       if (ifile.getParsingSuccess())
       {
         // Validate that we can access each single spectrum and chromatogram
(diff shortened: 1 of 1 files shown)


src/topp/FileInfo.cpp
```diff
@@ -167,7 +167,7 @@ class TOPPFileInfo : public TOPPBase
   }
 
   template <class Map>
```
timosachsenberg: do we need the template at all?

cbielow: we do. Its used for FeatureMaps as well.

timosachsenberg:
look good

PR #3176, opened 2018-01-19
TOPPView: fix initial zoom in identification view (stored annotations)

hendrikweisser:
Currently, when IDs with stored peak/fragment annotations are opened in TOPPView's identification view, MS2 spectra are shown zoomed out to the "maximal m/z data range" (of any MS2 spectrum?), although there may not be any peaks covering the full range. With this change, the spectra are initially zoomed in to the range where there are actually peaks.

### DIFF
--- a/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
+++ b/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
@@ -960,6 +960,15 @@ namespace OpenMS
       tv_->getActive1DWidget()->canvas()->activateLayer(current_spectrum_layer_index);
       tv_->getActive1DWidget()->canvas()->getCurrentLayer().setCurrentSpectrumIndex(current_spectrum_index);
 
+      // zoom visible area to real data range:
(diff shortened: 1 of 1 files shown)


hendrikweisser:
Screenshots:
Before:
toppview-idview-before

After:
toppview-idview-after

enetz:
is it possible to zoom in to a scale you want and keep that scale while switching between IDs?
sometimes people want to compare peaks in a specific region between several spectra.
could we add an "auto-zoom" checkbox, with the behaviour you described when it is checked and just not change the zoom between IDs when it is not? or would it be more complicated?

enetz:
otherwise this is a nice improvement

hendrikweisser:
> is it possible to zoom in to a scale you want and keep that scale while switching between IDs?

No. The scale is reset if you switch between spectra - both before and after this proposed change.

I agree that the functionality you describe would be nice to have, but it goes beyond what I'm prepared to do at the moment. I think the changes here are already an improvement over the current state, with no apparent downside.

timosachsenberg:
https://github.com/OpenMS/OpenMS/issues/3178

timosachsenberg:
I added a feature request for the auto-zoom checkbox. From my side, this PR is a nice addition that can be merged.

PR #3180, opened 2018-01-19
[FIX] #3179 

### LINKED ISSUE #3179: TOPPAS crashes when loading old workflow
when loading a TOPPAS workflow created with an older version of OpenMS
(or when transferring to another PC which has fewer GenericWrapper scripts/types installed),
TOPPAS might crash upon loading the workflow, because it tries to access in internal structure which has not been initialized yet.

If you run the script with ExecutePipeline, it will show a proper error message.

Workaround: 
 a) re-install the script 
 b) or delete the nodes from the workflow.

Real Solution: PR coming up.

cbielow:
scene() will return nullptr, since the TOPPASToolVertex is not registered as an object the scene yet (we're still constructing it.

Fixes #3179.


cbielow:
test errors are unrelated (some MacOS ABI-builder....).
Merge?

lars20070:
Fix works. Thanks @cbielow @timosachsenberg.

Issue #3172, opened 2018-01-18
Bug in TransitionTSVFile not reading the protein name


bretttully:
Looking at lines 373 and 374 in comparison with other lines around them -- should they have a `!extractName` (note the `!`)? Adding this in means that the protein name will be read when it is called `ProteinId`. 

https://github.com/OpenMS/OpenMS/blob/6edaa6492cd6d50d5e6e677e0df576d641bdba08/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp#L373

cc @hroest

bretttully:
This also means that when running the `TargetedFileConvertor` converting tsv to pqp the ProteinId is not copied across

grosenberger:
Thank you, indeed this is a bug. This should be fixed in the PR linked below.

PR #3181, opened 2018-01-19
[FIX] TargetedTSVFile: TSV import of ProteinId

grosenberger:
This PR fixes https://github.com/OpenMS/OpenMS/issues/3172.

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
@@ -370,8 +370,8 @@ namespace OpenMS
       extractName<int>(mytransition.fragment_modification, "FragmentModification", tmp_line, header_dict);
 
       //// Proteomics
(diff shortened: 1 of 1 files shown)


src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
```diff
@@ -370,8 +370,8 @@ namespace OpenMS
       extractName<int>(mytransition.fragment_modification, "FragmentModification", tmp_line, header_dict);
 
       //// Proteomics
-      extractName(mytransition.ProteinName, "ProteinName", tmp_line, header_dict) &&
-      extractName(mytransition.ProteinName, "ProteinId", tmp_line, header_dict); // Spectronaut
+      !extractName(mytransition.ProteinName, "ProteinName", tmp_line, header_dict) &&
+      !extractName(mytransition.ProteinName, "ProteinId", tmp_line, header_dict); // Spectronaut
```
timosachsenberg: I find using the short circuit logic a bit convoluted. Especially as the second ! has no meaning at all.
I would have preferred a simple if

PR #3182, opened 2018-01-20
Fix/tsv2traml

### LINKED ISSUE #3172: Bug in TransitionTSVFile not reading the protein name

hroest:
fixes #3172 
amends #3181


hroest:
also tests TSV import of other fields, including RelativeFragmentIntensity, ProteinId etc.

Issue #3188, opened 2018-01-22
Percolator reports all peptides with 1 score in OpenMS 2.3 with labeled samples


algom:
I have been trying to integrate Percolator to my workflow in OpenMS 2.3 (TOPPAS on Windows) but so far have not succeed. This issue might be related to https://github.com/OpenMS/OpenMS/issues/3071.

My workflow is simple and is just an adaptation from a post of Mathew The on the Gitter OpenMS/Knime chat, but has not been succesful.

MSGFPlusAdapter —> PeptideIndexer —> PercolatorAdapter

I can reproduce Mathew The results, with the yeast test file, however I have no luck with my test file which is a Formaldehyde labeled sample (K and n-term labeled) measured in a Qexactive. The changes to the workflow thus are only adding the fixed modification (CAM, K and n-term dimethylation).

I have also tried to do the search with a database without decoys and let the MSGFPlusAdapter add the decoys (and of course removing the peptide indexer node). I also have done it with the XtandemAdapter, with similar results. 

Percolator runs through, but using MSGF in the training portion reports 0 PSMs at q < 0.01, even though before training recognizes both targets and decoys, and finally reports "0 suitable peptides found". If I use the mzid output of msgfplus as input instead, it does recognize PSMs below q 0.01 at the training but it also ends with 0 suitable peptides. Looking at the produced idxml file, the first part of the file, protein section all seem to have a score of 0, whereas at the peptide level all get a score of 1. Running it with xTandem, does reports PSMs below q 0.01 at training but ends with the same result (All peptides with score 1). 

If I take the MSGFplusAdapter mzid output and do msgf2pin and percolator on the command line in the same PC, percolator runs fine, reports both peptides at training (at all iterations) and finds suitable peptides, with probabilities from 0-1 in the end.

Setting debugging at 5 in PercolatorAdapter to keep the intermediate files I get the pin input file (which I guess comes from the idxml) and what I guess the target and decoy output. Running percolator on the command line with that pin file (Of both xTandem and MSGFplus) also runs through, although it does not find peptides in all iterations at training with msgf input thus resets the vector, however in the end finds peptides below q 0.01.

[...]

PR #3189, opened 2018-01-22
Update PercolatorAdapter.cpp

timosachsenberg:
Attempt to fix https://github.com/OpenMS/OpenMS/issues/3188


jpfeuffer:
What about the double point? Although this might not be a problem for matching IDs it seems quite wrong.

timosachsenberg:
Was Just a quick check using the web interface

Issue #3191, opened 2018-01-23
PepXML Reader can't read some Comet enzyme names


Matthias313:
Hi,
I built a very simple workflow 
CometAdapter->PeptideIndexer->FalseDiscoveryRate->MzTabExporter/FileInfo

in the default_params_file which is given to the CometAdapter the enzyme is by default Trypsin (Code "1").
I specified another enzyme in the adapter (used glutamyl endopeptidase) and started the pipeline. Checking the Output the search run with the enzymatic specificity of trypsin, not glutamyl endopeptidase. 

To check if this could be fixed i changed the setting in the default_params_file to GluC = glutamyl endopeptidase (Code "8"). 
This change allowed the search for GluC specific Peptides. 

Is it possible to fix the Adapter so that the enzyme which the user defines in the tool is used rather than the enzyme specified in the default_params_file?

[Comet.txt](https://github.com/OpenMS/OpenMS/files/1656462/Comet.txt)

The toppas workflow saved as txt file

[comet.params.txt](https://github.com/OpenMS/OpenMS/files/1656457/comet.params.txt)

the default_params_file

jpfeuffer:
Why do you need the default params file? As specified in the docu, when a params file is given, it will use only that.

Matthias313:
load PepXMLFile
 95%Error: Unexpected internal error (the element 'Enzyme name cannot be found. '' could not be found)
Error occurred in line 67 of file C:\jenkins\ws\openms_release_packaging\c2e226b2\source\src\openms\source\CHEMISTRY\EnzymesDB.cpp (in function: const class OpenMS::Enzyme *__cdecl OpenMS::EnzymesDB::getEnzyme(const class OpenMS::String &) const) !

==============================================================================
17:24:12 ERROR: CometAdapter failed!

When using the CometAdapter without a default_params file i get this error. 
In the Adapter one only can select glutamyl endopeptidase whereas in the default in is written GluC with Code "8".

jpfeuffer:
Hmm I can't reproduce that error. Can you paste your command or ini file? I want to see how you specify the enzyme. For all adapters we use the same OpenMS nomenclature and translate to the specific nomenclature of the tool. Here "glutamyl endopeptidase" = "Glu_C" = "8".
This should be ok.

jpfeuffer:
I have found one bug though.

Matthias313:
I changed the file type to txt because .ini is not supported to be uploaded.
[...]

PR #3194, opened 2018-01-23
[Fix] Add synonymes to some comet enzymes so PepXML reader recognizes them

jpfeuffer:
Also fixes a wrong annotation of unspecific cleavage for the crux enzymes.
Also makes the lookup for enzymes supported by OMSSA and Comet a bit more robust by making their IDs ints and introducing a specific "-1" for unsupported enzymes.
Removed hardcoded strings in pepXML reader.


jpfeuffer:
Fixes #3191

src/openms/source/CHEMISTRY/DigestionEnzymeProtein.cpp
```diff
@@ -46,10 +46,10 @@ namespace OpenMS
     c_term_gain_(""),
     psi_id_(""),
     xtandem_id_(""),
-    comet_id_(),
-    crux_id_(),
+    comet_id_(-1),
```
timosachsenberg: not that important but in the future, we should transform these into =default constructors and define the default initialization value of the variables in the header.  Less error prone...

jpfeuffer:
hmm will fix the test tomorrow.

share/OpenMS/CHEMISTRY/Enzymes.xml
```diff
@@ -318,9 +328,14 @@
       <ITEM name="CTermGain" value="OH" type="string" />
       <ITEM name="PSIID" value="MS:1001956" type="string" />
       <ITEM name="XTandemID" value="[X]|[X]" type="string" />
+      <ITEM name="CruxID" value="no-enzyme" type="string" />
```
hroest: hmm, do you think this is what `no-enzyme ([X]|[X])` means? I thought that would be `([]|[])` if it cuts everywhere but we can ask

jpfeuffer:
So the latest changes include a major rewrite of the CometAdapter parameters. Some had wrong types. I also added more documentation and some I think important parameters.
Also terminal modifications were written wrongly to the params file.
While searching for errors in the build log, I catched a warning in MSFraggerAdapter and found another small bug.
Adapted the tests accordingly.

src/topp/CometAdapter.cpp
```diff
@@ -123,87 +123,97 @@ class TOPPCometAdapter :
...
     registerInputFile_("default_params_file", "<file>", "", "Default Comet params file. All parameters of this take precedence. A template file can be generated using comet.exe -p", false, false, ListUtils::create<String>("skipexists"));
     setValidFormats_("default_params_file", ListUtils::create<String>("txt"));
-    //registerIntOption_("threads", "<num>", 1, "number of threads", false, true);
 
     //Masses
-    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 10.0, "Precursor monoisotopic mass tolerance (Comet parameter: peptide_mass_tolerance)", false, false);
-    registerStringOption_("precursor_error_units", "<choice>", "ppm", "peptide_mass_units 0=amu, 1=mmu, 2=ppm", false, false);
+    registerDoubleOption_("precursor_mass_tolerance", "<tolerance>", 10.0, "Precursor monoisotopic mass tolerance (Comet parameter: peptide_mass_tolerance).  See also precursor_error_units to set the unit.", false, false);
+    registerStringOption_("precursor_error_units", "<choice>", "ppm", "Unit of precursor monoisotopic mass tolerance for parameter precursor_mass_tolerance (Comet parameter: peptide_mass_units)", false, false);
```
timosachsenberg: is "..._units" correct? would _unit make more sense

jpfeuffer:
Yes will change. I Have another commit coming.

jpfeuffer:
If that runs through. Ready to merge.

timosachsenberg:
+1 if passes

jpfeuffer:
Windows unrelated and macOS the unmatching boost problems.

Matthias313:
To test if the the CometAdapter now runs through, I will need the developer version (nightly version) of OMS?

jpfeuffer:
Yes exactly. Do you build it yourself or do you need an installer? Do you use the KNIME plugin?

Matthias313:
@lars20070 
Lars updated OpenMS on our Server (2.3). 
Is there the option to only update some tools (e.g. the CometAdapter)???
Currently, I am using TOPPAS, since I am starting the runs via a batch skript which is not KNIME compatible and I don´t have a lot of experience using KNIME.

Matthias313:
Alright, I will wait for Lars.
Thanks for the Info and the fast fix of the bug @jpfeuffer !

Issue #3193, opened 2018-01-23
FeatureFinder(Centroided) does not complain on 0 intensity peaks but just performs poorly


jpfeuffer:
See discussion on Gitter https://gitter.im/OpenMS/OpenMS?at=5a65f971ae53c159031ce9d8

jpfeuffer:
In general after looking at the code I think FFC needs to be improved to handle edge cases better or throw better exceptions.

jpfeuffer:
Also an issue seems to be that TOPPView does not load features with negative intensities. Maybe it should give warnings instead. Or the feature finders should not print them out. Although with the absolute value it might be solved. Affected features look good by the way. Ones that were negative before are in the range of surrounding features with similar intensity profiles.

IceFreez3r:
Old issue, but is probably now fixed: #4652

PR #3195, opened 2018-01-23
[FIX] Negative sigma when fitting asymm. EGH through optimization

jpfeuffer:
Related to #3193

### DIFF
--- a/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EGHTraceFitter.cpp
+++ b/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EGHTraceFitter.cpp
@@ -279,7 +279,8 @@ namespace OpenMS
   {
     // equation 21 from Lan & Jorgenson paper:
     double abs_tau = fabs(tau_);
-    double phi = atan(abs_tau / sigma_);
+    double abs_sigma = fabs(sigma_);
+    double phi = atan(abs_tau / abs_sigma);
     double epsilon = EPSILON_COEFS_[0];
     double phi_pow = phi;
     for (Size i = 1; i < 7; ++i)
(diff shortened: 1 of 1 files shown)


src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EGHTraceFitter.cpp
```diff
@@ -279,7 +279,8 @@ namespace OpenMS
   {
     // equation 21 from Lan & Jorgenson paper:
     double abs_tau = fabs(tau_);
-    double phi = atan(abs_tau / sigma_);
+    double abs_sigma = fabs(sigma_);
+    double phi = atan(abs_tau / abs_sigma);
```
hroest: how can sigma ever by negative? This is supposed to be the square root of the variance ... if this is negative, then the fit failed, right? I think simply setting it to the absolute value is a hack but leads to very wrong results ... should it not just be zero then? 

Also, do we have an example of such a degenerate case? maybe we can add it to the tests?

hroest:
@timosachsenberg @jpfeuffer that looks like a short term fix but I think that its a pretty bad choice in terms of the area and the other parameters that will be derived from this. I think if it fails to fit we should set sigma either to zero or throw an exception

jpfeuffer:
Yes I have an example here. I can add it. I agree that it looks weird but I don't think that it's necessarily wrong to use the absolute value here. I think the negative values only arise if they were fitted through the LM procedure which tries to fit Sigma and can result in local optima with a negative value. Not an expert here though. I can check if the fits still make sense.
You can see in the rest of the code that the abs was already taken at some points. In the others problems just didn't appear because the square of Sigma is used.

hroest:
>  I think the negative values only arise if they were fitted through the LM procedure which tries to fit Sigma and can result in local optima with a negative value

I agree but I dont quite see how the absolute value is any more correct than using zero -- unless what you are really fitting is sigma_squared ? Still, if possible lets add the example. Does the area seem sensible when you look at the example - I am just a bit worried that the area calculation does not make any sense any more once you have negative values from the fit.

PR #3197, opened 2018-01-25
[REFACTOR] MRMTransitionGroupPicker uses PeakIntegrator's methods

pasqoo:
**What changed**
I removed the old methods:
`calculateBgEstimationAverage_()`
`calculateBgEstimationExact_()`
`calculatePeakApexInt_()`
`calculatePeakShapeMetrics_()`

Instead, this class now uses _PeakIntegrator_'s methods:
`integratePeak()`
`estimateBackground()`
`calculatePeakShapeMetrics()`

**Why**
We're planning to implement the mean for this class to accept component-wise parameters, instead of relying on the same set of parameters for the entire set of components.

This feature will be possible thanks to two other classes I'm soon going to propose (_MRMFeaturePicker_ and _MRMFeaturePickerFile_).

**Side effects**
After these changes, many tests from TOPP tools failed.
Another test which failed (and I corrected this one) is: UTILS_MRMTransitionGroupPicker_test_1_out

If the code looks good to you, I will update all the tests files to make them pass. (basically I'm planning to copy/paste the generated files)


src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
```diff
@@ -119,114 +119,6 @@ namespace OpenMS
 
     picker_.setParameters(param_.copy("PeakPickerMRM:", true));
```
hroest: you should also expose and copy the peakpicker params:

```
pi_.setParameters(param_.copy("PeakIntegrationMRM:", true));
```

and above

```
defaults_.insert("PeakIntegrationMRM:", PeakIntegrator().getDefaults());
```

hroest:
overall looks good to me, of course the tests would need to pass and since this will change many of the parameters we should make sure its as easy as possible to migrate.

hroest:
why do we see such large changes in the tests? I suggest that we add a new test for the new settings and use the old settings for the current test `MRMTransitionGroupPicker_1_output.featureXML`

```
-					<UserParam type="float" name="peak_apex_int" value="30"/>
+					<UserParam type="float" name="peak_apex_int" value="390"/>
```

pasqoo:
We see such large changes because the previous method for integrating peaks (`calculatePeakApexInt_()`) used to take the peak's apex position in input. Said value was taken from the transition group with the instruction (line 217 of .h file):
`double peak_apex = picked_chroms[chr_idx][peak_idx].getRT();`
The value was also subject to change in some cases, eg. when `recalculate_peaks_` is `true` and `recalculatePeakBorders_()` would change the left and right limits:
```c++
if (recalculate_peaks_)
{
  // This may change best_left / best_right
  recalculatePeakBorders_(picked_chroms, best_left, best_right, recalculate_peaks_max_z_);
  if (peak_apex < best_left || peak_apex > best_right)
  {
    // apex fell out of range, lets correct it
    peak_apex = (best_left + best_right) / 2.0;
  }
}
```
And then `calculatePeakApexInt_()` would create a value for `peak_apex_int` with the following lines:
```c++
peak_apex_int = -1;
double peak_apex_dist = std::fabs(chromatogram.begin()->getMZ() - peak_apex_rt);
// ... hiding some code here ...
for (chromatogram loop here)
{
  DPosition<2> p;
  p[0] = it->getMZ();
  p[1] = it->getIntensity();
  hull_points.push_back(p);
  if (std::fabs(it->getMZ() - peak_apex_rt) <= peak_apex_dist)
  {
    peak_apex_int = p[1];
    peak_apex_dist = std::fabs(it->getMZ() - peak_apex_rt);
  }
}
```

The new method (`PeakIntegrator::integratePeak()`) doesn't take the peak's apex position in input. It searches for the peak's apex position for any given peak.
```c++
for (auto it = p.PosBegin(left); it != p.PosEnd(right); ++it)
{
  hull_points.push_back(DPosition<2>(it->getPos(), it->getIntensity()));
  if (peak_height < it->getIntensity())
  {
    peak_height = it->getIntensity();
    peak_apex_pos = it->getPos();
  }
}
```

The group's apex RT is not used here, but it's not completely discarded. Instead, it is saved in a new metavalue named `peak_apex_position`, which is what all those new green lines are in the new `.featureXML` committed file.  This also results in the highest intensity for the peak to be chosen because the retention time is left as a free parameter instead of specifying the retention time to choose the peak apex as.

Please let us know your thoughts on these changes.  If this change potentially means breaking too many existing workflows using _MRMTransitionGroupPicker_, we can add in a parameter (e.g., "align_peak_apices") to specify using the previous implementation, or the proposed implementation.

hroest:
@pasqoo great, I am 👍  with this change, go ahead and change the test files. I will think about how to best support the legacy options as I would like to at least still support the legacy background subtraction options in OpenSWATH.

pasqoo:
@hroest Hey Hannes, I'm a bit confused here because all tests pass on my machine, but here both travis and jenkins fail.
What could it be?

timosachsenberg:
rebuild win

hroest:
I can report on my machine, I see the same issues as on travis

```
The following tests FAILED:
        1655 - TOPP_OpenSwathWorkflow_3_out1 (Failed)
        1660 - TOPP_OpenSwathWorkflow_5_out1 (Failed)
        1663 - TOPP_OpenSwathWorkflow_6_out1 (Failed)
        1666 - TOPP_OpenSwathWorkflow_7_out1 (Failed)
        1687 - TOPP_OpenSwathWorkflow_16_out1 (Failed)
```

so this is not a fluke.

I have created a new branch here https://github.com/OpenMS/OpenMS/tree/fix/osw and it is currently run against travis so lets see what happens here: https://travis-ci.org/OpenMS/OpenMS/builds/340604145

pasqoo:
Thanks @hroest . I need to find out what's wrong with my environment.

Would it be possible for you to replace those tests' files? Because I don't have those .tmp files at hand, since the tests pass in my machine.

timosachsenberg:
+1

pasqoo:
@hroest Thanks again for your help! When I was updating this class, I just looked for all calls to the previous method `calculatePeakApexInt_()` and the others, thinking I was done.

I guess the only thing missing is the background subtraction to the precursor integration.

If this gets merged before I'm done with that, I'll just open a new PR.

PR #3199, opened 2018-01-25
[FEATURE] Support of new percolator parameter

grosenberger:
This PR enables support for a new Percolator hyperparameter to restrict training to the best scoring PSMs.

### DIFF
--- a/src/tests/topp/THIRDPARTY/PercolatorAdapter_1.ini
+++ b/src/tests/topp/THIRDPARTY/PercolatorAdapter_1.ini
@@ -8,7 +8,6 @@
       <ITEMLIST name="in_decoy" type="input-file" description="Input decoy file(s) in case of separate searches" required="false" advanced="false" supported_formats="*.mzid,*.idXML">
       </ITEMLIST>
       <ITEM name="out" value="" type="output-file" description="Output file in idXML format" required="false" advanced="false" />
-      <ITEM name="mzid_out" value="" type="output-file" description="Output file in mzid format" required="false" advanced="false" />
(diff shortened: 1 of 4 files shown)


jpfeuffer:
Hi @grosenberger +1 for this addition. We recently had some confusion in KNIME with the mutually exclusive output parameters. Since I think there is not sensible solution, could you add a warning when both types of output files are given?
E.g. "you can't have both outputs. please deactivate one. In KNIME go to the OutputTypes tab and set one to Inactive"

grosenberger:
Good point. I added a patch for this issue. A cleaner alternative might be to to merge all formats into single ``-in`` and ``-out`` parameters and then have also ``-in_type`` and ``-out_type`` parameters together with file ending detection. Will this result in any other problems? Else, I could also implement this solution if preferred.

jpfeuffer:
Hm good idea. The only issue with that approach is that you could not have both mzid and idXML at the same time anymore. And the user still has to go to the OutputTypes because we do not have an automatic association with the parameters and the creation of an appropriate output file name ending. So, I would say almost equal.

jpfeuffer:
I thought about it a bit. What about using one output and let the application choose the algorithm based on the file ending. Since there are restrictions on the file ending, the user can't choose an arbitrary one anyway. Maybe it is not perfectly represented in the ini then though.
@timosachsenberg any preference?

timosachsenberg:
I think having one mandatory output is best and choosing the algorithm would work.
If algorithms diverge more in the future, we could also think about having two tools.

jpfeuffer:
Cool. Default from extension plus overwrite param. +1

PR #3204, opened 2018-01-30
[FEATURE] AbsoluteQuantitation::getOptimizedCalibrationCurve()

pasqoo:
This function optimizes the parameters of the calibration for a single component, instead of a set of components (already implemented in `optimizeCalibrationCurves()`).
Internally, this method also calls said method.


timosachsenberg:
The interface (maybe only the name) of the new function confused me a bit.
It is certainly no getter and more like an "updateComponentConcentrations". Might be personal taste, but I don't like methods that update results in-place if it can be avoided.

An idea how this could be changed: provide a function that calculates the component concentrations of a **single** component. Building on that you could create the function that does it for all components at once. So this would basically reverse the logic you use now (currently you use a function that updates everything but you actually only want to update a single component).

I only gave it a brief look so maybe there is good reason to do it the other way around... so sorry about that if that is the case.

pasqoo:
I agree that the design of the new method is not exciting, and your solution is more elegant.
The thing is, this new method was not really needed. It's been implemented only to offer some python binding, because users could not currently use what was already implemented (`optimizeCalibrationCurves()`) due to the unsupported nested type argument.
This new method is not meant to be used from the "c++ side".

Do you think a change to the workflow is necessary?

In any case, I think that improving the method's name is a good idea.
Thanks for the proposed name, I'll take that into account.

timosachsenberg:
Ok I see, in this case it is fine with me. Feel free to change the name.

pasqoo:
PR is up to date @timosachsenberg @hroest

hroest:
👍 if it passes tests

pasqoo:
@hroest thanks :smile: 
looks like only the apple config is failing. It complains about files not being readable

timosachsenberg:
If you find time (and maybe just as an additional commit to some unrelated PR) you might want to document that this method is provided primarily to ease python bindings.

pasqoo:
Sure, I'll take care of that.
@timosachsenberg I noticed that I also staged some untracked files in my last commit (an .mzML file and 3 csv files). This was not intended.
The files are output files from other classes' tests.
Should I leave them there, or make a PR to remove those files?

timosachsenberg:
Please remove them in another PR

PR #3205, opened 2018-01-30
[FEATURE] AbsoluteQuantitationMethod operator== and operator!=

pasqoo:
Implemented the operators `==` and `!=` for _AbsoluteQuantitationMethod_.
Also updated the _AbsoluteQuantitationMethodFile_'s tests to take advantage of this feature.


src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
```diff
@@ -106,6 +106,46 @@ namespace OpenMS
     String transformation_model_; ///< transformation model
     Param transformation_model_params_; ///< transformation model parameters
   };
+
+  inline OPENMS_DLLAPI bool operator==(const AbsoluteQuantitationMethod& lhs, const AbsoluteQuantitationMethod& rhs)
+  {
+    if (
```
timosachsenberg: I think you can just use:
return std::tie(lhs.a, , ,  ) == std::tie(rhs.a, ..., ) with all fields you want to compare.
Note that this can also include the Param object (as these also define operator==).
this will simplify this function a lot.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
```diff
@@ -106,6 +106,46 @@ namespace OpenMS
...
+      lhs.getConcentrationUnits() != rhs.getConcentrationUnits() ||
+      lhs.getTransformationModel() != rhs.getTransformationModel()
+    )
+    {
+      return false;
+    }
+    // compare params
+    const Param lp = lhs.getTransformationModelParams();
+    const Param rp = rhs.getTransformationModelParams();
+    if (lp.size() != rp.size())
```
hroest: I think you could just replace this with 
`if (lp == rp)`
right?

timosachsenberg: or put the whole param into the tuple of the std::tie ... or do I miss something?

pasqoo:
Thanks, I did not think about checking if _Param_ had an `operator==` defined (or about implementing the functionality, if it was missing).
From what I understood about `std::tie()`, it seems it wouldn't help in this case. It's about "tying" lvalues to a tuple's elements.
The example from this page should explain better than my words: http://www.cplusplus.com/reference/tuple/tie/

Question: no mention of this method definition is made within the AQM class. Should I define it as `friend`, within the class? And then add tests in `AbsoluteQuantitationMethod_test.cpp`
If I define it as `friend`, would it be better to avoid using the getters, and use the private members directly? Avoiding those method calls.

timosachsenberg:
Example: you write something like:
```

bool operator==(MyStruct const& lhs, MyStruct const& rhs){
  return std::tie(lhs.one_member, lhs.another, lhs.yet_more) ==
         std::tie(rhs.one_member, rhs.another, rhs.yet_more);
}
```
and you already have your comparison operator. 
Kind of the new way how to write comparison operators in c11.
For operator< you would just write:
```
  return std::tie(lhs.one_member, lhs.another, lhs.yet_more) <
         std::tie(rhs.one_member, rhs.another, rhs.yet_more);
```
makes both implementations look very similar and clean.

pasqoo:
I think there is a misunderstanding about the way `std::tie()` works.
From what I understood, it is something to use the following way:
```c++
double somevar; // lvalue
int someothervar; // lvalue
// now i want to tie some lvalues to some rvalues, and these rvalues are tuple's members
std::tie(somevar, someothervar) = tuple of two elements, one double and one int
```
If the tuple's structure doesn't match `std::tie()`'s given args, one can use `std::ignore` to skip one or more tuple's members.

Please correct me if I got something wrong.

PS: What do you think about the `friend` keyword question in my previous post?

pasqoo:
@timosachsenberg Thanks for your support on Gitter. I found the way to use `std::tie()`.
I did not know I could access another object's private members, and this kept me blocked for some time (also I was missing what `std::tie()` returns).

I added some tests and a custom constructor to AQM which takes care of initializing the class' members.

pasqoo:
travis failed because of a timeout. the rest looks fine.

PR #3213, opened 2018-01-31
[FIX] draw text on 3D view

timosachsenberg:


timosachsenberg:
image
some other bugs persist when view from top is activated.

PR #3218, opened 2018-02-01
[DOCS] Add python binding note to AQ::optimizeSingleCalibrationCurve()

pasqoo:
Also removed unnecessary tests' output files from other classes.


timosachsenberg:
cool. Thanks!

PR #3220, opened 2018-02-02
[FEATURE] Also try to find static zlib

jpfeuffer:


jpfeuffer:
Nice. It immediately picked it up.  Any preferences on static/dynamic. Our zlib cmake module does not support choosing. This would be another issue. I changed it because suddenly zlib was not found on Win anymore. The dll went to bin instead of lib and the static lib was suddenly suffixes with static.

jpfeuffer:
I mean it was found for linking but not during testing. I guess I removed the contrib bin folder from the path to clean up a bit.

timosachsenberg:
+1

PR #3222, opened 2018-02-03
[FIX] set compression

hroest:
probably needs changes to the test files...


timosachsenberg:
After compression small binary blobs are larger than before. Guess this is not a problem in practice

hroest:
> After compression small binary blobs are larger than before. Guess this is not a problem in practice

no, its not :-)

timosachsenberg:
rebuild jenkins

PR #3223, opened 2018-02-05
some performance optimizations (especially XML parsing)

cbielow:
[FEATURE] faster whitespace removal function (2x - 6x faster than before)

On string with no whitespaces (stringlen 3000, 1e5 iterations):
old: 4.46 s (wall), 4.46 s (CPU), 0.00 s (system), 4.46 s (user)
new: 2.68 s (wall), 2.68 s (CPU), 0.00 s (system), 2.68 s (user)

On string with whitespaces: (stringlen 3000, 1e5 iterations):
old: 25.03 s (wall), 25.02 s (CPU), 0.00 s (system), 25.02 s (user)
new: 3.37 s (wall), 3.37 s (CPU), 0.00 s (system), 3.37 s (user)

[FIX] some cleaning up in XML parsing code, e.g. Base64 uses static functions only, so we can remove instanciation

In total, mzML parsing (FileInfo -in ...) is about 6% faster (~60 vs 63 MB/sec) on my machine.
There is a lot more potential IMHO, but I take what I can get :)

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
@@ -781,31 +781,26 @@ public:
     return this_s;
   }
 
-  static String& removeWhitespaces(String & this_s)
+  static String& removeWhitespaces(String& this_s)
   {
-    bool contains_ws = false;
-    for (std::string::const_iterator it = this_s.begin(); it != this_s.end(); ++it)
+    int shift(0); // windowing number of removed whitespaces
+    std::string::iterator it_end = this_s.end();
+    for (std::string::iterator it = this_s.begin(); it != it_end; ++it)
     {
-      char c = *it;
+      const char c = *it;
       if (c == ' ' || c == '\t' || c == '\n' || c == '\r')
       {
-        contains_ws = true;
-        break;
+        ++shift;
+        continue; // no need to shift this 
+      }
+      if (shift)
+      { // move current char to the left
+        *(it - shift) = c;
       }
     }
 
-    if (contains_ws)
-    {
-      std::string tmp;
-      tmp.reserve(this_s.size());
-      for (std::string::const_iterator it = this_s.begin(); it != this_s.end(); ++it)
-      {
-        char c = *it;
-        if (c != ' ' && c != '\t' && c != '\n' && c != '\r')
-          tmp += c;
-      }
-      this_s.swap(tmp);
-    }
+    // shorten result
+    if (shift) this_s.resize(this_s.size() - shift);
 
     return this_s;
   }
--- a/src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
@@ -783,24 +783,26 @@ public:
 
   static String& removeWhitespaces(String& this_s)
   {
-    int shift(0); // windowing number of removed whitespaces
+    std::string::iterator it = this_s.begin();
+    std::string::iterator dest = it;
     std::string::iterator it_end = this_s.end();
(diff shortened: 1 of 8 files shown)


timosachsenberg:
looks good. some minor style issues on CDASH http://cdash.openms.de/testDetails.php?test=47239110&build=256751
Others are probably related to the travis cache not being cleared...

timosachsenberg:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=256810

hroest:
👍 if the tests passes, some comments in the code

src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
```diff
@@ -781,32 +781,27 @@ namespace OpenMS
...
-    if (contains_ws)
-    {
-      std::string tmp;
-      tmp.reserve(this_s.size());
-      for (std::string::const_iterator it = this_s.begin(); it != this_s.end(); ++it)
-      {
-        char c = *it;
-        if (c != ' ' && c != '\t' && c != '\n' && c != '\r')
-          tmp += c;
+      if (shift)
```
hroest: why not use an iterator for the destination and use 
```
std::string::iterator dest = this_s.begin();
if (dest != it) *dest = *it;
```
and then increase `dest` and `it` together except if we hit a whitespace?

hroest:
did you compare this to `str.erase(remove_if(str.begin(), str.end(), isspace), str.end());` ?

src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
cbielow: I tried this, and performance-wise, there is no measureable difference.
So I guess its a matter of taste. Any strong feelings?

```
static String& removeWhitespaces2(String& this_s)
{
  std::string::iterator it = this_s.begin();
  std::string::iterator it_end = this_s.end();
  std::string::iterator dest = this_s.begin();
  while (it != it_end)
  {
    const char c = *it;
    if (c == ' ' || c == '\t' || c == '\n' || c == '\r')
    {
      ++it;
      continue;
    }
    if (dest != it) *dest = *it;
    ++dest;
    ++it;
  }

  // shorten result
  if (dest != it) this_s.resize(dest - this_s.begin());

  return this_s;
}
```

cbielow:
Not before, but our own implementation seems to do way better:

Without spaces:

Erase: 6.25 s (wall), 6.24 s (CPU), 0.00 s (system), 6.24 s (user)
this PR: 2.69 s (wall), 2.68 s (CPU), 0.00 s (system), 2.68 s (user)

With spaces:
Erase: 7.63 s (wall), 7.61 s (CPU), 0.00 s (system), 7.61 s (user)
this PR: 3.68 s (wall), 3.67 s (CPU), 0.00 s (system), 3.67 s (user)

cbielow:
tests look good (except for some false positives). will merge

hroest:
> remove mem test after free (not guaranteed to work on Linux) 2

hmm not sure I agree, why was this working before and now suddenly it doesnt work any more? sounds suspicious ?

cbielow:
I've been digging through documentation on how (virtual-)memory/paging is handled in C++ libs, but did not really get a good answer.
All I can say is: I'm very sure there is no (new) mem-leak since the code change is minimal and when used on Windows, you get all the memory back (even when iterating over loading/releasing an MSExperiment 100 times in the SysInfo test, the memory after the last release is the same as before the first).
So assuming that g++/clang does not have a buggy memory manager, the only explanation which is left is that upon 'free()' (if used under the hood of STL's delete operator), the memory manager decides not to return memory (yet) to the OS but rather hold onto it for the next malloc/new.
(see https://www.gnu.org/software/libc/manual/html_node/Freeing-after-Malloc.html#Freeing-after-Malloc).
I don't see why this would be a bad thing.

PR #3224, opened 2018-02-05
[FIX] pyOpenMS deleted constructor

timosachsenberg:
pyOpenMS seems to have problems with deleted constructors


timosachsenberg:
will merge to see if this fixes the nightlies... one travis build timed out

jpfeuffer:
Main nightly failure is crux adapter on macOS

jpfeuffer:
I put the pyopenms tests on ignore (for successful build). It is almost impossible right now to fix them all.

jpfeuffer:
But if it does fix them, I will enable them again.

hroest:
> pyOpenMS seems to have problems with deleted constructors

yes it does, since pyOpenMS works with a copy of the object it wont work if there is no constructor...

PR #3225, opened 2018-02-06
[REFACTOR][FIX] AbsoluteQuantitationMethodFile support for empty cells and missing headers

pasqoo:
The class used to assume that params cells could not be empty. Assumption that not always is true.

While fixing the bug, I found that some parts could be simpler. So I went for a refactor and I'm proposing this version, hoping I have reached my goal :smile: 

I got rid of `parseHeader_()`. The headers are now extracted within `load()`.
I added a helper method named `setCastValue_()`.

I added many more tests to make sure that missing headers or missing information (empty cell) do not cause troubles.


pasqoo:
Travis failed because of:

hroest:
can you resolve the merge conflicts, then we will look at it?

pasqoo:
@hroest Done, PR updated.

Issue #3203, opened 2018-01-30
Something during XML loading (re)sets locale settings


jpfeuffer:
This let's all C library printing and conversion functions "fail"/be inconsistent. (sprintf, atof, ...)
Although we keep a pointer to the locale settings in our Types.cpp, the contents get overwritten, too.
E.g. before https://github.com/OpenMS/OpenMS/blob/develop/src/topp/OpenSwathFeatureXMLToTSV.cpp#L450
the locale is "C", but after it is "de_DE.utf8" (which is set in my LANG environment variable).

I read about QApplications of QT resetting the locale but could not find where it would be constructed in that tool.
Could also be Xerces.
I did not find any other setlocale usage of our library.

timosachsenberg:
hmm just an idea: https://github.com/OpenMS/OpenMS/blob/876dbc38afa22e19111b39e016c681574ae57462/src/openms/source/SYSTEM/UpdateCheck.cpp#L137
This constructs a QCoreApplication which might set the local

jpfeuffer:
Hmm I was checking. But the tests correctly have a "-test" flag which should deactivate UpdateCheck right?

jpfeuffer:
Until we find the true reason: @grosenberger or @hroest can you have a look at OpenSwathFeatureXMLtoTSV and find an alternative to sprintf so that the tool does not depend on the locale?

jpfeuffer:
Not sure why this is showing up now on the build machines (QT5? threadsafety of the locale?) but as you see, I remember having the problem before on my Mac.

´´´
Offending lines:			(tab_width = 8, first_column = 1)

in1:  C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\tests\topp\OpenSwathFeatureXMLToTSV_3_output.short.csv   (line: 2, position/column: 360/531)
tr_gr1_run0	0	testfile.file	3119.09196821988	f_11965988948540676971	PEPTIDEA	NA	NA	500	3574.23		0	3118.65196821988	3096.28002929688	-21994.8526298809	2	3147.67993164062	3118.65196821988	86.0041379995282	3680.16062933207	0.985403060913086	0.971216417399615	1	0.108663236021716	4.45439541136954	3118.65196821988	0	0	0.998183460545928	0.997577947394571	803.023010!
[...]

PR #3226, opened 2018-02-06
[Feature] Packaging qt5

jpfeuffer:
Plus its platform plugins.
Major rework and simplification of all the packaging CMake scripts.
Kept dependencies to a minimum.
Relies on all Win libs to be static now. See current contrib PRs.
Tested on all platforms but due to many switches back and forth there might be little things that were overseen in the end.
Adapted docu accordingly.


cmake/knime_package_support.cmake
```diff
@@ -184,84 +184,60 @@ add_custom_target(
...
-    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:OpenSwathAlgo> ${PAYLOAD_LIB_PATH}
-    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:SuperHirn> ${PAYLOAD_LIB_PATH}
-  )
-
-  function(copy_library lib target_path)
-    string(REGEX REPLACE "lib$" "dll" target_dll "${lib}")
-    file(TO_NATIVE_PATH "${target_dll}" target_native)
+else()
+  ## Assemble common required libraries for win and lnx
+  ## Note that we do not need the QT plugins or QTGui libraries since we do not include GUI tools here.
```
timosachsenberg: question if we want to ship TOPPView in KNIME

timosachsenberg:
great job! Two things we might change:
- remove IDEvalutator as we kind of agreed to remove it
- bundle TOPPView in KNIME

jpfeuffer:
Sure we can think about that. We can remove the qt librarypath config hack. For TOPPView we would need to leave the guy libraries. TOPPView needs a ctd then. Is it using toppbase already? We have to check if it needs to be a bundle or an executable on Mac. Should we do this here or try to merge first?

jpfeuffer:
rebuild jenkins

timosachsenberg:
let's merge first

timosachsenberg:
maybe remove the hack and IDEvaluator if that's fine with you. TOPPView might need some more work

jpfeuffer:
I think finally we are struck by the locale error. We should fix it.
sprintf should not be used.
See #3203

hroest:
>  sprintf should not be used.

what is the alternative, should we use boost::format ?

jpfeuffer:
Good idea, maybe boost::locale::format?
http://www.boost.org/doc/libs/1_56_0/libs/locale/doc/html/group__format.html#gad7914df7b54382c1ad7f5360676fe2e8

hroest:
@jpfeuffer I have fixed this here: #3253 - was this what you had in mind?

timosachsenberg:
@hroest not sure if this solves the problem. From the doc, it seems to use the global local. Maybe we need to set it explicitly to the US local when we create the string?

timosachsenberg:
I merge to reduce the number of open qt5 branches. let's address the locale problem separately

Issue #3216, opened 2018-01-31
Toppview goes black when switching to 3D view for a second time


hroest:
Branch: `feature/qt5` ae0baaf40e51b359b2565873457e7e3b8b69f329
System: Linux

Problem: Toppview goes black when switching to 3D view for a second time
Reproduce problem: 
- Open TOPPView, select 2D view
- Right click, click "open 3D view"
- Go back to previous tab
- Right click, click "open 3D view"
- Result: black window, no work possible

![toppview](https://user-images.githubusercontent.com/464014/35651918-6b3ade54-06af-11e8-9c68-19e2528b3fae.png)

jpfeuffer:
Interesting. Can't confirm on Mac.

timosachsenberg:
can't confirm on Linux.

timosachsenberg:
Qt issues sometimes a warning that the paint device is not ready. I check this now explicitly and the warning goes away. https://github.com/OpenMS/OpenMS/pull/3217
Might be (probably not) connected to this issue

hroest:
definitely something weird going on here, I still have this issue on the newest branch. However I see a difference on Linux between Intel and NVIDIA gpu, on Intel I see white flickering while on NVIDIA I see black screens. Still while moving the cursor and switching tabs I see a black window:

https://www.dropbox.com/sh/fu87nvwi3z25shf/AACKKq6MzV12RnJh2fHXExWqa?dl=0

hroest:
it seems this is related to the `glEnable(GL_DEPTH_TEST);` call which is called in `paintGL();` and `initializeGL();`

PR #3233, opened 2018-02-08
[FIX] fixes OpenGL glitch

hroest:
fixes Linux glitch with OpenGL - no more flicker, no more black screen on Intel

### DIFF
--- a/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
@@ -202,7 +202,7 @@ namespace OpenMS
       openglwidget()->initializeGL();
     }
     openglwidget()->resizeGL(width(), height());
-    openglwidget()->paintGL();
+    // openglwidget()->paintGL();
   }
 
   void Spectrum3DCanvas::showCurrentLayerPreferences()


hroest:
also works well under NVIDIA graphics card and I dont see any negative impact so far. Please try on other platforms.

hroest:
maybe this is due to multiple calls to paintGL in short series, see https://stackoverflow.com/questions/17710514/when-is-qglwidgets-paintgl-called - maybe we are not supposed to call it manually? Also this is my understanding from http://doc.qt.io/qt-5/qopenglwidget.html it does not sound like one should call this function manually

hroest:
NOTE: still does not fix #3216 -- but now TOPPView ist at least useable

timosachsenberg:
I think its safe to remove the paintGL here. Will check on mac

hroest:
looks like travis passes, from my side this is ready to merge

src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
```diff
@@ -202,7 +202,7 @@ namespace OpenMS
       openglwidget()->initializeGL();
     }
     openglwidget()->resizeGL(width(), height());
-    openglwidget()->paintGL();
+    // openglwidget()->paintGL();
```
hroest: this is the main change here that fixes the Linux issue

PR #3234, opened 2018-02-08
[FIX] remove AbsoluteQuantitation::optimizeCalibrationCurves() from pyOpenMS

pasqoo:
All returned lists of `AbsoluteQuantitationStandard::featureConcentrations` are encapsulated in an extra list layer (e.g., `[[...]]` instead of `[...]`)

We are proposing to remove the wrapping. Python users would use `optimizeSingleCalibrationCurve()` instead.


pasqoo:
@hroest the failing tests don't seem related to this PR. What do you think?

PR #3235, opened 2018-02-08
[FIX] AQ's LLOQ and ULOQ not correctly computed when calibration curve not found

pasqoo:
In `AbsoluteQuantitation::optimizeCalibrationCurves()` the values set for the LLOQ and ULOQ are not correct when an optimal set of calibration points is not found.

This fix uses `min_element()` and `max_element()` instead of relying on elements sorting.


timosachsenberg:
looks good.
One (mainly style) thing: you could use a lambda for the simple comparison instead of defining a struct?
I find code like:
```
auto it = std::min_element(v.begin(), v.end(), 
    [](const Donut& x, const Donut& y) {  return x.diameter < y.diameter; });
```

easier to read

pasqoo:
@timosachsenberg even though I'm using it twice?

pasqoo:
Long variable names and types were involved. Therefore I preferred to go for a single object to call, instead of multiple lines of the same thing.
I prefer to avoid using `auto` too much because it hides information.
What do you think?

timosachsenberg:
It's mainly a matter of taste.  And in the case you use it twice its fine.
Regarding using auto for nearly everything... this is still under discussion in the C++ community.

pasqoo:
@timosachsenberg PR updated.

hroest:
>  Regarding using auto for nearly everything... this is still under discussion in the C++ community.

I dont think its good either, I like to be explicit when possible. Even better if we dont use complicated constructs of nested maps and vectors - better use small helper classes and typedefs. I think that would also solve the problem and not require auto everywhere.

hroest:
otherwise a 👍 from me, if Timo agrees we can merge

PR #3237, opened 2018-02-08
Feature/toppview black

### LINKED ISSUE #3216: Toppview goes black when switching to 3D view for a second time

hroest:
24d24a1 now also fixes #3216 

again probably somebody who knows whats going in here should have a look at the update functions, I am just finding these fixes by trial and error

### DIFF
--- a/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
@@ -202,7 +202,7 @@ namespace OpenMS
       openglwidget()->initializeGL();
     }
     openglwidget()->resizeGL(width(), height());
-    openglwidget()->paintGL();
+    // openglwidget()->paintGL();
   }
 
   void Spectrum3DCanvas::showCurrentLayerPreferences()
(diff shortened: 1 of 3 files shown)


timosachsenberg:
looks correct. In the past, I also observed that some drivers are very sensitive in aquiring an opengl context multiple times while other don't care. I think that's what  the extra call to initializeGL did.

src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
```diff
@@ -199,10 +199,10 @@ namespace OpenMS
       {
         openglwidget()->updateIntensityScale();
       }
-      openglwidget()->initializeGL();
+      // openglwidget()->initializeGL();
```
hroest: yes, I think initializing opengl twice is what caused issues with my drivers here

Issue #3229, opened 2018-02-06
Qt5: automatic tiling and linking zoom doesnt work


hroest:
try with these files:

TOPPView src/tests/topp/OpenSwathChromatogramExtractor_output*

![screenshot from 2018-02-06 17-34-09](https://user-images.githubusercontent.com/464014/35887943-fdb85974-0b63-11e8-9466-733e2c334b2c.png)
![screenshot from 2018-02-06 17-33-46](https://user-images.githubusercontent.com/464014/35887944-fdc471be-0b63-11e8-93eb-6a166b6a7d8e.png)
![screenshot from 2018-02-06 17-33-29](https://user-images.githubusercontent.com/464014/35887945-fdd46312-0b63-11e8-8f7f-d19bbb2338d8.png)

the window tiling produces strange results and the automated zoom linking does not work any more. These are limiting the options quite a lot for chromatographic viewing

PR #3238, opened 2018-02-08
[FIX] fix zoom link bug

hroest:
fixes the zoom link, see #3229


hroest:
now fixes also the horizontal and vertical tiling

PR #3227, opened 2018-02-06
[TEST/FIX] Try to fix crux tests by using reverse instead of shuffled

jpfeuffer:
Try to fix #3201

PR #3239, opened 2018-02-09
[Fix] xerces platform init win

jpfeuffer:
Don't know why this showed up only when linking to static Xerces. Or the newer version?
But I think this here is the correct way.
One could discuss of course where to correctly put the initialization. In the class?

This here solves it on win for me.

And oops. Since I tested this on the Crux test fix PR, it includes its commits, too.

### DIFF
--- a/src/tests/class_tests/openms/source/Bzip2InputStream_test.cpp
+++ b/src/tests/class_tests/openms/source/Bzip2InputStream_test.cpp
@@ -44,7 +44,7 @@ using namespace OpenMS;
 ///////////////////////////
 
 START_TEST(Bzip2InputStream, "$Id$")
-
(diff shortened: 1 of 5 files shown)


timosachsenberg:
+1

hroest:
>  One could discuss of course where to correctly put the initialization. In the class?

yes, especially since we should be calling `XMLPlatformUtils::Terminate` as well. So putting it into a class constructor / destructor seems like the optimal way of dealing with this. But lets first fix the tests and then think about this

jpfeuffer:
rebuild jenkins

jpfeuffer:
Nice. Apple had problems submitting to CDash. Happens sometimes.
And we probably should silence the CMake warning.

jpfeuffer:
supersedes #3227

PR #3243, opened 2018-02-10
Change default for decoy library generation

hroest:
Currently our decoy library generation is quite poor since we only have two options

- shuffle the whole protein
- reverse the whole protein

However, this strategy does not conserve the overall mass distribution of peptides and may introduce biases into the target-decoy approach. It seems better to be able to shuffle and/or reverse on peptide level with the possibility of keeping either C or N terminus constant. This PR will introduce this as default option. The new options will enable users to 

- shuffle with either C, N, CN or no terminus constant
- reverse with either C, N, CN or no terminus constant

One downside of this is that very short peptides (<3-4 AA) cannot be properly reversed, e.g. `PEK` reversed with C/N terminus constant will still be `PEK` while a peptide like `PETK` will simply become `PTEK` which is highly similar.

Another important point for redundant databases is whether the decoy and target DB have the same number of unique peptides. If not, there will be biases towards the decoy database as there are many more unique peptides than in the target DB (since we shuffle with different seeds each time). This PR ensures that the same seed is used for each peptide, resolving this problem. Alternatively, we could also keep a list of all peptides and their mapping to a decoy peptide to solve this problem, but this may be quite memory intensive.


src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -43,60 +43,68 @@
...
+  MRMDecoy::MRMDecoy() :
+    DefaultParamHandler("MRMDecoy"),
+    ProgressLogger()
   {
-    std::vector<std::pair<std::string::size_type, std::string> > idx;
-    std::vector<std::string> pattern;
-    pattern.push_back("K");
-    pattern.push_back("R");
-    pattern.push_back("P");
+    defaults_.setValue("non_shuffle_pattern", "K,R,P", "Residues to not shuffle (keep at a constant position when shuffling). Separate by comma, e.g. use 'K,P,R' here.");
```
timosachsenberg: maybe directly use a stringlist?

src/utils/DecoyDatabase.cpp
```diff
@@ -153,33 +224,61 @@ class TOPPDecoyDatabase :
...
-        else // reverse
+        else
         {
-          protein.sequence.reverse();
+          // sequence
+          if (shuffle)
+          {
+            String temp;
+            Size x = protein.sequence.size();
+            srand(time(nullptr));
```
timosachsenberg: why don't we use seed here?

timosachsenberg:
Hi,
Nice PR. 
Also worth to consider would be a hybrid approach in the future: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4941923/

timosachsenberg:
+1 from my side if seed is added. @cbielow any comments?

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -43,60 +43,68 @@
...
-      {
-        if (sequence.substr(i, 1) == pattern[j])
-        {
-          std::pair<std::string::size_type, std::string> idx_pair(i, pattern[j]);
-          idx.push_back(idx_pair);
-        }
-      }
-    }
-    return idx;
+    defaults_.setValue("keepN", "true", "Whether to keep N terminus constant", ListUtils::create<String>("advanced"));
```
cbielow: is this *protein* N/C-term?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMDecoy.h
```diff
@@ -132,44 +138,70 @@ namespace OpenMS
...
+                const std::vector<String> & const_pattern = std::vector<String>());
 
     /**
       @brief Find all K, R, P sites in a sequence to be set as fixed
 
       This method was adapted from the SpectraST decoy generator
     */
-    std::vector<std::pair<std::string::size_type, std::string> > findFixedResidues(
-      std::string sequence);
+    IndexType findFixedResidues(std::string sequence);
```
cbielow: could this be
`IndexType findFixedResidues(const std::string& sequence) const;`?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMDecoy.h
```diff
@@ -132,44 +138,70 @@ namespace OpenMS
...
 
     /**
       @brief Find all K, R, P and C-/N-terminal sites in a sequence to be set as fixed
 
       This method was adapted from the SpectraST decoy generator
     */
-    std::vector<std::pair<std::string::size_type, std::string> > findFixedAndTermResidues(
-      std::string sequence);
+    IndexType findFixedAndTermResidues(std::string sequence) const;
```
cbielow: could this be
`IndexType findFixedAndTermResidues(const std::string& sequence) const;`?

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
cbielow: This means peptide C/N terms, right?! we should make this clear.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -43,60 +43,68 @@
...
-          std::pair<std::string::size_type, std::string> idx_pair(i, pattern[j]);
-          idx.push_back(idx_pair);
+          idx.push_back(i);
         }
       }
     }
     return idx;
   }
 
+  MRMDecoy::IndexType MRMDecoy::findFixedResidues(std::string sequence)
```
cbielow: can also be const?

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -258,61 +265,70 @@ namespace OpenMS
...
+            reversed.mods[j].location = boost::numeric_cast<int>(k);
+            break;
+          }
         }
       }
+      return reversed;
     }
+  }
+
+  OpenMS::TargetedExperiment::Peptide MRMDecoy::pseudoreversePeptide(
```
cbielow: is `pseudo` a common term in this context?
If not, maybe there is something more descriptive? (reversePeptideConstCTerm()?)

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMDecoy.h
```diff
@@ -132,44 +138,70 @@ namespace OpenMS
...
+      @param peptide The peptide sequence and modifications
+      @param keepN Whether to keep N terminus in place
+      @param keepC Whether to keep C terminus in place
+      @param const_pattern A set of AA to leave in place
+    */
+    static OpenMS::TargetedExperiment::Peptide reversePeptide(
+                const OpenMS::TargetedExperiment::Peptide& peptide,
+                const bool keepN,
+                const bool keepC, 
+                const std::vector<String> & const_pattern = std::vector<String>());
```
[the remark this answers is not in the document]
hroest: no real reason to use a vector or string (except that it was there before) but it does make everything easier to write.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
hroest: we have been using this in the targeted proteomics world quite a bit, so I opt to leave it.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -116,8 +124,8 @@ namespace OpenMS
   }
 
   OpenMS::TargetedExperiment::Peptide MRMDecoy::shufflePeptide(
-    OpenMS::TargetedExperiment::Peptide peptide, double identity_threshold, int seed,
-    int max_attempts)
+    OpenMS::TargetedExperiment::Peptide peptide, const double identity_threshold, int seed,
+    const int max_attempts)
```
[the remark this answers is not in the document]
hroest: yes, that could be the next step. Since these methods are more general than just for MRM decoys, that should be the next step. However, we have to think about whether we then still want to use `OpenMS::TargetedExperiment::Peptide` or switch to `AASequence`

hroest:
I think I addressed all the comments. One question remains: is it possible to let the user provide a user-defined regex or does the user have to modify Enzymes.xml for this? I dont think I got it working as most methods supporting this are private / protected. That seems rather user-unfriendly?

timosachsenberg:
I think it is intended that the user edits the Enzymes.xml. That way a workflow can trace an Enzyme from start to end using our internal file formats. I agree that is is not super user-friendly but a way to ensure the evidence trace for enzymes. This also makes EnzymesDB read-only and we don't run into threading issues as we do with the residue db.

src/tests/class_tests/openms/source/MRMDecoy_test.cpp
```diff
@@ -377,15 +439,14 @@ START_SECTION(OpenMS::TargetedExperiment::Peptide pseudoreversePeptide(OpenMS::T
...
   pseudoreverseAASequence_expected.sequence = "DITPEPTSETE";
   OpenMS::TargetedExperiment::Peptide pseudoreverseAASequence_result;
-  pseudoreverseAASequence_result = gen.pseudoreversePeptide(pseudoreverseAASequence_target_sequence);
+  pseudoreverseAASequence_result = gen.pseudoreversePeptide_helper(pseudoreverseAASequence_target_sequence);
   TEST_EQUAL(pseudoreverseAASequence_result.sequence, pseudoreverseAASequence_expected.sequence)
 }
-
 END_SECTION
 
 START_SECTION(OpenMS::TargetedExperiment::Peptide reversePeptide(OpenMS::TargetedExperiment::Peptide peptide))
```
cbielow: technically, these are `[EXTRA]` now (since protected), but I'm not sure what the status of the checker is anyway.
Maybe discuss at retreat (will add it to the list).

cbielow:
only StopWatch test failed.
+1 for merge

PR #3245, opened 2018-02-12
PyOpenMS packaging in Qt5

hroest:
make some changes to deploy pyOpenMS with Qt5 - so far only Linux is tested. Some observations:

- there is an incompatibility in contrib regarding xerces
- it used to be sufficient to provide `-DQT_QMAKE_EXECUTABLE=` to find Qt4, now that wont work any more with Qt5 and I also have to provide `-DCMAKE_PREFIX_PATH="/qt/"`
- somehow `patchelf` is broken when used with `auditwheel repair` to fix the Qt5Core library in this case and when we try to change the soname of our `libQt5Core` then everything breaks and Python segfaults. I tried to work around this for a long time, wasting many hours until the only solution I found was to undo what  `auditwheel repair` does and just copy back the original libQt. I am not sure what is happening here but I can very clearly see that Python packages with the original `libQt5Core` work and as soon as you run `patchelf --set-soname libQt5Core-82077f0c.so.5.9.4 pyopenms/.libs/libQt5Core-82077f0c.so.5.9.4` (which is what auditwheel essentially does) then the whole library will not work any more and will segfault. This does not happen for any other libraries and it worked for libQt4 so I dont really know why this is. So far I have been able to narrow it down to that `patchelf` call which somehow trashes the Qt5Core library.

### DIFF
--- a/src/pyOpenMS/pyopenms/__init__.py
+++ b/src/pyOpenMS/pyopenms/__init__.py
@@ -8,16 +8,6 @@ import os
 here = os.path.abspath(os.path.dirname(__file__))
 os.environ["OPENMS_DATA_PATH"] = os.path.join(here, "share/OpenMS")
 
-import sys
-if sys.platform.startswith("linux"):
-    # load local shared libries before we import pyopenms.so, else
-    # those are not found. setting LD_LIBRARY_PATH does not work,
-    # see: http://stackoverflow.com/questions/1178094
-    import ctypes
-    ctypes.cdll.LoadLibrary(os.path.join(here, "libOpenSwathAlgo.so"))
-    ctypes.cdll.LoadLibrary(os.path.join(here, "libOpenMS.so"))
-    ctypes.cdll.LoadLibrary(os.path.join(here, "libSuperHirn.so"))
-
 try:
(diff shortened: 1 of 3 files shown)


jpfeuffer:
I recognized that with qt_qmake_executable too. I think I changed the docu accordingly in my PR.

jpfeuffer:
What incompatibility happened with xerces?

jpfeuffer:
Ah I think you have to add 3.2 to the list of possible library names in the findxerces module. Should be included in develop and my QT5 packaging pr by now.

hroest:
> I recognized that with qt_qmake_executable too. I think I changed the docu accordingly in my PR.

nice, thanks!

> Ah I think you have to add 3.2 to the list of possible library names in the findxerces module. Should be included in develop and my QT5 packaging pr by now.

ok I have to see what we have to do on the Python side. Right now I just made a symlink :-)

hroest:
@jpfeuffer can we merge here? now that qt5 is in develop, we will need this at some point to build wheels

jpfeuffer:
Switched target branches and updated in #3324

PR #3247, opened 2018-02-13
[FIX] MRMTransitionGroupPicker: Add background subtraction to precursors chromatograms

pasqoo:
In _MRMTransitionGroupPicker_, background subtraction on precursors chromatograms is currently missing.


pasqoo:
Looks like the stopwatch is the only failing test within the apple config.

hroest:
we should probably have a test for the smoothed data as well since I just found a bug here (not by you...)

PR #3249, opened 2018-02-15
Search engine adapter fixes

cbielow:
[FIX] use Carbamidomethylation (fixed), Oxidation (variable) as default modifications for all relevant adapters.

[FIX] annotate default XTandem config in /SHARE/ of what values are actually defaults. Also switch off refinement by default.

[FIX] FileInfo now considers N/C-terminal peptide modifications in its summary

[FEATURE] sort modifications by case-INSENSITIVE names (much easier to find)


cbielow:
StopWatch_test failed on Windows and Travis-clang timed out (even after restart).

any comments?

src/topp/FileInfo.cpp
```diff
@@ -596,7 +596,7 @@ class TOPPFileInfo : public TOPPBase
             for (Size ia = 0; ia < aa.size(); ++ia)
             {
               if (aa[ia].isModified())
-                ++mod_counts[aa[ia].getModificationName()];
+                ++mod_counts[aa[ia].getModificationName() + "@" + aa[ia].getOneLetterCode()];
```
timosachsenberg: getFullName besser?

cbielow:
merge?

timosachsenberg:
In principle, I am for it. 
I am just wondering about the N-term Acetylation. I think we want Protein N-term Acetylation. but currently only support peptide N-term. Smells like trouble ... ;)

timosachsenberg:
+1 if passes

PR #3251, opened 2018-02-16
[FIX] PeakIntegrator: Positions at peak's height percentages

pasqoo:
**Where**
_PeakIntegrator::calculatePeakShapeMetrics()_

**Why**
Shamelessly copying @dmccloskey description :smile: :
The current algorithm to calculate the peak shape metrics assumes an ideal peak whereby 5%, 10%, and 50% of the peak height actually exist on both sides of the peak. If the 5%, 10%, and 50% of the peak height cannot be found on either side, erroneous values are calculated.

**How**
I implemented a `protected` function template:
```c++
template <typename PeakContainerConstIteratorT>
double findPosAtPeakHeightPercent_(
  PeakContainerConstIteratorT it_begin,
  PeakContainerConstIteratorT it_end,
  const double peak_height,
  const double percent,
  const bool is_left_half
) const;
```
Documentation is also provided.

**Notes**
The proposed solution assumes a convex peak.


src/tests/class_tests/openms/source/PeakIntegrator_test.cpp
```diff
@@ -561,22 +592,44 @@ START_SECTION(PeakShapeMetrics calculatePeakShapeMetrics(
...
-  TEST_REAL_SIMILAR(psm.width_at_5, 0.231263425125414)
-  TEST_REAL_SIMILAR(psm.width_at_10, 0.134762234301732)
-  TEST_REAL_SIMILAR(psm.width_at_50, 0.0595791540757924)
-  TEST_REAL_SIMILAR(psm.start_position_at_5, 2.51268515480125)
-  TEST_REAL_SIMILAR(psm.start_position_at_10, 2.63222565817823)
-  TEST_REAL_SIMILAR(psm.start_position_at_50, 2.65391757114759)
-  TEST_REAL_SIMILAR(psm.end_position_at_5, 2.74394857992666)
-  TEST_REAL_SIMILAR(psm.end_position_at_10, 2.76698789247996)
-  TEST_REAL_SIMILAR(psm.end_position_at_50, 2.71349672522338)
+  TEST_REAL_SIMILAR(psm.width_at_5, 0.15955)
```
hroest: why are the values now so different? is this expected?

pasqoo: Yes, because we are now taking the point's position __before__ the percentage's goal. Not the one closest.
The one to the left in case of the left half, and the one to the right in case of the right half.

PR #3253, opened 2018-02-16
Feature/sprintf fix

hroest:
use boost::format instead of sprintf


src/topp/OpenSwathFeatureXMLToTSV.cpp
```diff
@@ -291,19 +290,17 @@ void write_out_body_(std::ostream &os, Feature *feature_it, TargetedExperiment &
...
+      String mz_str = boost::str(boost::format("%f") % sub_it->getMZ());
       String apex = "NA";
       if (sub_it->metaValueExists("peak_apex_int"))
       {
-        sprintf(intensity_apex_char, "%f", (double)sub_it->getMetaValue("peak_apex_int"));
-        apex = (String) intensity_apex_char;
+        apex = boost::str(boost::format("%f") % sub_it->getMetaValue("peak_apex_int"));
       }
-      os << line << meta_values << (String)intensity_char << "\t" << apex << "\t" << (String)sub_it->getMetaValue("native_id") << "\t" << (String)mz_char << std::endl;
+      os << line << meta_values << intensity_area << "\t" << apex << "\t" << (String)sub_it->getMetaValue("native_id") << "\t" << mz_str << std::endl;
```
timosachsenberg: maybei its sufficient to use
os.imbue(std::locale("En_US"));
and directly write out floats / ints

timosachsenberg:
Ok that's interesting. So the C++ locale does not have the problem? Fine with me if that works

hroest:
no 40 chars is not really important ...

PR #3258, opened 2018-02-21
[FEATURE] PeptideIndexer supports mismatches again

cbielow:
... it slows down the annotation considerably (as it was before), but the default (0) has no impact on performance.

new flag for PeptideIndexer `-mismatches_max <n>`

This complements #2898 (where mismatch support was temporarily dropped)


src/tests/class_tests/openms/source/AhoCorasickAmbiguous_test.cpp
```diff
@@ -245,9 +242,87 @@ START_SECTION(bool findNext(const FuzzyACPattern& pattern))
...
+  {
+    observed.push_back(String(pep_db[fuzzyAC.getHitDBIndex()].data_begin, pep_db[fuzzyAC.getHitDBIndex()].data_end) + "@" + fuzzyAC.getHitProteinPosition());
+  }
+  compareHits(__LINE__, prot, expected, observed);
+  ///
+  /// with prefix and B 
+  ///
+  AhoCorasickAmbiguous::initPattern(pep_db, 1, 2, pattern);
+  observed.clear();
+  prot = "aMMB"; // B = D|N,  Z = E|Q
```
timosachsenberg: remove ,Z = E|Q here?

timosachsenberg:
looks well covered by tests (I have to admit that I did not check every step of the algorithm)

timosachsenberg:
+1 if passes

PR #3262, opened 2018-02-22
Precompiled

cbielow:
Using Precompiled Headers (PCH) is supported by all compilers.
The cotire module (https://github.com/sakra/cotire/blob/master/MANUAL.md) for CMake allows easy incorporation of PCHs.

Compile time for OpenMS + OpenMS-GUI libs on VS2015
```
00:17:45  no PCH
00:16:04  with only system headers (vector, string, set, ...)  precompiled-header: 38 MB
00:13:28  with system + contrib headers; precompiled-header: 650 MB
00:09:08  with system + contrib  + some OpenMS headers; precompiled-header: 700 MB
```

So, compile time can be halved using precompiled headers.
Tests on Linux using G++ 7.2 did show only marginal speed improvements (0-10%), so they are deactivated by default.
For building all tests, our Jenkins takes 44 vs 58min, so still a 25% speedup (since compiling targets with single compilation units is not accelerated).

Thus, the current PR only uses PCHs for Visual Studio.

Along the way, I updated some outdated CMake commands which were outdated and thus confused cotire.

Using Cotire on anything other than libraries, but on Executables has no speed gain on any platform (as expected).

Cotire by default defines a unity_build target now for OpenMS and OpenMS_gui:
`OpenMS_unity` and `OpenMS_GUI_unity`. These builds are incredibly fast on g++ and might be worth integrating into CI (see https://github.com/sakra/cotire/blob/master/MANUAL.md#optimizing-the-build-process-for-multiple-processor-cores).
```
$ cmake -D COTIRE_MAXIMUM_NUMBER_OF_UNITY_INCLUDES=-j4 <path-to-source>
$ make -j 4
```

Drawbacks: only when touching any of the 'kernel' headers which are currently included, this will trigger a full rebuild. Workaround: change the pch to exclude the headers you are currently working on.

Any comments?
I.e. should we have a flag to completely switch it off (thus making the build system a little more complex)?


timosachsenberg:
In principle, I am all for it to include it on Win. Would like to know how well this plays with frequent switches between branches / different submodules etc. as often occurs in CI builds.

cbielow:
> Would like to know how well this plays with frequent switches between branches / different submodules etc. as often occurs in CI builds.

If you change the contrib_build directory for a given OpenMS build-tree, this changes the PCH, thus: full rebuild. Is this really done often? And even if, I'd guesstimate that most of the library is re-build anyway (since our contrib is very deeply anchored in OpenMS).
This all goes for Windows only. On Linux, nothing extraordinary will happen (just default dependency resolution).

timosachsenberg:
I think it is quite common on a CI platform. If MSVC handles this without caching errors etc. I think this might be a good addition. @jpfeuffer what do you think?

cbielow:
after inspection of compile times on TRAVIS and jenkins for non-windows, I have reverted to non-cotire, since it seems to negatively affect our CI.
If someone wants to give this another go, just revert 6c3904b91c27ce8cedf3b0b6b26dc0c94ecb5fde.

The way it is now, this PR basically just cleans up our build system.
Please merge :)

timosachsenberg:
did you try IncrediBuild in VS 2017 (I didn't)?

PR #3265, opened 2018-02-26
[FIX] Quantify and calculate ion ration by intensity

pasqoo:
_MRMFeatureFilter_ and _AbsoluteQuantitation_ methods to calculate the ratio between two components (i.e., subfeatures) is only supported for `metaValue`s. 
However, there should be an option to calculate the ratio by intensity.  The intensity (i.e., peak area) is given a dedicated attribute in the `Feature` class, and can only be accessed by `getIntensity()`.
This PR implements the functionality for intensity.


pasqoo:
jenkins appleclang error: `ERROR: Error cloning remote repo 'origin'`

PR #3267, opened 2018-02-27
[DOC,FEATURE] massive speedup for Digestion and update contrib

cbielow:
when verifying protein-peptide matches, the runtime was dominated by ProteaseDigestion::isValidProduct(),
which internally uses the very expensive `tokenize_()` to digest a full protein.

The digestion is now optimized, to only operate on the immediate vicinity of the peptide in question.
Also, there was redundant code in EnzymaticDigestion::isValidProduct(), which made maintenance really hard. Now, there is a single implementation.

Benchmark, verifiying 48k peptides (when the matching location is already known, but the trypticity needs to be determined).

before: 40 seconds
now: 3 seconds (factor x13)

Some documentation added.


jpfeuffer:
Looks great but needs conflict resolution. `git submodule update --init THIRDPARTY` might do the trick.

cbielow:
´git submodule update --remote --merge´ did work, but Im not sure it did the right thing.

jpfeuffer:
yes seemed to work. You will pull latest contrib changes, too, now.

cbielow:
I wasn't sure if these changes were already good to go.
But if thats ok, I#ll merge if tests pass.

jpfeuffer:
Should be fine. I wanted to merge them soon anyway. PRs use the latest contrib anyway (except for travis which needs a manual push to redo the cache)

cbielow:
not sure what happened to Jenkins... java error...

jpfeuffer:
If I knew... maybe IT knows.. but you can merge.

PR #3269, opened 2018-02-28
[FIX][FEATURE] Select peaks boundaries by widest peak

pasqoo:
**Description**
`findLargestPeak()` sometimes selects cutoff peaks or merged peaks.
The resulting `best_left` and `best_right` don't represent good choices for the rest of the work involved within the class.

**Solution proposal**
A method which instead selects the widest peak: `findWidestPeakIndices()`.
The user can choose the new technique by setting the parameter `"boundary_selection_method"` to `"widest"`.

**Question**
It would be useful to not rely on any consensus `best_left` or `best_right`.
We've been trying to come up with a way to bend the code toward this goal, but it's been proving difficult because a big part of the class _MRMTransitionGroupPicker_ relies on the existence of `chr_idx`, `peak_idx`, `best_left` and `best_right`.

@hroest Do you have any suggestion on this point? Or do you recommend to write a new class for such a goal?
In both cases, any advice / design outline recommendation is appreciated :smile:


hroest:
This looks good to me, I would be happy to merge this. I think the data we are working with is quite different, OpenSWATH is designed to work with (very) noisy data where you have a lot of overlapping peaks and not clean traces in every transition. Maybe your data are different? 

>  findLargestPeak() sometimes selects cutoff peaks or merged peaks.

I think this solution tends to prefer wide peaks over, does it overall increase the width of the peaks? 

>  It would be useful to not rely on any consensus best_left or best_right.

This is how the algorithm currently works: identify a place in the chromatogram that potentially has a peak and start with a left/right hypothesis. Its basically a greedy approach and as you mention, it can make mistakes that are hard to correct after the fact. On the hand it seems to be pretty robust even when confronted with noisy data. I dont think we can change that easily without changing the whole algorithm. I would be happy to talk on gitter/Skype about alternatives. We could potentially cluster peaks across transitions and then pick clusters. Maybe you can also share some examples where the current algorithm performs really badly? 

But depending on what you have in mind, the greedy approach I use may not work for what you want to do.

dmccloskey:
I have attached two examples where the algorithm cuts off a peak and over extends a peak incorrectly.  

[1280f967-bafb-48ad-8604-64a0914f19e5.pdf](https://github.com/OpenMS/OpenMS/files/1782462/1280f967-bafb-48ad-8604-64a0914f19e5.pdf)
[90976381-ba3f-4bb5-abfb-46fba00a72f4.pdf](https://github.com/OpenMS/OpenMS/files/1782464/90976381-ba3f-4bb5-abfb-46fba00a72f4.pdf)

The way in which the algorithm lines up the RT of peaks in each transition for the same transition group works great.   It is just the accuracy of the peak integration for each of the individual peaks in the transition that is problematic.

If you are able to meet over Skype, that would probably be the most ideal.  What is your time zone again?  Would you be able to meet sometime this week?

PR #3270, opened 2018-02-28
Easy indication if Absolute Quantitation found an optimal calibration or not

dmccloskey:
### Description
It not straight forward to determine based on the output of Absolute Quantitation if an optimal calibration curve was found.

### Fix
- Added a boolean output to `OptimizeCalibrationCurvesIterative` to indicate if an optimal fit was found or not
- Set the `n_points` and `correlation_coefficient` to default values of 0 if an optimal fit was not found to clearly indicate that an optimal was not found
- Updated the transformation model params only if an optimal fit was found.  Otherwise, the user provided params (if any) are returned)

### DIFF
--- a/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
+++ b/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
@@ -409,7 +409,6 @@ namespace OpenMS
     Param & optimized_params)
   {
 
-    bool optimal_calibration_found = false; 
     // sort from min to max concentration
     std::vector<AbsoluteQuantitationStandards::featureConcentration> component_concentrations_sorted = component_concentrations;
     std::sort(component_concentrations_sorted.begin(), component_concentrations_sorted.end(),
--- a/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
+++ b/src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
@@ -441,7 +441,7 @@ namespace OpenMS
       if (component_concentrations_sorted_indices.size() < min_points_)
       {
         LOG_INFO << "No optimal calibration found for " << component_concentrations_sub[0].feature.getMetaValue("native_id") << " .";
-        break;
+        return optimal_calibration_found;
       }
 
       // fit the model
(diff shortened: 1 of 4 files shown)


src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -440,7 +441,7 @@ namespace OpenMS
       if (component_concentrations_sorted_indices.size() < min_points_)
       {
         LOG_INFO << "No optimal calibration found for " << component_concentrations_sub[0].feature.getMetaValue("native_id") << " .";
-        break;
+        return optimal_calibration_found;
```
timosachsenberg: maybe replace with:
''' return false; // no optimal calibration found '''

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -475,7 +476,8 @@ namespace OpenMS
 
         // copy over the final optimized points before exiting
         component_concentrations = component_concentrations_sub;
-        break;
+        optimal_calibration_found = true;
+        return optimal_calibration_found;
```
timosachsenberg: return true; // optimal calibration found

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -512,9 +514,10 @@ namespace OpenMS
       }
       else
       {
-        break;
+        return optimal_calibration_found;
```
timosachsenberg: return false; // no optimal calibration found

src/openms/source/ANALYSIS/QUANTITATION/AbsoluteQuantitation.cpp
```diff
@@ -512,9 +514,10 @@ namespace OpenMS
...
       }
     }
+    return optimal_calibration_found;
```
timosachsenberg: return false; // no optimal calibration found

hope I got it right but it seemed to me that the variable is not needed and thus can be removed

dmccloskey:
Looks like the "TOPP_InspectAdapter_2" failed on windows.  Not sure what caused that.

dmccloskey:
Are there any suggestions on what todo about the windows failure?  Would it be worth just to try and re-running the check to see if it compiles and all tests pass?

jpfeuffer:
We upgraded our CI server and have to update the slaves accordingly but I am currently on a conference. If you can test it on windows somewhere I would be fine with merging without that check.

PR #3272, opened 2018-02-28
calculateRatio fails when feature_name = intensity and there is no IS

dmccloskey:
### Fix
- added additional conditionals to calculateRatio and calculateIonRatio methods
- added tests to test for the case of feature_name = intensity and no IS

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
@@ -388,7 +388,7 @@ namespace OpenMS
     // metaValue feature_name access
     else
     {
-      else if (component_1.metaValueExists(feature_name) && component_2.metaValueExists(feature_name))
+      if (component_1.metaValueExists(feature_name) && component_2.metaValueExists(feature_name))
       {
         const double feature_1 = component_1.getMetaValue(feature_name);
         const double feature_2 = component_2.getMetaValue(feature_name);
--- a/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
@@ -368,35 +368,42 @@ namespace OpenMS
   
   double MRMFeatureFilter::calculateIonRatio(const Feature & component_1, const Feature & component_2, const String & feature_name)
   {
-    
     double ratio = 0.0;
-    if (feature_name == "intensity" && component_1.metaValueExists("native_id") && component_2.metaValueExists("native_id"))
+    // member feature_name access
+    if (feature_name == "intensity")
     {
-      const double feature_1 = component_1.getIntensity();
-      const double feature_2 = component_2.getIntensity();
-      ratio = feature_1 / feature_2;
+      if (component_1.metaValueExists("native_id") && component_2.metaValueExists("native_id"))
+      {
+        const double feature_1 = component_1.getIntensity();
+        const double feature_2 = component_2.getIntensity();
+        ratio = feature_1 / feature_2;
+      }
+      else if (component_1.metaValueExists("native_id"))
+      {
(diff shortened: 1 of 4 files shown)


src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -370,28 +370,33 @@ namespace OpenMS
...
-    else if (feature_name == "intensity")
+    if (feature_name == "intensity" && component_1.metaValueExists("native_id") && component_2.metaValueExists("native_id"))
     {
       const double feature_1 = component_1.getIntensity();
       const double feature_2 = component_2.getIntensity();
       ratio = feature_1 / feature_2;
     }
+    else if (feature_name == "intensity" && component_1.metaValueExists("native_id"))
+    {
+      LOG_DEBUG << "Warning: no IS found for component " << component_1.getMetaValue("native_id") << ".";
```
timosachsenberg: in this if branch, the omponent_2.metaValueExists("native_id") is false. Though the debug message talks about component 1. Is this correct?

timosachsenberg: Ah I get it.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -370,28 +370,33 @@ namespace OpenMS
...
+      const double feature_1 = component_1.getIntensity();
+      ratio = feature_1;
+    }
+    else if (component_1.metaValueExists(feature_name) && component_2.metaValueExists(feature_name))
```
timosachsenberg: these cases are bit difficult to follow.
Does this also cover the case were feature="intensity"? Then I think it would be wrong to request the metaValue("intensity")

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -370,28 +370,33 @@ namespace OpenMS
...
+    {
+      double feature_1 = component_1.getMetaValue(feature_name);
+      double feature_2 = component_2.getMetaValue(feature_name);
+      ratio = feature_1/feature_2;
+    }
     else if (component_1.metaValueExists(feature_name))
```
timosachsenberg: same here... what if feature_name is "intensity"

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
dmccloskey: Essentially some feature_names are stored as metaValues and others are given members in the Feature class.  Specifically, I am trying to cover the case where one wants to quantify or filter based on peak area (i.e., "intensity") which is accessed via getIntensity() or peak height which is accessed by getMetaValue("peak_apex"int) for both the case where the first and second component exists or where the first component exists but the second does not.

timosachsenberg: Ok that's how I understood it. But this will e.g. fail if feature_name = intensity but no native_id meta value exists, right?
Then  else if (component_1.metaValueExists(feature_name)) will be called with feature_name = intensity which will result in an error.
I think it would be better to distinguish two cases:
if (feature_name == "intensity")
{ // access member
}
else // feature name refers to meta value 
{ // access meta values
} 
and put the corresponding code in that parts.

dmccloskey: If there is no metavalue called "native_id" it will fail not work.  I am testing to see if the feature is actually populated with values.  Is there a way you would recommend testing to see if the feature has an intensity value?  That would avoid the need to test for the "native_id".

timosachsenberg: Intensity is initialized to 0. Guess you could error out if this is zero and feature_name = intensity

timosachsenberg:
thanks! now it's easier to read

PR #3274, opened 2018-03-01
[FEATURE] TOPPView: peaks colored by label content

enetz:
Extension of #3007 by @hendrikweisser .

A new vector for peak colors has been added to LayerData.
Peaks are colored by the colors stored in there, if its length equals the length of the spectrum.
Currently it is filled by label content, in the same way the colors for the labels were determined.
In the future we can also use information in DataArrays or other sources to set colors for different kinds of data.

Screenshot (for Cross-Linking MS data):
peak_colors

### DIFF
--- a/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
+++ b/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
@@ -935,24 +935,30 @@ namespace OpenMS
                               ann_spectrum[i].getIntensity());
         const String& label = labels[i];
         QColor color;
+        QColor peak_color;
+        LayerData& annotated_layer = current_canvas->getCurrentLayer();
         // XL-MS specific coloring of the labels, green for linear fragments and red for cross-linked fragments
         if (label.hasSubstring("[alpha|") || label.hasSubstring("[beta|"))
         {
           if (label.hasSubstring("|ci$"))
           {
             color = Qt::darkGreen;
+            peak_color = Qt::green;
           }
           else if (label.hasSubstring("|xi$"))
           {
             color = Qt::darkRed;
+            peak_color = Qt::red;
           }
         }
         else // different colors for left/right fragments (e.g. b/y ions)
         {
           color = (label.at(0) < 'n') ? Qt::darkRed : Qt::darkGreen;
+          peak_color = (label.at(0) < 'n') ? Qt::red : Qt::green;
         }
 
         Annotation1DItem* item = new Annotation1DPeakItem(position, label.toQString(), color);
+        annotated_layer.peak_colors_1d.push_back(peak_color);
         item->setSelected(false);
         tv_->getActive1DWidget()->canvas()->getCurrentLayer().getCurrentAnnotations().push_front(item);
       }
--- a/src/openms_gui/include/OpenMS/VISUAL/LayerData.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/LayerData.h
@@ -308,6 +308,9 @@ public:
     /// Annotations of all spectra of the experiment (1D view)
     std::vector<Annotations1DContainer> annotations_1d;
 
+    /// Peak colors of the currently shown spectrum
+    std::vector<QColor> peak_colors_1d;
+
     /// Flag that indicates if the layer data can be modified (so far used for features only)
     bool modifiable;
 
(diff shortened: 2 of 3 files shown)


timosachsenberg:
nice. did you also find out why moving the labels does not work?

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
```diff
@@ -935,24 +935,30 @@ namespace OpenMS
...
           }
         }
         else // different colors for left/right fragments (e.g. b/y ions)
         {
           color = (label.at(0) < 'n') ? Qt::darkRed : Qt::darkGreen;
+          peak_color = (label.at(0) < 'n') ? Qt::red : Qt::green;
         }
 
         Annotation1DItem* item = new Annotation1DPeakItem(position, label.toQString(), color);
+        annotated_layer.peak_colors_1d.push_back(peak_color);
```
timosachsenberg: where are these deleted?

enetz: This vector gets filled in the new LayerData object containing only the labeled peaks and labels,
this is then drawn on top of the normal spectrum.
We should delete this whole layer after changing the spectrum / ID.
`deactivate1DSpectrum` removes the annotations from the current layer, but I think a new layer with a new spectrum (and now also a filled color vector) gets created every time a new ID is selected and these do not get deleted.
@hendrikweisser can you confirm that, or is there a mechanism in place to get rid of these additional layers and spectra? Then the color vectors should be erased with them.
I will look into this later.

hendrikweisser: @enetz: I think you're right, `deactivate1DSpectrum` and `removeTheoreticalSpectrumLayer_` (which is called by `deactivate...`) should perform the clean-up. But I'm not sure if they do everything correctly or need updating.

hendrikweisser:
Nice work, Eugen!
I'm also interested in the peptide coverage representation in the top right of your screenshot. This is very cool and I didn't know we could do this. Does this also work for non-crosslinked peptides (i.e. "normal" peptide IDs)? How can I get something like this for my RNAs?

enetz:
A new issue popped up while I was working on this PR:
the text labels can not be moved anymore.
This was the case already before I started with the peak colors, so it should be something from before this PR. I have not found the reason yet, will look into it.

enetz:
@hendrikweisser You can look into the function `TOPPViewIdentificationViewBehavior::extractCoverageStrings` that makes the colored arrow symbol Strings for each peptide and how it is used.
The code does not look nice, each of those green arrows is: `<font color=\"green\">&#8636;</font>`, stacked into long strings (`&#8636;` is an arrow).
Basically it is a text box with 5 lines of text and the right numbers of spaces to keep them stacked on top of each other correctly:
alpha arrows
alpha peptide
linker bar "...   |   ..."
beta peptide
beta arrows

It is currently tailored specifically for cross-link data and only done when some cross-linking specific meta values exist in the ID info, but implementing that for normal peptides should be much easier.
In general you just have to create this multi-line String with arrows and the sequence, e.g. `box_text`, and call
`widget_1D->canvas()->setTextBox(box_text.toQString());`

hendrikweisser:
@enetz: Thanks for the explanation! What do the colours of the arrows (red/green/blue) mean?

enetz:
green and red are the same as the peaks, green = linear peptide fragments from one peptide, red = cross-linked fragments containing parts from both peptides and the cross-linker.
Each fragmentation event would create one of both types, so if both fragments were found the blue double arrow is shown.
Since I am using strings like this, the arrows are not between residues for fragmentation sites, but rather above or below a residue indicating a fragment containing this residue and the rest of the sequence in the direction of the arrow. So a blue double arrow does not mean both fragments from one fragmentation site but rather two fragments from the fragmentation sites exactly left and right of this residue (both fragments containing and starting at this residue).

hendrikweisser:
@enetz: I don't know if you're already aware, but I wanted to point out that your usage of the arrows in the coverage annotation is a bit non-standard (and may be confusing for people who are used to the standard notation). See this example for what I mean by "standard notation": https://www.researchgate.net/profile/Ulrich_Goepfert/publication/229083272/figure/fig3/AS:300749193007108@1448715603476/Fragment-ion-spectra-of-the-HC-T13-peptide-originating-from-the-spiking-antibody-B.png
So in this notation you have "angles" on top and below the sequence corresponding to the b/y ions. Your arrows resemble these angles, but they're read in the opposite direction. For example, for the top-left red arrow in your screenshot (over the second "L" of "LEL"), the "standard" interpretation would be that it denotes a b ion with sequence "LEL". I think you could just reverse the direction of your single arrows to move to the "standard" notation.

timosachsenberg:
looks good! let's merge.

PR #3275, opened 2018-03-01
[FIX] Find Xerces module and pyOpenMS CMakeLists

jpfeuffer:
- Makes finding in pyOpenMS more generic
- Let's FindXerces prefer contrib
- Removes potentially harmful non-suffixed name from debug libraries.


hroest:
looks good, :+1:

cmake/modules/FindXercesC.cmake
```diff
@@ -54,8 +54,10 @@ find_path(
 
 # Find the xerces libraries
 if (NOT XercesC_LIBRARIES)
-    find_library(XercesC_LIBRARY_RELEASE NAMES xerces-c xerces-c_3 xerces-c_3_1 xerces-c-3.1 xerces-c_3_2 xerces-c-3.2 ${_XercesC_PATHS} PATH_SUFFIXES lib)
-    find_library(XercesC_LIBRARY_DEBUG NAMES xerces-c xerces-c_3D xerces-c_3_1D xerces-c-3.1D xerces-c_3_2D xerces-c-3.2D ${_XercesC_PATHS} PATH_SUFFIXES lib)
+    ## The NAMES_PER_DIR option will make sure that the PATHS are the "outer for loop" when searching for the libraries.
+	## We want that because we put the contrib as the first search path usually.
+    find_library(XercesC_LIBRARY_RELEASE NAMES xerces-c xerces-c_3 xerces-c_3_1 xerces-c-3.1 xerces-c_3_2 xerces-c-3.2 NAMES_PER_DIR ${_XercesC_PATHS} PATH_SUFFIXES lib)
```
timosachsenberg: still look for the old versions?

jpfeuffer: Yes otherwise we exclude most system contribs right?

PR #3276, opened 2018-03-02
Missing IS feature causes some sort of memory fault

dmccloskey:
### Description
When the IS for a feature is missing, but expected in the AbsoluteQuantitationMethod, strange behavior most likely resulting from some sort of memory fault occurs.  This strange behavior includes generating all 0s for each feature/IS ratio for seemingly random samples for the same component in what otherwise should be valid ratios.  

### Workaround
Filtering out all components that are missing an IS prior to calling OptimizeCalibrationCurves

### Fix (so as to not have to use the workaround)
- When an IS is expected, but not found during the call to `AbsoluteQuantitationStandards::mapComponentsToConcentrations`, the component is excluded.
- A check before weighting the data and calculating the pearson correlation coefficient is done so as to avoid an "invalidRange" error.

### DIFF
--- a/src/openms/source/METADATA/AbsoluteQuantitationStandards.cpp
+++ b/src/openms/source/METADATA/AbsoluteQuantitationStandards.cpp
@@ -84,7 +84,7 @@ namespace OpenMS
         }
         if (run.IS_component_name != "")
         {
-          if (!findComponentFeature_(fmap, run.IS_component_name, fc.IS_feature)
+          if (!findComponentFeature_(fmap, run.IS_component_name, fc.IS_feature))
           { // down stream methods will expect to find an IS
             continue;
           } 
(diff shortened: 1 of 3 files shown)


src/openms/source/METADATA/AbsoluteQuantitationStandards.cpp
```diff
@@ -84,7 +84,10 @@ namespace OpenMS
         }
         if (run.IS_component_name != "")
         {
-          findComponentFeature_(fmap, run.IS_component_name, fc.IS_feature);
+          if (!findComponentFeature_(fmap, run.IS_component_name, fc.IS_feature))
+          { // down stream methods will expect to find an IS
+            continue;
```
timosachsenberg: should a warning or some debug message be written?

dmccloskey: Good idea.

hroest:
@dmccloskey any update on this?

PR #3277, opened 2018-03-03
Feature/pepxml

### LINKED ISSUE #3268: PepXML Reader should read more PeptideProphet scores
... and store them as Metavalues so they could be switched with IDScoreSwitcher.

https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/PepXMLFile.cpp#L1062

hroest:
fixes #3268 

- parses peptideprophet score summaries and stores them as meta-values

### DIFF
--- a/src/openms/source/FORMAT/IdXMLFile.cpp
+++ b/src/openms/source/FORMAT/IdXMLFile.cpp
@@ -355,6 +355,26 @@ namespace OpenMS
           os << " >\n";
           writeFragmentAnnotations_("UserParam", os, p_hit.getPeakAnnotations(), 4);
           writeUserParam_("UserParam", os, p_hit, 4);
+
+          // write out the (optional) peptide prophet / interprophet results as UserParams
+          {
+            int k = 0;
+            for (std::vector<PeptideHit::PepXMLAnalysisResult>::const_iterator ar_it = p_hit.getAnalysisResults().begin();
+                ar_it != p_hit.getAnalysisResults().end(); ++ar_it, ++k)
+            {
+              os << "\t\t\t\t<UserParam type=\"string\" name=\"_ar_" << k << "_score_type\" value=\"" << ar_it->score_type << "\"/>" << "\n";
+              os << "\t\t\t\t<UserParam type=\"float\" name=\"_ar_" << k << "_score\" value=\"" << ar_it->main_score << "\"/>" << "\n";
+              if (!ar_it->sub_scores.empty())
+              {
+                for (std::map<String, double>::const_iterator subscore_it = ar_it->sub_scores.begin();
+                    subscore_it != ar_it->sub_scores.end(); ++subscore_it)
+                {
+                  os << "\t\t\t\t<UserParam type=\"float\" name=\"_ar_" << k << "_subscore_" << subscore_it->first <<"\" value=\"" << subscore_it->second << "\"/>" << "\n";
+                }
+              }
+            }
+
+          }
           os << "\t\t\t</PeptideHit>\n";
         }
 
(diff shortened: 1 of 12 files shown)


jpfeuffer:
Looks good. Failures unrelated.

jpfeuffer:
Are the additional search_scores   p-value and expect also parsed? Could find it after a quick look.
It seems to be on the same level as peptideprophet result.

src/tests/topp/CMakeLists.txt
```diff
@@ -819,6 +819,14 @@ set_tests_properties("TOPP_IDFileConverter_22_out1" PROPERTIES DEPENDS "TOPP_IDF
...
 add_test("TOPP_IDFileConverter_23_out1" ${DIFF} -in1 IDFileConverter_23_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_23_output.idXML -whitelist "IdentificationRun date" )
 set_tests_properties("TOPP_IDFileConverter_23_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_23")
+# Comet pep.xml
+add_test("TOPP_IDFileConverter_24" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/IDFileConverter_24_input.pep.xml -out IDFileConverter_24_output.tmp -out_type idXML)
+add_test("TOPP_IDFileConverter_24_out1" ${DIFF} -in1 IDFileConverter_24_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_24_output.idXML -whitelist "IdentificationRun date" )
+set_tests_properties("TOPP_IDFileConverter_24_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_24")
+# Comet pep.xml
+add_test("TOPP_IDFileConverter_25" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/IDFileConverter_25_input.idXML -out IDFileConverter_25_output.tmp -out_type pepXML)
+add_test("TOPP_IDFileConverter_25_out1" ${DIFF} -in1 IDFileConverter_25_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_25_output.pep.xml)
+set_tests_properties("TOPP_IDFileConverter_25_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_25")
```
hroest: @jpfeuffer here we have the complete conversion of a pep.xml file to idxml and back again (note that currently the output of 24 is identical to the input of 25 and the output of 25 is very similar to the input of 24). Does this answer your question?

hroest:
> Are the additional search_scores p-value and expect also parsed? Could not find it after a quick look.
It seems to be on the same level as peptideprophet result.

you mean the probability? yes it should also get parsed

src/openms/source/FORMAT/IdXMLFile.cpp
```diff
@@ -728,6 +748,32 @@ namespace OpenMS
...
+        {
+          if (!current_analysis_result_.score_type.empty())
+          {
+            pep_hit_.addAnalysisResults(current_analysis_result_);
+          }
+          current_analysis_result_.score_type = attributeAsString_(attributes, "value");
+        }
+        else if (val_name == "score")
+        {
+          current_analysis_result_.main_score = attributeAsDouble_(attributes, "value");
```
hroest: @jpfeuffer this is usually the probability, did you mean this with p-value ?

src/tests/topp/IDFileConverter_25_output.pep.xml
```diff
@@ -0,0 +1,92 @@
...
+					<search_score_summary>
+						<parameter name="fval" value="-2.3431"/>
+						<parameter name="isomassd" value="0"/>
+						<parameter name="massd" value="-0.087"/>
+						<parameter name="nmc" value="0"/>
+						<parameter name="ntt" value="0"/>
+					</search_score_summary>
+				</peptideprophet_result>
+			</analysis_result>
+			<search_score name="xcorr" value="0.998"/>
```
jpfeuffer: Sorry, I rather meant scores from the underlying search engine. Are they parsed? And what do you think about wrapping this into an option(s)? Could bloat the default idXMLs otherwise.

hroest:
> Sorry, I rather meant scores from the underlying search engine. Are they parsed? 

Yes, they are already parsed right now, that is not changing

> And what do you think about wrapping this into an option(s)? Could bloat the default idXMLs otherwise.

sure we can do that, even though running peptideprophet on the pep.xml files is not a standard workflow AFAIK

jpfeuffer:
Ah ok well no strong preference. I agree it is not a standard use case. Feel free to merge.

timosachsenberg:
StopWatch and InspectAdapter seem to be unrelated

PR #3279, opened 2018-03-05
[FIX?] Tries to fix the include mess

jpfeuffer:
By adding contrib includes first

### DIFF
--- a/src/openms/CMakeLists.txt
+++ b/src/openms/CMakeLists.txt
@@ -116,6 +116,26 @@ if (OPENMS_CONTRIB_LIBS)
  set (OPENMS_CONTRIB_INCLUDE_DIRS ${OPENMS_CONTRIB_LIBS}/include)
 endif()
 
+foreach (includedir
+	     ${SEQAN_INCLUDE_DIRS}
+	     ${SQLITE_INCLUDE_DIR}
+	     ${LIBSVM_INCLUDE_DIRS}
+	     ${EIGEN3_INCLUDE_DIR}
+	     ${WM5_INCLUDE_DIRS}
+	     ${CRAWDAD_INCLUDE_DIRS}
+	     ${COIN_INCLUDE_DIRS}
+	     ${GLPK_INCLUDE_DIRS}
+	     ${ZLIB_INCLUDE_DIRS}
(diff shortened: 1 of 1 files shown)


src/openms/CMakeLists.txt
```diff
@@ -116,29 +116,39 @@ if (OPENMS_CONTRIB_LIBS)
...
 #set(OpenMS_sources ${OpenMS_sources} openms_pch.cpp)
 openms_add_library(TARGET_NAME  OpenMS
                    SOURCE_FILES  ${OpenMS_sources}
                    HEADER_FILES  ${OpenMS_sources_h}
                                  ${OpenMS_configured_headers}
                    INTERNAL_INCLUDES ${CMAKE_CURRENT_SOURCE_DIR}/include
                                      ${CMAKE_CURRENT_BINARY_DIR}/include
-		               PRIVATE_INCLUDES ${EOL_BSPLINE_INCLUDE_DIRECTORY}
+		   PRIVATE_INCLUDES ${EOL_BSPLINE_INCLUDE_DIRECTORY}
                    EXTERNAL_INCLUDES ${OPENMS_CONTRIB_INCLUDE_DIRS}
```
hroest: why isnt having OPENMS_CONTRIB_INCLUDE_DIRS here sufficient? are the `xxx_INCLUDE_DIRS` different? I expect them all to be `/path/to/contrib/include`

hroest:
hmm it sounds like the right way to go, so :+1: from me

src/openms/CMakeLists.txt
jpfeuffer: Unfortunately not. That is the reason for the trouble. Sometimes libraries have a subfolder with their name (like boost) sometimes they don't. If we just don't care about system libs we can prefix our include statements with the subfolders and only have one include. But this won't work for system includes which sometimes do the prefixing and sometimes don't.

jpfeuffer: Actually I think it would be better to have different include dirs for every library because then you can have the system/contrib choice on a per Library basis. But then we need to adapt all find modules and even write own ones for the missing oned.

jpfeuffer: Missing in the sense of "currently using the default one from CMake"

PR #3282, opened 2018-03-06
include-what-you-use

cbielow:
After observing potentially large compile time savings in VS using precompiled headers (#3262),
this PR uses Timo's IWYU branch, fixes some problems for Windows and happily reports compile time savings for building OpenMS + TOPP from
17:45 before
15:10 after

If compile time is not negatively affected on other platforms, this gain would be free.
It also cleans up the include mess in some headers.


jpfeuffer:
Sounds great. There were some changes in Jenkins which I am fixing ASAP. Don't be surprised about missing Jenkins builds right now. I will rebuild here when I am finished.
Maybe the savings are even higher on machines with less cores? Let's hope.

timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=260428

cbielow:
fun fact: Codacy does not check non-template function definitions in headers (where they don't belong in the first place, but hey). Moving the code to a .cpp file, triggers the check.

cbielow:
Ok. Looks good.

hroest:
> If compile time is not negatively affected on other platforms, this gain would be free.

pretty sure that works the same way on all other platforms. Should speed up compilation somewhat everywhere (I hope)

hendrikweisser:
@cbielow: While merging these changes into a branch I noticed that you introduced spaces before "&" in type signatures of references (e. g. `const MSChromatogram & chromatogram`). Can you please configure your editor to not do that? It would save me from considerable mental anguish (and having to remove those spaces all the time).

hendrikweisser:
It appears that there is no longer a working overload for `const MSSpectrum& MSExperiment::getSpectrum(Size) const`. This used to work before the inline implementation for `getSpectrum` was removed (although there was no explicit overload for the `const` case). I assume it's the same for `getChromatogram`. We should probably add those overloads.

hroest:
> We should probably add those overloads.

and tests

PR #3288, opened 2018-03-08
FileInfo and Comet optimizations

cbielow:
[FIX] FileInfo's FASTA summary uses hashes to detect duplicates (old: quadratic runtime; new linear at the cost of 10% more memory; reporting of duplicate and AAA stats)

[FIX] minor changes to CometAdapter: remove excess includes; check for comet.exe before expensive mzML loading;

[FEATURE] add support for file rename in File.cpp (avoids inclusion of QDir in tools)

### DIFF
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -116,6 +116,11 @@ namespace OpenMS
     return !fi.exists() || fi.size() == 0;
   }
 
+  bool File::rename(const String& from, const String& to)
+  {
+    return QFile::rename(from.toQString(), to.toQString());
+  }
+
   bool File::remove(const String& file)
   {
     if (!exists(file))
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -39,6 +39,7 @@
 #include <OpenMS/CONCEPT/LogStream.h>
 
 #include <OpenMS/DATASTRUCTURES/DateTime.h>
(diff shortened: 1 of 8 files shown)


timosachsenberg:
How is sequence duplication defined? Also on ambiguous Aas?

cbielow:
> How is sequence duplication defined? Also on ambiguous Aas?

As exact string inequality.

cbielow:
ok, looks good

src/openms/source/SYSTEM/File.cpp
```diff
@@ -116,6 +117,23 @@ namespace OpenMS
     return !fi.exists() || fi.size() == 0;
   }
 
+  bool File::rename(const String& from, const String& to, bool overwrite_existing, bool verbose)
+  {
+    // existing file? Qt won't overwrite, so try to remove it:
+    if (overwrite_existing && exists(to) && !remove(to))
```
timosachsenberg: I think you should check if from == to

cbielow: mhh... that opens quite a can of worms...
It could be relative filenames, symlinks etc... which point to the same file. Checking `from == to` is only to tip of the iceberg

cbielow: http://doc.qt.io/archives/qt-4.8/qfileinfo.html#canonicalFilePath could solve it...

timosachsenberg: hmm otherwise you might easily remove a file that you just wanted to rename...

timosachsenberg:
+1 if passes

PR #3289, opened 2018-03-08
rename param in MSGFPlus adapter conforming with MS-GF+ API

lars20070:
Change of protocol `none` to `automatic` as used in MS-GF+ API.

### DIFF
--- a/src/topp/MSGFPlusAdapter.cpp
+++ b/src/topp/MSGFPlusAdapter.cpp
@@ -115,7 +115,7 @@ public:
     // parameter choices (the order of the values must be the same as in the MS-GF+ parameters!):
     fragment_methods_(ListUtils::create<String>("from_spectrum,CID,ETD,HCD")),
     instruments_(ListUtils::create<String>("low_res,high_res,TOF,Q_Exactive")),
-    protocols_(ListUtils::create<String>("automatic,phospho,iTRAQ,iTRAQ_phospho,TMT")),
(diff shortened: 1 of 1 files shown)


timosachsenberg:
This is a breaking change so we probably need to discuss that.
-p ProtocolID (Default: 0)
Protocol identifier. Protocols are used to enable scoring parameters for enriched and/or labeled samples.
0: No protocol (Default)

Looking at the documentation at https://omics.pnl.gov/software/ms-gf, the API is to only provide an integer value.
At the top it is called automatic
at the detailed description at the bottom it is called no protocol

So not sure if it is worth to change that

lars20070:
For example, for TMT one can work with `protocol = automatic`. Choosing `protocol = none` for TMT I found not intuitive.
https://gitter.im/lars20070/FreiburgOpenMS?at=5a9e87f9c3c5f8b90d3c6986

timosachsenberg:
if you look at: https://bix-lab.ucsd.edu/pages/viewpage.action?pageId=13533355
it seems our name just matches the description in the MSGFPlus documentation. 
So I guess they changed 0=auto later.
What about a non-breaking change? adding auto as default but also keeping "none"

lars20070:
Good solution.

@klemens1 Do you agree? Then I make the changes.

faceless123454:
sounds good

lars20070:
Yep. The MS-GF+ docu changed from `0: none` (2013) to `0: automatic` (2017).

PR #3292, opened 2018-03-08
[FEATURE] ScanNumber to SiriusAdapter output based on Accession in SourceFile

oliveralka:
[1]  SourceFile add native_id_type_accession
[2] SpectrumLookUp extractScanNumber based on native_id & native_id_type_accesion (Table 1 used as reference [table1] )
[3] ScanNumber was extracted and used additional to scan_index 

If a MzML-File is chunked the scan_index starts from 0 in each chunk. In contrast the ScanNumber is unique for the spectrum of a MzML-File. Here this information is used for unique identification of a compound spectrum. If only one MzML-File is used without merging the output documents the scan index can still be used for reference (e.g. to look at a spectrum in TOPPView).

[table1] https://github.com/HUPO-PSI/mzTab/blob/master/specification_document/1_1_draft_specs/mzTab_format_specification_1_1-M_draft.adoc#622-mztab-id


src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -77,6 +77,10 @@ namespace OpenMS
     // determine type of spectral data (profile or centroided) - only checking first spectrum (could be ms2 spectrum)
     SpectrumSettings::SpectrumType spectrum_type = spectra[0].getType();
 
+    // extract native id type accession (e.g. MS:1000768) corresponding to native id type (Thermo nativeID format)
+    const String native_id_type_accession = spectra.getExperimentalSettings().getSourceFiles()[0].getNativeIDTypeAccession();
+    std::cout << "native_id_type_accession: " << native_id_type_accession << std::endl;
```
timosachsenberg: change to LOG_DEBUG?

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -249,6 +249,67 @@ namespace OpenMS
     return -1;
   }
 
+  Int SpectrumLookup::extractScanNumber(const String& native_id,
+                                        const String& native_id_type_accession)
+  {
+    // check accession for data type to extract (e.g. MS:1000768 - Thermo nativeID format - scan=xsd:positiveInteger)
+    
+    boost::regex regexp;
+    std::vector<String> scan = {"MS:1000768","MS:1000769","MS:1000771","MS:1000772","MS:1000776"};
```
timosachsenberg: // list of CV accessions with native id format: "scan=NUMBER"

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -249,6 +249,67 @@ namespace OpenMS
...
+    std::vector<String> file = {"MS:1000773","MS:1000775"};
```
timosachsenberg: comment

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -249,6 +249,67 @@ namespace OpenMS
...
+      boost::smatch match;
+      bool found = boost::regex_search(native_id, match, regexp);
+      if (found && match["GROUP"].matched)
+      { 
+        try
+        {
+          String value = match["GROUP"].str();
+          return value.toInt();
+        }
+        catch (Exception::ConversionError&){}
```
timosachsenberg: this catch block ignores if a ConversionError occurred. I think this is fine but I would return -1 in the catch block

timosachsenberg:
+1 if passes

jpfeuffer:
Ok will be tested in the nightlies.

PR #3295, opened 2018-03-09
[REFACTOR] moved some tool methods to algorithm

timosachsenberg:


src/openms/include/OpenMS/MATH/STATISTICS/PosteriorErrorProbabilityModel.h
```diff
@@ -125,13 +150,13 @@ namespace OpenMS
       }
 
       ///computes the gaussian density at position x with parameters params.
-      double getGauss(double x, const GaussFitter::GaussFitResult & params)
+      static double getGauss_(double x, const GaussFitter::GaussFitResult & params)
```
cbielow: could this be a free function somewhere in `Math::` ?
I guess we'd have multiple use cases in OpenMS for this (same with Gumbel)

src/openms/include/OpenMS/MATH/STATISTICS/PosteriorErrorProbabilityModel.h
```diff
@@ -125,13 +150,13 @@ namespace OpenMS
...
       {
         return params.A * exp(-1.0 * pow(x - params.x0, 2) / (2 * pow(params.sigma, 2)));
```
cbielow: note: I remember `pow(x, 2)` being very slow compared to `x*x`.

src/openms/include/OpenMS/MATH/STATISTICS/PosteriorErrorProbabilityModel.h
```diff
@@ -168,6 +193,9 @@ namespace OpenMS
       void tryGnuplot(const String& gp_file);
 
 private:
+      /// transform different score types to a range and score orientation that the model can handle (engine string is assumed in upper-case)
```
cbielow: what Engines are allowed?
Use an enum?

src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
```diff
@@ -151,22 +151,23 @@ namespace OpenMS
...
-        //new mean
+        // new mean
         double sum_positive_x0 = sum_pos_x0(x_scores, incorrect_density, correct_density);
         double sum_negative_x0 = sum_neg_x0(x_scores, incorrect_density, correct_density);
 
         double positive_mean = sum_positive_x0 / one_minus_sum_posterior;
         double negative_mean = sum_negative_x0 / sum_posterior;
 
-        //new standard deviation
+        //i new standard deviation
```
cbielow: typo?

src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
```diff
@@ -574,5 +580,262 @@ namespace OpenMS
...
+          return (-1) * log10(max((double)hit.getMetaValue("expect"), smallest_e_value_));
+        }
+      }
+      else
+      {
+        throw Exception::UnableToFit(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No parameters for chosen search engine", "The chosen search engine is currently not supported");
+      }
+
+    // avoid compiler warning (every code path must return a value, even if there is a throw() somewhere)
+      return std::numeric_limits<double>::max();
```
cbielow: you can probably remove this, if the above throw is taken out of the `else { }`?

src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
```diff
@@ -574,5 +580,262 @@ namespace OpenMS
...
+      const vector<ProteinIdentification> & protein_ids,
+      const vector<PeptideIdentification> & peptide_ids,
+      const bool split_charge,
+      const bool top_hits_only,
+      const bool target_decoy_available,
+      const double fdr_for_targets_smaller)
+    {
+      std::set<Int> charges;
+      const StringList search_engines = ListUtils::create<String>("XTandem,OMSSA,MASCOT,SpectraST,MyriMatch,SimTandem,MSGFPlus,MS-GF+,Comet");
+
```
cbielow: suggestion (not neccessarily for this review):
can we convert this list into a central entity in OpenMS along with an Enum? Just to avoid typo's etc...
We will have a use-case for PeptideIndexer to detect XTandem search results soon as well.
Not sure if this is doable...

jpfeuffer: +1 yes please! Also important for ConsensusID.

src/openms/include/OpenMS/MATH/STATISTICS/PosteriorErrorProbabilityModel.h
timosachsenberg: I just tested it on godbolt and both produce the same asm code - except for debug mode (then x*x is faster).

src/openms/include/OpenMS/MATH/STATISTICS/PosteriorErrorProbabilityModel.h
```diff
@@ -71,6 +75,27 @@ namespace OpenMS
       ///Destructor
       ~PosteriorErrorProbabilityModel() override;
 
+      /// extract and transform different score types to a range and score orientation that the model can handle
+      static std::map<String, std::vector<std::vector<double>>> extractAndTransformScores(
```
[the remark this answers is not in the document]
timosachsenberg: I will add some. We are currently planning to replace most code as the current implementation has several flaws. I, thus, will not overdo it with putting work into proper documentation.

src/openms/include/OpenMS/MATH/STATISTICS/PosteriorErrorProbabilityModel.h
timosachsenberg: we will probably remove the fitting of Gauss and Gumble but I can move that to Math if it is considered useful

src/openms/include/OpenMS/MATH/STATISTICS/PosteriorErrorProbabilityModel.h
timosachsenberg: I will put it in the documentation. But I agree that we could use less String constants and use more enums instead

src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
timosachsenberg: Ok I think this is a bigger change. Using http://aantron.github.io/better-enums/
would be a possible solution

cbielow:
> Ok I think this is a bigger change. Using http://aantron.github.io/better-enums/ would be a possible solution
This would also be ideal for TOPP parameters with restrictions.
The current implementation is rather brittle…

cbielow:
+1 for merge if tests pass

cbielow:
python problem
http://cdash.openms.de/viewBuildError.php?buildid=260859

Issue #3207, opened 2018-01-31
QT5: potential drag and drop issue in TOPPAS


timosachsenberg:
Can't connect to output files. TOPPAS thinks I want to transfer tool parameters.

![image](https://user-images.githubusercontent.com/5803621/35615777-77cbc2fa-0673-11e8-8319-ff77065f1f50.png)

hroest:
I also get some weird errors, when e.g. trying to change the peak picker tutorial WF and adding a FileInfo parameter at the end:

```
Transferring parameters between nodes ...
Unknown (or deprecated) Parameter 'write_peak_meta_data' given in outdated parameter file! Ignoring parameter. 
Default-Parameter 'log' overridden: '' --> 'TOPP.log'!
Default-Parameter 'no_progress' overridden: 'true' --> 'false'!
Unknown (or deprecated) Parameter 'algorithm:signal_to_noise' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:centroid_percentage' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:peak_width' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:estimate_peak_width' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:fwhm_lower_bound_factor' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:fwhm_upper_bound_factor' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:optimization' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:thresholds:peak_bound' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:thresholds:peak_bound_ms2_level' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:thresholds:correlation' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:thresholds:noise_level' given in outdated parameter file! Ignoring parameter. 
Unknown (or deprecated) Parameter 'algorithm:thresholds:search_radius' given in outdated parameter file! Ignoring parameter. 
[...]

PR #3298, opened 2018-03-09
Feature/qt5: TOPPAS fix

cbielow:
fixes https://github.com/OpenMS/OpenMS/issues/3207

### DIFF
--- a/doc/doxygen/install/install-win.doxygen
+++ b/doc/doxygen/install/install-win.doxygen
@@ -153,6 +153,9 @@
           <li>we have no experience here, but it should be possible to get the DLL's and Link-Libraries here as well.
         </ol>
       <li>add the <tt>&lt;path-to-qt&gt;/bin</tt> (!) directory to your PATH Environment Variable, so the dlls are found during runtime.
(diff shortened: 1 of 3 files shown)


jpfeuffer:
Looks good. Is the QPA path necessary right after build or only if you move the binaries? I did not recognize it while testing Qt5 on Win. But maybe I did something similar. Or I only tested after the packaging/install steps which copy the plugins.

jpfeuffer:
There were some python failures though. I restarted.

jpfeuffer:
@cbielow Did you test packaging with this. I just can say from experience that making something a QApplication has a lot of consequences.

cbielow:
Qpa is required after build to run it (without install)
-- 
Sent via mobile phone.

Am 10. März 2018 23:11:48 MEZ schrieb jpfeuffer <notifications@github.com>:

jpfeuffer:
Yes I guess if the class is only called by something that already has an include before, it is fine. I mean stuff like ExecutePipeline. Would be a shame if they suddenly need these horrible Qt plugins.

PR #3299, opened 2018-03-10
Try to use same compiler for pyOpenMS and OpenMS

jpfeuffer:
On an older Mac it chooses clang (by default) instead of clang++ which links to a different std library.


timosachsenberg:
+1 if passes

hroest:
does it work?

jpfeuffer:
Yes works, but we seem to need to change more:
https://stackoverflow.com/questions/33738885/python-setuptools-not-including-c-standard-library-headers

I will try to pull over also the deployment target now.

Issue #3244, opened 2018-02-11
Private Thirdparty tests


jpfeuffer:
I added Msfragger and novor to the nightlies but they fail. On win the allocated jvm memory is probably too high for both. For Mac/Linux there are some differences in the results like missing search engine parameters. Not sure what XTandem is doing there.

hroest:
thanks, that would be great to have as nightly test.

jpfeuffer:
Ok novor should be solved. What about MSFragger @lukaszimmermann ?
```
Offending lines:			(tab_width = 8, first_column = 1)

in1:  /Users/builder/jenkins/ws/openms_nightly_packaging/4a4a5dc9/build/src/tests/topp/MSFraggerAdapter_7_out_tmp.pepXML   (line: 4, position/column: 75/76)
<msms_run_summary base_name="spectra" raw_data_type="raw" raw_data=".mzXML"!
                                                                           >

in2:  /Users/builder/jenkins/ws/openms_nightly_packaging/4a4a5dc9/source/src/tests/topp/THIRDPARTY/MSFraggerAdapter_7_out.pepXML   (line: 4, position/column: 75/76)
<msms_run_summary base_name="spectra" raw_data_type="raw" raw_data=".mzXML"!
                                                                            search_engine="X! Tandem">


Easy Access:
/Users/builder/jenkins/ws/openms_nightly_packaging/4a4a5dc9/build/src/tests/topp/MSFraggerAdapter_7_out_tmp.pepXML:4:76:
/Users/builder/jenkins/ws/openms_nightly_packaging/4a4a5dc9/source/src/tests/topp/THIRDPARTY/MSFraggerAdapter_7_out.pepXML:4:76:
```

oliveralka:
Maybe related to this:
"X!Tandem as the search engine in pepXML output
Peptide validation tools do not currently have support for MSFragger so MSFragger masquerades as X!Tandem for compatibility with tools like PeptideProphet. As the hyperscore calculated by MSFragger is slightly different than that of X!Tandem, this may be suboptimal in certain situations. We hope to change this in the future once there is support for MSFragger in peptide validation tools." - http://www.nesvilab.org/msfragger.html

hroest:
@jpfeuffer can we close this?

jpfeuffer:
[...]

PR #3300, opened 2018-03-10
[REFACTOR][FIX] NovorAdapter

oliveralka:
Add java_memory variable 
Fixed removeTempDir 
Edit test (add -java_memory variable #3244)

Further fix from Lukas (thanks a lot) 
QProcess was not working correctly -> Java Process was not started/finished due to a not closed stream.


hroest:
@oliveralka is there a chance that this will also run on travis or only on the nightly servers?

oliveralka:
@hroest I'm not sure about travis @jpfeuffer? But it should at least run on the nightly server.

jpfeuffer:
Yep. Runs on nightly. That is how we found out that the JVM memory was set too high. We were upgrading OS on our macs though. Not everything is smooth yet (e.g. here)

oliveralka:
rebuild jenkins

jpfeuffer:
Ok mac just failed because of cdash upload being too slow.

jpfeuffer:
MErge to test in nightlies.

PR #3304, opened 2018-03-13
[FEATURE] move IDConflict resolver to algorithm, keep conflicting (make unmapped)

timosachsenberg:
Aim of this PR:
- move tool code to algorithm
- keep conflicting IDs but move all but the best to "unmapped" 
- annotate feature identifier as meta value ("feature_id"="17962861576887891871")


jpfeuffer:
Why do we need the feature_leader metavalue? Isn't it clear if it is inside a feature and was not moved to unassigned?

timosachsenberg:
I was thinking the same when I opened the PR. Will check if we can do without it and remove it if that is the case.

jpfeuffer:
Only the python warnings. Looks good.

PR #3305, opened 2018-03-14
[FIX] PeakIntegrator::estimateBackground() correct formula for vertical division height

pasqoo:
The same formula was used for both `base_to_base` and `vertical_division` cases of the parameter `baseline_type`.

Now the following is used instead:
`height = std::min(int_r, int_l) + delta_int_apex;`
or
`height = std::min(int_r, int_l);`


pasqoo:
The pyopenms build has many errors. But it looks like those errors are present also in `develop`. My changes shouldn't affect pyopenms.

jpfeuffer:
They are actually just warnings. Don't know why they are reported as errors. +1 from my side. Ready to merge?

PR #3307, opened 2018-03-15
[FIX] fix Cython

hroest:
seems like Cython 0.28 creates trouble:

the error logs are full of warnings instead of errors:

```
warning: /.../OpenMS/src/pyOpenMS/pxds/ControlledVocabulary.pxd:80:8: 'NONE' redeclared
```

this is potentially because the warnings are reported to stderr instead of stdout ?

### DIFF
--- a/src/pyOpenMS/CMakeLists.txt
+++ b/src/pyOpenMS/CMakeLists.txt
@@ -431,10 +431,18 @@ configure_file(${_env_py_in} ${_env_py} @ONLY)
 #------------------------------------------------------------------------------
 # create targets in makefile
 
-add_custom_target(pyopenms_create_cpp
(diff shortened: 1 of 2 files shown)


hroest:
this is a pretty ugly way to fix the problem but it fixes travis

hroest:
it seems to fix travis for now, any comments? I suggest we merge first and can discuss potential improvements to the current solution

hroest:
@timosachsenberg @jpfeuffer any other ideas how to approach this problem? Is there a way we could filter out these stupid warnings, e.g. 

```
COMMAND ${PYTHON_EXECUTABLE} create_cpp_extension.py | grep -v warning 
```

sounds silly but it may just work :-)

PR #3312, opened 2018-03-15
TOPPView: sequence/ion diagrams for Identification View

hendrikweisser:
Inspired by work from @enetz for crosslinked peptides, this PR introduces diagrams for TOPPView's Identification View that show matched ions in the sequence context (for non-crosslinked peptides). The information comes from peak/fragment annotations in the input file.

An example from an RNPxl test file (the diagram is in the top-right of the spectrum view):

image

The ions to show (e.g. a/b/c, x/y/z) are configurable, but currently hard-coded in `TOPPViewIdentificationViewBehavior`.

The diagrams also support RNAs, but this code is currently commented out because the underlying library classes are not merged yet.


src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
```diff
@@ -12,7 +12,7 @@
 //    documentation and/or other materials provided with the distribution.
 //  * Neither the name of any author or any participating institution
 //    may be used to endorse or promote products derived from this software
-//    without specific prior written permission.
+//    without specific prior written permissi
```
hroest: can you fix this?

hroest:
there is another issue with the cpplint: 

```
/home/travis/build/OpenMS/OpenMS/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp:576:  Storage class (static, extern, typedef, etc) should be first.  [build/storage_class] [5]
/home/travis/build/OpenMS/OpenMS/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp:618:  Storage class (static, extern, typedef, etc) should be first.  [build/storage_class] [5]
/home/travis/build/OpenMS/OpenMS/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp:735:  Storage class (static, extern, typedef, etc) should be first.  [build/storage_class] [5]
Done processing /home/travis/build/OpenMS/OpenMS/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
Total errors found: 3
```

we cannot merge because then the tests will always fail. Either exclude that file from cpplint or we update our cpplint

cbielow:
instead of `for (const auto& aa : seq)`, you can use `for (auto const & aa : seq)`. Our current cpplint will not complain.

hendrikweisser:
We should update cpplint. The error (`Storage class (static, extern, typedef, etc) should be first.`) is a false-positive, because `auto` isn't a "storage class" but a type specifier. The order that I'd consider "normal" is e.g. `static const int x;`. So, agreeing with cpplint, "storage class" should come before `const`, but `auto` replaces the `int`, so should go last. Which is also the reason why I don't like Chris' `auto const&` suggestion.

hroest:
> We should update cpplint

Done: so we can move forward here #3334

timosachsenberg:
Is there a way to disable it?

hendrikweisser:
> Is there a way to hide it?

Nope. There's also no way to choose which fragment types are displayed (ideally this information should come through from the search settings).
Both would be useful additions.

PR #3313, opened 2018-03-15
[REFACTOR] MSFraggerAdapter

oliveralka:
#3244
- reduce java_heapmemory to 2600 (test_only)
- convert MSFragger pepXML to idXML
- optional_output pepXML
- multiple files input -> one file (looping KNIME)


src/utils/MSFraggerAdapter.cpp
```diff
@@ -678,45 +645,47 @@ class TOPPMSFraggerAdapter final :
...
       return EXTERNAL_PROGRAM_ERROR;
     }
 
-    // Copy the output files of MSFragger to the user location
-    for (int i = 0; i < this->input_files.size(); ++i)
+    // convert from pepXML to idXML
+    String pepxmlfile = File::removeExtension(link_name) + "." + "pepXML";
+    std::vector<PeptideIdentification> peptide_identifications;
+    std::vector<ProteinIdentification> protein_identifications;
+    PepXMLFile().load(pepxmlfile, protein_identifications, peptide_identifications);
```
jpfeuffer: Should we overwrite the weird XTandem hack of MSFragger here?

hroest:
how well does the conversion to idXML work? I assume that each modification is just identified by its mass and not by a name, does this break anything in OpenMS ?

oliveralka:
It seems not to break anything but you will get a non-fatal error:

> Non-fatal error while loading 'file/spectra_comet.pepXML': Modification '147.0354' is not uniquely defined by the given data. Using 'Oxidation (M)' to represent any of 'Oxidation (M), oxidation to L-methionine sulfoxide (M), oxidation to L-methionine (R)-sulfoxide (M), oxidation to L-methionine (S)-sulfoxide (M)'.

Please see test files:
MSFraggerAdapter_7_out.idXML
MSFraggerAdapter_7_opt_out.pepXML

The standard modifications seem to work fine. 
There may be problems with modifications which do not exist in unimod but can be used in the MSFragger.

src/utils/MSFraggerAdapter.cpp
oliveralka: Actually I don't know if it is necessary (will probably be updated in the following MSFragger Version) - if you think it is important - I can change it.

jpfeuffer: I think it would make more sense. In the end we have a lot of idxmls lying around with XTandem annotated. And for us it does not make sense as we do not directly use peptide prophet.

oliveralka:
@jpfeuffer  travis - job exceeded time limit - merge?

PR #3314, opened 2018-03-16
[FIX] small improvement

hroest:
after #3307 - its so silly it may just work


jpfeuffer:
Hmm but be careful, this prevents building pyopenms on a VS Command line or powershell

jpfeuffer:
Ah sorry was too quick. But what if there will be the new Cython on Windows?

hroest:
well its anyways only a fix for travis which interprets warnings in stderr as errors

jpfeuffer:
Aaah it's not ctest that decides on error here?

PR #3315, opened 2018-03-16
Some feature for Digestions

cbielow:
.digest() now reports back the number of discarded digestion products (e.g. used in Digestor UTIL) and allows for a length restriction for peptides as well.

Digestor UTIL now streams the FASTA file, instead of reading it all first.

AASequence stops complaining about unknown mass of AA 'X' when the warning is not relevant (but it will warn when invoking member functions which need a mass).


src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -298,7 +298,10 @@ namespace OpenMS
 
   EmpiricalFormula AASequence::getFormula(Residue::ResidueType type, Int charge) const
   {
-    OPENMS_PRECONDITION( !this->has( *ResidueDB::getInstance()->getResidue("X") ), "Cannot get formula of sequence with unknown AA with unknown mass.");
+    if (has(*ResidueDB::getInstance()->getResidue("X")))
```
timosachsenberg: could have a major speed penalty

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -394,7 +403,13 @@ namespace OpenMS
 
   double AASequence::getMonoWeight(Residue::ResidueType type, Int charge) const
   {
-    OPENMS_PRECONDITION( !this->has( *ResidueDB::getInstance()->getResidue("X") ), "Cannot get weight of sequence with unknown AA with unknown mass.");
+    // While PEPTIX[123]DE makes sense and represents an unknown mass of 123.0
+    // Da, the sequence PEPTIXDE does not make sense as it is unclear what a
+    // single, unknown residue should represent.
+    if (has(*ResidueDB::getInstance()->getResidue("X")))
```
timosachsenberg: same here

timosachsenberg:
looks good to me but I am a bit concerned about making the PRECONDITION tested in release builds.
Did you do any speed comparisons?

cbielow:
> could have a major speed penalty

My first approach was to add a bool data member with gets set during ::fromString() and is true/false depending on the presence of 'X'. But this adds to the sizeof() considerably (8 byte, due to alignment) and I guessed that getMonoWeight() is usually called only a few times on an AA Sequence object. Usually they are created anew .. just my feeling though.
Do you have a heavy use-case, so we can do some benchmarking?

timosachsenberg:
SimpleSearchEngine would be one I guess. I calls getMonoWeight() for potentially millions of peptides

cbielow:
My guess is that the code inside the functions also iterates over the entire sequence as does much heavier computations.

cbielow:
Benchmark numbers for SimpleSearchEngine (which makes heavy use of AASequence) on 38938 MS2 spectra (three runs each, best time shown):
before: 03:57 m (wall)
after: 04:14 m (wall)

I made some additional minor changes to AASeq, mainly for readability. They do not affect performance (except for hasSubstring(), but its not used in SimpleSearchEngine as far as I can tell).

timosachsenberg:
hmm that's on the full search including other computations I guess. 
Not that small considering that AASequence creation is already quite slow and a lot of other stuff is also happening in the search.

I would prefer the PRECONDITION...

In release mode, think it is better to check for 'X' before digestions and handle it once per protein.

cbielow:
I'm not really happy with PRECONDITION, since this might produce invalid results in release (when big datasets are crunched).
So I benchmarked getMonoWeight() using 1e5 calls to the function for 1000 internal AAs:
no check: 17 sec
with check (has('X'): 28 sec
Now, thats quite some overhead (and probably too heavy). However, when including the check in the internal for loop (see 2689851), the overhead disappears completely:
with check: 16.8 sec

So now we have the advantages of both solutions.

timosachsenberg:
I am fine with that. 
By the way, would it be save to make the residue X in that code static? Then ResidueDB would only be queried once.

cbielow:
> By the way, would it be save to make the residue X in that code static? Then ResidueDB would only be queried once.

good question. Order of initialization of static members is not defined AFAIK, so not sure if its safe. Or do you mean static local variables? This will work, but at the cost of having multiple static variables, wasting memory even when they are not used (very little, but hey).

timosachsenberg:
I meant static local. Compared to the other ResidueDB stuff it seems neglectible

cbielow:
static is faster: 
16.2 sec (w/o) 
vs.
15.4 sec (static)
So lets go with it.

Ready for merge.

timosachsenberg:
keep the output in the test?

cbielow:
whops. Yes :(

hroest:
@cbielow @timosachsenberg we may want to have a few cases where we use stopwatch tests but compile them out for travis using preprocessor directives. This would allow us to (somewhat easily) compare multiple OpenMS releases and see whether there are any speed regression. However, maybe we should to that on a tool level and not on a unit test level?

PR #3316, opened 2018-03-16
Background subtraction

hroest:
Added a further method for background subtraction, based on what Skyline and EncyclopeDIA is apparently doing. See for example: https://www.biorxiv.org/content/early/2018/03/07/277822

```
We calculate trapezoidal
peak areas across Savitsky-Golay smoothed chromatograms. Analogous to Skyline,
peak intensities are background subtracted by removing a peak area rectangle with a
height equal to the largest intensity of either of the boundary edges. If the area of the
rectangle is larger than the area of the peak the intensity is set to zero.
```

@dmccloskey @pasqoo I would also like to change "vertical division" to "rectangular" as that is what we are really doing, a rectangular background subtraction. Any thoughts on this?

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
@@ -118,7 +118,7 @@ public:
     ///@}
 
     /** @name calculatePeakShapeMetrics() output
-     *
+      
         The calculatePeakShapeMetrics() method uses this struct to save its results.
     */
     ///@{
@@ -210,7 +210,7 @@ public:
     ///@}
 
     /** @name Constant expressions for parameters
-     *
+      
         Constants expressions used throughout the code and tests to set
         the integration and baseline types.
     */
@@ -223,7 +223,7 @@ public:
     static constexpr const char* INTEGRATION_TYPE_SIMPSON = "simpson";
     /// Baseline type: base to base
     static constexpr const char* BASELINE_TYPE_BASETOBASE = "base_to_base";
-    /// Baseline type: vertical division (min of end points)
+    /// Baseline type: vertical division (min of end points; only for backwards compatibility)
     static constexpr const char* BASELINE_TYPE_VERTICALDIVISION = "vertical_division";
     /// Baseline type: vertical division (min of end points)
     static constexpr const char* BASELINE_TYPE_VERTICALDIVISION_MIN = "vertical_division_min";
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
@@ -45,8 +45,24 @@
 
 namespace OpenMS
 {
+
   /**
     @brief Compute the area, background and shape metrics of a peak.
+
+    The area computation is performed in integratePeak() and it supports
+    integration by simple sum of the intensity, integration by Simpson's rule
+    implementations for an odd number of unequally spaced points or integration
+    by the trapezoid rule.
+
+    The background computation is performed in estimateBackground() and it
+    supports three different approaches to baseline correction, namely
+    computing a rectangular shape under the peak based on the minimum value of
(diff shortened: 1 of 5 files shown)


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
```diff
@@ -102,7 +118,8 @@ namespace OpenMS
     ///@}
 
     /** @name calculatePeakShapeMetrics() output
-      The calculatePeakShapeMetrics() method uses this struct to save its results.
+     *
```
pasqoo: Is the asterisk unwanted?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
```diff
@@ -193,8 +210,9 @@ namespace OpenMS
     ///@}
 
     /** @name Constant expressions for parameters
-      Constants expressions used throughout the code and tests to set
-      the integration and baseline types.
+     *
```
pasqoo: Same here

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
```diff
@@ -205,8 +223,12 @@ namespace OpenMS
     static constexpr const char* INTEGRATION_TYPE_SIMPSON = "simpson";
     /// Baseline type: base to base
     static constexpr const char* BASELINE_TYPE_BASETOBASE = "base_to_base";
-    /// Baseline type: vertical division
+    /// Baseline type: vertical division (min of end points)
```
pasqoo: Maybe we should specify that this is kept for some backward compatibility, to not confuse future readers.

src/openms/source/ANALYSIS/OPENSWATH/PeakIntegrator.cpp
```diff
@@ -249,19 +249,28 @@ namespace OpenMS
...
-          area += it->getPos();
+          pos_sum += it->getPos();
         }
         UInt n_points = std::distance(p.PosBegin(left), p.PosEnd(right));
-        area = (area - n_points * p.PosBegin(left)->getPos()) * delta_int / delta_pos + n_points * int_l;
+
+        // We construct the background area as the sum of a rectangular part
+        // and a triangle on top. The triangle is constructed as the sum of the
+        // line's y value at each sampled point: \sum_{i=0}^{n} (x_i - x_0)  * m
+        double rectangle_area = n_points * int_l;
```
pasqoo: This and the two following can be `const`.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
hroest: @pasqoo do you need this for backwards compatibility or can we just remove it?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/PeakIntegrator.h
```diff
@@ -205,8 +223,12 @@ namespace OpenMS
     static constexpr const char* INTEGRATION_TYPE_SIMPSON = "simpson";
     /// Baseline type: base to base
     static constexpr const char* BASELINE_TYPE_BASETOBASE = "base_to_base";
-    /// Baseline type: vertical division
+    /// Baseline type: vertical division (min of end points; only for backwards compatibility)
```
dmccloskey: For naming, I would just suggest using "vertical_division" and "vertical_division_max" and not using "vertical_division_min".  The name "vertical_division" is pretty ubiquitous in most peak integration software (at least for small molecules).

hroest: I am ok with both, its just that using min/max would be less confusing for people.

dmccloskey: That works me.  Ready to be merged?

Issue #3010, opened 2017-11-13
cpplint / cppcheck and C++11


hroest:
We have a few false positives with cpplint:

- #2969
- #3007 

```
/home/travis/build/OpenMS/OpenMS/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp:864:  Storage class (static, extern, typedef, etc) should be first.  [build/storage_class] [5]
Done processing /home/travis/build/OpenMS/OpenMS/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
Total errors found: 1
```

hroest:
probably our version of cppcheck is also too old: 
https://github.com/danmar/cppcheck/commit/9d3359177d68f2236245fa1104e87672fe9b503b

hroest:
1.65: Update Makefile

    master 

    1.81 

    1.65 

1 parent c0131ee commit 9d3359177d68f2236245fa1104e87672fe9b503b danmar danmar committed on May 10, 2014

jpfeuffer:
What's the status? Do we still need to upgrade?

hroest:
yes this has been merged in #3011

hroest:
seems like we did not update cpplint, its still on a very old version 9bc8d670e939bc5dc77c34ba09eccb62054aad6d

hroest:
we probably should replace it with a new version https://raw.githubusercontent.com/cpplint/cpplint/master/cpplint.py

jpfeuffer:
see #3334

hroest:
Merged in 

- #3011
- #3334

PR #3317, opened 2018-03-16
Extra data arrays

hroest:
Allows representation of extra binary data arrays in Spectra / Chromatograms

- Extra data arrays could be ion mobility arrays, etc
- allow Chromatogram extractor to also filter by ion mobility


hroest:
@jpfeuffer it seems that cpplint complains about errors as it may not be C++11 aware:

/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/OPENSWATH/DATAACCESS/SpectrumAccessOpenMS.cpp:78:  Storage class (static, extern, typedef, etc) should be first.  [build/storage_class] [5]

hroest:
looks like the same issues as #3010 -- didnt we solve that?

hroest:
only LogStream_test failed on windows.

hroest:
rebuild win

hroest:
any comments on this? anybody opposed to merge this?

src/openms/source/FORMAT/CachedMzML.cpp
```diff
@@ -357,13 +474,42 @@ namespace OpenMS
...
+    for (const auto& ida : chromatogram.getIntegerDataArrays() )
+    {
+      Size len = ida.size();
+      ofs.write((char*)&len, sizeof(len));
+      Size len_name = ida.getName().size();
+      ofs.write((char*)&len_name, sizeof(len_name));
+      ofs.write((char*)&ida.getName().front(), len_name * sizeof(ida.getName().front()));
+      // now go to the actual data
+      Datavector tmp;
+      tmp.reserve(ida.size());
```
cbielow: optimization:
instead of allocating tmp in every loop, you could declare it outside, while still calling reserve() + clear() inside before pushing. Should be faster, since it saves the (de)allocation.

src/openms/source/FORMAT/CachedMzML.cpp
```diff
@@ -357,13 +474,42 @@ namespace OpenMS
...
+      for (const auto& val : ida) {tmp.push_back(val);}
+      ofs.write((char*)&tmp.front(), tmp.size() * sizeof(tmp.front()));
```
cbielow: is is even necessary to create an array first and then writing it? You could just write out each element piece by piece?

timosachsenberg:
I only had a quick look but looks good form my side

src/openms/source/FORMAT/CachedMzML.cpp
```diff
@@ -245,80 +269,142 @@ namespace OpenMS
...
+      data.back()->data.resize(len);
+      data.back()->description = buffer;
+      ifs.read((char*)&(data.back()->data)[0], len * sizeof(DatumSingleton));
+    }
+    delete[] buffer;
+    return data;
   }
 
-  void CachedmzML::readChromatogram_(Datavector& data1, Datavector& data2, std::ifstream& ifs) const
+  std::vector<OpenSwath::BinaryDataArrayPtr> CachedmzML::readChromatogramFast(std::ifstream& ifs)
```
cbielow: this looks like it does a subset of readSpectraFast().
Can this code be re-used instead of copied?

src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp
```diff
@@ -90,6 +90,266 @@ namespace OpenMS
...
+        tmp.setIntensity(*int_it);
+        tmp.setMZ(*mz_it);
+        ++mz_it;
+        ++int_it;
+        spectrum.push_back(tmp);
+      }
+    }
+  }
+
+  void MzMLSpectrumDecoder::decodeBinaryDataMSSpectrum_(std::vector<BinaryData>& data_, OpenMS::MSSpectrum& spectrum)
```
cbielow: is `data_` const here?
Remove `_`

src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp
```diff
@@ -90,6 +90,266 @@ namespace OpenMS
...
+                  x_precision_64, int_precision_64,
+                  x_index, int_index, default_array_length_);
+
+    if (data_.size() > 2) 
+    {
+      fillDataArrays(data_, spectrum);
+    }
+  }
+
+  void MzMLSpectrumDecoder::decodeBinaryDataMSChrom_(std::vector<BinaryData>& data_, OpenMS::MSChromatogram& chromatogram)
```
cbielow: same as above

src/openms/include/OpenMS/FORMAT/HANDLERS/IndexedMzMLHandler.h
```diff
@@ -100,25 +105,29 @@ namespace OpenMS
...
-    IndexedMzMLFile();
+    IndexedMzMLHandler();
 
     /**
       @brief Constructor
 
       Tries to parse the file, success can be checked with getParsingSuccess()
     */
-    explicit IndexedMzMLFile(String filename);
+    explicit IndexedMzMLHandler(String filename);
```
cbielow: string by const ref?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ChromatogramExtractor.h
```diff
@@ -165,16 +171,54 @@ namespace OpenMS
...
+     * @param im_extraction_window Extracts a window of this size in ion mobility
+     * @param filter Which filter to use (bartlett or tophat)
+     *
+     * @note: whenever possible, please use this ChromatogramExtractorAlgorithm implementation
+     *
+    */
+    void extractChromatograms(const OpenSwath::SpectrumAccessPtr input, 
+        std::vector< OpenSwath::ChromatogramPtr >& output, 
+        const std::vector<ExtractionCoordinates>& extraction_coordinates,
+        double mz_extraction_window, bool ppm, double im_extraction_window, String filter)
```
cbielow: string by const ref?

src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp
hroest: its not const

Issue #3301, opened 2018-03-12
[QT5] pyopenms compilation issue on travis


jpfeuffer:
http://cdash.openms.de/viewBuildError.php?buildid=261128
http://cdash.openms.de/viewBuildError.php?buildid=261130

It sometimes just does not find QDate.h. Sometimes restarts help, sometimes they don't.
@hroest Any idea?

jpfeuffer:
I changed the title. I found out that actually all of the failing runs are from the qt5 branch. So it's a qt5 issue. And it seems to occur always.

jpfeuffer:
https://github.com/OpenMS/OpenMS/commit/d32a07a8073368f0038e0d6455bda2faac7952b9

jpfeuffer:
https://github.com/OpenMS/OpenMS/commit/33a5cede872965bbe6f3903ea8251141d82a366d too

jpfeuffer:
Fixed by the above commits.

hroest:
@jpfeuffer isnt that going to be a problem with newer Qt versions where QDate.h is not  under QtCore, or did I misunderstand something in d32a07a8073368f0038e0d6455bda2faac7952b9 - see http://doc.qt.io/qt-5/qdate.html

jpfeuffer:
Hmm yes. You are right. If we prefix, the header should be found even when it changes the module (as long as we adapt cmake). In this case I don't think it changed why it does not matter. There are other includes like qstring where we use the prefix.
If we use the module prefixes however users would only need one include Path for QT (although we export all of them anyway)
So hmm slightly in favor of changing all of them to non-prefixed.

hroest:
>  So hmm slightly in favor of changing all of them to non-prefixed.

however then we need to find a way to make pyopenms compile for the non-prefixed version

jpfeuffer:
Ah, in the end that did not matter for pyopenms (but the new way of handling QT_INCLUDES).
I kept it for consistency. It was (I think) the only QT include that had no prefix.

grosenberger:
I'm having problems with building pyOpenMS on macOS due to the Qt5 merge with OpenMS/develop. It seems to be related to the added prefix in d32a07a. Are others experiencing the same issues?

jpfeuffer:
Weird. It compiled on travis back then. Feel free to remove it of course if that solves it alone.

hroest:
[...]

PR #3318, opened 2018-03-17
[FIX] pyOpenMS Qt5 for macOS

grosenberger:
`QT_LIBRARY_DIR` unfortunately doesn't exist with Qt5 anymore and pyOpenMS can't be built under macOS. This patch ensures that the libraries can be copied. Unfortunately, the problem #3301 persists.

clang can't find the Qt5 headers. Did anybody observe similar issues?


jpfeuffer:
You have to check which includes are passed from openms (it's external includes which include qt5core and qt5network at least). Probably the QT5 root dir is missing (which needs to be there because we prefix the includes with the subdirectory). Maybe we pass this directory by other means somehow on the build machines. I will take a close look on Monday.

grosenberger:
Thanks for the suggestion, the new patch enables building pyOpenMS on macOS again.

jpfeuffer:
No problem. The changes make sense. +1

Issue #3278, opened 2018-03-05
MSConvert reports in minutes instead of seconds in mzml file


nastjastrizh:
I have problems with RT in my mzml file. When I convert my raw file *.d to mzml by using MSConvert, I get RT in minutes. And in other programs, such as TOPPView or SeeMS, the retention time doesn't convert to seconds, so all this programs report minutes. 
For me it's problem to have time in minutes, because after conversion I run MRMMapper -> OpenSwathAnalyzer, and all this tools have some parameters in seconds by default (and with minutes it doesn't work)
I tried to convert firstly to mzXML and after to mzml, but I had an error (Error loading metadata: no spectra or chromatograms in seems ).
I also tried to open MSConvert and do conversion using command line (titleMaker <ScanStartTimeInminutes>), but it didn't solve the problem too.

Is there any solution to convert retention time from minutes to seconds to avoid problems with identification in OpenSwath ?

hroest:
I have not encountered this problem before, if the time is correctly annotated as "minutes" using `MS:1000016` then it should be stored as seconds inside OpenMS and there should be no issue. See this code snippet:

https://github.com/OpenMS/OpenMS/blob/f8f754421a7565787beb13e5a1ca127115a8f9fc/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp#L1930-L1941

so OpenMS internally always stores time in seconds, so if the mzML file from proteowizard is correctly written there should be no issue. Can you send us the file (or part of the file) and we have a look at it?

hroest:
ok, I think I understand now what you are saying - are you converting SRM data? Yes, we have a problem of OpenMS not correctly reading units and I have proposed a fix in PR #3320

hroest:
this should be fixed in #3320

PR #3320, opened 2018-03-19
Fix units

hroest:
Reads units correctly from binary data arrays (converts minutes to seconds)


hroest:
fixes #3278

hroest:
still needs tests

hroest:
add tests, turns out that some of our data was recorded in minutes and produced test errors now that we properly handle this

PR #3321, opened 2018-03-20
[FIX] fix special XML characters for TraML

hroest:
escape XML characters in TraML


cbielow:
just out curiosity: what happens to the double quote in the input? at what point is it removed?

hroest:
I was wondering that too, it turns out we remove it when reading the CSV file since fields in CSV often contain quotes.

cbielow:
two tests still fail
    projectroot.src.tests.topp.TOPP_ConvertTSVToTraML_test_6_out1
    projectroot.src.tests.topp.TOPP_ConvertTraMLToTSV_test_1_out1

+1 otherwise

PR #3325, opened 2018-03-21
Fix/mrm

hroest:
multiple fixes for MRM / OpenSWATH code

- correctly use minutes when data interpretation is in minutes
- EMG fitting was using wrong initial parameters that were wider than the peak
- code move

### DIFF
--- a/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EmgFitter1D.cpp
+++ b/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EmgFitter1D.cpp
@@ -95,7 +95,7 @@ namespace OpenMS
       // f'(h)
       derivative_height = w / s * sqrt_2pi * exp1 / exp2;
 
-      // f'(h)
+      // f'(w)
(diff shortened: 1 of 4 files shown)


timosachsenberg:
+1

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EmgFitter1D.cpp
```diff
@@ -265,15 +269,20 @@ namespace OpenMS
...
     // The computations can lead to an overflow error at very low values of symmetry (s~0).
     // For s~5 the parameter can be approximated by the Levenberg-Marquardt algorithms.
     // (the other parameters are much greater than one)
     if (symmetry_ < 1)
+    {
       symmetry_ += 5;
+    }
+
+    // Need to ensure that we do not go beyond the maximal width of the peak
+    symmetry_ = std::min(symmetry_, max_peak_width);
```
hroest: this is especially a problem when we are measuring time in minutes and not in seconds

PR #3326, opened 2018-03-21
Feature/filecacher

hroest:
- allow low memory file caching
- also remove float / integer data arrays from spectra and chromatograms after caching

### DIFF
--- a/src/openms/source/FORMAT/DATAACCESS/MSDataCachedConsumer.cpp
+++ b/src/openms/source/FORMAT/DATAACCESS/MSDataCachedConsumer.cpp
@@ -78,7 +78,16 @@ namespace OpenMS
     }
     writeSpectrum_(s, ofs_);
     spectra_written_++;
-    if (clearData_) { s.clear(false); }
+
+    // Clear all spectral data including all float/int data arrays
+    if (clearData_)
+    {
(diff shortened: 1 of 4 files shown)


src/openms/source/FORMAT/DATAACCESS/MSDataCachedConsumer.cpp
```diff
@@ -78,7 +78,16 @@ namespace OpenMS
...
     writeSpectrum_(s, ofs_);
     spectra_written_++;
-    if (clearData_) { s.clear(false); }
+
+    // Clear all spectral data including all float/int data arrays
+    if (clearData_)
+    {
+      s.clear(false);
+      MSSpectrum::FloatDataArrays fempty;
+      s.setFloatDataArrays(fempty);
```
timosachsenberg: would s.setFloatDataArrays({}); also work?

timosachsenberg:
looks good

src/openms/source/FORMAT/DATAACCESS/MSDataCachedConsumer.cpp
hroest: yes, it does!

PR #3330, opened 2018-03-26
[FIX] Only allow ssl when QT was compiled with it and if found at runtime

jpfeuffer:


src/openms/source/FORMAT/MascotRemoteQuery.cpp
```diff
@@ -77,8 +78,10 @@ namespace OpenMS
     defaults_.setValidStrings("login", ListUtils::create<String>("true,false"));
     defaults_.setValue("username", "", "Name of the user if login is used (Mascot security must be enabled!)");
     defaults_.setValue("password", "", "Password of the user if login is used (Mascot security must be enabled!)");
-    defaults_.setValue("use_ssl", "false", "Flag indicating wether the server uses HTTPS or not.");
+#ifndef QT_NO_SSL
+    defaults_.setValue("use_ssl", "false", "Flag indicating whether you want to send requests to an HTTPS server or not (HTTP). Requires OpenSSL to be installed (see openssl.org)");
     defaults_.setValidStrings("use_ssl", ListUtils::create<String>("true,false"));
+#endif
```
hroest: I would recommend against this as this can lead to all sorts of user confusion. Also I am sure KNIME and all other workflow engines will be confused. I suggest to leave the option in but throw an exception at runtime "this version of OpenMS was not compiled with SSL support" as you do below.

hroest:
:+1: I like it but I recommend against preprocessor stuff in the options, rather throw a runtime error

PR #3324, opened 2018-03-21
[QT5] pyOpenMS packaging from hroest

jpfeuffer:
Redirected to develop and adressed codacy issues.

PR #3331, opened 2018-03-26
Revert "[QT5] pyOpenMS packaging from hroest"

hroest:
Reverts OpenMS/OpenMS#3324


hroest:
since this broke develop

PR #3332, opened 2018-03-26
Feature/experimental design

lukaszimmermann:
Contains the most recent version of the experimental design, as implemented on the MSstats and ProteomicsLFQ branch. Most implementation was moved from `ExperimentalDesign.h` file to the `ExperimentalDesign.cpp` file.

Also include updated information about the sample section in `ExperimentalDesign.h`.


hroest:
I miss the unit tests and the pyOpenMS bindings. You fix the bindings but you dont provide new ones for the new classes 

@timosachsenberg @jpfeuffer should we have strict rules for this?
The rules for having tests for each class and public method are pretty clear but the rules for Python bindings are not. However, if it is part of the public API it should have Python bindings IMHO

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
-    design.runs.clear();
-
-    TextFile tf(tsv_file, true);
+    using RunRows = std::vector<ExperimentalDesign::RunRow>;
 
-    int line_number = 0;
 
-    for (TextFile::ConstIterator sit = tf.begin(); sit != tf.end(); ++sit)
+    // Parse Error of filename if test holds
+    void parseErrorIf(const bool test, const String &filename, const String &message)
```
hroest: again, this should be in `FORMAT`

hroest:
To me it seems like the class does too much and should be split up into a File IO class in `FORMAT` and a logical class that manages the data. Also, we need unit tests for these classes

src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
```diff
@@ -53,51 +61,270 @@ namespace OpenMS
...
 
-    /// return if each fraction number is associated with the same number of runs 
-    bool sameNrOfRunsPerFraction() const;
+    class OPENMS_DLLAPI SampleSection
+    {
+    public:
+      SampleSection() = default;
+
+      // Number of columns of the Sample Section
+      Size n_columns_;
```
timosachsenberg: do we need n_columns_? if so it should not be public

src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
```diff
@@ -53,51 +61,270 @@ namespace OpenMS
...
+
+      // The entries of the Sample Section, filled while parsing
+      // the Experimental Design File
+      std::vector< std::vector < String > > content_;
```
timosachsenberg: should be private? same for  other members

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+    {
+      for (Size i = 0; i < vec.size(); ++i) {
+        vec[i] = vec[i].trim();
       }
+    }
 
-      // run-level rows
-      StringList cells;
-      s.split("\t", cells);
+    inline bool contains(const StringList &vec, const String &item)
```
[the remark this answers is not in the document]
timosachsenberg: otherwise this could be part of the list classes (was it ListUtils?)

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+      auto e = std::end(vec);
+      return std::find(std::begin(vec), e, item) != e;
+    }
 
-      if (s.size() < 4)
-      {
-        throw Exception::ParseError(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, tsv_file,
-          "Error: Not all columns of experimental design provided: " + String(s) + ".");
-      }
+    void parse_header(const StringList &header,
```
timosachsenberg: parse_header should be parseHeader_ and private

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+          result = spec_file;
+        }
+      } else {
         // set to absolute path
-        r.file = spec_file;
+        result = spec_file;
       }
 
-      if (!File::exists(r.file))
+      /* Currently, existence of file is not required
```
timosachsenberg: check if commented section is needed

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+    {
+      ExperimentalDesign design;
+      design.run_section_.clear();
+
+      bool has_sample(false);
+      bool has_channel(false);
+
+      // Maps the column header string to the column index for
+      // the run section
+      std::map <String, Size> _run_column_header_to_index;
```
timosachsenberg: we don't prefix _ to variable names

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+
+      unsigned line_number(0);
+      int state(0);
+      Size n_col = 0;
+
+      const TextFile text_file(tsv_file, true);
+      for (String s : text_file) {
```
timosachsenberg: { into new line

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+          continue;
+        }
+
+        // Now split the line into individual cells
+        StringList cells;
+        line.split("\t", cells);
+        trimAll(cells);
+
+        // State = 0 : Run Section header
+        if (state == 0) {
```
timosachsenberg: consider using enums for states

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
-    return ret;
-  }
 
-  bool ExperimentalDesign::sameNrOfRunsPerFraction() const
-  {
-    map<unsigned, set<unsigned> > frac2run = getFractionToRunsMapping();
-    if (frac2run.size() > 1)
+    map<pair<String, unsigned>, unsigned> ExperimentalDesign::pathChannelMapper(
+            const bool basename,
+            unsigned (*f)(const ExperimentalDesign::RunRow &row)) const
```
timosachsenberg: why does it need to be that generic?

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+                {
+                  return std::tie(a.run, a.fraction, a.channel, a.sample, a.path) <
+                         std::tie(b.run, b.fraction, b.channel, b.sample, b.path);
+                });
+    }
+
+
+    /* Implementations of SampleSection */
+
+    void ExperimentalDesign::SampleSection::getSamples(std::set<unsigned> &samples) const
```
timosachsenberg: better to just return the set of samples?
filling the referenced vector is kind of old-style c++

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -41,152 +41,578 @@
...
+    {
+      samples.clear();
+      for (auto it = sample_to_rowindex_.begin();
+           it != sample_to_rowindex_.end(); ++it)
+      {
+        samples.insert(it->first);
+      }
+    }
+
+    void ExperimentalDesign::SampleSection::getFactors(std::set< String > &factors) const
```
timosachsenberg: same here...

timosachsenberg:
@lukaszimmermann looks pretty good. I agree with @hroest that we should separate the parsing and storing from the rest (and move that class to FORMAT).
Also please check the accessibility of the members (I think many of them can be private).
Regarding how mandatory pyOpenMS bindings are: I think this is a good topic for the developer retreat.

lukaszimmermann:
Thanks a lot for the feedback. I will split the class into IO and structure parts.

lukaszimmermann:
I addressed most of the points. It is not entirely clear to me how the class `ExperimentalDesignTable` can be split. The class does three things:
- Load Experimental design from various formats (static factory methods)
- Sanity checking of the format of the experimental design
- Providing several conventient accessors.

timosachsenberg:
For consistency reasons, it would make sense to move the IO only part (only the tsv reading) to an IO class .
Regarding the name ExperimentalDesignTable that @hroest proposed - I am not 100% happy with that.
For one, these are actually two tables in the file. For another, the tabular representation is the representation in the file. Otherwise one could argue that we have several classes named without the suffix *Table. The later is not that important to me but naming it ExperimentalDesign consisting of a RunSection and a SampleSection seemed not a worse choice in the first place. If one wants to emphasize the tabular nature one could rename RunSection to RunTable and SampleSection to SampleTable.
Again, not that important to me but maybe it was not clear that there are (up to) two tables in the design.

timosachsenberg:
@lukaszimmermann moving the IO to a separate class makes it easier to, e.g. modularize the IO part in OpenMS. We don't to this right now but in the future, the IO stuff could be a namespace thing, a separate link library, or a c20 and beyond module.

lukaszimmermann:
I moved the factory methods to the `ExperimentalDesignIO` class (which now has the `load` and `from*` static members).

timosachsenberg:
Hi Lukas, that was quick.
Two things:
1. We usually name the IO classes with suffix File
2. static ExperimentalDesign fromConsensusMap(const ConsensusMap& c); I would not put general conversion functions in a IO class

timosachsenberg:
also consider if friendship is needed

lukaszimmermann:
@timosachsenberg 
Ok, renamed `ExperimentalDesignIO` to `ExperimentalDesignFile` and put the converter methods back to `ExperimentalDesign`. 

Removed friend declarations.

Shall were remove the default constructors and setters? Then we could make the `ExperimentalDesign` class immutable.

timosachsenberg:
Good idea but I think we need the default constructor for pyOpenMS

src/openms/include/OpenMS/FORMAT/ExperimentalDesignFile.h
```diff
@@ -0,0 +1,69 @@
...
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer:	Timo Sachsenberg $
+// $Authors: Timo Sachsenberg, Lukas Zimmermann $
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_KERNEL_EXPERIMENTALDESIGNFILE_H
+#define OPENMS_KERNEL_EXPERIMENTALDESIGNFILE_H
```
hroest: change KERNEL to FORMAT ?

src/openms/include/OpenMS/FORMAT/ExperimentalDesign.h
```diff
@@ -0,0 +1,271 @@
...
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer:	Timo Sachsenberg $
+// $Authors: Timo Sachsenberg $
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_KERNEL_EXPERIMENTALDESIGN_H
+#define OPENMS_KERNEL_EXPERIMENTALDESIGN_H
```
hroest: this could go either into METADATA, but probably not into FORMAT

lukaszimmermann:
ExperimentalDesign.h was moved to METADATA

hroest:
@lukaszimmermann your pyOpenMS compilation failed, see https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_docker_pyopenms/75/console

hroest:
@lukaszimmermann please fix the following error:

```
git clone https://github.com/hroest/docker_files.git && \
cd docker_files/openms/pyopenms && \
docker build --build-arg branch=feature/experimental_design --build-arg repo=lukaszimmermann/OpenMS .
[...]
Level 25:root:parsing progress 200 out of 483
Level 25:root:parsing progress 250 out of 483
Level 25:root:parsing progress 300 out of 483
Level 25:root:parsing progress 350 out of 483

Error compiling Cython file:
...

cdef extern from "<OpenMS/METADATA/ExperimentalDesign.h>" namespace "OpenMS":
    
    cdef cppclass ExperimentalDesign "OpenMS::ExperimentalDesign":
        ExperimentalDesign(ExperimentalDesign) nogil except + #wrap-ignore
        libcpp_vector[ ExperimentalDesign_RunRow] rows
                                               ^

/OpenMS/src/pyOpenMS/pxds/ExperimentalDesign.pxd:8:48: unknown type in template argument

```

src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
```diff
@@ -34,70 +34,179 @@
...
+      sort_();
+      checkValidRunSection_();
+    }
+
+    const RunRows& getRunSection() const;
+
+    void setRunSection(const RunRows& run_section);
+
+    // Returns the Sample Section of the experimental design file
+    const ExperimentalDesign::SampleSection& getSampleSection() const;
```
hroest: I think the method names and documentation should be more clear

src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
```diff
@@ -34,70 +34,179 @@
...
-    ///   2	humanB.mzML	      1	              1          
-    ///   3	humanC.mzML	      2	              1              
-    ///   4	humanD.mzML	      2	              1
-    ///   5	humanE.mzML	      3	              1
-    ///   6	humanF.mzML	      3	              1          
-    ///   7	humanG.mzML	      3	              2       (<- example how a 2nd technical replicate is stored for fraction 3)
-    /// TODO: add possibility to provide optional columns that map to additional run-level specific meta data (see, for example, the mzTab specification)
-    class OPENMS_DLLAPI MSRun
+
+    class OPENMS_DLLAPI RunRow
```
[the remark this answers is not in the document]
timosachsenberg: For multiplexed data one row does not necessarily correspond to an injection.
The name run section and sample section have been used to distinguish the
meta data associated with mapping quant. channels in fractionated files
and the sample specific meta data (like conditions etc.)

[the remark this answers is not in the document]
timosachsenberg: we could also call this MSFileSection and MSFileRow? Would this be better?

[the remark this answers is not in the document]
hroest: ok, in that case I think it should just be documented better. I dont really like the naming of "rows" and "columns" here since it referers to a specific file layout instead of concepts used by experimentalisits

hroest:
@lukaszimmermann please merge https://github.com/lukaszimmermann/OpenMS/pull/1 for pyOpenMS to compile

src/openms/include/OpenMS/FORMAT/ExperimentalDesignFile.h
```diff
@@ -0,0 +1,125 @@
...
+	2	3		SPECTRAFILE_F3_TR2.mzML	3		7
+	2	1		SPECTRAFILE_F1_TR2.mzML	4		8
+	2	2		SPECTRAFILE_F2_TR2.mzML	4		8
+	2	3		SPECTRAFILE_F3_TR2.mzML	4		8
+
+  /// 2) Mandatory section with sample information of the experimental design.
+  ///    Required to process fractionated data. One Column must be 'Sample', other columns
+  ///    are unspecified and can contain arbitrary factors
+
+ Sample	Some_Condition	Technical_Replicate
```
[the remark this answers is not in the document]
timosachsenberg: what about Technical_Replicate? is this required by MSStats? if so this should be documented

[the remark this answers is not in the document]
timosachsenberg: Also it should be documented if something is an id used for grouping (e.g., all technical replicates of a sample have the same number) or for enumeration (e.g., 1,2,3 to denote TR1, TR2, and TR3)

src/openms/include/OpenMS/FORMAT/ExperimentalDesignFile.h
```diff
@@ -0,0 +1,125 @@
...
+                                      For multiplexed experiments, these might differ as multiple samples can be measured in single files
+         Fraction:                    1st, 2nd, .., fraction. Note: All runs must have the same number of fractions.
+         Path(Spectra File):          Path to mzML files
+         Channel:                     Channel in MS file:
+                                      label-free: always 1
+                                      TMT6Plex: 1..6
+                                      SILAC with light and heavy: 1..2
+         Sample:                      Index of sample measured in the specified channel X, in fraction Y of run Z
+
+	Fraction Group	Fraction	Path(Spectra File)	Channel		Sample
```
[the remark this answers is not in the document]
timosachsenberg: done

timosachsenberg:
TODO: clean up code required. move implementations to cpp file

src/openms/include/OpenMS/FORMAT/ExperimentalDesignFile.h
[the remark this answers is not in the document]
lukaszimmermann: The columns in the Sample table are arbitrary, you can name them whatever you like (with the exception of the Sample column, of course). But you need to tell MSstatsConverter, which column in the Sample table holds the Biological Replicate. This is a parameter of the converter. There is a default value for this parameter.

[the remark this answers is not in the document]
timosachsenberg: ok so it should be documented if Bio.Rep. is a grouping identifier or a counting identifier for MSstats

src/tests/class_tests/openms/source/ExperimentalDesign_test.cpp
```diff
@@ -75,115 +76,118 @@ END_SECTION
...
 {
-  // TODO
+  bool b = design.isFractionated(); 
 }
 END_SECTION
 
 START_SECTION((bool sameNrOfMSFilesPerFraction() const ))
 {
-  // TODO
+  bool b = design.sameNrOfMSFilesPerFraction();
```
hroest: please also test whether it returns the right value!!

src/tests/class_tests/openms/data/ExperimentalDesign_input_1.tsv
```diff
@@ -0,0 +1,29 @@
+Fraction_Group	Fraction	Spectra_Filepath	Label	Sample
+1	1	JD_06232014_sample1-A.raw	1	1
+2	1	JD_06232014_sample1_B.raw	1	2
+3	1	JD_06232014_sample1_C.raw	1	3
```
hroest: @timosachsenberg can you please make an example file that covers multiple cases, e.g. multiple fractionations or a fractionation combined with TMT ? so that people have some idea what such a file would look like

src/tests/class_tests/openms/source/ExperimentalDesignFile_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+START_SECTION(~ExperimentalDesignFile())
+{
+  delete ptr;
+}
+END_SECTION
+
+START_SECTION((static ExperimentalDesign load(const String &tsv_file, bool require_spectra_files)))
+{
+ExperimentalDesign design = ExperimentalDesignFile::load(
+  OPENMS_GET_TEST_DATA_PATH("ExperimentalDesign_input_1.tsv"), false);
```
hroest: please test here that the data was loaded properly, e.g. all the fields are populated with the right values

hroest:
starts to look good, however your unit tests should not just test that you can call the function but also test taht the function does what it is supposed to do. Think of it as a way of documenting your code ...

timosachsenberg:
sure ... still stubs. I push frequently to keep Lukas posted

src/tests/class_tests/openms/data/ExperimentalDesign_input_1.tsv
timosachsenberg: done in ExperimentalDesign_input_2.tsv fractionated 4plex

src/tests/class_tests/openms/source/ExperimentalDesignFile_test.cpp
timosachsenberg: done in ExperimentalDesign_test where I load the same files but test the content

PR #3333, opened 2018-03-27
Add option to disable MSFragger tests as they need too much RAM

jpfeuffer:


oliveralka:
Look good to me! 

We are in contact with the guys from MSFragger and will sort it out. 
But in regard to the failing nighlies - I think is a good idea.

PR #3334, opened 2018-03-27
update cpplint.py from googles master branch

jpfeuffer:


hroest:
hmm great:
```
58% tests passed, 779 tests failed out of 1848
```

I guess we have to work on some of our rules

hroest:
PS: how do you run it locally, is it through cmake or how do you do it?

hroest:
there are many of the same errors"

```

Test output

/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp:45:  Do not use namespace using-directives.  Use using-declarations instead.  [build/namespaces] [5]
/home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp:139:  Namespace should be terminated with "// namespace OpenMS"  [readability/namespace] [5]
Done processing /home/travis/build/OpenMS/OpenMS/src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
Total errors found: 2
```

if we disable these two rules we should hopefully have many more passing tests

jpfeuffer:
Yes exactly, we have to go over our rules.
Cmake basically runs 
`python cpplint.py --verbose=5 hello.cpp` on every source file

jpfeuffer:
down to 58. Manageable.
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=263592

hroest:
now it flags a lot of `do ... while` loops, see also this https://github.com/ros/roslint/issues/28

hroest:
also we should undo changes to `src/tests/coding/createcpplinttests.cmake` again

jpfeuffer:
How do you mean? Why undo?

hroest:
> How do you mean? Why undo?

sorry, disregard that comment

jpfeuffer:
Great! Thanks! @timosachsenberg did you look at clang-format to add the namespace comments automatically. Otherwise I think I will merge in this state, IMHO they are not super important.

timosachsenberg:
yes but it currently does not support applying one rule. let's keep it filtered for now

Issue #3343, opened 2018-04-02
NovorAdapter (NativeID, MS2, RT, Modification Name)


oliveralka:
(1) Conversion to mgf fails due to an incorrect conversion of the nativeID.
(2) MS2 spectra need to be filtered by hand (FileFilter).
(3) Round RT (Novor)
(4) modification not in "PSI-MS Name" fromat ( e.g. MTRQNAGQHLGLAMLC(Cam)SVPR -> MTRQNAGQHLGLAMLC(Carbamidomethyl)SVPR) - right now a few standard ones are parsed manually 

@lars20070 is in contact with the developer in regard to (3,4)

PR #3344, opened 2018-04-02
[FIX] NovorAdapter

oliveralka:
Will fix Issue (1,2) in #3343.

MascotGenericFile - get Number after last "=". This should be fine for most of the standard formats. 
SetLevels in MzML Options.


oliveralka:
Please not merge yet, there seems to be a parsing issue.

oliveralka:
New test files - real conditions;
extractScanNumber by using native_id_type_accession in MascotGenericFile

src/openms/source/FORMAT/MascotGenericFile.cpp
```diff
@@ -171,7 +173,6 @@ namespace OpenMS
     {
       os << "--" << param_.getValue("internal:boundary") << "\n" << "Content-Disposition: form-data; name=\"" << name << "\"" << "\n\n";
     }
-    else
```
oliveralka: was not intended will be fixed in next commit

src/utils/NovorAdapter.cpp
```diff
@@ -383,15 +387,22 @@ class TOPPNovorAdapter :
...
+      // v1.06.0634 (stable)
       vector<ProteinIdentification> protein_ids;
       StringList versionrow;
       csv.getRow(2, versionrow);
-      versionrow[0].suffix('#').trim();
-        
+      String version;
+      boost::smatch match;
+      boost::regex regexp("v.*");
+      if(boost::regex_search(versionrow[0], match, regexp))
```
cbielow: suggestion: this regex can be replaced by a simple
`version = versionrow[0].substr(versionrow[0].find("v."));`

src/utils/NovorAdapter.cpp
```diff
@@ -54,6 +54,8 @@
 #include <QtCore/QProcess>
 #include <fstream>
 
+#include <boost/regex.hpp>
```
cbielow: can be removed if regex goes (see below)

src/openms/source/FORMAT/MascotGenericFile.cpp
```diff
@@ -337,7 +338,7 @@ namespace OpenMS
            << "_" << spec.getNativeID() << "_" << filename << "\n";
         os << "PEPMASS=" << precisionWrapper(mz) <<  "\n";
         os << "RTINSECONDS=" << precisionWrapper(rt) << "\n";
-        os << "SCANS=" << spec.getNativeID().substr(spec.getNativeID().find("=")+1) << "\n";
+        os << "SCANS=" << SpectrumLookup::extractScanNumber(spec.getNativeID(), native_id_type_accession) << "\n";
```
cbielow: optimization: SpectrumLookup internally compares a lot of strings and is called for every spectrum.
It might be better to use an enum+switch. But maybe we should benchmark MGF writing first by simply timing FileConverter mgf->mgf

oliveralka:
Will try to fix it over the weekend.

oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

Issue #3311, opened 2018-03-15
Remove unused code/classes


cbielow:
candidates:
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1PepIterator.html
and subclasses

please add more here. We should have a short discussion during retreat and remove classes or TOPP tools accordingly.

timosachsenberg:
Inspect et al.

timosachsenberg:
maybe superhirn

timosachsenberg:
replace BinaryComposeFunctionAdapter.h use with lambda

cbielow:
fixed in #2560

PR #3347, opened 2018-04-02
Remove unused classes

cbielow:
as discussed in #3311 .


hroest:
looks good, what is the rationale for removing the additive series? I havent used it, but was it a useful tool? Was anybody using it?

cbielow:
no, it was there for a rather special purpose 12 years ago. according to usage stats nobdoy was using it.

PR #3351, opened 2018-04-03
Feature/mrm single transition

hroest:
allows usage of single-transition scoring model


timosachsenberg:
+1 if passes

timosachsenberg:
log stream only

PR #3352, opened 2018-04-04
[FEATURE] Update MRMFeatureFilter to support ComponentGroup new QCs

pasqoo:
The new supported QCs are: retention time, intensity, overall quality and metavalues.
Also, _MRMFeatureQC_ structs members now have default values, to avoid uninitialized random values.


src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -171,13 +197,12 @@ namespace OpenMS
               cg_qc_fail_message_vec.push_back("n_transitions");
             }
 
-            cg_tests_count += 6;
+            cg_tests_count += 9;
```
timosachsenberg: maybe ++cg_tests_count after each if block above? seems less error prone

PR #3360, opened 2018-04-04
Doc flags

cbielow:


jpfeuffer:
Looks good!

hroest:
@cbielow do you think it would make sense to have cmake fail if for example doxygen is not found? I have had issues with people where I told them to do `make doc` which actually fails silently if doxygen was not found and the target was disabled. I suggest therefore:

- fail hard with an error message if `make doc` target is not enabled
- fail hard during cmake and tell user to either install doxygen or disable docu

PR #3361, opened 2018-04-04
Updated contrib to include zlib1.2.11

jpfeuffer:
Builds with ALL on ubuntu 17.04+ should work again.
18.04s libz is 1.2.11, too.


hroest:
what happened to OSX ?

jpfeuffer:
There was a problem during the contrib build because the server with the contrib-sources was busy. I did not catch the error so the incomplete contrib was uploaded and used. I will restart that build so that the complete contrib is uploaded.

PR #3362, opened 2018-04-04
fix PercolatorAdapter

### LINKED ISSUE #3188: Percolator reports all peptides with 1 score in OpenMS 2.3 with labeled samples

lars20070:
Set dots in the AA sequence only, if no N/C terminal mods are present.
Fixes #3188.

### DIFF
--- a/src/topp/MSGFPlusAdapter.cpp
+++ b/src/topp/MSGFPlusAdapter.cpp
@@ -181,8 +181,9 @@ protected:
     registerIntOption_("matches_per_spec", "<num>", 1, "Number of matches per spectrum to be reported (MS-GF+ parameter '-n')", false);
     setMinInt_("matches_per_spec", 1);
 
-    registerFlag_("add_features", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')", true);
-
+    registerStringOption_("add_features", "<bool>", "true", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')");
+    setValidStrings_("add_features", ListUtils::create<String>("true,false"));
+    
     registerIntOption_("max_mods", "<num>", 2, "Maximum number of modifications per peptide. If this value is large, the search may take very long.", false);
     setMinInt_("max_mods", 0);
 
@@ -509,7 +510,7 @@ protected:
                    << "-minCharge" << QString::number(min_precursor_charge)
                    << "-maxCharge" << QString::number(max_precursor_charge)
                    << "-n" << QString::number(getIntOption_("matches_per_spec"))
-                   << "-addFeatures" << QString::number(int(getFlag_("add_features")))
+                   << "-addFeatures" << QString::number(int((getParam_().getValue("add_features") == "true")))
                    << "-thread" << QString::number(getIntOption_("threads"));
 
     if (!mod_file.empty())
--- a/src/topp/MSGFPlusAdapter.cpp
+++ b/src/topp/MSGFPlusAdapter.cpp
@@ -181,7 +181,7 @@ protected:
     registerIntOption_("matches_per_spec", "<num>", 1, "Number of matches per spectrum to be reported (MS-GF+ parameter '-n')", false);
     setMinInt_("matches_per_spec", 1);
 
-    registerStringOption_("add_features", "<bool>", "true", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')");
+    registerStringOption_("add_features", "<true/false>", "true", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')", false, false);
     setValidStrings_("add_features", ListUtils::create<String>("true,false"));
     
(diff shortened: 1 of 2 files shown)


timosachsenberg:
look good

timosachsenberg:
maybe add some comment to it

lars20070:
@algom `Percolator` now runs through on our dimethyl data. Let's test how much difference it makes after I am back.

lars20070:
rebuild jenkins

src/topp/MSGFPlusAdapter.cpp
```diff
@@ -181,7 +181,7 @@ class MSGFPlusAdapter :
     registerIntOption_("matches_per_spec", "<num>", 1, "Number of matches per spectrum to be reported (MS-GF+ parameter '-n')", false);
     setMinInt_("matches_per_spec", 1);
 
-    registerFlag_("add_features", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')", false);
+    registerFlag_("add_features", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')", true);
```
timosachsenberg: doesn't work that way. needs to be a string with true false

src/topp/MSGFPlusAdapter.cpp
```diff
@@ -181,8 +181,9 @@ class MSGFPlusAdapter :
     registerIntOption_("matches_per_spec", "<num>", 1, "Number of matches per spectrum to be reported (MS-GF+ parameter '-n')", false);
     setMinInt_("matches_per_spec", 1);
 
-    registerFlag_("add_features", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')", true);
-
+    registerStringOption_("add_features", "<bool>", "true", "Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter '-addFeatures')");
```
lars20070: @timosachsenberg That builds, but is not correct. Could you please point out an example?

timosachsenberg: something like in this snippet:
```
registerStringOption_("bla", "<true/false>", "true", "bla", false, false);
setValidStrings_("bla", ListUtils::create<String>("true,false"));
```

timosachsenberg: also update description to not talk about flag etc.

lars20070: Why? - Can I kick out the exception?

hroest:
:+1: for merge

Issue #2606, opened 2017-05-16
FileInfo output not consistent


linusb:
Using the FileInfo node on FeatureXML files reports many interesting properties. However, the tsv export (-out_tsv) does only contain very few of these and is especially missing the number of features in the run.

Example: 
normal command line output:
_Mem Usage while loading: 10 MB

-- General information --

File name: XXX.featureXML
File type: featureXML
Number of features: 2698

Ranges:
  retention time: 727.18 .. 7193.87
  mass-to-charge: 400.07 .. 648.38
  intensity:      35933.50 .. 5156480000.00

Total ion current in features: 5.78858e+10

Charge distribution:
  charge 1: 2298
  charge 2: 391
  charge 3: 4
  charge 4: 5

Distribution of peptide identifications (IDs) per feature:
  0 IDs: 2689
  1 IDs: 3
  2 IDs: 4
  3 IDs: 1
  4 IDs: 1

Unassigned peptide identifications: 9

FileInfo took 0.43 s (wall), 0.41 s (CPU), 0.00 s (system), 0.41 s (user)._

TSV_Output:
_file name	XXX.featureXML
file type	featureXML
retention time (min)	727.18
retention time (max)	7193.87
mass-to-charge (min)	400.07
mass-to-charge (max)	648.38
intensity (min)	35933.50
intensity (max)	5156480000.00_

The same holds for IdXML files (matched spectra and modifications are missing).

PR #3369, opened 2018-04-05
Fix/enrich fileinfo tsv output

sophiamersmann:
The tsv output file (*.csv) of FileInfo has been extended for featureXML and idXML formats.

In *.csv out files generated from featureXML files, the following fields were added:
* number of features
* total ion current in features
* charge distribution
* distribution of peptide identifications
* unassigned peptide identifications
* document ID
* statistics summaries for: intensities, feature FWHM in RT dimension, overall qualities, qualities in retention time dimension, qualities in mass-to-charge dimension

In *.csv out files generated from idXML files, the following fields were added:
* number of runs
* number of matched spectra
* number of modified top-hits
* document ID

Also, I slightly changed the format for better consistency with the command line output.
Implemented is a simple two-column format where the first and second column contain key and value, respectively. Information in the first column is organized into a colon-separated list.

See: #2606

### DIFF
--- a/src/topp/FileInfo.cpp
+++ b/src/topp/FileInfo.cpp
@@ -171,31 +171,31 @@ protected:
   template <class Map>
   void writeRangesMachineReadable_(const Map& map, ostream &os)
   {
-    os << "retention time (min)"
+    os << "general: ranges: retention time: min"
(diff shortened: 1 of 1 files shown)


sophiamersmann:
@timosachsenberg @oliveralka @jpfeuffer

hroest:
:+1: from me

PR #3370, opened 2018-04-06
[FEATURE] Additional scores for OpenSWATH

grosenberger:
This PR proposes a set of new scores for OpenSWATH/IPF. The implementation is based on [MIToolkit](https://github.com/Craigacp/MIToolbox), a small C++/Matlab toolkit released under the 3-clause BSD license that provides commonly used metrics from information theory.

To make these functions available in OpenSWATH, I suggest two alternative options:
- Adding MIToolbox to ``OpenMS/contrib``. I have implemented such a solution [here](https://github.com/grosenberger/contrib/tree/feature/mitoolbox).
- Adding MIToolbox directly to ``OpenSwathAlgo`` (this PR).

The benefit of the ``contrib`` approach is of course better modularity. On the other hand, this increases the maintenance efforts for all platforms agains. Further, MIToolbox is super small, so the overhead of directlu integrating this within ``OpenSwathAlgo`` would be minimal.

Which option is preferred by the core developers?


src/openswathalgo/include/OpenMS/OPENSWATHALGO/ALGO/Scoring.h
```diff
@@ -39,6 +39,9 @@
 #include <vector>
 
 #include <OpenMS/OPENSWATHALGO/OpenSwathAlgoConfig.h>
+#include <boost/lambda/casts.hpp>
+#include <boost/lambda/lambda.hpp>
+#include <OpenMS/OPENSWATHALGO/MITOOLBOX/MutualInformation.h>
```
hroest: I dont think this include is needed here, this should be all implementation

hroest:
there are a few issues to consider, one is maintenance of such a tool, the other question is whether we need the whole thing or can rip out parts of it. It is pretty small and we have done a similar approach with msnumpress. If the toolkit does not change a lot over time and is pretty stable, I would be happy with this current PR. Any comments @timosachsenberg @cbielow @jpfeuffer  ?

We also have to find ways of isolating it from the checkers, see http://cdash.openms.de/viewBuildError.php?buildid=264956 and  
```
The following tests FAILED:
	 11 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/ArrayOperations.cpp_cppcheck_test (Failed)
	 13 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/Entropy.cpp_cppcheck_test (Failed)
	 14 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/MutualInformation.cpp_cppcheck_test (Failed)
	 15 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/RenyiEntropy.cpp_cppcheck_test (Failed)
	 16 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/RenyiMutualInformation.cpp_cppcheck_test (Failed)
	 17 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/WeightedEntropy.cpp_cppcheck_test (Failed)
	 18 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/WeightedMutualInformation.cpp_cppcheck_test (Failed)
	934 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/ArrayOperations.cpp_cpplint_test (Failed)
	935 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/CalculateProbability.cpp_cpplint_test (Failed)
	936 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/Entropy.cpp_cpplint_test (Failed)
	937 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/MutualInformation.cpp_cpplint_test (Failed)
	938 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/RenyiEntropy.cpp_cpplint_test (Failed)
	939 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/RenyiMutualInformation.cpp_cpplint_test (Failed)
	940 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/WeightedEntropy.cpp_cpplint_test (Failed)
	941 - src/openswathalgo/source/OPENSWATHALGO/MITOOLBOX/WeightedMutualInformation.cpp_cpplint_test (Failed)
```

timosachsenberg:
its fine with me though I prefer single header-only dependencies inside the code base

jpfeuffer:
I like the header-lib approach in src/openms/thirdparty! +1

grosenberger:
That's a great suggestion, thanks @hroest! I will implement the changes accordingly.

grosenberger:
One further issue is the separation between the OpenSwathAlgo and OpenMS libraries. Currently, OpenSwathAlgo is built separately from OpenMS. However, if we add MIToolkit to OpenMS/thirdparty, we would have to link with OpenMS right?

hroest:
not if its header-only

src/openswathalgo/CMakeLists.txt
```diff
@@ -48,17 +48,20 @@ endif()
 # get information regarding the openswath files
 include(source/OPENSWATHALGO/OpenSwathAlgoFiles.cmake)
 
+#------------------------------------------------------------------------------
+# get information regarding the MIToolbox files
+set(MITOOLBOX_INCLUDE_DIRECTORY ../openms/thirdparty/MIToolbox/include)
```
grosenberger: These two variables are also set by the ``CMakeLists.txt`` file of ``openms/thirdparty``, however ``OpenSwathAlgo`` is built before ``OpenMS``. Is this the only way or is there a cleaner solution?

src/openswathalgo/source/OPENSWATHALGO/ALGO/Scoring.cpp
```diff
@@ -38,6 +38,12 @@
 
 #include <boost/numeric/conversion/cast.hpp>
 
+// Import dependencies from MIToolbox
```
grosenberger: Here we import all dependencies from MIToolbox.

hroest:
I agree that doing this in OpenSwathScoring adds additional complexity and maybe this is not the right way to go.

grosenberger:
> I agree that doing this in OpenSwathScoring adds additional complexity and maybe this is not the right way to go.

@hroest Could you please explain this in more detail? I have now moved MIToolbox to `src/openms/thirdparty/`, which is accessed by the OpenSWATH-specific scoring functions in `OpenSwathAlgo`. Maybe there are further CMake improvements possible, but this seems to work so far. If anybody has a better suggestion, I'm happy to implement the alternatives, else I would try to finish the last parts of this solution.

src/openswathalgo/CMakeLists.txt
jpfeuffer: Hmm yes. I think it would be cleaner the way it is done in eol_bspline. It basically creates a new target for that library and exports its include directories. Then I think both OpenSwathAlgo and OpenMS could refer to it.

grosenberger: I also liked the eol_bspline approach much better and tried to implement this. But then I ran into a circular issue, because the OpenMS target needs to be configured first to build MIToolbox, but also depends on OpenSwathAlgo, which depends on MIToolbox of OpenMS.

Is there a way to include the MIToolbox target in OpenSwathAlgo prior to configuring the OpenMS target?

jpfeuffer: Hmm I see. Why does it need OpenMS to be configured? Because it is in the OpenMS subfolder? In that case I think it would be fine to just make an openswathalgo/thirdparty folder.

grosenberger: Exactly, at least I couldn't figure out how to overcome the issue. I agree that putting it to `openswathalgo/thirdparty` might be a good option.

hroest:
> but this seems to work so far.

good enough for me :-)

hroest:
looks, good. Please merge and fix the tests

grosenberger:
I rebased on the current `OpenMS/develop` and fixed the tests. I further moved MIToolbox to `openswathalgo/thirdparty` to simplify the dependencies. The pyOpenMS issues seem unrelated, otherwise this is ready from my side.

hroest:
the failed run does seem to be related, see https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_docker_pyopenms/242/console
```
In file included from /home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openswathalgo/source/OPENSWATHALGO/ALGO/Scoring.cpp:42:0:
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openswathalgo/thirdparty/MIToolbox/src/ArrayOperations.c:19:33: fatal error: MIToolbox/MIToolbox.h: No such file or directory
 #include "MIToolbox/MIToolbox.h"
```

so it does seem that it compiles to `.c` file but the PATH is not properly set up here, but I wonder why it works for travis but not in the docker container. Maybe @jpfeuffer has some insight here?

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -685,6 +756,13 @@ namespace OpenMS
           }
         }
 
+        if (su_.use_mi_score_)
+        {
+          mrmfeature->addScore("var_mi_score", scores.mi_score);
+          mrmfeature->addScore("var_mi_weighted_score", scores.weighted_mi_score);
+          mrmfeature->addScore("var_mi_ratio_score", scores.mi_score  / (double)mrmfeature->getMetaValue("total_mi"));
```
hroest: can you check that this is not zero?

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -122,8 +122,12 @@ namespace OpenMS
     scores_to_use.setValidStrings("use_nr_peaks_score", ListUtils::create<String>("true,false"));
     scores_to_use.setValue("use_total_xic_score", "true", "Use the total XIC score", ListUtils::create<String>("advanced"));
     scores_to_use.setValidStrings("use_total_xic_score", ListUtils::create<String>("true,false"));
+    scores_to_use.setValue("use_total_mi_score", "true", "Use the total MI score", ListUtils::create<String>("advanced"));
+    scores_to_use.setValidStrings("use_total_mi_score", ListUtils::create<String>("true,false"));
     scores_to_use.setValue("use_sn_score", "true", "Use the SN (signal to noise) score", ListUtils::create<String>("advanced"));
     scores_to_use.setValidStrings("use_sn_score", ListUtils::create<String>("true,false"));
+    scores_to_use.setValue("use_mi_score", "true", "Use the MI (mutual information) score", ListUtils::create<String>("advanced"));
+    scores_to_use.setValidStrings("use_mi_score", ListUtils::create<String>("true,false"));
```
hroest: do you think it makes sense to turn them on by default? Does it improve our recall?

src/utils/OpenSwathWorkflow.cpp
```diff
@@ -820,6 +820,7 @@ class TOPPOpenSwathWorkflow
     {
       feature_finder_param.setValue("Scores:use_ms1_correlation", "true");
       feature_finder_param.setValue("Scores:use_ms1_fullscan", "true");
+      feature_finder_param.setValue("Scores:use_ms1_mi", "true");
```
hroest: should this be the default?

grosenberger:
Thanks for spotting the pyOpenMS issue. I have unrelated issues building pyOpenMS with OpenMS/develop on macOS 10.12, but I could not resolve the issue there yet.

In my tests, the MI scores also improve the sensitivity for standard OpenSWATH MS2 peak group scoring. But of course this comes at some computational costs and thus it might make sense to disable them by default. I will change this accordingly.

hroest:
I just tested it and on my Linux computer and on travis it builds, so that cannot be the issue.

grosenberger:
Thanks @jpfeuffer! I will try to finalize this.

grosenberger:
Ok, now every score can be turned on/off individually (default: off) and I added more checks and tests.

PR #3380, opened 2018-04-11
cexclude 11 header check that are not approved by google

timosachsenberg:


timosachsenberg:
http://cdash.openms.de/testDetails.php?test=48388653&build=265455

timosachsenberg:
https://github.com/google/styleguide/issues/194

PR #3385, opened 2018-04-12
[FIX] make Debugging:irt_mzml parameter functional

mffrank:
Until now OpenSwathWorkflow wrote the debugging output file containing iRT peptides in the current working directory with the fixed filename:  "debug_irts.mzML". The filename can now be set with the parameter -Debugging:irt_mzml


hroest:
failing test is unrelated

jpfeuffer:
Hmm I don't think so. There were many tests failing due to a failing build. 
VS does not implement the "and" "or".. keywords.
https://stackoverflow.com/questions/24414124/why-does-vs-not-define-the-alternative-tokens-for-logical-operators

jpfeuffer:
fixed in https://github.com/OpenMS/OpenMS/pull/3386

Issue #3348, opened 2018-04-02
Add numpy to automatic installable dependencies for pyOpenMS


jpfeuffer:
On mac I only notice that it needs numpy after I get a runtime error importing.
Can we add this to dependecies somewhere?
@hroest @grosenberger

hroest:
probably the best place is `setup.py` to add it as a dependency, right?

PR #3388, opened 2018-04-13
[Cleanup/Fix] Pyopenms build

### LINKED ISSUE #3373: zlib is not correctly exported or wrapped
Since standard python usually links to it. Forcing us to use a compatible version again.
I encountered problems when trying to link a static zlib on windows.
Just to let people know. Is this somewhere in the pyOpenMS docu?

Related:
https://stackoverflow.com/questions/16841674/python-c-extension-link-with-a-custom-shared-library

jpfeuffer:
In short:
- Copies and ships only what is dynamically linked and non-system
- Uses only one fix-dependency script on Mac now.
- Uses internal library targets to be copied (no hardcoded OpenMS filepaths)
- Fixes #3373  and #3348


jpfeuffer:
codacy is mainly a suggestion.

PR #3391, opened 2018-04-16
[FIX] Simpson and trapezoid integration for 1 and 2 pts cases

pasqoo:
The current implementation errors out when a peak with a number of points lower than 3 is encountered.
A more desirable behavior is to fallback to another integration method.

### Fixes proposed:
- if simpson and 2 pts: use trapezoid
- ~if simpson and 1 pt: use intensity_sum~ ~edit: `throw`ing exception~ area is `0`
- ~if trapezoid and 1 pt: use intensity_sum~ ~edit: `throw`ing exception~ area is `0`

This PR also fixes a bug that would let the user choose trapezoid even if the input had only 1 point.


pasqoo:
`LOG_WARN` is used to warn the user about the fallback behavior. The fallback happens only based on the number of points present in the input peak.

src/tests/class_tests/openms/source/PeakIntegrator_test.cpp
```diff
@@ -449,11 +459,23 @@ START_SECTION(PeakArea integratePeak(
...
+  pa = ptr->integratePeak(chromatogram, left, 2.478);
+  TEST_REAL_SIMILAR(pa.area, 881)
+  TEST_REAL_SIMILAR(pa.height, 881)
+  TEST_REAL_SIMILAR(pa.apex_pos, 2.477966667)
+
+  STATUS("Integration type: simpson (2 points)")
+  pa = ptr->integratePeak(chromatogram, left, 2.489);
+  TEST_REAL_SIMILAR(pa.area, 11.6081250000001)
+  TEST_REAL_SIMILAR(pa.height, 1384)
+  TEST_REAL_SIMILAR(pa.apex_pos, 2.488216667)
```
hroest: I feel like that is a big problem if the peak area is substantially larger when integrating a single peak compared to two peaks - e.g. I integrate a small area and I get `881` and then I integrate a larger area and I get `11.6`. Maybe the correct behavior would be to throw an exception instead of producing two peak area estimates that are completely incomparable. 

The other approach would be to at least take the spacing of the next peak into account and multiply by that ? Basically just compute a rectangle in that case where the x-axis base is the distance to the next peak?

hroest:
```
if simpson and 1 pt: use intensity_sum
if trapezoid and 1 pt: use intensity_sum
```

I feel that we will be mixing two different integration approaches that produce values on a very different scale, making them completely incomparable and introduce all sorts of problems down the line for quantitative comparisons. I dont think that this is an optimal choice

pasqoo:
Something we'd like to have is a way for this algorithm to not error out / throw an exception and just keep working (in the context of _MRMTransitionGroupPicker_).

Douglas proposed something like this:
for the case of trapezoid/simpson and 1 pt:
I take 1 more point on the left and one more on the right (assuming they both exist), take the average distance from the input point and use 1/100 of this value as a "base width" with which calculate an area.
Example:
input peak is: `1.2   1.3   1.4   1.5   1.6`
left: `1.35`, right: `1.45`
base width would be: `[(0.5 + 0.5) / 2 ] * 0.01`
Supposing the intensity on position `1.4` is `123456`, then the area would be `617.28`

In case additional points on left/right side don't exist, I'd just let the algorithm return `0`.

Would something like this work for you? @hroest @timosachsenberg 
If not, what path would you follow to have a fault-free implementation?

Thanks!

timosachsenberg:
My general take on this is that I think the low-level calculations should fail (e.g., throw an exception) if there is no single, best fall back. Instead, the caller should catch the exception and decide if in his context e.g. fall-back or program termination makes sense.
My feeling is, this switching the method or extending the region is in the realms of extrapolation/imputation. Question is, what you gain from that instead of just reporting a 0, ie., quantification was not successful. But I guess this is hard to answer because there might be situations where the single peak intensity is trustworthy enough.

hroest:
I agree with Timo that returning either zero or throwing an exception makes most sense. In principle we can make it the job of the calling function to ensure that there are at least 3 points in a peak (anyways, it does not make much sense to call something a peak if it has fewer data points). I dont really like the approach of dividing the area by 100 as people will take it at face value, I think 0 or NA is much more honest. 

However, to ensure that we dont actually abort, we could force the calling function to ensure that there are always 3 data points by taking the left/right neighbor in case there is for whatever reason only a single peak

hroest:
I agree with Timo that returning either zero or throwing an exception makes most sense. In principle we can make it the job of the calling function to ensure that there are at least 3 points in a peak (anyways, it does not make much sense to call something a peak if it has fewer data points). I dont really like the approach of dividing the area by 100 as people will take it at face value, I think 0 or NA is much more honest. 

However, to ensure that we dont actually abort, we could force the calling function to ensure that there are always 3 data points by taking the left/right neighbor in case there is for whatever reason only a single peak

pasqoo:
@hroest @timosachsenberg The PR is up to date.
It's not clear to me why LogStream failed.
I'm trying with the merge of `develop`, let's see.

pasqoo:
All tests succeeded :smile:

PR #3392, opened 2018-04-16
[FEATURE] PeakIntegrator: fit EMG peak model

pasqoo:
The implemented methods help in reconstructing a saturated and/or cutoff peak.
Both documentation and tests are provided to ensure the features work correctly.

From the docs:

> The method is able to recapitulate the actual peak area of saturated or cutoff peaks.
> In addition, the method is able to fine tune the peak area of well acquired peaks.
> The output is a reconstruction of the input peak. Additional points are often added
> to produce a peak with similar intensities on boundaries' points.

**Inspired by the results of:**
https://doi.org/10.1002/cem.1343
_Reconstruction of chromatographic peaks using the exponentially modified Gaussian function_
Yuri Kalambet, Yuri Kozmin, Ksenia Mikhailova, Igor Nagaev, Pavel Tikhonov

### New `public` methods:
`fitEMGPeakModel()` for both `MSChromatogram` and `MSSpectrum`

### New `protected` methods:
`extractTrainingSet()`
`computeMuMaxDistance()`
`computeInitialMean()`

### New `private` methods:
`emg_gradient_descent()`
`iRpropPlus()`
`Loss_function()`
`E_wrt_*()` for each parameter to optimize (h, mu, sigma, tau)
`compute_z()`
`emg_vector()`
`emg_point()`

### New class parameters:
`UInt print_debug_`
`UInt max_gd_iter_`
`bool compute_additional_points_`


hroest:
1. why not use the existing EMG fitter ? https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/EmgFitter1D.h 
2. this should for sure go into a different class

pasqoo:
1. The goal was to implement something similar to what the reference explains: https://doi.org/10.1002/cem.1343
The interface offered in this PR is very easy to use, because the user doesn't have to "help" the algorithm when it comes to parameters initialization or boundaries. It just accepts a `MSChromatogram` or `MSSpectrum` in input, and the rest is taken care of automatically. At least for the kind of data I had the chance to test it with :smiley: I hope the tests are good enough to ensure future use by different people or organizations.

2. I agree that all the new code could go to a new class. I felt this need when I noticed I had to test many private and protected methods, but I didn't want to clash with what was already present as a 1D fitter. How do you recommend doing this? Should I implement something like _EmgFitter1D\_GD_ and not inherit from _LevMarqFitter1D_?

hroest:
we have this one that runs through for pyOpenMS: https://travis-ci.org/OpenMS/OpenMS/builds/367548512?utm_source=github_status&utm_medium=notification which is good enough for me

hroest:
maybe @timosachsenberg has some comment on this. In either case, I would suggest to have the EMG model as a separate class and maybe even move the two into the same folder. However right now I dont see a great place to put them, I dont think TRANSFORMATIONS works well, maybe ANALYSIS?

timosachsenberg:
Would the https://github.com/OpenMS/OpenMS/tree/develop/src/openms/source/MATH/MISC folder work? There is already some fitting going on ;)

pasqoo:
Ok, I will move the code to `MATH/MISC` and call the new class _EmgGradientDescent_.
Open to better name proposals :smile:

timosachsenberg:
@hroest is the Math namespace fine with you?

hroest:
I like it, it is much cleaner now. I guess in the next PR we can move our previous EMG model to the same place @timosachsenberg ?

src/pyOpenMS/pxds/EmgGradientDescent.pxd
```diff
@@ -12,5 +13,5 @@ cdef extern from "<OpenMS/MATH/MISC/EmgGradientDescent.h>" namespace "OpenMS":
 
         void getDefaultParameters(Param&) nogil except +
 
-        void fitEMGPeakModel(const MSChromatogram& input_peak, MSChromatogram& output_peak) nogil except +
-        void fitEMGPeakModel(const MSSpectrum& input_peak, MSSpectrum& output_peak) nogil except +
+        void fitEMGPeakModel(MSChromatogram input_peak, MSChromatogram& output_peak) nogil except +
```
pasqoo: @hroest How come the removal of "&" from the input? And what about const? I thought they were both legal

hroest: you are right in this case it does not make a big difference. I general adding `&` can incur large overheads on the pyOpenMS wrappers because they copy data back even though the object did not change. Here both would work

pasqoo: What do you (or Cython team?) recommend doing when passing input data that doesn't change? Should I use `const` and a reference (`&`) or not?
Cythong docs don't really help me...

timosachsenberg:
looks good from my side

src/pyOpenMS/pxds/EmgGradientDescent.pxd
hroest: yes, but autowrap does not yet fully support `const ref` at the moment.

hroest: I am working on proper support for `const X &`, see https://github.com/uweschmitt/autowrap/pull/70

timosachsenberg:
So should we wait with this PR until this is done or merge now and adapt later?

hroest:
I think we should merge

@timosachsenberg we maybe should talk about how we could merge these methods that we now have for OpenSWATH data (mainly working on `MSChromatogram`) and feature data (working on `Feature` and `MassTrace`) so that people can use integration and FWHM techniques throughout OpenMS

timosachsenberg:
@hroest I think we could greatly benefit from consolidating the calculations throughout OpenMS. Might be kind of involved but can be done step-by-step. Are there some specific ones that you have in mind on the OpenSWATh side that the MassTrace class could benefit from? I was also thinking about the new EMG fits for incomplete traces that @pasqoo is developing.

hroest:
@timosachsenberg I am thinking about

- correlating mass traces (which is what we do a lot to find coeluting fragment ions) to find isotopic traces / charge states / adducts in MS1 
- peak model fitting such as EMG fits
- peak integration from mass traces / chromatograms where we now have multiple integration rules that are more accurate for quantification
- peak picking in mass traces / finding FWHM / splitting double peaks in chromatographic domain

PR #3399, opened 2018-04-18
[FIX] Try again to make non-QT libs private so that they are not link…

jpfeuffer:
…ed to dependents of OpenMS (e.g. superhirn)

### DIFF
--- a/cmake/add_library_macros.cmake
+++ b/cmake/add_library_macros.cmake
@@ -169,7 +169,7 @@ function(openms_add_library)
   if(openms_add_library_LINK_LIBRARIES)
     ## check for consistent lib arch (e.g. all 64bit)?
     check_lib_architecture(openms_add_library_LINK_LIBRARIES)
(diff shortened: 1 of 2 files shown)


jpfeuffer:
This is currently not possible since the Find modules of many libraries do not set the type of the found library (static or shared) and this is essential for the propagation rules of target_link_libraries.
We can:
1) adapt the find modules
2) write own CMake magic according to the filenames
3) add Cmake command-line options for every dependency (like for boost), if it should be found as shared or static and set accordingly

jpfeuffer:
We should re-evaluate this when we look at using imported targets (Boost and QT actually made this their default export now). But this won't be super easy since we use the xyz_LIBRARY variables to fill external files (e.g. for pyopenms or knime).
With imported targets, those variables would only include xyz::component as a location-independent reference. But also imported targets might not always specify if they are shared or static libraries (we have to check).

timosachsenberg:
what is the state of this PR?

jpfeuffer:
It probably needs a complete overhaul. See the linked issue.

timosachsenberg:
ok let's close for now...

jpfeuffer:
superseded by #5757

PR #3400, opened 2018-04-19
[FIX] enable scrolling again while preventing bug

hroest:
- re-enables scrolling in 3D view using the OpenGL Canvas
- works around a bug that caused black / black windows during scrolling
  or when opening multiple 3D windows

### DIFF
--- a/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
@@ -199,7 +199,7 @@ namespace OpenMS
       {
         openglwidget()->updateIntensityScale();
       }
-      // openglwidget()->initializeGL();
+      openglwidget()->initializeGL();
     }
     openglwidget()->resizeGL(width(), height());
     // openglwidget()->paintGL();
--- a/src/openms_gui/source/VISUAL/Spectrum3DOpenGLCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum3DOpenGLCanvas.cpp
@@ -180,8 +180,11 @@ namespace OpenMS
 
   void Spectrum3DOpenGLCanvas::initializeGL()
   {
-    makeCurrent();
-    glEnable(GL_DEPTH_TEST);
+    // The following line triggered a bug where the whole screen would turn
+    // black during scrolling (specifically it seems that multiple calls to
+    // this function causes the issue):
+    // glEnable(GL_DEPTH_TEST);
+
     glEnable(GL_BLEND);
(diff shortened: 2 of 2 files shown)


hroest:
The underlying issue seems to be that calling `paintGL();` and `initializeGL();` both call `glEnable(GL_DEPTH_TEST);` which is where I have narrowed down the problem.

src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
```diff
@@ -199,7 +199,7 @@ namespace OpenMS
       {
         openglwidget()->updateIntensityScale();
       }
-      // openglwidget()->initializeGL();
+      openglwidget()->initializeGL();
```
hroest: re-enabling this makes scrolling possible again in 3D (at least on Linux)

oliveralka:
Was not able to test on OSX. TOPPView is crashing when switching to 3D view.

```
macOS High Sierra
Version 10.13.4
```
```

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000248
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [0]
```

cbielow:
I tried this PR with mutliple 3D windows, changing forth and back, scrolling etc...
Seems fine on Windows 7, Qt. 5.6.2, MSVC2015

hroest:
@oliveralka was the crash caused by the patch or was 3D view broken before already?

src/openms_gui/source/VISUAL/Spectrum3DOpenGLCanvas.cpp
```diff
@@ -180,8 +180,11 @@ namespace OpenMS
 
   void Spectrum3DOpenGLCanvas::initializeGL()
   {
-    makeCurrent();
-    glEnable(GL_DEPTH_TEST);
+    // The following line triggered a bug where the whole screen would turn
+    // black during scrolling (specifically it seems that multiple calls to
+    // this function causes the issue):
+    // glEnable(GL_DEPTH_TEST);
```
hroest: basically what happened on Linux: when Spectrum3DCanvas.cpp:202 was commented out, then 3D view would work but scrolling was not possible. When not commented out, the scrolling would kindof work but the screen would go black and only if this line `glEnable(GL_DEPTH_TEST);` is commented out then it would all work.

PR #3401, opened 2018-04-20
Feature/cached ms2 spectra

hroest:
Load MS2 spectra on demand for viewing in TOPPView. This works because most MS2 spectral viewing is spectra-by-spectra and there is no need to have all spectra in memory at once.

This saves substantial amount of memory, depending on the data file. The new behavior is off by default but can be turned on through `.TOPPView.ini`. For some files this makes a huge difference, for a sample test file that I generated there is a 10x difference in memory consumption (note that I am loading the same file):

screenshot from 2018-04-20 00-17-33

The impact for regular users is minimal as this will have no effect unless turned on - but it could help in many cases where loading all data would be too memory-intensive. This is for example true for SWATH-MS data.

I have tested this on a HDD and an SSD drive with MS2 spectra that have 5000 peaks, there is no noticeable lag when opening a specific MS2 spectrum. Basically this is almost as fast as in memory for my test case.

### DIFF
--- a/doc/code_examples/Tutorial_GUI_Spectrum1D.cpp
+++ b/doc/code_examples/Tutorial_GUI_Spectrum1D.cpp
@@ -32,6 +32,7 @@
 #include <OpenMS/VISUAL/Spectrum1DWidget.h>
 #include <OpenMS/FORMAT/DTAFile.h>
 #include <OpenMS/KERNEL/StandardTypes.h>
+#include <OpenMS/KERNEL/OnDiscMSExperiment.h>
 #include <OpenMS/VISUAL/LayerData.h>
 
 using namespace OpenMS;
@@ -49,8 +50,9 @@ Int main(int argc, const char ** argv)
   exp.resize(1);
   DTAFile().load(tutorial_data_path + "/data/Tutorial_Spectrum1D.dta", exp[0]);
   LayerData::ExperimentSharedPtrType exp_sptr(new PeakMap(exp));
+  LayerData::ODExperimentSharedPtrType on_disc_exp_sptr(new OnDiscMSExperiment());
   Spectrum1DWidget * widget = new Spectrum1DWidget(Param(), nullptr);
-  widget->canvas()->addLayer(exp_sptr);
+  widget->canvas()->addLayer(exp_sptr, on_disc_exp_sptr);
   widget->show();
 
   return app.exec();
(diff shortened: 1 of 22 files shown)


timosachsenberg:
This is cool but I am worried that it adds some extra complexity to the code.
Wouldn't it be possible to wrap that in a class that handles spectra retrieval from disc and memory?
Doesn't need to have fancy additional logic (basically the same code but just moved to a class).

hroest:
>  Wouldn't it be possible to wrap that in a class that handles spectra retrieval from disc and memory?

this is basically what `LayerData` is doing, see the abstraction in `getCurrentSpectrum`

replacing the `MSExperiment` with a wrapper will be an almost monumental task, mainly because most of the code assumes that will get a `MSExeperiment` at some point

hroest:
Here I report a substantial improvement in useability: 

Loading a 21 GB highly compressed (zlib + numpress) file that has an index now takes 40 seconds on my SSD and it takes 150 MB of memory. The whole file uncompressed file is 93 GB on disk, so this is more or less the size that we would be looking at in memory, simply reading and converting the file with OpenMS takes ca 20 minutes and loading it into memory is out of the question on most machines. I therefore believe that this is the only option we currently have to display this data and I think the measurements are pretty convincing:

- 20 minutes (+) loading time -> 40 seconds
- 90+ GB in memory -> 150 MB

of course this only allows one to browse the data spectrum by spectrum but its better than not being able to load the data.

hroest:
Note: implementing this properly has proven to be a bit more difficult than anticipated. That has mostly to do with implicit assumptions in the code and leaky abstractions - mainly that a lot of code accesses the raw peak map and modifies it by sorting, computing ranges on it etc. I have started to fix this by removing the muteable `getPeakData` call and the `currentPeakData_` call that indirectly provides muteable access to peak data. However, removing all muteable access to peak data will be more challenging as the underlying `const shared_ptr< X >` will simply give muteable access to `X` to all callers even if they dont need it.

hroest:
now it also works for chromatograms (instead of crashing) and I have separated calls to the muteable interface from those that dont need a muteable peak map.

timosachsenberg:
I think the separate calls for mutable / immutable is nice. Maybe I would not give them different names (we usually also don't give different names for mutable / immutable getter in OpenMS)

hroest:
another issue of course is that `const shared_ptr< X >` is not actually a const reference and should not be labelled `const`

hroest:
@timosachsenberg @cbielow This is now ready to merge, I have also added the new options directly to the preference dialogue so that users can turn on the option from TOPPView directly.

PR #3403, opened 2018-04-20
MzTabExporter filetype

MFreidank:
This pull request addresses #1462. 
It adds `FileTypes::MZTAB`, makes `FileHandler::getTypeByFileName` aware of it by checking for substring `"MTD mzTab-version" `and changes `MzTabFile` and `MzTabExporter` to use this new file type as output type. 

I also added two passing tests for `MzTabExporter` - one with a `consensusXML` and another with an `idXML` input. 
 
An attempt to add a third test with a `featureXML` input failed. For `featureXML` input only, mzTabExporter adds a line starting with "uri" and containing the (path to) the given input file. 
This is a problem, because in cmake input files are specified using `${DATA_DIR_TOPP}/path/to/file` which expands to a full path. Thus, the output file during testing contains a string with the full path to the input file which causes my test to fail. Any ideas on how to proceed on this?


src/openms/source/FORMAT/FileHandler.cpp
```diff
@@ -409,6 +409,13 @@ namespace OpenMS
       }
     }
 
+    // mzTab file format
+    for (Size i = 0; i != complete_file.size(); ++i) {
+        if (complete_file[i].hasSubstring("MTD mzTab-version")) {
```
timosachsenberg: I think separator is not correct. should be \t between MTD and mzTab

src/tests/topp/CMakeLists.txt
```diff
@@ -2034,6 +2034,26 @@ add_test("TOPP_PeptideIndexer_14" ${TOPP_BIN_PATH}/PeptideIndexer -test -fasta $
...
+add_test("TOPP_MzTabExporter_1_out" ${DIFF} -in1 MzTabExporter_1_output.tmp -in2 ${DATA_DIR_TOPP}/MzTabExporter_1_output.mzTab)
+set_tests_properties("TOPP_MzTabExporter_1_out" PROPERTIES DEPENDS "TOPP_MzTabExporter_1")
+
+add_test("TOPP_MzTabExporter_2" ${TOPP_BIN_PATH}/MzTabExporter -test -in ${DATA_DIR_TOPP}/MzTabExporter_2_input.idXML -out MzTabExporter_2_output.tmp)
+add_test("TOPP_MzTabExporter_2_out" ${DIFF} -in1 MzTabExporter_2_output.tmp -in2 ${DATA_DIR_TOPP}/MzTabExporter_2_output.mzTab)
+set_tests_properties("TOPP_MzTabExporter_2_out" PROPERTIES DEPENDS "TOPP_MzTabExporter_2")
+
+# TODO: This test does not yet pass. 
+# Reason: For featureXML output, `mzTabExporter` adds a line starting with 
+# "uri" and containing the (entire user-specified) path to its input file
```
timosachsenberg: there is an option to ignore (whitelist) lines in the comparison.
See the other tests, they have similar problems

src/openms/source/FORMAT/FileHandler.cpp
MFreidank: Thanks, good catch. I fixed it.

PR #3404, opened 2018-04-22
[GUI] different default colors for the layers

hroest:
Changes the default colors in the 1D spectral view:

- the default color was switched to black here fa46276d6293b90bc09ae47 which I dont particularly like
- the colors for the additional layers were way to bright and hard to see, so I made them more dark
- in the future it would be nice to be able to control the pen width for the `DM_CONNECTEDLINES` settings

we may want to look into real color schemes, such as http://colorbrewer2.org/ or http://paletton.com/#uid=7000u0kllllaFw0g0qFqFg0w0aF

### DIFF
--- a/src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
@@ -86,7 +86,7 @@ namespace OpenMS
     //Parameter handling
     defaults_.setValue("highlighted_peak_color", "#ff0000", "Highlighted peak color.");
     defaults_.setValue("icon_color", "#000000", "Peak icon color.");
-    defaults_.setValue("peak_color", "#000000", "Peak color.");
+    defaults_.setValue("peak_color", "#0000ff", "Peak color.");
     defaults_.setValue("annotation_color", "#000055", "Annotation color.");
     defaults_.setValue("background_color", "#ffffff", "Background color.");
     defaults_.setValue("show_legend", "false", "Annotate each layer with its name on the canvas.");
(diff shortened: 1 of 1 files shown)


timosachsenberg:
@enetz I think this change to black as a default was originally introduced to have black peaks for unannotated ones and colored ones for peaks with annotations. (which I think makes sense) 
Should we have different defaults (e.g. black) in the identification view?

hroest:
I kind of liked the blue ones, but if there is a good reason we can switch to black. However, the other colors are definitely too bright and we should maybe change them.

Also it would be nice to allow people to set the line width, but to have fractional width properly displayed we probably need

```
painter->setRenderHint(QPainter::Antialiasing)
```

src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
```diff
@@ -86,7 +86,7 @@ namespace OpenMS
     //Parameter handling
     defaults_.setValue("highlighted_peak_color", "#ff0000", "Highlighted peak color.");
     defaults_.setValue("icon_color", "#000000", "Peak icon color.");
-    defaults_.setValue("peak_color", "#000000", "Peak color.");
+    defaults_.setValue("peak_color", "#0000ff", "Peak color.");
     defaults_.setValue("annotation_color", "#000055", "Annotation color.");
```
enetz: this default annotation color should be the same as the case 0 color in line 1072, right?

enetz:
I agree that the colors for annotated peaks were too bright, but when I made them darker it was not as easy to distinguish them from the black peaks. I guess that also depends on the monitor / projector.
I do not particularly care whether we use black or blue for the standard peak color, but when you have black and other colors it is obvious the colored ones are annotated somehow. If there are blue, green and red peaks, it is not as obvious that the blue ones are supposed to be the not annotated background.
If it is possible to make black the standard peak color only for annotated spectra or the ID View, that might be optimal.

timosachsenberg:
Let's merge this. I think we can set hard coded colors for the identification view

PR #3412, opened 2018-04-25
Feature/version info

hroest:
somehow the version is not displayed any more in TOPPView -> re-add this feature (this is useful when talking to collaborators about which versions they are working with).

Also, I feel that we should switch the naming convention right after a freeze to "x.y.alpha" for the new release where y = prev_release + 1. According to this we are now at `2.4.alpha` - I would also accept `pre-alpha`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,8 +95,8 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Windows")
 #------------------------------------------------------------------------------
 
 set(OPENMS_PACKAGE_VERSION_MAJOR "2")
-set(OPENMS_PACKAGE_VERSION_MINOR "3")
-set(OPENMS_PACKAGE_VERSION_PATCH "0")
+set(OPENMS_PACKAGE_VERSION_MINOR "4")
+set(OPENMS_PACKAGE_VERSION_PATCH "alpha")
 
 #------------------------------------------------------------------------------
 # retrieve detailed informations on the working copy (git)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Windows")
 
 set(OPENMS_PACKAGE_VERSION_MAJOR "2")
 set(OPENMS_PACKAGE_VERSION_MINOR "4")
-set(OPENMS_PACKAGE_VERSION_PATCH "alpha")
+set(OPENMS_PACKAGE_VERSION_PATCH "0-alpha")
 
 #------------------------------------------------------------------------------
 # retrieve detailed informations on the working copy (git)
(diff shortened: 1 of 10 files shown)


cbielow:
TOPPView still shows 2.3 in the splash screen (which we should get rid of entirely -- the number that is, the splash screen should stay)
But the PR looks good and is a step in the right direction.
+1 for merge

hroest:
I suggest that as soon as we have a feature freeze, we would label the `master` branch with `beta` and increase the version number on `develop.

jpfeuffer:
At a feature freeze, we usually branch off a release branch from which we do additional release builds, before merging it into master. But I agree, once we branched off for release, we can/should increase the version number in develop (maybe with a beta suffix).

cbielow:
python reports `Failure: ImportError (No module named pyopenms_1) ... ERROR`. this is probably unrelated, right?
Clang restarted

jpfeuffer:
Nope. related:
`pyopenms/pyopenms_3.cpp:120941:13: error: cannot convert 'OpenMS::String' to 'OpenMS::Int {aka int}' in assignment
   __pyx_t_1 = __pyx_v_self->inst.get()->version_patch;
             ^`

hroest:
ok I will have a look

hroest:
the good thing is that now also the Python packages are labelled as alpha releases

timosachsenberg:
before we merge this, I will probably need to update the REST server for the version checks. Currently expects Ints so he can compare versions

cbielow:
to you count developer installations as well in REST? Thats a good way to tell dev vs release then :)

hroest:
> before we merge this, I will probably need to update the REST server for the version checks. Currently expects Ints so it can compare versions

seems like this has become more involved than I originally wanted

jpfeuffer:
in theory we also have to check packaging. It uses the versions for the installer names etc.
Should work with strings though.

jpfeuffer:
And we wanted to change naming anyway, to include date or something, right?

hroest:
> in theory we also have to check packaging. It uses the versions for the installer names etc.
Should work with strings though.

maybe changing this to strings was not such a good idea - I am happy to revert this

hroest:
> And we wanted to change naming anyway, to include date or something, right?

either the date or git revision number for the nightly executables

timosachsenberg:
maybe we keep the three ints and add a suffix string? could be a date or "-alpha" etc.
seems also more compatible with https://semver.org/

hroest:
> maybe we keep the three ints and add a suffix string? could be a date or "-alpha" etc.

so basically `2.4.0-alpha` ?still leaves questions how different tools will deal with it the string - or we add a new field to VersionInfo in addition to patch level?

hroest:
PS: we are anyways not doing so great with semver:



Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. It MAY include minor and patch level changes. Patch and minor version MUST be reset to 0 when major version is incremented.

timosachsenberg:
yes if we want to add it then as an additional parameter so it does not interfere with other tools.

And its true that we are not close to semver. But could be an option in the future.

hroest:
@jpfeuffer @timosachsenberg looks ready from my side, any further comments?

PR #3415, opened 2018-04-29
[FIX] calculation for MassTrace estimateFWHM using intercept theorems

Zethson:
Hi,

excited to be submitting my first pull request to OpenMS!

# The issue
The issue with the former estimateFWHM implementation is that it always uses the start and end indexes just after the intensities passed the half_max intensity. Therefore they do **not** use the RT values exactly at the half_max intensity but somewhere past it.
This leads to some ugly lines in plots, since the FWHM clusters around common values past the half_max. 
See: Attached Image 1

# The proposed fix
The proposed fix uses an intercept theorem in order to calculate the exact RT values at the half_max intensity and not past it. 
Calculation is based on:
(yB - yA) / (xB - xA) = (yC - yA) / (xC - xA)
where yC corresponds to half_max and the rest are the respective RT and respective Intensity values.
If we solve this for xC:
xC = xA + ((yC - yA) * (xB - xA) / (yB - yA))
We get the x-coordinate corresponding to the half_max.
This corresponds to the RT, that we are looking for.

# Attachments/Images
The attached image 2 demonstrates a reduced formation of straight lines. I used the same dataset for both images, 50k of lines were plotted. The zoom level is about the same. 

Image 1 (old):
intensity_width

Image 2 (new!):
1_a_withnewalgo_intensity_width


hroest:
this is clearly an improvement :+1: I do have a few comments 

Can you fix the tests here: http://cdash.openms.de/viewTest.php?onlyfailed&buildid=267973

src/openms/include/OpenMS/KERNEL/MassTrace.h
```diff
@@ -321,6 +321,11 @@ namespace OpenMS
     /// median of trace intensities
     double computeMedianIntensity_() const;
 
+    /// calculate x coordinate of start/end indexes at half_max
+    /// calculation is based on (yB - yA) / (xB - xA) = (yC - yA) / (xC - xA)
+    /// solve for xC: xC = xA + ((yC - yA) * (xB - xA) / (yB - yA))
+    double calculateXCoordinateAtHalfMax_(double xA, double xB, double yA, double yB, double fwhm_int) const;
```
hroest: I suggest to change this to `linearInterpolationAtY_` and change `fwhm_int` to `y_eval` or `y` or something to make it more clear.

hroest:
also I suggest to change the docu  of `double estimateFWHM(bool use_smoothed_ints = false);` in the code to reflect the linear interpolation assumption

src/openms/include/OpenMS/KERNEL/MassTrace.h
Zethson: I'll try to fix the tests. 

Yeah, renamed that while debugging, will revert the name to something reflecting the calculation again.

src/tests/topp/FeatureFinderMetabo_2_noEPD_output.featureXML
```diff
@@ -527,7 +527,7 @@
 				<pt x="1.38960882" y="776.556250572984" />
 			</convexhull>
 			<UserParam type="string" name="label" value="T26_T5_T1"/>
-			<UserParam type="float" name="FWHM" value="43.5767288208008"/>
+			<UserParam type="float" name="FWHM" value="61.5109634399414"/>
```
hroest: why does the FWHM increase here? Looking at the code, it should only decrease - is there something wrong with my thinking?

src/openms/source/KERNEL/MassTrace.cpp
```diff
@@ -283,12 +283,12 @@ namespace OpenMS
...
   }
 
-    double MassTrace::calculateXCoordinateAtHalfMax_(double xA, double xB, double yA, double yB, double fwhm_int) const
+    double MassTrace::linearInterpolationAtY_(double xA, double xB, double yA, double yB, double y_eval) const
     {
       // no solution -> return an estimate
       if (std::fabs(xA - xB) == 0 || std::fabs(yA - yB) == 0)  { return xA; }
 
-      double xC = (xA + ((fwhm_int - yA) * (xB - xA) / (yB - yA)));
+      double xC = (xA + ((y_eval - yA) * (xB - xA) / (yB - yA)));
```
hroest: can you add a `POSTCONDITION` to check that `xC` is between `xA` and `xB` to make sure we are interpolating not extrapolating?

src/openms/source/KERNEL/MassTrace.cpp
```diff
@@ -283,12 +283,12 @@ namespace OpenMS
     return fwhm_;
   }
 
-    double MassTrace::calculateXCoordinateAtHalfMax_(double xA, double xB, double yA, double yB, double fwhm_int) const
+    double MassTrace::linearInterpolationAtY_(double xA, double xB, double yA, double yB, double y_eval) const
```
hroest: can you add a `PRECONDITION` to check that `y_ecal` is between `yA` and `yB` to make sure we are interpolating not extrapolating?

src/tests/topp/FeatureFinderMetabo_2_noEPD_output.featureXML
Zethson: It's possible that those are the edge cases. @timosachsenberg and me will take another look at it again next week.

src/openms/source/KERNEL/MassTrace.cpp
Zethson: Sure.

src/openms/source/KERNEL/MassTrace.cpp
Zethson: This preconditions is thrown if no trace point exists beforehand. 
If we do not want to allow this we should decide how to respond: 
Extrapolate or return 0. 

The new algorithm now correctly returns 0 for many sidecases where no value can be calculated. The old algorithm returned 0.'verysmall' values.
If we return 0 we can filter those values out later on.

What do you think should be done in this case?

hroest: well its quite simple, you cannot interpolate if there is no data point. Again, it is quite hard to think of a case where this would happen but it may happen if the right and left border are exactly equal because a peak is basically a single individual outlier peak. You should probably check for that case because I cannot see any other case where you would want to extrapolate

src/tests/topp/FeatureFinderMetabo_2_noEPD_output.featureXML
hroest: I dont see a a case where this could be correct, logically. If it cannot be correct then either the code is faulty or my reasoning is faulty. I am happy to be corrected.

src/openms/source/KERNEL/MassTrace.cpp
Zethson: I perfectly agree.
Basically what throws me off is the following:
https://imgur.com/a/BDUYnrq
y_eval is smaller than yA and yB, so no matter what you do with the algorithm, this does not make any sense.
Is the input data already wrong or is there something that I miss with my lacking MS data knowledge?

hroest: well can you also check whether this code works as expected?

```
     Size left_border(max_idx), right_border(max_idx);
 
     while (left_border > 0 && tmp_ints[left_border] >= half_max_int)
     {
       --left_border;
     }
 
     while (right_border + 1 < tmp_ints.size() && tmp_ints[right_border] >= half_max_int)
     {
       ++right_border;
     }
```

maybe the `max_idx` is already at 1.3e6 and there is simply not data point picked further to the right? Again, in this case we should not extrapolate but (i) fix the mass trace itself or (ii) simply take the rightmost data point. I think for both cases you should simply check that your leftmost / rightmost datapoint is below the half intensity.

timosachsenberg: and return 0 otherwise to indicate that we don't have a proper FWHM?

hroest: > and return 0 otherwise to indicate that we don't have a proper FWHM?

hmm I would not do that but simply return the leftmost / rightmost point that we have. I dont think zero is a good solution in that case

timosachsenberg: this will often lead to fwhm that are half the size as expected and in my view not much better than extrapolating (or in some cases even worse). But I have no strong feelings about that and we just need to decide if we want to throw them out

hroest: > this will often lead to fwhm that are half the size as expected and in my view not much better than extrapolating (or in some cases even worse). But I have no strong feelings about that and we just need to decide if we want to throw them out

I disagree, if you only capture half of the elution trace than your FWHM estimate is the least of your worries - also your quantification will be off etc. So unless we also set those to zero I dont think we should set FWHM to zero

timosachsenberg: Yes, I think we are on the same line and that's what I actually wanted to say with "need to decide if we want to throw them out". If we don't have a proper mass trace then we might consider setting FWHM, quality and quantification to 0

timosachsenberg: So should we, in these cases, go the more conservative way of setting everything to 0 to clearly indicate that something went wrong during mass trace extraction / quantification? Or choose something that could be kind of close to the FWHM (but also totally off in some cases).

hroest: > So should we, in these cases, go the more conservative way of setting everything to 0 to clearly indicate that something went wrong during mass trace extraction / quantification? Or choose something that could be kind of close to the FWHM (but also totally off in some cases).

In principle I agree but I feel that this is the wrong place to discuss because (i) we had this issue of incomplete masstraces being reported before and this is not the PR to address this issue but to actually improve OpenMS at another place and (ii) determining whether to throw out a masstrace should depend on other factors than the FWHM and the user should have control over which traces to throw out. Finally (iii) the best option could be to retrace our steps and improve the masstrace in case we find an issue like this

timosachsenberg:
@Zethson ok so we basically decided to use the distance between the border points without extrapolation.

Zethson:
new_plot

Above is the plot generated with the same fileset and also 50k points. 
The newest implementation passes all PRE/POST Conditions. 
I also added another POSTCONDITION to verify that the interpolation indeed produces lower fwhm values than the previous implementation.

Zethson:
Did the FeatureFinderMetabo_4 test recently get added (again)? It does not exist in my openMS build, even when I run ctest, all pass. 
I can see that test in the cmakelists: 
https://github.com/OpenMS/OpenMS/blob/12267c2060d1981bfbd016fde4e93d944b5b4f48/src/tests/topp/CMakeLists.txt
However the required files are not even present in my openMS build, I don't have the required files to run this test. 
```

[lheumos@aragorn-old openms-build]$ ctest -R TOPP_FeatureFinderMetabo_4 -V
UpdateCTestConfiguration  from :/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
Parse Config file:/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
Parse Config file:/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
Test project /home/lheumos/CLionProjects/OpenMS/openms-build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
No tests were found!!!

[lheumos@aragorn-old openms-build]$ ctest -R TOPP_FeatureFinderMetabo_4_out -V
UpdateCTestConfiguration  from :/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
Parse Config file:/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
Parse Config file:/home/lheumos/CLionProjects/OpenMS/openms-build/DartConfiguration.tcl
Test project /home/lheumos/CLionProjects/OpenMS/openms-build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
No tests were found!!!
```

hroest:
> Did the FeatureFinderMetabo_4 test recently get added (again)? It does not exist in my openMS build, even when I run ctest, all pass.

I suggest to merge with `develop` and ensure that all tests pass

Zethson:
"Build step 'Execute shell' marked build as failure"

The tests seem to run fine though.

timosachsenberg:
I think all points are addressed an the plots show a nice improvement

PR #3417, opened 2018-04-29
[FIX] review OpenSwath interfaces

hroest:
review the OpenSwath interfaces, create some improvements (e.g. const correctness)


src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -840,7 +840,7 @@ namespace OpenMS
       if (stop_report_after_feature_ >= 0 && i >= (Size)stop_report_after_feature_) {break;}
       output.push_back(feature_list[i]);
     }
-    transition_group = transition_group_detection;
+    transition_group = transition_group_detection; // TODO why is this here?
```
hroest: @grosenberger do you know why we are doing this? I dont quite understand the point of overwriting the `transition_group` here?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMAssay.h
```diff
@@ -325,12 +338,12 @@ namespace OpenMS
       TODO: why make copy of IonMapT ?
```
hroest: @grosenberger can you comment on whether we could change pass `DecoyIonMap` and `TargetIonMap` by reference?

grosenberger: Yes, we should change that, thanks for spotting this. They are only used as reference and should not be modified.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
hroest: ok great, I have changed it and now `transition_group` is `const`

hroest: nope, seems that its important to copy back the information...

grosenberger:
Great, thanks for doing this! 👍

hroest:
I think this can be merged, failing tests are unrelated

PR #3419, opened 2018-04-30
[FEATURE] FileFilter: New flag to allow only exact matches from id:sequences_whitelist

MFreidank:
Previously, when we specify `id:sequences_whitelist` for `FileFilter` sequences that match any 
substring from this whitelist are not discarded. 
This pull request adds a new flag `id:sequences_whitelist_enforce_exact_matches`. If this flag is set, only sequences that match a full sequence from our whitelist are not discarded. 
Default behavior of `id:sequences_whitelist` is left unchanged. 
New tests `FileFilter_45` and `FileFilter_46` pass and show: (1) When `id:sequences_whitelist_enforce_exact_matches` is not set, matching is done by substring as before (2) When the flag is set, full matches against the whitelist are necessary.


src/topp/FileFilter.cpp
```diff
@@ -356,10 +375,11 @@ class TOPPFileFilter :
...
 
     addEmptyLine_();
-    registerTOPPSubsection_("id", "ID options. The Priority of the id-flags is: remove_annotated_features / remove_unannotated_features -> remove_clashes -> keep_best_score_id -> sequences_whitelist / accessions_whitelist");
+    registerTOPPSubsection_("id", "ID options. The Priority of the id-flags is: remove_annotated_features / remove_unannotated_features -> remove_clashes -> keep_best_score_id -> sequences_whitelist (sequences_whitelist_enforce_exact_matches) / accessions_whitelist");
     registerFlag_("id:remove_clashes", "Remove features with id clashes (different sequences mapped to one feature)", true);
     registerFlag_("id:keep_best_score_id", "in case of multiple peptide identifications, keep only the id with best score");
     registerStringList_("id:sequences_whitelist", "<sequence>", StringList(), "keep only features with white listed sequences, e.g. LYSNLVER or the modification (Oxidation)", false);
```
hroest: can you update the docu for `id:sequences_whitelist` to better reflect that it matches substrings?

MFreidank: Yes, will do.

timosachsenberg:
Could later be Extended to support advanced regex etc. just by adding an extra method

MFreidank:
Seems like a good idea, but where would one best implement such a method? Seems like it might be relevant outside of FileFilter potentially.

src/topp/FileFilter.cpp
MFreidank: Not sure about the formulation, but I gave it a shot, feel free to refactor.

timosachsenberg:
Doesn't need to be implemented now.
What should be done now is to replace the flag with a:
id:sequence_comparison_method
and have "substring" and "exact" as options (substring being the default).

Later one *could* add "regex" as option too but this doesn't need to be done now.

MFreidank:
@timosachsenberg Done. Can you have a look if this can be merged?

src/topp/FileFilter.cpp
```diff
@@ -135,7 +135,42 @@ class TOPPFileFilter :
...
 
 private:
-  static bool checkPeptideIdentification_(BaseFeature& feature, const bool remove_annotated_features, const bool remove_unannotated_features, const StringList& sequences, const StringList& accessions, const bool keep_best_score_id, const bool remove_clashes)
+  static bool sequenceIsWhiteListed_(const AASequence& peptide_hit_sequence, 
+                                     const StringList& whitelist, 
+                                     const String& sequence_comparison_method) {
+    //loop over all sequence entries of the StringList
+    for (StringList::const_iterator seq_it = whitelist.begin(); seq_it != whitelist.end(); ++seq_it)
+    {
+      if (sequence_comparison_method == "substring") {
```
timosachsenberg: see coding convention... brackets go in new line

MFreidank: oh snap, yes, thanks. I have resolved all brackets/whitespace issues I could spot now.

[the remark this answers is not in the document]
MFreidank: Decided to go with option (1) and split into two loops each guarded by a single if.

src/topp/FileFilter.cpp
```diff
@@ -135,7 +135,42 @@ class TOPPFileFilter :
...
+        if (peptide_hit_sequence.toString().hasSubstring(*seq_it)
```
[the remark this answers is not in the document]
MFreidank: I implemented caching in 610d6a6

lars20070:
Merge?

cbielow:
good work

MFreidank:
@cbielow Thank you!

timosachsenberg:
Thank you @MFreidank

PR #3426, opened 2018-05-01
Package versions need to be numbers for NSIS

jpfeuffer:
More changes to come.


CMakeLists.txt
```diff
@@ -124,6 +125,19 @@ else()
   message(STATUS "  [CMake is tracking Git commits and branching. To disable use '-D GIT_TRACKING=OFF'.]")
 endif()
 
+# Additional versioning information for packaging
```
cbielow: Its a bit hard to read the the code and imagine what it does.
Can you document this a bit more extensively?
and give an example of the result in either case.

jpfeuffer: I can. Basically: If it is not built from a release branch it will add the branch name and the last change date to everything that can handle strings in the version (eg. installer filenames).

jpfeuffer:
Any further comments?

hroest:
looks good from my side - lets merge!

PR #3428, opened 2018-05-02
Splashscreen versions and About dialogs in TOPPView/TOPPAS

cbielow:
- prints the current OpenMS Version dynamically on the Splash screens (no need to change the png templates anymore). Example: 
image

 - unifies the about-dialogs for TOPPView and TOPPAS in one single place

follow-up to  #3412

( - cleans up some unused classes)


src/openms_gui/source/VISUAL/APPLICATIONS/GUITOOLS/TOPPAS.cpp
```diff
@@ -168,10 +171,14 @@ int main(int argc, const char** argv)
 
     a.connect(&a, SIGNAL(fileOpen(QString)), mw, SLOT(openToppasFile(QString)));
 
-    // Create the splashscreen that is displayed while the application loads
-    QSplashScreen* splash_screen = new QSplashScreen(QPixmap(":/TOPPAS_Splashscreen.png"));
+    // Create the splashscreen that is displayed while the application loads (version is drawn dynamically)
+    QPixmap qpm(":/TOPPAS_Splashscreen.png");
+    QPainter pt_ver(&qpm);
+    pt_ver.setFont(QFont("Helvetica [Cronyx]", 15, 2, true));
```
timosachsenberg: is this font available on all systems?

cbielow: yes, it should be. If not Font-matching kicks in http://doc.qt.io/qt-5/qfont.html#fontmatching
So this is safe

cbielow:
updates image templates @ https://github.com/OpenMS/images/commit/fc6dbcec7a193ac8dc104a60bcc05bfb7c6455ca

PR #3430, opened 2018-05-02
[FIX] pyOpenMS fix for macOS

grosenberger:
pyOpenMS currently doesn't build on my macOS system. Specifically, ``fix_dependencies.rb`` tries to patch ``__init__.py`` using ``handleDyLib()``. This fix ensures that only ``dylib`` or ``so`` files are being patched.


jpfeuffer:
Thanks! I tried to combine both fix_dependencies scripts into one and probably overlooked that. I thought the ruby one was more generic so I chose that.

hroest:
:+1:

PR #3432, opened 2018-05-03
[FIX] determine peak/profile type from meta data (if possible)

timosachsenberg:
if spectrum is annotated as centroided, we don't do a peak type estimation based on the data (which often goes wrong)


hroest:
:+1:

PR #3435, opened 2018-05-04
[REFACTOR] moved deisotoping into new class

timosachsenberg:
@enetz moved RNPxl deisotoping to a different class and reuse it in SimpleSearchEngine
Do you already have some tests for the protein-protein spectra deisotoping?

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/RNPXL/RNPxlDeisotoper.h
+++ b/src/openms/include/OpenMS/ANALYSIS/RNPXL/RNPxlDeisotoper.h
@@ -1,0 +1,71 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2017.
+//
+// This software is released under a three-clause BSD license:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of any author or any participating institution
+//    may be used to endorse or promote products derived from this software
+//    without specific prior written permission.
+// For a full list of authors, refer to the file AUTHORS.
+// --------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


timosachsenberg:
improved locking in slightly less simple search engine

enetz:
I have a simple test here:
https://github.com/enetz/OpenMS/blob/feature/xlms-tool/src/tests/class_tests/openms/source/OPXLSpectrumProcessingAlgorithms_test.cpp#L122

I use my XL-MS theoretical spectrum generator to produce spectra and deisotope these, but it is a very rudimentary test for now (especially since isotopic patterns are not really implemented in my generator, just the option for the 2nd isotopic peak). Using the normal generator without meta info would be a better idea.

timosachsenberg:
Thanks. I added the tests and replaced the generic NEUTRON_MASS_U with the more accurate C13C12_MASSDIFF_U

src/openms/include/OpenMS/ANALYSIS/RNPXL/RNPxlDeisotoper.h
```diff
@@ -0,0 +1,77 @@
...
+{
+
+
+class OPENMS_DLLAPI Deisotoper
+{
+  public:
+
+  /* @brief Detect isotopic clusters in a fragment spectrum.
+
+     Note: spectrum must must be sorted by m/z
```
hroest: can you add a PRECONDITION?

hroest:
the apple tests may be related, the simepl search engine tests failed: http://cdash.openms.de/viewTest.php?onlyfailed&buildid=268860

src/pyOpenMS/pxds/RNPxlDeisotoper.pxd
```diff
@@ -9,5 +9,14 @@ cdef extern from "<OpenMS/ANALYSIS/RNPXL/RNPxlDeisotoper.h>" namespace "OpenMS":
 
 # COMMENT: wrap static methods
 cdef extern from "<OpenMS/ANALYSIS/RNPXL/RNPxlDeisotoper.h>" namespace "OpenMS::Deisotoper":
-        void deisotopeAndSingleCharge(MSSpectrum & in, double fragment_tolerance, bool fragment_unit_ppm, int min_charge, int max_charge, bool keep_only_deisotoped, unsigned int min_isopeaks,  unsigned int max_isopeaks, bool make_single_charged, bool annotate_charge) nogil except + # wrap-attach:Deisotoper
+        void deisotopeAndSingleCharge(MSSpectrum & sepctra,
```
hroest: @timosachsenberg actually it would have been sufficient to only change the pxd :-) but its better and clearer that way

timosachsenberg: Except for the typo;)

PR #3437, opened 2018-05-06
[FEATURE] PeakPickerHiRes level automation

Zethson:
Hi,

this pull request introduces an auto mode for the ms_levels of the PeakPickerHiRes. 

The user will now be able to skip setting the ms_levels parameter when running the PeakPickerHiRes. If the user does not supply a ms_levels parameter the tool will now run in auto mode. In auto mode the tool will pick everything that hasn't been picked yet.
However, the user still has the option to supply a ms_levels parameter, just like before.

Up for discussion: 

- Instead of extracting the if statements and introducing a for loop for each case (auto/manual) I chose to keep one single for loop with all if statements in it for readability. The if checks expense is neglectable.

- Should the user get warned if his whole input had already been picked before? In auto mode this is completely silent now, although not an issue.


cbielow:
looks good to me.
The comment "manual mode" is a  bit confusing, since it applies to the second AND third block, but its ok.

One thing (which does not relate to this PR, but to PPHR) is staggering: chromatograms are always picked, no matter their type. Thus, if I run the tool twice, chromatograms will be broken. Ad hoc, I did not find a "spectrumtype" for chromatograms, is that the reason why its the way it is?

timosachsenberg:
@hroest what do you think. Would it make sense to also introduce a PROFILE / CENTROIDED for chromatogram peaks?

timosachsenberg:
+1
Existing workflows still will have level "1" as  default and should not break.

Issue #3382, opened 2018-04-12
Error in multithreading.cmake


bretttully:
On linux (at least, don't have other systems to hand) this `/DOPENMS_HAS_TBB` needs to be `-DOPENMS_HAS_TBB`

https://github.com/OpenMS/OpenMS/blob/207614542166080d373c5b874c0447ea32c18ee5/cmake/multithreading.cmake#L54

jpfeuffer:
Oh yes! Thanks for letting us know. I will fix it, but be aware that since it wasn't caught for a long time, it is probably not well tested.

hroest:
maybe @timosachsenberg @aiche remembers the time when we had TBB - it may be useful to use the TBB data structures even together with OpenMP - especially for some of our singletons! https://stackoverflow.com/questions/9448613/are-tbb-data-objects-usable-with-openmp

hroest:
>  Does this perhaps suggest a wider issue with the TBB build?

yes, mainly that it has not been used and tested since a couple of years!

Issue #957, opened 2014-08-21
Threadsafe OpenMS


hroest:
It would be nice if OpenMS would in general be threadsafe, e.g. if there were no side effects on global states when calling functions. Some of the main obstacles may be
- global variables (hope we dont have any of those)
- singletons and static members

To address singletons and static members we need to either i) remove them or ii) wrap them into an OpenMP `critical` section. 

Currently we seem to have 100+ static variables if I counted correctly:

```
$ grep -R "static " src/openms | grep -v "static void" | grep -v "static const"  | grep -v "//.*static" | grep -v "(" | grep -v NNLS | wc -l
115
```

Explanation of grep command
- remove all `static void` functions (static functions are not a problem)
- remove all `static const` - when it is not mutable its fine
- remove all static in comments
- remove all lines that have a opening paranthesis since they are likely static functions (some false positives here)
- remove NNLS because honestly, nobody will touch that thing

we are left with 115 occurences of static somethings which probably need to be looked at (plus 379 instances of static and an opening paranthesis `grep -R "static " src/openms | grep -v "static const"  |  grep -v "static void"  | grep -v "//.*static" | grep  "(" | wc` which are hopefully mostly static functions and thus not a problem). This issue will track the progress of making OpenMS threadsafe.

timosachsenberg:
could something like: https://github.com/facebook/folly/blob/master/folly/AtomicHashMap.h be part of a solution?

PR #3439, opened 2018-05-07
Feature/openms concurrency

### LINKED ISSUE #905: ResidueDB not thread safe.

### LINKED ISSUE #3431: Error with Singleton factory & OpenSwath
We are getting intermittent and unpredictable failures on runs when the EMG scoring is switched on that are coming from within the Singleton factory.
```
error message: the value 'EmgModel' was used but is not valid; This FactoryProduct is not registered!
```
My best guess is that this approach isn't thread safe and we've got a race condition where one thread is creating the singleton and the other is trying to use it?

https://github.com/OpenMS/OpenMS/blob/0f6f4e3c2ac9380eccaf6c7e74efec2b2c63ac13/src/openms/include/OpenMS/CONCEPT/Factory.h#L120

hroest:
- adds a common framework for managing concurrency in `CONCEPT/MultiThreading.h`
- implements a simple read-write lock using macros:
  - `OPENMS_UNIQUELOCK` indicates a section where only one (writer) thread can execute
  - `OPENMS_NONUNIQUELOCK` indicates a section where multiple (reader) threads can concurrently execute but not concurrently with a writer thread
- allows user to easily switch between different forms of mutexes (boost, STL)
- adds missing mutexes and adds multithreaded tests for several singletons:
  - ProteaseDB
  - RNaseDB
  - ModificationsDB
  - ElementDB
  - Factory
  - UniqueIdGenerator
  - MetaInfoRegistry
- fixes #905
- fixes #3431
- another step on the way to address #957

## Testing
while multithreading is pretty tough to test and debug, the following is a pretty straight forward approach:

```
while ctest -R Factory_test -V; do :; done
```

## Performance
Performance testing of a realistic sample is even harder, this PR basically just ensures that we wont produce any errors, unexpected segfaults etc when using the singletons above in a multithreaded environment. I did benchmark `boost::shared_mutex`, `std::mutex` and `pragma omp critical` against each other and it seems that for simple code where lots of code has a lock, `critical` performs best whereas for more complex code where only part of the code is locked, the `shared_mutex` performs best. Note there are some platform specific issues with `boost::shared_mutex`, see https://stackoverflow.com/questions/37303342/boostshared-mutex-slower-than-a-coarse-stdmutex-on-linux/37306432?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

### DIFF
--- a/cmake/multithreading.cmake
+++ b/cmake/multithreading.cmake
@@ -51,7 +51,7 @@ endif()
 
 if (TBB_FOUND)
   INCLUDE_DIRECTORIES(${TBB_INCLUDE_DIRS})
-  add_compile_options(/DOPENMS_HAS_TBB)
+  add_compile_options(-DOPENMS_HAS_TBB)
 endif()
 
 #------------------------------------------------------------------------------
--- a/src/openms/cmake_findExternalLibs.cmake
+++ b/src/openms/cmake_findExternalLibs.cmake
@@ -57,7 +57,7 @@ find_package(XercesC REQUIRED)
 
 #------------------------------------------------------------------------------
 # BOOST
-find_boost(iostreams date_time math_c99 regex)
+find_boost(iostreams date_time math_c99 regex system thread)
 
 if(Boost_FOUND)
   message(STATUS "Found Boost version ${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}" )
--- a/src/openms/cmake_findExternalLibs.cmake
+++ b/src/openms/cmake_findExternalLibs.cmake
@@ -57,7 +57,7 @@ find_package(XercesC REQUIRED)
 
 #------------------------------------------------------------------------------
 # BOOST
-find_boost(iostreams date_time math_c99 regex system thread)
+find_boost(iostreams date_time math_c99 regex thread)
 
 if(Boost_FOUND)
   message(STATUS "Found Boost version ${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}" )
--- a/src/openms/source/CHEMISTRY/ElementDB.cpp
+++ b/src/openms/source/CHEMISTRY/ElementDB.cpp
@@ -32,7 +32,9 @@
 // $Authors: Andreas Bertsch, Timo Sachsenberg $
 // --------------------------------------------------------------------------
 //
+
 #include <OpenMS/CHEMISTRY/ElementDB.h>
+
 #include <OpenMS/CHEMISTRY/Element.h>
 
 #include <OpenMS/DATASTRUCTURES/Param.h>
@@ -41,12 +43,16 @@
 
 #include <OpenMS/SYSTEM/File.h>
 
+#include <OpenMS/CONCEPT/MultiThreading.h>
+
 #include <iostream>
 
 using namespace std;
 
 namespace OpenMS
 {
+  STATIC_LOCK(mutex_edb) 
+
   ElementDB::ElementDB()
   {
     readFromFile_("CHEMISTRY/Elements.xml");
@@ -57,23 +63,45 @@ namespace OpenMS
     clear_();
   }
 
+  const ElementDB * ElementDB::getInstance()
+  {
+    static ElementDB * db_ = nullptr;
+
+    // unique lock (make sure we only create one instance)
+    OPENMS_UNIQUELOCK(mutex_edb, lock)
+
+    if (db_ == nullptr)
+    {
+      db_ = new ElementDB;
+    }
+    return db_;
+  }
+
   const Map<String, const Element*>& ElementDB::getNames() const
   {
+    // TODO: this is dangerous in a multithreaded environment, we cannot guarantee that these references are valid
+    OPENMS_NONUNIQUELOCK(mutex_edb, lock)
     return names_;
   }
 
   const Map<String, const Element*>& ElementDB::getSymbols() const
   {
+    // TODO: this is dangerous in a multithreaded environment, we cannot guarantee that these references are valid
+    OPENMS_NONUNIQUELOCK(mutex_edb, lock)
     return symbols_;
   }
 
   const Map<UInt, const Element*>& ElementDB::getAtomicNumbers() const
   {
+    // TODO: this is dangerous in a multithreaded environment, we cannot guarantee that these references are valid
+    OPENMS_NONUNIQUELOCK(mutex_edb, lock)
     return atomic_numbers_;
   }
 
   const Element* ElementDB::getElement(const String& name) const
   {
+    OPENMS_NONUNIQUELOCK(mutex_edb, lock)
+
     if (names_.has(name))
     {
       return names_[name];
@@ -90,6 +118,8 @@ namespace OpenMS
 
   const Element* ElementDB::getElement(UInt atomic_number) const
   {
+    OPENMS_NONUNIQUELOCK(mutex_edb, lock)
+
     if (atomic_numbers_.has(atomic_number))
     {
       return atomic_numbers_[atomic_number];
@@ -99,11 +129,13 @@ namespace OpenMS
 
   bool ElementDB::hasElement(const String& name) const
   {
+    OPENMS_NONUNIQUELOCK(mutex_edb, lock)
     return names_.has(name) || symbols_.has(name);
   }
 
   bool ElementDB::hasElement(UInt atomic_number) const
   {
+    OPENMS_NONUNIQUELOCK(mutex_edb, lock)
     return atomic_numbers_.has(atomic_number);
   }
 
--- a/src/openms/source/CHEMISTRY/ResidueDB.cpp
+++ b/src/openms/source/CHEMISTRY/ResidueDB.cpp
@@ -34,6 +34,7 @@
 //
 
 #include <OpenMS/CHEMISTRY/ResidueDB.h>
+
 #include <OpenMS/CHEMISTRY/ResidueModification.h>
 #include <OpenMS/CHEMISTRY/ModificationsDB.h>
 #include <OpenMS/CHEMISTRY/Residue.h>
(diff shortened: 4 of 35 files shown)


timosachsenberg:
Cool
I would be interested how speed drops in e.g. SimpleSearchEngine when run on a large database. Or alternatively, how digestions and conversion to a AASequence object performs in the single threaded and multi-threaded context vs. old implementation.

cmake/multithreading.cmake
```diff
@@ -51,7 +51,7 @@ endif()
 
 if (TBB_FOUND)
   INCLUDE_DIRECTORIES(${TBB_INCLUDE_DIRS})
-  add_compile_options(/DOPENMS_HAS_TBB)
+  add_compile_options(-DOPENMS_HAS_TBB)
```
hroest: fixes #3382

src/openms/cmake_findExternalLibs.cmake
```diff
@@ -57,7 +57,7 @@ find_package(XercesC REQUIRED)
 
 #------------------------------------------------------------------------------
 # BOOST
-find_boost(iostreams date_time math_c99 regex)
+find_boost(iostreams date_time math_c99 regex system thread)
```
hroest: this will link to additional boost libraries, I hope we are building those in contrib?

src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -1161,6 +1165,44 @@ START_SECTION([EXTRA] testing terminal modifications)
...
+  // Parallel execution of code:
+  // original code:
+  // 5e4 iterations -> 4.01 seconds with boost::shared_mutex
+  // 5e4 iterations -> 8.05 seconds with std::mutex
+  //
+  // refactored code that throws exception outside block:
+  // 5e4 iterations -> 8.55 seconds with omp critical
+  // 5e4 iterations -> 8.59 seconds with std::mutex
+  // 5e4 iterations -> 3.94 seconds with boost::shared_mutex
+  int nr_iterations (5e4), test (0);
```
hroest: > how digestions and conversion to a AASequence object performs

@timosachsenberg does this answer your question?

timosachsenberg: partly... also interested in the speed drop compared to the non-thread safe code .
E.g. parallel context but no proper locking

hroest: > partly... also interested in the speed drop compared to the non-thread safe code .

that comparison wont work because you get segfaults and uncaught Exceptions all around (e.g. code trying to add Modifications to the DB but iterators becoming invalid midway through the code etc...). There is a reason we call it *thread-safe* code :-)

timosachsenberg: I worked around the thread safety in SimpleSearchEngine by registering all Residues in a single-threaded context. So I am a bit curious how this now affects speed ;). But I can do it myself later.

src/openms/cmake_findExternalLibs.cmake
hroest: seems like we are not doing that for the contrib and for any of the test systems. @jpfeuffer  @timosachsenberg  any comments on how to handle this? should we use `std::mutex` instead of the more fine grained locking?

timosachsenberg: I would go with boost here. c11 and even c14 is lacking too many basic features. It’s afaik also much more stable.

jpfeuffer: I agree with Timo. And since we include boost already anyhow and it seems to be the fastest one, let's go with boost. Should be easy to add to the contrib as configure flags.

cmake/multithreading.cmake
hroest: I did look into whether the TBB data structures did give any benefit but I did not pursue this for this one reasons:

- we still need the singleton access, which means that we dont really want concurrent insertions and reads, rather we want to have sequential insertions and reads. Otherwise we would have multiple entries of the same residue in the singleton etc...
- TBB is great if you need to insert different element from multiple threads but not if you are potentially inserting the same element, there is no other way around this
- TBB could still help by making the check whether the element exists "atomic" and only then go into the read loop. I have not tested this so far
- all the code does compile right now for Residue and ModificationsDB if you simply swap out the OpenMS data structures with TBB concurrent data structures (this is why all the `auto` is now in the code to make it work seamlessly)

src/openms/source/CHEMISTRY/ResidueDB.cpp
```diff
@@ -452,29 +481,36 @@ namespace OpenMS
...
   const Residue* ResidueDB::getModifiedResidue(const String& modification)
   {
+    OPENMS_NONUNIQUELOCK(mutex_rdb, lock)
+
     // terminal mods. don't apply to residue (side chain), so don't consider them:
     const ResidueModification& mod = ModificationsDB::getInstance()->getModification(modification, "", ResidueModification::ANYWHERE);
     return getModifiedResidue(getResidue(mod.getOrigin()), mod.getFullId());
   }
 
   const Residue* ResidueDB::getModifiedResidue(const Residue* residue, const String& modification)
```
hroest: basically if this function would not modify ResidueDB and / or ModifcationDB then all access would be read-only and we would not need locks in the first place.

src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -1161,6 +1165,42 @@ START_SECTION([EXTRA] testing terminal modifications)
...
+#endif
+  for (int k = 1; k < nr_iterations + 1; k++)
+  {
+    auto aa = AASequence::fromString("TEST[" +  String(0.14*k) + "]PEPTIDE");
+
+#ifdef _OPENMP
+#pragma omp critical (add_test)
+#endif
+    {
+      test += aa.size();
```
cbielow: this should not be done within the loop (since it adds a constant overhead and biases the measurements). use omp-reduction instead.

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -57,23 +63,45 @@ namespace OpenMS
     clear_();
   }
 
+  const ElementDB * ElementDB::getInstance()
+  {
+    static ElementDB * db_ = nullptr;
```
cbielow: replacing the body with
```
static ElementDB db_;
return &db_;
```
is faster, easier and thread-safe (Myers Singleton)

src/openms/source/CHEMISTRY/ResidueDB.cpp
```diff
@@ -51,19 +53,39 @@ using namespace std;
...
   ResidueDB::ResidueDB()
   {
+    // private constructor that is called from getInstance which acquires the lock (no lock needed here)
     readResiduesFromFile_("CHEMISTRY/Residues.xml");
     buildResidueNames_();
   }
 
+  ResidueDB* ResidueDB::getInstance()
+  {
+    static ResidueDB* db_ = nullptr;
```
cbielow: same here. Myers singleton will suffice

cbielow:
If I get this right, by this implementation can only use one locking type for the whole library (supporting anything else would be hard work I guess). Now, its also generally not advised to mix and match multithreading frameworks (e.g. TBB and OMP). So there is actually no way of ensuring this right now, since the developer cannot know what frameworks are used by either functions he is using or what functions call his code.
So this looks like a serious issue to me.
We might be better off to write code which is read-only + myers singleton in threaded contexts. If code requires write access to singletons, could this be solved on the algorithm that manipulates the data, i.e. on the higher level, where locking might also be much more efficient (since it can be used targeted?)?
What are our current use-cases for write-access to singletons?

timosachsenberg:
I think the main problem is how we handle modifications/residues not contained in the initial database. All other use cases could potentially be worked around easily (e.g., using dependency injection).

hroest:
> If I get this right, by this implementation can only use one locking type for the whole library (supporting anything else would be hard work I guess).

in principle yes, but you can of course still do

```
#define USE_STD_MUTEX
#include <OpenMS/CONCEPT/MultiThreading.h> 
```

for your code if you want to.


> We might be better off to write code which is read-only + myers singleton in threaded contexts.

I agree, that would be the best way to handle it. I think we made quite a mess of the singletons. Optimally, the singleton would be a 1:1 read-only representation of an XML file that is somewhere on disk. 

> What are our current use-cases for write-access to singletons?

I think the main use-case is modifications that are not in the database, as Timo pointed out. I suggted somewhere that the solution to this could be to define a single "unknown modification" `ResidueModification` and then make the database read-only.

hroest:
@timosachsenberg @cbielow maybe I should prepare a separate PR that addresses #905
and #3431 and #3481 while we discuss this in more detail? Still, it will be surprising to people that parsing `AASequence` leads to segfaults ...

PR #3443, opened 2018-05-08
[FIX] fixes #3355 1D Tab in TOPPView broken

### LINKED ISSUE #3355: 1D tabs in TOPPView broken
When opening 1D spectra from 2D view, two tabs are opened instead of one, see screenshot.
screen shot 2018-04-04 at 13 07 46

timosachsenberg:


hroest:
Did you test whether the spectra view and identification view both work?

timosachsenberg:
Yes. I only quickly checked but they seemed to work

PR #3444, opened 2018-05-08
[FIX] removed extra call to widget update

timosachsenberg:


timosachsenberg:
we recently had either crashes or speed issues on mac. maybe its related to this extra call to update - can't currently test on mac though

hroest:
turns out this also solves #3445

timosachsenberg:
I will merge so I can quickly check tomorrow if this solves the segfaults. 3D view seems to be broken on mac anyways

PR #3446, opened 2018-05-08
Fix peaktypeestimator

cbielow:
this PR makes the PeakTypeEstimator very robust. Its works on all data tested (Orbi, FTMS, IonTrap),
even on very sparse spectra (which have been very problematic in the past) and extreme m/z ranges (where the inter-peak spacing usually becomes very wide).

Accuracy is 100% on the BSA example data (centroided) in share, PeakTypeTest data and my 16k Orbitrap spectra test set (profile & centroided).


timosachsenberg:
Cool. Did you test this on high-mass + high-charge isotope patterns (e.g., top-down?).

timosachsenberg:
And you could also use: https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/MetaProSIP_1_input.mzML
as test-case, if you want.
It contains a 50% 13C labeled peptide (unfortunately only charge 2 so I don't know if all 5 peaks fall into the 1 Da window)

cbielow:
good idea. I did not think of heavily labeled peptides.

MetaProSIP dataset has 57 centroid spectra, correct .-)

I tried on a top down dataset from pride. Profile spectra were correctly identified 99% of the time.
Centroid spectra 100%. But the test dataset was not very extensive.

hroest:
seems like there are 29 failing tests right now, can you have a look at these?

src/openms/include/OpenMS/FORMAT/PeakTypeEstimator.h
```diff
@@ -106,8 +106,9 @@ namespace OpenMS
         if (idx == -1) break;
 
         // check left and right peak shoulders and count number of sample points
-        PeakConstIterator it_max = data.begin() + idx;
-        PeakConstIterator it = it_max;
+        typedef typename std::vector<PeakT>::iterator PeakIterator; // non-const version, since we need to modify the peaks
```
timosachsenberg: would:
using PeakIterator = std::vector<PeakT>::iterator; 
work?

timosachsenberg: Actually I meant with the brackets but seems this now also works ;)

Issue #3413, opened 2018-04-26
TOPPView crashes when uses 3D View


oliveralka:
TOPPView is crashing when switching to 3D view.

```
macOS High Sierra
Version 10.13.4
```

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000248
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [0]
```

**Question from @hroest** 

@oliveralka so the 3D view was broken beforehand already? I guess that was due to the switch to Qt5 - do you know which Qt version was used?

I have similar issues on Windows with QT 5.9 to get the 3D view working, even though it does not crash, the window is just blank.

oliveralka:
@hroest I am using QT 5.10.1

lars20070:
FYI, any panel switches do lead to TOPPView crashes under macOS in my case.

hroest:
> I have similar issues on Windows with QT 5.9 to get the 3D view working, even though it does not crash, the window is just blank.

I also just ran into similar issues, in our case the window was black. It turned out to be an issue with the ANGLE library and how we compiled Qt (you have to use either `-opengl dynamic` or `-opengl desktop`).

hroest:
@timosachsenberg @oliveralka can we close this?

timosachsenberg:
Let’s close and reopen if not fixed on all systems.

jpfeuffer:
Fixed it on my mac, too. BUT: zooming is almost impossible. I see no pattern in what goes wrong it just arbitrarily zooms in and out when I touch the mouse wheel.

jpfeuffer:
At least it seems to scale correctly again, too.

hroest:
also it seems that in the 3D view, the mouse wheel is the only way to zoom and Ctrl - / Ctrl + do not work as they do in 1D and 2D view @timosachsenberg is that expected behavior?

@jpfeuffer does Ctrl and then selecting an area with your mouse work for zooming?

timosachsenberg:
Not sure would need to check the key bindings

hroest:
> Not sure would need to check the key bindings

[...]

Issue #3354, opened 2018-04-04
intensity scale in TOPPView 3D view broken


lars20070:
The scaling of intensities in 3D view is incorrect, see screenshot. Tested under macOS.
![screen shot 2018-04-04 at 12 24 35](https://user-images.githubusercontent.com/6149805/38305106-83dafe78-3803-11e8-8d03-60d7a2d75376.png)

timosachsenberg:
hi @lars20070 do you know if this is a regression or an old bug?

lars20070:
Regression.

jpfeuffer:
@timosachsenberg didn't you reimplement the renderText function? maybe it needs to consider scaling?

timosachsenberg:
works with current develop on my ubuntu.
A MacOS thing?

jpfeuffer:
Can reproduce it on macOS Sierra.
What happens is that the x and y axes scale and "zoom" when you scroll but the plot does not update (and therefore also the z axis does not scale).
Also: It does not cutoff the axes labels that are out of the plotting range after scaling. (Might be solved by a correct plot updating altogether)

hroest:
I had a similar issue with Ubuntu and I fixed it with #3400 - maybe have a look at `void Spectrum3DCanvas::update_` and the currently commented out code to see whether its necessary?

PR #3448, opened 2018-05-10
[FIX] check if OpenGL was properly initialized before calling initial…

timosachsenberg:
…izeGL manually

### DIFF
--- a/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
@@ -192,6 +192,10 @@ namespace OpenMS
   {
 #endif
 
+    // make sure OpenGL already properly initialized
+    QOpenGLContext *ctx = QOpenGLContext::currentContext();
+    if (!ctx || !ctx->isValid()) return;
+    
     if (update_buffer_)
     {
       update_buffer_ = false;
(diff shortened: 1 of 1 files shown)


timosachsenberg:
the newer mac driver seem to be more sensitive in regards to calling opengl functions before proper initialization. This should fix the issue.

timosachsenberg:
fixes https://github.com/OpenMS/OpenMS/issues/3413 on my mac

timosachsenberg:
tested on linux - also works

timosachsenberg:
might also fix https://github.com/OpenMS/OpenMS/issues/3354

src/openms_gui/source/VISUAL/Spectrum3DCanvas.cpp
```diff
@@ -192,6 +192,10 @@ namespace OpenMS
   {
 #endif
 
+    // make sure OpenGL already properly initialized
+    QOpenGLContext *ctx = QOpenGLContext::currentContext();
+    if (!ctx || !ctx->isValid()) return;
```
hroest: how do we deal with failure here? according to the docs:

```
The typical way to handle context loss scenarios in applications is to check via this function whenever makeCurrent() fails and returns false. If this function then returns false, recreate the underlying native OpenGL context by calling create(), call makeCurrent() again and then reinitialize all OpenGL resources.
```
http://doc.qt.io/qt-5/qopenglcontext.html#isValid

timosachsenberg: I don’t think we need to at this point. I think we now just ignore premature update calls.

hroest:
works on Linux

PR #3451, opened 2018-05-11
[FEATURE] Remove OpenGL dependency (use Qt)

jpfeuffer:
Qt was included outside of the macro which could lead to system includes prepended before our
contrib includes.

### DIFF
--- a/src/openms/CMakeLists.txt
+++ b/src/openms/CMakeLists.txt
@@ -107,9 +107,6 @@ endif()
 if (TBB_FOUND)
  list(APPEND OPENMS_DEP_LIBRARIES ${TBB_LIBRARIES})
 endif()
-if (MSVC) ## TODO check if necessary, does it need to go to OpenMS_GUI?
-	list(APPEND OPENMS_DEP_LIBRARIES opengl32.lib)
-endif()
 if (WITH_CRAWDAD) ## TODO check if still necessary
   list(APPEND OPENMS_DEP_LIBRARIES ${Crawdad_LIBRARY})
 endif()
--- a/src/openms/cmake_findExternalLibs.cmake
+++ b/src/openms/cmake_findExternalLibs.cmake
@@ -155,7 +155,7 @@ find_package(Qt5 COMPONENTS ${OpenMS_QT_COMPONENTS} REQUIRED)
 
 IF (NOT Qt5Core_FOUND)
   message(STATUS "QT5Core not found!")
-	message(FATAL_ERROR "To find a custom Qt installation use: cmake <..more options..> -D QT_QMAKE_EXECUTABLE='<path_to_qmake(.exe)' <src-dir>")
+  message(FATAL_ERROR "To find a custom Qt installation use: cmake <..more options..> -DCMAKE_PREFIX_PATH='<path_to_parent_folder_of_lib_folder_withAllQt5Libs>' <src-dir>")
 ENDIF()
 
 
--- a/src/openms_gui/CMakeLists.txt
+++ b/src/openms_gui/CMakeLists.txt
@@ -60,15 +60,6 @@ set(CMAKE_AUTOUIC ON)
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
-
-# The Qt5Widgets_INCLUDES also includes the include directories for
-# dependencies QtCore and QtGui
-include_directories(${Qt5Widgets_INCLUDE_DIRS})
-include_directories(${Qt5Svg_INCLUDE_DIRS})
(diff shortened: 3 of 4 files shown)


timosachsenberg:
+1

jpfeuffer:
Wait a little. I will add stuff for cleanup.

timosachsenberg:
feel free to merge yourself

jpfeuffer:
Did anyone ever think about how to include OpenGL correctly? It seems a bit redundant to include specific libraries additionally on top of QT. Especially on Windows I thought QT would choose at runtime which OpenGL to use.
I tried removing it on my Mac but it didn't work. The gl.h was not found. So I leave it for now.

hroest:
looks good, I will test that

hroest:
looks good, I will test that

hroest:
> Did anyone ever think about how to include OpenGL correctly?

well, but then we should link against `Qt5OpenGL_LIBRARIES` which we currently do not - as far as I see it, we either need to directly link against opengl or to the Qt5OpenGL libs but one of these is mandatory, otherwise there is no OpenGL. Currently on Linux we have:

```
$ ldd lib/libOpenMS_GUI.so  | grep -i gl
	libCgl.so.1 => /usr/lib/x86_64-linux-gnu/libCgl.so.1 (0x00007fba35f15000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007fba33c0d000)
	libGL.so.1 => /usr/lib/nvidia-370/libGL.so.1 (0x00007fba331c1000)
	libGLX.so.0 => /usr/lib/nvidia-370/libGLX.so.0 (0x00007fba2f0ae000)
	libGLdispatch.so.0 => /usr/lib/nvidia-370/libGLdispatch.so.0 (0x00007fba2edc5000)

$ ldd lib/libOpenMS_GUI.so  | grep -i qt
	libQt5Svg.so.5 => /usr/lib/x86_64-linux-gnu/libQt5Svg.so.5 (0x00007f8ff410c000)
	libQt5Network.so.5 => /usr/lib/x86_64-linux-gnu/libQt5Network.so.5 (0x00007f8ff3fb2000)
	libQt5Widgets.so.5 => /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5 (0x00007f8ff1d2e000)
	libQt5Gui.so.5 => /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5 (0x00007f8ff17e6000)
	libQt5Core.so.5 => /usr/lib/x86_64-linux-gnu/libQt5Core.so.5 (0x00007f8ff1310000)

$ ldd /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5 | grep -i gl
	libGL.so.1 => /usr/lib/nvidia-370/libGL.so.1 (0x00007f2205f5a000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f2204377000)
	libGLX.so.0 => /usr/lib/nvidia-370/libGLX.so.0 (0x00007f2203859000)
	libGLdispatch.so.0 => /usr/lib/nvidia-370/libGLdispatch.so.0 (0x00007f2203570000)

```

So we do get OpenGL and libGL through Qt but maybe we need to link directly since we make direct calls to the library? Also we do not currently link against `libQt5OpenGL.so` which may be a way for us to get rid of the direct GL calls and do everything through QT. Maybe @timosachsenberg can comment on this?

hroest:
@jpfeuffer are the failures related? Otherwise I would like to only merge 
d65b077

jpfeuffer:
Hm I see. Yes unfortunately I guess the failures are related to the changes in the OpenGL linking. Yes it's better to merge the include order first

jpfeuffer:
I removed OpenGL because we actually do not use Qts old OpenGL wrapper. We only use the new OpenGLwidgets which are part of QtGui.

timosachsenberg:
Not 100% sure but I think we should replace the direct calls to the OpenGL function with the QOpenGLFunctions equivalent:
"When making OpenGL function calls, it is strongly recommended to avoid calling the functions directly. Instead, prefer using QOpenGLFunctions (when making portable applications) or the versioned variants (for example, QOpenGLFunctions_3_2_Core and similar, when targeting modern, desktop-only OpenGL). This way the application will work correctly in all Qt build configurations, including the ones that perform dynamic OpenGL implementation loading which means applications are not directly linking to an GL implementation and thus direct function calls are not feasible."

timosachsenberg:
Not necessarily. Seems it is possible to derive the canvas from QOpenGLFunctions to make all members available

jpfeuffer:
yeah.

PR #3452, opened 2018-05-11
[FEATURE] FileFilter new flag: "f_and_c:make_unique"

MFreidank:
This pull request introduces a new flag `f_and_c:make_unique` for FileFilter. 

Our flag serves as a filter for unique sequence/charge state combinations. In a single (feature/consensus) map, peptides with the same sequence and charge states may appear. If this flag is set only the feature with the highest intensity will pass this filter. (Features without identification pass our filter automatically, a different flag is used to remove these if desired.).

In its current state, the flag only supports `featureXML`. We would like to extend its functionality to support  `consensusXML` as well but were not sure how a good design would look. One idea is to process a `std::vector<BaseFeature>` in `makeUnique` instead and perform the copying into a `featureMap` or `consensusMap` inside the main loop of `FileFilter`. Better ideas would be much appreciated however. 

This is only a rough draft towards the final feature. It works, but the implementation is suboptimal and not feature complete (e.g. `consensusXML`).

@lars20070 Could you have a look?
@ALL: Feedback on how to support `featureXML` and `consensusXML` side-by-side would be much appreciated. (as well as any input on better naming for the flag/function)


jpfeuffer:
Do you really want to remove the complete feature? Couldn't it be that a second annotation of this feature is the true generator of this feature then?

MFreidank:
@lars20070 This is only a rough first draft. Please have a look and see if this is what you had in mind.

lars20070:
@jpfeuffer For one of our workflows we require the most reliable (i.e. most intense) peptide features for each sequence/charge combination.

timosachsenberg:
can you reduce test file sizes?

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
   }
 
 private:
+  static FeatureMap makeUnique(const FeatureMap& feature_map)
```
timosachsenberg: private methods end with underscore

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
...
+    // 1. create a std::map `feature_set` that maps pairs of <charge, sequence> 
+    // to a pointer to the feature with the highest intensity for this sequence.
+    // 2. construct a FeatureMap that contains all unannotated features in `feature_map`
+    // and all features that are in `feature_set`.
+    
+    // Copy (unique) features from `feature_set` over into a fresh FeatureMap.
+    FeatureMap unique_features = feature_map;
+    unique_features.clear(false);
+
+    // A set of features disguised as a std::map.
```
timosachsenberg: a set of features disguised as a map - sounds fishy ;)

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
...
+    // Uniqueness criterion/key is a pair <charge, sequence> for each feature.
+    typedef std::map<std::pair<Int, AASequence>, const Feature*> FeatureSet;
+    FeatureSet feature_set;
+
+    // 1. create a std::map `feature_set` mapping pairs <charge, sequence> to a pointer to 
+    // the feature with the highest intensity for this sequence.
+    for (FeatureMap::const_iterator fm_it = feature_map.begin(); fm_it != feature_map.end(); ++fm_it)
```
timosachsenberg: consider using a reference in rang for loop

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
...
+    // A set of features disguised as a std::map.
+    // Uniqueness criterion/key is a pair <charge, sequence> for each feature.
+    typedef std::map<std::pair<Int, AASequence>, const Feature*> FeatureSet;
+    FeatureSet feature_set;
+
+    // 1. create a std::map `feature_set` mapping pairs <charge, sequence> to a pointer to 
+    // the feature with the highest intensity for this sequence.
+    for (FeatureMap::const_iterator fm_it = feature_map.begin(); fm_it != feature_map.end(); ++fm_it) 
+    {
+      const std::vector<PeptideIdentification> pep_ids = fm_it->getPeptideIdentifications();
```
timosachsenberg: copy necessary?

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
...
+      if (!pep_ids.empty())
+      {
+        if (pep_ids.size() != 1) 
+        {
+          throw OpenMS::Exception::IllegalArgument(__FILE__, __LINE__, __FUNCTION__, "Features may contain at most one identification. Run IDConflictResolver first to remove ambiguities!");
+        }
+
+        // Assumption: first hit returned by `getHits()` has always highest search engine score.
+        // TODO: Is this assumption reasonable or do we need to sort the hits first?
+        const PeptideHit highest_score_hit = pep_ids.begin()->getHits().front();
```
timosachsenberg: first hit might not exist

timosachsenberg: copy required?

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
...
+
+        // Pair <charge, sequence> of charge of the new feature and the sequence of its highest scoring peptide hit.
+        const std::pair<Int, AASequence> pair = std::make_pair(fm_it->getCharge(), highest_score_hit.getSequence());
+
+        // If a <charge, sequence> pair is not yet in the FeatureSet or new feature `fm_it` 
+        // has higher intensity than its counterpart `feature_set[<charge, sequence>]`
+        // store a pointer to `fm_it` in `feature_set`.
+        const FeatureSet::iterator feature_in_set = feature_set.find(pair);
+        if (feature_in_set == feature_set.end() || feature_in_set->second->getIntensity() < fm_it->getIntensity())
```
timosachsenberg: will cause exception if feature_in_set points on end

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
...
+        // TODO: Is this assumption reasonable or do we need to sort the hits first?
+        const PeptideHit highest_score_hit = pep_ids.begin()->getHits().front();
+
+        // Pair <charge, sequence> of charge of the new feature and the sequence of its highest scoring peptide hit.
+        const std::pair<Int, AASequence> pair = std::make_pair(fm_it->getCharge(), highest_score_hit.getSequence());
+
+        // If a <charge, sequence> pair is not yet in the FeatureSet or new feature `fm_it` 
+        // has higher intensity than its counterpart `feature_set[<charge, sequence>]`
+        // store a pointer to `fm_it` in `feature_set`.
+        const FeatureSet::iterator feature_in_set = feature_set.find(pair);
```
timosachsenberg: const iterator or const_iterator?

src/topp/FileFilter.cpp
lars20070: A `PeptideHit` without first hit? - Ok, we will test.

src/topp/FileFilter.cpp
```diff
@@ -135,6 +135,69 @@ class TOPPFileFilter :
...
+      {
+        // Maintain features without peptide identifications.
+        unique_features.push_back(*fm_it);
+      }
+    }
+
+    // 2. copy features from `feature_set` into our new FeatureMap.
+    for (auto const& element : feature_set) 
+    {
+      const Feature feature = *(element.second);
```
timosachsenberg: const reference removes one copy

src/topp/FileFilter.cpp
timosachsenberg: Could potentially be a template to be used with feature map and consensus map.
Consider using BaseFeature* as this is exists for both feature and consensus feature

lars20070: We thought so as well. We recently removed a lot of templates, but guess it makes sense here.

src/topp/FileFilter.cpp
timosachsenberg: sure - there can be PeptideIdentifications without hits

timosachsenberg:
I think Julianus has a valid point.
We have IDConflictResolver that, so far, removes conflicting IDs from single features.
Why not add this functionality there? I think it might be a better place than FileFilter.

MFreidank:
@timosachsenberg 
While moving our code to `IDConflictResolver` we encountered the following issue:
We would like to add a flag to enable/disable our new behavior, but `IDConflictResolver` is split in `IDConflictResolver` and `IDConflictResolverAlgorithm` -- with only the first parsing arguments. Making `IDConflictResolver` parse arguments (by inheriting from `DefaultParamHandler`) requires changes to preexisting code, since all of its methods are static, and we'd need an instance to parse arguments in its constructor.
We propose the following two steps:
1. Implement the current functionality in `IDConflictResolver` in this pull request, adding the flag in the  `IDConflictResolver.cpp` (instead of in `IDConflictResolverAlgorithm.cpp`) and passing its value to the algorithm.
2. In a seperate pull request, refactor the interface, e.g. making the static methods of `IDConflictResolverAlgorithm` non-static, inheriting from `DefaultParamHandler` and using an instance of the algorithm class in `IDConflictResolver`.
Does this seem reasonable or are we missing something?

timosachsenberg:
>Implement the current functionality in IDConflictResolver in this pull request, adding the flag in the IDConflictResolver.cpp (instead of in IDConflictResolverAlgorithm.cpp) and passing its value to the algorithm.

Yes, I would do it that way. Not really required to derive from DefaultParamHandler. Actually, if I can go easily without it I woull prefer not using the inheritance for various reasons.

>In a seperate pull request, refactor the interface, e.g. making the static methods of IDConflictResolverAlgorithm non-static, inheriting from DefaultParamHandler and using an instance of the algorithm class in IDConflictResolver.

Yes, this could be done in the future but has no high priority.

MFreidank:
@timosachsenberg 
I just pushed commits that implement these changes and address your code style concerns from the last one. Could you have a look?

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
```diff
@@ -81,7 +81,79 @@ class OPENMS_DLLAPI IDConflictResolverAlgorithm
   **/
   static void resolve(ConsensusMap & features);
 
+  /** @brief TODO DOKU
```
timosachsenberg: address TODO

MFreidank:
@timosachsenberg got you, will today (and maybe tomorrow)

lars20070:
The proposed/altered functionality makes both workflows and implementation more complex.

In order to get the most reliable peptide quants, we now have to run (1) current `IDConflictResolver` (remove multiple IDs) (2) new `IDConflictResolver` (remove low intensity IDs from features) and (3) `FilteFilter` (remove unannotated features). An altered implementation will also make the code [more complex](https://github.com/OpenMS/OpenMS/pull/3452/files#diff-9c3a51675da1366d6e9ef42fbc359288R150).

I would prefer to run (1,2,3) as single `make_quants_unique` method in `IDConflictResolver`. *Give me the most reliable peptide quants* is what the end user needs and easily understands. A Python script can always achieve what Julianus proposed.

timosachsenberg:
I don't think so. Note that there are many ways to make quants unique and reserving the method name just for the special use case you have in mind is probably not the best way. 

What about the discussion we had that adding a flag "remove_unannotated_features" to IDConflictResolver would also be an option so you don't need to run FileFilter afterwards?
That way you would just run the tool ONCE (select to remove multiple IDs per freature AND remove low intensity IDs over features of the same peptide) and the workflow would not get more complex.
 
I also don't see why the implementation would be more complex.

Where are you struggling? Maybe we can assist.

lars20070:
Ok. Calling `IDConflictResolver` only once is a solution. Instead of choosing between two mutually exclusive methods (1) and (2), as [implemented now](https://github.com/OpenMS/OpenMS/pull/3452/files#diff-8f89624bc5a79c36839909811522c6d3R115), `IDConflictResolver` will run always (1), and (2) and (3) are two optional flags.

Regarding implementation of (2), in addition to this [map](https://github.com/OpenMS/OpenMS/pull/3452/files#diff-9c3a51675da1366d6e9ef42fbc359288R98) we will construct a vector of feature pointers *not* in the map. From both map and vector we then construct the final output.

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
```diff
@@ -88,50 +88,58 @@ class OPENMS_DLLAPI IDConflictResolverAlgorithm
...
+          typename FeatureSet::iterator feature_in_set = feature_set.find(pair);
           if (feature_in_set != feature_set.end()) 
           {
+            // Identical (charge, sequence) key found. Remove annotations from either the old or new feature.
+            
             if (feature_in_set->second->getIntensity() < element.getIntensity()) 
             {
+              // Remove annotations from the old low-intensity feature.
+              std::vector<PeptideIdentification> pep_ids_empty;
+              feature_in_set->second->setPeptideIdentifications(pep_ids_empty);
```
timosachsenberg: please move them to unassignedPeptideIdentifications (see other IDConflictResolver code)

lars20070:
@timosachsenberg In the end, I left the `remove_unannotated_features` functionality out. I don't like that the functionality is duplicated in `FileFilter`. Just more parameters and confusion.

src/topp/IDConflictResolver.cpp
```diff
@@ -102,34 +102,52 @@ class TOPPIDConflictResolver :
...
 
+protected:
+
   void registerOptionsAndFlags_() override
   {
     registerInputFile_("in", "<file>", "", "Input file (data annotated with identifications)");
     setValidFormats_("in", ListUtils::create<String>("featureXML,consensusXML"));
     registerOutputFile_("out", "<file>", "", "Output file (data with one peptide identification per feature)");
     setValidFormats_("out", ListUtils::create<String>("featureXML,consensusXML"));
+    registerFlag_("make_unique", "A map may contain multiple features with both identical sequence and charge state. The feature with the highest intensity is very likely the most reliable one. When switched on, the filter removes the sequence annotation from the lower intensity features, thereby resolving the multiplicity. Only the most reliable features maintain annotated.");
```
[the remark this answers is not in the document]
timosachsenberg: Apart from these minor issues, overall, it looks pretty good and I think we are close to a merge. Thanks!

One might argue that the feature with lowest m/z deviation to the ID or lowest RT deviation might be better and it is hard to reason why one is better than the other. Proper resolving this might be much more complex than just taking the highest intensity one (e.g., by formulating a global optimization problem). Reserving the name make_unique is, thus, way too general to assume that we will not change it in the future (and which potentially will break many workflows). So I am a bit hesitant.

Maybe naming the parameter -resolve_features "highest_intensity" would lead to more stable tool interface as we could add more algorithms in the future. The name "feature" implies that we are actually resolving IDs of ambiguous features (modified sequence + charge) and not peptides.

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
```diff
@@ -81,7 +81,85 @@ class OPENMS_DLLAPI IDConflictResolverAlgorithm
...
+    (Features without any identification pass this filter automatically)
+  **/
+  template<class T>
+  static void makeUnique(T & map)
+  {
+    // unassigned peptide identifications in this map
+    std::vector<PeptideIdentification>& unassigned = map.getUnassignedPeptideIdentifications();
+    
+    // A std::map tracking the set of unique features.
+    // Uniqueness criterion/key is a pair <charge, sequence> for each feature.
```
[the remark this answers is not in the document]
timosachsenberg: see below... a different name might be better

src/topp/IDConflictResolver.cpp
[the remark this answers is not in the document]
jpfeuffer: I would also change it to "... maintain annotated with that peptide". It could still be annotated by other peptides.
And I agree that make_unique is too general. But I also don't like "resolve_features" because that actually sounds like you would always end up with only one annotation to a feature but it does it the other way: it makes the features unique to a peptidoform. 
I think there need to be two categories, one for resolution across peptides and one for resolution inside a feature.
One could argue, that some approach will try to optimize for both at the same time and probably also the order of the resolution will make a big difference, but it would be a first step.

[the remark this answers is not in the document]
timosachsenberg: I think the order could be:
In feature:
- removing labeled ids that map to wrong channel (aka multiplex resolution)
- removing all lower scoring ones (standard)
- remove same final ids from features - except for the highest intensity one

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
```diff
@@ -81,7 +81,85 @@ class OPENMS_DLLAPI IDConflictResolverAlgorithm
...
+
+      if (!pep_ids.empty())
+      {
+        if (pep_ids.size() != 1) 
+        {
+          throw OpenMS::Exception::IllegalArgument(__FILE__, __LINE__, __FUNCTION__, "Feature does contain multiple identifications.");
+        }
+
+        // Assumption: first hit returned by `getHits()` has always highest search engine score.
+        // TODO: Is this assumption reasonable or do we need to sort the hits first?
```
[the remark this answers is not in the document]
timosachsenberg: On the other hand - we should only have one hit at this point don't we? See my comment on the order below

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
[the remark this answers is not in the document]
lars20070: At the moment only modified sequences i.e. [`AASequence`](https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/RC/2.3.0/html/classOpenMS_1_1AASequence.html#details).

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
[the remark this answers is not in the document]
lars20070: We should have only one sequence annotation at this point ([this](https://github.com/MFreidank/OpenMS/blob/d221cb8db56df16b16b55c97cef0f786eb6d13f6/src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h#L107) should never happen). But this annotation can have multiple hits.

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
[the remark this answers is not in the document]
timosachsenberg: k please extend documentation on this point

src/openms/include/OpenMS/ANALYSIS/ID/IDConflictResolverAlgorithm.h
```diff
@@ -81,7 +81,85 @@ class OPENMS_DLLAPI IDConflictResolverAlgorithm
...
+            if (feature_in_set->second->getIntensity() < element.getIntensity()) 
+            {
+              // Remove annotations from the old low-intensity feature. But only after moving these annotations to the unassigned list.
+              std::vector<PeptideIdentification>& obsolete = feature_in_set->second->getPeptideIdentifications();
+              unassigned.insert(unassigned.end(), obsolete.begin(), obsolete.end());
+              std::vector<PeptideIdentification> pep_ids_empty;
+              feature_in_set->second->setPeptideIdentifications(pep_ids_empty);
+              
+              // Replace feature in the set.
+              feature_in_set->second = &(element);
```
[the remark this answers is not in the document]
timosachsenberg: but not that important to fix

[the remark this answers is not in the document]
lars20070: For the original functionality, pointers were more convenient here.
Change pointer --> index  necessary?

timosachsenberg:
@jpfeuffer comments
"And I agree that make_unique is too general. But I also don't like "resolve_features" because that actually sounds like you would always end up with only one annotation to a feature but it does it the other way: it makes the features unique to a peptidoform."

The way it is implemented now it does condense annotations first down to a single ID per feature (in default method before). Then the make_unqiue method checks for features with the same annotation and retains only one feature annotated.

src/topp/IDConflictResolver.cpp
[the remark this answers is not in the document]
jpfeuffer: I still think the description needs a "... maintain annotated with that peptide"

jpfeuffer:
What about "one_per_peptide" or "unique_for_peptides"?

timosachsenberg:
can you make the template method private and add two members (one for FeatureMap and one for ConsensusMap) like it is done for the resolve method. That way you can also add the pyOpenMS bindings.

timosachsenberg:
so now we just need to trim down the huge test files:  src/tests/topp/FileFilter_47_output.featureXML

PR #3453, opened 2018-05-11
Fix segfaults & other issues

### LINKED ISSUE #3345: tests create temp files in source tree
src/tests/class_tests/openms/data/20171013_HMP_C61_ISO_P1_GA1_UV_VIS_2.mzML
	src/tests/class_tests/openms/data/AbsoluteQuantitationMethodFile_out.csv
	src/tests/class_tests/openms/data/CsvFile_3.csv
	src/tests/class_tests/openms/data/CsvFile_4.csv

they should be created in the build tree instead.

cbielow:
I've been running the tests on Visual Studio in debug mode, and it found some quite alarming bugs, which are not caught by our CI/nightlies. We should think about running VS debug builds from time to time....
I fixed most things except for: This PR will make XFDR tests fail, but it only exposes the problem. 

 - fixes #3345 (temp test files in source tree)
 - fix less-operator in IDConflictResolver (needs to be strict weak ordering) @hendrikweisser: can you please verify it still does the right thing f762051 ?
 - fix segfault in LPWrapper (when adding empty row, the first element is accessed)
 - fix invalid set operation (intersection input needs to be sorted)
 - add assert (this will lead to failing XFDR tests, but highlights the issue) @lukaszimmermann: fb31e6f now exposes the segfaults, so XFDR tests will fail. Can you fix?
 - augments #3408 (which broke the script), not escaping special chars "()<>|"

### DIFF
--- a/src/openms/source/ANALYSIS/ID/IDConflictResolverAlgorithm.cpp
+++ b/src/openms/source/ANALYSIS/ID/IDConflictResolverAlgorithm.cpp
@@ -105,8 +105,11 @@ namespace OpenMS
   {
     // if any of them is empty, the other is considered "greater"
     // independent of the score in the first hit
-    if (left.getHits().empty()) return true;
-    if (right.getHits().empty()) return false;
+    if (left.getHits().empty() || right.getHits().empty()) 
+    { // also: for strict weak ordering, comp(x,x) needs to be false
+      return left.getHits().size() < right.getHits().size();
+    }
+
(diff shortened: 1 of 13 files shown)


hroest:
I agree, we should incorporate this into the nightly tests - maybe we should also do that with `AdressSanitizer`

cbielow:
I think this is ready for merge. the XFDR issue should be adressed in a separate PR.

PR #3454, opened 2018-05-13
[FEATURE]/ PeptideIndexer auto mode for decoy string + position

Zethson:
Hey,

this PR introduces an automode for the PeptideIndexer. When no decoy string and/or prefix/suffix string is provided, then the decoy string + position will now be automatically determined. 

We scan the identifiers for a preset of decoy strings and increase each occurence.
If the occurence of a decoy string is >80% we'll use it as the decoy string, if it occured less than 100% we display a warning. 
Furthermore we determine whether the decoy string is prefix/suffix and set it accordingly. 

Up for discussion:
If you know any additional commonly or rarely used decoy strings, please tell me about them. I'll then happily add them.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
@@ -55,7 +55,6 @@
 #include <atomic>
 #include <algorithm>
 #include <fstream>
-#include <boost/algorithm/string/find.hpp>
 #include <OpenMS/DATASTRUCTURES/StringUtils.h>
 
 namespace OpenMS
@@ -756,6 +755,7 @@ public:
                }
                else if (seq_lower.hasSuffix(pair.first))
                {
+                 // count observed (ignoring case)
                  decoy_count[pair.first].second++;
                  // store observed (case sensitive)
                  std::string seq_decoy = StringUtils::suffix(seq, pair.first.length());
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
@@ -55,6 +55,7 @@
 #include <atomic>
 #include <algorithm>
 #include <fstream>
+#include <boost/algorithm/string/find.hpp>
 
 namespace OpenMS
 {
@@ -128,7 +129,8 @@ public:
       PEPTIDE_IDS_EMPTY,
       DATABASE_CONTAINS_MULTIPLES,
       ILLEGAL_PARAMETERS,
-      UNEXPECTED_RESULT
+      UNEXPECTED_RESULT,
+      DECOYSTRING_EMPTY,
     };
 
     /// Default constructor
@@ -137,7 +139,94 @@ public:
     /// Default destructor
     ~PeptideIndexing() override;
 
-    /// forward for old interface and pyOpenMS; use run<T>() for more control
+     template<typename T>
+     const std::string findDecoyString(FASTAContainer<T> &proteins, std::vector<ProteinIdentification> &prot_ids, std::vector<PeptideIdentification> &pep_ids)
+     {
+       const size_t PROTEIN_CACHE_SIZE = 4e5;
+
+       while (true)
+       {
+         bool has_active_data = proteins.activateCache();
+
+         if (!has_active_data) break;
+         SignedSize prot_count = (SignedSize)proteins.chunkSize();
+
+         {
+           proteins.cacheChunk(PROTEIN_CACHE_SIZE);
+           for (SignedSize i = 0; i < prot_count; ++i)
+           {
+             const String& seq = proteins.chunkAt(i).identifier; //protein description?
+             boost::iterator_range<std::string::const_iterator> rng;
+
+             // remove all alphanumeric characters
+             seq.erase(std::remove_if(seq.begin(), seq.end(), [](char c) { return !std::isalpha(c); } ), seq.end());
+
+             // search for case insensitive occurence of the possible decoys
+             for (const auto &pair : decoy_occur_prefix_) {
+               if (boost::ifind_first(seq, pair.first))
+               {
+                 if (seq.hasPrefix(pair.first))
+                 {
+                   decoy_occur_prefix_[pair.first]++;
+                 }
+                 else
+                 {
+                   decoy_occur_suffix_[pair.first]++;
+                 }
+               }
+             }
+           }
+         }
+       }
+
+       // determine decoy
+       const std::string &determinedDecoyString = determinedDecoyString;
+
+       if (determinedDecoyString.empty())
+       {
+         LOG_ERROR << "Error: Unable to determine decoy string" << std::endl;
+         return DECOYSTRING_EMPTY; // function of course does not return ExitCodes
+       }
+
+       return determinedDecoyString;
+     }
+
+     std::string determineDecoyString() const
+     {
+       std::string determinedDecoyString;
+       int all_prefix_occur;
+       int all_suffix_occur;
+       for (const auto &pair : decoy_occur_prefix_)
+       {
+         all_prefix_occur += pair.second;
+       }
+
+       for (const auto &pair : decoy_occur_suffix_)
+       {
+         all_suffix_occur += decoy_occur_suffix_;
+       }
+
+       for (const auto &pair : decoy_occur_prefix_)
+       {
+         if (pair.second / all_prefix_occur > 0.5)
+         {
+           determinedDecoyString = pair.first;
+         }
+       }
+
+       for (const auto &pair : decoy_occur_suffix_)
+       {
+         if (pair.second / all_suffix_occur > 0.5)
+         {
+           determinedDecoyString = pair.first;
+         }
+       }
+
+       return determinedDecoyString;
+     }
+
+
+     /// forward for old interface and pyOpenMS; use run<T>() for more control
     inline ExitCodes run(std::vector<FASTAFile::FASTAEntry>& proteins, std::vector<ProteinIdentification>& prot_ids, std::vector<PeptideIdentification>& pep_ids)
     {
       FASTAContainer<TFI_Vector> protein_container(proteins);
@@ -819,6 +908,13 @@ protected:
 
     Int aaa_max_;
     Int mm_max_;
+
+    // common decoy string to number of occurences in fasta
+    // could use <std::pair<std::string, std::string> as keys???
+    std::map<std::string, int> decoy_occur_prefix_ = {{"decoy", 0}, {"rev", 0}, {"reverse", 0}, {"dev", 0}, {"iddecoy", 0},
+                                                  {"xxx", 0}, {"shuffle", 0}, {"shuffled", 0}, {"random", 0}};
+     std::map<std::string, int> decoy_occur_suffix_ = {{"decoy", 0}, {"rev", 0}, {"reverse", 0}, {"dev", 0}, {"iddecoy", 0},
+                                                    {"xxx", 0}, {"shuffle", 0}, {"shuffled", 0}, {"random", 0}};
 
   };
 }
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
@@ -202,7 +202,7 @@ public:
           return DECOYSTRING_EMPTY;
         }
 
-        proteins.reset(); // need to reset
+        proteins.reset(); // reset caching
       }
 
       //-------------------------------------------------------------
@@ -713,10 +713,10 @@ public:
        // Common decoy strings in fasta files. Used to determine if a protein is target/decoy automatically.
        // Note: decoy prefixes/suffices must be provided in lower case and must only contain alphanumeric characters
         MapDecoyStringToPrefixSuffixCount decoy_count = {{"decoy", {0,0}}, {"rev", {0,0}},
-                                                                 {"reverse", {0,0}}, {"dev", {0,0}},
-                                                                 {"iddecoy", {0,0}}, {"xxx", {0,0}},
-                                                                 {"shuffle", {0,0}}, {"shuffled", {0,0}},
-                                                                 {"random", {0,0}}};
+                                                          {"reverse", {0,0}}, {"dev", {0,0}},
+                                                          {"iddecoy", {0,0}}, {"xxx", {0,0}},
+                                                          {"shuffle", {0,0}}, {"shuffled", {0,0}},
+                                                          {"random", {0,0}}};
 
 
        // map case insensitive strings back to original case (as used in fasta)
@@ -728,7 +728,6 @@ public:
        {
          proteins.cacheChunk(PROTEIN_CACHE_SIZE);
          bool has_active_data = proteins.activateCache();
-
 
          if (!has_active_data) break;
          auto prot_count = (SignedSize)proteins.chunkSize();
@@ -768,7 +767,7 @@ public:
        }
 
        // DEBUG only: print counts of found decoys
-       for (auto & a : decoy_count) LOG_DEBUG << a.first << "\t" << a.second.first << "\t" << a.second.second << std::endl;
+       for (auto &a : decoy_count) LOG_DEBUG << a.first << "\t" << a.second.first << "\t" << a.second.second << std::endl;
 
        return determineDecoyStringFromCounts_(decoy_count, decoy_case_sensitive, decoy_string, is_prefix);
      }
@@ -794,7 +793,7 @@ public:
 
        if (all_prefix_occur == all_suffix_occur)
        {
-         LOG_ERROR << "Error: Unable to determine decoy string" << std::endl;
+         LOG_ERROR << "Unable to determine decoy string" << std::endl;
          return DECOYSTRING_EMPTY;
        }
 
@@ -816,7 +815,6 @@ public:
            }
 
          }
-
        }
 
        // if a decoy suffix occured at least 80% -> set it as suffix decoy
@@ -841,7 +839,7 @@ public:
 
        if (decoy_string.empty())
        {
-         LOG_ERROR << "Error: Unable to determine decoy string" << std::endl;
+         LOG_ERROR << "Unable to determine decoy string" << std::endl;
          return DECOYSTRING_EMPTY;
        }
 
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
@@ -56,6 +56,7 @@
 #include <algorithm>
 #include <fstream>
 #include <boost/algorithm/string/find.hpp>
+#include <OpenMS/DATASTRUCTURES/StringUtils.h>
 
 namespace OpenMS
 {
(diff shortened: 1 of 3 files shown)


src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -182,6 +185,25 @@ namespace OpenMS
     template<typename T>
     ExitCodes run(FASTAContainer<T>& proteins, std::vector<ProteinIdentification>& prot_ids, std::vector<PeptideIdentification>& pep_ids)
     {
+      // no decoy string provided? try to deduce from data
+      if (decoy_string_.empty())
+      {
+        LOG_INFO << "No decoy string provided. Trying to determine from data." << std::endl;
+        ExitCodes e = findDecoyString_(proteins, decoy_string_, prefix_);
```
cbielow: why return `ExitCodes`? there are only two outcomes: found, not-found, so a bool would be much more intuitive.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -182,6 +185,25 @@ namespace OpenMS
...
+        if (e == EXIT_SUCCESS)
+        {
+          String position = prefix_ ? "prefix" : "suffix";
+          LOG_INFO << "Using " << position << " decoy string '" << decoy_string_ << "'" << std::endl;
+        }
+        else
+        {
```
cbielow: If you return from the function, why not do 
``` 
  if (!findDecoyString....)
 {
   return DECOYSTRING_EMPTY;
  }
  LOG_INFO << "Using " << (prefix_ ? "prefix" : "suffix") << " decoy string '" << decoy_string_ << "'" << std::endl;
```
is much easier to read

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
 
-protected:
-    struct PeptideProteinMatchInformation
+ protected:
+
+     using MapDecoyStringToPrefixSuffixCount = std::map<std::string, std::pair<int, int>>;
+     using MapCaseInsensitiveToCaseSensitiveDecoyString = std::map<std::string, std::string>;
+
+     template<typename T>
+     ExitCodes findDecoyString_(FASTAContainer<T> & proteins,
```
cbielow: These are quite useful functions, and I think they deserve their own class. I would not associate them with PeptideIndexer.
Since `determineDecoyStringFromCounts` is called only once, you could think about merging the functionality into `findDecoyString()` and save some boilerplate code since many things can be counted/computed when iterating over the proteins already.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
       return EXECUTION_OK;
     }
 
-protected:
-    struct PeptideProteinMatchInformation
+ protected:
+
+     using MapDecoyStringToPrefixSuffixCount = std::map<std::string, std::pair<int, int>>;
```
cbielow: ThatsQuiteAMouthfulAndHardToRead :-)
How about `DecoyCounts`

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
+     using MapCaseInsensitiveToCaseSensitiveDecoyString = std::map<std::string, std::string>;
```
cbielow: same here. `LowerToObserved`?

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
+
+       // map case insensitive strings back to original case (as used in fasta)
+       MapCaseInsensitiveToCaseSensitiveDecoyString decoy_case_sensitive;
+
+       const size_t PROTEIN_CACHE_SIZE = 4e5;
+
+       while (true)
+       {
+         proteins.cacheChunk(PROTEIN_CACHE_SIZE);
+         bool has_active_data = proteins.activateCache();
```
cbielow: Just a comment: You use these on-time-variables quite often. Makes the code really hard to read IMHO
```
if (!proteins.activateCache())
{
  break; // easier?
}
```

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
+
+         {
+           for (SignedSize i = 0; i < prot_count; ++i)
+           {
+             String seq = proteins.chunkAt(i).identifier;
+             String seq_lower = seq;
+             seq_lower.toLower();
+
+             // remove all non-alphanumeric characters
+             seq_lower.erase(std::remove_if(seq_lower.begin(), seq_lower.end(),
```
cbielow: what is the rationale here?
Have you seen weird characters in the wild, which warrant this overhead?

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
+       }
+
+       if (all_prefix_occur == all_suffix_occur)
+       {
+         LOG_ERROR << "Unable to determine decoy string" << std::endl;
+         return DECOYSTRING_EMPTY;
+       }
+
+       // if a decoy prefix occured at least 80% -> set it as prefix decoy
+       for (const auto & pair : decoy_count)
```
cbielow: formatting: please attach the `&` to the type (no spaces), ie. `const auto& pair`

occurs in a few places

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
+       {
+         const std::string & case_insensitive_decoy_string = pair.first;
+         const std::pair<int, int> & prefix_suffix_counts = pair.second;
+         double freq_prefix = static_cast<double>(prefix_suffix_counts.first) / static_cast<double>(all_prefix_occur);
+         if (freq_prefix > 0.8)
```
cbielow: Does this mean, that finding a single random decoy string once, will make it the winner?
Mhh... usually one has 50% decoys + 50% foward, so there might be a better heuristic using the total protein counts?

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
+         {
+           is_prefix = true;
+           decoy_string = decoy_case_sensitive.at(case_insensitive_decoy_string);
+
+           if (prefix_suffix_counts.first != all_prefix_occur)
+           {
+             LOG_WARN << "More than one decoy prefix observed!" << std::endl;
+             LOG_WARN << "Using most frequent decoy prefix (" << freq_prefix * 100 <<"%)" << std::endl;
```
cbielow: this probably prints a bit ugly. Maybe cast to int to get a nominal percentage value.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +699,155 @@ namespace OpenMS
...
+       // if a decoy prefix occured at least 80% -> set it as prefix decoy
+       for (const auto & pair : decoy_count)
+       {
+         const std::string & case_insensitive_decoy_string = pair.first;
+         const std::pair<int, int> & prefix_suffix_counts = pair.second;
+         double freq_prefix = static_cast<double>(prefix_suffix_counts.first) / static_cast<double>(all_prefix_occur);
+         if (freq_prefix > 0.8)
+         {
+           is_prefix = true;
+           decoy_string = decoy_case_sensitive.at(case_insensitive_decoy_string);
```
cbielow: do you need bounds checking? why not `[]`?

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
hroest: I agree, you could change this to > 40% of the total proteins carrying the string (and the string being the most common one across all of them)

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: Another contributor suggested this. I personally don't mind, so that's up to you guys ;)

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: Yup, thanks. Noted.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: Input files seem to containt decoy strings with sometimes arbitrary amounts of '_' characters.
In order to compare things more easily I simply remove all non-alphanumeric characters. Since the identifiers are usually pretty small strings this should be fine?

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: I was told that there's a friendly war about this in OpenMS ;)
Will do.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: Since we're iterating over the decoy_count here, the first decoystring with > 80% occurence wins. So, the first random decoy string shouldn't just win, it would also have to satisfy the condition:
>80% of all pre/suf occurences of the decoy string are of this string

Just checking prefixes first here. 
However, if you guys feel like the total protein counts result in a more accurate heuristic - sure!

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: Will do.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: <code>[]</code> only works if you do not pass a const reference. I guess this is because it does not contain any preset values at this stage?

Up to you whether const reference + .at or no const reference and <code>[]</code>

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: See above.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: Fine with me. 
If any other people think that this deserves its own class, then I'll gladly outsource it.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
jpfeuffer: I would vote for DecoyStringToAffixCount. The "To" implies a Map and Affix combines Suffix and Prefix.

Zethson:
Thank you for your in depth review!
I did not yet move the function into a new class, if other people suggest this as well, then I will do so.

Now only affixes which satisfy:
80% of all specific affix + 40% of all proteins
get used as decoy string

Furthermore I identified an issue with sidecase of the nature: 'reverse' and 'rev". Here 'reverse' and 'rev' get incremeted, so I added a sidecase for this.

However: I'm looking into a more general solution with some regex black magic.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
cbielow: how would these underscore cases look like?
e.g. "rev____albumin" would be detected in any case. So there are people out there who use "_rev____albumin"?
If thats the case, ok.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +692,143 @@ namespace OpenMS
...
+       for (auto &a : decoy_count) LOG_DEBUG << a.first << "\t" << a.second.first << "\t" << a.second.second << std::endl;
+
+       if (all_prefix_occur == all_suffix_occur)
+       {
+         LOG_ERROR << "Unable to determine decoy string!" << std::endl;
+         return false;
+       }
+
+       // side case: prefix_reverse == prefix_rev -> if reverse occurs, then rev will be found as well -> remove rev findings
+       if (decoy_count["rev"].first == decoy_count["reverse"].first)
```
cbielow: the `==` might break. Imagine the protein `REV1_HUMAN` (yes, it exists).
the safest option would be to include the underscores, e.g. `rev_`, which would make it unique, but I guess one cannot rely on underscores being used?!
What if you check for `reverse` first and if its found, stop iterating over the other affixes? Then you'd have at most a few spurious hits for `rev`, but`reverse` would still win if its truly present and you dont need to handle the special case any further.

Zethson: This sidecased has been removed. 
See: https://github.com/OpenMS/OpenMS/pull/3454/commits/c4f95fbef454a9c548811bbd737f8fc6ba7f98dc

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: I have observed cases like:
__id_decoy__blabla

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +692,162 @@ namespace OpenMS
...
+       // Setup regex vector, sort descending
+       std::vector<std::string> prefix_regex;
+       for (const auto& pair : decoy_count)
+       {
+         prefix_regex.push_back(pair.first);
+       }
+       std::sort(prefix_regex.rbegin(), prefix_regex.rend());
+
+       // Setup regex string -> create prefix- and suffix pattern to match to
+       std::string regexstr_prefix = "^(";
```
cbielow: how about `std::string regexstr_prefix = "^(" + ListUtils::concatenate<String>(prefix_regex, "|") + ")";` :)

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +692,162 @@ namespace OpenMS
...
+             seq_lower.toLower();
+
+             // remove all non-alphanumeric characters
+             seq_lower.erase(std::remove_if(seq_lower.begin(), seq_lower.end(),
+                                            [](char c) { return !std::isalpha(c); }), seq_lower.end());
+
+
+             boost::smatch sm;
+             // search for prefix
+             bool found_prefix = boost::regex_search(seq_lower, sm, pattern_prefix);
```
cbielow: regex is fine, since it solves the rev/reverse problem, I guess.
But I've seen regex queries take longer than the actual peptideindexing procedure. So could you please provide some numbers on how long this takes on a larger FASTA file? Thanks

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: That's pretty!

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +692,156 @@ namespace OpenMS
...
+       for (const auto& pair : decoy_count)
+       {
+         prefix_regex.push_back(pair.first);
+       }
+       std::sort(prefix_regex.rbegin(), prefix_regex.rend());
+
+       // Setup regex string -> create prefix- and suffix pattern to match to
+       std::string regexstr_prefix = std::string("^(") + ListUtils::concatenate<std::string>(prefix_regex, "|") + ")";
+       // suffix: remove leading '^', add latter '$'
+       std::string regexstr_suffix(regexstr_prefix.begin() + 1, regexstr_prefix.end());
```
timosachsenberg: this is as fast as creating it from scratch (like you did with regexstr_prefix) so I would go with a easier readable one line expression

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -677,8 +692,156 @@ namespace OpenMS
...
+       // Setup regex vector, sort descending
+       std::vector<std::string> prefix_regex;
+       for (const auto& pair : decoy_count)
+       {
+         prefix_regex.push_back(pair.first);
+       }
+       std::sort(prefix_regex.rbegin(), prefix_regex.rend());
+
+       // Setup regex string -> create prefix- and suffix pattern to match to
+       std::string regexstr_prefix = std::string("^(") + ListUtils::concatenate<std::string>(prefix_regex, "|") + ")";
```
timosachsenberg: check if variables can be declared const

Zethson: Missed this, thanks

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: Sure.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
Zethson: -- done [took 2.67 s (CPU), 2.68 s (Wall)] -- 
Using suffix decoy string 'rev'
Mapping 4974 peptides to 173450 proteins.

Ran on a thinkpad t440s with an i54200U -> 2,6 GHz max 
4 Gig Ram

Might be faster if I close chromium + Clion.

cbielow:
ok. looks all good to me. last but not least: a test-case in PeptideIndexing_test would be nice (wait for #3458 to have it activated)

timosachsenberg:
ok tests for PeptideIndexing have been enabled.
Agree with chris that it would be good to have a test (especially for the problem of prefixes being parts of other prefixes aka: "rev_ vs reverse_")

src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
```diff
@@ -300,6 +311,11 @@ class FASTAContainer<TFI_Vector>
     return data_.size();
   }
 
+  /// required for template parameters!
+  void reset()
+  {
```
cbielow: please insert
```
  activate_count_ = 0;
  cache_count_ = 0;
```
to reset this.
Also here, please augment the tests.
i.e. after resetting, `activateCache() == true` again

Zethson: Will do.

cbielow:
1) its fine to fail right away, if `missing_decoy_action_` does not say otherwise. You should write an error message which allows the user to work around the failure, i.e. 'Automatic detection of a decoy_string failed. If you are sure there is a decoy, OR you want to index against a database without decoys, please specify a decoy_string manually." .. or something along these lines.

2) Good point! `rev` is problematic in cases of forward protein names containing rev, see https://github.com/OpenMS/OpenMS/pull/3454#discussion_r188552521
So without the underscore, this will create a false positive. So one would need `rev_`. Any idea how to do that? Extending the candidates would be an option, but that should happen on the first pass of the fasta (parsing it a second time seems potentially dangerous, especially when considering 4 GB metaproteomics FASTA files)

timosachsenberg:
maybe the regex can be extended to also greedily take an "_"  ?

Zethson:
What if we changed the regex in order to match not a cleaned string like 'rev' like we do now, but rather
[arbitrary amounts of _ ]rev[arbitrary amounts of _ ] and then save the exact match? I'm far from a regex expert, but this should be doable? I hope that this wouldn't introduce the rev_ vs reverse_ issue again...

timosachsenberg:
hmm I don't see why not. You have two RegEx, one for prefix and one for suffix.
Or do you mean a prefix that starts with "_rev" ?

timosachsenberg:
I think I actually meant the same as you:
"What if we changed the regex in order to match not a cleaned string like 'rev' like we do now, but rather
[arbitrary amounts of _ ]rev[arbitrary amounts of _ ] and then save the exact match? I hope that this wouldn't introduce the rev_ vs reverse_ issue again..."
That's basically what I meant with extending the regex to match also (zero or more) "_" in greedy mode.
AFAIK, the key is using a greedy matching which means, the regex will try to match the longest pattern. Then you won't get any issue with rev and reverse (and probably don't need to sort the regex).

timosachsenberg:
I just checked. Boost regex is greedy by default

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -680,8 +695,154 @@ namespace OpenMS
       return EXECUTION_OK;
     }
 
-protected:
-    struct PeptideProteinMatchInformation
+     const String& getDecoy_string_() const;
```
cbielow: coding convention. Should be `const String& getDecoyString() const`

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -680,8 +695,154 @@ namespace OpenMS
...
+
+ protected:
+     using DecoyStringToAffixCount = std::map<std::string, std::pair<int, int>>;
+     using CaseInsensitiveToCaseSensitiveDecoy = std::map<std::string, std::string>;
+
+     template<typename T>
+     bool findDecoyString_(FASTAContainer<T>& proteins,
+                                std::string& decoy_string,
```
cbielow: I just realized that decoy_string and is_prefix are references to internal members. If you are not pulling out the function, you might as well just work directly on the members. Makes assignments in the code a bit clearer as well.

src/tests/class_tests/openms/source/PeptideIndexing_test.cpp
```diff
@@ -219,7 +223,20 @@ START_SECTION((ExitCodes run(std::vector<FASTAFile::FASTAEntry>& proteins, std::
   r = pi.run(proteins, prot_ids, pep_ids);
   for (Size i = 0; i < pep_ids.size(); ++i) TEST_EQUAL(pep_ids[i].getHits()[0].extractProteinAccessionsSet().size(), 0); // no hits
 
- 
+  // resetting caching
```
cbielow: the reset test belongs into the class-test where reset() is implemented

src/utils/RNPxlSearch.cpp
```diff
@@ -2311,6 +2311,7 @@ class RNPxlSearch :
     // reindex ids
     PeptideIndexing indexer;
     Param param_pi = indexer.getParameters();
+    param_pi.setValue("decoy_string", "rev_");
```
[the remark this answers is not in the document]
Zethson: We don't. However we have to supply ANY decoy string or else we get the "Unable to determine decoy string" error. There are no decoys in this test.
This specific case was always using the default decoy_string "DECOY_". When I fixed this I forgot that the default was "DECOY_", but for consistencies sake I should use "DECOY_".

However this and other cases are the origin of my question 1 here: https://github.com/OpenMS/OpenMS/pull/3454#issuecomment-390661536

If it s impossible to determine a decoy string (because there are no decoys) the function just errors out. In the old implementation it would simply take "DECOY_" and never use it. It didn't error out however.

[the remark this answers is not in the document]
timosachsenberg: hmm @cbielow what do you thing. Shoud we, in the case the zero occurrences of any possible decoy in the was found just fall back to DECOY_ and just rely on the missing decoy warning that is produced later?

PR #3455, opened 2018-05-13
UTIL AssayGeneratorMetabo

oliveralka:
Basic Utility to build an assay library using metabolomics DDA data. 

Internal procedure AssayGeneratorMetabo:
  1. Input mzML and featureXML
  2. Annotate precursor mz and intensity
  3. Filter feature by convexhull size
  4. Assign precursors to specific feature
  5. Extract feature meta information (if possible)
  6. Find MS2 spectrum with highest intensity precursor for one feature
  7. Dependent on the method use the MS2 with the highest intensity precursor or a conensus spectrum for the transition calculation
  8. Calculate thresholds (maximum and minimum intensity for transition peak)
  9. Extract and write transitions

Method: 
- highest intensity precursor (use MS2 from the precursor with the highest intensity to assign the transition )
-  consensus_spectrum (calculate the cosine similariy between the MS2 of the same precursor in a feature and merge them before transition assignment)

For targeted (AccurateMassSearch) and semi-targeted (FeatureFinderMetabo/AccurateMassSeach + use_known_unkowns).

further TODOs:
[1] Test the util in combination with OpenSWATH 
[2] Add a validation for the transitions (e.g. Sirius) - if you have any ideas to go about that please let me know. 

Remark (Things to further validate):
[1] I added an adduct column will this work will OpenSWATH ignore the column or will there be problems reading the file 

Please let me know what you think and if you have further suggestions in regard to the transition caluclation and the progamming/coding style (e.g. move functions to header file / further split into functions).

Thank you!


src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+    setValidFormats_("in_id", ListUtils::create<String>("featurexml"));
+
+    registerOutputFile_("out", "<file>", "", "Assay library output file");
+    setValidFormats_("out", ListUtils::create<String>("tsv"));
+
+    registerStringOption_("method", "<choice>", "highest_intensity", "",false);
+    setValidStrings_("method", ListUtils::create<String>("highest_intensity,consensus_spectrum"));
+
+    registerDoubleOption_("precursor_mz_tolerance", "<num>", 0.005, "Tolerance window (left and right) for precursor selection", false);
+    registerDoubleOption_("precursor_rt_tolerance", "<num>", 5, "Tolerance window (left and right) for precursor selection", false);
```
hroest: it would be nice to have those in `ppm` and to in seconds (is it already or is it not?)

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    registerDoubleOption_("cosine_similarity_threshold", "<num>", 0.98, "Threshold for cosine similarity of MS2 spectras of same precursor used for consensus spectrum", false);
+
+    registerIntOption_("filter_by_convex_hulls", "<num>", 2, "Features have to have at least x MassTraces", false);
+
+    registerDoubleOption_("transition_threshold", "<num>", 10, "Further transitions need at least x% of the maximum intensity (default 10%)", false);
+
+    registerFlag_("use_known_unknowns", "Use features without identification", false);
```
hroest: do you mean spectra without identification? how else are you getting transitions?

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+        // if id information is missing set use_known_unknowns to true
+        if (feature_map.getProteinIdentifications().empty())
+        {
+          use_known_unknowns = true;
+        }
+      }
+    }
+
+    // annotate precursor mz and intensity
+    annotatePrecursorIntensity(spectra, 0.2, 0.2);
```
hroest: these values should be adjustable, for TOF and Orbitrap data you would want to have different tolerances here!

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+    MSSpectrum::ConstIterator left = spectrum1.MZBegin(test_mz - left_tolerance);
+    MSSpectrum::ConstIterator right = spectrum1.MZEnd(test_mz + right_tolerance);
+
+    // no MS1 precursor peak in +- tolerance window found
+    if (left == right || left->getMZ() > test_mz + right_tolerance)
+    {
+      return -1;
+    }
+
+    MSSpectrum::ConstIterator max_intensity_it = std::max_element(left, right, Peak1D::IntensityLess());
```
hroest: are the input data expected to be centroided? this may not be the best approach for profile data

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    if (max_intensity_it == right)
```
hroest: why only check right? and not left?

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+        LOG_WARN << "No MS1 Spectrum was found to the specific precursor" << std::endl;
+        continue;
+      }
+
+      const MSSpectrum& precursor_spectrum = *s_it2;
+      Int mono_index = getHighestIntensityPeakInMZRange(test_mz, precursor_spectrum, left_tolerance, right_tolerance);
+      if (mono_index == -1)
+      {
+        LOG_WARN << "No precusor peak in MS1 spectrum found." << std::endl;
+        continue;
```
hroest: or the tolerance is too small ...

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+                                                 });
+    feature_map.erase(map_it, feature_map.end());
+
+    KDTreeFeatureMaps fp_map_kd;
+    vector<FeatureMap> v_fp;
+    v_fp.push_back(feature_map);
+    fp_map_kd.addMaps(v_fp);
+
+    // read FeatureMap in KDTree for feature-precursor assignment
+    map<const BaseFeature*, vector<size_t> > map_feature_to_precursors = extractMetaInformation(spectra, fp_map_kd, precursor_mz_tol, precursor_rt_tol);
```
hroest: this is a bit confusing as you are actually mapping features to triggered MS2 spectra (while the feature represents the precursor). I suggest to rename this to `feature_ms2_spectra_map` or something similar

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+      // find precursor/spectrum with highest intensity precursor
+      std::vector<size_t> index = it->second;
+      for (std::vector<size_t>::iterator index_it = index.begin(); index_it != index.end(); ++index_it)
+      {
+        const MSSpectrum &spectrum = spectra[*index_it];
+        const vector<Precursor> &precursor = spectrum.getPrecursors();
+
+        // get m/z and intensity of precursor
+        double precursor_mz = precursor[0].getMZ();
```
hroest: add a comment here that only spectra with precursors are in the map, therefore no need to check for presence of precursors

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+        // calculate consensus spectrum
+        exp.sortSpectra();
+        SpectraMerger merger;
+        Param p;
+        p.setValue("precursor_method:mz_tolerance", precursor_mz_tol*2);
+        p.setValue("precursor_method:rt_tolerance", precursor_rt_tol*2);
+        merger.setParameters(p);
+
+        // all MS spectra should have the same precursor
+        merger.mergeSpectraPrecursors(exp);
```
hroest: this is a good start but we will have to look into whether this is really the optimal way of averaging / merging spectra

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+    // write output
+    TextFile tf;
+    tf.addLine("PrecursorMz\tProductMz\tLibraryIntensity\tRetentionTime\tCompoundName\tSMILES\tSumFormula\tTransitionId\tTransitionGroupId\tAdduct\tDecoy\n");
+    for (auto & entry : assaylib)
+    {
+      tf.addLine(String(entry.precursor_mz)+"\t"+String(entry.product_mz)+"\t"+String(entry.library_int)+"\t"+String(entry.normalized_rt)+
+                 "\t"+entry.compound_name+"\t"+entry.smiles+"\t"+entry.sumformula+"\t"+entry.transition_group_id+"\t"+entry.transition_id+
+                 "\t"+entry.adduct+ "\t"+entry.decoy+"\n");
+    }
+    tf.store(out);
```
hroest: this works for now, in the future it would be nice to use the TransitionTSVWriter ...

src/tests/topp/CMakeLists.txt
```diff
@@ -2104,6 +2104,29 @@ add_test("UTILS_AccurateMassSearch_1" ${TOPP_BIN_PATH}/AccurateMassSearch -test
...
+add_test("UTILS_AssayGeneratorMetabo_4_out1" ${DIFF} -in1 AssayGeneratorMetabo_ams_output_consensus.tmp -in2 ${DATA_DIR_TOPP}/AssayGeneratorMetabo_ams_output_consensus.tsv -whitelist "")
+set_tests_properties("UTILS_AssayGeneratorMetabo_4_out1" PROPERTIES DEPENDS "UTILS_AssayGeneratorMetabo_4")
+# use AccurateMassSearch data with method consensus_spectrum and use_known_unkowns
+add_test("UTILS_AssayGeneratorMetabo_5" ${TOPP_BIN_PATH}/AssayGeneratorMetabo -test -in ${DATA_DIR_TOPP}/AssayGeneratorMetabo_input.mzML -in_id ${DATA_DIR_TOPP}/AssayGeneratorMetabo_ams_input.featureXML -out AssayGeneratorMetabo_ams_uku_output_consensus.tmp -method consensus_spectrum -use_known_unknowns)
+add_test("UTILS_AssayGeneratorMetabo_5_out1" ${DIFF} -in1 AssayGeneratorMetabo_ams_uku_output_consensus.tmp -in2 ${DATA_DIR_TOPP}/AssayGeneratorMetabo_ams_uku_output_consensus.tsv -whitelist "")
+set_tests_properties("UTILS_AssayGeneratorMetabo_5_out1" PROPERTIES DEPENDS "UTILS_AssayGeneratorMetabo_5")
```
hroest: can you add a test to ensure the output can be converted to `TraML` format using the `TargetedFileConverter` and then check the identity of the TraML file?

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+        // write row for each transistion
+        // current int has to be higher than transition thresold and should not be smaller than threshold noise
+        if (current_int > threshold_transition && current_int > threshold_noise)
+        {
+          float rel_int = current_int/max_int;
+          row.precursor_mz = highest_precursor_mz;
+          row.product_mz = current_mz;
+          row.library_int = rel_int;
+          row.normalized_rt = spectrum_rt;
```
hroest: I think it may be smarter to take the *Feature* RT here since this is simply the RT of where the MS2 spectrum was triggered. Due to dynamic exclusion, this may be far from the apex of the actual RT peak.

hroest:
> [1] Test the util in combination with OpenSWATH

sounds like a good idea :-) 

> [2] Add a validation for the transitions (e.g. Sirius) - if you have any ideas to go about that please let me know.

what do you mean with validation? E.g. it may make sense to annotate the spectra with whatever Sirus thinks these spectra are based on the fragmentation trees. Do we currently have such an implementation, e.g. are data from Sirius read back into the spectra as additional data traces or how does that currently work?

> Remark (Things to further validate):
> [1] I added an adduct column will this work will OpenSWATH ignore the column or will there be problems reading the file

Will currently be ignored, but again it may make sense to go through the TransitionTSVReader class to make sure that we are in sync with what you write out and what gets read into a TraML file

> Please let me know what you think and if you have further suggestions in regard to the transition caluclation and the progamming/coding style (e.g. move functions to header file / further split into functions).

I think the transition selection right now is not very sophisticated yet and we may be able to hijack some code from https://github.com/OpenMS/OpenMS/blob/develop/src/topp/OpenSwathAssayGenerator.cpp specifically the annotation and restriction of transitions:

https://github.com/OpenMS/OpenMS/blob/7ef56453f13d18609d0ff46fb0128f4480999580/src/topp/OpenSwathAssayGenerator.cpp#L300-L308

of course our use-case is quite different and we may have to adjust `MRMAssay` to also work with metabolite transitions. But I think at least `restrictTransitions` is something we should be able to add and we could write a new transition filtering functions that is based on Sirius annotation since these are not peptides. Basically following the idea of [Scheubert et al](https://www.nature.com/articles/s41467-017-01318-5)

src/utils/AssayGeneratorMetabo.cpp
oliveralka: will add ppm and rt is in seconds

src/utils/AssayGeneratorMetabo.cpp
oliveralka: I mean that a feature was found in the FeatureFinderMetabo but no identification was performed on that feature beforehand (like AccurateMassSearch or de-novo identification). Which would mean that It is a compound (feature) but you do not know which one - known_unknown).

src/utils/AssayGeneratorMetabo.cpp
oliveralka: should have precursor_mz_tolarance as a variable (will change that)

src/utils/AssayGeneratorMetabo.cpp
oliveralka: The input data is expected to be centroided (will add an error if it is profile data.)

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    registerStringOption_("method", "<choice>", "highest_intensity", "",false);
+    setValidStrings_("method", ListUtils::create<String>("highest_intensity,consensus_spectrum"));
+
+    registerDoubleOption_("precursor_mz_tolerance", "<num>", 0.005, "Tolerance window (left and right) for precursor selection", false);
+    registerDoubleOption_("precursor_rt_tolerance", "<num>", 5, "Tolerance window (left and right) for precursor selection", false);
+
+    registerDoubleOption_("cosine_similarity_threshold", "<num>", 0.98, "Threshold for cosine similarity of MS2 spectras of same precursor used for consensus spectrum", false);
+
+    registerIntOption_("filter_by_convex_hulls", "<num>", 2, "Features have to have at least x MassTraces", false);
```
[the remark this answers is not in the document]
oliveralka: The rational behind this is to reduce the number of compounds dependent on mass traces in a feature - e.g. You may think that a compound should not only be defined by a mass trace but should have isotope traces as well - here you would filter by the size of the convex hulls. Which would then define your anticipated "real" feature. ( I was not able to filter by subordinates ).

hroest:
>> Do we currently have such an implementation, e.g. are data from Sirius read back into the spectra as additional data traces or how does that currently work?
> 
> As far as I know there is right now no read back to featureXML - I will have to think about that.
> 

Actually, we will have peak-level information from Sirius - e.g. each peak in the MS2 spectrum will be annotated. That information is best stored either in SpectraST format or as a StringDataArray in mzML. I think both approaches will work but it would give us more information for each peak and whether we should use it for a transition or not (e.g. if Sirius cannot find an annotation, maybe we should filter these peaks).

> It would be best to have the ID information in the featureXML or should that be stored in an idXML in the end - other suggestions?
> 

yes, featureXML works for the *overall* id of the whole MS2 spectrum but not well for the peak-level information.

>>   Will currently be ignored, but again it may make sense to go through the TransitionTSVReader class to make sure that we are in sync with what you write out and what gets read into a TraML file
> 
> I will try to integrate the TransitionTSVWriter and check for consistency.
> 

we can postpone this to another PR - I suggest right now to just add a test to make sure your output can be read in by the `TargetedFileConverter`

>>  I think the transition selection right now is not very sophisticated yet and we may be able to hijack some code from https://github.com/OpenMS/OpenMS/blob/develop/src/topp/OpenSwathAssayGenerator.cpp specifically the annotation and restriction of transitions:
> 
> Ok, I will think about this further and see would I can do in this regard.

we can postpone this to another PR

hroest:
> Further changes regarding the spectra merging and OpenSWATHAssayGenerator will be postponed in another PR.

sounds good, I am fine with that

PR #3461, opened 2018-05-15
[FIX] use QOpenGLFunctions instead of direct calls

timosachsenberg:


jpfeuffer:
Interesting. If this passes we could look at removing the linking to a separate OpenGL.
The added inherited class seems to be part of QtGui too. So we can leave out QtOpenGL (Which seems to be deprecated anyway)

timosachsenberg:
sounds like magic :P let's see

jpfeuffer:
haha why? Shouldn't the Qt build system just find the right OpenGL library? How can it otherwise dynamically choose an OpenGL lib on Windows?

timosachsenberg:
passed

hroest:
hmm do I see this correctly, this changes basically all calls of OpenGL functions from global calls to member functions? E.g. `glEnable` is now actually `this->glEnable` ? Indeed, seems like magic - but maybe so much magic that we should add a comment on how it works?

PR #3462, opened 2018-05-15
[FIX] target+decoy are now treated as target

timosachsenberg:
@MatthewThe is treating the target+decoy as target correct? Then this should fix this issue


jpfeuffer:
Makes much more sense imho. Let's see what mathew says.

MatthewThe:
Yes, that's correct.

PR #3464, opened 2018-05-15
Copy correct metadata

timosachsenberg:


hroest:
looks good, would be even better with a regression test...

PR #3465, opened 2018-05-16
[FIX] force repaint after update_ was called

timosachsenberg:


timosachsenberg:
@jpfeuffer can you test on Mac? Only tested on linux

timosachsenberg:
log stream failed

jpfeuffer:
works flawlessly

hroest:
:+1: great - I think I may have removed that update at some point when trying to fix a bug on Linux. I will test again later today

hroest:
:+1: can confirm on Linux, works very well

PR #3468, opened 2018-05-17
Feature/fix flq tuse adducts

FabianAicheler:
Fixed FeatureLinker base featureXml branch to not convert to consensusxml, as that lost crucial adduct information. Now, FLQT really uses adduct information for the linking.

Also for convenience added adduct information to Decharger consensus features output.


src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -337,7 +338,12 @@ namespace OpenMS
          it = elements.begin(); it != elements.end(); ++it)
     {
       feature.insert(it->first, it->second->getFeature());
+      if (it->second->getFeature().metaValueExists("dc_charge_adducts"))
```
timosachsenberg: it->second->getFeature() is accessed a few times - introduce const & to point to it?

src/topp/FeatureLinkerBase.cpp
```diff
@@ -178,7 +178,9 @@ class TOPPFeatureLinkerBase :
         }
       }
 
-      vector<ConsensusMap > maps(ins.size());
+
+      vector<FeatureMap > maps(ins.size());
+      //vector<MapType> maps(ins.size());
```
timosachsenberg: remove comment

src/topp/FeatureLinkerBase.cpp
```diff
@@ -206,12 +208,25 @@ class TOPPFeatureLinkerBase :
...
+          String adduct;
+          //exception: addduct information
+          if (it->metaValueExists("dc_charge_adducts"))
+          {
+            adduct = it->getMetaValue("dc_charge_adducts");
+          }
           it->getSubordinates().clear();
           it->getConvexHulls().clear();
           it->clearMetaInfo();
+          if (adduct != "")
```
timosachsenberg: if (!adduct.empty())

src/topp/FeatureLinkerBase.cpp
```diff
@@ -206,12 +208,25 @@ class TOPPFeatureLinkerBase :
...
+          {
+            it->setMetaValue("dc_charge_adducts", adduct);
+          }
+
         }
 
-        MapConversion::convert(i, tmp, maps[i]);
```
timosachsenberg: why was the conversion used in the first place?
remove comment

PR #3469, opened 2018-05-17
[FEATURE] added option to filter single mass traces

timosachsenberg:
simple option to filter unassembled traces before writing them out.


hroest:
do we need a 7MB test file for this?

timosachsenberg:
No not really as it is already used in another test. I will remove the extra cooy

timosachsenberg:
ok I removed it

hroest:
> ok I removed it

not much point to it unless you squash the commits :-) now its still in the history - however since its duplicated, its compressed well ...

timosachsenberg:
Git does store duplicated files as pointers to 1 file.

PR #3471, opened 2018-05-18
Refactor/mrmtgp no consensus boundaries

dmccloskey:
Provides users an option to either enforce a consistent peak boundary (i.e., best left and best right) between all transitions in a transition group or use the individual transition peak boundaries when reporting the intensity for each individual transition.


src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
```diff
@@ -201,66 +270,99 @@ END_SECTION
...
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("1").getIntensity(), 140);
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("2").getIntensity(), 58); // consensus = 117
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("3").getIntensity(), 45);
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("1").getMetaValue("peak_apex_int"), 16);
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("2").getMetaValue("peak_apex_int"), 11);
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("3").getMetaValue("peak_apex_int"), 3);
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("1").getMetaValue("peak_apex_position"), 14);
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("2").getMetaValue("peak_apex_position"), 16); // consensus = 7
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("3").getMetaValue("peak_apex_position"), 7);
   }
```
hroest: can you test the same data with and without consensus?

dmccloskey: Test added in commit dd28aa1

hroest:
looks like the test failed:

```
 -  line 351:  TEST_EQUAL(transition_group.getFeatures().size(),2): got 1, expected 2

The following tests FAILED:
	549 - MRMTransitionGroupPicker_test (Failed)
```

dmccloskey:
fixed

dmccloskey:
It seems like the build failures with Jenkins are due to documentation not being produced?

dmccloskey:
@hroest and  @timosachsenberg It seems like the build failures with Jenkins are due to documentation not being produced?  Is this correct?  If so, how do I fix it?

jpfeuffer:
Which is now ;)

hroest:
@dmccloskey can you merge `develop` into your branch and push it here? Then it should re-build and tall tests should run through

src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
```diff
@@ -358,41 +491,41 @@ START_SECTION((template <typename SpectrumT, typename TransitionT> MRMFeature cr
...
     picker_param.setValue("background_subtraction", "exact");
-    picker_param.setValue("PeakIntegrator:integration_type", "simpson");
     picker.setParameters(picker_param);
 
     MRMFeature mrmfeature = picker.createMRMFeature(transition_group, picked_chroms, smoothed_chroms, chr_idx, peak_idx);
 
-    TEST_REAL_SIMILAR(mrmfeature.getFeature("2").getIntensity(), 1.42194e+06); // slightly reduced value due to background subtraction
+    TEST_REAL_SIMILAR(mrmfeature.getFeature("2").getIntensity(), 514590); // slightly reduced value due to background subtraction
     TEST_REAL_SIMILAR((double)mrmfeature.getFeature("2").getMetaValue("peak_apex_int"), 155331.37806798); // slightly reduced value due to background subtraction
-    TEST_REAL_SIMILAR(mrmfeature.getFeature("1").getIntensity(), 168685); // slightly reduced value due to background subtraction
```
timosachsenberg: this large change is expected I guess?

src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
```diff
@@ -358,41 +491,41 @@ START_SECTION((template <typename SpectrumT, typename TransitionT> MRMFeature cr
...
     Param picker_param = picker.getDefaults();
     picker_param.setValue("PeakPickerMRM::method", "legacy"); // old parameters
     picker_param.setValue("PeakPickerMRM::peak_width", 40.0); // old parameters
     picker_param.setValue("background_subtraction", "exact");
-    picker_param.setValue("PeakIntegrator:integration_type", "simpson");
     picker.setParameters(picker_param);
 
     MRMFeature mrmfeature = picker.createMRMFeature(transition_group, picked_chroms, smoothed_chroms, chr_idx, peak_idx);
 
-    TEST_REAL_SIMILAR(mrmfeature.getFeature("2").getIntensity(), 1.42194e+06); // slightly reduced value due to background subtraction
```
timosachsenberg: also a large change hier

dmccloskey: That was changed in a previous PR for this class.  It is showing up in the diff because I had done a copy and paste operation on those lines by accident.  I did not change the code that was there previously.

src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
dmccloskey: See comment below.

src/tests/class_tests/openms/source/MRMTransitionGroupPicker_test.cpp
hroest: see line 501, I think @dmccloskey just switched the two tests around so that the simpson integration is now first.

PR #3474, opened 2018-05-19
Fix/maracluster

Leon-Bichmann:
Discussed with Matthew - edited the documentation and the cluster index that will be annotated to idXML to see which spectra are in the same cluster is now equal to the scan number in the produced clustered mzML.


Leon-Bichmann:
@MatthewThe

timosachsenberg:
looks good but could you add some comments to the variables in the actual code.

Leon-Bichmann:
or which variables do you mean?

timosachsenberg:
+1

PR #3478, opened 2018-05-22
Feature/gui

hroest:
- display Ion Mobility data in TOPPView
- display SWATH-MS / PRM / DIA data in TOPPView

this works through an addition option when right-clicking on the 1D spectrum view to either enter Ion Mobility view (if present) or SWATH-MS view (if selected when loading the file). Works well together with the cached MS1 / MS2 spectra from previous PR.

Ion mobility data can be visualized in the m/z *vs* Ion Mobility domain from a 1D spectrum:

screenshot from 2018-04-25 14-06-27

screenshot from 2018-04-25 14-11-20

this works to visualize individual scans in 2D, 3D as well as 1D (again). 

For the SWATH-MS / DIA part:

Allows loading of full SWATH-MS data files in ca 20 seconds (full file is ca 40-50 GB raw data in memory, 75k spectra) with less than 300 MB of RAM usage. Loading of individual SWATH-MS slices takes about 10 seconds during which ca 2300 spectra are loaded into memory and need to be decompressed (1.5 GB of memory). This allows users to explore individual SWATH-MS windows with the whole data on disk using IndexedMzML.

screenshot from 2018-05-21 22-58-52

screenshot from 2018-05-21 22-58-02

note how all the scans have the same precursor in the above screenshot since they are all from the same SWATH-MS window. This should also work with spectrum-level SRM data, with PRM data, DIA data etc. 

Finally, this PR modifies the loading TOPPViewOpenDialog to the following:

screenshot from 2018-05-21 23-00-56

in order to make space (low intensity cutoff does not need 2 boxes) and allow user to specify loading data in DIA / SWATH-MS mode.

### DIFF
--- a/src/openms_gui/include/OpenMS/VISUAL/APPLICATIONS/TOPPViewBase.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/APPLICATIONS/TOPPViewBase.h
@@ -322,6 +322,8 @@ public slots:
     void showCurrentPeaksAs3D();
     /// Shows the current peak data of the active layer as ion mobility
     void showCurrentPeaksAsIonMobility();
+    /// Shows the current peak data of the active layer as DIA data
+    void showCurrentPeaksAsDIA();
     /// Shows the 'About' dialog
     void showAboutDialog();
     /// Saves the whole current layer data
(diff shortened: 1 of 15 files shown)


timosachsenberg:
looks good

PR #3479, opened 2018-05-22
[Fix] Versions with strings

jpfeuffer:
Problems with "_" in Tutorials
Problems with mismatch in mac package.


jpfeuffer:
Also KNIME/Eclipse does not support strings.

PR #3480, opened 2018-05-22
Feature/rn pxl search ladder

timosachsenberg:
Refactoring of RNPxlSearch engine.
Created separate classes for
- reporting
- modification generation

and removed deprecated RNPxl tool


timosachsenberg:
The new classes and extra files are only of use to RNPxlSearch. Should I make everything private and RNPxlSearch a friend so I don't need to write extra tests / pyOpenMS bindings? They are tested via the tool test anyways. Or is it fine to keep them as is?

src/openms/include/OpenMS/ANALYSIS/RNPXL/RNPxlFragmentAnnotationHelper.h
```diff
@@ -0,0 +1,231 @@
...
+#include <vector>
+#include <algorithm>
+
+namespace OpenMS 
+{
+
+/* @brief Convenience functions to construct appealing fragment annotation strings
+ *
+ */
+class FragmentAnnotationHelper
```
hroest: @timosachsenberg can you either put this into `OpenMS::Internal` or consider exposing it through pyOpenMS ?

timosachsenberg:
I move it to internal and cleaned up the code a bit. Any additional reviewer comments?

src/openms/include/OpenMS/ANALYSIS/RNPXL/RNPxlReport.h
```diff
@@ -71,6 +71,7 @@ struct OPENMS_DLLAPI RNPxlReportRow
   double m_3H;
   double m_4H;
   String fragment_annotation_string;  
+  int rank;
```
hroest: can you sync this here? https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/pxds/RNPxlReport.pxd#L50

src/openms_gui/source/VISUAL/SpectraIdentificationViewWidget.cpp
```diff
@@ -676,28 +653,46 @@ namespace OpenMS
               }
               item->setBackgroundColor(c);
               table_widget_->setItem(table_widget_->rowCount() - 1, 4, item);
+
+              // set precursor intensity
+              item = table_widget_->itemPrototype()->clone();
+              item->setData(Qt::DisplayRole, first_precursor.getIntensity());
```
hroest: as discussed in  #3486 this may often be zero

src/utils/RNPxlSearch.cpp
```diff
@@ -2350,13 +2162,684 @@ class RNPxlSearch :
...
+{
+  const double xl_mz = (fixed_and_variable_modified_peptide_weight + precursor_rna_weight +
+                  static_cast<double>(charge) * Constants::PROTON_MASS_U)
+                 / static_cast<double>(charge);
+  partial_loss_spectrum.push_back(Peak1D(xl_mz, 1.0));
+  partial_loss_spectrum_charge.push_back(charge);
+  partial_loss_spectrum_annotation.push_back(String("[M+") + precursor_rna_adduct + "]");
+}
+
 int main(int argc, const char** argv)
```
hroest: this class / file looks like its way to large. Is there a way that some this can make its way into the library?

hroest:
@timosachsenberg can you comment on what the changes in the GUI code are doing?

timosachsenberg:
Thanks for taking another look at it.

I agree that the util gained some holiday pounds over time. I already moved most of the code into either the library for general use in the other search engines in previous PRs (or the specialized stuff to the RNPxlXXX classes) to shrink its size. Most of the remaining code so targeted to our method that I doubt that it is more generally applicable. I will continue to move more code (especially the scoring) to extra RNPxlXYZ helper classes in future PRs when I tweak the scoring if that's ok with you guys. 

Regarding the GUI I did not change anything except cleaning up the code (there is still quite some code duplication from past developments but at least its a start) and adding an additional column with the precursor intensity (based on a user request). It might be 0 in many cases but for orbitrap data it is filled with values. I figured it might be kind of useful in the future for search engine related things (e.g., reannotating precursor intensities and correlating spectra quality etc.) and its pretty hidden in the rightmost column.

hroest:
> and adding an additional column with the precursor intensity (based on a user request). It might be 0 in many cases but for orbitrap data it is filled with values.

it would be nice to either have pwiz or the vendors populate that with a meaningful value then. Or we would try and compute it ourselves on the fly, it doesnt sound super hard.

src/openms/include/OpenMS/ANALYSIS/RNPXL/RNPxlReport.h
hroest: can you also add `rank` as a member to the pxd?

timosachsenberg:
rebuild win

PR #3484, opened 2018-05-23
Refactoring XFDR

lukaszimmermann:
Fixing XFDR resulted in quite a large refactoring and some considerations:

- The code should now be more clearly structured. But there are still some things that could be improved, though. The main method is still a bit large.

- There was another bug which resulted in wrong delta scores being written to the output. This is now fixed and the test files have been regenerated. 

- As before, the delta score of the last peptide (last means lowest score) in a spectrum is the score of the `PeptideHit`. I think this is quite a bad idea, because it simply is not a delta score. I think
that the last peptide in a spectrum should not have a delta score at all.

-  I think functions such as `getRelativeError` or `getRank` should be moved to a CrossLink specific data structure or to a utility class. These functions are not specific for XFDR.

- Also IMO the logic for generating cross-link IDs as it is implemented in `getId` should clearly be moved
somewhere else. This has nothing to do with XFDR.

- There are a lot of String literals everywhere: such as `getMetaValue("OpenXQuest:id")`. Can these at least be moved as static methods to the utility class?  Currently, it is completely not obvious (at least for me) which implementors of MetainfoInterface for Cross Linking stuff can have which meta data.

- Extending the previous point, why is the meta value called `OpenXQuest:<something>`?.
There is no such thing as `OpenXQuest`.


jpfeuffer:
Is this ready? We really should have working nightlies again.

enetz:
Thanks for addressing this @lukaszimmermann .

What exacly was the bug with the delta scores? Was the absolute difference written out instead of the ratio? If it is the ratio now, then the absolute score does not make any sense for the last hit anymore.
The ratio is how xProphet does it, so I guess since XFDR is a reimplementation it is a good idea to stick to that.

You are right, that the hit score is not a delta score and the last hit should not have one, but if we apply a filter that uses the delta score to a dataset, what would happen to the hits without one?
E.g. if there is a spectrum with only one hit, but that hit has a good score. If we just ignore all "last hits" without a delta score, this kind of hit would be filtered out as well. A delta score of 1.0 would have the same problem. A delta score of 1.0 would mean the 2nd hit has the same score (bad 1st hit) or there is none at all (might be a very good hit). The best option might be to make it 0.0. This would not happen with any pair of actual scores, so we can specifically filter for those, and it would not be filtered out by using any delta score threshold.

`OpenXQuest::score` and similar meta value names are the PSI CV terms, that I asked for back when the tool was still called like that, for the MzIdentML output files. We will contact the person responsible for the PSI CV and change this when we have a fixed set of meta values. Currently our CrossLinking meta values are still changing quite often. But you are right that it is not a good idea to keep them as string literals. 
I do not have the time currently, but I will have to do some major refactoring for OpenPepXL and XFDR anyway and this stuff is on my TODO list.

For now we should focus on getting it to compile and pass the tests with the corrected delta score computation.

hroest:
> Is this ready? We really should have working nightlies again.

can we have a small PR that fixes the nightlies and then discuss larger refactoring?

lukaszimmermann:
@jpfeuffer  The tests should run successfully now, so we can merge to fix the nightlies. 

@enetz  For the delta scores: 
The delta score is implemented as the ratio. I was not aware that the delta score was at some point
implemented as the absolute difference. For the bug: the delta score for the last hit in a spectrum was missing and the n-1 hit's delta score was equal to the score. I think there was a problem with a shifted index. But perhaps I missing something, I can check again next week.

For the filtering. I think to define the delta score as 0 for the last hit is a valid solution.

@hroest I think we can use this PR to fix the nightlies and then open another one for futher discussions.

timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=271849

timosachsenberg:
@lukaszimmermann I initialized the double to but I guess we still need to resolve the delta score issue?

timosachsenberg:
I will merge to fix nightly builds

lukaszimmermann:
Perhaps we can collect here which changes to the delta score should be done and how the delta score should be defined for a last hit in a spectrum. I will open another PR where the delta score of the last hit is set to 0.0.

PR #3485, opened 2018-05-24
Feature/msstats

timosachsenberg:


timosachsenberg:
Ok I trimmed it down to 6 mb... hope thats fine

timosachsenberg:
TODO: TOPP tool test

src/utils/MSstatsConverter.cpp
```diff
@@ -0,0 +1,635 @@
...
+              String _accession,
+              String _sequence,
+              String _precursor_charge,
+              String _fragment_ion,
+              String _frag_charge,
+              String _isotope_label_type,
+              String _condition,
+              String _bioreplicate,
+              String _run,
+              String _fraction
```
hroest: can you make all these `const String&` ?

src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
```diff
@@ -34,70 +34,210 @@
...
-      std::string file; ///< file name, mandatory
-      unsigned fraction; ///< fraction 1..m, mandatory, 1 if not set
-      unsigned technical_replicate; ///< technical replicate 1..k of a fraction, 1 if not set
+
+      SampleSection() = default;
+
+      SampleSection(
+        std::vector< std::vector < String > > content,
+        std::map< unsigned, Size > sample_to_rowindex,
+        std::map< String, Size > columnname_to_columnindex
```
hroest: make all of these const-ref?

src/openms/include/OpenMS/METADATA/ExperimentalDesign.h
```diff
@@ -34,70 +34,210 @@
...
+    void setMSFileSection(const MSFileSection& msfile_section);
 
+    // Returns the Sample Section of the experimental design file
+    const ExperimentalDesign::SampleSection& getSampleSection() const;
+
+    void setSampleSection(const SampleSection& sample_section);
+
+
+    /// return fraction index to file paths (ordered by fraction_group)
+    std::map<unsigned int, std::vector<String> > getFractionToMSFilesMapping() const;
```
hroest: return const-ref?

timosachsenberg:
I'll add the pyOpenMS tests.
quick question: is it intended that def testString(): has no @report ?

timosachsenberg:
@hroest stupid question but where does the test file need to go?

hroest:
> @hroest stupid question but where does the test file need to go?

probably `src/pyOpenMS/tests/unittests` and then do this

```
dirname = os.path.dirname(os.path.abspath(__file__))
self.filename = os.path.join(dirname, "test.TraML").encode()
```

hroest:
great, looks ready to merge from my end :+1:

PR #3492, opened 2018-05-26
[FIX] Fixing debug build, accessing const map key with .at instead of []

Zethson:
Hi,

I hope that this is not solely on my end, but rather a problem for everyone ;)

The following compiler error occurs when trying to build OpenMS in debug mode: /home/lheumos/CLionProjects/OpenMS/src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h:258:96: error: no match for 'operator[]' (operand types are 'const std::map<OpenMS::String, int>' and 'const OpenMS::String')

It is not allowed to access const maps keys with parentheses [], because values not present in the map will get inserted using the [] operator. Therefore, it is recommended to use .at, which checks whether or not the key is present in the map and does not insert any new keys.


jpfeuffer:
Ah thanks. I didn't look into the Debug nightlies because they were failing because of another problem. But yes, this problem shows up, too.
http://cdash.openms.de/viewBuildError.php?buildid=271628

PR #3493, opened 2018-05-28
fix quickbuild script

timosachsenberg:


jpfeuffer:
Can you explain why?

timosachsenberg:
I think the normal cmake script does not have -DNUMBER_OF_JOBS does it?
At least it complained on my PC

jpfeuffer:
Ah yes, sure. Only the contrib. For openMS it's a make parameter.

PR #3494, opened 2018-05-28
const-fix for currentSpectrum

hroest:
changes muteable access to different function
allows fixing places in the code where we apparently need a muteable spectrum

### DIFF
--- a/src/openms_gui/source/VISUAL/LayerData.cpp
+++ b/src/openms_gui/source/VISUAL/LayerData.cpp
@@ -79,7 +79,7 @@ namespace OpenMS
     }
   }
 
-  LayerData::ExperimentType::SpectrumType & LayerData::getCurrentSpectrum()
+  LayerData::ExperimentType::SpectrumType & LayerData::getCurrentSpectrumMuteable()
   {
     return cached_spectrum_;
   }
@@ -94,7 +94,8 @@ namespace OpenMS
     // Return if no valid peak layer attached
     if (getPeakData()->size() == 0 || type != LayerData::DT_PEAK) { return; }
 
-    MSSpectrum & spectrum = getCurrentSpectrum();
+    // TODO: check if this is correct?
+    MSSpectrum & spectrum = getCurrentSpectrumMuteable();
     int ms_level = spectrum.getMSLevel();
 
     if (ms_level == 2)
@@ -219,7 +220,8 @@ namespace OpenMS
     // no ID selected
     if (peptide_id_index == -1 || peptide_hit_index == -1) { return; }
 
-    MSSpectrum & spectrum = getCurrentSpectrum();
+    // TODO: check if this is correct?
+    MSSpectrum & spectrum = getCurrentSpectrumMuteable();
     int ms_level = spectrum.getMSLevel();
 
     // wrong MS level
--- a/src/openms_gui/source/VISUAL/LayerData.cpp
+++ b/src/openms_gui/source/VISUAL/LayerData.cpp
@@ -81,11 +81,6 @@ namespace OpenMS
     }
   }
 
-  LayerData::ExperimentType::SpectrumType & LayerData::getCurrentSpectrumMuteable()
-  {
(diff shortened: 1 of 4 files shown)


src/openms_gui/source/VISUAL/LayerData.cpp
```diff
@@ -94,7 +94,8 @@ namespace OpenMS
     // Return if no valid peak layer attached
     if (getPeakData()->size() == 0 || type != LayerData::DT_PEAK) { return; }
 
-    MSSpectrum & spectrum = getCurrentSpectrum();
+    // TODO: check if this is correct?
+    MSSpectrum & spectrum = getCurrentSpectrumMuteable();
```
hroest: @timosachsenberg can you check whether this should be `MSSpectrum & spectrum = getPeakDataMuteable()->getSpectrum(current_spectrum_);`

src/openms_gui/source/VISUAL/LayerData.cpp
```diff
@@ -219,7 +220,8 @@ namespace OpenMS
     // no ID selected
     if (peptide_id_index == -1 || peptide_hit_index == -1) { return; }
 
-    MSSpectrum & spectrum = getCurrentSpectrum();
+    // TODO: check if this is correct?
+    MSSpectrum & spectrum = getCurrentSpectrumMuteable();
```
hroest: @timosachsenberg can you check whether this should be `MSSpectrum & spectrum = getPeakDataMuteable()->getSpectrum(current_spectrum_);`

src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
```diff
@@ -1130,7 +1130,8 @@ namespace OpenMS
     {
       (*getCurrentLayer_().getPeakDataMuteable())[i].sortByPosition(); // TODO 
     }
-    getCurrentLayer_().getCurrentSpectrumMuteable().sortByPosition();
+    Size current_spectrum = getCurrentLayer_().getCurrentSpectrumIndex();
+    getCurrentLayer_().getPeakDataMuteable()->getSpectrum(current_spectrum).sortByPosition();
```
hroest: this is actually now not the same thing as before, as we were sorting the cached spectrum

src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
```diff
@@ -1130,7 +1130,7 @@ namespace OpenMS
     {
       (*getCurrentLayer_().getPeakDataMuteable())[i].sortByPosition();
     }
-    getCurrentLayer_().getCurrentSpectrum().sortByPosition();
+    getCurrentLayer_().sortCurrentSpectrumByPosition();
```
hroest: not sure how necessary this is here, but this does the same thing as the code did before

timosachsenberg: If we use our default mzML reader the spectra should be sorted on load (I think I implemented this quite a while ago). Question is, if this is also the case for the OnDisc data. If so we would not need mutable access here I guess

hroest: we currently do not, we could add it here: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp#L310 and here https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp#L343

timosachsenberg: what do you think? should we do so?
Could otherwise lead to different behavior if in memory and on disc is switched

hroest: > should we do so?

yes, we should - there is no reason not to

PR #3496, opened 2018-05-28
[FEATURE] msgfplus Trypsin -> Trypsin/P; xtandem full -> semi

Zethson:
Hey,

some users who have xtandem generated data are running PeptideIndexer with the wrong enzyme specificity. Enzyme specificity 'full' should not be used when xtandem was used. This PR will automatically switch to enzyme specificity semi when full was detected for xtandem.

Furthermore, MSGFPLUS uses Trypsin/P instead of Trypsin, therefore if Trypsin is detected with MSGFPLUS data this PR automatically switches to Trypsin/P.


timosachsenberg:
@jpfeuffer @hroest do you know of any other search engine that might need special treatment here?

jpfeuffer:
See https://github.com/OpenMS/OpenMS/pull/3059
@cbielow Do you know if this is part of PepIdxr already?

jpfeuffer:
I talked to him and it could be added.
In the form: If XTandem or ConsensusID with XTandem, change the cutting rule flags to the new ones in #3059 .
Don't forget to forward the information on the search engines in Consensus ID if you want full support for automation.

PR #3498, opened 2018-05-29
[FIX] annotateSpectra() correctly annotates multiple spectra per transition

pasqoo:
The implementation should not `break;` when a transition in the target list annotates a spectrum (meaning at most one spectrum is annotated, per transition).
Instead, each transition should potentially annotate all the spectra that satisfy the conditions.

I noticed I had not implemented a check for the sizes of inputs in `selectSpectra()`. I am using this same PR to fix this. It shouldn't generate confusion :smiley:


src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -373,6 +372,10 @@ namespace OpenMS
     const bool compute_features
   ) const
   {
+    if (compute_features && scored_spectra.size() != features.size())
+    {
+      throw Exception::InvalidSize(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
```
hroest: can you use `TEST_EXCEPTION` to test this code path?

pasqoo: Thanks for merging. I'll do this in some other PR, noted :+1:

PR #3499, opened 2018-05-29
Fix/ion annotation

timosachsenberg:
- show/hide ion ladder
- simplified update of fragment annotations by user
- support for both "++" and "2+" charge suffixes in ion annotations


hroest:
havent tested it but it looks generally good.

As a side-note: it seems that we could / should move some parsing of ion annotations into the main library, e.g. `y6++`, `y12+2` etc

Issue #3495, opened 2018-05-28
ConsensusMap operator+=


timosachsenberg:
Adding two consensus maps using operator+= just appends the two lists of consensus features. In other words, if we add a cmap with 2 maps (e.g. light and heavy column) and another one with 2 maps (light and heavy column) we get a cmap with 2 maps (=2 columns).
I think this is unexpected and rarely what one wants.
Better would be to have two methods:
appendRows (to do what operator+= now does)
and
appendColumns (that adds the second map to yield a cmap with map size 4 = basically 4 columns of which either the first 2 or the last two are filled)

PR #3500, opened 2018-05-29
[FEATURE] rename operator+ to appendRows

timosachsenberg:
- renamed operator+= to appendRows()
- added appendColumns() 

The first has probably originally been used to merge e.g. consensusMaps from different fractions. 
appendColumns was not existent but is IMHO the right way to merge consensusMaps (e.g. from different fractions) as it preserves the meta information (which quant value came from what file and which channel/label).

I also changed the name FileDescriptions to ColumnHeaders as FileDescriptions was a bit misleading and might date back to the times when we only supported label-free.

I also changed tracking of the primary MS run to a single place (the interface stays the same but we don't use an extra meta value but store it directly in the ColumnHeaders) and made this consistent throughout OpenMS.

Now it should be possible to take a consensusMap and map the contained quant columns to the entries in the OpenMS experimental design.


timosachsenberg:
will add more

timosachsenberg:
adresses https://github.com/OpenMS/OpenMS/issues/3495

hroest:
please fix
```
/home/travis/build/OpenMS/OpenMS/src/openms/source/KERNEL/ConsensusMap.cpp:510:  If an else has a brace on one side, it should have it on both  [readability/braces] [5]
Done processing /home/travis/build/OpenMS/OpenMS/src/openms/source/KERNEL/ConsensusMap.cpp


/home/travis/build/OpenMS/OpenMS/src/utils/MSstatsConverter.cpp:234:  Missing space before ( in if(  [whitespace/parens] [5]
Done processing /home/travis/build/OpenMS/OpenMS/src/utils/MSstatsConverter.cpp
Total errors found: 1
```

timosachsenberg:
More comments/suggestions? Otherwise I‘ll merge

jpfeuffer:
Question on appendColumns:
How do you join the rows? Did you think about this? Like an incremental linking if consensus features actually match.

timosachsenberg:
they are just appended as new columns.
So if you have column 0 and 1
and you append another consensus map with column 0 and 1
you get column 0 to 3
If you want to join rows you need a feature linker

timosachsenberg:
Say you have 6 runs - 3 samples with 2 fractions each.
You would use feature linker to create 2 consensus maps (one for faction 1 and one for fraction 2).
Then you append columns and get on consensus map with 6 columns and the number of rows is the sum of both original maps.

timosachsenberg:
TODO: Probably add something like this to docu

<table>
  <tr>
    <th></th>
    <th>label 1 / MS file A</th>
    <th>...</th>
    <th>label 9 / MS file A</th>
    <th>...</th>
    <th>label 1 / MS file Z</th>
    <th>...</th>
    <th>label 9 / MS file Z</th>
  </tr>
  <tr>
    <td>Pep. ID-MS file A-1</td>
    <td>1234</td>
    <td>...</td>
    <td>1234</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <td>...</td>
    <td>1234</td>
    <td>...</td>
    <td>-<br></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <td>Pep. ID-MS file A-999</td>
    <td>-</td>
    <td>...</td>
    <td>1234</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <td>...</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <td>Pep. ID-MS file Z-1</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td>1234</td>
    <td>...</td>
    <td>1234</td>
  </tr>
  <tr>
    <td>...</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td>1234</td>
    <td>...</td>
    <td>-</td>
  </tr>
  <tr>
    <td>Pep. ID-MS file Z-999</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td>-</td>
    <td>...</td>
    <td>1234</td>
  </tr>
</table>

cbielow:
remark/question on the -1/numeric_limits;;max() part: is there a way to get rid of the warning?!
`size_t x = -1;` is perfectly valid C++ since the standard guarantees wrapping for unsigned types (not for signed ones, but thats another topic). I find it reads much smoother than the numeric_limits version.

timosachsenberg:
I don't know of any (except maybe disabling the warning with a pragma ;) 
Maybe putting a (size_t)-1 ?
did not test it

Issue #3501, opened 2018-05-29
XML Reader performance


hroest:
This paper has some interesting description on how to improve XML reader performance for mzML files:

Fast pure ion chromatograms extraction method for LC-MS. https://www.sciencedirect.com/science/article/pii/S0169743917303726#appsec1 -- see also their supplemental material: https://ars.els-cdn.com/content/image/1-s2.0-S0169743917303726-mmc1.doc

hroest:
use https://github.com/aklomp/base64

cbielow:
we have base64 SIMD now. Closing.

PR #3504, opened 2018-05-30
[FEATURE] quick spectra and chromatogram count

cbielow:
the core of the PR is b3478c0:
MzMLFileHandler can now quickly determine the number of spectra with respect to options, e.g. count the number of MS2 spectra, without loading the actual content. As before the total number of spectra and chromatograms can be extracted as well, but even more efficient (jumping over the entire spectrumList, instead of over spectrum only).

The parsing itself should also be faster, even though this is just a byproduct.

The other commits basically do cleanup only:
 - move implementations from .h to .cpp
 - adhere to coding conventions (naming of functions and members mainly)
 - use includes only when needed (e.g. changing MzMLHandler.h led to recompilation of PeakPickerCWT -- go figure .. the main reason was wasteful header-usage and implementations in headers, which required to specify more dependencies than needed).


cbielow:
fun fact: cpplint ignores implementation errors in header files and only complains when they are moved to the cpp (I hope that's only true for non-templates). Yet another reason not to do it.

cbielow:
btw: I played with a constexpr hash function which enables `switch` on strings along the lines of
```
    constexpr uint32_t hash(const char *s) {                        
      return !(*s) ? 5381 : (hash((char*)s+1)*33) ^ (*s);                           
    } 

    constexpr uint32_t hash(const XMLCh *s) {                        
      return !(char)(*s) ? 5381 : (hash((XMLCh *)s+1)*33) ^ (char)(*s);                           
    }
```
, to get rid of the long chains of `if (tag == "spectrum") ... else if tag == "chromatogram") ... else if tag = ...`.
Performance-wise this gave a measurable speedup of 1 sec per 50k spectra, so not overwhelming.
Advantages of hash-functions would be: 
 - compiler error for duplicated blocks (since switch checks for it)
 - small speed boost
Disadvantages:
 - if the hash of an unlisted tag has the same hash as a tag which we handle, there will be nasty errors. One could somehow deal with this by checking for collisions in debug mode, but that wont protect from unknown mzMLs during release-mode.

src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
```diff
@@ -169,6 +169,14 @@ namespace OpenMS
         STORE               ///< Storing a file
       };
 
+      enum LOADDETAIL 
+      {  
+        LD_ALLDATA,       // default; load all data
+        LD_RAWCOUNTS,     // only count the total number of spectra and chromatograms (usually very fast)
+        LD_COUNTS_WITHOPTIONS // count the number of spectra, while respecting PeakFileOptions (msLevel and RTRange) and chromatograms (fast)
+      };
```
hroest: wrap in pyOpenMS ?

hroest:
something seems wrong here, can you try and merge #3506 and see whether that passes?

hroest:
> Performance-wise this gave a measurable speedup of 1 sec per 50k spectra, so not overwhelming.

I dont think that is really worth it for something that may break our parsing - on the other hand, we know which tags are currently allowed in mzML so we could check this for collisions, right? There wont be any new tags for the same standard - new stuff is added through CV terms, right?

cbielow:
@hroest right. This is a good way forward, but I agree its probably not worth the effort atm. Base64 decoding using ss2/avx as you mentioned already is probably much more beneficial.


All tests passed (1 spurious LogStream test fail).
Ready for merge.

hroest:
yes, I think following up on #3501 would be the most promising approach.

PR #3507, opened 2018-05-31
[FEATURE] More unified and improved error handling in Adapters.

jpfeuffer:
- Use of QTempDirectory with auto remove functionality if debug level <2
- Always parse std err and std out if exit code != 0
- More unified logging

One could even go further and abstract external process creation/logging as well as improve the temp file handling in TOPPBase (e.g. to cleanup in the destructor).

### DIFF
--- a/src/openms/source/APPLICATIONS/TOPPBase.cpp
+++ b/src/openms/source/APPLICATIONS/TOPPBase.cpp
@@ -105,10 +105,10 @@ namespace OpenMS
     tool_name_(tool_name),
     tool_description_(tool_description),
     instance_number_(-1),
-    version_(""),
-    verboseVersion_(""),
     working_dir_(""),
     working_dir_keep_debug_lvl_(-1),
+    version_(""),
+    verboseVersion_(""),
     official_(official),
     citations_(citations),
     log_type_(ProgressLogger::NONE),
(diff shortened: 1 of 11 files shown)


hroest:
looks good but tons of duplicated code. Maybe we can put some of that into `OpenMS/APPLICATIONS/TOPPBase.h` or even a new class `OpenMS/APPLICATIONS/AdapterBase.h` ?

jpfeuffer:
Ok I did it. Although I also liked the AdapterBase idea more, I think it is not worth it yet for the one function that I had in mind.

timosachsenberg:
+1

src/openms/source/APPLICATIONS/TOPPBase.cpp
```diff
@@ -1609,6 +1627,36 @@ namespace OpenMS
...
+    if (success == false || qp.exitStatus() != 0 || qp.exitCode() != 0)
+    {
+      writeLog_("FATAL: External invocation of " + executable.toStdString() + " failed. Standard output and error were:");
+      const QString stdout(qp.readAllStandardOutput());
+      const QString stderr(qp.readAllStandardError());
+      writeLog_(stdout);
+      writeLog_(stderr);
+      writeLog_(String(qp.exitCode()));
+
+      return EXTERNAL_PROGRAM_ERROR;
```
hroest: do you catch this error?

jpfeuffer: puh.. no.. I just want the program to exit. I see that it does not work anymore.

hroest: yes, but now this returns from the function, not from `main` any more

PR #3510, opened 2018-06-02
void MzIdentMLFile::load(const String& filename, Identification& id) is broken

hroest:
There does not seem to be a test for `void MzIdentMLFile::load(const String& filename, Identification& id)` and I have tried to add one in 0fc26e1 but it seems that the result is empty.  So does the `void MzIdentMLFile::load(const String& filename, Identification& id)` not work? It does not seem tested. If it is broken, can we remove it?


timosachsenberg:
I think it should be removed (including everything related to AnalysisXML). As I understood it it is outdated and has never been really used. maybe @mwalzer knows more about it

hroest:
PS: I assume the `store` function is equally broken?

PR #3511, opened 2018-06-02
Feature/ff

hroest:
write indexedMzML also by default in the `FileFilter` tool


hroest:
nice sideeffect: fixes a bug in the indexedMzML code!

hroest:
only test failure is     projectroot.src.tests.class_tests.openms.LogStream_test

PR #3515, opened 2018-06-02
Feature/binned offset

timosachsenberg:
Added support for peak offset in binned spectrum implementation.
@pasqoo @dmccloskey could come in handy for low-res stuff like GC/MS.


timosachsenberg:
Not yet ready for review

pasqoo:
@timosachsenberg Thanks a lot for implementing this, I'll test it asap :+1:

dmccloskey:
@timosachsenberg I agree it should come in handy.  We can test it out on our GC-MS data and let you know how it works out.

PR #3516, opened 2018-06-03
Fix/false discovery rate

Leon-Bichmann:
keep unreferenced peptide hits when add_decoy_peptides is true

### DIFF
--- a/src/topp/FalseDiscoveryRate.cpp
+++ b/src/topp/FalseDiscoveryRate.cpp
@@ -195,7 +195,15 @@ protected:
     if (filter_applied)
     {
       IDFilter::removeUnreferencedProteins(prot_ids, pep_ids);
-      IDFilter::updateProteinReferences(pep_ids, prot_ids, true);
+
+      if (alg_param.getValue("add_decoy_peptides").toBool() == 1)
+      {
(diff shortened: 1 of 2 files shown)


src/topp/FalseDiscoveryRate.cpp
```diff
@@ -198,7 +195,15 @@ class TOPPFalseDiscoveryRate :
     if (filter_applied)
     {
       IDFilter::removeUnreferencedProteins(prot_ids, pep_ids);
-      IDFilter::updateProteinReferences(pep_ids, prot_ids, true);
+
+      if (alg_param.getValue("add_decoy_peptides").toBool() == 1)
```
timosachsenberg: == true

timosachsenberg: Please add a comment why protein references should not be updated if decoy peptides are added

Leon-Bichmann: /Users/bichmann/Programs/OpenMS/OpenMS/src/topp/FalseDiscoveryRate.cpp:199:52: error: use of overloaded operator '==' is ambiguous
      (with operand types 'const OpenMS::DataValue' and 'bool')
      if (alg_param.getValue("add_decoy_peptides") == true)

PR #3518, opened 2018-06-03
readme: add badge link

hroest:
side question: is there such a thing as too many badges?


hroest:
also: we could look at using https://codedocs.xyz/ for our nightly documentation build

hroest:
> can this be combined with our build infrastructure / travis? We generate quite some extra doxygen code from the command line description during the build process.

not sure, I am not even sure whether it work in the first place due to our build times...

PR #3519, opened 2018-06-04
[FIX] correctly pass exit code from external process call

jpfeuffer:


timosachsenberg:
+1

hroest:
seems like there is an issue:


 
--
7>C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\openms\source\APPLICATIONS\TOPPBase.cpp(1631): error C2556: 'OpenMS::TOPPBase::ExitCodes OpenMS::TOPPBase::runExternalProcess_(const QString &,const QStringList &) const': overloaded function differs only by return type from 'int OpenMS::TOPPBase::runExternalProcess_(const QString &,const QStringList &) const' [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\openms\OpenMS.vcxproj]                               C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\openms\include\OpenMS/APPLICATIONS/TOPPBase.h(819): note: see declaration of 'OpenMS::TOPPBase::runExternalProcess_'

PR #3522, opened 2018-06-06
Hyperlink DOIs against preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). Yes, a bit ironic, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org) and the old links with the `dx` subdomain continue to work.

So, there is no urgent need to do anything, but nonetheless, I'd like to suggest to hereby update all static DOI links.

Cheers!


katrinleinweber:
Since other PRs fail those builds as well, I presume they aren't due to my changes.

jpfeuffer:
Indeed the build had problems uploading the results because one of our servers was down. I'll try to re-build.

jpfeuffer:
rebuild jenkins

PR #3523, opened 2018-06-06
fix bug in unspecific digestion of unmodified strings

timosachsenberg:
added regression test


hroest:
any reason why the tests fail?

timosachsenberg:
I think our build server was down. According to Jenkins these did not fail

timosachsenberg:
rebuild jenkins

PR #3524, opened 2018-06-06
Feature/sse peptide motif

timosachsenberg:
Only consider peptides that match a regular expression:
e.g., -peptide:motif "FV.*" will match all peptides starting with FV


timosachsenberg:
especially useful in the context of unspecific cleavage

timosachsenberg:
to be merged after https://github.com/OpenMS/OpenMS/pull/3523

timosachsenberg:
rebuild jenkins

Issue #3528, opened 2018-06-08
Change all mzTab type output file parameters to mzTab


jpfeuffer:
Since there was a change to make mzTab the default extension,
we should either change all the tools to write mzTab or make the mzTabWriter backwards
compatible so it can still write the old tsv extension.
Affected is e.g. AccurateMassSearch but potentially more.

timosachsenberg:
k will fix

PR #3529, opened 2018-06-08
[FIX] fix mzTab extension

timosachsenberg:


timosachsenberg:
adresses https://github.com/OpenMS/OpenMS/issues/3528

jpfeuffer:
Looks good! I *think* this should solve it. If not it is a step in the right direction.

PR #3532, opened 2018-06-10
 Fix/fuzzy

hroest:
fixes a pretty serious bug in FuzzyDiff:

- Fixes the following bug: when a large ratio was found but does not
  lead to failure (due to small absolute difference) then smaller ratios
  than the original one that *should* lead to failure are ignored.
- adds tests for FuzzyDiff (really, we did not have tests so far?)

### DIFF
--- a/src/openms/source/CONCEPT/FuzzyStringComparator.cpp
+++ b/src/openms/source/CONCEPT/FuzzyStringComparator.cpp
@@ -398,18 +398,19 @@ namespace OpenMS
                   {
                     ratio = 1.0 / ratio;
                   }
+
                   // by now, we are sure that ratio >= 1
                   if (ratio > ratio_max_) // update running max
                   {
-                    ratio_max_ = ratio;
                     line_num_1_max_ = line_num_1_;
                     line_num_2_max_ = line_num_2_;
                     line_str_1_max_ = line_str_1;
                     line_str_2_max_ = line_str_2;
-                    if (ratio_max_ > ratio_max_allowed_)
+                    if (ratio > ratio_max_allowed_)
                     {
                       if (!is_absdiff_small_)
                       {
+                        ratio_max_ = ratio;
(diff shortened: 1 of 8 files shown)


src/openms/source/CONCEPT/FuzzyStringComparator.cpp
```diff
@@ -398,18 +403,25 @@ namespace OpenMS
...
                     ratio = 1.0 / ratio;
                   }
+#ifdef DEBUG_FUZZY
+                  std::cout << " check ratio:  " << ratio << " vs " << ratio_max_ << std::endl;
+#endif
+
                   // by now, we are sure that ratio >= 1
                   if (ratio > ratio_max_) // update running max
                   {
-                    ratio_max_ = ratio;
```
hroest: this did not work and led to the wrong behavior...

src/tests/topp/FuzzyDiff_1_in2.featureXML
```diff
@@ -0,0 +1,34 @@
...
+			<position dim="0">100.999999701977</position>
+			<position dim="1">412.5</position>
+			<intensity>25944.8</intensity>
+			<quality dim="0">0</quality>
+			<quality dim="1">0</quality>
+			<overallquality>2.27157</overallquality>
+			<UserParam type="float" name="var_isotope_correlation_score" value="0.865706760687806"/>
+			<UserParam type="float" name="var_isotope_overlap_score" value="0.181819841265678"/>
+			<UserParam type="float" name="var_massdev_score" value="3.78948546767785e-11"/>
+			<UserParam type="float" name="var_massdev_score_weighted" value="6.31580911279641e-11"/>
```
hroest: large relative difference, small absolute difference -> it gets ignored. But then failure is not reported upon `var_dotprod_score` on line 22

timosachsenberg:
nice catch!

PR #3539, opened 2018-06-12
fix: removed default conversion from minutes to seconds

dmccloskey:
fix: removed default conversion from minutes to seconds in MzMLHandle

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandlerHelper.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandlerHelper.cpp
@@ -375,11 +375,11 @@ namespace OpenMS
     {
       data.back().meta.setName(name);
 
-      // time array is given in minutes instead of seconds, we need to convert
-      if (accession == "MS:1000595" && unit_accession == "UO:0000031")
-      {
-        data.back().unit_multiplier = 60.0;
-      }
+      // // time array is given in minutes instead of seconds, we need to convert
+      // if (accession == "MS:1000595" && unit_accession == "UO:0000031")
+      // {
+      //   data.back().unit_multiplier = 60.0;
+      // }
     }
     else
     {


timosachsenberg:
hmm I am not exactly sure but recall that some other third-party tools had problems when minutes were annotated.
OpenMS internal data structures assume seconds so changing this in the mzML handler is probably not correct.

dmccloskey:
The problem is that none of the other readers (e.g., TraML, MRMFeatureQC, etc.,) make a similar implicit conversion.  Therefore it is up to the user to know themselves that a conversion is taking place.  I think this is quite problematic.

dmccloskey:
Also, if the user is expecting the output to be in matching units to the input mzML file, this can be quite frustrating (as is the case for our analytics unit).

hroest:
see the conversation here: #3320 - yes, OpenMS internally assumes that all the data is stored in seconds but until recently it did not explicitly check that. Maybe we can make it an option to revert to "legacy" behavior if that is required? But you are right, so far code that wrote minutes into the TraML file and then extracted mzML without the conversion was working and is now not working any more. So maybe we can load the mzML with an option to "ignore_units" or something?

hroest:
> Also, if the user is expecting the output to be in matching units to the input mzML file, this can be quite frustrating (as is the case for our analytics unit).

the problem is that some vendors would store time in minutes and others in seconds, so your solution is not cross-vendor compatible (each time you would have to guess which unit you are dealing with). Probably the right approach would be to also have TraML read in the unit correctly?

hroest:
we also do the conversion here: 

https://github.com/OpenMS/OpenMS/blob/6c74b40bb8950b1cbf89c8e426dce45483569de2/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp#L2003-L2012

so removing it from one place but not the other again would lead to inconsistencies

dmccloskey:
Should the conversion in the TraML file be done automatically then?

hroest:
> Should the conversion in the TraML file be done automatically then?

well currently the data is annotated in the TraML - so your code can basically read out whether your TraML data is in minutes or seconds. In principle we could even have an automated conversion there to seconds, if you prefer. What do you think?

dmccloskey:
One of our classes that we implemented that is reading in a TraML file is currently doing the conversion in the class.  We could modify the other classes to do the same.  However, if the conversion was done automatically to match the mzML conversion, then we would not need to worry about implementing the conversion on a case by case basis.  

Do the OpenSwath pipelines currently convert from minutes to seconds using the TraML annotations?

hroest:
> Do the OpenSwath pipelines currently convert from minutes to seconds using the TraML annotations?

no they dont. In OpenSWATH we almost never work with "raw" RT but rather use normalized RT. What this means is that our library / reference standards have RT values from 0 to 100 and we then use a few normalization peaks in the data to recalibrate all experimental RT points (be it minutes or seconds) to the normalized RT space between 0 and 100. This means that we dont care what units we input, we just need a transformation function from the library RT to the experimental RT.

PR #3540, opened 2018-06-12
Feature/decoy switch kr

hroest:
adds two new features to the OpenSwathDecoyGenerator:

- non-equal number of decyos and targets
- a feature to switch terminal K and R residues for decoy peptides
- recomputes precursor m/z based on new decoy peptide sequence


src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -324,6 +322,40 @@ namespace OpenMS
...
+  void switchKR(OpenMS::TargetedExperiment::Peptide& peptide)
+  {
+    static std::string aa[] =
+    {
+      "A", "N", "D", "C", "E", "Q", "G", "H", "I", "L", "M", "F", "S", "T", "W",
+      "Y", "V"
+    };
+    int aa_size = 17;
+
+    static boost::mt19937 generator(42);
```
grosenberger: Should a seed similar as in shufflePeptide() be used?

hroest: only if we want to actually test this part, right now all peptides in the test have K/R at the end

PR #3542, opened 2018-06-13
Update MSQuantifications.pxd

hroest:


timosachsenberg:
Thanks for the fix. MSQuantification is not really used right now (mzQuantML beeing not really adopted).
Any plans on your side to use it?

hroest:
> MSQuantification is not really used right now (mzQuantML beeing not really adopted).
> Any plans on your side to use it?

not really but I noticed that we did not have great Python support.

PR #3543, opened 2018-06-13
[FEATURE] MSSpectrum methods to return reference to a named data array

pasqoo:
Implements:
```
getStringDataArrayByName()
getIntegerDataArrayByName()
getFloatDataArrayByName()
```
Both `const` and mutable versions.
Tests are provided.

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/MSSpectrum.h
+++ b/src/openms/include/OpenMS/KERNEL/MSSpectrum.h
@@ -249,6 +249,73 @@ public:
 
     /// Sets the integer meta data arrays
     void setIntegerDataArrays(const IntegerDataArrays& ida);
+
+    /**
+      Return the `StringDataArray` with the given name
+
+      @throw Exception::ElementNotFound if no matching `StringDataArray` is found
+
+      @param[in] name The desired data array's name
+
+      @return The const reference to the found data array
+    */
+    const MSSpectrum::StringDataArray& getStringDataArrayByName(const String& name) const;
+
+    /**
+      Return the `StringDataArray` with the given name
+
+      @throw Exception::ElementNotFound if no matching `StringDataArray` is found
+
+      @param[in] name The desired data array's name
+
+      @return The mutable reference to the found data array
+    */
+    MSSpectrum::StringDataArray& getStringDataArrayByName(const String& name);
+
+    /**
+      Return the `FloatDataArray` with the given name
+
+      @throw Exception::ElementNotFound if no matching `FloatDataArray` is found
+
+      @param[in] name The desired data array's name
+
+      @return The const reference to the found data array
+    */
+    const MSSpectrum::FloatDataArray& getFloatDataArrayByName(const String& name) const;
+
+    /**
+      Return the `FloatDataArray` with the given name
+
+      @throw Exception::ElementNotFound if no matching `FloatDataArray` is found
+
+      @param[in] name The desired data array's name
+
+      @return The mutable reference to the found data array
+    */
+    MSSpectrum::FloatDataArray& getFloatDataArrayByName(const String& name);
+
+    /**
+      Return the `IntegerDataArray` with the given name
+
+      @throw Exception::ElementNotFound if no matching `IntegerDataArray` is found
+
+      @param[in] name The desired data array's name
+
+      @return The const reference to the found data array
+    */
+    const MSSpectrum::IntegerDataArray& getIntegerDataArrayByName(const String& name) const;
+
+    /**
+      Return the `IntegerDataArray` with the given name
+
+      @throw Exception::ElementNotFound if no matching `IntegerDataArray` is found
+
+      @param[in] name The desired data array's name
+
+      @return The mutable reference to the found data array
+    */
+    MSSpectrum::IntegerDataArray& getIntegerDataArrayByName(const String& name);
+
     //@}
 
     ///@name Sorting peaks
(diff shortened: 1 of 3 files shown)


hroest:
we just had a discussion about these and the problems they cause: https://github.com/OpenMS/OpenMS/pull/3438 - and in general I think we should be using the https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/KERNEL/SpectrumHelper.h for access

also we dont have a good way to check whether the array was found with your approach, checking for an exception is not a good approach. The main upside I see is that it is wrappable in Python

pasqoo:
I remember the removed methods because I was using them :smile: 
I will try to provide a better reasoning for this PR, because I understand it's important to not add back something that has been previously removed.

Those methods were removed because dangerous: they would return a reference to a `.end()` position if the element was not found. I think we briefly discussed this also on Gitter.

I kept the issue in mind while implementing the methods proposed in this PR.
If the expected array is not found, an exception is thrown.

What are the reasons against a `try/catch` block?

Imagine this used without such block. While developing a new feature in _TargetedSpectraExtractor_, this kind of methods came very handy.
In that case, I know that the element would be there, therefore I'd not use the `try/catch` block.
And, if any future developer erroneously assumes that an element is present, then an exception will greet him/her :smile:

The PR adds a feature, but does not remove what OpenMS already offers, meaning those two methods in _SpectrumHelper_ (that are actually used in this PR's `getTypeDataArrayByName()`).

I'll show an example of where this kind of helper method is useful:
```cpp
  TEST_EQUAL(s1.getStringDataArrayByName("Synon")[0], "name2 of 1st")
  TEST_EQUAL(s1.getStringDataArrayByName("Synon")[1], "name3 of firsttt")
  TEST_EQUAL(s1.getStringDataArrayByName("Formula")[0], "A11B22C333")
  TEST_EQUAL(s1.getStringDataArrayByName("MW")[0], "156")
  TEST_EQUAL(s1.getStringDataArrayByName("CAS#")[0], "0123-45-6")
  TEST_EQUAL(s1.getStringDataArrayByName("NIST#")[0], "654321")
  TEST_EQUAL(s1.getStringDataArrayByName("DB#")[0], "1")
  TEST_EQUAL(s1.getStringDataArrayByName("Comments")[0], "Some comment")
  TEST_EQUAL(s1.getStringDataArrayByName("Num Peaks")[0], "14")
```

Doing this without these methods would be many more lines. So we'd be choosing between a `try/catch` block (optional) and a `if (not end iterator)` block (necessary).
Also, from some answers on stackoverflow, it doesn't look like one block is faster/slower than the other.

Other snippets where these methods are used in my code, to provide more context (both are part of a parser for metabolites msp files, which I'll soon propose):
```cpp
// Ensure Comments metadatum is present and, if not, set it to an empty string
try
{
  spectrum.getStringDataArrayByName("Comments");
}
catch (const Exception::ElementNotFound& e)
{
  pushParsedInfoToNamedDataArray(spectrum, "Comments", "");
}
```
```cpp
// Check that all expected points are parsed
const String& num_peaks { spectrum.getStringDataArrayByName("Num Peaks").front() };
if (spectrum.size() != std::stoul(num_peaks) )
{
  throw Exception::ParseError(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
    num_peaks,
    "The number of points parsed does not coincide with `Num Peaks`."
  );
}
```
I believe this PR would only add  value to the library, without undermining its quality and the work of current/future developers. I hope it'll be given another round of thoughts. :crossed_fingers:

hroest:
> EDIT: I had the chance to test this, and I noticed a 20% slowdown when using the try/catch block. Noticeable, if used in intensive code (in my case, a parser and a file with hundreds of thousands spectra)

speed is one reason (catching and throwing exception is slow in c++ - however after reading online I am not so sure any more that is always tree). However the main reason is that from a software engineering point of view it is an anti-pattern: https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

>   TEST_EQUAL(s1.getStringDataArrayByName("Synon")[0], "name2 of 1st")

this code is nice and short but *dangerous* for various reasons, first it does not catch the exception and second you do not check whether `array.empty()` which could lead to a segfault. I think the correct could would be

```
auto it = getDataArrayByName(s1.getStringDataArrays(), "Synon")
if (it != s1.getStringDataArrays().end())
{
  if (! it->empty())
  {
    TEST_EQUAL( (*it)[0], "name2 of 1st")
  }
}
```

Finally, if you strongly feel like we should add those, we may think about adding them to SpectrumHelper.h as well to keep the `Spectrum` interface cleaner. Maybe @jpfeuffer @timosachsenberg have some insight here?

timosachsenberg:
Same opinion as Hannes. Better than the .end() solution but still an anti-pattern that usually _should_ be avoided. Others might be tempted to use try / catch with that function while in reality, the check should be more like a precondition (if the caller - who should know better - requests it but it doesn't exist, then it should IMHO be a fatal error). Given that it is nicer to wrap with pyOpenMS why not have the methods but instead of throwing an exception we could fail hard with an assertion? That way it is as fatal as an out of bounds access in a vector.

pasqoo:
I was not aware of the anti-pattern fact. Thanks for the link and the examples.

Would the technique proposed by Timo help with the issue? Meaning the precondition/assertion, and removing the exception. If so, how would you code it? The `OPENMS_PRECONDITION()` macro seems to be relevant only in `Debug` mode. Is it correct?

In general, I don't feel like it's very important to merge this PR, because it basically offers syntactic sugar to what's already present (in a safer version) in _SpectrumHelper_. So it's nice to have, but not necessary.
If it comes at the expense of code security or anything else important (ie. introducing an anti-pattern), then I feel I should also vote against the merge.

PR #3546, opened 2018-06-14
add peptide motif regex feature

Leon-Bichmann:
Added a flag to PSMFeatureExtractor to include Peptide Motifs as feature for Percolator.

This is similar to using the enzyme specificity feature for example from a tryptic digest in case you have special peptide motifs like in MHC binding peptides.


timosachsenberg:
- please remove swap files
- can there be some bias introduced if the regex only matches targets (vs. decoys). If so this needs to be documented.
- please add at least a class test

timosachsenberg:
Sounds like a useful addition. Thanks!

hroest:
> not sure how the pyopenms bindings work.. ill have a look

see https://pyopenms.readthedocs.io/en/latest/wrap_classes.html# - basically you have to add the new function to the corresponding `pxd` file

src/tests/class_tests/openms/source/PercolatorFeatureSetHelper_test.cpp
```diff
@@ -44,6 +45,20 @@
...
 
+bool check_regex(const OpenMS::String& peptide_motif)
+{
+    try
+    {
+      const OpenMS::String sequence = "TESTPEPTIDE";
+      boost::regex peptide_motif_regex(peptide_motif);
+      boost::regex_match(sequence, peptide_motif_regex);
+      return true;
+    } catch (boost::regex_error& e) {
```
hroest:
```suggestion
    } catch (boost::regex_error& e)
    {
```

src/openms/include/OpenMS/ANALYSIS/ID/PercolatorFeatureSetHelper.h
```diff
@@ -96,7 +96,15 @@ namespace OpenMS
          * Concatenates SearchParameter of multiple search engine runs and merges PeptideEvidences, collects used search engines in MetaInfo for collection in addMULTISEFeatures for feature registration.
          */
         static void mergeMULTISEProteinIds(std::vector<ProteinIdentification>& all_protein_ids, std::vector<ProteinIdentification>& new_protein_ids);
-        
+
+        /**
+         * @brief addMotifRegExFeatures
```
hroest: can you add a brief description hwere of what this does?

src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
```diff
@@ -34,11 +34,43 @@
...
+        {
+          String sequence = hit->getSequence().toUnmodifiedString();
+
+          // set motif regex feature to 1 if hit sequence matches provided regex
+          if (boost::regex_match(sequence, peptide_motif_regex))
+          {
+
+            hit->setMetaValue("MotifRegEx:match", 1);
+
+          } else {
```
hroest:
```suggestion
          } else
          {
```

PR #3547, opened 2018-06-15
[FEATURE] Implement MSPMetaboFile class

pasqoo:
The class parses a .msp file containing a metabolite library and saves it into an `MSExperiment`.

The usual `load()` method is provided, and a handy constructor with the same parameters.

Documentation and tests are provided as well.


hroest:
seems like something throws an exception during the test:
```
checking MSPMetaboFile() ... 
 +  line 58:  TEST_NOT_EQUAL(ptr,null_ptr): got 0x1b65d40, forbidden is 0
: passed

checking ~MSPMetaboFile() ... : passed

checking void load(const String& filename, MSExperiment& experiment) const ... 
Loading spectra from .msp file. Please wait.

Error: Caught std::exception
 - Message: regex_error
: failed
Warning: no subtests performed in 'void load(const String& filename, MSExperiment& experiment) const' (line 245)!

checking void pushParsedInfoToNamedDataArray( MSSpectrum& spectrum, const String& name, const String& info ) const ... 
 +  line 265:  TEST_EQUAL(SDAs.size(),1): got 1, expected 1
```

hroest:
maybe use boost regex? we had problems before with std::regex if I remember correctly.

hroest:
we had this issue before with std::regex and boost::regex, please see https://github.com/OpenMS/OpenMS/pull/3149#issuecomment-356353855

pasqoo:
rebuild win

pasqoo:
@hroest @timosachsenberg https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-15,os_label=win10_64/2022/console
not sure if I am using the correct command to trigger the CI

timosachsenberg:
rebuild jenkins

pasqoo:
Seems like three unrelated tests failed:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=274971

hroest:
@pasqoo I am not sure that you are using String Data Arrays as they are intended to be used. Basically what this will do is generate a lot of StringArrays of size 1, right? It seems that you are abusing the StringData array to store meta data about spectra - while we already have a better way to store Spectrum Meta data through the `MetaInfoInterface`. The StringDataArrays are intended for cases where you need to store information about each *peak* but not meta data about the spectrum. I think for that it is much more memory efficient and easier to use `spectrum.setMetaValue("Formula", "COOH")` and that will also be stored in the mzML file.

hroest:
we also have already a MSPFile implementation in OpenMS, but I agree that one is very much geared towards peptides. Still, it may be a bit confusing to have two implementations

pasqoo:
@hroest Oh yeah, that would also make the code easier. How would I deal with the `Synon` metadata? A spectrum can have multiple synonyms, but the metavalue key must be unique. Concatenating the synonyms and separating them with some character like `;` could work but I'd like a different approach.

Regarding the confusion between the two _MSP_ classes, what do you recommend?

pasqoo:
I updated the code with the proposed changes. The parsing also got faster (24 seconds for 220k spectra, from around 35 seconds), but it's only saving the last synonym for each spectrum. Timings might change once we decide on how to deal with synonyms, as mentioned in my last comment.

Currently, `load()` skips spectra with the same value for `Name`. It saves only the first one. This is going to change in the near future, because often libraries propose different (raw and metadata) representations for spectra of the same `Name`.

pasqoo:
failed test: `The job exceeded the maximum time limit for jobs, and has been terminated.`

pasqoo:
rebuild jenkins

hroest:
> Concatenating the synonyms and separating them with some character like ; could work but I'd like a different approach.

yes, I think that is currently the best approach and it depends how often this happens. Also you need to make sure there is no `;` character in the synonyms. Finally, the question is how important the synonyms really are - I assume they are not really for computing but for human inspection? With the StringDataArrays, there is an implicit assumption that these arrays relate to peaks and are also *of the same length* as the peak data - even though this is not part of the standard and it is perfectly valid to do what you do. Also, as you discovered, it is slow :-) 
I dont think there is a perfect solution, maybe the StringDataArray is the correct approach for the synonyms? 

> Regarding the confusion between the two MSP classes, what do you recommend?

hmm not sure yet, but I dont have strong feelings. However, I think that your class is not metabolite-specific so it should not be called metaboliteMSP, it is actually a more general version of an MSP parser.

pasqoo:
@hroest PR updated.

The name is now _MSPGenericFile_. If better ideas come up, I'll update accordingly.

Should _MSPFile_ name also be updated? If so, to what?

pasqoo:

PR #3548, opened 2018-06-15
[FIX] OpenSWATH: OSW transition-level reports without MI scores

grosenberger:
This PR fixes a bug in the OSW export functionality if transition-level scoring is enabled but MI scores are not used.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
@@ -392,7 +392,6 @@ namespace OpenMS
           std::vector<double> id_target_area_intensity = ListUtils::create<double>((String)feature_it->getMetaValue("id_target_area_intensity"),';');
           std::vector<double> id_target_total_area_intensity = ListUtils::create<double>((String)feature_it->getMetaValue("id_target_total_area_intensity"),';');
           std::vector<double> id_target_apex_intensity = ListUtils::create<double>((String)feature_it->getMetaValue("id_target_apex_intensity"),';');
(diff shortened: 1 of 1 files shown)


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
```diff
@@ -392,23 +392,50 @@ namespace OpenMS
...
+            for (int i = 0; i < id_target_num_transitions; ++i)
+            {
+              if (id_target_total_mi[i] == "") {
+                id_target_total_mi[i] = "NULL";
+              }
+              if (id_target_ind_mi_score[i] == "") {
+                id_target_ind_mi_score[i] = "NULL";
+              }
+              if (id_target_ind_mi_ratio_score[i] == "") {
+                id_target_ind_mi_ratio_score[i] = "NULL";
```
hroest: why do you expect empty strings? does OpenSWATH write out empty strings?

hroest: PS: also can you change the style to
```
if ... 
{
}
```

hroest:
looks good, some minor questions

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
grosenberger: The transition-level scores are still reported as a concatenated meta feature (e.g. "2;3;4"). If (some of) the MI scores are disabled, an empty string might be reported. This part might be worthwhile to refactor in the future.

I changed the style accordingly.

PR #3550, opened 2018-06-15
Feature/cachedmzml cleanup

hroest:
- create new class CachedMzMLHandler
- move functionality to handler
- create new class CachedmzML with clean interface


hroest:
projectroot.src.tests.class_tests.openms.StopWatch_test  -otherwise it compiles

timosachsenberg:
looks good

PR #3553, opened 2018-06-17
Doc/move

hroest:
moved documentation to https://pyopenms.readthedocs.io/en/latest/index.html


timosachsenberg:
how are these build now?

hroest:
they are now here: https://github.com/OpenMS/pyopenms-extra/blob/master/docs/README.md

and they are built with every push to the repo through https://readthedocs.org/ - they build and host the documentation for us. It also means that people can contribute more easily to the documentation of pyOpenMS

timosachsenberg:
Ok I guess this currently doesn't work easily for OpenMS as we include the command line parameters. But maybe we could use our nightly jenkins jobs to push them to readthedocs?

timosachsenberg:
there seem to be some workarounds required for doxygen: https://breathe.readthedocs.io/en/latest/readthedocs.html

hroest:
it looks like seqan is also using readthedocs: http://seqan.readthedocs.io/en/master/index.html# -- we should have a look at what they are doing, it looks pretty clean, extensive and well put together.

I anyways think that our documentation needs some update, it is outdated in quite a few areas (especially the developer-oriented doc). Also visually it needs some update :-) So we could adjust it in doxygen or think about a (partial) re-write using .rst and doxygen and thus separate the API documentation (which will always be doxygen) from the more user-oriented manual style documentation that seqan has. We are currently lacking a good developer tutorial that incorporates the last 5 years of OpenMS development and I think this is somewhat hampering new developers joining in.

hroest:
> there seem to be some workarounds required for doxygen: https://breathe.readthedocs.io/en/latest/readthedocs.html

I have been using breathe for the Python extension checker and it works quite well

hroest:
> there seem to be some workarounds required for doxygen: https://breathe.readthedocs.io/en/latest/readthedocs.html

I have been using breathe for the Python extension checker and it works quite well

PR #3554, opened 2018-06-17
Feature/aaseq bug

hroest:
Bugfix for AASequence, all bracket modifications would automatically be added to the C-terminus and we would get the following:

```
AASequence::fromString(".DFPIAM[+16]GER.") == AASequence::fromString("DFPIAMGER.(Oxidation)")
```

fixes the bug and adds test cases

PR #3558, opened 2018-06-19
Feature/aaseq fixes

hroest:
multiple fixes for AASequence, mainly for terminal mods

- ensure that bracket mods `[+57]` honor C/N terminal as well as AA specificity
- use mass delta (with sign) or absolute mass output

=> basically make unknown C/N terminal modifications work


hroest:
this was supposed to be follow-up on #3554 however it turned out to be more involved than originally anticipated.

hroest:
so the first part of this is merged in #3559 but there are still a few outstanding issues and bugs -- mainly conceptual and I am not sure whether I can/will address them in the near future. It turns out the current implementation is slightly broken for some edge cases, however the current PR is also slightly broken and I dont have time to address this at the moment.

timosachsenberg:
BTW https://www.ncbi.nlm.nih.gov/pubmed/29397739 seem to not care about the distinction abs mass vs delta mass

src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -908,74 +917,122 @@ START_SECTION([EXTRA] Arbitrary tag in peptides using square brackets)
...
-    AASequence test_seq = AASequence::fromString("[1600.230654]IDE");
+    AASequence test_seq = AASequence::fromString(".[1601.2384790319]IDE");
     TEST_EQUAL(test_seq.size(), 3)
-    TEST_EQUAL(test_seq.toString(), ".[1600.230654]IDE")
-    TEST_STRING_SIMILAR(test_seq.toUniModString(), ".[1600.230654]IDE")
-    TEST_STRING_SIMILAR(test_seq.toBracketString(false), "n[1600.230654]IDE")
+    TEST_STRING_SIMILAR(test_seq.toString(), ".[1601.2384790319]IDE")
+    TEST_STRING_SIMILAR(test_seq.toUniModString(), ".[1601.2384790319]IDE")
+    TEST_STRING_SIMILAR(test_seq.toBracketString(false, false), "n[1601.2384790319]IDE") // an extra H for the N-terminus
+    TEST_STRING_SIMILAR(test_seq.toBracketString(false, true), "n[+1600.230654]IDE")
```
hroest: this is what has been fixed here: `n[1601.2384790319]IDE` is now equivalent to `n[+1600.230654]IDE`

hroest:
looks like travis runs through but none of the other tests...

hroest:
rebuild jenkins

hroest:
only logstream failed, @timosachsenberg @cbielow merge?

src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -1278,6 +1337,15 @@ START_SECTION([EXTRA] testing terminal modifications)
...
   TEST_EQUAL( AASequence::fromString(".DFPIANGEM(UniMod:10)"), AASequence::fromString(".DFPIANGEM.(UniMod:10)") )
+  TEST_EQUAL( AASequence::fromString(".DFPIANGEM.c[-30]"), AASequence::fromString(".DFPIANGEM.(UniMod:10)") )
+  TEST_EQUAL( AASequence::fromString(".DFPIANGEM.c[-29.99]"), AASequence::fromString(".DFPIANGEM.(UniMod:10)") )
+
+  // TO FIX: this is an ambiguous case, toBracketString produces the "absolute"
+  // mass of the n terminal "AA" but its mass is negative, therefore when
+  // parsing it would be interpreted as a delta mass and its off by one!
+  TEST_EQUAL( AASequence::fromString(".DFPIANGEM(UniMod:10)").toBracketString(true), "DFPIANGEMc[-13]") // absolute
+  TEST_STRING_SIMILAR( AASequence::fromString(".DFPIANGEM(UniMod:10)").toBracketString(false), "DFPIANGEMc[-12.9900659681]") // absolute
+  TEST_EQUAL( AASequence::fromString(".DFPIANGEM(UniMod:10)").toBracketString(true, true), "DFPIANGEMc[-30]") // relative
```
hroest: Why does this problem occur?

UniMod 10 is a "Cyanogen bromide (CNBr) cleavage converts the C-term Met to either homoserine or homoserine lactone, depending on pH." so it does not actually modify the C-terminus, it modifies the amino acid, but only if located at the C-terminus. So the modification in that sense is *not* a c-terminal modification, but a modification of the C-terminal amino acid. Otherwise, of course, we would never have a negative absolute mass of anything, that is simply not possible. However in this case its the mass of the C-terminus + the mass of the final M = positive whereas the mass of the C-terminus alone is negative.

http://www.unimod.org/modifications_view.php?editid1=10

hroest:
only logstream failed, @timosachsenberg @cbielow merge?

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -178,15 +188,17 @@ namespace OpenMS
         const String & mod_name = mod.getFullId();
         if (std::find(fixed_modifications.begin(), fixed_modifications.end(), mod_name) == fixed_modifications.end())
         {
+          double nominal_mass = r.getMonoWeight(Residue::Internal);
+          if (mass_delta) nominal_mass = mod.getDiffMonoMass();
+          String sign = (mass_delta && nominal_mass > 0) ? "+" : "";
+          if (aa == "X") {nominal_mass = r.getMonoWeight(Residue::Internal); sign = "";} // cannot have delta mass for X
```
cbielow: this is hard to read, can you make it a proper block?

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -208,15 +220,17 @@ namespace OpenMS
       // only add to string if not a fixed modification
       if (std::find(fixed_modifications.begin(), fixed_modifications.end(), cterm_mod_name) == fixed_modifications.end())
       {
-        double nominal_mass = mod.getDiffMonoMass(); // just get input mass
-        if (!mod.isUserDefined()) nominal_mass += Residue::getInternalToCTerm().getMonoWeight();
+        double nominal_mass = Residue::getInternalToCTerm().getMonoWeight() + mod.getDiffMonoMass();
```
cbielow: there are a few places where this computation is performed, but the result might be discarded immediately after if `mass_delta == true`. For readibility and performance in Debug mode, can you switch to ternary op or if-else. 
~~I dont think its a performance problem in Release~~  (scratch that... no compiler I checked will get rid of the computation... ouch https://godbolt.org/z/xds49Y). Even more reason do change it.

cbielow: e.g.
```
double nominal_mass = mod.getDiffMonoMass();
if (mass_delta) nominal_mass =+ Residue::getInternalToCTerm().getMonoWeight() ;
```

cbielow:
+1

src/openms/source/CHEMISTRY/AASequence.cpp
hroest: > if (mass_delta) nominal_mass =+ Residue::getInternalToCTerm().getMonoWeight() ;

I guess you mean if (!mass_delta)

PR #3560, opened 2018-06-19
[FEATURE] SpectraExtractor: matchSpectrum()

pasqoo:
**Interface:**
```c++
void matchSpectrum(
  const MSSpectrum& input_spectrum,
  const MSExperiment& library,
  std::vector<Match>& matches
);

struct Match
{
  Match(MSSpectrum a, double b) : spectrum(a), score(b) {}
  MSSpectrum spectrum;
  double score;
};
```
The method compares the given spectrum to all the spectra contained in `library`.
The comparison is based on _BinnedSpectrum_ and _BinnedSpectralContrastAngle_.
It is easy to add other comparison techniques.

A cache mechanism is implemented (`bs_cache_`) to avoid creating a `BinnedSpectrum` multiple times for the same `MSSpectrum`. Since the user can choose the bin size, peak spread and the bin offset, the key for such a `BinnedSpectrum` is based not only on the spectrum's name, but also on those params' values (check `extractBinnedSpectrum()`).

The output is supposed to use the MzTab format in future.

You can ignore the _MSPMetaboFile_ code, because it is under review in another PR: https://github.com/OpenMS/OpenMS/pull/3547

The class parameter `min_score_` is renamed to `min_select_score_` to avoid confusion with a new parameter (`min_match_score_`).

**New class parameters:**
```c++
String similarity_function_;
Size top_matches_to_report_;
double bin_size_;
double peak_spread_;
double bin_offset_;
double min_match_score_;
```
Documentation and tests are provided.


timosachsenberg:
Hi,
Cool! Some remarks (as these are rather central algorithms I will be a bit picky ;) .

You are currently trying to solve the design goals (amongst others):
1. The user should be able to select a scoring function on runtime.
2. The scoring functions might process either MSSpectra or BinnedSpectra (which are a different type - bummer)
3. Your class should have the same interface (accept MSExperiment)

This leads to the kind of ugly situation that your class has a lot of binned spectrum related stuff in there (parameter and a cache). Also, if the user chooses a different scoring function (let's say one that works on MSSpectra) what happens with the cache and all the other stuff? 

IMHO a better design would be not to mix these as it will most likely lead to quite a lot of bookkeeping and code complexity that is not needed.

An interface roughly like (just to get the idea):
```
  SpectrumMatchType matchSpectrum(const MSSpectrum & s, const MSExperiment & m, PeakSpectrumCompareFunctor * compare);
```

and:
```
 SpectrumMatchType matchBinnedSpectrum(const BinnedSpectrum & s, const vector<BinnedSpectrum> & m, BinnedPeakSpectrumCompareFunctor * compare);
```
looks cleaner to me. SpectrumMatchType could be e.g. a score, spectrum index pair.

The caller could then make sure to provide the appropriate spectrum type.

Regarding memory consumption (if that was the reason for the cache): if reading the full database is really a problem there are better ways to address the problem (actually I think it isn't for metabolites and you could just convert everything to the binned representation). E.g., you can do the search in two steps (first determining candidates (by matching precursor m/z) and then loading/converting the actual spectra to match them. But I would only implement this if required.

pasqoo:
Thanks @timosachsenberg for the input. I will try to address the issues asap and get something elegant out of it.

About the reasons behind the cache, it was about time consumption on creating 220k `BinnedSpectrum`s (the size of the library I was using). Basically the first call to `matchSpectrum()` (so when all the `BinnedSpectra` are created) would take 2.5+ seconds, but the subsequent calls would take 1 (or better) second less (which is a big difference, if you're going to compare many spectra against the library).

timosachsenberg:
I see. But if you create them all once before matching, then it should even be slightly faster to match them

timosachsenberg:
Thinking about it a bit more, I think it would be sufficient for now to just support the scoring functions that take a BinnedSpectrum. Then you could use the factory method to retrieve the comparison functor. 
I would then probably precreate the BinnedSpectrum (just takes 2.5 seconds as you said so this is probably no big deal) and match against these. I really would not do the caching on the fly but precalculate the binned spectra if possible.

pasqoo:
Pre-caching also means that I would not pass the library to the method. It's also assumed that the library doesn't change between calls, at that point (sounds fair). One should also add methods to manage the library (ie. add a spectrum, lookup for one, etc). I wonder how all of this would look in the same class.

pasqoo:
I don't think I had understood the problem correctly in my previous comment.
I went for a rewrite of the feature.
I implemented a base class `Comparator` with two pure virtual functions inside.
To add another similarity function, one would need to derive from such base class as I did with `BinnedSpectrumComparator`.

Is this close to what you were recommending, @timosachsenberg ?

If it looks good, I'll proceed with a cleanup of the code/comments/update docs.

pasqoo:
I fixed the tests by making sure that the sorting is correct, instead of checking for a specific match name.
So it is checked that the first match has a higher score than the second.

Now only pyOpenMS needs to be fixed.

pasqoo:
I'm not sure of how to solve the pyOpenMS related issues. @hroest any hint?

hroest:
there is a problem with "generateScores" here, mainly it should be `libcpp_pair` instead of `libcpp_cpp`:

```
Error compiling Cython file:
...
    cdef cppclass TSE_Comparator "OpenMS::TargetedSpectraExtractor::Comparator":

        TSE_Comparator() nogil except +
        TSE_Comparator(TSE_Comparator) nogil except +

        void generateScores(MSSpectrum& spec, libcpp_vector[libcpp_cpp[Size,double]]& scores, double min_score) nogil except +
                                                                      ^

/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/pyOpenMS/pxds/TargetedSpectraExtractor.pxd:54:71: unknown type in template argument
```

https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_docker_pyopenms/651/console

pasqoo:
Is this relevant? https://stackoverflow.com/questions/13976809/abstract-classes-with-pure-virtual-methods-in-cython

pasqoo:
The tests keep timing out

hroest:
looks good to me, the only failing tests are not related to the PR

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -131,6 +135,51 @@ namespace OpenMS
...
+    //   "Peak spread for binned spectral contrast angle similarity function."
+    // );
+    // params.setMinFloat("peak_spread", 0.0);
+
+    // params.setValue(
+    //   "bin_offset",
+    //   0.4,
+    //   "Bin offset for binned spectral contrast angle similarity function."
+    // );
+    // params.setMinFloat("bin_offset", 0.0);
```
hroest: can you remove this?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.h
```diff
@@ -309,17 +424,60 @@ namespace OpenMS
...
+//     TargetedSpectraExtractor_friend() = default;
+//     ~TargetedSpectraExtractor_friend() = default;
+
+//     // const BinnedSpectrum& extractBinnedSpectrum(const MSSpectrum& s)
+//     // {
+//     //   return tse_.extractBinnedSpectrum(s);
+//     // }
+
+//     TargetedSpectraExtractor tse_;
+//   };
```
hroest: can you remove unused comments / unused code?

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -473,4 +522,113 @@ namespace OpenMS
...
+  //   const String bs_key = s.getName() + String(bin_size_) + String(peak_spread_) + String(bin_offset_);
+  //   std::unordered_map<std::string,BinnedSpectrum>::const_iterator it = bs_cache_.find(bs_key);
+  //   if (it == bs_cache_.cend())
+  //   {
+  //     std::pair<std::unordered_map<std::string,BinnedSpectrum>::const_iterator,bool>
+  //       p = bs_cache_.emplace(bs_key, BinnedSpectrum(s, bin_size_, false, peak_spread_, bin_offset_));
+  //     it = p.first;
+  //   }
+  //   return it->second;
+  // }
```
hroest: same here, please remove unused code

hroest:
from my side it looks good, please remove some commented out code. @timosachsenberg any comments from your side?

pasqoo:
@hroest The pxd file is not fixed, though. I just commented all the lines pertaining to this feature. Do you have suggestions?

timosachsenberg:
Looks good from my side.
Some final thoughts:
If precursor m/z annotations are present in the spectral library, this information could be used to greatly reduce the search space. e.g. only spectra matching in m/z tolerance window need to be compared

pasqoo:
The failed test is a timeout, which actually shows all tests as passed: https://api.travis-ci.org/v3/job/402605629/log.txt
It failed during the last steps of the CI (packaging?)

jpfeuffer:
should be fine. was probably a matter of seconds. Ready to merge from your side?

pasqoo:
Yes, the PR is up to date. Can't do much on my own about the .pxd. Other than that, everything is in place.

PR #3563, opened 2018-06-19
Feature/toppview chrom

hroest:
- cleanup of the chromatogram code in TOPPView
- allow loading of chromatograms from disk using indexed mzML


hroest:
mainly refactor `TOPPViewSpectraViewBehavior.cpp` and extract a function with common functionality. Also some fixes to allow chromatograms to be cached on disk (similar to raw spectra)

PR #3565, opened 2018-06-19
Feature/py fixes

hroest:


timosachsenberg:
the is a potential unused import "breath" see codacy

hroest:
> the is a potential unused import "breath" see codacy

actually no, it is used to determine whether `breathe` is there and inform the user how to run the script.

PR #3566, opened 2018-06-20
[FIX] FalseDiscoveryRate.cpp: fixed a bug where having decoy hits but…

poshul:
… no actual hits would cause a seg fault


jpfeuffer:
Looks reasonable! Thanks.

timosachsenberg:
sound like something that could have a regression test added

PR #3568, opened 2018-06-22
[FIX] parsing of unknown modifications

timosachsenberg:


timosachsenberg:
@lars20070 as I am on vacation could you add a regression test?

lars20070:
@timosachsenberg No crashes. Thanks. - But mods still as `unknown` in `idXML`.
```
  <Peptide id="Pep_SYSPRSSPPR">
    <PeptideSequence>SYSPRSSPPR</PeptideSequence>
    <Modification location="0" monoisotopicMassDelta="17.031558">
      <cvParam cvRef="PSI-MS" accession="MS:1001460" name="unknown modification" value="Methyl:2H(2)13C"/>
    </Modification>
  </Peptide>
```
screen shot 2018-06-22 at 12 26 33

timosachsenberg:
The code snipped is from the mzID

lars20070:
@timosachsenberg Yes, I will add the tests.

timosachsenberg:
Regarding the idXML I can check tomorrow

timosachsenberg:
Ok I fixed this. 
One failing test remains that I need to take a look. Currently, if name is "unknown_modification"
but "value" of the cvParam is "" we need to fall back to adding a bracket modification.

timosachsenberg:
File and tests are fixed now.
 @hroest you recently took a look at the unknown modification code. Do we have modifications with name "" , the empty string in our ModificationsDB?
I needed to add a check that not modification "" is set (which seems to be valid as no error was thrown)

if (cv.hasValue() && ModificationsDB::getInstance()->has(cvvalue) && !cvvalue.empty())  // why do we need to check for empty?

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -2362,15 +2368,28 @@ namespace OpenMS
...
                     }
                     if (index == 0)
                     {
                       try // does not work for cross-links yet, but the information is finally stored as MetaValues of the PeptideHit
                       {
-                        aas.setNTerminalModification(cv.getName());
+                        if (cvname == "unknown modification")
+                        {
+                          const String & cvvalue = cv.getValue();
+                          if (ModificationsDB::getInstance()->has(cvvalue))
```
hroest: I guess you should also check for empty here?

hroest:
> Do we have modifications with name "" , the empty string in our ModificationsDB?

not that I know of, but apparently we do. I could look into this

src/tests/topp/CMakeLists.txt
```diff
@@ -837,6 +837,10 @@ set_tests_properties("TOPP_IDFileConverter_24_out1" PROPERTIES DEPENDS "TOPP_IDF
 add_test("TOPP_IDFileConverter_25" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/IDFileConverter_25_input.idXML -out IDFileConverter_25_output.tmp -out_type pepXML)
 add_test("TOPP_IDFileConverter_25_out1" ${DIFF} -in1 IDFileConverter_25_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_25_output.pep.xml)
 set_tests_properties("TOPP_IDFileConverter_25_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_25")
+# MS-GF+ mzid with non-Unimod modifications
+add_test("TOPP_IDFileConverter_26” ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/IDFileConverter_26_input.mzid -out IDFileConverter_26_output.tmp -out_type idXML)
+add_test("TOPP_IDFileConverter_26_out1" ${DIFF} -in1 IDFileConverter_26_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_26_output.idXML)
```
lars20070: This `TOPP_IDFileConverter_26_out1` is not run. But I cannot see the problem.

timosachsenberg: " vs ”

timosachsenberg:
@lars20070  I'll fix it

timosachsenberg:
k I added the extra checks

PR #3570, opened 2018-06-23
Fix/kr fix

hroest:
fix KR switch for case of terminal AA modification (not C/N terminal)


grosenberger:
This seems to work now. There is a debug statement still active from the last PR:

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp:
```
       std::cout << " last AA " << peptide.sequence[lastAA]  << std::endl;
       peptide.sequence[lastAA] = 'R';
       std::cout << " last AA " << peptide.sequence[lastAA]  << std::endl;
       std::cout << " after " << peptide.sequence << std::endl;
```

grosenberger:
Ok, I now did some more testing with the KR switch mode and it seems to have a positive effect by lowering the amount of too many high scoring decoys. Maybe we should consider using this mode by default in the future.

grosenberger:
We should probably make it a default for shuffle. For pseudo-reverse, it might be useful option, but except in very difficult situations, it shouldn't be a big problem.

I tested this with the empirical p-value option in PyProphet and while the estimated TPs are similar at q-value 0.02-0.5, the estimated TPs are higher with the new option with q-value < 0.01. We previously ran into issues with TRIC, where not enough TPs with q-value < 0.0001 existed and with the new option, this is not an issue anymore.

grosenberger:
From a user point of view, my recommendation would be to move the option to the advanced settings in ``OpenSwathDecoyGenerator`` and by default set it to true. The vast majority of people will want to use the default, minimal options.

hroest:
> advanced settings in OpenSwathDecoyGenerator and by default set it to true.

then it would also be true for pseudo-reverse, would that be ok?

grosenberger:
I think that would be ok, with the new more robust defaults in "shuffle", that's the main mode that we would recommend anyway. Also, it will be good to have a different precursor m/z from the corresponding targets.

PR #3574, opened 2018-06-25
[FIX] MSPGenericFile: Assume one metadatum per line

pasqoo:
The `develop` version of this class parses multiple metadata on the same line, assuming they are separated by a `;`.

I noticed that some libraries use the character `;` within the `Comments` metadatum, and that breaks the implementation.

If one exports the NIST library to msp format (demo: https://chemdata.nist.gov/dokuwiki/doku.php?id=chemdata:nist17), he/she'll find that `CAS#` and `NIST#` are on the same line, separated by a `;`.

It is nice to support an important library as NIST's, therefore I went for a simple solution: keep supporting the above line, but do not assume multiple metadata on a single line anywhere else, because character `;` is not universally used as a separator for metadata.


pasqoo:
Only _LogStream_test_ failed: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-15,os_label=win10_64/2060/testReport/junit/projectroot.src.tests.class_tests/openms/LogStream_test/

PR #3575, opened 2018-06-25
[FIX, FEATURE] some code refactoring and small fixes

timosachsenberg:
- moved some code from FFM tool to algorithm
- FIX: update meta value "map_index" if consensus maps are appended
- more verbose import of experimental design from consensus map


lars20070:
Do we still require unique IDs for the features?
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderMultiplexAlgorithm.cpp#L749

timosachsenberg:
How to test that? Tests passed...

lars20070:
IMO, `unique_id` can be optional.

Hence, +1 for merge.

lars20070:
I think `-test` takes care of the unique ids, i.e. we get the same, identical ID every time the tool is run.

PR #3577, opened 2018-06-26
[FEATURE] SpectrumHelper: slicePeakContainer() and reZeroIntensities()

pasqoo:
`slicePeakContainer()` takes a MS container (`MSSpectrum` or `MSChromatogram`) and a range. It modifies the container in place, removing all those peaks that do not fall within the range (start and end positions are inclusive).

`reZeroIntensities()` takes a MS container (`MSSpectrum` or `MSChromatogram`) and modifies it in place, making sure the lowest intensity has value `0`.


timosachsenberg:
Some thought:
- the stl way would be to use remove_if (for the first method) and transform for the second method.
- this does not apply to the data arrays in stored in a Spectrum (which may lead to nasty bugs if these arays are then of different size)

An idea:
We could add a transformPeaks(f) and removePeaks(f, bool ignoreDataArrayss=false) method to MSSpectrum/MSChromatogram that takes a functor/lambda f as argument. That way we have full control of updating the associated data arrays.
@hroest what is your take on that?

pasqoo:
Would `std::remove_if()` be slower than two calls of `container.erase()`?

I can work into using `std::transform()` for the second method. It shouldn't add any annoyance.

I understand it might be undesirable to ignore the data arrays, and that's why I inserted a note within the code. I like the `bool` argument idea.

timosachsenberg:
Regarding speed... that's a good question and probably depends on the size of the spectrum and where stuff gets erased. If only a small fraction is removed, I guess your solution is faster for large spectra.

If I am not mistaken:
Your solution uses two log(n) look-ups of start and end. A constant time erase + a second linear time erase (as it involves destruction + move).
The remove_if (note that it also requires an erase after remove_if ) solution requires a linear scan through the vector + a linear time complexity destruction of erased elements.

pasqoo:
If elements are _moved_ at every _removal_, that might be a negative point. Especially if the container is visited starting from the left side. I guess this can be avoided by passing a different kind of iterator ie. reverse? Never used these, might be recommending something incorrect. The idea is to go from right to left.
But doing _n^2_ moves (which is the worst case: a container outside of the requested start/end range) looks less promising than the current implementation.
An advantage of the `std::remove_if` approach is that the container could also be not sorted, and the algorithm would work anyway.
Should I add a `.isSorted()` check and, in case of negative, sort the array? Or would documenting this point be enough? I.e. `@warn The input container MUST be sorted`.

timosachsenberg:
In both implementations, if you remove elements at the beginning or middle of the vector, the gaps need to be filled by swapping (moving) in the elements (e.g., after the gap).

pasqoo:
What I'm wondering is if the following scenario is happening for the two cases.

Assume elements with RT = 3 4 5 6 7 8 9 10 11 12. I desire range [10 11].

### `std::remove_if` case:
1. remove element 3
2. move all the subsequent elements to the left
3. remove element 4
4. remove all the subsequent elements to the left
5. and so on

**Disadvantage:** the same elements are moved multiple times

### **Current implementation case:**
1. remove the elements at the end (_O(1)_)
2. remove the elements at the beginning (_O(n)_)
3. subsequent elements are moved to the left

**Advantage:** the subsequent elements are moved only once instead of a number of times equal to the number of elements with RT lower than input `pos_start`

timosachsenberg:
No. See https://en.cppreference.com/w/cpp/algorithm/remove

pasqoo:
Thanks Timo, I have already checked that page, and its description (which I might misinterpret) and _Possible implementation_ section show that a `std::move()` is executed for each element. Basically the situation I explained in the `remove_if` case of my previous comment.

The complexity is said to be "Exactly std::distance(first, last) applications of the predicate." which in our case translates to a number of calls to `p.erase()`.

Am I getting something wrong? If so, could you please point me the relevant part of the page?

timosachsenberg:
I think the difference is that not all subsequent ones are moved again and again but only once.

pasqoo:
By looking at the gcc implementation, basically the elements are simply moved (`std::move()`) to the left in the sequence. The size of the container is not modified, which means one would have to call `erase()` at the end.
So the moves should be efficient, and `erase` is only called once.
Better explained here: https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom
And the code is here: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_algo.h#L858

src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
```diff
@@ -66,6 +66,84 @@ namespace OpenMS
...
+
+  template <typename PeakContainerT>
+  void removePeaks(
+    PeakContainerT& p,
+    const double pos_start,
+    const double pos_end,
+    const bool ignoreDataArrays = false
+  )
+  {
+    // typename PeakContainerT::iterator it = std::remove_if(p.begin(), p.end(),
```
timosachsenberg: I think your implementation is fine and you can remove the comment

pasqoo:
@timosachsenberg PR updated. One test timed out.

PR #3587, opened 2018-07-02
Fix/aa as search

timosachsenberg:
fixes:
https://github.com/OpenMS/OpenMS/issues/3583
https://github.com/OpenMS/OpenMS/issues/3582


timosachsenberg:
time outs

PR #3592, opened 2018-07-05
[FEATURE] Integrate EMG into PeakIntegrator workflow

pasqoo:
_EmgGradientDescent_ now supports `left` and `right` arguments.
_PeakIntegrator_ has a `"fit_EMG"` parameter that, if `true`, fits the EMG peak model to the interested peak part. The integration/bg_estimation/metrics computation continues on the interested part.


jpfeuffer:
This broke the library I guess. Can you have a look?

Issue #3581, opened 2018-06-28
TOPPView deselection of window leads to jump to last column in identification view


timosachsenberg:

hroest:
also happens for spectrum view

PR #3595, opened 2018-07-05
[FIX] only update table model if new layer is selected

timosachsenberg:


timosachsenberg:
I could not reproduce the jump / deselection in the scan view (only in id view)
(see https://github.com/OpenMS/OpenMS/issues/3581)

enetz:
With these changes the issue of TOPPView resetting the ID view upon regaining focus again is resolved, but the jump to the end of the ID table (horizontally, to the last column) when e.g. switching between tabs within TOPPView as mentioned in #3581 still persists.  This happens e.g. when you still have the MS1 map open after loading IDs and switch between the MS1 view and the ID view with MS2 in a second tab.

timosachsenberg:
k lets merge this fix then and I will try to add another one

PR #3596, opened 2018-07-05
[Fix]ed XTandem bug with multiply ocurring proteins.

### LINKED ISSUE #3594: xtandemAdapter reports incorrect protein evidance
Hello,

It seems like xtandemAdapter reports wrong proteins for peptides in idXML (OpenMS version 2.3.0). 

I have attached a very small example of XTandem xml file. The workflow is simply an IDConverter followed by TextExporter. 

It reports that peptide VPLRSILE is coming from sp|Q03517|SCG2_MOUSE and sp|Q99JA0|CALCA_MOUSE but it should report CALC_MOUSE and CALCA_MOUSE. 

In version 2.1.0 it works good. 

[test.zip](https://github.com/OpenMS/OpenMS/files/2166295/test.zip)

Best,
Payam

jpfeuffer:
fixes #3594

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/XTandemXMLFile.h
+++ b/src/openms/include/OpenMS/FORMAT/XTandemXMLFile.h
@@ -104,6 +104,10 @@ private:
     // true during "note" element containing spectrum ID
     bool is_spectrum_note_;
 
+    // true after non-new protein entries, so that with the next "protein note" the
+    // accession will not be updated again
+    bool skip_protein_acc_update_;
+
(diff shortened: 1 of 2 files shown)


timosachsenberg:
There seems to be a failing test on windows

jpfeuffer:
Weird, I would have thought it only affects parsing and here it seems as the xtandem executable fails.

jpfeuffer:
rebuild jenkins

timosachsenberg:
failed test unrelated

PR #3599, opened 2018-07-06
[FIX] jump to column in TOPPView

timosachsenberg:
simplified code a bit and fixes https://github.com/OpenMS/OpenMS/issues/3581 in id view


enetz:
:+1:

PR #3600, opened 2018-07-06
[FIX] don't annotate singly charged immonium ions with multiple charg…

timosachsenberg:
…ed fragment shifts

Reported and fix validated by Aleksandar


enetz:
The changes in RNPxlSearch look fine to me, but how are the contrib changes in this PR related to the fix?

timosachsenberg:
@jpfeuffer the contrib changes are probably in the newer contrib. should I pull them in?

timosachsenberg:
talked to @jpfeuffer . should go in

Issue #3466, opened 2018-05-17
protein inference support in Percolator adapter


lars20070:
Protein inference in `Percolator` is a powerful algorithm. In our tests, the tool scales well and performs similarly to `ProteinProphet`. It would therefore be very useful to implement support for the protein inference part of `Percolator` in the `PercolatorAdapter` as well.

MatthewThe:
I think there's already support for this, have you tried setting `protein-level-fdrs` to true?

wenjiebi:
@MatthewThe In the aspect of protein inference, I compared PercolatorAdapter with FidoAdapter. The workflow, filtered mzML file, database and results are [Here](https://www.dropbox.com/sh/gv2r0gol4ztd1hl/AACk3B5eXLHqIbnWL_8PdjGOa?dl=0). There are much fewer protein generated by Percolator

  | Percolator | Fido
-- | -- | --
PRT | 5725 | 8784
PSM | 9092 | 9054

I don't know if the parameters were set correctly. Can you please help to check it?

MatthewThe:
There are several independent issues at play here:
- FidoAdapter creates protein groups based on the observed peptides (last column in the mzTab = `indistinguishable_group` instead of `single_protein`). If you just count the `single_protein` rows you will also have 5725 proteins.
- If you're trying to assess the performance based on the number of _significant_ proteins, you will have to set some threshold, instead of simply counting _all_ proteins as you do now. However, this is a bit tricky since FidoAdapter only seems to report the `Posterior Probability` and no FDR. One way to compare would be to compare "1 - `Posterior Probability`" of the FidoAdapter output to the field `MS:1001493` (Percolator:PEP, which is the Posterior Error Probability) of the PercolatorAdapter output. You can also calculate the FDR from the FidoAdapter results by sorting by "1 - `Posterior Probability`" in ascending order and calculate a running average of the resulting Posterior Error Probabilities until this exceeds your FDR threshold.
- It seems your workflow has not incorporated #3462 yet, so the PercolatorAdapter results are not really useful at the moment.

wenjiebi:
Thank you very much for your response.
To the first point, indeed the amount of `single_protein` are the same in FidoAdapter and PercolatorAdapter. Can we also get the `ìndistinguishable_group` in PercolatorAdapter? Because they are also useful information so we would expect them in the output of PercolatorAdapter.

MatthewThe:
[...]

PR #3601, opened 2018-07-06
[FIX] Correction of the observed mass if isotope error is given

sophiamersmann:
If an isotope error is given (e.g. when using the search engine MSGF+), the calulated mass is corrected by subtracting `(isotopeError * protonMass) / charge`.

### DIFF
--- a/src/topp/PercolatorAdapter.cpp
+++ b/src/topp/PercolatorAdapter.cpp
@@ -418,12 +418,17 @@ protected:
         double calc_mass = hit.getSequence().getMonoWeight(Residue::Full, charge)/charge;
         hit.setMetaValue("CalcMass", calc_mass);
 
-        if (hit.metaValueExists("IsotopeError"))
+        if (hit.metaValueExists("IsotopeError"))  // MSGFPlus
         {
-          float isoErr = stof(hit.getMetaValue("IsotopeError").toString());
+          float isoErr = hit.getMetaValue("IsotopeError").toString().toFloat();
           exp_mass = exp_mass - (isoErr * Constants::C13C12_MASSDIFF_U) / charge;
         }
-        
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Related to #3466

timosachsenberg:
thanks! did you check if different search engines: e.g., Comet, MSGFPlus and our SimpleSearchEngine write use the same MetaValue key? and also numbering scheme (e.g., one could use 1, the other -1)

timosachsenberg:
Hmm did not find the meta value for comet. Can check tomorrow, maybe we don’t parse it?

timosachsenberg:
we also use "isotope_error" error and use 0 for the monoisotopic peak and 1 if the experimental precursor annotation points to the first isotopic peak

timosachsenberg:
I am not 100% sure that the caluclation is correct.
calc_mass as in this code does not need any isotopic correction as it is calculated from the theoretical composition. The actual istopic misassignment error is on the experimental precursor mass.

sophiamersmann:
You are absolutely right about calc_mass / exp_mass. I updated the code - exp_mass is now corrected rather than calc_mass.

sophiamersmann:
I will need to check if different search engines use the same MetaValue key and numbering scheme.

src/topp/PercolatorAdapter.cpp
```diff
@@ -416,7 +417,12 @@ class PercolatorAdapter :
         
         double calc_mass = hit.getSequence().getMonoWeight(Residue::Full, charge)/charge;
         hit.setMetaValue("CalcMass", calc_mass);
-        
+
+        if (hit.metaValueExists("IsotopeError"))
+        {
+          float isoErr = stof(hit.getMetaValue("IsotopeError").toString());
```
timosachsenberg: stof uses the current locale and might thus not work for files produced on a system with a different locale.
use the toString().toFloat() instead (if toFloat() doesn't work)

PR #3603, opened 2018-07-09
[FIX] added delta mass mod to sequence so e.g. percolator can calcula…

timosachsenberg:
…te a proper mass error


src/openms/source/ANALYSIS/RNPXL/RNPxlReport.cpp
```diff
@@ -248,6 +248,37 @@ namespace OpenMS
...
+              AASequence new_aa = AASequence::fromString("[" + String(rna_weight) + "]." + seq);
+              ph.setSequence(new_aa);
+            }
+            else // place it anywhere
+            {
+              for (auto a : aa)
+              {
+                if (!a.isModified())
+                {
+                  a.setModification("[" + String(rna_weight) + "]");
```
hroest: @timosachsenberg arent you placing the mod multiple times ?

timosachsenberg: good catch! never happened on the test data that both N and C term were modified

PR #3604, opened 2018-07-09
[FIX, DOC] increases min Qt Version to 5.5

timosachsenberg:


jpfeuffer:
Yep. Only thing that I am not sure about is if this variable is propagated down correctly. Later on, we search Qt again for packaging (Windows and KNIME).
But this is more a general problem (we should not search again). See #3569 where we e.g. used that quick fix.

PR #3612, opened 2018-07-10
correct peptide intensities in FeatureFinderMultiplex

lars20070:
The simple uncorrected peptide intensities in FFM are better than the corrected ones. The corrections are simply wrong and should be removed for the OpenMS 2.4 release. See *corrected* vs *un-corrected* in the labelled fixed-ratio 1:1 analyses below. 

A more elaborate peptide intensity calculation with correct volume calculation will follow.

old new


timosachsenberg:
Also,
How are peptide Quants if rt shifts are present

timosachsenberg:
also please remove afaik unused code like the correction stuff

hroest:
> also please remove afaik unused code like the correction stuff

I would not remove it but I wonder whether for the release its better to use the correction version or the raw version. I would prefer a solution with an option to choose one or the other

lars20070:
@hroest The corrections are simply wrong. That's why I removed them as Timo suggested.

I now systematically optimise params on *labelled fixed ratio* and *unlabelled spike-in* datasets. In the same go, I compare to MaxQuant, see [`MQ2mzTab`](https://github.com/ProteomicsScripts/ProteomicsScripts/pull/10), and modify the algorithm. A good algorithm with bad params might perform worse than a bad algorithm with optimised params.

timosachsenberg:
can you think about a regression test to prevent such errors in the future? Maybe extracting a tiny region with one (ideally more) features from the fixed ratio dataset and explicitly testing for the ratio (e.g., in the class test)

lars20070:
@timosachsenberg The TOPP tests are based on a fixed ratio `H:L = 1:4` dataset. `FFM` reports the correct ratios.
ffm_1to4

timosachsenberg:
can we add this fixed ratio to a test? e.g., check if median ratio is close to 4? probably easiest done in a class_test

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderMultiplexAlgorithm.cpp
```diff
@@ -236,7 +236,7 @@ namespace OpenMS
   
   std::vector<double> FeatureFinderMultiplexAlgorithm::determinePeptideIntensitiesCentroided_(const MultiplexIsotopicPeakPattern& pattern, const std::multimap<size_t, MultiplexSatelliteCentroided >& satellites)
   {
-    // determine RT shift between the peptides
+    // determine peptide intesities and RT shift between the peptides
```
timosachsenberg: typo

hroest:
looks good, should we merge?

PR #3614, opened 2018-07-11
[FEATURE] Implement targeted/untargeted spectra matching

pasqoo:
Methods implemented:
```c++
void targetedMatching(
  const std::vector<MSSpectrum>& spectra,
  const Comparator& cmp,
  FeatureMap& features
);

void untargetedMatching(
  const std::vector<MSSpectrum>& spectra,
  const Comparator& cmp,
  FeatureMap& features
);
```

These two methods make use of `matchSpectrum()`.
These take a `FeatureMap` in input, and try to enrich its content with what is found within the spectral library.
The difference between the two methods is that the latter also executes peak picking on the input spectra, filling the `FeatureMap` with the initial info (which, in the case of the former method, is already present as the output of `extractSpectra()`).

Documentation and tests are provided.


hroest:
see for failed test https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=appleclang-9.0.0,os_label=highsierra/2124/testReport/junit/projectroot.src.tests.class_tests/openms/TargetedSpectraExtractor_test/

pasqoo:
@hroest It is not the first time I experience this kind of issue with the spectra matching method.
I work on ubuntu 18.04, and that CI machine 's tests all passed.
On windows and apple it fails. Is it normal that computation changes so much in scoring?
I will look for bugs within the code.

pasqoo:
Maybe it is because of `selectSpectra()` using an `unordered_map()`. Which would explain why the tests on the "untargeted" version don't fail between different platforms.

I'll try using a `map`, to see if tests' results change.

pasqoo:
@hroest @timosachsenberg Is there some issue with the win/mac/ubuntu CIs? These did not queue, twice.

jpfeuffer:
Win VM was unreachable.

jpfeuffer:
It's up again

pasqoo:
Thanks @jpfeuffer 

@hroest @timosachsenberg Seems like the issue was really the spectra selection step.
I could change the tests by making sure that the first score > second score > etc and keep using `unordered_map`, as I did in `matchSpectrum()` tests, but for now it's ok to me to accept the `map` usage.

If I find out that `unordered_map` gives us a performance boost, then I'll change it back in future and update the tests in a different PR.

PR #3615, opened 2018-07-11
[FEATURE] Option to specify the type of peptide main scores

sophiamersmann:
Allow the user to specify which variable is set as main score for peptide hits. Valid scores to choose from are q-value, posterior error probability (PEP), and posterior probability (1 - PEP).

### DIFF
--- a/src/topp/PercolatorAdapter.cpp
+++ b/src/topp/PercolatorAdapter.cpp
@@ -229,6 +229,8 @@ protected:
     registerFlag_("peptide-level-fdrs", "Calculate peptide-level FDRs instead of PSM-level FDRs.");
     registerFlag_("protein-level-fdrs", "Use the picked protein-level FDR to infer protein probabilities. Use the -fasta option and -decoy-pattern to set the Fasta file and decoy pattern.");
     registerStringOption_("osw_level", "<osw_level>", "ms2", "OSW: Either \"ms1\", \"ms2\" or \"transition\"; the data level selected for scoring.", !is_required);
+    registerStringOption_("score_type", "<type>", "q-value", "Type of the peptide main score", false);
+    setValidStrings_("score_type", ListUtils::create<String>("q-value,Posterior Error Probability,Posterior Probability"));
 
     //Advanced parameters
     registerFlag_("generic-feature-set", "Use only generic (i.e. not search engine specific) features. Generating search engine specific features for common search engines by PSMFeatureExtractor will typically boost the identification rate significantly.", is_advanced_option);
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Great. I think though that both PP and PEP are not necessary. We usually have automatic conversions in place where needed. Instead I would also allow the svm_score as it is now.

src/topp/PercolatorAdapter.cpp
```diff
@@ -229,6 +229,8 @@ class PercolatorAdapter :
     registerFlag_("peptide-level-fdrs", "Calculate peptide-level FDRs instead of PSM-level FDRs.");
     registerFlag_("protein-level-fdrs", "Use the picked protein-level FDR to infer protein probabilities. Use the -fasta option and -decoy-pattern to set the Fasta file and decoy pattern.");
     registerStringOption_("osw_level", "<osw_level>", "ms2", "OSW: Either \"ms1\", \"ms2\" or \"transition\"; the data level selected for scoring.", !is_required);
+    registerStringOption_("score_type", "<type>", "q-value", "Type of the peptide main score", false);
+    setValidStrings_("score_type", ListUtils::create<String>("q-value,Posterior Error Probability,SVM Score"));
```
grosenberger: Maybe the input options would be better without spaces? E.g. "qvalue", "pep", "svm"? This makes it easier to pass the parameters via CLI.

src/topp/PercolatorAdapter.cpp
```diff
@@ -1028,12 +1032,32 @@ class PercolatorAdapter :
...
           else
           {
-            hit->setScore(1.0); // set q-value to 1.0 if hit not found in results
+            if (scoreType == "q-value" || scoreType == "Posterior Error Probability")
+            {
+              hit->setScore(1.0); // set q-value or PEP to 1.0 if hit not found in results
+            }
+            else if (scoreType == "SVM Score")
+            {
+              hit->setScore(0.0); // set SVM score to 0.0 if hit not found in results
```
grosenberger: Is it possible that Percolator produces scores between -1.0 and 1.0? I'm not sure if I remember correctly, but in that case we should set the SVM score to the lower boundary.

sophiamersmann: SVM scores can be negative indeed but they are not constrained to lie in a (-1,1) interval. The absolute value of an SVM score corresponds to the distance to the SVM's margin. So, I am not really sure what to use as a lower boundary here. Just a high negative number, e.g. 1000?

sophiamersmann: Sry, -1000 of course...

sophiamersmann: Apparently, it happens quite a lot - at least for the test data set I used (over 500 times). Not sure why though.

timosachsenberg: I am not sure if that is the same reason for the missing scores but I fixed it for my data in:
https://github.com/OpenMS/OpenMS/pull/3630
The reason in my case was, that terminal modifications with bracket notation e.g. PEPTIDE.[1.2345] was removed and the peptide not properly mapped to the PercolatorResult.

Issue #3588, opened 2018-07-03
TOPPView readd fragment annotation column in advanced view (user request)


timosachsenberg:

timosachsenberg:
done

PR #3618, opened 2018-07-12
[FEATURE] TOPPView column showing peak annotations as text

enetz:
The column appears when "Show advanced annotations" is ticked and `PeakAnnotations` exist in the `PeptideHits`.
Adresses #3588 .

For me it looks like the resizing of the column may not work correctly, it resizes to the longest annotation for one dataset with shorter annotations, but not for another with some much longer strings.
There may be a maximal limit to the automatic resizing with `table_widget_->resizeColumnsToContents()`

Also the horizontal scrolling method that sticks to columns makes it difficult to scroll this wide column.

At least exporting the table as .csv exports all the annotations correctly, regardless of the column size.

### DIFF
--- a/src/openms/include/OpenMS/METADATA/PeptideHit.h
+++ b/src/openms/include/OpenMS/METADATA/PeptideHit.h
@@ -130,11 +130,27 @@ public:
 
     bool operator==(const PeptideHit::PeakAnnotation& other) const
     {
-      if (charge != other.charge || mz != other.mz || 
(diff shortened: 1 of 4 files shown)


timosachsenberg:
I think it is fine as a temporary solution. Better would probably be just add "show annotations" to the column and open a new window that contains a table with ion name, mz, intensity columns.
Should also be rather easy to implement...

timosachsenberg:
looks good!

PR #3619, opened 2018-07-12
Feature/doc tutorial

hroest:
fix tutorial, add documentation for python module


timosachsenberg:
+1

doc/code_examples/Tutorial_IdentificationClasses.cpp
```diff
@@ -99,8 +100,8 @@ int main()
   peptide_id.setMZ(440.0);
   peptide_id.setScoreType("ScoreType");
   peptide_id.setHigherScoreBetter(false);
-  peptide_id.setIdentifier("Method");
-  
+  peptide_id.setIdentifier("Identifier");
```
hroest: this is actually a bug in the tutorial, if the peptide and protein identifier dont match, they dont get stored!

PR #3620, opened 2018-07-12
idXML bug: rank not stored

hroest:
this adds a test and uncovers a bug where the rank of the peptide hit will not be stored in the idXML and is lost after reading in the file from disk


src/tests/class_tests/openms/source/IdXMLFile_test.cpp
```diff
@@ -289,6 +289,104 @@ START_SECTION(([EXTRA] XLMS data labeled cross-linker))
...
+
+  // We could now store the identification data in an idXML file
+  IdXMLFile().store(actual_file, protein_ids, peptide_ids);
+  // And load it back with
+  IdXMLFile().load(actual_file, protein_ids, peptide_ids);
+
+  TEST_EQUAL(peptide_ids.size(), 1)
+  TEST_EQUAL(peptide_ids[0].getHits().size(), 2)
+  TEST_EQUAL(peptide_ids[0].getHits()[0].getRank(), 1)
+  TEST_EQUAL(peptide_ids[0].getHits()[1].getRank(), 2)
```
hroest: NOTE: this will fail since the information is lost upon reading the file

timosachsenberg: Thanks for uncovering this. Do we need the rank concept at all? Either the rank is determined by the position in the PeptideHits vector or its just another "score" and could be stored in a meta value.

hroest: yes, I agree its not necessary since we can compute it from the score and the is_higher_score_better

timosachsenberg:
@hendrikweisser how do you handle ranks in the new datastructure?
Would you preserve a rank meta value between storing and loading or make the rank always based on the current scores of a PSM?

jpfeuffer:
Can you add that to the dozens of questions in the project? https://github.com/OpenMS/OpenMS/projects/29

hendrikweisser:
@timosachsenberg: Ranks aren't stored. Since multiple scores per ID are supported, the ranking might depend on which score you're considering. So better to just compute it when you need it.

hendrikweisser:
@jpfeuffer: Is there a way to reply to a "card" on the project page?

jpfeuffer:
Good question. Unfortunately, I don't think so. Either you edit the cards or "convert to issue" I think.
I would be interested in if something was already considered and if people think it is worth adding while "we are at it".
Maybe for the bigger cards with actual "planning/discussion" an issue would make sense.

timosachsenberg:
So the current design is that ranks are not stored but only available if assign ranks was called.
As pointed out above this is likely a bad design choice and fixed in the new data structure.
I think we can close this issue and add a comment that ranks are not preserved during storing

PR #3621, opened 2018-07-13
OpenPepXL Cross-Linking MS search tool update

enetz:
Added precursor monoisotopic peak correction.
Added additional fragment types to TheoreticalSpectrumGeneratorXLMS (TSG-XLMS), mainly neutral losses of H2O and NH3.
Exposed some TSG-XLMS parameters through the tools (e.g. for people using ETD or EThcD fragmentation).
Improved scoring functions and alignment algorithm.
XL-MS output data in mzid 1.2 format now validates with the newest mzid validator (although unfortunately some peak annotation types that we write to idXML I could not squeeze into the mzid peak annotation format yet).
OpenPepXL, OpenPepXLLF and XFDR now work properly together with each other and also with IDFileConverter and TextExporter.
Removed static column for Protein XL position from TOPPView ID table, now treated as normal UserParam as advanced annotation.
TOPPView does not show 2 rows for each XL-MS hit anymore, all necessary info is shown in one row (some stuff for the second peptide only in advanced annotations).

Improvements in efficiency, especially memory usage is reduced drastically.

Added an optional fast pre-scoring algorithm for quicker, rougher scans of big datasets (no optimality guarantees, but usually finds the most obvious stuff if anything is there to be found).

Will work on documentation and python bindings in the next few days.


src/openms/include/OpenMS/CHEMISTRY/TheoreticalSpectrumGeneratorXLMS.h
```diff
@@ -57,6 +58,23 @@ namespace OpenMS
...
+
+        bool operator==(const LossMass & other) const
+        {
+          return name == other.name && mass == other.mass;
+        }
+
+        // even if there are multiple losses with the same mass, having every mass once is enough
+        bool operator<(const LossMass & other) const
+        {
+          return mass < other.mass;
```
timosachsenberg: you need to also compare the name to get a proper less operator
std::tie(mass, name) < std::tie(other.mass, other.name)

src/openms/include/OpenMS/CHEMISTRY/TheoreticalSpectrumGeneratorXLMS.h
```diff
@@ -166,14 +193,68 @@ namespace OpenMS
...
+       * @param peptide The fragmented peptide
+       * @param link_pos position of the linker on the fragmented peptide
+       * @param precursor_mass The mass of the whole cross-link candidate or the precursor mass of the experimental MS2 spectrum.
+       * @param frag_alpha True, if the fragmented peptide is the Alpha peptide. Used for ion-name annotation.
+       * @param charge The charge of the ion
+       */
+      virtual void addKLinkedIonPeaks_(PeakSpectrum & spectrum, DataArrays::IntegerDataArray & charges, DataArrays::StringDataArray & ion_names, AASequence & peptide, Size link_pos, double precursor_mass, bool frag_alpha, int charge) const;
+
+
+      virtual void addComplexXLinkIonPeaks_(PeakSpectrum & spectrum, DataArrays::IntegerDataArray & charges, DataArrays::StringDataArray & ion_names, OPXLDataStructs::ProteinProteinCrossLink & crosslink, int charge, std::vector< std::set< LossMass > > & forward_losses_alpha, std::vector< std::set< LossMass > > & backward_losses_alpha, std::vector< std::set< LossMass > > & forward_losses_beta, std::vector< std::set< LossMass > > & backward_losses_beta) const;
```
timosachsenberg: add missing documentation (here and else but I guess this will come with some additional commits)

src/openms/include/OpenMS/CHEMISTRY/TheoreticalSpectrumGeneratorXLMS.h
```diff
@@ -166,14 +193,68 @@ namespace OpenMS
...
 
       // TODO copied from normal TSG, but it is protected over there. Move it to Residue class maybe?
       /// helper for mapping residue type to letter
       char residueTypeToIonLetter_(Residue::ResidueType res_type) const;
 
+
+
 //      /// TODO helper to add full neutral loss ladders
-//      //void addLosses_(RichPeakSpectrum & spectrum, const AASequence & ion, double intensity, Residue::ResidueType res_type, int charge, bool fragment_alpha_chain) const;
+      // void addLosses_(PeakSpectrum & spectrum, const AASequence & ion, double intensity, Residue::ResidueType res_type, int charge, bool fragment_alpha_chain) const;
```
timosachsenberg: remove commented out code?

src/openms/source/FORMAT/XQuestResultXMLFile.cpp
```diff
@@ -134,14 +134,14 @@ namespace OpenMS
         spec_xml_file << "</spectrum>" << std::endl;
 
         // the preprocessed pair spectra are sorted by another index
-        // because some pairs do not yield any reasonable hits, the index from the spectrum matches or spectrum_indices does not address the right pair anymore
-        // use find with the pair of spectrum indices to find the correct index for the preprocessed common and cross-linked ion spectra
+        // because some pairs do not yield any resonable hits, the index from the spectrum matches or spectrum_indices does not address the right pair anymore
```
timosachsenberg: reasonable?

src/openms_gui/source/VISUAL/SpectraIdentificationViewWidget.cpp
```diff
@@ -446,15 +446,15 @@ namespace OpenMS
         // ppm error
         addTextItemToBottomRow_("-", 15, c);
 
-        // cross-link position in Protein
-        addTextItemToBottomRow_("-", 16, c);
+        // // cross-link position in Protein
```
timosachsenberg: please remove commented out code

src/utils/OpenPepXLLF.cpp
```diff
@@ -189,6 +188,8 @@ class TOPPOpenPepXLLF :
 
     registerIntOption_("precursor:min_charge", "<num>", 3, "Minimum precursor charge to be considered.", false, true);
     registerIntOption_("precursor:max_charge", "<num>", 7, "Maximum precursor charge to be considered.", false, true);
+    registerIntList_("precursor:corrections", "<num>", ListUtils::create<int>("-2, -1, 0"), "Monoisotopic peak correction. Looks for possible monoisotopic precurser peaks at given number * (C13-C12) difference to the monoisotopic peak assumed by the instrument. These should be ordered from more extreme to less extreme corrections. Numbers later in the list will be preferred in case of ambiguities.", false, false);
```
timosachsenberg: can you check if MSGFPlus / Comet / SimpleSearchEngine uses the same numbering scheme? e.g. if -1 or +1 is used?

src/utils/OpenPepXLLF.cpp
```diff
@@ -558,7 +551,7 @@ class TOPPOpenPepXLLF :
 #endif
       {
         spectrum_counter++;
-        cout << "Processing spectrum " << spectrum_counter << " / " << spectra.size() << "\tSpectrum ID: " << spectrum.getNativeID()  << endl;
+        cout << "Processing spectrum " << spectrum_counter << " / " << spectra.size() << "\tSpectrum ID: " << spectrum.getNativeID()  << "\t| at: " << DateTime::now().getTime() << endl;
```
timosachsenberg: replace cout with e.g. writeDebug_

src/utils/OpenPepXLLF.cpp
```diff
@@ -569,130 +562,134 @@ class TOPPOpenPepXLLF :
...
+      for (Size pc = 0; pc < precursor_correction_positions.size(); ++pc)
+      {
+        precursor_corrections.push_back(precursor_correction_steps[precursor_correction_positions[pc]]);
       }
 
+      vector <OPXLDataStructs::ProteinProteinCrossLink> cross_link_candidates = OPXLHelper::buildCandidates(candidates, precursor_corrections, precursor_correction_positions, filtered_peptide_masses, cross_link_residue1, cross_link_residue2, cross_link_mass, cross_link_mass_mono_link, spectrum_precursor_vector, allowed_error_vector, cross_link_name);
+
+      progresslogger.endProgress();
+
+      cout << "Size of enumerated candidates: " << double(cross_link_candidates.size()) * sizeof(OPXLDataStructs::ProteinProteinCrossLink) / 1024.0 / 1024.0 << " mb" << endl;
```
timosachsenberg: what about these 1024 / 1024?

src/utils/OpenPepXLLF.cpp
```diff
@@ -569,130 +562,134 @@ class TOPPOpenPepXLLF :
...
-
-        LOG_DEBUG << "Pre-scoring | score: " << pre_score << endl;
-
-
-        prescore_csms_spectrum.push_back(csm);
       }
 
-      sort(prescore_csms_spectrum.rbegin(), prescore_csms_spectrum.rend());
+      // TODO major memory eater, use a couple of vectors instead?
+      cout << "Size of prescored candidates: " << double(prescore_csms_spectrum.size()) * sizeof(OPXLDataStructs::CrossLinkSpectrumMatch) / 1024.0 / 1024.0 << " mb" << endl;
```
timosachsenberg: same here

src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp
```diff
@@ -137,7 +140,7 @@ namespace OpenMS
 
       if (deisotope_spectra)
       {
-        PeakSpectrum deisotoped = OPXLSpectrumProcessingAlgorithms::deisotopeAndSingleChargeMSSpectrum(exp[exp_index], 1, 7, fragment_mass_tolerance_xlinks, fragment_mass_tolerance_unit_ppm, false, 3, 10, false);
+        PeakSpectrum deisotoped = OPXLSpectrumProcessingAlgorithms::deisotopeAndSingleChargeMSSpectrum(exp[exp_index], 1, 7, 100, fragment_mass_tolerance_unit_ppm, false, 3, 10, false);
```
timosachsenberg: fix this to 100ppm?

src/openms/include/OpenMS/CHEMISTRY/TheoreticalSpectrumGeneratorXLMS.h
enetz: I agree in general, but I used this only to make a unique list of loss masses. What should I do if two losses have the same mass but different names? Apply both and have two theoretical peaks stacked at the same mass? But I just noticed that this is not really used currently. The losses are now restricted to H2O and NH3, so there won't be any equal masses anyway. I will remove this function for now.

src/openms/source/FORMAT/XQuestResultXMLFile.cpp
enetz: this just means that some spectra do not have any hits, that are worth reporting (I need to check the exact criteria, but I think for now I only exclude stuff without any aligned peaks) and are therefore skipped in the output. because of that, the index of the vector of results does not have a direct relation to the index of the spectrum or pair of spectra that the current result relates to.

src/utils/OpenPepXLLF.cpp
enetz: this is just a conversion from bytes to mb and I think this makes it easier to read and understand than   / 1048576    but now that you mentioned it, does it make sense to write this out at all? I used this to see how much memory I saved with some changes, but this will not be very useful for users since it is only one chunk of the used memory anyway, and not always the biggest one. I will make this and some other outputs to debug outputs.

src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp
enetz: oh right, I think I should change that. Do you think using the search tolerance for deisotoping as well is the best way to go? And I think I will also include a parameter into the tools to let the user decide, if they want to apply deisotoping. Currently I only deisotope the spectra if the tolerance is below 100 ppm, but that was just a sloppy way to separate the two types of data I have seen so far: > 0.1 Da tolerances from ion traps and < 30 ppm tol from orbitraps. Maybe I will leave that as an "auto" mode, but give the user the ability to decide as an advanced parameter.

src/openms/include/OpenMS/CHEMISTRY/TheoreticalSpectrumGeneratorXLMS.h
timosachsenberg: in this case it should not be an operator<.
Not providing a full comparison operator actually leads to very nasty bugs in some cases (you can imagine that some algorithms, also in the STL, only use < to determine if two elements are exactly equal)

src/openms/source/FORMAT/XQuestResultXMLFile.cpp
timosachsenberg: typoe "resonable"

timosachsenberg:
please also move the code in the tools to an algorithm class (like you did for PeptideIndexer). I know it's some work but we will need it anyway.

src/utils/OpenPepXLLF.cpp
enetz: @timosachsenberg the MSGF+ documentation says this:
for `-ti -1,2` ( = n in the range from -1 to +2) it makes the corrections
`abs(exp-calc-n*1.00335Da)<20ppm`
so they start off from the experimental mass and subtract the given numbers.
Comet seems to do the same. I will adapt my scheme to theirs then.

enetz:
There is a failed test due to a function giving nonsensical results on some operating systems or compilers. I will look into that today.
The rest is a pyOpenMS issue. On Friday I had the error message `SyntaxWarning: import * only allowed at module level` once and  `ImportError: No module named all_modules` for most of the tests. That seems to indicate imports within a class or function definition. I don't think I had any of those. I found some imports that were unnecessary or too broad and reduced them. The first error seems to have gone, but now all pyOpenMS tests fail with the second one.
@timosachsenberg @hroest any ideas on what could have gone wrong? I can not reproduce these on my own computer, compiles and runs fine on Ubuntu 14 with python 2.7.
I had to add (for now incomplete) python bindings for StringView and EnzymaticDigestion to be able to add bindings for some functions in my OPXLHelper. Was there maybe a good reason, why we did not have those before?

enetz:
The pyOpenMS tutorial tests on Travis fail with `SyntaxWarning: import * only allowed at module level`, because of several lines of `from pyopenms import *` within function definitions,
while all other pyOpenMS tests fail with `ImportError: No module named all_modules`
after 0 compiler errors.
The Jenkins docker container fails with `NotImplementedError: void has no matching python type` during compiling (although I am sure there have been a lot of functions with void return types before?)

I could not reproduce this on linux, and today I successfully built the pyOpenMS.sln solution with VS2015 on Windows. Tried removing the pxds that I changed an dadded but that didn't make any difference.

Does anyone else have an idea?

enetz:
rebuild jenkins

src/utils/OpenPepXL.cpp
```diff
@@ -423,14 +482,15 @@ class TOPPOpenPepXL :
     Int max_precursor_charge = getIntOption_("precursor:max_charge");
     double precursor_mass_tolerance = getDoubleOption_("precursor:mass_tolerance");
     bool precursor_mass_tolerance_unit_ppm = (getStringOption_("precursor:mass_tolerance_unit") == "ppm");
+    IntList precursor_correction_steps = getIntList_("precursor:corrections");
```
hroest: your main method is almost 1000 LOC, this is clearly too long. please consider extracting some functions

src/utils/OpenPepXL.cpp
```diff
@@ -270,17 +287,44 @@ class TOPPOpenPepXL :
...
+      DataArrays::FloatDataArray dummy_array;
+      DataArrays::IntegerDataArray dummy_charges;
+      OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentFastCharge(matched_fragments_without_shift, fragment_mass_tolerance, fragment_mass_tolerance_unit_ppm, spectrum_light, spectrum_heavy, dummy_charges, dummy_charges, dummy_array, 0.3);
       LOG_DEBUG << " heavy_light comparison, matching peaks without shift: " << matched_fragments_without_shift.size() << endl;
 
       // transform by m/z difference between unlabeled and labeled cross-link to make heavy and light comparable.
       PeakSpectrum xlink_peaks;
       PeakSpectrum::IntegerDataArray spectrum_heavy_charges;
-      if (spectrum_heavy.getIntegerDataArrays().size() > 0)
+      PeakSpectrum::IntegerDataArray spectrum_light_iso_peaks;
```
hroest: same here, almost 1000 LOC main method

src/utils/OpenPepXLLF.cpp
```diff
@@ -413,71 +439,61 @@ class TOPPOpenPepXLLF :
 
     vector<PeptideIdentification> peptide_ids;
```
hroest: your main method is almost 1000 LOC, this is clearly too long. please consider extracting some functions

timosachsenberg: @enetz I agree with Hannes. I think a good option would be to move most of the code in main to an algorithm class. This would already get rid of some boilerplate parameter code / reading and writing of input output files. From my side, this could also be done after a merge as checks are currently passing and these changes would further expand this already big PR.

enetz: There is also still some duplicated functionality between the two binaries. I will start by abstracting some more static functions that the tools can share into the OPXLHelper class. Then move the main functions in algorithm classes.
And I also agree that this should be a separate PR.

timosachsenberg: Sounds good. Just consider putting that code into dedicated classes and not everything into a Helper class. The reason is, that Helper classes are considered an anti-pattern under most circumstances as they don't define a clear class responsibility (rather a dump for methods ;) ).

enetz: In this case I would rather have a dump for methods, since this Helper class is basically just abstraction for reused functionality between two main classes. Howerver I will think about what could be usable outside of those main functions and put them in other classes.

PR #3638, opened 2018-07-25
[FIX] CsiFingerIdMzTAbWriter, SiriusMzTabWriter 

oliveralka:
Feature_id was not reported correctly in the CsiFingerIdMzTabWriter.

Use the number of candidates for Sirius Output.
Use the number of top hits for CSIFingerID Output.

Fix corresponding CSIFingerID-test.

### DIFF
--- a/src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
+++ b/src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
@@ -64,10 +64,13 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
     {
       CsvFile compounds(pathtocsicsv, '\t');
       const UInt rowcount = compounds.rowCount();
-    
+
       if (rowcount > 1)
       {
-        const UInt top_n_hits_cor = (top_n_hits >= rowcount) ? rowcount : top_n_hits;
+        // correction if the rowcount is smaller than the number of hits used as parameter
+        // rowcount-1 because the csv header will be skipped in the loop later on.
+        int header = 1;
+        const UInt top_n_hits_cor = (top_n_hits >= rowcount) ? rowcount-header : top_n_hits;
 
         // fill identification structure containing all candidate hits for a single spectrum
         CsiFingerIdMzTabWriter::CsiAdapterIdentification csi_id;
@@ -86,12 +89,11 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
         boost::regex regexp_feature("_(?<SCAN>\\d+)-");
         bool found = boost::regex_search(str, match, regexp_feature);
         if (found && match["SCAN"].matched) {feature_id = "id_" + match["SCAN"].str();}
-        // results from scan were not assigned to a feautre
-        if (feature_id == "id_0") {feature_id = "null";}
+        String unassigned = "null";
 
-        for (Size j = 1; j < top_n_hits_cor; ++j)
+        // j = 1 because of .csv file format (header)
+        for (Size j = 1; j <= top_n_hits_cor; ++j)
         {
-          
           StringList sl;
           compounds.getRow(j, sl);
           CsiFingerIdMzTabWriter::CsiAdapterHit csi_hit;
@@ -110,7 +112,15 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
 
         csi_id.scan_index = scan_index;
         csi_id.scan_number = scan_number;
-        csi_id.feature_id = feature_id;
+        // check if results were assigned to a feature
+        if (feature_id != "id_0")
+        {
+          csi_id.feature_id = feature_id;
+        }
+        else
+        {
+          csi_id.feature_id = unassigned;
+        }
         csi_result.identifications.push_back(csi_id);
 
         // write metadata to mzTab file
--- a/src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
+++ b/src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
@@ -56,7 +56,6 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
 
   for (std::vector<String>::const_iterator it = sirius_output_paths.begin(); it != sirius_output_paths.end(); ++it)
   {
-
     const std::string pathtocsicsv = *it + "/summary_csi_fingerid.csv";
 
     ifstream file(pathtocsicsv);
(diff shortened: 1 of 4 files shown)


oliveralka:
Failed test:
`projectroot.src.tests.class_tests.openms.File_test`
seems unrelated

oliveralka:
rebuild jenkins

src/utils/SiriusAdapter.cpp
```diff
@@ -199,7 +199,7 @@ class TOPPSiriusAdapter :
     bool no_mt_info = getFlag_("no_masstrace_info_isotope_pattern");
 
     // needed for counting
-    int top_n_hits = getIntOption_("top_n_hits"); 
+    int top_n_hits = getIntOption_("top_n_hits") + 1;
```
timosachsenberg: why +1?

src/utils/SiriusAdapter.cpp
```diff
@@ -209,7 +209,7 @@ class TOPPSiriusAdapter :
     const QString isotope = getStringOption_("isotope").toQString();
     const QString noise = QString::number(getIntOption_("noise"));
     const QString ppm_max = QString::number(getIntOption_("ppm_max"));
-    const QString candidates = QString::number(getIntOption_("candidates"));
+    const QString candidates = QString::number(getIntOption_("candidates")+1);
```
timosachsenberg: why +1

timosachsenberg: better no to use a string here but a Size type.
Conversion to string can be done if required in line 311

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -68,7 +67,8 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
     
       if (rowcount > 1)
       {
-        
+        const UInt top_n_hits_cor = (top_n_hits >= rowcount) ? rowcount : top_n_hits;
```
lukaszimmermann: Maybe use `std::min`. Comment what happens here. Variable name unclear. (cor?)

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -89,7 +89,6 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
         // results from scan were not assigned to a feautre
         if (feature_id == "id_0") {feature_id = "null";}
```
lukaszimmermann: would not change the value of `feature_id`. Perhaps only use the "null" value where `feature_id` is read, otherwise I'd use a new variable name.

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -89,7 +89,6 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
...
 
-        const UInt top_n_hits_cor = (top_n_hits > rowcount) ? rowcount : top_n_hits;
         for (Size j = 1; j < top_n_hits_cor; ++j)
```
lukaszimmermann: What about `j = 0`?

src/utils/SiriusAdapter.cpp
```diff
@@ -151,7 +151,7 @@ class TOPPSiriusAdapter :
     // internal sirius parameters
     registerStringOption_("profile", "<choice>", "qtof", "Specify the used analysis profile", false);
     setValidStrings_("profile", ListUtils::create<String>("qtof,orbitrap,fticr"));
-    registerIntOption_("candidates", "<num>", 5, "The number of candidates in the output.", false);
+    registerIntOption_("candidates", "<num>", 5, "The number of candidates in the sirius output.", false);
```
lukaszimmermann: sirius -> SIRIUS

src/utils/SiriusAdapter.cpp
```diff
@@ -161,7 +161,7 @@ class TOPPSiriusAdapter :
     registerStringOption_("elements", "<choice>", "CHNOP[5]S[8]Cl[1]", "The allowed elements. Write CHNOPSCl to allow the elements C, H, N, O, P, S and Cl. Add numbers in brackets to restrict the maximal allowed occurrence of these elements: CHNOP[5]S[8]Cl[1].", false);
     registerIntOption_("compound_timeout", "<num>", 10, "Time out in seconds per compound. To disable the timeout set the value to 0", false);
     registerIntOption_("tree_timeout", "<num>", 0, "Time out in seconds per fragmentation tree computation.", false);
-    registerIntOption_("top_n_hits", "<num>", 10, "The top_n_hit for each compound written to the output", false);
+    registerIntOption_("top_n_hits", "<num>", 10, "The top_n_hit for each compound written to the CSI:FingerID output", false);
```
lukaszimmermann: Maybe: The number of top hits for each compound written...

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
oliveralka: Right now, feature_id is processed from name of the .csv file.
If no feature_id was associated the regex will still pick up an id_0. 
The writer will then write a "null" in the feautre_id column. 
In the case a feature_id is present it should look like:
id_3456783223456734

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
oliveralka: +1

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -64,10 +64,13 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
     {
       CsvFile compounds(pathtocsicsv, '\t');
       const UInt rowcount = compounds.rowCount();
-    
+
       if (rowcount > 1)
       {
-        const UInt top_n_hits_cor = (top_n_hits >= rowcount) ? rowcount : top_n_hits;
+        // correction if the rowcount is smaller than the number of hits used as parameter
```
timosachsenberg: Minor suggestion as I found the comments/variable names could be a bit more descriptive.
you could define
``` 
const UInt n_header_rows = 1;
const UInt n_hits_rows = rowcount - n_header_rows;
and
const UInt n_hits_to_extract = std::min(top_n_hits, n_hits_rows);
```
instead of using the top_n_hits_cor variable name .
comment in line 92 would probably than not be necessary
and you could loop from j=0 to n_hits_to_extract but inside of the loop call .getRow(n_header_rows + j) instead.
All in all slightly clearer (at least to me) :P

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
timosachsenberg: I think Lukas means that it is confusing/hard to follow if the feature_id is overwritten.
I think what he means is to better set the id to e.g. "UNKNOWN_ID" (ideally a constant defined as static constexpr) instead of id_0.
When writing to mzTab you would check if it is  == this value and if so write out "null".

timosachsenberg:
I will merge this PR as it contains a fix to a bug. As Oliver is on holidays, I would like to ask him to resolve the remaining minor comments in a new PR.

PR #3642, opened 2018-07-26
[FIX, API, TEST] mass calculations for isotope distributions

### LINKED ISSUE #3589: Average weight calculations
Dear developers

I have an issue where I dont quite understand how isotopic distributions and average weight are calculated in OpenMS. Specifically I am looking at BSA using pyOpenMS:

```
>>> from pyopenms import *
>>> import urllib
>>> urllib.urlretrieve ("http://www.uniprot.org/uniprot/P02769.fasta", "bsa.fasta")
>>>
>>> bsa = "".join([l.strip() for l in open("bsa.fasta").readlines()[1:]])
>>> bsa_prot = AASequence.fromString(bsa, True)
>>> bsa_processed = AASequence.fromString(bsa[24:], True)
>>>
>>> bsa_prot.getAverageWeight(Residue.ResidueType.Full, 0)
69292.88650621701
>>> bsa_processed.getAverageWeight(Residue.ResidueType.Full, 0)
66432.45593654823
>>>
>>> seq_formula = bsa_processed.getFormula(Residue.ResidueType.Full, 0)
>>> isotopes = seq_formula.getIsotopeDistribution(80)
>>> max( isotopes.getContainer(), key=lambda x : x[1] )
(66399, 0.05713559158303688)
>>> for iso in isotopes.getContainer()[-10:]:
...     print (iso)
...
(66427, 7.452259391000588e-05)
(66428, 4.8067464291277997e-05)
(66429, 3.0645369519834025e-05)
(66430, 1.931540348099346e-05)
(66431, 1.2037621963010855e-05)
(66432, 7.419015362283155e-06)
(66433, 4.522610417503779e-06)
(66434, 2.7273207269190423e-06)
(66435, 1.627242728505254e-06)
(66436, 9.607287237740715e-07)
```

so what I dont understand are two things: first, why the average weight iof BSA is 66432.45593654823 which is close to the literature value of 66430.3 [1] but not exactly the same as the literature value. Secondly, and  more strangely, the maximum of the isotopic distribution is at 66399 with 5.7% while the average value has 1.9e-5 probability. I expected the maximum of the isotopic distribution to be close to the average weight.

1. https://www.ncbi.nlm.nih.gov/pubmed/2260975

DennisGoldfarb:
fixes #3589

CoarseIsotopePatternGenerator used to output atomic numbers (# protons + # neutrons) for the mass of each isotope. This can be quite different from the rounded mass of an isotope (see #3589)

I added a calc_mass_ flag that can be set during initialization of the CoarseIsotopePatternGenerator or afterwards. It determines if the expected masses (new functionality) or atomic numbers (old functionality) should be returned. By default, the calc_mass_ is set to true. The previous behavior can be achieved by setting calc_mass_ to false.

Expected masses are the monoisotopic mass of the molecule + (C13_C12_MASS_DIFFERENCE * isotope number). I'm assuming that most of a coarse isotopic peak is due to C13 molecules.

Currently, OpenMS has the following capabilities:

- With a known EmpiricalFormula
  - Precursor isotope distribution
    - Compute fine isotope probabilities and mass
    - Compute coarse isotope probabilities and atomic numbers for the mass
    - Compute coarse isotope probabilities and expected mass
  - Fragment isotope distribution
    - Compute coarse isotope probabilities and atomic numbers for the mass
    - Compute coarse isotope probabilities and expected mass
- With unknown EmpiricalFormula (using observed masses instead)
  - Precursor isotope distribution
    - Compute fine isotope probabilities and mass
    - Compute coarse isotope probabilities and atomic numbers for the mass
    - Compute coarse isotope probabilities and expected mass
  - Fragment isotope distribution
    - Compute coarse isotope probabilities and atomic numbers for the mass
    - Compute coarse isotope probabilities and expected mass

These changes only broke a handful of unit tests (because default is now to return expected mass) and they have been updated. I looked for tools that used the returned masses but it seems like they re-compute the masses themselves. I need to look more thoroughly.

Examples:
https://github.com/OpenMS/OpenMS/blob/5a70018d9e03ce32e64fcbb1c985b7a1256efc7a/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderAlgorithmPicked.cpp#L487-L491 

https://github.com/OpenMS/OpenMS/blob/5a70018d9e03ce32e64fcbb1c985b7a1256efc7a/src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp#L115-L122

Note that for the FeatureFinder, it doesn't use C13_C12_MASS_DIFFERENCE. I'm going to try changing this later to see if the number features increases/decreases. I haven't gone through the openSWATH code yet.

I can also add functionality to return the rounded expected masses. I would probably replace the bool calc_mass_ variable with an enum {ROUND, EXPECTED_MASS, ATOMIC_NUMBER}.

Thoughts and advice are welcome!

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
@@ -75,6 +75,8 @@ namespace OpenMS
 
     CoarseIsotopePatternGenerator(const Size& max_isotope);
 
+    CoarseIsotopePatternGenerator(const Size& max_isotope, const bool calc_mass);
+
     virtual ~CoarseIsotopePatternGenerator();
 
     /// @name Accessors
@@ -87,8 +89,14 @@ namespace OpenMS
     */
     void setMaxIsotope(const Size& max_isotope);
 
+    /// sets the calc_mass_ flag to calculate and return expected masses (true) or atomic numbers (false).
+    void setCalcMass(const bool calc_mass);
+
     /// returns the currently set maximum isotope
     Size getMaxIsotope() const;
+
+    /// returns the current value of the flag to return expected masses (true) or atomic numbers (false).
+    bool getCalcMass() const;
     
     Size getMin() const;
     Size getMax() const;
@@ -269,8 +277,9 @@ namespace OpenMS
        @param fragment_isotope_dist the isotopic distribution of the fragment (as if it was a precursor).
        @param comp_fragment_isotope_dist the isotopic distribution of the complementary fragment (as if it was a precursor).
        @param precursor_isotopes a list of which precursor isotopes were isolated. 0 corresponds to the mono-isotopic molecule (M0), 1->M1, etc.
+       @param fragment_mono_mass the monoisotopic mass of the fragment. It is only used to compute masses if getCalcMass() is true.
     */
-    IsotopeDistribution calcFragmentIsotopeDist(const IsotopeDistribution& fragment_isotope_dist, const IsotopeDistribution& comp_fragment_isotope_dist, const std::set<UInt>& precursor_isotopes);
+    IsotopeDistribution calcFragmentIsotopeDist(const IsotopeDistribution& fragment_isotope_dist, const IsotopeDistribution& comp_fragment_isotope_dist, const std::set<UInt>& precursor_isotopes, const double fragment_mono_mass) const;
 
     CoarseIsotopePatternGenerator& operator=(const CoarseIsotopePatternGenerator& iso);
 
@@ -283,7 +292,10 @@ namespace OpenMS
     /// convolves the distribution @p input with itself and stores the result in @p result
     IsotopeDistribution::ContainerType convolveSquare_(const IsotopeDistribution::ContainerType & input) const;
 
-protected:
+    /// converts the masses of distribution @p input from atomic numbers to atomic masses
+    IsotopeDistribution::ContainerType correctMass_(const IsotopeDistribution::ContainerType & input, const double mono_weight) const;
+
+  protected:
 
     /** @brief calculates the fragment distribution for a fragment molecule and stores it in @p result.
 
@@ -291,7 +303,7 @@ protected:
         @param comp_fragment_isotope_dist the isotopic distribution of the complementary fragment (as if it was a precursor).
         @param precursor_isotopes which precursor isotopes were isolated. 0 corresponds to the mono-isotopic molecule (M0), 1->M1, etc.
     */
-    IsotopeDistribution calcFragmentIsotopeDist_(const IsotopeDistribution::ContainerType& fragment_isotope_dist, const IsotopeDistribution::ContainerType& comp_fragment_isotope_dist, const std::set<UInt>& precursor_isotopes);
+    IsotopeDistribution calcFragmentIsotopeDist_(const IsotopeDistribution::ContainerType& fragment_isotope_dist, const IsotopeDistribution::ContainerType& comp_fragment_isotope_dist, const std::set<UInt>& precursor_isotopes) const;
 
     /// fill a gapped isotope pattern (i.e. certain masses are missing), with zero probability masses
     IsotopeDistribution::ContainerType fillGaps_(const IsotopeDistribution::ContainerType& id) const;
@@ -299,6 +311,8 @@ protected:
  protected:
     /// maximal isotopes which is used to calculate the distribution
     Size max_isotope_;
+    /// flag to determine whether expected masses or atomic numbers are returned
+    bool calc_mass_;
 
   };
 
--- a/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
@@ -53,6 +53,12 @@ namespace OpenMS
     *         For example for (13)Carbon it assumes that the mass of the isotope is 13Da
     *         instead of 13.0033548378
     *
+    *         The output is a list of nominal isotope probabilities
+    *         with either accurate or rounded masses. The accurate masses assume the
+    *         nominal isotopes are mostly due to (13)Carbon.
+    *         To return accurate vs rounded masses, use setRoundMasses accordingly.
+    *         The default is to return accurate masses.
+    *
     *         The most important value which should be set is the max isotope value.
     *         This value can be set using the setMaxIsotope method. It is an upper
     *         bound for the number of isotopes which are calculated
@@ -62,8 +68,6 @@ namespace OpenMS
     *         number of values, if the mass value is large!
     *
     *         See also method run()
-    *
-    *         Distributions can be added to one another or scaled by an integer.
     **/
 
   class OPENMS_DLLAPI CoarseIsotopePatternGenerator 
@@ -75,7 +79,7 @@ namespace OpenMS
 
     CoarseIsotopePatternGenerator(const Size& max_isotope);
 
-    CoarseIsotopePatternGenerator(const Size& max_isotope, const bool calc_mass);
+    CoarseIsotopePatternGenerator(const Size& max_isotope, const bool round_masses);
 
     virtual ~CoarseIsotopePatternGenerator();
 
@@ -89,14 +93,14 @@ namespace OpenMS
     */
     void setMaxIsotope(const Size& max_isotope);
 
-    /// sets the calc_mass_ flag to calculate and return expected masses (true) or atomic numbers (false).
-    void setCalcMass(const bool calc_mass);
+    /// sets the round_masses_ flag to round masses to integer values (true) or return accurate masses (false)
+    void setRoundMasses(const bool round_masses);
 
     /// returns the currently set maximum isotope
     Size getMaxIsotope() const;
 
     /// returns the current value of the flag to return expected masses (true) or atomic numbers (false).
-    bool getCalcMass() const;
+    bool getRoundMasses() const;
     
     Size getMin() const;
     Size getMax() const;
@@ -277,7 +281,8 @@ namespace OpenMS
        @param fragment_isotope_dist the isotopic distribution of the fragment (as if it was a precursor).
        @param comp_fragment_isotope_dist the isotopic distribution of the complementary fragment (as if it was a precursor).
        @param precursor_isotopes a list of which precursor isotopes were isolated. 0 corresponds to the mono-isotopic molecule (M0), 1->M1, etc.
-       @param fragment_mono_mass the monoisotopic mass of the fragment. It is only used to compute masses if getCalcMass() is true.
+       @param fragment_mono_mass the monoisotopic mass of the fragment.
+       @pre fragment_isotope_dist and comp_fragment_isotope_dist are gapless (no missing isotopes between the min/max isotopes of the dist)
     */
     IsotopeDistribution calcFragmentIsotopeDist(const IsotopeDistribution& fragment_isotope_dist, const IsotopeDistribution& comp_fragment_isotope_dist, const std::set<UInt>& precursor_isotopes, const double fragment_mono_mass) const;
 
@@ -292,7 +297,7 @@ namespace OpenMS
     /// convolves the distribution @p input with itself and stores the result in @p result
     IsotopeDistribution::ContainerType convolveSquare_(const IsotopeDistribution::ContainerType & input) const;
 
-    /// converts the masses of distribution @p input from atomic numbers to atomic masses
+    /// converts the masses of distribution @p input from atomic numbers to accurate masses
     IsotopeDistribution::ContainerType correctMass_(const IsotopeDistribution::ContainerType & input, const double mono_weight) const;
 
   protected:
@@ -311,8 +316,8 @@ namespace OpenMS
  protected:
     /// maximal isotopes which is used to calculate the distribution
     Size max_isotope_;
-    /// flag to determine whether expected masses or atomic numbers are returned
-    bool calc_mass_;
+    /// flag to determine whether masses should be rounded or not
+    bool round_masses_;
 
   };
 
--- a/src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
+++ b/src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
@@ -205,8 +205,7 @@ namespace OpenMS
     IsotopeDistribution fragment_isotope_dist = getIsotopeDistribution(CoarseIsotopePatternGenerator(max_depth));
     IsotopeDistribution comp_fragment_isotope_dist = complementary_fragment.getIsotopeDistribution(CoarseIsotopePatternGenerator(max_depth));
 
-    IsotopeDistribution result;
-    result = solver.calcFragmentIsotopeDist(fragment_isotope_dist, comp_fragment_isotope_dist, precursor_isotopes, getMonoWeight());
+    IsotopeDistribution result = solver.calcFragmentIsotopeDist(fragment_isotope_dist, comp_fragment_isotope_dist, precursor_isotopes, getMonoWeight());
 
     // Renormalize to make these conditional probabilities (conditioned on the isolated precursor isotopes)
     result.renormalize();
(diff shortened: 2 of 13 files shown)


hroest:
> I can also add functionality to return the rounded expected masses. I would probably replace the bool calc_mass_ variable with an enum {ROUND, EXPECTED_MASS, ATOMIC_NUMBER}.

I agree, however I dont see any point in having the atomic number (except maybe backwards compatibility) because it is simply *wrong* -- I would rather argue that if a user wants it they can have an integer back that starts at 0, basically

```
0 : 0.03
1 : 0.16
2 : 0.5
3 : 0.2
...
```

where its then the users job to figure out what to do with these numbers in case they dont want any molecular weight. I like the idea with ROUND as for low mass compounds it should return the same result as we currently have and it should still be accurate for high mass compounds (accurate within 1 Da that is).

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
```diff
@@ -87,8 +89,14 @@ namespace OpenMS
     */
     void setMaxIsotope(const Size& max_isotope);
 
+    /// sets the calc_mass_ flag to calculate and return expected masses (true) or atomic numbers (false).
+    void setCalcMass(const bool calc_mass);
```
hroest: I am unhappy about the term "expected masses", maybe use "accurate monoisotopic mass"

src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
```diff
@@ -206,8 +206,7 @@ namespace OpenMS
     IsotopeDistribution comp_fragment_isotope_dist = complementary_fragment.getIsotopeDistribution(CoarseIsotopePatternGenerator(max_depth));
 
     IsotopeDistribution result;
-    CoarseIsotopePatternGenerator solver;
-    result = solver.calcFragmentIsotopeDist(fragment_isotope_dist, comp_fragment_isotope_dist, precursor_isotopes);
```
hroest: is this simply to improve performance and not create a new object each time?

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
hroest: maybe even explain the two options in the class description and explain the default behavior (true)

DennisGoldfarb:
Looks like I got some pyopenms tests to fix

DennisGoldfarb:
@hroest I agree that rounded makes more sense than the atomic numbers. Up to if you want to break the backwards compatibility. I expect that it wouldn't affect most people because the molecule needs to be large to cause an issue. Do you have a preference? It would be less work for me to have the only two options be rounded vs accurate masses

src/tests/class_tests/openms/source/EmpiricalFormula_test.cpp
```diff
@@ -366,56 +366,87 @@ END_SECTION
...
+
+  // Formula of Bovine Serum Albumin (BSA) that is has been processed and is missing the first 24 AA
+  EmpiricalFormula ef2("C2934H4615N781O897S39");
+  IsotopeDistribution iso3 = ef2.getIsotopeDistribution(CoarseIsotopePatternGenerator(20));
+  // monoisotopic mass
+  TEST_REAL_SIMILAR(iso3.begin()->getMZ(), 66389.863)
+
+  IsotopeDistribution iso4 = ef2.getIsotopeDistribution(CoarseIsotopePatternGenerator(20, false));
+  // due to rounding, the atomic number should be smaller than the monoisotopic mass of BSA
+  TEST_REAL_SIMILAR(iso4.begin()->getMZ(), 66357)
```
hroest: it is really a question whether we want to keep computing these wrong values

hroest:
> Up to if you want to break the backwards compatibility. I expect that it wouldn't affect most people because the molecule needs to be large to cause an issue.

I think for most small molecules the rounded is equal to the atomic number, so there we should be ok. For large molecules, it will be different but it will wrong and I dont see a good reason to output wrong numbers.

DennisGoldfarb:
Yea, I imagine everyone who used it before was expecting rounded masses anyway, not atomic numbers. I'll switch it to only those two options

src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
DennisGoldfarb: The solver is now a parameter to the function and therefore doesn't need to be created inside the function. This way you can choose if you want rounded masses or not.

hroest:
@DennisGoldfarb please merge https://github.com/DennisGoldfarb/OpenMS/pull/1

hroest:
it looks like now the MSSim test failes: http://cdash.openms.de/testDetails.php?test=51009106&build=279507 @cbielow do you have some insight into why that may be the case?

```


checking (SimTypes::MSSimExperiment const& getExperiment() const ) ... 
 -  line 251:  TEST_EQUAL(mssim.getExperiment().getNrSpectra(),230): got 233, expected 230
 +  line 252:  TEST_EQUAL(nr_ms1,127): got 127, expected 127
 -  line 253:  TEST_EQUAL(nr_ms2,103): got 106, expected 103
 +  line 255:  TEST_EQUAL(nr_ms2 + nr_ms1,mssim.getExperiment().getNrSpectra()): got 233, expected 233
 +  line 267:  TEST_EQUAL(no_sim.getExperiment().getSize(),empty_experiment.getSize()): got 0, expected 0
: failed
```

src/tests/class_tests/openms/source/EmpiricalFormula_test.cpp
```diff
@@ -366,56 +366,87 @@ END_SECTION
...
+    TEST_REAL_SIMILAR(it->getIntensity(), result_prob[i])
+    // atomic numbers should have been calculated
+    TEST_REAL_SIMILAR(it->getMZ(), result_rounded_mass[i])
+  }
+
+  // Formula of Bovine Serum Albumin (BSA) that is has been processed and is missing the first 24 AA
+  EmpiricalFormula ef2("C2934H4615N781O897S39");
+  IsotopeDistribution iso3 = ef2.getIsotopeDistribution(CoarseIsotopePatternGenerator(20));
+  // monoisotopic mass
+  TEST_REAL_SIMILAR(iso3.begin()->getMZ(), 66389.863)
```
hroest: @DennisGoldfarb once you are here, why not also test the probability? Lets also test one of the more interesting ones, eg 20 or 30 isotopes in to see whether we get the correct result

src/tests/class_tests/openms/source/EmpiricalFormula_test.cpp
```diff
@@ -387,15 +387,30 @@ START_SECTION(IsotopeDistribution getIsotopeDistribution(UInt max_depth) const)
...
+
+  IsotopeDistribution iso4 = ef2.getIsotopeDistribution(CoarseIsotopePatternGenerator(100, true));
   // rounded monoisotopic mass
-  TEST_REAL_SIMILAR(iso4.begin()->getMZ(), 66390)
+  TEST_EQUAL(iso4.begin()->getMZ(), 66390)
+  TEST_REAL_SIMILAR(iso4.averageMass(), ef2.getAverageWeight())
+
+  most_abundant = iso4.getMostAbundant();
+  TEST_EQUAL(most_abundant.getMZ(), 66432);
+  TEST_REAL_SIMILAR(most_abundant.getIntensity(), 0.057);
```
hroest: great, this is the result we want to see!

src/tests/class_tests/openms/source/EmpiricalFormula_test.cpp
```diff
@@ -387,15 +387,30 @@ START_SECTION(IsotopeDistribution getIsotopeDistribution(UInt max_depth) const)
...
-  IsotopeDistribution iso3 = ef2.getIsotopeDistribution(CoarseIsotopePatternGenerator(20));
+  IsotopeDistribution iso3 = ef2.getIsotopeDistribution(CoarseIsotopePatternGenerator(100));
   // monoisotopic mass
   TEST_REAL_SIMILAR(iso3.begin()->getMZ(), 66389.863)
-
-  IsotopeDistribution iso4 = ef2.getIsotopeDistribution(CoarseIsotopePatternGenerator(20, true));
+  // the average mass of the IsotopeDistribution is computed from its masses and probabilities
+  // and should be extremely close to the result from the EmpiricalFormula which uses the
+  // average weights of each atom.
+  TEST_REAL_SIMILAR(iso3.averageMass(), ef2.getAverageWeight())
```
hroest: nice!

src/tests/class_tests/openms/source/EmpiricalFormula_test.cpp
```diff
@@ -387,15 +387,30 @@ START_SECTION(IsotopeDistribution getIsotopeDistribution(UInt max_depth) const)
...
+
+  Peak1D most_abundant = iso3.getMostAbundant();
+  // Partially a regression test. For such a large molecule, consecutive isotopes have very similar probabilities
+  // and a small change could change the result by 1Da (e.g. numerical error, elemental isotope probabilities)
+  TEST_REAL_SIMILAR(most_abundant.getMZ(), 66432.0037);
```
hroest: why isnt this `66389.863 + C12_C13 * k` for an integer value of `k` ? Did I misinterpret something here?

src/tests/class_tests/openms/source/IsotopeDistribution_test.cpp
```diff
@@ -170,6 +170,18 @@ START_SECTION(Size getMin() const)
...
+START_SECTION(Size getMostAbundant() const)
+	IsotopeDistribution iso(EmpiricalFormula("C1").getIsotopeDistribution(CoarseIsotopePatternGenerator(11, true)));
+    // The most abundant isotope is the monoisotope
+	TEST_EQUAL(iso.getMostAbundant().getMZ(), 12)
+	IsotopeDistribution iso2(EmpiricalFormula("C100").getIsotopeDistribution(CoarseIsotopePatternGenerator(11, true)));
+    // In this case, the most abundant isotope isn't the monoisotope
+	TEST_EQUAL(iso2.getMostAbundant().getMZ(), 1201)
+	IsotopeDistribution iso3;
+    // Making sure an empty distribution doesn't crash it.
+    TEST_EQUAL(iso3.getMostAbundant().getMZ(), 0);
```
hroest: this *should* not work, `std::max_element` will return `last()` on which you cannot call `getMZ()`

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsotopeDistribution.cpp
```diff
@@ -108,6 +108,10 @@ namespace OpenMS
     return distribution_[0].getMZ();
   }
 
+  Peak1D IsotopeDistribution::getMostAbundant() const
+  {
+      return *std::max_element(begin(), end(), MassAbundance::IntensityLess());
```
hroest: you may have to check here that its not empty, otherwise you return `last()` which will lead to segfaults and issues in the Python bindings, see https://en.cppreference.com/w/cpp/algorithm/max_element

> Iterator to the greatest element in the range [first, last). If several elements in the range are equivalent to the greatest element, returns the iterator to the first such element. Returns last if the range is empty.

src/tests/class_tests/openms/source/IsotopeDistribution_test.cpp
DennisGoldfarb: Ah good point. Any thoughts on what I should return if it's empty? Peak1D of 0mz and 1 probability? 0 and 0?

hroest: either that or throw an exception that you cannot get a max peak on an empty distribution - which will be better than a segfault down the line :-)

src/tests/class_tests/openms/source/EmpiricalFormula_test.cpp
DennisGoldfarb: Your got the equation right: C12_C13 = 1.00335 and in this case k = 42, so 66389.863 + (1.00335 * 42) = 66389.0037.

Related to this, I see that OpenSWATH DIAHelper uses a 1.00048 distance between isotopes. https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h#L63-L67. 

I'm comparing that with using 1.00335 right now. I've never used the tool before but I imagine some of those isotope fields in the output might improve. When I switched FeatureFinderMultiplex from using 1.00335 to 1.0, the number of features went from ~27k to ~25k

DennisGoldfarb: No effect on OpenSWATH, but I'm not going to dig into why.

hroest:
can look into the following failing test: 

The following tests FAILED:
	531 - MSSim_test (Failed)
Submit files (using http)

src/tests/class_tests/openms/source/IsotopeDistribution_test.cpp
hroest: > I see now that the constructor initializes the distribution to have a single entry of 0,1. No wonder it didn't crash.

what happens if someone initializes it on purpose with 0 entries?

DennisGoldfarb:
Seems to be related to the version of boost. It passes on my build which uses 1.64:

https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/MSSim_test.cpp#L250-L259

src/tests/class_tests/openms/source/IsotopeDistribution_test.cpp
DennisGoldfarb: I added a check for that. If the distribution is empty, it will return 0,1

hroest:
> Seems to be related to the version of boost. It passes on my build which uses 1.64:

the issue is that it passed before your PR

DennisGoldfarb:
I've got a boost 1.55 build working now. Should be able to pinpoint the reason soon

hroest:
> I've got a boost 1.55 build working now. Should be able to pinpoint the reason soon

this is a bit strange, I am more worried about why a change here in this part of the code will affect a completely different part of the Simulation code. Do you use a RNG somewhere or does the MSSim code somehow generated different random numbers depending on the isotopic envelopes it produces? Maybe @cbielow can help here?

src/tests/class_tests/openms/source/EmpiricalFormula_test.cpp
hroest: > No effect on OpenSWATH, but I'm not going to dig into why.

ok, that is fine :-)

DennisGoldfarb:
It's difficult to pin down precisely the situation that causes the problem. But here's a high-level explanation. Different retention times are predicted depending on the version of boost. Possibly due to a RNG. This leads to different features being simulated. Other predictions might vary too. The set of simulated features with boost 1.55 happens to lead to the following case:

The m/z of one of the isotopes falls outside an <isolation window, retention time> bounding box (though previously it was within the box), and maybe no longer gets added to the exclusion list. It therefore gets targeted later on for an MS2.

There are a few sections of code in the simulator that compute isotope m/z's. Sometimes it uses the masses from the IsotopicDistribution, sometimes it's computed manually and assumes 1da intervals. This PR caused those approaches to give slightly different answers and changed the result.

Examples:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/SIMULATION/RawMSSignalSimulation.cpp#L655-L665

https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/SIMULATION/RawMSSignalSimulation.cpp#L730-L742

https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/SIMULATION/RawMSSignalSimulation.cpp#L712-L720

To keep things consistent with the previous unit tests, I set IsotopeModel to use rounded masses. This way, the sections of code that used the IsotopeDistribution masses will give the same answers as before.

Whether the result is correct could be the subject of another issue though. At some point I'd guess that IsotopeModel needs to be refactored to support any IsotopeDistribution solver (coarse round, coarse exact, fine)

timosachsenberg:
@DennisGoldfarb wow thanks a lot for taking on that issue!
@hroest good to merge from your side?

hroest:
wow, thanks so much for all your work, this is really great

> To keep things consistent with the previous unit tests, I set IsotopeModel to use rounded masses. This way, the sections of code that used the IsotopeDistribution masses will give the same answers as before.

if that is the case, I am also fine with using accurate masses in `IsotopeModel` and changing the results of the unit test. However, I think before recent changes all code expected rounded, integer masses so maybe returning rounded masses is the safe bet for now until we understand the impact. But it is clear that we need to look at this again, also with the high-res isotope pattern generator that hopefully comes soon

PR #3644, opened 2018-07-27
fix peptide detection in FeatureFinderMultiplex

lars20070:
In `FeatureFinderMultiplex` filter results are stored as peak indices in the original MS experiment. Previously, an already intensity-filtered MS experiment was used instead of the complete, original experiment. This resulted in a mix-up of indices and consequently incorrect features. The problem remained unobserved, since in all tests a very low intensity cutoff was used.

The previous intensity filtering of the complete MS experiment is replaced by the blacklisting of low-intensity peaks. Now the reference experiment is the complete, original MS experiment, and low-intensity peaks are merely blacklisted right from the start.

An additional TOPP test with a high intensity-cutoff was added. Two peptide pairs are in the `mzML` input, but only one is detected.


src/openms/source/TRANSFORMATIONS/FEATUREFINDER/MultiplexFiltering.cpp
```diff
@@ -55,15 +55,15 @@ namespace OpenMS
...
+  exp_picked_(exp_picked), patterns_(patterns), isotopes_per_peptide_min_(isotopes_per_peptide_min), isotopes_per_peptide_max_(isotopes_per_peptide_max),
   intensity_cutoff_(intensity_cutoff), rt_band_(rt_band), mz_tolerance_(mz_tolerance), mz_tolerance_unit_in_ppm_(mz_tolerance_unit),
   peptide_similarity_(peptide_similarity), averagine_similarity_(averagine_similarity),
   averagine_similarity_scaling_(averagine_similarity_scaling), averagine_type_(averagine_type)
   {
     // initialise experiment exp_picked_
     // Any peaks below the intensity cutoff cannot be relevant and are therefore removed.
     // loop over spectra
-    for (MSExperiment::ConstIterator it_rt = exp_picked.begin(); it_rt < exp_picked.end(); ++it_rt)
+    /*for (MSExperiment::ConstIterator it_rt = exp_picked.begin(); it_rt < exp_picked.end(); ++it_rt)
```
timosachsenberg: what about this code?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/MultiplexFiltering.cpp
```diff
@@ -96,11 +97,27 @@ namespace OpenMS
...
+      }
+    }
+    
   }
 
-  MSExperiment MultiplexFiltering::getWhiteMSExperiment_(White2Original& mapping)
+  void MultiplexFiltering::updateWhiteMSExperiment_()
   {
-    MSExperiment exp_picked_white;
+    // reset both the white MS experiment and the corresponding mapping to the complete i.e. origibal MS experiment
```
timosachsenberg: typo original

PR #3645, opened 2018-07-30
Feature/float data array compression

hroest:
- add easy access to float data arrays from Python
- allow compression of float data arrays in API
- allow compression of float data arrays in FileFilter and FileConverter

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -3258,7 +3258,7 @@ namespace OpenMS
       return cvTerm;
     }
 
-    void MzMLHandler::writeUserParam_(std::ostream& os, const MetaInfoInterface& meta, UInt indent, String path, Internal::MzMLValidator& validator) const
+    void MzMLHandler::writeUserParam_(std::ostream& os, const MetaInfoInterface& meta, UInt indent, String path, const Internal::MzMLValidator& validator) const
     {
       std::vector<String> cvParams;
       std::vector<String> userParams;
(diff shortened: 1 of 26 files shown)


hroest:
I did a few measurements using different ways of copying data between C++ and Python:

For assignment

```
        cdef _FloatDataArray * fda_ = self.inst.get()
        cdef unsigned int n = fda_.size()
             
        # Obtain a raw ptr to the beginning of the C++ array
        cdef libcpp_vector[float] * vec_ptr = <libcpp_vector[float]*> fda_
        cdef float * raw_ptr =  address(deref(vec_ptr)[0]) 

        cdef float[:] fda_view = <float[:n]>raw_ptr # cast to memoryview, refer to the underlying buffer without copy
        xarr = np.asarray(fda_view) # numpy array refer to the underlying buffer without copy
        return xarr
```

vs:

```
        cdef _FloatDataArray * fda_ = self.inst.get()
        cdef libcpp_vector[float].iterator it = fda_.begin()
        cdef unsigned int n = fda_.size()
        cdef np.ndarray[np.float32_t, ndim=1] data
        data = np.zeros( (n,), dtype=np.float32)
        
        cdef int i = 0
        while it != fda_.end():
            data[i] = deref(it)
            inc(it)
            i += 1
        
        return data
```

Performance measurements:
```
        $ python -m timeit -r 3 -s 'import pyopenms, numpy; d = numpy.random.rand( 100000).astype(numpy.float32); f = pyopenms.FloatDataArray();f.set_data(d)' 'f.get_data()'
        
        Using the naive assignment method:
            10000 loops, best of 3: 71.5 usec per loop
        Using the memory view method:
            1000000 loops, best of 3: 1.52 usec per loop
        
        # cmp to using pure Python push_back
        $ python -m timeit -r 3 -s 'import pyopenms, numpy; d = numpy.random.rand( 100000).astype(numpy.float32); f = pyopenms.FloatDataArray();f.set_data(d)' 'q = [val for val in d]'
            100 loops, best of 3: 4.01 msec per loop
```
We see an improvement of ca 50x using the memory view method (and a 2600x fold improvement compared to pure Python).

hroest:
Similarly for setting data:

```
        cdef _FloatDataArray * fda_ = self.inst.get()
        fda_.clear()

        cdef float * array_start = <float*>data.data
        cdef int N
        N = data.size
        fda_.assign(array_start, array_start + N)
```

vs

```
        fda_.reserve(N)
        for i in range(N):
            fda_.push_back(data[i])
```

Performance measurements:
```
        $ python -m timeit -s 'import pyopenms, numpy; d = numpy.random.rand( 100000).astype(numpy.float32); f = pyopenms.FloatDataArray();' 'f.set_data(d)'    

        Using the push_back method:
            10000 loops, best of 3: 167 usec per loop
        Using the assign method:
            100000 loops, best of 3: 14 usec per loop

        # cmp to using Python push_back
        $ python -m timeit -r 3 -s 'import pyopenms, numpy; d = numpy.random.rand( 100000).astype(numpy.float32); f = pyopenms.FloatDataArray();' 'for val in d: f.push_back(val)'
            100 loops, best of 3: 8.14 msec per loop
```
We see an improvement of ca 10x using the assign method (and an 600x improvement compared to pure Python).

hroest:
all tests pass, only codacy complains for some reason (unclear why)

hroest:
@jpfeuffer @timosachsenberg can we merge this?

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -4949,44 +4963,23 @@ namespace OpenMS
...
-            os << "\t\t\t\t\t\t<cvParam cvRef=\"MS\" accession=\"" << bi_term.id << "\" name=\"" << bi_term.name << "\" />\n";
-          }
-          else
-          {
-            os << "\t\t\t\t\t\t<cvParam cvRef=\"MS\" accession=\"MS:1000786\" name=\"non-standard data array\" value=\"" << array.getName() << "\" />\n";
-          }
-          writeUserParam_(os, array, 6, "/mzML/run/spectrumList/spectrum/binaryDataArrayList/binaryDataArray/cvParam/@accession", validator);
-          os << "\t\t\t\t\t\t<binary>" << encoded_string << "</binary>\n";
-          os << "\t\t\t\t\t</binaryDataArray>\n";
+          writeBinaryFloatDataArray_(os, options_, array, s, m, true, validator);
```
hroest: this is the main change here in the code (on top of cosmetic changes): we now have support for writing compressed float arrays.

timosachsenberg: cool but should this go into the release? wanted to branch now

PR #3646, opened 2018-07-30
PercolatorAdapter parameter post-processing-tdc = true by default

lars20070:
In majority of use cases, we will search against a combined target-decoy database. Hence, the Percolator flag `post-processing-tdc` should be set `true` by default. This PR, changes the default value from `false` to `true`.


timosachsenberg:
+1

lars20070:
[By default](https://github.com/OpenMS/OpenMS/blob/develop/src/topp/PercolatorAdapter.cpp#L254) the `testFDR` parameter is 1% FDR. How useful is a [TOPP test](https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/THIRDPARTY/PercolatorAdapter_1.ini#L19) with 50% FDR?

timosachsenberg:
What's the problem with setting the testFDR to 50%?

lars20070:
The `ini` file parameters in the TOPP test are [outdated](https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/THIRDPARTY/PercolatorAdapter_1.ini#L4) and no longer consistent with the current version.

src/tests/topp/THIRDPARTY/PercolatorAdapter_1_out.idXML
```diff
@@ -93,13 +93,10 @@
 				<UserParam type="float" name="COMET:lnNumSP" value="9.48311183169221"/>
 				<UserParam type="float" name="COMET:lnRankSP" value="0"/>
 				<UserParam type="float" name="COMET:IonFrac" value="0.208333333333333"/>
-				<UserParam type="float" name="MS:1001492" value="0.343995"/>
-				<UserParam type="float" name="MS:1001491" value="0.357143"/>
```
timosachsenberg: why are these missing?

timosachsenberg: ok I can add these in a different branch

timosachsenberg:
http://cdash.openms.de/testDetails.php?test=51063668&build=279832

timosachsenberg:
please don't use word or similar to edit source files as they add strange characters (again the quotes in this case)

lars20070:
The `PercolatorAdapter_1.idXML` input file comes from a combined target-decoy search. But `post-processing-tdc = false` in the original TOPP test, which is not correct (as I understand this parameter). IMO, we should not get the old output `PercolatorAdapter_1_out.idXML`.

But feel free to correct the test.

timosachsenberg:
no my comment was about the meta values. please continue updating the tests

timosachsenberg:
http://cdash.openms.de/testDetails.php?test=51155506&build=280138

lars20070:
@jpfeuffer Ready to merge?

Might be useful for @MaxEsefeld.

timosachsenberg:
Hmm did you check that this is compatible with percolator in the third party folder

timosachsenberg:
I think I tried it recently and it did not have the parameter

MaxEsefeld:
As I mentioned in the gitter group: It wasn´t clearly documented if the PercolatorAdapter is filtering the hits by itself or if its necessary to use the IDFilter or the FDR node. And something different: @lars20070 send me some old knwf and my OpenMS Version (2.4.-HEAD-2018-08-28) wasn´t able to open the dialog window.

timosachsenberg:
PercolatorAdapter does no filtering so IDFilter is still required

lars20070:
@timosachsenberg The param exists. Advanced.

screenshot from 2018-11-14 16-41-45

timosachsenberg:
Yes but is it available in the percolator version we ship in THIRDPARTY

lars20070:
Yes. I just did run help.
```
--post-processing-tdc                        Replace the mix-max method by 
                                              target-decoy competition for 
                                              assigning q-values and PEPs. If 
                                              the input PSMs are from separate 
                                              target and decoy searches, 
                                              Percolator's SVM scores will be 
                                              used to eliminate the lower 
                                              scoring target or decoy PSM(s) of 
                                              each scan+expMass combination. If 
                                              the input PSMs are detected to be 
                                              coming from a concatenated search, 
                                              this option will be turned on 
                                              automatically, as this is 
                                              incompatible with the mix-max 
                                              method. In case this detection 
                                              fails, turn this option on 
                                              explicitly.
```

MaxEsefeld:
> PercolatorAdapter does no filtering so IDFilter is still required

No Problem about that. :) It is just because the Percolator implements a new score (SVM?) and when setting the pep score in the IDFilter straight to 0.01 the idXML turns out to be empty...  Therefore, the FalseDiscoveryRate Node is also necessary. Am I right?

timosachsenberg:
Ah yes. I thought it, by default, annotates the q-value. But if it is the SVM score then FDR tool is also needed.

lars20070:
@timosachsenberg Merge?

src/topp/PercolatorAdapter.cpp
```diff
@@ -268,8 +268,9 @@ class PercolatorAdapter :
...
-    registerFlag_("post-processing-tdc", "Use target-decoy competition to assign q-values and PEPs.", is_advanced_option);
-    registerFlag_("train-best-positive", "Enforce that, for each spectrum, at most one PSM is included in the positive set during each training iteration. If the user only provides one PSM per spectrum, this filter will have no effect.", is_advanced_option);
+    registerStringOption_("post-processing-tdc", "<value>", "true", "Use target-decoy competition to assign q-values and PEPs.", !is_required, is_advanced_option);
+    setValidStrings_("post-processing-tdc", ListUtils::create<String>("true,false"));
+     registerFlag_("train-best-positive", "Enforce that, for each spectrum, at most one PSM is included in the positive set during each training iteration. If the user only provides one PSM per spectrum, this filter will have no effect.", is_advanced_option);
```
timosachsenberg: why is this a flag and the other one a true / false option? should we make this consistent?

timosachsenberg: true. can you add the string option for the other flag as well?

lars20070:
Merge?

timosachsenberg:
can you merge develop into it to trigger a rebuild?

lars20070:
Under Ubuntu/Mac all tests pass.

lars20070:
@jpfeuffer In my local tests, the test numbers match. See for example `16`.
```
        Start  16: Multithreading_test
 16/2222 Test  #16: Multithreading_test ................................................   Passed    0.01 sec
```
On Travis they do not. `2109` and `2110` are supposed to fail.
```
2109/2270 Test #2013: TOPP_ProteinQuantifier_1_out1 ......................................   Passed    0.09 sec
        Start 2016: TOPP_ProteinQuantifier_2_out1
        Start 2017: TOPP_ProteinQuantifier_2_out2
        Start 2019: TOPP_ProteinQuantifier_3_out1
2110/2270 Test #2017: TOPP_ProteinQuantifier_2_out2 ......................................   Passed    0.09 sec
```
This PR is a simple flag change false -> true. All tests pass locally on my systems.

timosachsenberg:
a lot of the jenkins tests are failing. 
Internal error: Request for parameter with wrong type 'train-best-positive'

lars20070:
Cannot re-produce the failing test. All tests pass under Ubuntu.

Any advice how to debug?

timosachsenberg:
can you check again if that parameter exists in the tools from THIRDPARTY?

lars20070:
Bingo. - No `train-best-positive` is no param.

(`post-processing-tdc` is a param, which I checked earlier. `train-best-positive` I only touched for sake of consitency.)

How come params in tool and adapter don't match? `#workflowtests`

timosachsenberg:
might only be in the crux percolator: http://crux.ms/commands/percolator.html

lars20070:
rebuild jenkins

jpfeuffer:
https://github.com/OpenMS/OpenMS/blob/e6e73768f87ff6dfc29e249aa277c25c739fae17/src/topp/PercolatorAdapter.cpp#L1151

still used here.

But I am not sure if removing the parameter is a good idea. Shouldn't newer version support it?

timosachsenberg:
yes, I think it is included starting 3.2 https://github.com/percolator/percolator/releases

jpfeuffer:
@timosachsenberg you commited merge conflicts. do you have it locally and can remove them?

timosachsenberg:
replaced by new PR

PR #3647, opened 2018-07-30
[FIX] add warning if ID can't be mapped back

timosachsenberg:
but still write out the meta values (for consistency)


timosachsenberg:
any comments? just debug code...

PR #3649, opened 2018-07-30
[FEATURE] OpenPepXL score update

enetz:
new score for cross-linking MS hits based on Linear Discriminant Analysis and Regression Analysis by Tjeerd Dijkstra on a representative dataset,


enetz:
I just noticed, that I need to redo this.
Tried to put it into the PeptideHit generation, since I don't want to compute the precursor error,
but the sorting and filtering of results is done before that, so this won't work.

PR #3650, opened 2018-07-30
[FEATURE] OpenPepXL score update, 2nd try

enetz:
new score for cross-linking MS hits based on Linear Discriminant Analysis and Regression Analysis by Tjeerd Dijkstra on a representative dataset.

This duplicates some code for precursor error calculation, but I will take care of that with another refactoring PR.


timosachsenberg:
do you distinguish CID and HCD?

PR #3654, opened 2018-07-31
[Fix] doxygen js

### LINKED ISSUE #3631: Doxygen html pages are missing javascript
This happens for nightly and past builds:

- http://ftp.mi.fu-berlin.de/pub/OpenMS/release-documentation/html/classOpenMS_1_1ReactionMonitoringTransition.html
- https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1ReactionMonitoringTransition.html

```
Uncaught ReferenceError: $ is not defined
    at classOpenMS_1_1React…gTransition.html:30
```

so this means that we do not have a navigation tree (I am fine with that), but also no expanding of inherited members (that is annoying). 

it seems that jquery is missing and you can manually get it all working without errors if you include the following in your html:

```
    <script src="http://ajax.googleapis.com/ajax/libs/jquery/1.6.1/jquery.min.js" type="text/javascript"></script>
    <script src="navtreedata.js" type="text/javascript"></script>
    <script src="navtree.js" type="text/javascript"></script>
    <script src="dynsections.js" type="text/javascript"></script>
```

jpfeuffer:
fixes #3631


jpfeuffer:
also updates contrib

PR #3657, opened 2018-08-01
OpenPepXLAlgorithm, XL-MS search class

enetz:
refactored OpenPepXL into a class, with the TOPPTool mainly handling file input and output.

Also abstracted some code into functions.

Planning to do the same with the second tool OpenPepXLLF, but that is not finished yet.


timosachsenberg:
cool. Tests and pxd for the additional classes are needed though...

enetz:
Added class for OpenPepXLLF and python bindings for new functions and the two new classes.
Making tests for the new functions in OPXLHelper now,

but does it make sense to make test for the OpenPepXLAlgorithm classes?
They only have a `run` function and all the test would do is just copy the OpenPepXL tool code to read test files, which the tool tests already test.

timosachsenberg:
I think the tests don't need to be extensive but merely test if parameter passing and algorithm execution work. So you probably can just reuse the test data.

timosachsenberg:
looks good.
minor comments in general:
- document score direction of scores in further PRs
- make more variables const

timosachsenberg:
Is this correct? log transform only in ppm part?
```
    if (fragment_mass_tolerance_unit_ppm)
    {
      range = std::log(theoretical_spec.back().getMZ()) - std::log(theoretical_spec[0].getMZ());
      used_tolerance = fragment_mass_tolerance / 1e6;
    }
    else
    {
      range = theoretical_spec.back().getMZ() - theoretical_spec[0].getMZ();
      used_tolerance = fragment_mass_tolerance;
    }
```

enetz:
yes, the log transform is to get the range into the same "space" as ppm tolerances, where we can treat pmm tolerances as absolute instead of relative. Basically instead of computing absolute tolerances from a mass and a relative tolerance, we just make the whole spectrum "relative"

enetz:
but in practice that did not change the results much compared to just pre-calculating the absolute tolerance from the mean of peak m/zs, as it is done in the matchOdds

PR #3659, opened 2018-08-02
Update CHANGELOG for 2.4

hroest:
This is a place to collect changes for the changelog


jpfeuffer:
I was trying to add the details from changelog_helper but I forgot the adapters because the executables were not in my path.

jpfeuffer:
I actually used the wrong release. Should be fine now, please double check.

jpfeuffer:
Ah I guess it was a leftover in my bin folder.

PR #3661, opened 2018-08-02
[FIX, GUI] fixed output of "Curated" column in TOPPView csv export

enetz:
writes out the status of the "Curated" checkboxes as "1" for checked and "0" for unchecked


timosachsenberg:
failed test unrelated

PR #3663, opened 2018-08-02
[FEATURE] Compute Percolator features across replicates

sophiamersmann:
Add an option in `PSMFeatureExtractor` to compute Percolator features across multiple replicate runs (i.e. the same sample analysed multiple times). For each given input file, one output file is produced.  

Currently implemented features are more like suggestions / examples and include:
- identical unmodified peptide found in the same or other runs (bool)
- number of identical unmodified peptides found in the same and other runs
- same peptide with different modification found in the same or other runs (bool)
- number of different peptide modifications found in the same and other runs
- identical unmodified peptide found in other runs (bool)
- number of identical unmodified peptides found in other runs
- same peptide with different modification found in other runs (bool)
- number of different peptide modifications found in other runs

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PercolatorFeatureSetHelper.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PercolatorFeatureSetHelper.h
@@ -157,6 +157,15 @@ namespace OpenMS
         static void addCONCATSEFeatures(std::vector<PeptideIdentification>& peptide_id_list, StringList& search_engines_used, StringList& feature_set);
 
         /**
+         * @brief addREPLICATEFeatures
+         * @param peptide_id_list PeptideIdentification vector to create Percolator features in
+         * @param feature_set register of added features
+         *
+         * Creates and adds Percolator features across replicates and registers them in feature_set
+         */
+        static void addREPLICATEFeatures(std::vector<PeptideIdentification>& peptide_id_list, StringList& feature_set);
+
+        /**
          * @brief checkExtraFeatures
          * @param psms the vector of PeptideHit to be checked
          * @param extra_features the list of requested extra features
--- a/src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
+++ b/src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
@@ -405,11 +405,11 @@ namespace OpenMS
                   double score = inner_pep->getScore();
                   if (is_e_value) score = -log(score + 0.0001);
 
-                  if (outer_pep->getMetaValue("TMP:sources") == inner_pep->getMetaValue("TMP:sources"))  // PSMs from different runs
+                  if (outer_pep->getMetaValue("TMP:sources") == inner_pep->getMetaValue("TMP:sources"))  // PSMs from the same run
                   {
                     replicate_spectra += score;
                   }
-                  else  // PSMs from the same run
+                  else  // PSMs from different runs
                   {
                     sibling_experiments += score;
                   }
--- a/src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
+++ b/src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
@@ -316,7 +316,7 @@ namespace OpenMS
         for (vector<PeptideHit>::iterator hit = it->getHits().begin(); hit != it->getHits().end(); ++hit)
         {
           String sequence = hit->getSequence().toUnmodifiedString();
-          String sources = hit->getMetaValue("TMP:sources").toString();
+          String sources = hit->getMetaValue("TMP:sources");
           matches[sequence][sources].push_back(*hit);
         }
       }
@@ -336,7 +336,7 @@ namespace OpenMS
           for (std::map<String, vector<PeptideHit>>::iterator item = match_group.begin(); item != match_group.end(); ++item)
           {
             String sources = item->first;
-            if (sources != hit->getMetaValue("TMP:sources").toString())
+            if (sources != hit->getMetaValue("TMP:sources"))
             {
               vector<PeptideHit> peptide_matches = item->second;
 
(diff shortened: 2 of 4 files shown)


timosachsenberg:
@jpfeuffer how do you annotate the MS run as meta value?

timosachsenberg:
also interesting: if peptide was found with different charge states

jpfeuffer:
I did not annotate simple idxmls yet since I was working on consensusXMLs. There the IDs have the map_index which tells you the consensusHeader (file, label) they belong to. We need to think about a scheme (probably outside of this PR) where we can take over that information into standalone Idxmls ( e.g. ripped from a consensusXML). 
As an idea, we could add the (sub)map of consensusHeaders each IDRun corresponds to, to each IDRun (and take over the map_index). Idxml somehow needs to know about the design I think. Otherwise we have to store file (and label?) in every peptide Id.

sophiamersmann:
@jpfeuffer @timosachsenberg Is my workaround okay then for the moment?

sophiamersmann:
I updated the features to features similar to those implemented in iProphet. For all those features that currently have 2 versions (considering all hits or only the top hit for each file), I am currently testing which one is the better option; I suspect considering the top hit only will be better though.

jpfeuffer:
Yes I think the workaround is fine for your needs. Maybe we have to check, since you merge the searches we might need to keep the annotated metavalue to figure out later, from which file it came from.
But I really like the new features and I think it's a good idea to check the two versions.

sophiamersmann:
What do you mean by merge the searches? I think I keep them separated; If x input files are given, x output files are written. Or do you mean something else?

sophiamersmann:
Percolator features based on top hits only seem to result in superior performance: [https://sophiamersmann.github.io///2018/08/07/replicate-features/](https://sophiamersmann.github.io///2018/08/07/replicate-features/)

sophiamersmann:
I removed them in the last commit. Might be a bit confusing because I simplified the names as well (REP:siblingSearchesTop is now simply REP:siblingSearches).

Leon-Bichmann:
I think it would be great to have this feature for a merged idXML of several runs as well. I usually merge replicates using IDMerger and set the flag -annotate_file_origin. This results in a metavalue for each peptide identifictation:

UserParam type="string" name="file_origin"

How about length variants of +- 1 or 2 residues? would that also raise the confidence?

timosachsenberg:
@Leon-Bichmann couldn't you just run the feature extractor before merging?
regarind the +-1 2 residues. this might mess a bit with the enzyme settings. would you restrict yourself to semi-tryptic cleavage or the non-specific cleavage as for your MHC peptides?

Leon-Bichmann:
True psm feature extractor could be run before merging.
And yes that might be better for semi- or unspecific cleaved peptides.

src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
```diff
@@ -287,6 +287,142 @@ namespace OpenMS
...
+                // two peptide hits with identical charge and sequence
+                if (outer_pep->getCharge() == inner_pep->getCharge() &&
+                    outer_pep->getSequence().toUnmodifiedString() == inner_pep->getSequence().toUnmodifiedString())
+                {
+                  double score = inner_pep->getScore();
+                  if (is_e_value) score = -log(score + 0.0001);
+
+                  if (outer_pep->getMetaValue("TMP:sources") == inner_pep->getMetaValue("TMP:sources"))  // PSMs from different runs
+                  {
+                    replicate_spectra += score;
```
timosachsenberg: did you also test absolute counts instead of score?
is the comparison correct? if the source is the same, shouldn't it be the same run?

sophiamersmann: The code is correct - the comments should be swapped.

For features taking into account the precursor ion, I did not test absolute counts. Do you think that would work better?

timosachsenberg: hmm I have no idea. probably not but if it is not much work may be worth a shot.

timosachsenberg: Ok. Can you check and correct the comments in the code? I can update the pyOpenMS part if you want. In these dataset it indeed seems to be the case that these features are not very important. How costly are these to calculate?

sophiamersmann: Sorry - I was busy writing my thesis. I'm going to correct the comments, update the pyOpenMS part, and implement the new features asap.

I'd say it takes like 5-10 minutes - so, it's not that bad (but a bit annoying).

src/openms/include/OpenMS/ANALYSIS/ID/PercolatorFeatureSetHelper.h
```diff
@@ -156,6 +156,15 @@ namespace OpenMS
...
         static void addCONCATSEFeatures(std::vector<PeptideIdentification>& peptide_id_list, StringList& search_engines_used, StringList& feature_set);
 
+        /**
+         * @brief addREPLICATEFeatures
+         * @param peptide_id_list PeptideIdentification vector to create Percolator features in
+         * @param feature_set register of added features
+         *
+         * Creates and adds Percolator features across replicates and registers them in feature_set
+         */
+        static void addREPLICATEFeatures(std::vector<PeptideIdentification>& peptide_id_list, StringList& feature_set);
```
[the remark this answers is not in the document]
sophiamersmann: Will do 👍

timosachsenberg:
rebuild jenkins

hroest:
@sophiamersmann @timosachsenberg where are we on this PR? Can this be merged soon?

timosachsenberg:
I think it is basically finished but Julianus  wanted to give it another deeper look in the context of his protein inference research.

sophiamersmann:
Is there anything I can do to help? I just saw that I promised to check the precursor ion features using absolute counts but never delivered. I was busy celebrating the end of my Master's I guess 💃 🚀  but I'm based in Barcelona now and could give this some time on the weekend if you think it's worth a shot.

timosachsenberg:
Regarding summing up the scores I was just a bit concerned that this might lead to very different results depending on whether PEP or 1-PEP is used (or other types of scores). Counts might be more robust in that regards. If it is not much trouble it would be great to see if that makes any difference.

hroest:
it would also be great to have some tests for these new options

PR #3669, opened 2018-08-03
[FEATURE] conversion between different ion mobility formats

hroest:
Convert between two different storage format for ion mobility data:

- one `<spectrum>` tag per ion mobility value (this is how the MS records the data)
- one `<spectrum>` tag per frame, basically storing 3D data in m/z, ion mobility, intensity

the latter is more space efficient as we can see here on an example dataset:

```
-rw-rw-r-- 1 hr   hr    61M Aug  3 12:55 file.single.mzML
-rw-rw-r-- 1 hr   hr    75M Aug  3 12:55 file.multiple.mzML
-rw-rw-r-- 1 hr   hr    15M Aug  3 12:56 file.single.zlib.mzML
-rw-rw-r-- 1 hr   hr    62M Aug  3 12:56 file.multiple.zlib.mzML
-rw-rw-r-- 1 hr   hr    20M Aug  3 12:56 file.multiple.zlib.mzML.gz
-rw-rw-r-- 1 hr   hr    12M Aug  3 12:56 file.single.zlib.mzML.gz

```

as you can see when using zlib compression we get 4x reduction in size and even when gzipping the whole file, we get a 1.7x reduction in size, overall going from 75MB to 12MB.

also we can visualize the latter easily:

```
./bin/TOPPView /tmp/file.single.zlib.mzML
```

screenshot from 2018-08-03 12-57-33

screenshot from 2018-08-03 12-57-46

TODO:

- add tests
- make conversion more efficient (currently we copy all spectra)
- support low_memory processing
- use a proper tag for the IM array


hroest:
probably

```
[Term]
id: MS:1002893
name: ion mobility array
def: "An array of ion mobility data." [PSI:PI]
is_a: MS:1000513 ! binary data array
```

is the right term to support...

hroest:
PS: the nice thing is that on the single-spectrum data all of our tools such as FeatureFinder should work as expected as they simply ignore the IM FloatDataArray

timosachsenberg:
rebuild jenkins

src/topp/FileConverter.cpp
```diff
@@ -444,6 +481,104 @@ class TOPPFileConverter :
           s.getInstrumentSettings().setScanMode(InstrumentSettings::SRM);
         }
       }
+
+      if (store_im == "multiple")
+      {
+        bool has_im = false;
```
timosachsenberg: maybe we could add a helper function getIMType() that returns an enum IM_NONE, IM_STACKED_ ...

Currently, you only specify how to store it right? Is the input format always the same? Because if not I don't know if the check for dirft time / data arrays work for every input combination (I am probably missing something here ;) ).

hroest: > Currently, you only specify how to store it right? Is the input format always the same?

yes, this currently only works to convert from one format to the other and you have to know which format your data is in, otherwise it wont work

hroest:
rebuild jenkins

src/topp/FileConverter.cpp
```diff
@@ -143,6 +144,142 @@ String extractCachedMetaFilename(const String& in)
   return in_meta;
 }
 
+namespace OpenMS {
+enum IMType
```
hroest: @timosachsenberg any idea where this should go? Where would be a good place to put this? E.g. `KERNEL/SpectrumHelper.h`?

Issue #3660, opened 2018-08-02
IM support in TOPPView


hroest:
We need to improve ion mobility support in toppview, right now it only works if there is a float data array called "Ion Mobility" - this is not yet portable!

PR #3670, opened 2018-08-03
[FIX] allow multiple array names for IM

hroest:
allow multiple different values for the ion mobility array, including 

```
[Term]
id: MS:1002893
name: ion mobility array
def: "An array of ion mobility data." [PSI:PI]
is_a: MS:1000513 ! binary data array
```


hroest:
fixes #3660

timosachsenberg:
+1

PR #3672, opened 2018-08-07
fixed OPXLHelper test to fix nightly build

enetz:
made test more robust by using local objects
and explicitly setting a precursor correction


hroest:
it seems that the issue here is missing default values for data structures and then forgetting to set these values, maybe we should add a requirement to initialize data structures to the https://github.com/OpenMS/OpenMS/wiki/Coding-conventions#class-requirements @timosachsenberg @jpfeuffer ?

PR #3677, opened 2018-08-10
Feature/header fixes

hroest:
- removes unnecessary includes in OpenMS
- adds forward declarations where possible


hroest:
this may reduce overall compile and should drastically reduce partial compile time if only certain header files are changed which are now included in fewer cpp files.

hroest:
seems like there are some linking issues now on Windows: http://cdash.openms.de/viewBuildError.php?buildid=281723

```
8>SwathFile.obj : error LNK2019: unresolved external symbol "public: void __cdecl OpenMS::FullSwathFileConsumer::retrieveSwathMaps(class std::vector<struct OpenSwath::SwathMap,class std::allocator<struct OpenSwath::SwathMap> > &)" (?retrieveSwathMaps@FullSwathFileConsumer@OpenMS@@QEAAXAEAV?$vector@USwathMap@OpenSwath@@V?$allocator@USwathMap@OpenSwath@@@std@@@std@@@Z) referenced in function "public: class std::vector<struct OpenSwath::SwathMap,class std::allocator<struct OpenSwath::SwathMap> > __cdecl OpenMS::SwathFile::loadMzML(class OpenMS::String,class OpenMS::String,class boost::shared_ptr<class OpenMS::ExperimentalSettings> &,class OpenMS::String)" (?loadMzML@SwathFile@OpenMS@@QEAA?AV?$vector@USwathMap@OpenSwath@@V?$allocator@USwathMap@OpenSwath@@@std@@@std@@VString@2@0AEAV?$shared_ptr@VExperimentalSettings@OpenMS@@@boost@@0@Z) [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\openms\OpenMS.vcxproj]
```

but it works on g++ and clang - could somebody look into the Windows stuff that has some experience?

timosachsenberg:
I have been struggling with these issues in the past when I used include-what-you-use.
I just guess that our VC is not entirely standard conforming.

timosachsenberg:
let's close for now

PR #3678, opened 2018-08-14
Fix/pyopenms

hroest:
Update Python bindings for Release 2.4


hroest:
@jpfeuffer can you update the docker pyopenms to the newest version of autowrap?

hroest:
@timosachsenberg @jpfeuffer can we merge this? These are pyOpenMS changes which I can either merge right now or I will prepare a separate release of pyOpenMS based on these changes. Whatever is easier...

PR #3679, opened 2018-08-14
Feature/nested p

hroest:
Use nested parallelization in OpenSWATH - we have two loops and if we have a certain number of threads, we cannot really increase the number of threads on the outer loop in any useful manner. Instead, we will parallelize over the batches using a set of inner thread workers.


hroest:
failures unrelated:

The following tests FAILED:
	2087 - TOPP_SiriusAdapter_1_out (Failed)
	2089 - TOPP_SiriusAdapter_2_out (Failed)
Submit files (using http)

hroest:
@timosachsenberg merge?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathWorkflow.h
```diff
@@ -248,14 +283,16 @@ namespace OpenMS
...
   };
 
   /**
    * @brief Class to execute an OpenSwath Workflow
    *
-   * The workflow will perform a complete OpenSWATH analysis. Optionally, an RT
-   * transformation (mapping peptides to normalized space) can be obtained
-   * beforehand using the OpenSwathRetentionTimeNormalization class.
+   * The workflow will perform a complete OpenSWATH analysis, see
+   * performExtraction(). Optionally, a calibration of m/s and retention time
```
timosachsenberg: m/s to m/z

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -496,21 +503,47 @@ namespace OpenMS
...
+          // If we have a multiple of threads_outer_loop_ here, then use nested
+          // parallelization here. E.g. if we use 8 threads for the outer loop,
+          // but we have a total of 24 cores available, each of the 8 threads
+          // will then create a team of 3 threads to work on the batches
+          // individually.
+          //
+          // We should avoid oversubscribing the CPUs, therefore we use integer division.
+          // -- see https://docs.oracle.com/cd/E19059-01/stud.10/819-0501/2_nested.html
+          int outer_thread_nr = omp_get_thread_num();
+          omp_set_num_threads(std::max(1, total_nr_threads / threads_outer_loop_) );
```
timosachsenberg: In a different context, I saw the rule not to use more than #cores - 1 threads (except if hyperthreading).
We do allow for #cores thread here right?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -465,6 +465,13 @@ namespace OpenMS
     // in which they were given to the program / acquired. This gives much
     // better load balancing than static allocation.
 #ifdef _OPENMP
+    int total_nr_threads = omp_get_max_threads(); // store total number of threads we are allowed to use
+    if (threads_outer_loop_ > -1)
+    {
+      omp_set_nested(1);
```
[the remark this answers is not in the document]
hroest: hmm, it might but its easy to just set it back to `total_nr_threads` after the loop. Sounds like a good idea

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
hroest: well the user specifies the number of threads anyways. So in principle the user can set the number of threads to 128 on a single threaded machine and would get really poor performance. I am not sure whether there is an easy way to control for that if the user makes an error, we could check the maximum threads available but on a cluster where multiple jobs run per node that is not a good idea either (e.g. you have 3 jobs on a node with 48 cores, you get assigned 10 of these but your code doesnt know that it cannot use the other 38 and it looks like you have 48 for yourself). So in all these cases, the user needs to know and control how many cores to use. 

We should anyways test this on some real data, I am curious, where did you see the advise to use `#cores -1` threads?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -563,6 +597,13 @@ namespace OpenMS
 
     }
     this->endProgress();
+    
+#ifdef _OPENMP
+    if (threads_outer_loop_ > -1)
+    {
+      omp_set_num_threads(total_nr_threads); // set number of available threads back to initial value
+    }
+#endif
```
hroest: @timosachsenberg I have now added this code that sets the number of threads back to its original value

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
timosachsenberg: Mostly anecdotal ... would not give too much on that. I think it is intended to leave one core to the OS. Best is to benchmark I guess.

hroest: I think that may also differ on Desktop machines that need to run an X server and other stuff vs HPC nodes that are dedicated for compute.

timosachsenberg:
feel free to merge after resolving the conflict.

timosachsenberg:
I don't know if I understand all implications but could this be used to also speed up the FeatureFinderIdentification?

hroest:
> I don't know if I understand all implications but could this be used to also speed up the FeatureFinderIdentification?

I dont think so, since it does not use the `OpenSwathWorkflow` class, but the same principle could be applied.

PR #3682, opened 2018-08-15
Feature/fix fda

hroest:
bugfix after #3645 - fixes a bug where the wrong terms are written out and the resulting file is invalid


hroest:
since this is a bugfix and all tests pass, I will merge

PR #3688, opened 2018-08-20
Feature/open swath nonlinear

hroest:
- allow separate mz extraction window for MS1
- speed up mass calibration
- allow for non-linear RT calibration using more data points
  - using a two pass approach with iRT first, then a larger set of peptides for calibration
  - allows fast initial calibration and accurate second *m/z* and RT calibration


hroest:
rebuild jenkins

hroest:
@timosachsenberg @grosenberger looks ready to merge, any comments?

PR #3690, opened 2018-08-21
Feature/osw fixes

hroest:
A series of fixes and features for OpenSWATH using ion mobility:

- [FEATURE] add isolation window to Precursor
- [FEATURE] store isolation window for chromatograms 
- [FEATURE] simplify chromatogram extraction and scoring in OpenSwathWorkflow
- [FEATURE] filter SWATH Spectrum by ion mobility

PR #3691, opened 2018-08-21
Feature/precursor iso

hroest:
use multiple precursors

merge after #3690


hroest:
@grosenberger should we merge this?

grosenberger:
I tested this branch and if the number of isotopes is set to 0, everything works as usually. If it is set to a higher value, some downstream issues occur. But please already merge now whenever you think it is ready. I'm now working on the revamped precursor scoring that will resolve this issue, which will be a separate PR, building on top of this.

PR #3692, opened 2018-08-22
[DOC] add documentation

hroest:
substantial improvement of the OpenSWATH part of the docu (tools, classes)


hroest:
@jpfeuffer since this is documentation only it may make sense to merge it into the release if there is still time

timosachsenberg:
+1 for also merging that into the release docu if that isn't too much hassle right now

PR #3694, opened 2018-08-24
Feature/osw cli

hroest:
update OpenSWATH CLI, remove some default options (like 14s default peak width) and replace -ppm option with -mz_extraction_window_unit


hroest:
rebuild jenkins

grosenberger:
Great! Are there any consequences by removing the 14s requirement?

grosenberger:
Ok, sounds good. We also still have ``-Scoring:TransitionGroupPicker:compute_peak_quality true`` enabled by default that remove some peaks. I'm currently assessing whether that makes sense to disable in some scenarios.

timosachsenberg:
merge?

PR #3695, opened 2018-08-24
Update HMDB to 4.0

heckendorfc:
There was a lot of churn because of the HMDB accession numbers changing recently. Elements.xml also needed to be updated with some missing elements now used in HMDB.


heckendorfc:
Thanks for catching that. It was pulling from the right element "name" but at the wrong location in the xml hierarchy. The latest commit should fix that.

oliveralka:
@jpfeuffer Is the nightly status here critical? 
Continuous integration out of time +1 for merge.
@FabianAicheler  can you confirm?

FabianAicheler:
Old 3.5 HMDBMappingFile.tsv:
26.981538441	Al	HMDB:HMDB01247	HMDB:**HMDB15456**
27.010899037	CHN	HMDB:HMDB02084	HMDB:**HMDB35752**

New one:
26.981538441	Al	HMDB:HMDB0001247
27.010899037	CHN	HMDB:HMDB0002084	HMDB:HMDB0060292

So there seems to (at least sometimes) be a loss of old secondary identifiers. However, HMDB15456 can still be resolved at hmdb.ca, whereas HMDB35752 can't. So I am not sure how much importance we should place on carrying over old secondary identifiers.

heckendorfc:
Is that file supposed to display secondary accessions? It's currently only showing primary accessions.

FabianAicheler:
Hmm no, primary should be fine for HMDBMapping. Not sure why some identifiers dissapeared, but that seems to be probably because they changed the database,  I think our script works fine. So I would propose to just use the new version as is.

I would also be fine with merging

oliveralka:
@heckendorfc Thanks again!

PR #3696, opened 2018-08-24
[FEATURE] OpenSWATH: Precursor isotope scoring

grosenberger:
This PR complements #3691 from @hroest and adapts the scoring and reporting to be capable of dealing with multiple precursor isotopes.

- The cross-correlation co-elution and shape, as well as the mutual information scores are reported as precursor isotope scores and precursor contrast scores.
- Precursor isotope scores (var_xcorr_coelution, var_xcorr_shape, var_mi_score) are computed within the precursor isotopes if the monoisotopic peak and at least one isotope produced a signal.
- The precursor contrast scores are computed by comparing each precursor isotope chromatograms against the fragment ion chromatograms. In the special case where only the monoisotopic peak is present, the scores are exactly as in the old scores.

All precursor isotope quantities are reported in the OSW table ``feature_precursor``.  The existing table ``feature_ms1`` contains the summarized area and apex intensities of all isotopes, analogous to ``feature_ms2``.


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
```diff
@@ -270,23 +279,15 @@ namespace OpenMS
                                         << sub_it->getMetaValue("peak_apex_int") << ", " 
                                         << total_mi << "); ";
           }
-          else if (sub_it->metaValueExists("FeatureLevel") && sub_it->getMetaValue("FeatureLevel") == "MS1")
+          else if (sub_it->metaValueExists("FeatureLevel") && sub_it->getMetaValue("FeatureLevel") == "MS1" && sub_it->getIntensity() > 0)
```
grosenberger: We only report isotopes if any signal was detected at all. In my benchmark data set, this was the case for ~60% with four isotopes.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -652,23 +649,24 @@ namespace OpenMS
...
           native_ids_detection.push_back(transition_group_detection.getTransitions()[i].getNativeID());
         }
         for (Size i = 0; i < transition_group_detection.getPrecursorChromatograms().size(); i++)
         {
-          // try to identify the correct precursor native id
           String precursor_chrom_id = transition_group_detection.getPrecursorChromatograms()[i].getNativeID();
-          if (OpenSwathHelper::computePrecursorId(transition_group.getTransitionGroupID(), 0) == precursor_chrom_id)
+
+          // append precursor monoisotopic peak and additional isotopes with a signal for scoring
+          if (OpenSwathHelper::computePrecursorId(transition_group.getTransitionGroupID(), 0) == precursor_chrom_id || mrmfeature->getPrecursorFeature(precursor_chrom_id).getIntensity() > 0)
```
grosenberger: We always score the precursor monoisotopic peak and only those isotopes that have a signal.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -246,7 +246,7 @@ namespace OpenMS
     if (su_.use_coelution_score_)
     {
       scores.xcorr_coelution_score = mrmscore_.calcXcorrCoelutionScore();
-      scores.weighted_coelution_score = mrmscore_.calcXcorrCoelutionScore_weighted(normalized_library_intensity);
+      scores.weighted_coelution_score = mrmscore_.calcXcorrCoelutionWeightedScore(normalized_library_intensity);
```
grosenberger: I tried to improve the consistency of the function names in ``OpenSwathScoring`` as there are now many more options than in the very original version.

src/openswathalgo/include/OpenMS/OPENSWATHALGO/ALGO/MRMScoring.h
```diff
@@ -95,44 +95,67 @@ namespace OpenSwath
...
+    double calcXcorrShapeScore();
 
     /// calculate the weighted cross-correlation shape score
-    double calcXcorrShape_score_weighted(const std::vector<double>& normalized_library_intensity);
+    double calcXcorrShapeWeightedScore(const std::vector<double>& normalized_library_intensity);
 
-    /// calculate the weighted cross-correlation score
-    double calcXcorrCoelutionScore_weighted(const std::vector<double>& normalized_library_intensity);
+    /// calculate the cross-correlation contrast shape score
+    double calcXcorrContrastShapeScore();
```
grosenberger: We currently use ``initializeXCorrContrastMatrix`` only for transition-level scoring in IPF. However we could now directly implement SILAC scoring in combination with the new function ``calcXcorrContrastShapeScore``.

hroest: that would be interesting, however for SILAC scoring we would not have the second set of transitions -- but we could correlate the two MS1 traces

hroest:
@grosenberger can you fix these:

```
/.../OpenMS/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h:285:69: error: character constant too long for its type [-Werror]                                   
/.../OpenMS/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h:285:69: error: character constant too long for its type [-Werror]                                   
/.../OpenMS/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp:59:6: error: unused parameter ‘quantification_cutoff_’ [-Werror=unused-parameter]
```

@grosenberger did you test the resulting pqp files?

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -68,12 +68,9 @@ void processFeatureForOutput(OpenMS::Feature& curr_feature, bool write_convex_hu
...
 
-  // Sum up intensities of the MS2 features
-  if (curr_feature.getMZ() > quantification_cutoff_ && ms_level == "MS2")
-  {
-    total_intensity += curr_feature.getIntensity();
-    total_peak_apices += (double)curr_feature.getMetaValue("peak_apex_int");
-  }
+  // Sum up intensities of the features
+  total_intensity += curr_feature.getIntensity();
+  total_peak_apices += (double)curr_feature.getMetaValue("peak_apex_int");
```
hroest: maybe we should split MS1 and MS2 intensity - does this here sum up the total intensity across MS1 and MS2 features? is that a good idea?

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
hroest: hmm, I wonder whether that is really fair because you then compare precursors with 3 isotopes to those with only 1 -- also it will lead to a situation where when you have a tiny signal your score becomes much worse...

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -255,16 +255,23 @@ namespace OpenMS
...
     {
-      mrmscore_.initializeMS1XCorr(imrmfeature, native_ids, precursor_feature_id); // perform cross-correlation on monoisotopic precursor
-      scores.xcorr_ms1_coelution_score = mrmscore_.calcMS1XcorrCoelutionScore();
-      scores.xcorr_ms1_shape_score = mrmscore_.calcMS1XcorrShape_score();
+      // we need at least two precursor isotopes
+      if (precursor_ids.size() > 1)
+      {
+        mrmscore_.initializeXCorrPrecursorMatrix(imrmfeature, precursor_ids);
+        scores.ms1_xcorr_coelution_score = mrmscore_.calcXcorrPrecursorCoelutionScore();
+        scores.ms1_xcorr_shape_score = mrmscore_.calcXcorrPrecursorShapeScore();
```
hroest: what are the default values here (I think 0) -- however 0 is also the best possible score you can get. So in case you dont have any 2nd isotope, you get the best possible score? That may mess up the scoring.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
grosenberger: I will have a closer look at this. The problem is that on MS1-level the signal is much more sparse (or clean) than on MS2-level in the Orbitrap data.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
grosenberger: MS1 and MS2 signals are summed up separately, combining them would probably be a bad idea. The function is called twice and ``total_intensity`` and ``total_peak_apices`` are attached as separate MS1 and MS2 meta values to the main feature for easy access when writing the reports.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
grosenberger: Indeed, this is very dangerous. It will work when only MS1 traces with a signal are supplied, but not otherwise. I will look into improving the exception management here.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
hroest: yes, we probably have to look into this more -- the MS1 signal being "cleaner" could be an artefact of low fill time?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
hroest: I am more worried about the scoring, lets assume 2 precursors:

1. precursor has low but detectable MS1_isotope1 signal, therefore gets scored and gets a score of 0.6
2. precursor has no detectable MS1_isotope1 signal, therefore Line 267 to 271 is skipped and it gets a default score of 0 -- which makes its score better than the other signal. I think that will lead to inflation of scores around zero

hroest:
looks good, do you have a test where you extract more than 1 precursor trace?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
grosenberger: I now implemented a failsafe to address this problem, however we might want to consider whether we should make this even more generic, see below.

grosenberger:
I now added several changes:
- ``Combined`` scores: The combined scores treat transition and precursor isotope chromatograms identically. This might work since the used cross-correlation and MI scores are normalized.
- Variable number of transitions and precursor isotopes. I have the impression that the Orbitrap data that we are using is much more preprocessed than the QTOF data that the scores were originally developed on. This results in MS1 and MS2 chromatograms being completely empty. When we score them using our cross-correlation and MI scores, the result is that the scores are downvoted substantially, which is of course the desired behavior. So maybe it might make sense to drop these chromatograms or alternatively to dynamically select transitions that result in chromatograms with a signal. A new set of parameters sets the maximum and minimum number of transitions or precursor isotopes to select. A big drawback is that selected transitions might now differ between different runs. I'm not sure whether we actually want that or whether we should chose a different strategy.
- Unit tests are added for multiple precursor isotopes.

src/tests/topp/FeatureFinderIdentification_3_output.featureXML
```diff
@@ -2903,7 +2927,7 @@
 		<feature id="f_17447818880105247092">
 			<position dim="0">1658.35957551813</position>
 			<position dim="1">532.248746583271</position>
-			<intensity>202348</intensity>
+			<intensity>202355</intensity>
```
grosenberger: This difference seems to be a newly introduced bug that needs to be addressed.

src/tests/topp/OpenSwathWorkflow_11_output.featureXML
```diff
@@ -12,7 +12,7 @@
 			<intensity>141615</intensity>
 			<quality dim="0">0</quality>
 			<quality dim="1">0</quality>
-			<overallquality>-1.58727</overallquality>
```
grosenberger: Several scores differ with SONAR parameters enabled. Something is not working right yet.

src/tests/topp/FeatureFinderIdentification_3_output.featureXML
grosenberger: Ok, it seems that this value is within the error tolerance as the current head of ``OpenMS/develop`` produces the same value. The difference is thus not originating from this PR.

hroest:
> Variable number of transitions and precursor isotopes.

lets hold off with that a bit, maybe keep it on your own branch and evaluate how well this works before we merge this? I think we can go ahead and merge the rest and make some progress here already

grosenberger:
I agree, this would get too convoluted otherwise. This PR is ready from my side.

PR #3702, opened 2018-08-28
[FIX] annotate "spectra_data"

oliveralka:
Annotation was set beforehand but lost on the way. 

This will fix the issue. The fix is critical for the Release 2.4.0, since the Pipeline using the FeautreFinderMetabo will otherwise not annotate the filenames correctly.
The tools used further downstream will use "UNKOWN" instead.


hroest:
@oliveralka in that case I suggest to create a PR against the release branch and not `develop`

PR #3703, opened 2018-08-28
[FIX] annotate "spectra_data"

oliveralka:
Annotation was set beforehand but lost on the way.

This will fix the issue. The fix is critical for the Release 2.4.0, since the Pipeline using the FeautreFinderMetabo will otherwise not annotate the filenames correctly.
The tools used further downstream will use "UNKOWN" instead.

(cheery-picked commit 35caee889efd87002dbb1243e1d3c9b9e769c772)

@hroest thank you for the suggestion


timosachsenberg:
Looks good

PR #3712, opened 2018-09-04
minor improvments to SWATHWorkflow

cbielow:
- improve docs
- optimize loop end condition
- avoid new/delete
- use platform specific temp dir (which also changes with environment automatically)
- force directory separator at the end of tempDir to avoid late crash due to file-not-found for internal temp files

### DIFF
--- a/src/utils/OpenSwathWorkflow.cpp
+++ b/src/utils/OpenSwathWorkflow.cpp
@@ -440,13 +440,13 @@ protected:
     setValidFormats_("tr_irt", ListUtils::create<String>("traML,tsv,pqp"));
 
     // one of the following two needs to be set
-    registerInputFile_("tr_irt_nonlinear", "<file>", "", "transition file ('TraML')", false);
+    registerInputFile_("tr_irt_nonlinear", "<file>", "", "additional nonlinear transition file ('TraML')", false);
     setValidFormats_("tr_irt_nonlinear", ListUtils::create<String>("traML,tsv,pqp"));
 
     registerInputFile_("rt_norm", "<file>", "", "RT normalization file (how to map the RTs of this run to the ones stored in the library). If set, tr_irt may be omitted.", false, true);
     setValidFormats_("rt_norm", ListUtils::create<String>("trafoXML"));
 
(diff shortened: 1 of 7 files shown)


hroest:
looks good, it seems like travis has some issues now: 

```
Error: Unable to write file (the file '/home/travis/build/OpenMS/OpenMS/_build/src/tests/topp/osw_file5/openswath_tmpfile_ms1.mzML' could not be created)
```

hroest:
> use platform specific temp dir (which also changes with environment automatically)

I dont think that quite worked as expected, as now it tries to write the file `osw_file5/openswath_tmpfile_ms1` instead of the intended `osw_file5_openswath_tmpfile_ms1` file. @cbielow you will have to change the CMake test file, but also add an additional dependency to make sure the temp files are not overwritten as they will have the same name ... 

Currently it was possible to specify filename and directory, but now this is not possible any more, leading to the test failures.

src/utils/OpenSwathWorkflow.cpp
```diff
@@ -682,7 +684,10 @@ class TOPPOpenSwathWorkflow
 
     String readoptions = getStringOption_("readOptions");
     String mz_correction_function = getStringOption_("mz_correction_function");
-    String tmp = getStringOption_("tempDirectory");
+    
+    // make sure tmp is a directory with proper separator at the end (downstream methods simply do path + filename)
+    // (do not use QDir::separator(), since its platform specific (/ or \) while absolutePath() will always use '/')
+    String tmp_dir = String(QDir(getStringOption_("tempDirectory").c_str()).absolutePath()).ensureLastChar('/');
```
hroest: before it was possible to specify filename and directory, but now this is not possible any more, leading to the test failures. This was a good idea if you wanted to analyze multiple runs without worrying about them overwriting each others files.

cbielow:
ok, I see the problem, but IMHO this was misleading. `tempDirectory` was not a directory, but rather a prefix to a fixed suffix.
The proper fix would be to use proper random filenames in the specified directory. This would avoid to specifiy artificial dependencies in ctest as well. Does that give headaches to anyone? If desired, one could print the filename to the console, in case someone wants to check the content (but the cache file is binary, so its hard to inspect anyway).

hroest:
> The proper fix would be to use proper random filenames in the specified directory.

:+1: 

I think right now the test 5 and 6 will overwrite each others files ...

cbielow:
ready for merge

oliveralka:
+1 
macOS fail is unrelated

PR #3716, opened 2018-09-05
[FEATURE] new class PrecursorPurity

enetz:
This class computes metrics for precursor isolation window purity (or noise).
The function extracts the peaks from an isolation window targeted for fragmentation
and determines which peaks are isotopes of the target (MS2 precursor m/z) and which come from other sources.
The intensities of the assumed target peaks are summed up as the target intensity.
Using this information it calculates an intensity ratio for the relative intensity of the target
compared to other sources.
These metrics are combined over the previous and the next MS1 spectrum.

Includes example use case in OpenPepXL/LF.


src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.h
```diff
@@ -72,7 +72,7 @@ namespace OpenMS
        * @param labeled
        * @return A PeakMap of preprocessed spectra
        */
-      static PeakMap preprocessSpectra(PeakMap& exp, double fragment_mass_tolerance, bool fragment_mass_tolerance_unit_ppm, Size peptide_min_size, Int min_precursor_charge, Int max_precursor_charge, bool deisotope, bool labeled);
+      static PeakMap preprocessSpectra(PeakMap& exp, double fragment_mass_tolerance, bool fragment_mass_tolerance_unit_ppm, Size peptide_min_size, Int min_precursor_charge, Int max_precursor_charge, std::vector<Size>& discarded_spectra, bool deisotope, bool labeled);
```
hroest: please adjust the pyOpenMS bindings, see 
```
  ^
pyopenms/pyopenms_3.cpp: In function 'PyObject* __pyx_pf_8pyopenms_10pyopenms_3_32OPXLSpectrumProcessingAlgorithms_6preprocessSpectra(__pyx_obj_8pyopenms_10pyopenms_3_OPXLSpectrumProcessingAlgorithms*, __pyx_obj_8pyopenms_10pyopenms_3_MSExperiment*, double, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*)':
pyopenms/pyopenms_3.cpp:152676:246: error: no matching function for call to 'OpenMS::OPXLSpectrumProcessingAlgorithms::preprocessSpectra(boost::shared_ptr<OpenMS::MSExperiment>::element_type&, double, bool, size_t, int, int, bool, bool)'
     __pyx_t_11 = __pyx_v_self->inst.get()->preprocessSpectra((*__pyx_v_exp->inst.get()), ((double)__pyx_v_fragment_mass_tolerance), ((bool)__pyx_t_5), ((size_t)__pyx_t_6), ((int)__pyx_t_7), ((int)__pyx_t_8), ((bool)__pyx_t_9), ((bool)__pyx_t_10));
                                                                                                                                                                                                                                                      ^
pyopenms/pyopenms_3.cpp:152676:246: note: candidate is:
In file included from pyopenms/pyopenms_3.cpp:856:0:
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.h:75:22: note: static OpenMS::PeakMap OpenMS::OPXLSpectrumProcessingAlgorithms::preprocessSpectra(OpenMS::PeakMap&, double, bool, OpenMS::Size, OpenMS::Int, OpenMS::Int, std::vector<long unsigned int>&, bool, bool)
       static PeakMap preprocessSpectra(PeakMap& exp, double fragment_mass_tolerance, bool fragment_mass_tolerance_unit_ppm, Size peptide_min_size, Int min_precursor_charge, Int max_precursor_charge, std::vector<Size>& discarded_spectra, bool deisotope, bool labeled);
                      ^
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.h:75:22: note:   candidate expects 9 arguments, 8 provided

```

src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLDataStructs.h
```diff
@@ -167,6 +167,12 @@ namespace OpenMS
 
         int precursor_correction = 0;
 
+        double precursor_total_intensity = 0;
+        double precursor_target_intensity = 0;
+        double precursor_signal_proportion = 0;
+        Size precursor_target_peak_count = 0;
+        Size precursor_residual_peak_count = 0;
```
hroest: please adjust the pyOpenMS bindings, see

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -0,0 +1,231 @@
...
+      isolated_window.push_back(*lower_it);
+      lower_it++;
+    }
+
+    // std::cout << "Isolation window peaks: " << isolated_window.size();
+    for (auto peak : isolated_window)
+    {
+      // std::cout << " | " << peak.getMZ();
+    }
+    // std::cout << std::endl;
```
hroest: can you remove / comment out that construct?

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+      // stop loop when new mz is outside the isolation window
+      // changes through the isotope index iso
+      if (next_peak > upper)
+      {
+        break;
+      }
+      int next_iso_index = isolated_window.findNearest(next_peak, precursor_tolerance_abs);
+      if (next_iso_index != -1)
+      {
```
hroest: it seems here you are only using the closest peak, this will not work well for profile data or poorly centroided data.

Why not simply sum up all intensity in `next_peak +/- tolerance` ?

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -0,0 +1,231 @@
...
+      // always increment iso to progress the loop
+      iso++;
+    }
+    // std::cout << std::endl;
+
+    // std::cout << "noise peaks: ";
+    double noise_intensity(0);
+    for (auto peak : isolated_window)
+    {
+      noise_intensity += peak.getIntensity();
```
hroest: this is incorrect, as you erased all noise on line 117 except the noise between the last isotopic peak and the end

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -0,0 +1,231 @@
...
+    {
+      rel_sig = target_intensity / total_intensity;
+    }
+
+    score.total_intensity = total_intensity;
+    score.target_intensity = target_intensity;
+    score.residual_intensity = noise_intensity;
+    score.signal_proportion = rel_sig;
+    score.target_peak_count = target_peak_count;
+    score.residual_peak_count = isolated_window.size();
```
hroest: same issue here?

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -0,0 +1,231 @@
...
+      iso++;
+    }
+    // std::cout << std::endl;
+
+    // std::cout << "noise peaks: ";
+    double noise_intensity(0);
+    for (auto peak : isolated_window)
+    {
+      noise_intensity += peak.getIntensity();
+      // std::cout << peak.getMZ() << " | ";
```
hroest: why not just use `total_intensity - target_intensity` ?

hroest:
overall looks good, please fix the Python issues and consider creating pyOpenMS bindings for the new class

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
enetz: yes, that is a problem I have already run into. For centroided data it seems to work better after running `HighResPrecursorMassCorrector` with the `nearest_peak' method. Will summing up the intensity within the tolerance window not also catch some noise? I am also not sure what tolerance to use here in general, for now I use the precursor search tolerance from my search engine * 2, because the 10 ppm I usually use for the search engine is too narrow.

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
enetz: that was mainly a double check that all my numbers sum up to the correct totals.
you are right, in the end we only need the total intensity and the target intensity,
everything else can be calculated from that.
I will fix that

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
hroest: > Will summing up the intensity within the tolerance window not also catch some noise?

isnt that the point? 

> I am also not sure what tolerance to use here in general,

that should be the *resolution* of the instrument, e.g. how wide the peak is. Your 10 ppm is for the mass accuracy and there is a big difference (peak apex vs peak width)

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -0,0 +1,231 @@
...
+        // std::cout << "MS1 Spectrum: " << spectra[current_parent_index].getNativeID() << " | MS2 : " << spectra[i].getNativeID() << std::endl;
+        PrecursorPurity::PurityScores score1 = PrecursorPurity::computePrecursorPurity(spectra[current_parent_index], spectra[i].getPrecursors()[0], precursor_mass_tolerance, precursor_mass_tolerance_unit_ppm);
+        // use default values of 0, if there is an MS2 spectrum without an MS1 after it (may happen for the last group of MS2 spectra)
+        PrecursorPurity::PurityScores score2;
+        if (!lastMS1) // there is an MS1 after this MS2
+        {
+          score2 = PrecursorPurity::computePrecursorPurity(spectra[next_parent_index], spectra[i].getPrecursors()[0], precursor_mass_tolerance, precursor_mass_tolerance_unit_ppm);
+        }
+        PrecursorPurity::PurityScores score = PrecursorPurity::combinePrecursorPurities(score1, score2);
+        purityscores.push_back(score);
```
hroest: how do you link the scores to the MS spectra? It seems it would make sense to have an index / identifier to link the two?

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -0,0 +1,231 @@
...
+    Size next_parent_index = 0;
+    bool lastMS1(false);
+    for (Size i = 0; i < spectra.size(); ++i)
+    {
+      // change current parent index if a new MS1 spectrum is reached
+      if (spectra[i].getMSLevel() == 1)
+      {
+        current_parent_index = i;
+      }
+      else if (spectra[i].getMSLevel() == 2)
```
hroest: have you thought about MS3 spectra?

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
timosachsenberg: I think it makes most sense for centroided data. Resolution also might vary with m/z so I am not sure how to integrate this information properly

hroest: should work when its in ppm

timosachsenberg: I thought the resolution is about distinguishing two peaks and not about the total peak width (there are probably multiple definitions) so I am not totally convinced this works well for profile data. The accuracy for precursor mass assignment might be different than the accuracy in the MS1 (that's probably why he needs the precursor mass correction step). It will work for centroided data though but probably lead to effects summing up peaks twice, overlapping windows etc. For now I would go with a centroided only solution until more experiments are done.

enetz: @hroest this part of the algorithm sums up the intensities of the "target peaks", meaning the precursor ion targeted for fragmentation and possible additional isotopic peaks. I sum these up here and remove the peaks from this sliced out spectrum. Everything else, all the peaks I can not match to isotopes of the precursor mass here, I consider as noise.
@timosachsenberg Yes I wrote this code mainly for centroided data, that is why I only consider the one closest peak for each theoretical isotopic mass. If there are additional peaks within the tolerance window, they do not match the isotopic pattern and are considered as noise. I think making it work for profile data would require a very different approach.

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
enetz: true, currently they are linked only by their order, the 5th `PurityScore` belongs to the 5th MS2 spectrum in the map (not counting the MS1 spectra).
That is slightly annoying to deal with. But to simplify it I would rather need some kind of index in the `PeakMap` that points to the correct PrecursorPurity. In my search engine I have some filtering steps, that remove the MS1 spectra and some useless MS2 spectra after running this and I have to keep track of which spectra were discarded. So keeping a spectrum index in `PurityScore` would be useless. The most convenient solution for me would be to add this whole data structure to the `Precursor` class, but I don't want to add stuff there, that only a few people will need.

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
enetz: @hroest the function to do this for one spectrum is agnostic to the MS level. I guess one possibility would be to add the MS levels as input parameters to the looping function.

hroest:
it looks almost ready but on OSX there are 2 tests that consistently fail:

```
    projectroot.src.tests.topp.TOPP_RNPxlSearch_4_out2
    projectroot.src.tests.topp.TOPP_RNPxlSearch_4_out1
```

maybe @timosachsenberg has some idea what is happening there?

timosachsenberg:
I preprared a fix in https://github.com/OpenMS/OpenMS/pull/3726
Reason were two false (rank > 1) spectra that got the same score. Depending on the sorting algorithm on the different platforms one or the other was reported.

hroest:
> I preprared a fix in #3726

ok, so lets merge here. From my side it looks good and we can merge.

PR #3719, opened 2018-09-05
Fix/add protein ids

timosachsenberg:


timosachsenberg:
minor fix to add back the protein id to the (kind of deprecated) tsv output

src/openms/source/ANALYSIS/RNPXL/RNPxlReport.cpp
```diff
@@ -107,7 +107,7 @@ namespace OpenMS
         sl << String(ma.first + "_" + ma.second[i].first);
       }
     }
-    sl << "abs prec. error Da" << "rel. prec. error ppm" << "M+H" << "M+2H" << "M+3H" << "M+4H" << "fragment_annotation" << "rank";
+    sl << "abs prec. error Da" << "rel. prec. error ppm" << "M+H" << "M+2H" << "M+3H" << "M+4H" << "rank";
```
enetz: @timosachsenberg wasn't this in another PR, that you closed without merging recently? (Just to make sure this was on purpose.) Also, are you just removing the column header without changing the actual output in the following rows, or did you do that somewhere else? Because this seems unrelated to the other change in this PR.

timosachsenberg:
replaced by #3720

PR #3724, opened 2018-09-07
[FEATURE] disable nested parallelism

hroest:
This disables nested parallelism for OpenSwath (at least for now) due to some segfaults that were observed on a collaborators machine.


hroest:
not sure whether this is needed, the issues may potentially come from using two different OpenMP versions

hroest:
however, removing the nested parallelism solved the problem for the user, so it may be a good idea to offer this as a cmake configuration option

timosachsenberg:
Hmm, so can we be sure that this is really related to the OpenMP options for nested parallelism? Otherwise, it just might be a bug for which I would not like to add a cmake option. Probably unrelated, but we also get some ASan erros in OpenSWATH http://cdash.openms.de/viewTest.php?onlyfailed&buildid=285589 .

cbielow:
I also see segfaults on OpenSWATH from time to time, usually in very weird places (e.g. mzMLHandler). It might be worth investigating in detail. Not sure what the reason is. Might be usage of raw pointers...

hroest:
> Hmm, so can we be sure that this is really related to the OpenMP options for nested parallelism?

yes, because when compiled with this patch, the issue goes away. The person compiled OpenMS on a VM and then transferred the binaries to another system (CentOS 6 which does not support C++11 natively) so that could explain it. 

> we also get some ASan erros in OpenSWATH

these are related to the recent ion mobility extraction, and they are not ASan errors but precondition failures -- I will look into those

> I also see segfaults on OpenSWATH from time to time, usually in very weird places (e.g. mzMLHandler). It might be worth investigating in detail.

are you talking about the issues with destructors on Windows? On Linux we are pretty much battle-tested, we just had a person run over 50k instances of OpenSWATH using different parameters and they did find a few problems which we fixed

cbielow:
I'm running a debug build on Windows right now, to dig a bit deeper.
One thing that came up already is
` User Error 1001: omp_set_num_threads should only be called in serial regions`.
That might explain some things...

cbielow:
the crash I observe goes away when reverting back to 1 thread

hroest:
> the crash (segfault in dtors) I observed earlier goes away when reverting back to 1 thread

but I guess that is unrelated to this PR and nested parallelism?

cbielow:
> 
> 
> > `User Error 1001: omp_set_num_threads should only be called in serial regions`.
> 
> that is not what I found in this documentation https://docs.oracle.com/cd/E19059-01/stud.10/819-0501/2_nested.html

Visual Studio unfortunately only supports OpenMP2.0, which has no support for nested parallelism
https://msdn.microsoft.com/en-us/en-en/library/d8wkzt26.aspx
(calling omp_set_num_threads in parallel regions has undefined behaviour) :(

hroest:
>  (calling omp_set_num_threads in parallel regions has undefined behaviour) :(

so we could either make the nested parallelism configurable at build time or we could move the `omp_set_num_threads` call into an `if` clause to make sure it is not called when `threads_outer_loop_ < 0`

hroest:
@cbielow ready for merge? It would fix Windows builds

cbielow:
so, to enable nesting, you need to modify the `ENABLE_OPENMS_NESTED_PARALLELISM` flag in `OpenSwathWorkflow.cpp` now?
Fine with me.

hroest:
> so, to enable nesting, you need to modify the `ENABLE_OPENMS_NESTED_PARALLELISM` flag in `OpenSwathWorkflow.cpp` now?

yes, I will look into making this a cmake flag in the future

PR #3726, opened 2018-09-11
fix RNPxlTest

timosachsenberg:
- fix tests related to sorting of equal scored spectra.
+ experimental feature to constrain nucleotides


hroest:
looks good, travis had issues with sirius again

src/openms/source/ANALYSIS/RNPXL/RNPxlModificationsGenerator.cpp
```diff
@@ -343,12 +344,26 @@ RNPxlModificationMassesResult RNPxlModificationsGenerator::initModificationMasse
...
         continue;
       }
 
+      // check if nucleotides from more than one nt_group are present (e.g. from DNA and RNA)
+      Size found_in_n_groups(0);
+      for (const String & n : nt_groups)
+      { 
+        if (nucleotide_style_formula.find_first_of(n) != string::npos) { ++found_in_n_groups; }
+      }
+      // nucleotide stile formula (e.g. AATU matches to more than one group (e.g., RNA and DNA))?
```
enetz: spelling error: "stile"

src/utils/RNPxlSearch.cpp
```diff
@@ -188,6 +188,13 @@ class RNPxlSearch :
                         false, 
                         false);
 
+    registerStringList_("RNPxl:nt_groups",
+        "",
+        {},
+	"Restrict which nucleotides can cooccur in a precursor adduct to be able to search both RNA and DNA (Formate e.g.: AU CG).",
```
hroest: can you explain what for example "AU CG" would actually do? I am still not sure what it does, you then search for "AU" and "CG" in the result?

timosachsenberg: No it means that precursor adduced may be composed of As and Us or alternatively Cs and Gs. But no mix between both e.g. AAC-H2O

PR #3729, opened 2018-09-12
[Feature] SiriusAdapter converter_mode, sirius_workspace_directory 

oliveralka:
As a wish by the community, I would like to add a converter_mode.
Here the OpenMS preprocessed ".ms" file can be exported for further validation and use in other pipelines. 

Additionally, the sirius_workspace_directory can be copied to a new location, which will allow checking the findings in greater detail (e.g. Information about fragments). 
This will not hinder further implementation effort to write fragment information to the output mzTab. 

One question is if this feature should be retained in the initial SiriusAdapter implementation or transferred into a new tool.  
(I am not sure if the two EXECUTION_OK statements are alright - please see SiriusAdapter.cpp)


src/utils/SiriusAdapter.cpp
```diff
@@ -131,12 +131,16 @@ class TOPPSiriusAdapter :
...
     setValidFormats_("in_featureinfo", ListUtils::create<String>("featurexml"));
 
-    registerOutputFile_("out_sirius", "<file>", "", "MzTab Output file for SiriusAdapter results");
+    registerOutputFile_("out_sirius", "<file>", "", "MzTab Output file for SiriusAdapter results", false);
     setValidFormats_("out_sirius", ListUtils::create<String>("mzTab"));
 
     registerOutputFile_("out_fingerid","<file>", "", "MzTab output file for CSI:FingerID, if this parameter is given, SIRIUS will search for a molecular structure using CSI:FingerID after determining the sum formula", false);
     setValidFormats_("out_fingerid", ListUtils::create<String>("mzTab"));
 
+    registerOutputFile_("out_ms","<file>", "", "Internal SIRIUS .ms format after openMS preprocessing", false);
```
timosachsenberg: OpenMS

src/utils/SiriusAdapter.cpp
```diff
@@ -386,16 +405,42 @@ class TOPPSiriusAdapter :
...
+    if(!out_ms.empty())
+    {  
+      QFile::copy(tmp_ms_file.toQString(), out_ms.toQString());
+      LOG_WARN << "Preprocessed .ms files was moved to " << out_ms << std::endl; 
+    }
+
+
     // clean tmp directory if debug level < 2
-    if (debug_level_ >= 2)
+    if (debug_level_ >= 2 && out_ms.empty() && sirius_workspace_directory.empty())
```
timosachsenberg: why will it only be kept if out_ms == "" and sirius_workspace_directory == ""?
because it was already moved? please document this

oliveralka:
continuous exceeded time limit.

PR #3733, opened 2018-09-13
[FIX] reduce size of meta data

timosachsenberg:
Small change to reduce the size of meta info interface
before:
Memory usage (loading consensusXML): 2205 MB (working set delta)
FileInfo took 02:41 m (wall), 02:41 m (CPU), 2.10 s (system), 02:39 m (user).
after:
Memory usage (loading consensusXML): 1638 MB (working set delta)
FileInfo took 02:32 m (wall), 02:32 m (CPU), 1.72 s (system), 02:30 m (user).

### DIFF
--- a/src/openms/include/OpenMS/METADATA/MetaInfo.h
+++ b/src/openms/include/OpenMS/METADATA/MetaInfo.h
@@ -40,6 +40,8 @@
 #include <OpenMS/METADATA/MetaInfoRegistry.h>
 #include <OpenMS/DATASTRUCTURES/DataValue.h>
 
+#include <boost/container/flat_map.hpp>
+
 namespace OpenMS
 {
   class String;
@@ -119,7 +121,7 @@ public:
     void clear();
 
 private:
-    using MapType = std::map<UInt, DataValue>;
+    using MapType = boost::container::flat_map<UInt, DataValue>;
     /// Static MetaInfoRegistry
     static MetaInfoRegistry registry_;
     /// The actual mapping of indexes to values


hroest:
:+1: sounds great, 

I have also just looked at DataValue itself https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/DATASTRUCTURES/DataValue.h

```
DataValue 56 Bytes:
- 32 Bytes : String unit_;
- 4 Bytes: value_type_;
- 8 Bytes: data_;
- 8 Bytes: vtable
```

so it looks like

hroest:
I have just done some benchmarking on MetaInfo and DataValue using the following code:

```
  typedef std::map<UInt, DataValue> Map1Type;
  std::cout << " using map type 1" << std::endl;
  std::vector< Map1Type > v1;
  for (Size k = 0; k <= 1e7; k++) 
  {
    Map1Type tmp; 
    tmp[k] = DataValue(42.0);
    tmp[k+5] = DataValue(52.0);
    tmp[k+15] = DataValue(52.0);
    v1.push_back(tmp);
}
```

my findings so far using the vanilla implementation `typedef std::map<UInt, DataValue> Map1Type;` of the proposed implementation `typedef boost::container::flat_map<UInt, DataValue> Map1Type` is that I dont see a large difference in memory usage:

- `std::map`: 385.941504 Bytes for 3 entries [128.647168 bytes per entry]
- `boost::container::flat_map` : 391.29088 Bytes for 3 entries [130.4302 per entry]

However, this is quite shocking because we are storing a single float (8 bytes) and it takes us 130 bytes to store it, so there is an overhead of 16-fold! That is not really great and I started to look at bit more into `DataValue` and `MetaInfo`. I noticed that you can do two things immediately, namely remove the virtual destructor and replace the `String` with a `String*`

- 337.6234496 Bytes for 3 entries [ 112.54114 per entry] -- without vtable
- 230.20544 Bytes for 3 entries [ 76.7351466667 per entry] -- without vtable and string ptr

so this almost cuts the memory usage in half of the code above. Still there is a 10-fold overhead that remains, which could addressed by replacing the `flat_map` with a `std::pair`:

- 109.4606848 Bytes for 3 entries [ 36.4868949333 per entry] -- using a `std::pair`

this is now much closer to what we would expect with 24 bytes for the `DataValue` and 8 bytes for the `UInt` giving a total of 32 bytes

timosachsenberg:
ok it's interesting that it doesn't make a difference on that code but on the large consensusmap. could be related to memory fragmentation, different implementations of the maps (especially the grow strategie of the flat implementations) I guess. I tested it on other data as well and I am sure that I always observed less memory usage. Another reason could be that typically few meta values stored in many flat_maps need less space than few stored in the std::map.

timosachsenberg:
also, the flat map should actually be implemented as vector of pairs so I don't see why there is such an overhead

hroest:
> also, the flat map should actually be implemented as vector of pairs so I don't see why there is such an overhead

sorry, I think there was something wrong with my benchmark code that I have now fixed. I think it had to do with the `vector` growing larger than it should and doing extra allocations. I have fixed this now and updated the benchmark, and I consistently see lower memory with the flat_map

PR #3734, opened 2018-09-13
Swath QC

cbielow:
first of many to come :)

- number of missed cleavages for each transition (group)
- report FWHM in RT of each transition
- report m/z error in ppm for each transition

fixed some other minor/major issues:
+ PeakIntegrator segfaults on degenerate input data
+ reading TSV files from other platforms (with different line endings) does not throw away data anymore
+ invalid Visual Studio solution due to name clashes (duplicate DataStructures_test)
+ no more magic numbers for indices into floatdataarrays in Swath


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathTSVWriter.h
```diff
@@ -106,7 +106,7 @@ namespace OpenMS
      * @note Only call inside an OpenMP critical section
      *
      */
-    void writeLines(std::vector<String> to_output);
+    void writeLines(const std::vector<String>& to_output);
```
hroest: thanks :+1:

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -942,6 +946,7 @@ namespace OpenMS
       if (pep->isPeptide())
       {
         pep_hit_.setSequence(AASequence::fromString(pep->sequence));
+        mrmfeature->setMetaValue("missedCleavages", pd.peptideCount(pep_hit_.getSequence()));
```
hroest: I dont think this is a great idea, since this basically only measures the library cleavages -- actually I was planning to remove the peptide hits from here.

cbielow: yeah sure, one can only see MCs if the library contains them. But OpenSWATH is targeted after all, so I don't see a way around that. And if the library contains the MC's then there should be no problem.

hroest: as we discussed, I think we would need a more targeted approach to detect MC, e.g. if you have a protein 
PEPTIDEKPEPAKPEPBK

then we would have to compare intensities of 
- PEPTIDEK
- PEPAK
- PEPBK
- PEPTIDEKPEPAK
- PEPAKPEPBK
- PEPTIDEKPEPAKPEPBK

across batches and get some information on trypsin efficiency and I think that would be more effective.

hroest:
looks like you forgot a few tests: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=appleclang-9.0.0,os_label=highsierra/2377/


```
    projectroot.src.tests.topp.UTILS_MRMTransitionGroupPicker_test_2_out
    projectroot.src.tests.topp.UTILS_MRMTransitionGroupPicker_test_1_out
    projectroot.src.tests.topp.TOPP_OpenSwathWorkflow_21_out1
```

cbielow:
something weird is happening here:
On some systems, the peptide sequence `pep->sequence` is empty, on others not:
https://github.com/cbielow/OpenMS/blob/SWATHQC/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp#L965
When running TOPP_OpenSwathWorkflow_21, the number of missed cleavages is either -1 (for empty peptide), or 0 (for sequence PEPTIDEA).
Any idea why the sequence differs between systems?

cbielow:
ok, got it.
Its a problem in TransitionTSVFile: when reading a file from another platform, the default getline() will not remove '\r', and hence the last header name (here `PeptideSequence`) will be `PeptideSequence\r`. The subsequent parsing of all lines, ignores the column.

Fix comping up.

src/tests/class_tests/openms/source/PeakIntegrator_test.cpp
```diff
@@ -925,6 +942,28 @@ START_SECTION(PeakShapeMetrics calculatePeakShapeMetrics(
...
+  // corner cases: just a single point in range
+  pos = pit.findPosAtPeakHeightPercent(spectrum.begin(), spectrum.begin() + 1, spectrum.begin() + 1, 0.0, 0.0, false);
+  TEST_EQUAL(pos, spectrum[0].getPos()); // return the only peak there is
+  pos = pit.findPosAtPeakHeightPercent(spectrum.begin(), spectrum.begin() + 1, spectrum.begin() + 1, 0.0, 0.0, true);
+  TEST_EQUAL(pos, spectrum[0].getPos()); // return the only peak there is
+
+
+  // corner cases: empty range
+  TEST_EXCEPTION(Exception::InvalidRange, pit.findPosAtPeakHeightPercent(spectrum.end(), spectrum.end(), spectrum.end(), 0.0, 0.0, false))
+  TEST_EXCEPTION(Exception::InvalidRange, pit.findPosAtPeakHeightPercent(spectrum.end(), spectrum.end(), spectrum.end(), 0.0, 0.0, true))
```
hroest: @cbielow does this fix the bug?

src/tests/topp/MRMTransitionGroupPicker_1_output.featureXML
```diff
@@ -372,7 +372,7 @@
 					<UserParam type="string" name="native_id" value="60349_NSFVSYENFYQNHENAEIR/3_y16_2"/>
 					<UserParam type="float" name="peak_apex_int" value="50"/>
 					<UserParam type="float" name="total_xic" value="6777"/>
-					<UserParam type="float" name="width_at_50" value="6.82999999999993"/>
+					<UserParam type="float" name="width_at_50" value="3.42000000000007"/>
```
hroest: are these values now correct?

hroest: the feature goes from 

```
<UserParam type="float" name="leftWidth" value="4152.9599609375"/>
<UserParam type="float" name="rightWidth" value="4166.6201171875"/>
```

which are 13 seconds, so it seems like the `width_at_50` for the other ones is simply the full width of the feature except for this one feature here. Have you looked at the feature?

src/tests/topp/MRMTransitionGroupPicker_1_output.featureXML
```diff
@@ -194,6 +205,7 @@
 					<UserParam type="string" name="native_id" value="60352_NSFVSYENFYQNHENAEIR/3_y10"/>
 					<UserParam type="float" name="peak_apex_int" value="403"/>
 					<UserParam type="float" name="total_xic" value="4746"/>
+					<UserParam type="float" name="width_at_50" value="17.0700000000002"/>
```
hroest: ok, this is the "real" peak, and it makes a ton of sense, see 
screenshot from 2018-09-28 12-17-25

src/tests/topp/MRMTransitionGroupPicker_1_output.featureXML
```diff
@@ -138,6 +145,7 @@
 					<UserParam type="string" name="native_id" value="60348_NSFVSYENFYQNHENAEIR/3_y15_2"/>
 					<UserParam type="float" name="peak_apex_int" value="2110"/>
 					<UserParam type="float" name="total_xic" value="15604"/>
+					<UserParam type="float" name="width_at_50" value="13.6500000000001"/>
```
hroest: I also measured exactly 13.6, see 

screenshot from 2018-09-28 12-17-25

src/tests/class_tests/openms/source/PeakIntegrator_test.cpp
cbielow: yes, exactly

src/tests/topp/MRMTransitionGroupPicker_1_output.featureXML
cbielow: I looked at a few examples, and it was always on spot (given that it does not interpolate, but reports multiples of the swath-cycle time)

PR #3735, opened 2018-09-13
[FIX] OpenSWATH: Disabling transition-level MI scores

grosenberger:
This PR provides a hotfix to prevent a segmentation fault for OpenSwathWorkflow when transition-level scoring is enabled but the MI scores are disabled.


hroest:
looks good, what was the problem?

grosenberger:
``su_.use_mi_score_`` enables/disables upstream scoring and we then later try to access undefined data structures.

PR #3744, opened 2018-09-18
[FEATURE] save memory for MetaValues

hroest:
This is an extension of  #3733:

# Rationale 

1. Uses a `boost::container::flat_map` as in #3733 which is more efficient for many small maps
2. Reduces `DataValue` size:

```
DataValue 56 Bytes before:
- 32 Bytes : String unit_;
- 8 Bytes: value_type_;
- 8 Bytes: data_;
- 8 Bytes: vtable
DataValue 16 Bytes after:
- 1 Bytes: value_type_;
- 1 Bytes: unit_type_;
- 4 Bytes: unit_;
- 8 Bytes: data_;
```

# Benchmarking

Benchmarking on MetaInfo and DataValue using the following code:

```
  typedef std::map<UInt, DataValue> Map1Type;
  std::cout << " using map type 1" << std::endl;
  std::vector< Map1Type > v1;
  v1.reserve(1e7);
  for (Size k = 0; k < 1e7; k++) 
  {
    Map1Type tmp; 
    tmp[k] = DataValue(42.0);
    tmp[k+5] = DataValue(52.0);
    tmp[k+15] = DataValue(52.0);
    v1.push_back(tmp);
}
```

yields that in a comparison of the vanilla implementation `typedef std::map<UInt, DataValue> Map1Type;` against `typedef boost::container::flat_map<UInt, DataValue> Map1Type` :

- `std::map`: 385.941504 Bytes for 3 entries [128.647168 bytes per entry]
- `boost::container::flat_map` : 234.0593664 Bytes for 3 entries [78.01978 per entry]

So that means to store a single float (8 bytes) and takes 130 bytes, so there is an overhead of 16-fold! That is not really great and I started to look at bit more into `DataValue` and `MetaInfo`. I noticed that you can do two things immediately, namely remove the virtual destructor and replace the `String` with a `String*`

- 202.0610048 Bytes for 3 entries [ 78.034944 per entry] -- without vtable in `DataValue`
- 138.0769792 Bytes for 3 entries [ 46.0256597333 per entry] -- without vtable and string ptr in `DataValue`

so this substantially cuts the memory usage. There is of course still overhead, but the theoretical size now fits pretty well `3*(8+24) + 24 = 120` where the flat map needs to store 24 bytes for the DataValue and 8 bytes for the integer and each 3 objects require a full vector.

So overall, this should cut down memory usage by 2.8-fold, from 385.941504 to 138.0769792. Of course the downside is that for metavalues that contain units there is an extra level of indirection, but I dont think there are that many (actually the feature is rarely used except in a few places in the mzML and mzIdentML parser).

# Bugfixes / Features

While working on this I also noticed several issues with handling of units and I fixed them in most locations, specifically

- comparisons of `DataValue` now also compare the unit itself and the unit type
- we now also carry units also for `userParam` values which were lost before
- for mzIdentML we now only write the accession number and not the unit name any more


src/tests/class_tests/openms/source/DataValue_test.cpp
```diff
@@ -623,10 +624,15 @@ START_SECTION((void setUnit(const String& unit)))
...
   TEST_EQUAL(a1.getUnit(), "ppm")
+  auto tmpref = a1.getUnit();
 
   a1.setUnit("kg");
   TEST_EQUAL(a1.getUnit(), "kg")
 
+  a1.setUnit(a1.getUnit());
+  TEST_EQUAL(a1.hasUnit(), true)
+  TEST_EQUAL(a1.getUnit(), "kg")
+  // note: tmpref has now been deleted (access is illegal)
```
hroest: this may be the downside of the PR: it is now possible to invalidate references to the unit due to manual memory management

hroest:
Benchmarking on some real data

```
import pyopenms
file = "./src/tests/topp//MzTabExporter_1_input.consensusXML"

c = pyopenms.ConsensusMap()
pyopenms.ConsensusXMLFile().load(file, c)

all_f = [cc for cc in c]

c2 = pyopenms.ConsensusMap()
for f in all_f:
  for k in range(50):
    fq = pyopenms.ConsensusFeature(f)
    fq.clearUniqueId()
    fq.ensureUniqueId()
    c2.push_back(fq)

pyopenms.ConsensusXMLFile().store("/tmp/large.50.no.consensusXML", c2)
c2.setProteinIdentifications( c.getProteinIdentifications() )
pyopenms.ConsensusXMLFile().store("/tmp/large.50.consensusXML", c2)
```

And these are the results:

```
./bin/FileInfo -in /tmp/large.50.consensusXML
Memory usage (loading consensusXML): 807 MB (working set delta)

## using flat_map -> 81%
Memory usage (loading consensusXML): 661 MB (working set delta)

# using String ptr -> 73%
Memory usage (loading consensusXML): 591 MB (working set delta)

# using no vtable -> 70%
Memory usage (loading consensusXML): 569 MB (working set delta)
```

so overall a 30% reduction in total memory size from 800 MB to 569 MB.

src/openms/include/OpenMS/DATASTRUCTURES/DataValue.h
```diff
@@ -116,7 +116,7 @@ namespace OpenMS
     /// copy constructor
     DataValue(const DataValue&);
     /// destructor
-    virtual ~DataValue();
+    ~DataValue();
```
hroest: I dont think that people should use DataValue polymorphically or even derive directly from it at all.

timosachsenberg:
I just checked, unit.obo only has < 400 entries so two bytes would already be sufficient
to store the integer id:
https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/CV/unit.obo#L1523
for the value_type one byte would be sufficient. 
So this could in principle be squeezed into 16 bytes comfortably.

timosachsenberg:
@hroest great! One thing I dislike is the way the unit pointer is managed so I would strongly vote for a different solution (i.e., using a unit obo int).

hroest:
>  So this could in principle be squeezed into 16 bytes comfortably.

yes, depending on the memory layout it fits nicely into 16 bytes:

```
class OPENMS_DLLAPI DataValue
{

  DataType value_type_;
  uint32_t unit_int;
  /// Space to store the data
  union
  { 
    SignedSize ssize_;
    double dou_;
    String* str_;
    StringList* str_list_;
    IntList* int_list_;
    DoubleList* dou_list_;
  } data_;
}
```

then `sizeof(DataValue)` is 16. Also note that the 8 bytes we gain are only about 23 MB in a ConsensusXML as shown above or basically each 8 bytes are about 3.5% of the total size in memory. 

So your idea would be to store the integer part of the `id: UO:0000171` in there, and then you need a separate lookup object (global object) to map those ids to names? The only issue that I can see with that is if we ever have a second ontology that is not `UO` but I dont think we should worry about that right now.

timosachsenberg:
Yes, I think it is sensible to look it up as it is seems only be used in a few file handlers

hroest:
> Yes, I think it is sensible to look it up as it is seems only be used in a few file handlers

we could even add an enum to refer to which unit table is used, but that is getting ahead of ourselves

hroest:
> Yes, I think it is sensible to look it up as it is seems only be used in a few file handlers

it turned out to be substantially more involved that I would have liked, but now its down to 16 bytes. It turns out that we do use 2 different ontologies, `UO` and `MS` for units. So that is a bit annoying but in 27cb5c7c81279ccc416d69c873bf3d40fc8e5522 I have now implemented the new DataValue and I am open for comments.

src/tests/class_tests/openms/source/DataValue_test.cpp
```diff
@@ -69,49 +70,59 @@ using namespace std;
...
 DataValue* dv_nullPointer = nullptr;
 START_SECTION((DataValue()))
-	dv_ptr = new DataValue;
+
+  // Just as a sanity check, the size of DataValue should be exactly 16 bytes
+  // on a 64 bit system:
+  // - 1 byte for the data type
+  // - 1 byte for the unit type
+  // - 4 bytes for the unit identifier (32bit integer)
+  // - 1 byte padding
```
hroest: note that 2 bytes would actually be sufficient for the identifier for `UO` but not for the `MS` ontology which all start with a 1, e.g. `MS:1000040 ! m/z` which means we have to represent values of 1 million and above

src/tests/topp/IDFileConverter_9_output.mzid
```diff
@@ -203,7 +203,7 @@
 					<cvParam accession="MS:1001143" cvRef="PSI-MS" name="PSM-level search engine specific statistic"/>
 					<userParam name="MOWSE" unitName="xsd:double" value="1.4"/>
 				</SpectrumIdentificationItem>
-				<cvParam accession="MS:1000894" cvRef="PSI-MS" name="retention time" value="1234.5" unitAccession="UO:0000010" unitName="second" unitCvRef="UO"/>
+				<cvParam accession="MS:1000894" cvRef="PSI-MS" name="retention time" value="1234.5" unitAccession="UO:0000010" unitCvRef="UO"/>
```
hroest: here we dont have a solution (yet)

src/tests/class_tests/openms/source/DataValue_test.cpp
timosachsenberg: The 1 is just a padding number to ensure the number of digits

src/tests/topp/IDFileConverter_9_output.mzid
timosachsenberg: I used a slightly bigger consensusXML with more ID data.  
Before flat_map: 3297 MB
After flat_map: 2334 MB 
After slim data value: 1952 MB. 
I think this is a nice improvement. Also FileInfo is done from 3:00m to 2:40m

timosachsenberg: with #pragma pack(push, 1) 
and #pragma pack(pop)
I get: 1857 MB but 2:52m

hroest: > with #pragma pack(push, 1)

great for checking, I wonder though whether that is a good idea for a library? I am a bit worried whether that may lead to binary incompatibilities

>  I get: 1857 MB but 2:52m

so it actually takes longer as expected due to mis-alignment, however I dont understand how you can gain another 95 MB because it seems like you are almost saving another 10 bytes here - but how would that be possible on a struct of 16 bytes that only has 1 byte of padding?

Can you check the `sizeof(DataValue)` before and after the pragma?

hroest: > After flat_map: 2334 MB
> After slim data value: 1952 MB.
> After pack:  1857 MB but 2:52m

so I am wondering how you go from 2334 -> 1952 (382 MB) when going from 56 to 16 bytes but then you gain another 95 MB going from 16 to 14 / 12 where the math would predict that a gain of 95 MB should be another gain of 8 bytes.  Maybe it has something to do with how `vector` stores packed structs?

src/openms/source/DATASTRUCTURES/DataValue.cpp
```diff
@@ -676,7 +675,7 @@ namespace OpenMS
 
   bool operator==(const DataValue& a, const  DataValue& b)
   {
-    if (a.value_type_ == b.value_type_)
+    if (a.value_type_ == b.value_type_ && a.unit_type_ == b.unit_type_ && a.unit_ == b.unit_)
```
hroest: now we actually also compare the unit, which led to a lot of test failures since we *forgot* to carry the unit in `userParam`

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -539,7 +539,8 @@ namespace OpenMS
         String unitCvRef = XMLString::transcode(param->getAttribute(XMLString::transcode("unitCvRef")));
         String type = XMLString::transcode(param->getAttribute(XMLString::transcode("type")));
         DataValue dv;
-        dv.setUnit(unitAcc + ":" + unitName);
+        // dv.setUnit(unitAcc + ":" + unitName);
+        // TODO
```
hroest: I dont fully understand the DOM code and I am not sure whether this will be a problem or not.

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -1393,7 +1400,23 @@ namespace OpenMS
         }
       }
 
-      if (unit_accession != "") termValue.setUnit(unit_accession);
+      if (unit_accession != "")
+      {
+        if (unit_accession.hasPrefix("UO:"))
+        {
+          termValue.setUnit(unit_accession.suffix(unit_accession.size() - 3).toInt());
```
hroest: @timosachsenberg this should be pretty safe, I dont see a case for `MS:xxx` or `UO:xxx` where xxx is not a integer

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -566,6 +567,26 @@ namespace OpenMS
         {
           dv = value;
         }
+
+        // Add unit *after* creating the term
+        if (unitAcc != "")
```
hroest: this code actually used to drop the unit name since any assignment operator first clears the unit, so the code was broken before and should work now.

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -3085,26 +3110,56 @@ namespace OpenMS
         warning(LOAD, String("Unhandled cvParam '") + accession + "' in tag '" + parent_tag + "'.");
     }
 
-    void MzMLHandler::handleUserParam_(const String& parent_parent_tag, const String& parent_tag, const String& name, const String& type, const String& value)
+    void MzMLHandler::handleUserParam_(const String& parent_parent_tag,
+                                       const String& parent_tag,
+                                       const String& name,
+                                       const String& type,
+                                       const String& value,
+                                       const String& unit_accession)
```
hroest: userParam now also writes out unit accession

src/openms/include/OpenMS/FORMAT/HANDLERS/MzMLHandler.h
```diff
@@ -282,14 +282,18 @@ namespace OpenMS
...
                           const String& unit_accession = "");
 
       /// Handles user terms
-      void handleUserParam_(const String& parent_parent_tag, const String& parent_tag, const String& name, const String& type, const String& value);
+      void handleUserParam_(const String& parent_parent_tag,
+                            const String& parent_tag,
+                            const String& name,
+                            const String& type,
+                            const String& value,
+                            const String& unit_accession = "");
```
hroest: userParam now also writes out unit

src/tests/topp/IDFileConverter_9_output.mzid
hroest: > very cool. thanks! I will try that branch on the data and report back. So I guess the lookup is still missing in that part?

yes the lookup is missing and I am not sure that its very important.

hroest:
@timosachsenberg any more comments or are we ready to merge?

timosachsenberg:
looks good from my side. regarding the struct packing, I would have gone for that, too. Though I don't know if this somehow interferes with pyOpenMS or binary serialization that we have.

hroest:
> looks good from my side. regarding the struct packing, I would have gone for that, too.

not sure whether that is a good idea since it causes mis-aligned data structures which slows down performance. Also I am not sure about the implication of using this in a library and other code potentially linking to it which seems like a great place for strange errors - if its in a TOPP tool or internal struct, I would be ok but since its a public interface I feel a bit uneasy about it. Maybe we merge this and discuss the `pack` later?

jpfeuffer:
Hey @hroest . Can you have a look at the failing ASan tests here:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=287658

All the tests boil down to that "BaseLabeler" fails in mergeProteinIdentificationMaps_.
It might have something to do with copy constructing an object that has a MetaInfoInterface (here a ProteinHit).
So this might be related to the PR here.

hroest:
> Hey @hroest . Can you have a look at the failing ASan tests here:

I think this is already fixed, see https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp#L438

PR #3745, opened 2018-09-18
Swathqc - fixes carried over

hroest:
fixes from #3734


cbielow:
some unrelated failures, I guess. What about the py memcheck?

Issue #3310, opened 2018-03-15
Coding conventions discussion


hendrikweisser:
Prompted by a discussion with @timosachsenberg, I checked the Coding Conventions page on our wiki: https://github.com/OpenMS/OpenMS/wiki/Coding-conventions

There are some conventions there that are incomplete or don't agree with current best practices. We may have reasons to stick with them anyway, but I believe it would be good to have a discussion and possibly make changes.

Examples:
- The code [here](https://github.com/OpenMS/OpenMS/wiki/Coding-conventions#class-requirements) says to always check for self-assignment. The C++ FAQ has a more nuanced view [here](https://isocpp.org/wiki/faq/assignment-operators#self-assignment-how) - always *handle* self-assignment, but avoid the explicit check if possible (because self-assignments happen very rarely).
- For destructors (same section), the general recommendation is to make them virtual if a class has other virtual members: https://isocpp.org/wiki/faq/virtual-functions#virtual-dtors
- There is no mention in this section of when member functions can be automatically generated by the compiler - see [here](http://www.enyo.de/fw/notes/cpp-auto-members.html). Personally, I think we should avoid implementing functions that can be auto-generated, but the convention seems to suggest otherwise. No mention of using `= default` or `= delete` (C++11) either.

Please feel free to add other points for discussion as well.

timosachsenberg:
we could just say: if possible follow the https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md :)

hendrikweisser:
@timosachsenberg: We should include a link to them, but they are too long and in-depth to point a new contributor to and say: "Here, read this before writing code for OpenMS."

hroest:
> The code here says to always check for self-assignment. The C++ FAQ has a more nuanced view here - always handle self-assignment, but avoid the explicit check if possible (because self-assignments happen very rarely).

I dont have a strong view on that one

> For destructors (same section), the general recommendation is to make them virtual if a class has other virtual members: https://isocpp.org/wiki/faq/virtual-functions#virtual-dtors

I agree, we should add that - I wonder which should be the default (virtual or not)

> There is no mention in this section of when member functions can be automatically generated by the compiler - see here. Personally, I think we should avoid implementing functions that can be auto-generated, but the convention seems to suggest otherwise. No mention of using = default or = delete (C++11) either.

I think the coding conventions were written before C++11 so feel free to add that

hroest:
[...]

PR #3746, opened 2018-09-19
Poshul/hendrikweisser rna ms2

poshul:
NucleicAcidSearchEngine integration.


timosachsenberg:
as I understood it, this PR is mainly intended for early review and installer creation via our build system

poshul:
I've gotten rid of THRASH (which was a remnant of another project), and reverted the adducts files to match the develop branch.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -0,0 +1,556 @@
...
+                  double value);
+
+
+    /*!
+      @brief Set a data processing step that will apply to all subsequent "register..." calls.
+
+      This step will be appended to the list of processing steps for all relevant elements that are registered subsequently (unless it is already the last entry in the list).
+      If a score type without a software reference is registered, the software reference of this processing step will be applied.
+
+      Effective until @ref clearCurrentProcessingStep() is called.
```
timosachsenberg: Could this be done stateless? Current design harbors some potential for hard to debug code.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -0,0 +1,556 @@
...
+    */
+    ProcessingStepRef getCurrentProcessingStep();
+
+    /// Cancel the effect of @ref setCurrentProcessingStep().
+    void clearCurrentProcessingStep();
+
+    std::vector<QueryMatchRef> getBestMatchPerQuery(ScoreTypeRef
+                                                    score_ref) const;
+
+    std::pair<ScoreTypeRef, bool> findScoreType(const String& score_name) const;
```
timosachsenberg: in general more comments are required and a general overview

timosachsenberg:
I think peak annotations should be on the PSM level and not on the peptide level (or did I misunderstood sth?)

hroest:
you will also need to update the Python bindings...

src/tests/topp/CMakeLists.txt
```diff
@@ -2400,3 +2400,14 @@ add_test("UTILS_XFDR_4_out43" ${DIFF} -whitelist "date=" -in1 XFDR_test_out4_tem
...
+#------------------------------------------------------------------------------
+# NucleicAcidSearchEngine:
+add_test("UTILS_NucleicAcidSearchEngine_1" ${TOPP_BIN_PATH}/NucleicAcidSearchEngine -test
+-ini ${DATA_DIR_TOPP}/NucleicAcidSearchEngine_1.ini -in
+${DATA_DIR_TOPP}/NucleicAcidSearchEngine_1.mzML -id_out NucleicAcidSearchEngine_11_out.tmp -out NucleicAcidSearchEngine_12_out.tmp
+-database ${DATA_DIR_TOPP}/NucleicAcidSearchEngine_1.fasta)
+add_test("UTILS_NucleicAcidSearchEngine_11_out" ${DIFF} -in1 NucleicAcidSearchEngine_11_out.tmp -in2 ${DATA_DIR_TOPP}/NucleicAcidSearchEngine_11_out.idXML -whitelist "IdentificationRun date" "SearchParameters id=\"SP_0\" db=")
+add_test("UTILS_NucleicAcidSearchEngine_12_out" ${DIFF} -in1 NucleicAcidSearchEngine_12_out.tmp -in2 ${DATA_DIR_TOPP}/NucleicAcidSearchEngine_12_out.mzTab -whitelist "ms_run[1]-location")
+set_tests_properties("UTILS_NucleicAcidSearchEngine_11_out" PROPERTIES DEPENDS "UTILS_NucleicAcidSearchEngine_1")
+set_tests_properties("UTILS_NucleicAcidSearchEngine_12_out" PROPERTIES DEPENDS "UTILS_NucleicAcidSearchEngine_1")
```
hroest: can you also add tests for your other tool `RNAMassCalculator`?

src/tests/class_tests/openms/source/Ribonucleotide_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, 
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; 
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, 
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR 
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF 
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+// 
+// --------------------------------------------------------------------------
+// $Maintainer: Timo Sachsenberg $
+// $Authors: Andreas Bertsch $
```
hroest: can you replace the maintainer and author in all files?

src/tests/class_tests/openms/source/Ribonucleotide_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+START_TEST(Ribonucleotide, "$Id$")
+
+/////////////////////////////////////////////////////////////
+
+Ribonucleotide* e_ptr = nullptr;
+Ribonucleotide* e_nullPointer = nullptr;
+START_SECTION((Ribonucleotide()))
+  e_ptr = new Ribonucleotide();
+	TEST_NOT_EQUAL(e_ptr, e_nullPointer)
```
hroest: please replace all tabs with 2 whitespaces in all code

src/tests/topp/CMakeLists.txt
hroest: also for `DecoyDatabase`

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -0,0 +1,556 @@
...
+#include <OpenMS/METADATA/ID/ScoreType.h>
+#include <OpenMS/FORMAT/MzTab.h>
+#include <OpenMS/METADATA/PeptideIdentification.h>
+#include <OpenMS/METADATA/ProteinIdentification.h>
+
+#include <boost/unordered_set.hpp>
+
+namespace OpenMS
+{
+  class OPENMS_DLLAPI IdentificationData: public MetaInfoInterface
```
hroest: @hendrikweisser is this the new ID data structure people are talking about? If so, can you add some documentation on how it works and what it will do and how to use it?

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -0,0 +1,556 @@
...
+#include <OpenMS/METADATA/ID/MetaData.h>
+#include <OpenMS/METADATA/ID/MoleculeParentMatch.h>
+#include <OpenMS/METADATA/ID/MoleculeQueryMatch.h>
+#include <OpenMS/METADATA/ID/ParentMolecule.h>
+#include <OpenMS/METADATA/ID/ParentMoleculeGroup.h>
+#include <OpenMS/METADATA/ID/QueryMatchGroup.h>
+#include <OpenMS/METADATA/ID/ScoreType.h>
+#include <OpenMS/FORMAT/MzTab.h>
+#include <OpenMS/METADATA/PeptideIdentification.h>
+#include <OpenMS/METADATA/ProteinIdentification.h>
```
hroest: are all these includes required? I see at least no mention of `PeptideIdentification` or `ProteinIdentification` anywhere in the file.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -0,0 +1,556 @@
...
+    {
+    }
+
+    // Copy constructor - not allowed, as references would be invalidated:
+    // @TODO: implement using deep copy
+    IdentificationData(const IdentificationData& other) = delete;
+
+    /// Move constructor
+    IdentificationData(IdentificationData&& other):
+      current_step_ref_(other.current_step_ref_)
```
hroest: can you replace this with  
```
IdentificationData(IdentificationData&& other) noexcept = default;
```

src/openms/include/OpenMS/METADATA/ID/IdentifiedCompound.h
```diff
@@ -0,0 +1,87 @@
...
+      IdentifiedCompound(const IdentifiedCompound& other) = default;
+    };
+
+    // identified compounds indexed by their identifiers:
+    typedef boost::multi_index_container<
+      IdentifiedCompound,
+      boost::multi_index::indexed_by<
+        boost::multi_index::ordered_unique<boost::multi_index::member<
+          IdentifiedCompound, String, &IdentifiedCompound::identifier>>>
+      > IdentifiedCompounds;
```
hroest: I dont quite understand this, I thought the point of `boost::multi_index_container` was to index by `A` and also sort by `B` -- but here you only use the indexing part and not the second sorting part. Is that correct? Could this be replaced by a sorted list by `IdentifiedCompound::identifier` ?

src/openms/include/OpenMS/METADATA/ID/MetaData.h
```diff
@@ -0,0 +1,100 @@
...
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Hendrik Weisser $
+// $Authors: Hendrik Weisser $
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_METADATA_ID_METADATA_H
+#define OPENMS_METADATA_ID_METADATA_H
```
hroest: I dont think that MetaData.h is a great name, especially as the folder is already called METADATA

src/tests/topp/CMakeLists.txt
poshul: Done.

src/openms/include/OpenMS/METADATA/CVTerm.h
```diff
@@ -66,28 +66,24 @@ namespace OpenMS
...
-        name(rhs.name),
-        cv_ref(rhs.cv_ref)
-      {
-      }
+
+      /// Copy constructor
+      Unit(const Unit & rhs) = default;
+      
+      /// Move constructor
+      Unit(Unit&&) = default;
```
hroest: @hendrikweisser this should be 

```
       Unit(Unit&&) noexcept = default; 
```

src/openms/include/OpenMS/METADATA/Software.h
```diff
@@ -50,14 +50,18 @@ namespace OpenMS
   {
 public:
     /// Constructor
-    explicit Software(const String& name = "", const String& version = "");
+    Software() = default;
     /// Copy constructor
-    Software(const Software& source) = default;
+    Software(const Software & source) = default;
+    /// Move constructor
+    Software(Software&&) = default;
```
hroest: same here

src/openms/include/OpenMS/METADATA/Software.h
```diff
@@ -50,14 +50,18 @@ namespace OpenMS
...
     /// Destructor
     ~Software() override;
 
     /// Assignment operator
-    Software& operator=(const Software& source) = default;
+    Software & operator=(const Software & source) = default;
+    /// Move assignment operator
+    Software& operator=(Software&&) & = default;
```
hroest: @hendrikweisser we are currently debating here how to handle these "simple" cases https://github.com/OpenMS/OpenMS/issues/3310#issuecomment-424066542 and we are torn between your approach (Rule of Five) and the cleaner "Rule of Zero"

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -3381,6 +3382,144 @@ namespace OpenMS
       // ensure spectrum is drawn as sticks
       draw_group_1d_->button(Spectrum1DCanvas::DM_PEAKS)->setChecked(true);
       setDrawMode1D(Spectrum1DCanvas::DM_PEAKS);
+
+  //      }
+  /*
```
hroest: should this be in here?

src/openms/source/METADATA/ID/sources.cmake
```diff
@@ -0,0 +1,20 @@
+### the directory name
+set(directory source/METADATA/ID)
+
+### list all filenames of the directory here
+set(sources_list
+IdentificationData.cpp
+IdentificationDataConverter.cpp
```
hroest: where are all the cpp files? are all the ID structures header-only?

src/utils/NucleicAcidSearchEngine.cpp
```diff
@@ -0,0 +1,1171 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2017.
```
hroest: can you change all of these to 2018 ?

src/openms/include/OpenMS/CHEMISTRY/Ribonucleotide.h
```diff
@@ -0,0 +1,196 @@
...
+#include <OpenMS/CHEMISTRY/EmpiricalFormula.h>
+#include <iostream>
+
+namespace OpenMS
+{
+  /** @ingroup Chemistry
+
+      @brief Representation of a ribonucleotide (modified or unmodified)
+
+      The available information is based on the Modomics database (http://modomics.genesilico.pl/modifications/).
```
hroest: comment seems to be from DB, please change

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
```diff
@@ -0,0 +1,125 @@
...
+
+#include <OpenMS/CHEMISTRY/Ribonucleotide.h>
+#include <OpenMS/DATASTRUCTURES/String.h>
+#include <unordered_map>
+
+namespace OpenMS
+{
+  /** @ingroup Chemistry
+
+      @brief Database of ribonucleotides (modified and unmodified)
```
hroest: maybe you can comment on how this works, are modified and unmodified entries treated the same ? is there  a link between them?

hroest:
@hendrikweisser so how do the new data structures work? Can you load/store files in idXML or mzIdentML format? Do you convert them to the old PeptideIdentification / ProteinIdentification structures first? 

Have you done any benchmarking in terms of speed / memory compared to the old structures? Do you expect them to have lower overhead?

hroest:
great, seems like it compiles and only these tests fail:

```
    projectroot.src.tests.topp.UTILS_DecoyDatabase_4_out
    projectroot.src.tests.topp.UTILSWRITEINI_NucleicAcidSearchEngine_SectionName
    projectroot.src.tests.topp.UTILS_NucleicAcidSearchEngine_11_out
    projectroot.src.tests.topp.UTILS_NucleicAcidSearchEngine_12_out
    projectroot.src.tests.topp.UTILS_RNAMassCalculator_1_out
```

https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=appleclang-9.0.0,os_label=highsierra/2384/#showFailuresLink

src/openms/include/OpenMS/METADATA/ID/IdentifiedCompound.h
hendrikweisser: Yes, in principle it could be replaced with a set (I think), but the multi_index_container makes it easy to index by "identifier" and provides additional useful features (range retrieval, updating of elements).

src/openms/source/METADATA/ID/sources.cmake
hendrikweisser: Most of them currently are.

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
hendrikweisser: They are treated the same.

src/openms/include/OpenMS/CHEMISTRY/Ribonucleotide.h
hendrikweisser: I think having the reference here as well is useful.

hendrikweisser:
@timosachsenberg:
> I think peak annotations should be on the PSM level and not on the peptide level (or did I misunderstood sth?)

Yes, they are. They are stored in class `MoleculeQueryMatch`, which represents e.g. PSMs.

hendrikweisser:
@hroest:
> Can you load/store files in idXML or mzIdentML format? Do you convert them to the old PeptideIdentification / ProteinIdentification structures first?

You can store in idXML or mzIdentML via (imperfect) conversion to PeptideIdentification/ProteinIdentification, or directly in mzTab. (I've extended mzTab to handle RNA identification results.) The code is in class `IdentificationDataConverter`.
We will have to make new file formats to handle all this properly; the idea is to base them on SQLite, like you did for some of the OpenSWATH formats.

hendrikweisser:
> Have you done any benchmarking in terms of speed / memory compared to the old structures? Do you expect them to have lower overhead?

Yes, Timo insisted on it. ;-)
Here are the results:

> I've tested loading peptide/protein identification data from Julianus' example file (471 MB idXML with ~11,000 proteins, ~74,000 peptides, ~335,000 PSMs). Loading the data into the old "PeptideIdentification"/"ProteinIdentification" structures uses 1510 MB of memory. Importing from there into the new "IdentificationData" class currently takes only 247 MB - smaller by a factor of five.
> 
> I've added some look-up tables of memory addresses to help with validity checking of references in "IdentificationData" (for example, if you add a peptide that references a protein, that protein must be already known). That reduced the time for conversion from "PeptideIdentification"/"ProteinIdentification" to "IdentificationData" from 12 minutes down to 5 seconds on my server.
> 
> A problem that I haven't solved yet is the following: In "IdentificationData", spectra are identified by their ID ("spectrum_reference" in idXML) and input file. Julianus' example file is a merge of several files, so some spectrum IDs are repeated, but input file information isn't available. This means some spectra from different files get collapsed into one entry in "IdentificationData" and the different RT, m/z information is lost.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hroest: PS: it should probably be

```
       current_step_ref_(std::move(other.current_step_ref_) )
```

hroest:
> I've tested loading peptide/protein identification data from Julianus' example file (471 MB idXML with ~11,000 proteins, ~74,000 peptides, ~335,000 PSMs). Loading the data into the old "PeptideIdentification"/"ProteinIdentification" structures uses 1510 MB of memory. Importing from there into the new "IdentificationData" class currently takes only 247 MB - smaller by a factor of five.

since I wasnt at the retreat, can you quickly recap how that is achieved? Is it because links / identifiers are not strings any more? Or something else?

hroest:
@hendrikweisser could you try it on this file https://github.com/OpenMS/OpenMS/pull/3750#issuecomment-423726499 or provide me with the file that you used so that I can do some benchmarking of the current code as well?

src/openms/include/OpenMS/METADATA/ID/ScoredProcessingResult.h
```diff
@@ -0,0 +1,117 @@
...
+#include <OpenMS/METADATA/ID/ScoreType.h>
+
+namespace OpenMS
+{
+  namespace IdentificationDataInternal
+  {
+    /// Base class for ID data with scores and processing steps (and meta info)
+    struct ScoredProcessingResult: public MetaInfoInterface
+    {
+      ScoreList scores;
```
hroest: I dont think that this should be a public member, please make something like this private and give access through a clean interface (e.g. getScore / setScore)

that basically translates to `std::vector<std::pair< IteratorWrapper<std::set<ScoreType> ::iterator> , double>> ScoreList;`

src/openms/include/OpenMS/METADATA/ID/QueryMatchGroup.h
```diff
@@ -0,0 +1,96 @@
...
+    };
+
+    typedef boost::multi_index_container<
+      QueryMatchGroup,
+      boost::multi_index::indexed_by<
+        boost::multi_index::ordered_unique<
+          boost::multi_index::member<QueryMatchGroup, std::set<QueryMatchRef>,
+                                     &QueryMatchGroup::query_match_refs>>>
+      > QueryMatchGroups;
+    typedef IteratorWrapper<QueryMatchGroups::iterator> MatchGroupRef;
```
hroest: to me this is part of an implementation detail and should not be part of the public API. Also these containers will be hard to access in Python, therefore we should have a public interface that provides access to the objects and the implementation should be private and hidden.

src/openms/include/OpenMS/METADATA/ID/QueryMatchGroup.h
```diff
@@ -0,0 +1,96 @@
...
+    /*!
+      Group of related (co-identified) molecule-query matches
+
+      E.g. for cross-linking data or multiplexed spectra.
+    */
+    struct QueryMatchGroup: public ScoredProcessingResult
+    {
+      std::set<QueryMatchRef> query_match_refs;
+
+      bool allSameMolecule() const
```
hroest: I think this should go into the cpp file and not into the  header

src/openms/include/OpenMS/METADATA/ID/ParentMoleculeGroup.h
```diff
@@ -0,0 +1,84 @@
...
+{
+  namespace IdentificationDataInternal
+  {
+    /*!
+      Group of ambiguously identified parent molecules (e.g. protein group)
+    */
+    // @TODO: derive from MetaInfoInterface?
+    struct ParentMoleculeGroup
+    {
+      ScoreList scores;
```
hroest: I dont think that this should be a public member, please make something like this private and give access through a clean interface (e.g. getScore / setScore)

hroest:
> I've tested loading peptide/protein identification data from Julianus' example file (471 MB idXML with ~11,000 proteins, ~74,000 peptides, ~335,000 PSMs). Loading the data into the old "PeptideIdentification"/"ProteinIdentification" structures uses 1510 MB of memory. Importing from there into the new "IdentificationData" class currently takes only 247 MB - smaller by a factor of five.

ok, great, I god the same file from Julianus. Good News: I was able to compare it with the current trunk and using `FileInfo` there are indeed substantial savings:

```
Memory usage (loading idXML): 478 MB (working set delta)

-- General information --

File name: /media/doc/tmp/openms_issues/mergedIDs_Fido_qVals_noFilter_fdrNew.idXML
File type: idXML
Number of:
  runs:                       1
  protein hits:               11098
  non-redundant protein hits: 11098
  (only hits that differ in the accession)

  matched spectra:    328591
  peptide hits:               335250 (avg. length: 13)
  modified top-hits:          60640/328591 (18.5%)
  non-redundant peptide hits: 73950
  (only hits that differ in sequence and/or modifications)
  Modification count (top-hits only): Carbamidomethyl (C) 45463, Oxidation (M) 28777

FileInfo took 23.86 s (wall), 23.84 s (CPU), 0.67 s (system), 23.17 s (user).
```

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -0,0 +1,556 @@
...
+      query_matches_.swap(other.query_matches_);
+      query_match_groups_.swap(other.query_match_groups_);
+      swap(current_step_ref_, other.current_step_ref_);
+      // look-up tables:
+      data_query_lookup_.swap(other.data_query_lookup_);
+      parent_molecule_lookup_.swap(other.parent_molecule_lookup_);
+      identified_peptide_lookup_.swap(other.identified_peptide_lookup_);
+      identified_compound_lookup_.swap(other.identified_compound_lookup_);
+      identified_oligo_lookup_.swap(other.identified_oligo_lookup_);
+      query_match_lookup_.swap(other.query_match_lookup_);
```
hroest: I think the default should be `mvar( std::move(other.mvar) )` for all member variables `mvar` unless you have a good reason - `std::swap` will actually create a temporary object and is extra work.

https://stackoverflow.com/questions/9456910/how-to-define-a-move-constructor

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hroest: > Yes. @poshul opened the PR before this was ready!

so I assume that this is mainly to discuss some of the main concepts and give early feedback - not for merging?

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
```diff
@@ -0,0 +1,888 @@
...
+        peptide.processing_step_refs.push_back(step_ref);
+        for (const PeptideEvidence& evidence : hit.getPeptideEvidences())
+        {
+          const String& accession = evidence.getProteinAccession();
+          if (accession.empty()) continue;
+          IdentificationData::ParentMolecule parent(accession);
+          parent.processing_step_refs.push_back(step_ref);
+          // this will merge information if the protein already exists:
+          IdentificationData::ParentMoleculeRef parent_ref =
+            id_data.registerParentMolecule(parent);
```
hroest: do I understand this correctly that you store a list of proteins in `id_data` and here you just look up whether that protein already exists in the list, and if so you always reference the same molecule? Is that also how the memory savings are achieved?

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
```diff
@@ -0,0 +1,888 @@
...
+          IdentificationData::ProcessingStepRef sub_step_ref =
+            id_data.registerDataProcessingStep(sub_step);
+          match.processing_step_refs.push_back(sub_step_ref);
+          for (const pair<String, double>& sub_pair : ana_res.sub_scores)
+          {
+            IdentificationData::ScoreType sub_score;
+            sub_score.name = sub_pair.first;
+            sub_score.software_opt = software_ref;
+            IdentificationData::ScoreTypeRef sub_score_ref =
+              id_data.registerScoreType(sub_score);
```
hroest: the same for the scores, you keep a global list of the names of the scores and then you only store a `pair` of `<ref, double` to reference the score? I assume that is way more efficient than using the MetaValue ?

hroest: Is that is the case, I am wondering why not simply store a `std::pair<int, double>` where the `int` references the entry in a global list of scores. I believe the interface may be a bit cleaner, even though you introduce another set of indirection but there is less dependency on the actual type of `IdentificationData::ScoreTypeRef` and all you would need is a function `IdentificationData::getScore(int score_id)` to retrieve the score, its name etc. Also it may be even more memory efficient as an `int` is only 4 bytes whereas an iterator is usually the same size as a ptr = 8 bytes

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
```diff
@@ -0,0 +1,888 @@
...
+        IdentificationData::ParentMoleculeRef parent_ref = pair.first;
+        for (const IdentificationData::MoleculeParentMatch& parent_match :
+               pair.second)
+        {
+          PeptideEvidence evidence;
+          evidence.setProteinAccession(parent_ref->accession);
+          evidence.setStart(parent_match.start_pos);
+          evidence.setEnd(parent_match.end_pos);
+          evidence.setAABefore(parent_match.left_neighbor[0]);
+          evidence.setAAAfter(parent_match.right_neighbor[0]);
```
hroest: what happens if the string is empty? sounds like a nasty segfault is possible here!

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
```diff
@@ -0,0 +1,888 @@
...
+              psm_data[key].second = score_ref;
+              break;
+            }
+          }
+        }
+      }
+    }
+
+    set<IdentificationData::ProcessingStepRef> steps;
+    for (auto psm_it = psm_data.begin(); psm_it != psm_data.end(); ++psm_it)
```
hroest: why not simply 
```
for (const auto & psm_it : psm_data)
```

src/openms_gui/include/OpenMS/VISUAL/TOPPViewIdentificationViewBehavior.h
```diff
@@ -34,7 +34,7 @@
 
 #pragma once
 
-// #include <OpenMS/CHEMISTRY/NASequence.h>
+#include <OpenMS/CHEMISTRY/NASequence.h>
```
hroest: actually this could go into the cpp and you can forward-declare `class NASequence` here!

src/tests/class_tests/openms/source/IdentificationDataConverter_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  IdentificationData ids;
+  IdentificationDataConverter::importIDs(ids, proteins_in, peptides_in);
+
+  vector<ProteinIdentification> proteins_out;
+  vector<PeptideIdentification> peptides_out;
+  IdentificationDataConverter::exportIDs(ids, proteins_out, peptides_out);
+
+  TEST_EQUAL(peptides_in.size(), peptides_out.size());
+  TEST_EQUAL(proteins_in.size(), proteins_out.size());
```
hroest: can you please also test that they are equal, e.g. `TEST_EQUAL(proteins_in == proteins_out, true)` ?

src/tests/class_tests/openms/source/IdentificationDataConverter_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+  IdentificationDataConverter::importIDs(ids, proteins_in, peptides_in);
+
+  vector<ProteinIdentification> proteins_out;
+  vector<PeptideIdentification> peptides_out;
+  IdentificationDataConverter::exportIDs(ids, proteins_out, peptides_out);
+
+  // the additional Peptide-/InterProphet scores cause multiple ID runs upon
+  // export:
+  TEST_EQUAL(proteins_out.size(), 3);
+  TEST_EQUAL(peptides_out.size(), 4);
```
hroest: is that a problem? does this need to be addressed?

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+  // tested further below
+}
+END_SECTION
+
+START_SECTION((ProcessingSoftwareRef registerDataProcessingSoftware(const Software& software)))
+{
+  Software sw("Tool", "1.0");
+  sw_ref = data.registerDataProcessingSoftware(sw);
+  TEST_EQUAL(data.getDataProcessingSoftware().size(), 1);
+  TEST_EQUAL(*sw_ref == sw, true); // "TEST_EQUAL(*sw_ref, sw)" doesn't compile
```
hroest: can you also test if you register the same tool twice, you should get the same reference back (e.g. no double entries, right) ?

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+  peptide_ref = data.registerIdentifiedPeptide(peptide);
+  TEST_EQUAL(data.getIdentifiedPeptides().size(), 1);
+  TEST_EQUAL(*peptide_ref == peptide, true); // "TEST_EQUAL(*peptide_ref, peptide)" doesn't compile
+
+  // peptide with protein reference:
+  peptide.sequence = AASequence::fromString("PEPTIDE");
+  peptide.parent_matches[protein_ref];
+  peptide_ref = data.registerIdentifiedPeptide(peptide);
+  TEST_EQUAL(data.getIdentifiedPeptides().size(), 2);
+  TEST_EQUAL(*peptide_ref == peptide, true); // "TEST_EQUAL(*peptide_ref, peptide)" doesn't compile
```
hroest: what happens when you register the same peptide, once with and without protein reference? does it generate 2 entries?

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+  // registering a peptide with RNA reference doesn't work:
+  peptide.parent_matches[rna_ref];
+  TEST_EXCEPTION(Exception::IllegalArgument,
+                 data.registerIdentifiedPeptide(peptide));
```
hroest: can you add an example where it does work?

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+END_SECTION
+
+START_SECTION((const IdentifiedPeptides& getIdentifiedOligos() const))
+{
+  TEST_EQUAL(data.getIdentifiedOligos().empty(), true);
+  // tested further below
+}
+END_SECTION
+
+START_SECTION((IdentifiedPeptideRef registerIdentifiedOligo(const IdentifiedOligo& oligo)))
```
hroest: should be `     IdentifiedOligoRef registerIdentifiedOligo(const IdentifiedOligo& oligo); `

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+                 data.registerIdentifiedOligo(oligo));
+  TEST_EQUAL(data.getIdentifiedOligos().empty(), true);
+
+  // oligo without RNA reference:
+  oligo.sequence = NASequence::fromString("ACGU");
+  oligo_ref = data.registerIdentifiedOligo(oligo);
+  TEST_EQUAL(data.getIdentifiedOligos().size(), 1);
+  TEST_EQUAL(*oligo_ref == oligo, true); // "TEST_EQUAL(*oligo_ref, oligo)" doesn't compile
+
+  // oligo with protein reference:
```
hroest: protein?

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+END_SECTION
+
+START_SECTION((const IdentifiedPeptides& getIdentifiedCompounds() const))
+{
+  TEST_EQUAL(data.getIdentifiedCompounds().empty(), true);
+  // tested further below
+}
+END_SECTION
+
+START_SECTION((IdentifiedPeptideRef registerIdentifiedCompound(const IdentifiedCompound& compound)))
```
hroest: signature is wrong

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+{
+  TEST_EQUAL(data.getMoleculeQueryMatches().empty(), true);
+  // tested further below
+}
+END_SECTION
+
+START_SECTION((QueryMatchRef registerMoleculeQueryMatch(const MoleculeQueryMatch& match)))
+{
+  // match with a peptide:
+  IdentificationData::MoleculeQueryMatch match(peptide_ref, query_ref, 3);
```
hroest: for readability and testability, can you please make the test self-contained? This now references a peptide_ref that from a different test which is somewhat confusing, it would be better for documentation purposes and separability of the tests to generate a new peptide_ref and query_ref to see what is going on here.

src/tests/class_tests/openms/source/ModifiedNASequenceGenerator_test.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    NASequence sequence = NASequence::fromString("AUAUAUA");
+  vector<NASequence> ams;
+
+	// (1) Add at most one modification. (true) return the unmodified version
+  ModifiedNASequenceGenerator::applyVariableModifications(
+		var_mods.begin(),
+  	var_mods.end(),
+	  sequence,
+	  1,
```
hroest: can you fix whitespace?

src/tests/class_tests/openms/source/NASequence_test.cpp
```diff
@@ -0,0 +1,751 @@
...
+  NASequence c = NASequence::fromString("C");
+  c = NASequence::fromString("AAA");
+  TEST_EQUAL(c.size(), 3);
+}
+END_SECTION
+
+START_SECTION((NASequence(std::vector<const Ribonucleotide*> s, const RibonucleotideChainEnd* five_prime, const RibonucleotideChainEnd* three_prime)))
+{
+  NASequence aaa = NASequence::fromString("AAA");
+  NASequence aaa2(aaa.getSequence(), nullptr, nullptr);
```
hroest: can you simplify this by having another constructor `explicit NASequence(std::vector<const Ribonucleotide*> s,` ?

src/tests/class_tests/openms/source/NASequence_test.cpp
```diff
@@ -0,0 +1,751 @@
...
+END_SECTION
+
+START_SECTION((void setThreePrimeMod(const RibonucleotideChainEnd* r)))
+{
+  NASequence aaa = NASequence::fromString("AAA");
+  TEST_EQUAL(aaa.hasThreePrimeMod(), false);
+  aaa.setThreePrimeMod(db->getRibonucleotide("(pN)"));
+  TEST_EQUAL(aaa.hasThreePrimeMod(), true);
+  TEST_EQUAL(aaa.getThreePrimeMod()->getCode(), "(pN)");
+  TEST_STRING_EQUAL(aaa.toString(), "AAA[(pN)]");
```
hroest: can you explain the notation? In peptides we use `[]` to indicate mass differences not mapped to a specific modification. Does it make sense to use the same notation for a different purpose here?

src/tests/class_tests/openms/source/RibonucleotideDB_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+END_SECTION
+
+START_SECTION(virtual ~RibonucleotideDB())
+  NOT_TESTABLE
+END_SECTION
+
+START_SECTION(ConstIterator begin())
+{
+  RibonucleotideDB::ConstIterator it = ptr->begin();
+  TEST_STRING_EQUAL((*it)->getCode(), "m1Am");
```
hroest: does this make sense to test? doesnt this depend a lot on the file used to populate the database?

hroest:
Overall this is a lot of work and it looks like its going in the right direction. However, I feel  that not all parts are quite ready yet, maybe it would make sense to merge parts of the code in steps? I think the whole `NASequence` and associated ribonucleotide classes as well as `RNAMassCalculator` are probably close to being ready and could be merged separately. For the other stuff (id classes etc) there is still a lot documentation and unit tests missing. Any comments on this suggestion?

hroest:
there are some compiler errors here: http://cdash.openms.de/viewBuildError.php?buildid=287360
```


/.../source/src/utils/NucleicAcidSearchEngine.cpp:1042:26: error: use of undeclared identifier 'MetaboliteSpectralMatching'

```

and it does not seem to compile on Windows or under gcc 4.8 for pyopenms

poshul:
> there are some compiler errors here: http://cdash.openms.de/viewBuildError.php?buildid=287360
> 
> ```
> 
> 
> /.../source/src/utils/NucleicAcidSearchEngine.cpp:1042:26: error: use of undeclared identifier 'MetaboliteSpectralMatching'
> ```
> 
> and it does not seem to compile on Windows or under gcc 4.8 for pyopenms

I'm working on addressing the Windows compile issue. I will then take a look at the Mac one.

src/openms/include/OpenMS/METADATA/ID/MoleculeQueryMatch.h
```diff
@@ -0,0 +1,171 @@
...
+
+      explicit MoleculeQueryMatch(
+        IdentifiedMoleculeRef identified_molecule_ref,
+        DataQueryRef data_query_ref, Int charge = 0,
+        const ScoreList& scores = ScoreList(),
+        const std::vector<ProcessingStepRef>& processing_step_refs =
+        std::vector<ProcessingStepRef>(),
+        const PeakAnnotationSteps& peak_annotations = PeakAnnotationSteps()):
+        ScoredProcessingResult(scores, processing_step_refs),
+        identified_molecule_ref(identified_molecule_ref),
```
hroest: this is a problem, you shadow your member parameter here! you should probably use 
```
IdentifiedMoleculeRef identified_molecule_ref_tmp,
...
identified_molecule_ref(identified_molecule_ref_tmp)
```

src/openms/include/OpenMS/METADATA/ID/MoleculeQueryMatch.h
```diff
@@ -0,0 +1,171 @@
...
+
+      Int charge;
+
+      // peak annotations (fragment ion matches), potentially from different
+      // data processing steps:
+      PeakAnnotationSteps peak_annotations;
+
+      explicit MoleculeQueryMatch(
+        IdentifiedMoleculeRef identified_molecule_ref_tmp,
+        DataQueryRef data_query_ref, Int charge = 0,
```
timosachsenberg: @poshul some parameter of the constructor have the same name as the member variables. Thus clang complains see e.g. charge and data_query_ref 
I would also make the members private and add (+ public getter / setter if needed).
Then you can set charge_ (the private variable) to charge (the parameter) in the constructor

src/tests/class_tests/openms/source/IdentificationDataConverter_test.cpp
poshul: I've added these tests, but they don't work out of the box. @hendrikweisser can you take a look as to why?

src/tests/class_tests/openms/source/RibonucleotideDB_test.cpp
poshul: I don't foresee any use case of the software that would remove existing mods from the DB file. Any changes in the future should be additions.

poshul:
> Overall this is a lot of work and it looks like its going in the right direction. However, I feel that not all parts are quite ready yet, maybe it would make sense to merge parts of the code in steps? I think the whole `NASequence` and associated ribonucleotide classes as well as `RNAMassCalculator` are probably close to being ready and could be merged separately. For the other stuff (id classes etc) there is still a lot documentation and unit tests missing. Any comments on this suggestion?

The original goal of starting the pull request was to create an installer to distribute to the biology-centric people who have asked for working version of this code. I'd be happy to hop on a call next week sometime and talk with Timo, Hendrik and you about the direction of this. Thanks for all the hard work reviewing!

hroest:
But why need parantheses and brackets?  It may be confusing to have
different meanings in AA and NA context

hendrikweisser:
@hroest, @poshul:
It is perfectly legal to use the same name for an argument in a constructor as for a member variable; there is no "shadowing" if you use an initializer list. See e.g. here: https://stackoverflow.com/questions/6185020/initializing-member-variables-using-the-same-name-for-constructor-arguments-as-f
I like to use the same name because it makes everything cleaner. So in my opinion, @poshul's change [here](https://github.com/OpenMS/OpenMS/pull/3746/commits/45f5c668098f6ff139d87eb84ced0e2a1049157b) is not an improvement and I'd like to see it reversed.

hendrikweisser:
> But why need parantheses and brackets?

It so happens that the short code for the modification already includes the parentheses. I don't think it's enough of an issue to warrant fiddling with the "official" notation.

hendrikweisser:
> Overall this is a lot of work and it looks like its going in the right direction. However, I feel that not all parts are quite ready yet, maybe it would make sense to merge parts of the code in steps? I think the whole NASequence and associated ribonucleotide classes as well as RNAMassCalculator are probably close to being ready and could be merged separately. For the other stuff (id classes etc) there is still a lot documentation and unit tests missing. Any comments on this suggestion?

We discussed with Timo whether it would make sense to split this into multiple PRs and came to the conclusion that it wouldn't be worth the additional effort.

src/openms/include/OpenMS/METADATA/ID/MoleculeQueryMatch.h
hendrikweisser: I disagree. See here: https://github.com/OpenMS/OpenMS/pull/3746#issuecomment-427810167

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: Unfortunately this is not so easy. Because of all the cross-references between data, making tests self-contained would lead to a lot of code duplication. So I think it's better to keep the tests incremental (same as a lot of other OpenMS tests).

src/openms/include/OpenMS/METADATA/ID/MoleculeQueryMatch.h
hendrikweisser: See my comment here: https://github.com/OpenMS/OpenMS/pull/3746#issuecomment-427810167
I think it was better before the renaming.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: @hroest: Both. We want to get this merged as soon as possible.

src/openms/include/OpenMS/METADATA/ID/ScoredProcessingResult.h
hendrikweisser: This now supports multiple scores, so I would argue that the list of scores *is* a clean interface. I don't agree with writing unnecessary getters/setters if there are no invariants to check.

I am however still thinking about changing the implementation behind `ScoreList` (e.g. to directly include the processing step that assigned a score, or to allow efficient retrieval by `ScoreType` and in order via `boost::multi_index_container`).

src/openms/include/OpenMS/METADATA/ID/QueryMatchGroup.h
hendrikweisser: Not sure if I understand you correctly. The `QueryMatchGroups` (in this example) needs to be part of the public interface, but the fact that `QueryMatchGroups` is implemented as a multi-index container should be hidden - which I think it is.

src/openms/include/OpenMS/METADATA/ID/ParentMoleculeGroup.h
hendrikweisser: See https://github.com/OpenMS/OpenMS/pull/3746#discussion_r223354069.

src/openms/include/OpenMS/METADATA/ID/ScoredProcessingResult.h
hroest: > I am however still thinking about changing the implementation behind `ScoreList` (e.g. to directly include the processing step that assigned a score, or to allow efficient retrieval by `ScoreType` and in order via `boost::multi_index_container`).

having a simple, clean interface would make it easier to change the underlying implementation. Also, it is not possible to wrap something like `std::vector<std::pair< IteratorWrapper<std::set<ScoreType> ::iterator> , double>> ` in Python. 

> I don't agree with writing unnecessary getters/setters if there are no invariants to check.

I agree for simple POD. However, if the implementation is complex and subject to change, it can make sense. Also, unless the implementation is POD / a trivial member variable, we will need that for Python

src/openms/include/OpenMS/METADATA/ID/QueryMatchGroup.h
hroest: > Not sure if I understand you correctly. The `QueryMatchGroups` (in this example) needs to be part of the public interface, but the fact that `QueryMatchGroups` is implemented as a multi-index container should be hidden 

I agree

>which I think it is.

if you expose a public typdef like this and use it in other classes, then it becomes part of the public interface, does it not?

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
```diff
@@ -0,0 +1,91 @@
...
+#include <map>
+#include <set>
+
+namespace OpenMS
+{
+  class OPENMS_DLLAPI ModifiedNASequenceGenerator
+  {
+   /*
+    * @brief Modifications can be generated and combinatorically applied to NASequences. 
+    */
```
hroest: the docu string should be before the class definition:

```
 /*
  * @brief Modifications can be generated and combinatorically applied to NASequences. 
  */
  class OPENMS_DLLAPI ModifiedNASequenceGenerator
  {
```

also there should be more docu

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
```diff
@@ -0,0 +1,91 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Timo Sachsenberg $
+// $Authors: Timo Sachsenberg $
```
hroest: I suggest to change this to Sam or Hendrik?

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
```diff
@@ -0,0 +1,91 @@
...
+// --------------------------------------------------------------------------
+
+#ifndef OPENMS_ANALYSIS_CHEMISTRY_MODIFIEDNASEQUENCEGENERATOR_H
+#define OPENMS_ANALYSIS_CHEMISTRY_MODIFIEDNASEQUENCEGENERATOR_H
```
hroest: replace with `#pragma once`

src/openms/include/OpenMS/CHEMISTRY/NASequence.h
```diff
@@ -0,0 +1,504 @@
...
+      Precursor,      //< MS:1001523 Precursor ion
+      BIonMinusH20,   //< MS:1001222 b ion without water
+      YIonMinusH20,   //< MS:1001223 y ion without water
+      BIonMinusNH3,   //< MS:1001232 b ion without ammonia
+      YIonMinusNH3,   //< MS:1001233 y ion without ammonia
+      NonIdentified,  //< MS:1001240 Non-identified ion
+      Unannotated,    //< no stored annotation
+      WIon,           //< W ion, added for nucleic acid support
+      AminusB,        //< A ion with base loss, added for nucleic acid support
+      DIon,           //< D ion, added for nucleic acid support
```
hroest: all of these need to be `///<` (note the three slashes) for doxygen to pick them up.

src/openms/include/OpenMS/CHEMISTRY/NASequence.h
```diff
@@ -0,0 +1,504 @@
...
+
+  public:
+    /*
+     * Default constructors and assignment operators.
+     */
+    NASequence() = default; /// default constructor
+    NASequence(const NASequence&) = default; // Copy constructor
+    NASequence(NASequence&&) = default; // Move constructor
+    NASequence& operator=(const NASequence&) & = default;  // Copy assignment operator
+    NASequence& operator=(NASequence&&) & = default; // Move assignment operator
```
hroest: please use `///<` for this, right now doxygen will annotate `NASequence(const NASequence&) = default;` with "default constructor"

src/openms/include/OpenMS/CHEMISTRY/Ribonucleotide.h
```diff
@@ -0,0 +1,196 @@
...
+    String name_; //< full name
+    String code_; //< short name
+    String new_code_; //< Modomics code
+    String html_code_; //< RNAMods code
+    EmpiricalFormula formula_; //< sum formula
+    char origin_; //< character of unmodified version of ribonucleotide
+    double mono_mass_; //< monoisotopic mass
+    double avg_mass_; //< average mass
+    enum TermSpecificity term_spec_; //< terminal specificity
+    EmpiricalFormula baseloss_formula_; //< sum formula after loss of the nucleobase
```
hroest: should be `///<`

hroest: maybe search and replace the whole codebase?

src/openms/include/OpenMS/METADATA/ID/QueryMatchGroup.h
```diff
@@ -62,7 +62,7 @@ namespace OpenMS
         for (auto it = ++query_match_refs.begin(); it != query_match_refs.end();
              ++it)
         {
-          if ((*it)->identified_molecule_ref != ref) return false;
+          if (!((*it)->identified_molecule_ref == ref)) return false;
```
hroest: looks like 5b72555e16e845c4db8a3fff09f2160e020c7cfc (almost) fixed the pyOpenMS build

hroest:
ok, pyOpenMS compiles again. Please consider adding further bindings using the approach detailed here: https://github.com/OpenMS/OpenMS/wiki/pyOpenMS#wrap - its not perfect but it will give you a starting point for your wrapping

hendrikweisser:
The whole "shadowing variables in the c'tor" problem still bugs me. As we know, naming things (e.g. variables) is one of the two hard problems in computer science. I want to use proper names both for member variables in my code and for function arguments (that are important in the documentation!), not some weird variants.

> it leads to compiler warnings for "shadowing" and we would like to reduce those. In this case clang is right to complain (as you can read on SO), the member variable is shadowed by the function argument.

I don't agree that clang is right to complain - and apparently, neither does clang: http://llvm.org/viewvc/llvm-project?view=revision&revision=267957

> The alternative is to add pragmas to all these cases which is even more ugly. So either (i) we add the pragmas or (ii) rename the parameters. Up to you.

False dichotomy. Can we find a better set of clang parameters that warn for shadowing only in relevant cases? The link above would suggest that such parameters exist. (Note that I'm using GCC, where it looks like we currently compile without `-Wshadow`.)

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -436,7 +436,7 @@ END_SECTION
 
 START_SECTION((result<ScoreTypeRef, bool> findScoreType(const String& score_name, ProcessingSoftwareRef software_ref) const))
 {
-  Software sw("Other");
+  Software sw("Other","1.0");
```
hendrikweisser: This shouldn't be necessary. What was your reason for making this change?

hroest:
> The whole "shadowing variables in the c'tor" problem still bugs me. As we know, naming things (e.g. variables) is one of the two hard problems in computer science. I want to use proper names both for member variables in my code and for function arguments (that are important in the documentation!), not some weird variants.

I totally agree with you. I think my argument was only to explain *why* clang thinks its shadowing (because purely technically speaking it is), but I agree its not harmful shadowing and that is the important point.

> I don't agree that clang is right to complain - and apparently, neither does clang: http://llvm.org/viewvc/llvm-project?view=revision&revision=267957

Yes, it seems like an issue that was fixed here and there is an easy way to try it out: https://godbolt.org/z/HzTkr2 -- it looks like the following complain:

- clang 3.8 or lower
- not clang 3.9 or higher
- gcc: all of them (with `-Wshadow` but not `-Wall -Wextra`)

so it seems the problem only occurs in older clang (before 4.0 which was released March 2017) or any gcc

hendrikweisser:
@hroest: Cool, I'm glad we agree in principle. As a last note, new GCC versions (>7.0) apparently support `-Wshadow=[compatible-]local` (see e.g. [here](http://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)), which doesn't seem to trigger for shadowed member variables.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: Oh, it's probably that "result" in the function signature (after "START_SECTION") should be "pair". I'll fix it.

src/openms/include/OpenMS/METADATA/ID/MoleculeParentMatch.h
```diff
@@ -54,10 +54,10 @@ namespace OpenMS
...
       String left_neighbor, right_neighbor; // neighboring sequence elements
 
-      static const Size UNKNOWN_POSITION; // = Size(-1)
-      static const char UNKNOWN_NEIGHBOR; // = 'X'
-      static const char LEFT_TERMINUS; // = '['
-      static const char RIGHT_TERMINUS; // = ']'
+      static const Size UNKNOWN_POSITION = Size(-1);
+      static const char UNKNOWN_NEIGHBOR = 'X';
+      static const char LEFT_TERMINUS = '[';
+      static const char RIGHT_TERMINUS = ']';
```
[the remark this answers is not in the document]
hendrikweisser: Thanks for pointing that out. "odr-use" is a concept I hadn't come across. I'll change it to `constexpr`.

src/openms/include/OpenMS/CHEMISTRY/NASequence.h
```diff
@@ -0,0 +1,504 @@
...
+namespace OpenMS
+{
+
+  class OPENMS_DLLAPI NASequence
+  {
+
+  public:
+    enum NASFragmentType
+    {                 // NB: Not all fragments types are valid for all residue types, this class should probably get split
+      Full = 0,       // with N-terminus and C-terminus
```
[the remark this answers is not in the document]
hendrikweisser: As Hannes himself pointed out below, it should be `///<` (or `//!<`) for Doxygen.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: I've added more documentation.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: I've added documentation.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: You're right. I'm removing them.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: Good point. I've changed it to `... = default;` and it works.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: Nope, I see the problem is that my changes to `Software` got lost. I changed the constructor to `explicit Software(const String& name = "", const String& version = "")` (and removed the annoying spaces before the `&` of references).
@poshul, can you please update "Software.h" and "Software.cpp" with the versions from my branch?

src/openms/include/OpenMS/METADATA/Software.h
hendrikweisser: Some changes to `Software` got lost, presumably in a merge conflict. Let's fix that and have a look again.

src/openms/include/OpenMS/METADATA/ID/ScoredProcessingResult.h
hendrikweisser: Let's discuss these interface questions in our upcoming call.

src/openms/include/OpenMS/METADATA/ID/QueryMatchGroup.h
hendrikweisser: Let's discuss these interface questions in our upcoming call.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: The implementation has been replaced by `= default`.

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
hendrikweisser: Yes.

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
hendrikweisser: I haven't compared against `MetaValue` specifically.

> why not simply store a std::pair<int, double>

As you said, mainly due to the additional indirection. I also think that using `ScoreTypeRef` makes things more explicit than having `int`-that-indexes-a-list-of-scores.

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
```diff
@@ -0,0 +1,888 @@
...
+            evidence.getStart(), evidence.getEnd(), evidence.getAABefore(),
+            evidence.getAAAfter());
+          peptide.parent_matches[parent_ref].insert(match);
+        }
+        IdentificationData::IdentifiedPeptideRef peptide_ref =
+          id_data.registerIdentifiedPeptide(peptide);
+
+        IdentificationData::MoleculeQueryMatch match(peptide_ref, query_ref);
+        match.charge = hit.getCharge();
+        static_cast<MetaInfoInterface&>(match) = hit;
```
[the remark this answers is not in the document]
hendrikweisser: This is an interesting idea. My problem was: "How can you tell whether a meta value represents an additional score?" Perhaps your suggestion of treating every `double` meta value as a score would be a good heuristic. (I don't think we would typically use `int`s for scores.)

src/tests/class_tests/openms/source/IdentificationData_test.cpp
poshul: PR 37 (https://github.com/poshul/OpenMS/pull/37) didn't compile due to a lack of constructor signature mismatch. I fixed that in the least invasive way possible. I will pull in the rest of your changes.

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
hendrikweisser: Makes sense. Changed.

src/tests/class_tests/openms/source/IdentificationDataConverter_test.cpp
hendrikweisser: The conversion isn't perfect and the order isn't necessarily preserved.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: I've now added that check wherever it's relevant.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: I don't understand. A peptide can't be part of an RNA, only of a protein, which is what's checked here.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: Fixed.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: Yes. Fixed this and a bunch of similar copy-paste errors.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
hendrikweisser: Fixed.

src/tests/class_tests/openms/source/IdentificationDataConverter_test.cpp
hroest: @hendrikweisser can you sort and check that the main attributes are preserved and no data is lost?

src/openms/include/OpenMS/METADATA/Software.h
poshul: Fixed

hroest:
great, it looks like pyOpenMS worked! nice job!

src/pyOpenMS/pxds/Ribonucleotide.pxd
```diff
@@ -0,0 +1,104 @@
...
+        void setOrigin(char origin) nogil except +
+
+        # Set the HTML (RNAMods) code
+        void setHTMLCode(String html_code) nogil except +
+
+        # Get the HTML (RNAMods) code
+        String getHTMLCode() nogil except +
+
+        # Set the terminal specificity
+        #void setTermSpecificity(TermSpecificityNuc term_spec) nogil except +
```
hroest: any reason this doesnt work?

hendrikweisser:
I've investigated the Windows compile error. Funny thing: Replacing the `boost::unordered_map` in "CHEMISTRY/DigestionEnzymeDB.h" with a `std::unordered_map` makes it work under Windows, but breaks my Linux (GCC) build. :-(

poshul:
> I've investigated the Windows compile error. Funny thing: Replacing the `boost::unordered_map` in "CHEMISTRY/DigestionEnzymeDB.h" with a `std::unordered_map` makes it work under Windows, but breaks my Linux (GCC) build. :-(
> 
> EDIT: A workaround may be to just use a plain old `map`.

Can you use a preprocessor directive to include the correct unordered_map for the correct OS?

hendrikweisser:
> Can you use a preprocessor directive to include the correct unordered_map for the correct OS?

Could do, but I just changed the implementation to use `Map` now. There isn't really a performance reason for using `unordered_map` in this case. See https://github.com/poshul/OpenMS/pull/40.

hendrikweisser:
There's an open pull request (https://github.com/poshul/OpenMS/pull/41/) that addresses points raised during the review (more thorough testing, handling of scores), has some small fixes/improvements, and resolves the current merge conflicts.

After that, I'd be ready to open another PR that adds support for label-free quantification (via FeatureFinderMetaboIdent) to NucleicAcidSearchEngine.

hendrikweisser:
> That should be all from my side!

Nope, there's one more: https://github.com/poshul/OpenMS/pull/43
That should fix the OpenMP invocations and enable multithreading in NucleicAcidSearchEngine again.

poshul:
pyopenms bindings are busted again:

`pyopenms/pyopenms_2.cpp:229044:0: error: no matching function for call to ‘OpenMS::IdentificationDataConverter::exportIDs(boost::shared_ptr<OpenMS::IdentificationData>::element_type&, std::vector<OpenMS::ProteinIdentification>&, std::vector<OpenMS::PeptideIdentification>&, boost::shared_ptr<OpenMS::String>::element_type&, boost::shared_ptr<OpenMS::String>::element_type&, bool)’
`

hendrikweisser:
> pyopenms bindings are busted again

Yes, the function signatures need to be updated.

I've also just noticed a problem with the progress logging in NucleicAcidSearchEngine, caused by the dynamic scheduling. The `* NUMBER_OF_THREADS` needs to be removed here: https://github.com/OpenMS/OpenMS/blob/77f2e92defa7555225cf91d9a10dbb5d8cf81f36/src/utils/NucleicAcidSearchEngine.cpp#L1045

poshul:
@hroest @timosachsenberg do you folks have thoughts on other things that need to be added prior to merging? I think we've addressed the outstanding items.

timosachsenberg:
when I click on Files Changed in  this PR, I see some comments that do not show up in this thread.
e.g. replace with #pragma once
Could you once more go through this list and check if all has been addressed or address the outstanding ones?

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: Apparently the `= default` non-implementation didn't work on some platform (Windows or Mac?), so I've made the changes that Hannes suggested.

hendrikweisser:
> Could you once more go through this list and check if all has been addressed or address the outstanding ones?

I've just done that. (I haven't looked at the ones again that were already marked as "resolved".)
PR is here: https://github.com/poshul/OpenMS/pull/44

hendrikweisser:
> Did you remove this bit because it broke something for your sequence annotations (RNPXL)?

Looking at an example RNPxlSearch output file, I think the answer is "yes" (e.g. "y1-H3N1+" should just resolve to "y1"). I'll add a specific case for the "-B" in RNA annotations.

hendrikweisser:
Pull request with fix is here: https://github.com/poshul/OpenMS/pull/45

poshul:
@timosachsenberg Does this look okay to you now?

src/pyOpenMS/pxds/NASequence.pxd
```diff
@@ -0,0 +1,117 @@
...
+        # returns the name (ID) of the N-terminal modification, or an empty string if none is set
+        const RibonucleotideChainEnd * getFivePrimeMod() nogil except +
+
+        # sets the 3' modification
+        void setThreePrimeMod(const RibonucleotideChainEnd * modification) nogil except +
+
+        const RibonucleotideChainEnd * getThreePrimeMod() nogil except +
+
+        # returns the residue at position index
+        #Ribonucleotide * get(Size index) nogil except +
```
hroest: @poshul can you actually do anything with this Python object? If you cannot access the individual nucleotides, it seems to me its difficult to use, right?

hroest:
```suggestion
        const Ribonucleotide * get(Size index) nogil except +
```

this should work (I think), see 

https://github.com/OpenMS/OpenMS/blob/4902e996a3c64fe8d0beb8f49eaa54408da925c0/src/pyOpenMS/pxds/AASequence.pxd#L55

src/pyOpenMS/pxds/Ribonucleotide.pxd
```diff
@@ -0,0 +1,104 @@
+from Types cimport *
+from String cimport *
+from EmpiricalFormula cimport *
+#from RibonucleotideModification cimport *
```
hroest: does `RibonucleotideModification` exist? Or this a copy-paste error?

poshul:
Pinging @hroest  and @timosachsenberg  on this. Any other bits that I should fix, or is this ready to merge?

timosachsenberg:
is the test for the conversion of the fido (protein inference) result included? I think I ask about that a few times and can't recall if we settled on something

hendrikweisser:
@timosachsenberg: Yes, there is a test now, as part of the `IdentificationDataConverter_test`: https://github.com/OpenMS/OpenMS/blob/32428bfffd7debfd35926dadaeb640a56bc124e9/src/tests/class_tests/openms/source/IdentificationDataConverter_test.cpp#L120

timosachsenberg:
Ok I would be fine with the merge but could you please take a look at this failing test on windows:
´´´
Progress of 'converting peptide identifications':

 +  line 71:  TEST_EQUAL(peptides_in.size(),peptides_out.size()): got 3, expected 3
 +  line 81:  TEST_EQUAL(hits_in.size(),hits_out.size()): got 5, expected 5
 +  line 86:  TEST_EQUAL(find(hits_in.begin(), hits_in.end(), hit) != hits_in.end(),true): got 1, expected 1
 +  line 86:  TEST_EQUAL(find(hits_in.begin(), hits_in.end(), hit) != hits_in.end(),true): got 1, expected 1
 +  line 86:  TEST_EQUAL(find(hits_in.begin(), hits_in.end(), hit) != hits_in.end(),true): got 1, expected 1
 +  line 86:  TEST_EQUAL(find(hits_in.begin(), hits_in.end(), hit) != hits_in.end(),true): got 1, expected 1
 +  line 86:  TEST_EQUAL(find(hits_in.begin(), hits_in.end(), hit) != hits_in.end(),true): got 1, expected 1
 +  line 89:  TEST_EQUAL(proteins_in.size(),proteins_out.size()): got 2, expected 2
 -  line 97:  TEST_EQUAL(proteins_in[0].getHits() == proteins_out[0].getHits(),true): got 0, expected 1
 -  line 98:  TEST_EQUAL(proteins_in[1].getHits() == proteins_out[1].getHits(),true): got 0, expected 1
: failed
´´´

hroest:
there is also something strange going on with boost regex:

http://cdash.openms.de/testDetails.php?test=54563929&build=299885

```
terminate called after throwing an instance of 'std::regex_error'
  what():  regex_error
```

hroest:
> Pinging @hroest and @timosachsenberg on this. Any other bits that I should fix, or is this ready to merge?

all the tests have to work before a merge

hroest:
> Pinging @hroest and @timosachsenberg on this. Any other bits that I should fix, or is this ready to merge?

all the tests have to work before a merge

hendrikweisser:
I've opened a PR that updates the Modomics file, synchronizes with "develop" and fixes the merge conflicts:
https://github.com/poshul/OpenMS/pull/46
I get only a failure of the "SVMWrapper_test" (that always fails) on Linux. Let's see if the other errors persist.

poshul:
I've incorporated the updates from Hendrik, and am still seeing errors in
the test. Specifically:
105:  -  line 97:  TEST_EQUAL(proteins_in[0].getHits() ==
proteins_out[0].getHits(),true): got '0', expected '1'
105:  -  line 98:  TEST_EQUAL(proteins_in[1].getHits() ==
proteins_out[1].getHits(),true): got '0', expected '1'

hendrikweisser:
@poshul: Can you be more specific about where you see the errors? Which test, which platform? All I see in the automated checks at the moment are "continuous integration" failing (due to [this one](https://travis-ci.org/OpenMS/OpenMS/jobs/496609634) timing out) and "nightlies" failing (but I'm not sure what I'm looking at there).

hendrikweisser:
I've just built the branch on my Windows laptop (after some struggles*) and all tests passed.

* It seems that various libraries (incl. Qt and BZip2) were picked up from my Anaconda installation, although I set `OPENMS_CONTRIB_LIBS` and `CMAKE_PREFIX_PATH` (for Qt) in my CMake call. I did not have this issue in the past, so my suspicion is that some of the recent changes in the build system caused the problem (maybe different Qt etc. versions being used for compiling and linking?). Uninstalling Anaconda and reconfiguring/rebuilding everything solved it.

timosachsenberg:
what I would love to see in the future is some example code how to write out the new ID structure into a csv file (e.g., with multiple sections). This would help us coming up with a file format (let it be SQLite or parquet or whatever) to store this data efficiently.

hendrikweisser:
Yay, it is done! @hroest, @timosachsenberg - thanks for your help!

> what I would love to see in the future is some example code how to write out the new ID structure into a csv file (e.g., with multiple sections)

Not exactly what you asked for, but there is already code that fills an `MzTab` object for mzTab export. Maybe that helps.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/METADATA/ID/IdentificationDataConverter.cpp#L578

hendrikweisser:
@poshul: Please don't delete the "poshul/hendrikweisser-rna-ms2-id" branch. I'll use it to provide a link to the code for the journal.

Issue #624, opened 2014-01-09
Internal representation of idXML files very inefficient [624]


timosachsenberg:
Submitted by cbielow on 2013-11-04 10:33:46

Loading a 1GB file uses about 7GB of memory (Win64), plus it takes ages to load (indicating that the parser might have potential for speedup).

So, memory usage must go down, since this affects all tools using IDXML.
While at it, we can use proper members for MZ and RT, instead of meta-values -- this might actually be (one of) the reasons why memory consumption is so high.

---

Commented by hendrikweisser on 2013-11-06 12:21:45:

> While at it, we can use proper members for MZ and RT, instead of meta-values -- this might actually be (one of) the reasons why memory consumption is so high.

That should have happened a long time ago, but does it still make sense now that we're planning to switch to mzIdentML?

hroest:
I have done some benchmarking on the file provided from https://github.com/OpenMS/OpenMS/pull/3750#issuecomment-423726499 to document progress on this issue:

# Benchmarking 
```
$ /usr/bin/time ../release_2.1/bin/FileInfo  -in E3035_HCV_LAB1.idXML
^CCommand terminated by signal 2
176.97user 24.04system 7:57.48elapsed 42%CPU (0avgtext+0avgdata 13351316maxresident)k
$ /usr/bin/time ../release_2.3/bin/FileInfo  -in E3035_HCV_LAB1.idXML
File '/home/hr/.OpenMS/OpenMS.ini' is deprecated.
Updating missing/wrong entries in '/home/hr/.OpenMS/OpenMS.ini' with defaults!
^CCommand terminated by signal 2
153.69user 6.60system 2:41.31elapsed 99%CPU (0avgtext+0avgdata 13515052maxresident)k
$ /usr/bin/time ../release_2.4/./bin/FileInfo -in E3035_HCV_LAB1.idXML 
Memory usage (loading idXML): 10520 MB (working set delta)
FileInfo took 02:15 m (wall), 02:15 m (CPU), 3.20 s (system), 02:11 m (user).
```
Note that I had to abort the test with OpenMS 2.1 and 2.3 due to excessive memory consumption and swapping on my system after reaching 13 GB of memory usage.

after merging https://github.com/OpenMS/OpenMS/pull/3744
```
$ /usr/bin/time ../pr_3744/bin/FileInfo -in E3035_HCV_LAB1.idXML
Memory usage (loading idXML): 5327 MB (working set delta)
```
[...]

PR #3750, opened 2018-09-21
[Feature] fast basic protein scoring

jpfeuffer:
For a fast way to score proteins and group them.
@lars20070 

I will write tests and add documentation/python bindings for the graph class in later pull requests.

I can add pybindings for the algorithm class though soon.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/BasicProteinInferenceAlgorithm.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/BasicProteinInferenceAlgorithm.h
@@ -1,0 +1,64 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2017.
+//
+// This software is released under a three-clause BSD license:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of any author or any participating institution
+//    may be used to endorse or promote products derived from this software
+//    without specific prior written permission.
+// For a full list of authors, refer to the file AUTHORS.
+// --------------------------------------------------------------------------
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL ANY OF THE AUTHORS OR THE CONTRIBUTING
+// INSTITUTIONS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Julianus Pfeuffer $
+// $Authors: Julianus Pfeuffer $
+// --------------------------------------------------------------------------
+#pragma once
+
+#include <OpenMS/DATASTRUCTURES/DefaultParamHandler.h>
+#include <OpenMS/CONCEPT/ProgressLogger.h>
+#include <OpenMS/METADATA/ProteinIdentification.h>
+
+namespace OpenMS
+{
+  /// \brief Algorithm class that implements simple protein inference by aggregation of peptide scores
+  /// It has multiple parameter options like the aggregation method, when to distinguish peptidoforms,
+  /// and if you want to use shared peptides.
+  class OPENMS_DLLAPI BasicProteinInferenceAlgorithm :
+    public DefaultParamHandler,
+    public ProgressLogger
+  {
+    public:
+
+    enum AggregationMethod
+    {
+      PROD, SUM, MAXIMUM
+    };
+
+    /// Default constructor
+    BasicProteinInferenceAlgorithm();
+
+    /// main method of BasicProteinInferenceAlgorithm
+    /// inputs are not const, since it will get annotated with results
+    /// annotation of protein groups is currently only possible for a single protein ID run
+    void run(std::vector<PeptideIdentification> &pep_ids, std::vector<ProteinIdentification> &prot_ids) const;
+  };
+} //namespace OpenMS
--- a/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
@@ -35,6 +35,7 @@
 #ifndef OPENMS_ANALYSIS_ID_IDBOOSTGRAPH_H
 #define OPENMS_ANALYSIS_ID_IDBOOSTGRAPH_H
 
+//#include <OpenMS/ANALYSIS/ID/MessagePasserFactory.h> //included in BPI
 #include <OpenMS/CONCEPT/LogStream.h>
 #include <OpenMS/CONCEPT/Types.h>
 #include <OpenMS/KERNEL/StandardTypes.h>
@@ -45,7 +46,10 @@
 
 #include <boost/function.hpp>
 #include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/depth_first_search.hpp>
+#include <boost/graph/breadth_first_search.hpp>
 #include <boost/graph/filtered_graph.hpp>
+#include <boost/graph/subgraph.hpp>
 #include <boost/graph/properties.hpp>
 #include <boost/variant.hpp>
 #include <boost/variant/detail/hash_variant.hpp>
@@ -72,6 +76,8 @@ namespace OpenMS
   public:
 
     BOOST_STRONG_TYPEDEF(char, PeptideCluster)
+
+    // TODO make a double out of it and store the posterior
     BOOST_STRONG_TYPEDEF(char, ProteinGroup)
 
     //typedefs
@@ -79,9 +85,15 @@ namespace OpenMS
 
     typedef boost::variant<ProteinHit*, ProteinGroup*, PeptideCluster*, PeptideHit*> IDPointer;
     typedef boost::variant<const ProteinHit*, const ProteinGroup*, const PeptideCluster*, const PeptideHit*> IDPointerConst;
-    //TODO check the impact of different datastructures to store nodes/edges (maybe also use directed graph?)
+    //TODO check the impact of different data structures to store nodes/edges
+    // Directed graphs would make the internal computations much easier (less in/out edge checking) but boost
+    // does not allow computation of "non-strongly" connected components for directed graphs, which is what we would
+    // need
     typedef boost::adjacency_list <boost::setS, boost::vecS, boost::undirectedS, IDPointer> Graph;
-    typedef boost::adjacency_list <boost::setS, boost::vecS, boost::undirectedS, IDPointerConst> GraphConst;
+    typedef boost::subgraph<boost::adjacency_list <boost::setS, boost::vecS, boost::undirectedS, IDPointer>> SubGraph;
+    typedef std::vector<Graph> Graphs;
+    typedef std::vector<SubGraph> SubGraphs;
+    typedef boost::adjacency_list <boost::setS, boost::vecS, boost::undirectedS, IDPointer> GraphConst;
     typedef boost::graph_traits<Graph>::vertex_descriptor vertex_t;
     typedef boost::graph_traits<Graph>::edge_descriptor edge_t;
     typedef boost::filtered_graph<Graph, boost::function<bool(edge_t)>, boost::function<bool(vertex_t)> > FilteredGraph;
@@ -91,8 +103,8 @@ namespace OpenMS
     /// Constructor
     IDBoostGraph(ProteinIdentification &proteins, std::vector<PeptideIdentification>& idedSpectra);
 
-    /// Do sth on connected components
-    void applyFunctorOnCCs(std::function<void(FilteredGraph &)> functor);
+    /// Do sth on connected components (your functor object has to inherit from std::function)
+    void applyFunctorOnCCs(std::function<void(Graph &)> functor);
 
     /// Add intermediate nodes to the graph that represent indist. protein groups and peptides with the same parents
     /// this will save computation time and oscillations later on.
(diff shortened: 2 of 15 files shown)


src/openms/include/OpenMS/ANALYSIS/ID/BasicProteinInferenceAlgorithm.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <OpenMS/DATASTRUCTURES/DefaultParamHandler.h>
+#include <OpenMS/CONCEPT/ProgressLogger.h>
+#include <OpenMS/METADATA/ProteinIdentification.h>
+
+namespace OpenMS
+{
+  /// \brief Algorithm class that implements simple protein inference by aggregation of peptide scores
+  /// It has multiple parameter options like the aggregation method, when to distinguish peptidoforms,
+  /// and if you want to use shared peptides.
```
timosachsenberg: Please add additional information/details how inference is performed.

src/openms/include/OpenMS/ANALYSIS/ID/BasicProteinInferenceAlgorithm.h
```diff
@@ -0,0 +1,64 @@
...
+    {
+      PROD, SUM, MAXIMUM
+    };
+
+    /// Default constructor
+    BasicProteinInferenceAlgorithm();
+
+    /// main method of BasicProteinInferenceAlgorithm
+    /// inputs are not const, since it will get annotated with results
+    /// annotation of protein groups is currently only possible for a single protein ID run
```
timosachsenberg: please specify what input is required and what will be annotated in detail

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -0,0 +1,278 @@
...
+
+    /// Constructor
+    IDBoostGraph(ProteinIdentification &proteins, std::vector<PeptideIdentification>& idedSpectra);
+
+    /// Do sth on connected components
+    void applyFunctorOnCCs(std::function<void(FilteredGraph &)> functor);
+
+    /// Add intermediate nodes to the graph that represent indist. protein groups and peptides with the same parents
+    /// this will save computation time and oscillations later on.
+    void clusterIndistProteinsAndPeptides();
```
timosachsenberg: is it mandatory to call this method? or just to speed up the computation?

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -0,0 +1,278 @@
...
+
+      OpenMS::String operator()(const PeptideCluster* /*pc*/) const
+      {
+        return String("PepClust");
+      }
+
+    };
+
+    /// Visits nodes in the boost graph (ptrs to an ID Object) and depending on their type prints the address. For debug
+    class PrintAddressVisitor:
```
timosachsenberg: maybe construct with stream and use that in output?

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -0,0 +1,278 @@
...
+    public:
+
+      void operator()(PeptideHit* pep, double posterior) const
+      {
+        pep->setScore(posterior);
+      }
+
+      void operator()(ProteinHit* prot, double posterior) const
+      {
+        std::cout << "set score " << posterior << " for " << prot->getAccession() << std::endl;
```
timosachsenberg: debug message?

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -0,0 +1,278 @@
...
+
+      void operator()(const PeptideCluster* /*pc*/, double /*posterior*/) const
+      {
+        // do nothing
+      }
+
+    };
+
+    /// Visits nodes in the boost graph (ptrs to an ID Object) and depending on their type creates a random
+    /// variable or "Dependency" to add into the InferenceGraph
```
timosachsenberg: clean up?

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -0,0 +1,278 @@
...
+    /// @param use_all_psms If all or just the FIRST psm should be used
+    void buildGraph(bool use_all_psms);
+    //void buildGraph(const ProteinIdentification& protein, const std::vector<PeptideIdentification>& peptides);
+
+  private:
+    Graph g;
+
+    /// static objects created from a hard typedef to differentiate between different types of nodes.
+    /// they will be reused throughout the graph when necessary
+    static PeptideCluster staticPC;
```
timosachsenberg: are these needed?

src/openms/source/ANALYSIS/ID/BasicProteinInferenceAlgorithm.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <map>
+#include <unordered_map>
+
+namespace OpenMS
+{
+  BasicProteinInferenceAlgorithm::BasicProteinInferenceAlgorithm():
+      DefaultParamHandler("BasicProteinInferenceAlgorithm"),
+      ProgressLogger()
+  {
+    //registerIntOption_("min_peptides_per_protein", "<num>", 2, "Minimal number of peptides needed for a protein identification", false);
```
timosachsenberg: used in future version?

src/openms/source/ANALYSIS/ID/IDBoostGraph.cpp
```diff
@@ -0,0 +1,470 @@
...
+  {
+    //TODO add support for (consensus) feature information
+
+    unordered_map<IDPointer, vertex_t, boost::hash<IDPointer>> vertex_map{};
+
+    unordered_map<string, ProteinHit*> accession_map{};
+
+    auto protIt = proteins_.getHits().begin();
+    auto protItEnd = proteins_.getHits().end();
+    for (; protIt != protItEnd; ++protIt)
```
timosachsenberg: a few lines shorter with: 
for (auto const & ph : proteins_.getHits())

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
jpfeuffer: Yes this is not used yet. It's for later and yes mostly for saving duplicate work.

src/openms/include/OpenMS/ANALYSIS/ID/BasicProteinInferenceAlgorithm.h
jpfeuffer: Sure. Will do.

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
jpfeuffer: Yup. Needs some cleanup. Agree. I can do it in this pr

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
jpfeuffer: Yes they are placeholders for intermediate nodes that connect proteins and peptides that share computations. Not used for basic inference/grouping.

src/openms/source/ANALYSIS/ID/BasicProteinInferenceAlgorithm.cpp
jpfeuffer: I was thinking about including filtering in that tool but I am not sure because you could easily do it afterwards in an idfilter.

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
timosachsenberg: Can we these be non static?

lars20070:
@jpfeuffer Great. [Here](https://www.dropbox.com/s/0904koqpf8legqo/E3035_HCV_LAB1.idXML?dl=1) and example `idXML`. What's the runtime?

hroest:
@lars20070 how did you generate this idXML file?

@timosachsenberg @jpfeuffer there are some strange entries in here


```
               <PeptideIdentification score_type="q-value" higher_score_better="false" significance_threshold="0" MZ="1066.57165527344" RT="12001.7388" spectrum_reference="controllerType=0 controllerNumber=1 scan=156335" >
                        <PeptideHit score="0.00435916238898395" sequence="IKREPPEIHRAGMVPSRSC(Methylthio)PSITHGPRLSSVPRTRRR" charge="4" aa_before="K K K K K" aa_after="A A A A A" start="333 239 99 333 299" end="369 275 135 369 335" protein_refs="PH_25225 PH_25704 PH_28726 PH_30042 PH_31516" >
                                <UserParam type="string" name="target_decoy" value="target"/>
                                <UserParam type="float" name="MS:1002049" value="31"/>
                                <UserParam type="float" name="MS:1002050" value="133"/>
                                <UserParam type="float" name="MS:1002052" value="3.325968e-10"/>
                                <UserParam type="float" name="MS:1002053" value="0.009342001"/>
                                <UserParam type="string" name="AssumedDissociationMethod" value="CID"/>
                                <UserParam type="string" name="CTermIonCurrentRatio" value="0.19019023"/>
                                <UserParam type="string" name="ExplainedIonCurrentRatio" value="0.26852977"/>
                                <UserParam type="string" name="IsotopeError" value="0"/>
                                <UserParam type="string" name="MS2IonCurrent" value="51260.24"/>
                                <UserParam type="string" name="MeanErrorAll" value="204.33833"/>
                                <UserParam type="string" name="MeanErrorTop7" value="172.44984"/>
                                <UserParam type="string" name="MeanRelErrorAll" value="6.537234"/>
                                <UserParam type="string" name="MeanRelErrorTop7" value="145.52988"/>
                                <UserParam type="string" name="NTermIonCurrentRatio" value="0.07833954"/>
                                <UserParam type="string" name="NumMatchedMainIons" value="24"/>
                                <UserParam type="string" name="StdevErrorAll" value="176.65616"/>
                                <UserParam type="string" name="StdevErrorTop7" value="110.185905"/>
                                <UserParam type="string" name="StdevRelErrorAll" value="270.03485"/>
                                <UserParam type="string" name="StdevRelErrorTop7" value="143.8782"/>
                                <UserParam type="float" name="calcMZ" value="1066.57739257813"/>
                                <UserParam type="int" name="pass_threshold" value="1"/>
                                <UserParam type="int" name="start" value="300"/>
                                <UserParam type="int" name="end" value="336"/>
                                <UserParam type="string" name="protein_references" value="non-unique"/>
                                <UserParam type="float" name="SpecEValue_score" value="3.325968e-10"/>
                        </PeptideHit>
                        <UserParam type="string" name="MS:1001115" value="156335"/>
```

there are tons of userParams that are floating point numbers but annotated as `type="string"` which will lead to inefficient storage in OpenMS. Any chance somebody knows how these could have gotten in there? These are 26 userParams of which 15 are annotated as String but are double values (could be stored as 16 byte `DataValue` but instead will be stored as 16 byte + 32 byte `String` + `len(str)`

jpfeuffer:
True. Could be Psmfeatureextractor? But not sure if this was just not seem until now or if it is a bug.

hroest:
I did some performance analysis:

```
$ /usr/bin/time ../release_2.4/./bin/FileInfo -in E3035_HCV_LAB1.idXML 
Memory usage (loading idXML): 10520 MB (working set delta)
```

with the changes in #3744 
```
$ /usr/bin/time ./bin/FileInfo -in E3035_HCV_LAB1.idXML
Memory usage (loading idXML): 5327 MB (working set delta)
```

and when removing all string values
```
grep -v 'UserParam type="string"' E3035_HCV_LAB1.idXML  > /tmp/q.idXML
$ /usr/bin/time ./bin/FileInfo -in /tmp/q.idXML
Memory usage (loading idXML): 2561 MB (working set delta)
```
So I calculate that on average each of the string values gives rise to about 90 bytes in memory which seems in the expected range. We probably waste a lot of space storing the same identifiers over and over again, such as 
```
<UserParam type="string" name="target_decoy" value="target"/>
<UserParam type="string" name="AssumedDissociationMethod" value="CID"/>
<UserParam type="string" name="protein_references" value="non-unique"/>
```

I think either we should remove those or store them explicitly in a member variable, especially binary variables such as unique/non-unique or target/decoy/target+decoy that can be encoded with 1 or 2 bits.

I am still amazed how a 2.7 GB file gives rise to a 10 GB in-memory structure.

timosachsenberg:
I think this might be a bug in the mzIdentML part. These values are generated by MSGF+

jpfeuffer:
I agree with @hroest that probably uniqueness and TD annotation are so fundamental that they could get a member. Two things to consider: before peptideindexer these members do not exist. And TD annotations can currently AFAIK take three values (e.g. if a peptide maps to both)

timosachsenberg:
I think it is the MzIdentMLDomHandler:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp#L1444
Seems every userParam is treated as string

hroest:
> I did some performance analysis:

I have an update on the memory performance, using  #3755

```
$ /usr/bin/time ./bin/FileInfo -in /media/doc/tmp/openms_issues/E3035_HCV_LAB1.idXML 
Memory usage (loading idXML): 3168 MB (working set delta)
FileInfo took 01:55 m (wall), 01:55 m (CPU), 1.76 s (system), 01:53 m (user).
113.70user 1.76system 1:55.48elapsed 99%CPU (0avgtext+0avgdata 3411876maxresident)k
1768inputs+0outputs (15major+899236minor)pagefaults 0swaps
```

and when removing all string values
```
$ grep -v 'UserParam type="string"' E3035_HCV_LAB1.idXML  > /tmp/qqq.idXML
$ /usr/bin/time ./bin/FileInfo -in /tmp/qqq.idXML 
Memory usage (loading idXML): 1709 MB (working set delta)
FileInfo took 49.02 s (wall), 49.01 s (CPU), 0.90 s (system), 48.11 s (user).
FileInfo took 49.02 s (wall), 49.01 s (CPU), 0.90 s (system), 48.11 s (user).
48.13user 0.90system 0:49.09elapsed 99%CPU (0avgtext+0avgdata 1930636maxresident)k
```

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
jpfeuffer: Also, I will probably change it in the future because I want to store info in the groups at least.

src/openms/source/ANALYSIS/ID/BasicProteinInferenceAlgorithm.cpp
timosachsenberg: I think I would add basic filtering to the tool.

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
timosachsenberg: if these are read-only then you could keep them but it must be static const.
If they are also written to, then they should not be static members as this will kill any concurrency (e.g., if multiple instances exist).

jpfeuffer:
Ok I am pretty sure, that the user param problem is a problem coming from MSGFPlus / an under-specification in the mzIdentML standard.
For mzIdentML it is not mandatory to set the data type of a UserParam. So MSGFPlus does not do that and we have no choice but reading it as string or write a hardcoded conversion method that (e.g. in MSGFPlusAdapter removes all string MetaValues, checks its actual type with a hardcoded map and readds the MetaValue correctly).

jpfeuffer:
@lars20070 I cannot process your data on my computer, it's too big. If you want it tested, please send me a smaller set. Maybe around 10 runs first. And I would suggest not to pass through protein sequences+descriptions from the beginning and rather reannotate in the end. Each protein run will have a big overlap in proteins and huge sequences will be stored redundantly.

hroest:
@jpfeuffer @timosachsenberg @lars20070 I have done some further analysis and I now estimate if all the floats are read in as floats instead of strings, the file should take about 2200 MB in memory in total. See https://github.com/OpenMS/OpenMS/issues/624#issuecomment-425163006

hroest:
> For mzIdentML it is not mandatory to set the data type of a UserParam. So MSGFPlus does not do that and we have no choice but reading it as string or write a hardcoded conversion method

or we could check if the first character is a number whether the whole thing can be converted to int or float and then store it like that. Seems like an easy fix and not much overhead, if the first char is not a number then its probably a true string. Not sure about hardcoding things ...

lars20070:
@jpfeuffer Let me try to run the analysis on my 64 GB RAM Ubuntu machine. I will build this branch. Can you please post the wf?

lars20070:
I will also re-run the complete MS-GF+ search without protein sequence/description annotation.

lars20070:
@jpfeuffer The MS-GF+ search of the complete datatset is finished. Can send you the `idXML` via FTP.

Cannot build your branch.
```
[ 42%] Building CXX object src/openms_gui/CMakeFiles/OpenMS_GUI.dir/OpenMS_GUI_automoc.cpp.o
[ 43%] Linking CXX shared library ../../lib/libOpenMS_GUI.so
[ 43%] Built target OpenMS_GUI
Makefile:138: recipe for target 'all' failed
make: *** [all] Error 2
```

jpfeuffer:
I am currently fixing stuff and improving (including automatic merging).
I could make it run on a reduced version of your dataset. PR updated soon.

lars20070:
I want to see how the algorithm scales, i.e. run the **complete dataset** on our 8-core 64GB RAM Ubuntu machine.

jpfeuffer:
It's very basic. Less than half an hour if I extrapolate. It was around 15 mins in debug mode on my quadcore on 35 of those 42 data sets.

src/openms/source/ANALYSIS/ID/IDBoostGraph.cpp
```diff
@@ -191,15 +194,16 @@ namespace OpenMS
       int i = 0;
 
       #pragma omp parallel for
-      for (const Graph& fg : ccs_)
+      for (auto g_it = ccs_.begin() ; g_it < ccs_.end(); ++g_it)
```
hroest: I dont think that works on Windows, for this you need OpenMP 3.0 : https://stackoverflow.com/questions/2513988/iteration-through-std-containers-in-openmp  @cbielow is that correct?

timosachsenberg:
is this outdated after PFLQ branch?

PR #3751, opened 2018-09-21
[INTERNAL, FEATURE] Refactor of SvmTheoreticalSpectrumGeneratorSet

rinmiEft:
As according to:
C++ Core Guidelines:
* [C.defop.20 - If you can avoid defining default operations, do.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero)
* [C.copy.62 - Make copy assignment safe for self-assignment.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-self)
* [P.3 - Express intent.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rp-what)
* [P.13 - Use support libraries as appropriate.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rp-lib)
* [F.def.1 - "Package" meaningful operations as carefully named functions.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-package)

Clang-Tidy:  
* [modernize-use-auto](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html#iterators)

Notes / Observations:
* [N4165 - Unified Call Syntax](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4165.pdf)

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
@@ -61,29 +61,7 @@ namespace OpenMS
   class OPENMS_DLLAPI SvmTheoreticalSpectrumGeneratorSet
   {
 public:
-
-    /** @name Constructors and Destructors
-     */
-    //@{
-    /// Default constructor
-    SvmTheoreticalSpectrumGeneratorSet();
-
-    /// Copy constructor
-    SvmTheoreticalSpectrumGeneratorSet(const SvmTheoreticalSpectrumGeneratorSet & source);
-
-    /// Move Constructor
-    SvmTheoreticalSpectrumGeneratorSet(SvmTheoreticalSpectrumGeneratorSet&& source) noexcept;
-
-    /// Destructor
-    virtual ~SvmTheoreticalSpectrumGeneratorSet();
-    //@}
-
-    /// Assignment operator
-    SvmTheoreticalSpectrumGeneratorSet & operator=(const SvmTheoreticalSpectrumGeneratorSet & tsg);
-
-    /// Move Assignment operator
-    SvmTheoreticalSpectrumGeneratorSet & operator=(SvmTheoreticalSpectrumGeneratorSet&& tsg) noexcept;
-
+    // Rule of Zero in effect for this class.
 
     /// Generate the MS/MS according to the model for the given precursor_charge
     void simulate(PeakSpectrum & spectrum, const AASequence & peptide, boost::random::mt19937_64& rng, Size precursor_charge);
--- a/src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
@@ -82,7 +82,7 @@ public:
     void simulate(PeakSpectrum & spectrum, const AASequence & peptide, boost::random::mt19937_64& rng, Size precursor_charge);
 
     ///Load a trained Svm and Prob. models
-    void load(String);
+    void load(String filename);
 
     ///Return precursor charges for which a model is contained in the set
     void getSupportedCharges(std::set<Size> & charges);
--- a/src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
@@ -71,12 +71,19 @@ public:
     /// Copy constructor
     SvmTheoreticalSpectrumGeneratorSet(const SvmTheoreticalSpectrumGeneratorSet & source);
 
+    /// Move Constructor
+    SvmTheoreticalSpectrumGeneratorSet(SvmTheoreticalSpectrumGeneratorSet&& source) noexcept;
+
     /// Destructor
     virtual ~SvmTheoreticalSpectrumGeneratorSet();
     //@}
 
     /// Assignment operator
     SvmTheoreticalSpectrumGeneratorSet & operator=(const SvmTheoreticalSpectrumGeneratorSet & tsg);
+
+    /// Move Assignment operator
+    SvmTheoreticalSpectrumGeneratorSet & operator=(SvmTheoreticalSpectrumGeneratorSet&& tsg) noexcept;
+
 
     /// Generate the MS/MS according to the model for the given precursor_charge
     void simulate(PeakSpectrum & spectrum, const AASequence & peptide, boost::random::mt19937_64& rng, Size precursor_charge);
--- a/src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
+++ b/src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
@@ -86,32 +86,34 @@ namespace OpenMS
       filename = File::find(filename);
     }
 
-    Param sim_param = SvmTheoreticalSpectrumGenerator().getDefaults();
-
     TextFile file(filename);
-    TextFile::ConstIterator it = file.begin();
-
-    if (it == file.end()) return; // no data to load
+    // Boost::empty()
+    if (std::begin(file) == std::end(file)) { return; }// no data to load
 
     // skip header line
-    ++it;
-    // process content
-    for (; it != file.end(); ++it)
+    const auto begin_from_skipped_header = std::next(std::begin(file));
+
+    Param sim_param = SvmTheoreticalSpectrumGenerator().getDefaults();
+
+    const auto process_content = [this, filename, &sim_param] (TextFile::Iterator::value_type i)
     {
-      std::vector<String> spl;
-      it->split(":", spl);
-      Int precursor_charge = spl[0].toInt();
+        std::vector<String> spl;
+        i.split(":", spl);
+        Int precursor_charge = spl[0].toInt();
 
-      if (spl.size() != 2 || precursor_charge < 1)
-      {
-        throw Exception::ParseError(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, *it, " Invalid entry in SVM model File");
-      }
+        if (spl.size() != 2 || precursor_charge < 1)
+        {
+          throw Exception::ParseError(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, i, " Invalid entry in SVM model File");
+        }
 
-      //load the model into the map
-      sim_param.setValue("model_file_name", File::path(filename) + "/" + spl[1]);
-      simulators_[precursor_charge].setParameters(sim_param);
-      simulators_[precursor_charge].load();
-    }
+        //load the model into the map
+        sim_param.setValue("model_file_name", File::path(filename) + "/" + spl[1]);
+        simulators_[precursor_charge].setParameters(sim_param);
+        simulators_[precursor_charge].load();
+    };
+
+    // process content
+    std::for_each(begin_from_skipped_header, std::end(file), process_content);
   }
 
   //Return precursor charges for which a model is contained in the set
--- a/src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
+++ b/src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
@@ -67,8 +67,8 @@ namespace OpenMS
   // Generate the MS/MS according to the given probabilistic model
   void SvmTheoreticalSpectrumGeneratorSet::simulate(PeakSpectrum& spectrum, const AASequence& peptide, boost::random::mt19937_64& rng, Size precursor_charge)
   {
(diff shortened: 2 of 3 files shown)


src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
```diff
@@ -61,28 +61,13 @@ namespace OpenMS
...
-    /// Assignment operator
-    SvmTheoreticalSpectrumGeneratorSet & operator=(const SvmTheoreticalSpectrumGeneratorSet & tsg);
+    // Rule of Zero in effect for this class.
 
     /// Generate the MS/MS according to the model for the given precursor_charge
     void simulate(PeakSpectrum & spectrum, const AASequence & peptide, boost::random::mt19937_64& rng, Size precursor_charge);
 
     ///Load a trained Svm and Prob. models
-    void load(String);
+    void load(String filename);
```
hroest: why not `const String& filename` ?

src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
```diff
@@ -86,54 +59,56 @@ namespace OpenMS
...
+    if (std::begin(file) == std::end(file)) { return; }// no data to load
+
+    // skip header line
+    const auto begin_from_skipped_header = std::next(std::begin(file));
+
     Param sim_param = SvmTheoreticalSpectrumGenerator().getDefaults();
 
-    TextFile file(filename);
-    TextFile::ConstIterator it = file.begin();
+    const auto process_content = [this, filename, &sim_param] (TextFile::Iterator::value_type i)
```
hroest: @timosachsenberg I am not sure I agree with this, this is turning a simple loop into a function without any real benefit. As a simple "code maintenance" I dont really agree with such a change

src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
```diff
@@ -86,54 +59,56 @@ namespace OpenMS
...
     charges.clear();
-    std::map<Size, SvmTheoreticalSpectrumGenerator>::const_iterator it;
-    for (it = simulators_.begin(); it != simulators_.end(); ++it)
-    {
-      charges.insert(it->first);
-    }
+
+    std::transform(std::begin(simulators_), std::end(simulators_),
+                   std::inserter(charges, std::begin(charges)),
+                   [] (std::pair<Size, SvmTheoreticalSpectrumGenerator> it) { return it.first; });
```
hroest: same here, I feel this code would be much more readable, concise and express intent to anybody
```
for (const auto& it : simulators_) charges.insert(it.first)
```

hroest:
thanks a lot, we really appreciate these improvements. I have two changes I dont really like where I feel the code becomes more convoluted and harder to read, specifically using algorithm functions such as `std::transform` and `std::for_each` which may sound elegant but 

- make the code hard to read
- can easily be expressed using a simple loop

In my opinion, simpler code leads to fewer bugs, more people looking at the code while at the same time people coming from Python and other languages also capable of understanding it (while the STL constructs are very C++ specific and a newcomer will not be able to read the code).

In general we dont have any specific guidelines on this, however I feel that a simple code maintenance should not make the code more complex 

@timosachsenberg @jpfeuffer any comments?

src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
timosachsenberg: I agree with @hroest

src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
timosachsenberg: agree

rinmiEft:
Thank you for the feedback and encouragement!
Fuller questions / replies to PR review to be written after (possible) correspondence from @timosachsenberg and @jpfeuffer.

As a side note, the PRs produced are (hopefully) to contribute code and also to encourage debate on practices accepted / not accepted in a project. The PRs are also written to highlight things lacking in the project guideline's documents so it's great that this came through in the PR.

@hroest, could you clarify on your thoughts in a bit more detail before follow-up replies to the review:

> In my opinion, simpler code leads to fewer bugs, more people looking at the code while at the same time people coming from Python and other languages also capable of understanding it (while the STL constructs are very C++ specific and a newcomer will not be able to read the code).

Who is the project to be written / maintained by? Assuming the project is a C++ project but this paragraph makes this slightly unclear. Is there a target audience of developers the project is for? (This clarity will help in framing replies to responses for future reference.)

src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
```diff
@@ -61,28 +61,13 @@ namespace OpenMS
...
-    /// Copy constructor
-    SvmTheoreticalSpectrumGeneratorSet(const SvmTheoreticalSpectrumGeneratorSet & source);
-
-    /// Destructor
-    virtual ~SvmTheoreticalSpectrumGeneratorSet();
-    //@}
-
-    /// Assignment operator
-    SvmTheoreticalSpectrumGeneratorSet & operator=(const SvmTheoreticalSpectrumGeneratorSet & tsg);
+    // Rule of Zero in effect for this class.
```
hroest: My preference would actually be 

```
    
     /// Default constructor
    SvmTheoreticalSpectrumGeneratorSet() = default;
    /// Copy constructor
    SvmTheoreticalSpectrumGeneratorSet(const SvmTheoreticalSpectrumGeneratorSet&) = default;               
    /// Move constructor
    SvmTheoreticalSpectrumGeneratorSet(SvmTheoreticalSpectrumGeneratorSet&&) = default;                    
    /// Copy assignment operator
    SvmTheoreticalSpectrumGeneratorSet& operator=(const SvmTheoreticalSpectrumGeneratorSet&) & = default;  
    /// Move assignment operator
    SvmTheoreticalSpectrumGeneratorSet& operator=(SvmTheoreticalSpectrumGeneratorSet&&) & = default;       
    /// Destructor
    virtual ~SvmTheoreticalSpectrumGeneratorSet() = default;    
```

also note that your change has the subtle effect of changing `SvmTheoreticalSpectrumGeneratorSet` from a class with a virtual destructor to one without one! This changes its size as well as the behavior of ptrs to derived classes. 

@timosachsenberg @jpfeuffer any comments regarding the rule of zero / rule of zero defaults / rule of five?

My preference is discussed here https://github.com/OpenMS/OpenMS/issues/3310#issuecomment-411141372 and see also http://scottmeyers.blogspot.com/2014/03/a-concern-about-rule-of-zero.html

timosachsenberg:
@hroest nicely summarized our primary target audience. One thing that I want to highlight is that we provide python binding. This poses some additional constraints on how we design our C++ interfaces as not everything available in C++ is available in python. One of these things are templated code (especially member functions), for which we usually provide additional, specialized members to be wrapped in python. Also, I think a wrapped class needs to have a public default constructor and copy constructor (which is not ideal but as I understood it hard to change in our wrapper code). So maybe we should also include the information in http://ftp.mi.fu-berlin.de/pub/OpenMS/release-documentation/html/pyOpenMS.html to update our coding conventions

rinmiEft:
@hroest 
>  In practice it is mostly written by computer scientists but is used by a variety of people that also use our Python bindings to execute our code. The idea, in my understanding, is to write modern, understandable, object-oriented C++ code. Part of our hope is that we can draw developers from the Python users to also develop in C++. 

@timosachsenberg 
> One thing that I want to highlight is that we provide python binding. This poses some additional constraints on how we design our C++ interfaces as not everything available in C++ is available in python. 

The reason for asking for clarity was exactly for these statements. Yes, the project for now is a C++ project but over time this could change into a "Python + bindings around C++ libraries" project.

@hroest 
> I suggest to participate in #3310 since we are just working on adopting our coding conventions to C++11 and the current one is still based on C++98...

A refactoring path with projects of similar nature to this (upgrading the accepted codebase to C++11 standards) is the "refactor to pybind11 (a c++11-written python binding library) for all "mission critical" c++ algorithms and then do all other non-critical code in python. It seemed unclear if this was where the project was heading. The clarity helps in reframing potential candidates for refactoring.

src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
rinmiEft: The `lambda` form helps encapsulate the `for-loop` form. It reveals the `for-loop` form's dependencies to help with further refactoring. 
The basic `std::for_each` change was to help keep the PR small. Further refactoring can be done however this PR is to check if refactoring would be accepted or not.

src/openms/source/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.cpp
rinmiEft: The `lambda` form encapsulates the `for-loop` form. Using `std::transform` + `std::inserter` gives certain guarantees and intent about the datatype for `simulators_` and `charges` that are not promised in the `for-loop` form.
Technically this should be a `const std::pair<X,Y>` (or `const auto` in c++14), but mainly this commit was to check if this style of PR refactoring would be accepted into the codebase.
n.b. The form prepares for later c++ standards for parallel standard library algorithms.

rinmiEft: An author familiar wrote about this exact topic for their book in 2001[[1]](http://www.drdobbs.com/stl-algorithms-vs-hand-written-loops/184401446).
 Instead, `lambdas` superseded `std::bind2nd` in c++11 standards and beyond.


1: http://www.drdobbs.com/stl-algorithms-vs-hand-written-loops/184401446

src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
rinmiEft: If no move semantics are guaranteed, `const String& filename` should be preferred.

timosachsenberg:
@rinmiEft we already provide pyOpenMS bindings to many OpenMS classes and data structures (see https://github.com/OpenMS/OpenMS/wiki/pyOpenMS ) . We use autowrap though (see https://github.com/uweschmitt/autowrap ).

src/openms/include/OpenMS/CHEMISTRY/SvmTheoreticalSpectrumGeneratorSet.h
hroest: > Maybe a compromise would suit, a "Rule of All or Nothing": As long as you can, stick to the Rule of Zero, but if you have to write at least one of the Big Five, default the rest.

I can agree to that, I think it makes sense. Maybe the proper way to for this case would actually be to write a test to ensure that the move operators are present and use the "Rule of All or Nothing" so in that case we have clean code and if somebody decides to accidentally remove the move operator, it would trigger a test failure. Seems like the best of both worlds.

>  This PR is testing to see if this type of refactor would be accepted into the codebase in the future.

that sounds good and you caught us at a somewhat unprepared state where we have not come to complete conclusions in #3310 how to handle C++11 features. However, now would be a good time to conclude that discussion and your input is very welcome.

> `SvmTheoreticalSpectrumGeneratorSet` is a datatype (class) and not an interface[[1]](http://www.stroustrup.com/bs_faq2.html#virtual-dtor).

I agree in this case that removal of the virtual dtor is correct. It also makes class size larger and adds nothing. So I am happy to remove the virtual dtor, I just wanted to highlight how subtle differences can manifest in C++ when one is implicit instead of explicit ...

hroest:
>  "Python + bindings around C++ libraries" project.

I think it is that already.

As timo mentioned, we currently dont use pybind11 but autowrap which gives us a bit more control over the process. While it is some work to maintain autowrap and the bindings, we are currently not planning to switching to anything else (the benefits have not been discussed and there is nobody to do the work).

src/tests/class_tests/openms/source/SvmTheoreticalSpectrumGeneratorSet_test.cpp
```diff
@@ -65,17 +65,25 @@ START_SECTION(SvmTheoreticalSpectrumGeneratorSet(const SvmTheoreticalSpectrumGen
   NOT_TESTABLE //is tested in getSupportedCharges test
 END_SECTION
 
+START_SECTION(SvmTheoreticalSpectrumGeneratorSet(SvmTheoreticalSpectrumGeneratorSet&& source))
```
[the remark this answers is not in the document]
hroest: one could then add some more checks that `source` is in a valid state after the move and/or that really a move occured.

timosachsenberg:
hi feel free to join https://gitter.im/OpenMS/CodingConventions

hroest:
@rinmiEft please feel free to contact us on https://gitter.im/OpenMS/CodingConventions

PR #3755, opened 2018-09-24
Feature/move semantics

hroest:
- add noexcept keyword to move constructor (this is crucial for std::vector to work properly and actually move data instead of copying it)
- add move semantics to `DataValue`
- add tests for DataValue_test, MetaInfoInterface_test. and PeptideIdentification_test.cpp which explicitly test the move semantics
- use move semantics in `IdXMLFile` as proof of concept

Using the test file from https://github.com/OpenMS/OpenMS/pull/3750#issuecomment-423726499 I get some nice memory savings:

```
$ /usr/bin/time ../release_2.4/./bin/FileInfo -in E3035_HCV_LAB1.idXML 
Memory usage (loading idXML): 10520 MB (working set delta)
FileInfo took 02:15 m (wall), 02:15 m (CPU), 3.20 s (system), 02:11 m (user).

$ /usr/bin/time ../pr_3744/bin/FileInfo -in E3035_HCV_LAB1.idXML
Memory usage (loading idXML): 5327 MB (working set delta)

$ /usr/bin/time ./bin/FileInfo -in E3035_HCV_LAB1.idXML
Memory usage (loading idXML): 3168 MB (working set delta)
FileInfo took 01:55 m (wall), 01:55 m (CPU), 1.76 s (system), 01:53 m (user).
```

It looks like memory is reduced by a factor of 3 and execution speed is slightly faster (around 15%).

### DIFF
--- a/src/openms/source/DATASTRUCTURES/DataValue.cpp
+++ b/src/openms/source/DATASTRUCTURES/DataValue.cpp
@@ -172,7 +172,7 @@ namespace OpenMS
   }
 
   //--------------------------------------------------------------------
-  //                       copy constructor
+  //                   copy and move constructors
   //--------------------------------------------------------------------
   DataValue::DataValue(const DataValue& p) :
     value_type_(p.value_type_),
@@ -198,7 +198,19 @@ namespace OpenMS
     }
   }
 
-  void DataValue::clear_()
+  DataValue::DataValue(DataValue&& rhs) noexcept :
+    value_type_(rhs.value_type_),
+    unit_type_(rhs.unit_type_),
+    unit_(rhs.unit_),
+    data_(rhs.data_)
+  {
+    // clean up rhs 
+    rhs.value_type_ = EMPTY_VALUE;
+    rhs.unit_type_ = OTHER;
+    rhs.unit_ = -1;
+  }
+
+  void DataValue::clear_() noexcept
   {
     if (value_type_ == STRING_LIST)
     {
@@ -223,7 +235,7 @@ namespace OpenMS
   }
 
   //--------------------------------------------------------------------
-  //                      assignment operator
+  //                    copy and move assignment operators
   //--------------------------------------------------------------------
   DataValue& DataValue::operator=(const DataValue& p)
   {
@@ -262,6 +274,33 @@ namespace OpenMS
     value_type_ = p.value_type_;
     unit_type_ = p.unit_type_;
     unit_ = p.unit_;
+
+    return *this;
+  }
+
+  /// Move assignment operator
+  DataValue& DataValue::operator=(DataValue&& rhs) noexcept
+  {
+    std::cout << "DataValue Move assignemnt operator " << std::endl;
+    // Check for self-assignment
+    if (this == &rhs)
+    {
+      return *this;
+    }
+
+    // clean up *this
+    clear_();
+
+    // assign values to *this
+    data_ = rhs.data_;
+    value_type_ = rhs.value_type_;
+    unit_type_ = rhs.unit_type_;
+    unit_ = rhs.unit_;
+
+    // clean up rhs 
+    rhs.value_type_ = EMPTY_VALUE;
+    rhs.unit_type_ = OTHER;
+    rhs.unit_ = -1;
 
     return *this;
   }
--- a/src/openms/source/DATASTRUCTURES/DataValue.cpp
+++ b/src/openms/source/DATASTRUCTURES/DataValue.cpp
@@ -199,12 +199,13 @@ namespace OpenMS
   }
 
   DataValue::DataValue(DataValue&& rhs) noexcept :
-    value_type_(rhs.value_type_),
-    unit_type_(rhs.unit_type_),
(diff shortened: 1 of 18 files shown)


hroest:
I have now added 15bd0f for the time being and interestingly this improves performance:

```
$ /usr/bin/time ./bin/FileInfo -in /media/doc/tmp/openms_issues/E3035_HCV_LAB1.idXML
Memory usage (loading idXML): 3002 MB (working set delta)
FileInfo took 01:55 m (wall), 01:52 m (CPU), 1.85 s (system), 01:50 m (user).
```

so that is 5% less memory (or 150 MB) than before! No idea why though!

src/openms/source/DATASTRUCTURES/DataValue.cpp
```diff
@@ -198,7 +198,19 @@ namespace OpenMS
...
+    value_type_(rhs.value_type_),
+    unit_type_(rhs.unit_type_),
+    unit_(rhs.unit_),
+    data_(rhs.data_)
+  {
+    // clean up rhs 
+    rhs.value_type_ = EMPTY_VALUE;
+    rhs.unit_type_ = OTHER;
+    rhs.unit_ = -1;
+  }
```
hroest: this is a move constructor, the idea is that we simply "steal" the rhs stuff including all pointers and we store it in our object and then reset the rhs to an empty state. The object is now in a "a valid but unspecified state." 
https://stackoverflow.com/questions/7930105/does-moving-leave-the-object-in-a-usable-state
Note that similar to `clear_`, we still have the actual ptr / or data in `rhs.data_` and did not reset that

timosachsenberg: I still consider myself a move newbie so I have a few questions. Is this move constructor needed? I would assume a synthesized one (that uses memmove internally) would even be faster? Cleaning up rhs seems to be additional overhead and from what I read not necessary.

hroest: A simpler answer is here:  https://stackoverflow.com/questions/9965461/in-what-scenarios-should-i-expect-to-explicitly-need-to-implement-a-move-constru

Q: In what scenarios should I expect to explicitly need to implement a move constructor and move assignment 
A: If you find yourself implementing, any of:
-    destructor
-    copy constructor
-    copy assignment

hroest: > Cleaning up rhs seems to be additional overhead and from what I read not necessary.

yes, I think we have a few alternatives. We could go for either

- `rhs.value_type_ = EMPTY_VALUE;`
- `rhs.data_ = nullptr;`

and we dont really need to reset `rhs.unit_type_` and `rhs.unit_` so we could remove that

src/openms/source/METADATA/PeptideHit.cpp
```diff
@@ -81,17 +81,18 @@ namespace OpenMS
...
+    sequence_(std::move(source.sequence_)),
+    score_(std::move(source.score_)),
+    analysis_results_(std::move(source.analysis_results_)),
+    rank_(std::move(source.rank_)),
+    charge_(std::move(source.charge_)),
+    peptide_evidences_(std::move(source.peptide_evidences_)),
+    fragment_annotations_(std::move(source.fragment_annotations_))
   {
+    // see http://thbecker.net/articles/rvalue_references/section_05.html
     source.analysis_results_ = nullptr;
```
hroest: @timosachsenberg I think this is another example of what the default move operator would do and then what we need to do on line 96 where we need to ensure that the *ownership* is transferred for the ptr

timosachsenberg: ok I see, thanks!... to summarize: here be dragons  :P

hroest:
> looks pretty cool! +1

yes, but it will be quite a bit of work until we have transitioned all the classes.

src/openms/source/METADATA/MetaInfoInterface.cpp
```diff
@@ -53,7 +54,8 @@ namespace OpenMS
     }
   }
 
-  MetaInfoInterface::MetaInfoInterface(MetaInfoInterface&& rhs) :
+  /// Move constructor
+  MetaInfoInterface::MetaInfoInterface(MetaInfoInterface&& rhs) noexcept :
     meta_(rhs.meta_)
```
cbielow: should be `std::move(rhs.meta_)`, just to make sure its still correct when someone changes the type of meta_ at some point.

src/openms/source/DATASTRUCTURES/DataValue.cpp
hroest: > Cleaning up rhs seems to be additional overhead and from what I read not necessary.

I dont think there is much overhead and it does put the object into a nice, well defined empty state so that it could even be used again. Also, one may even argue that all three values above are in one 64 bit block of space (8 bytes), on some architectures setting all three values may even be *faster* since you can set a single 64 value for the whole block -- if the compiler is smart enough :-) Not that you would ever notice the difference

PR #3757, opened 2018-09-28
[FIX] removed some compile warnings on unused parameters in default

timosachsenberg:


timosachsenberg:
there seems to be a regression in https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=g++-7,os_label=ubuntu_1804/2393/testReport/junit/projectroot.src.tests.class_tests/openms/OpenPepXLLFAlgorithm_test/

timosachsenberg:
https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/OpenPepXLLFAlgorithm_test.cpp
there is a segmentation fault and a warning: AASequence::getMonoWeight: Mass for ResidueType 0 not defined for sequences of length 0.

timosachsenberg:
@hroest you made the move constructor = default. We have some pointers in AASequence so I guess that is the problem. Can you confirm that this can't be defaulted?

timosachsenberg:
Maybe because they are declared const * and not pointer to const?:
```
     std::vector<const Residue*> peptide_;	     
     const ResidueModification* n_term_mod_;
     const ResidueModification* c_term_mod_;
```

hroest:
hmm interesting, here everything worked on the CI side: https://github.com/OpenMS/OpenMS/pull/3755

hroest:
> @hroest you made the move constructor = default. We have some pointers in AASequence so I guess that is the problem. Can you confirm that this can't be defaulted?

hmm these ptrs should point to a global instance of the Residue / Residue Modification so my impression was that they would just get copied. So if you physically copy the pointers, you now have 2 AA objects that have the same ptrs but since they are not "owning" the ptrs but simply compute on them I dont see how that would be an issue.

timosachsenberg:
Me neither but it’s just a guess right now.

timosachsenberg:
on my PC it only crashes from time to time and might be related to parallelization... 
will make a debug build an hope that it still crashes.
or @enetz do you have an idea what the issue might be?
' ' '
(gdb) bt
#0  0x00007ffff6e70b96 in OpenMS::OPXLHelper::buildPeptideIDs(std::vector<OpenMS::PeptideIdentification, std::allocator<OpenMS::PeptideIdentification> >&, std::vector<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch, std::allocator<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch> > const&, std::vector<std::vector<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch, std::allocator<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch> >, std::allocator<std::vector<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch, std::allocator<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch> > > >&, unsigned long, OpenMS::MSExperiment const&, unsigned long, unsigned long) () from /home/sachsenb/OpenMS/openms-build/lib/libOpenMS.so
#1  0x00007ffff6e6233c in OpenMS::OpenPepXLLFAlgorithm::run(OpenMS::MSExperiment&, std::vector<OpenMS::FASTAFile::FASTAEntry, std::allocator<OpenMS::FASTAFile::FASTAEntry> >&, std::vector<OpenMS::ProteinIdentification, std::allocator<OpenMS::ProteinIdentification> >&, std::vector<OpenMS::PeptideIdentification, std::allocator<OpenMS::PeptideIdentification> >&, std::vector<std::vector<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch, std::allocator<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch> >, std::allocator<std::vector<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch, std::allocator<OpenMS::OPXLDataStructs::CrossLinkSpectrumMatch> > > >&, OpenMS::MSExperiment&) [clone ._omp_fn.0] () from /home/sachsenb/OpenMS/openms-build/lib/libOpenMS.so
#2  0x00007ffff516a435 in ?? () from /lib64/libgomp.so.1
#3  0x00007ffff4d29e25 in start_thread () from /lib64/libpthread.so.0
#4  0x00007ffff4a5734d in clone () from /lib64/libc.so.6
' ' '

timosachsenberg:
debug did not crash unfortunately

timosachsenberg:
@enetz could it be that the access to all_top_csms is not properly synchronized? It is in a critical section but this section but not the same as the one where all_top_csms is modified: https://github.com/OpenMS/OpenMS/blob/6d4d6b32633056b2fe1d7a92831570f58d174468/src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp#L888
I only checked it briefly but this seems to be a likely cause.

hroest:
What if you remove the parallelization? Does it stop crashing?

hroest:
@timosachsenberg can you try and name these two critical section the same so that they are mutually exclusive:
https://github.com/OpenMS/OpenMS/blob/6d4d6b32633056b2fe1d7a92831570f58d174468/src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp#L882-L891

https://github.com/OpenMS/OpenMS/blob/2e951a6e45bbe2674db0316e7757d62235dee652/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp#L937-L945

e.g. make both ` #pragma omp critical (all_top_csms_access) `

timosachsenberg:
yep, seems to work now. I'll push that change but would like to have @enetz to take a look at that if this is how it was intended.

hroest:
Pretty sure there is no negative side effects here by synchronization of
these

enetz:
:+1:  thanks, I did encounter this error before, but I could not reliably reproduce it and never figured out why this happened. I will be more careful with critical sections in the future.

PR #3758, opened 2018-09-28
Feature/move semantics

hroest:
for MSSpectrum


hroest:
seems like 

```
 -  line 78:  TEST_EQUAL(noexcept(MSSpectrum(std::declval<MSSpectrum&&>())),true): got 0, expected 1
```

works on clang and gcc, but not MSVS - can somebody with MSVS experience have a look? @cbielow @timosachsenberg

timosachsenberg:
For std::string as value type in container, it produces the correct result on VS 2015,17,18PRE
https://godbolt.org/z/yuKZZJ

src/openms/source/METADATA/PeptideHit.cpp
```diff
@@ -198,6 +198,11 @@ namespace OpenMS
     sequence_ = sequence;
   }
 
+  void PeptideHit::setSequence(AASequence&& sequence)
+  {
+    sequence_ = std::move(sequence);
```
timosachsenberg: could this assignment potentially be marked as throwing? This would probably call the move operator (and not the move constructor). See my comments below.

src/tests/class_tests/openms/source/MSSpectrum_test.cpp
```diff
@@ -72,6 +72,10 @@ MSSpectrum* nullPointer = nullptr;
 START_SECTION((MSSpectrum()))
   ptr = new MSSpectrum();
   TEST_NOT_EQUAL(ptr, nullPointer)
+
+  // Ensure that MSSpectrum has a no-except move constructor (otherwise
+  // std::vector is inefficient and will copy instead of move).
+  TEST_EQUAL(noexcept(MSSpectrum(std::declval<MSSpectrum&&>())), true)
```
cbielow: could this be implemented using a static_assert in the class itself?

hroest: > could this be implemented using a static_assert in the class itself?

yes, except MSVS 2015 will not work on this: https://godbolt.org/z/utMmsl (MSVS 2017 will work)

timosachsenberg: static_assert(noexcept(M(std::declval<M&&>())), "failed" ); works for me

src/openms/source/METADATA/PeptideHit.cpp
hroest: hmm it could since its not labeled `noexept` yes, but on the other hand its not used in a `std::vector`, right? I am not sure I get your point?

timosachsenberg: ok I think I get it now. Even if the elements of the underlying vector  would throw then this std::move would still perform a cheap e.g., std::swap

src/tests/class_tests/openms/source/ExperimentalDesign_test.cpp
```diff
@@ -65,6 +65,12 @@ START_SECTION(ExperimentalDesign())
 {
   ptr = new ExperimentalDesign();
   TEST_NOT_EQUAL(ptr, null_ptr)
+
+  // Ensure that ExperimentalDesign has a no-except move constructor (otherwise
+  // std::vector is inefficient and will copy instead of move).
+  TEST_EQUAL(noexcept(ExperimentalDesign(std::declval<ExperimentalDesign&&>())), true)
+  TEST_EQUAL(noexcept(ExperimentalDesign::MSFileSection(std::declval<ExperimentalDesign::MSFileSection&&>())), true)
+  TEST_EQUAL(noexcept(ExperimentalDesign::SampleSection(std::declval<ExperimentalDesign::SampleSection&&>())), true)
```
hroest: ok, I think I can reproduce this here: https://godbolt.org/z/VNob7-

it seems that MSVS cannot create default noexcept move constructors for classes with members using `std::map`, while it works for `std::vector`. I found a stackoverflow post about this regarding std::unordered_map (not quite the same): https://stackoverflow.com/questions/47604029/move-constructors-of-stl-containers-in-msvc-2017-are-not-marked-as-noexcept

src/openms/source/METADATA/PeptideHit.cpp
hroest: still not sure I understand what this has to do with throwing, the only thing this will do is to call

```
AASequence& operator=(AASequence&&);
```

which is default constructed (e.g. `AASequence& operator=(AASequence&&); = default`) and most likely will do the following (this seems most likely):

```    
peptide_ == std::move(rhs.peptide_);
n_term_mod_ == std::move(rhs.n_term_mod_);
c_term_mod_ == std::move(rhs.n_term_mod_);
``` 

which then will call the implementation of `std::vector<const Residue*>& operator=(std::vector<const Residue*>&&);` which is the one from `std::vector` - again it does not matter whether its noexcept or not, this will be called. Now there is an interesting question of whether it should be 

```
std::vector<T>& operator=(std::vector<T>&&);
--- or
std::vector<T>& operator=(std::vector<T>&&) noexcept;
```

see here, https://stackoverflow.com/questions/12332772/why-arent-container-move-assignment-operators-noexcept -- but again irrelevant for our case here. So what will  `std::vector<const Residue*>& operator=(std::vector<const Residue*>&&);` probably do? (Again its not `noexcept` so it does not even guarantee anything) From my understanding, it will simply swap the ptrs to memory of `*this` and `rhs`.

I think you may be confusing the move assignment and move constructors of the `std::vector` and their `noexcept`ness with the question of whether the elements of `std::vector<T>` have noexcept move assignment and move constructors (e.g. `T` has noexcept move assignment and move constructors ). This is relevant when the vector resizes.

src/tests/class_tests/openms/source/MSSpectrum_test.cpp
hroest: > could this be implemented using a static_assert in the class itself?

I just discussed this with Timo and we probably would want to do this as a macro. There are two places where this could go, either in the class or in the test file. There are pros and cons to both, of course it increases bloat if it goes into the class itself but it will also ensure that every future compilation of OpenMS will ensure that this is true and it adds to the "documentation" if every client can see how the class behaves. On the other hand, we do all of our testing in specialized files and this seems a logical place to do testing about assumptions of the class. We may actually end up testing more than just this, see for example https://philippegroarke.com/posts/2018/easy_defensive_programming/

hroest:
MSVS fails to produce a move constructor for `MSSpectrum`, probably because of 

- OpenMS/METADATA/Precursor.h has a `std::set<ActivationMethod>` member

src/tests/class_tests/openms/source/MSSpectrum_test.cpp
```diff
@@ -72,6 +89,24 @@ MSSpectrum* nullPointer = nullptr;
...
+  TEST_EQUAL(noexcept(M2(std::declval<M2&&>())), true)
+  TEST_EQUAL(noexcept(M3(std::declval<M3&&>())), true)
+
+  TEST_EQUAL(noexcept(InstrumentSettings(std::declval<InstrumentSettings&&>())), true)
+  TEST_EQUAL(noexcept(SourceFile(std::declval<SourceFile&&>())), true)
+  TEST_EQUAL(noexcept(AcquisitionInfo(std::declval<AcquisitionInfo&&>())), true)
+  TEST_EQUAL(noexcept(Precursor(std::declval<Precursor&&>())), true)
+  TEST_EQUAL(noexcept(Product(std::declval<Product&&>())), true)
+  TEST_EQUAL(noexcept(PeptideIdentification(std::declval<PeptideIdentification&&>())), true)
+  TEST_EQUAL(noexcept(DataProcessing(std::declval<DataProcessing&&>())), true)
```
hroest: only under MSVS:

```
checking (MSSpectrum()) ... 
 +  line 91:  TEST_NOT_EQUAL(ptr,nullPointer): got 000001FD6D4548C0, forbidden is 0000000000000000
 -  line 95:  TEST_EQUAL(noexcept(MSSpectrum(std::declval<MSSpectrum&&>())),true): got 0, expected 1
 +  line 98:  TEST_EQUAL(noexcept(M1(std::declval<M1&&>())),true): got 1, expected 1
 +  line 99:  TEST_EQUAL(noexcept(M2(std::declval<M2&&>())),true): got 1, expected 1
 -  line 100:  TEST_EQUAL(noexcept(M3(std::declval<M3&&>())),true): got 0, expected 1
 +  line 102:  TEST_EQUAL(noexcept(InstrumentSettings(std::declval<InstrumentSettings&&>())),true): got 1, expected 1
 -  line 103:  TEST_EQUAL(noexcept(SourceFile(std::declval<SourceFile&&>())),true): got 0, expected 1
 +  line 104:  TEST_EQUAL(noexcept(AcquisitionInfo(std::declval<AcquisitionInfo&&>())),true): got 1, expected 1
 -  line 105:  TEST_EQUAL(noexcept(Precursor(std::declval<Precursor&&>())),true): got 0, expected 1
 -  line 106:  TEST_EQUAL(noexcept(Product(std::declval<Product&&>())),true): got 0, expected 1
 +  line 107:  TEST_EQUAL(noexcept(PeptideIdentification(std::declval<PeptideIdentification&&>())),true): got 1, expected 1
 -  line 108:  TEST_EQUAL(noexcept(DataProcessing(std::declval<DataProcessing&&>())),true): got 0, expected 1
: failed
```

src/tests/class_tests/openms/source/MSSpectrum_test.cpp
```diff
@@ -46,6 +46,11 @@
 using namespace OpenMS;
 using namespace std;
 
+static_assert(OpenMS::Test::fulfills_rule_of_5<MSSpectrum>(), "Must fulfill rule of 5");
+static_assert(OpenMS::Test::fulfills_rule_of_6<MSSpectrum>(), "Must fulfill rule of 6");
+static_assert(OpenMS::Test::fulfills_fast_vector<MSSpectrum>(), "Must have fast vector semantics");
+static_assert(std::is_nothrow_move_constructible<MSSpectrum>::value, "Must have nothrow move constructible");
```
hroest: @cbielow @timosachsenberg what do you think of these? Here we can use `constexpr` functions to test these different properties

hroest:
@timosachsenberg @cbielow this should be ready to merge

hroest:
@timosachsenberg @cbielow merge?

PR #3762, opened 2018-10-02
[CLEAN] remove remnants of TBB and CUDA from build-system

cbielow:


timosachsenberg:
I recall that @hroest had some ideas regarding TBB and potential fixes for the threading issues in OpenMS. Apart from that, I am for removal of CUDA and TBB (if not used right now)

cbielow:
I think we need to weigh the maintenance effort of TBB (including it in contrib, make sure there are no conflicts with other dependencies) against the benefits of TBB vs. OpenMP.

hroest:
I am also fine with this for now, but we also have some ideas for using CUDA and TBB so some of that code may come back at some point. I hope that is ok, if we have a good use-case?

PR #3763, opened 2018-10-02
Swathbug

cbielow:
fix a bug introduced in 10c2d31eb71269fd7ed8b9633bce79dbe9ef74a4

when OpenSwath is run with `use_consensus == false`, the selection of the closest peak is wrong (best candidate is not updated - the last candidate is always chosen).

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
@@ -433,12 +433,7 @@ public:
         {
           double background{0};
           double avg_noise_level{0};
-          if ((peak_integration_ == "smoothed") && smoothed_chroms.size() <= k)
-          {
-            throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
-              "Tried to calculate background estimation without any smoothed chromatograms");
-          }
-          else if (background_subtraction_ == "original")
+          if (background_subtraction_ == "original")
           {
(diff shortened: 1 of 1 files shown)


cbielow:
> You mean `use_consensus == false`, right?

yes. corrected.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h
```diff
@@ -258,60 +258,62 @@ namespace OpenMS
...
-              l = picked_chroms[k].getFloatDataArrays()[PeakPickerMRM::IDX_LEFTBORDER][min_dist];
-              r = picked_chroms[k].getFloatDataArrays()[PeakPickerMRM::IDX_RIGHTBORDER][min_dist];
-              picked_chroms[k][min_dist].setIntensity(0.0); // only remove one peak per transition
+            PeakIntegrator::PeakArea pa_tmp = pi_.integratePeak(  // get the peak apex
+                picked_chroms[k],
+                picked_chroms[k].getFloatDataArrays()[PeakPickerMRM::IDX_LEFTBORDER][i], 
+                picked_chroms[k].getFloatDataArrays()[PeakPickerMRM::IDX_RIGHTBORDER][i]); 
+            if (pa_tmp.apex_pos > 0.0 && std::fabs(pa_tmp.apex_pos - peak_apex) < peak_apex_dist_min)
+            { // update best candidate
+              peak_apex_dist_min = std::fabs(pa_tmp.apex_pos - peak_apex);
```
hroest: thx, this was missing!

hroest:
sounds good, even better would be a regression test :-)

PR #3766, opened 2018-10-03
Isospec

hroest:
first approach to incorporate IsoSpec and fine-grained isotopic patterns into OpenMS. Some issues remain

- no separate library is built for IsoSpec (its only 500 kb, so its hardly worthwhile)
- currently the isotopic tables of ~~IsoSpec~~ OpenMS are used
- currently the sum formula parsing of ~~IsoSpec~~ OpenMS is used


timosachsenberg:
seem to be some discepancies on mac

michalsta:
Hi,

One quick thing I've noticed: you probably don't need the cwrapper.cpp/cwrapper.h stuff. These files are needed to provice an interface to use IsoSpec from the C programming language, and seeing as OpenMS is C++ they should not be necessary.

hroest:
> seem to be some discepancies on mac

It may just be the sorting, I now sort the output before testing.

hroest:
> One quick thing I've noticed: you probably don't need the cwrapper.cpp/cwrapper.h stuff. These files are needed to provice an interface to use IsoSpec from the C programming language, and seeing as OpenMS is C++ they should not be necessary.

yes, I already exclude a few of these here https://github.com/OpenMS/OpenMS/pull/3766/files#diff-e5d91bee7c44ad19971ed8bfd698ad84R48

```
// #include "misc.cpp"
// #include "spectrum2.cpp" // contains unix-specific code
// #include "cwrapper.cpp"
```

so the wrapper is not included - can you comment on what `spectrum2` is doing?

michalsta:
spectrum2 is some preliminary code for quickly producing binned spectra from the stream of infinitely-resolved configurations as produced by the Iso* classes (with support for multithreading) that we are developing. Ultimately it will probably be nice to have it in OpenMS, but for now feel free to rip it out, that part of the code is still firmly in the "experimental" stage for now.

hroest:
@timosachsenberg @cbielow it looks like this is working and ready for merging, all tests pass

hroest:
@cbielow I did some stress tests to check for memory and performance issues (this is using a value of `1e6`):

```
$ /usr/bin/time /home/hr/openmsall/builds/openms_trunk2/src/tests/class_tests/bin/IsoSpec_test
16.62user 0.00system 0:16.63elapsed 99%CPU (0avgtext+0avgdata 22868maxresident)k
$ /usr/bin/time /home/hr/openmsall/builds/openms_trunk2/src/tests/class_tests/bin/FineIsotopeDistribution_test 
28.95user 0.00system 0:28.95elapsed 99%CPU (0avgtext+0avgdata 24184maxresident)k
```

as you can see there are no memory leaks and the speed of `IsoSpec` is comparable to their implementation. However, `FineIsotopeDistribution` is about a factor 2 slower which is due to (i) using our own isotopic tables, (ii) copying of the data from two `std::vector` into a `std::vector<Peak1D>` and (iii) sorting. I think all three steps are necessary to provide a clean interface that is consistent with the other interfaces in OpenMS - even though I am not sure about the sorting as its not documented but I am pretty sure that people / clients would expect a sorted output?

PR #3767, opened 2018-10-03
[FIX] ASan

timosachsenberg:
disclaimer: edited online with web edit

### DIFF
--- a/src/openms/source/METADATA/MetaInfo.cpp
+++ b/src/openms/source/METADATA/MetaInfo.cpp
@@ -92,7 +92,7 @@ namespace OpenMS
     else
     {
       DataValue tmp = value;
-      index_to_value_.insert({index,  tmp});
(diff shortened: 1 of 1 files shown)


src/openms/source/METADATA/MetaInfo.cpp
```diff
@@ -78,13 +78,22 @@ namespace OpenMS
...
     // @TODO: check if that index is registered in MetaInfoRegistry?
-    index_to_value_[index] = value;
+    auto it = index_to_value_.find(index);
+    if ( it != index_to_value_.end()) 
+    {
+      it->second = value; 
+    }
+    else
+    {
+      DataValue tmp = value;
```
hroest: @timosachsenberg I think you should add some documentation / explanation for this here, otherwise it is unclear why this was done and will be reverted at some point

PR #3770, opened 2018-10-04
document why a copy is created

timosachsenberg:


hroest:
I suggest to add a test that fails in ASan mode, e.g. 100x pushing back of the same metaValue into the same vector

PR #3773, opened 2018-10-05
Rename RNPXLDeisotoper to Deisotoper

cbielow:
to make it more prominently visible (its universal)

+ some minor code restructuring

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
@@ -520,8 +520,10 @@ namespace OpenMS
           //
           // We should avoid oversubscribing the CPUs, therefore we use integer division.
           // -- see https://docs.oracle.com/cd/E19059-01/stud.10/819-0501/2_nested.html
+#ifdef _OPENMP
           int outer_thread_nr = omp_get_thread_num();
           omp_set_num_threads(std::max(1, total_nr_threads / threads_outer_loop_) );
+#endif
 #pragma omp parallel for schedule(dynamic, 1)
           for (SignedSize pep_idx = 0; pep_idx <= nr_batches; pep_idx++)
           {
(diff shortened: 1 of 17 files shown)


src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -523,7 +523,6 @@ namespace OpenMS
           int outer_thread_nr = omp_get_thread_num();
           omp_set_num_threads(std::max(1, total_nr_threads / threads_outer_loop_) );
```
timosachsenberg: sure that this does not result in any error if we have no OpenMP?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -582,9 +581,7 @@ namespace OpenMS
             // Step 4: write all chromatograms and features out into an output object / file
             // (this needs to be done in a critical section since we only have one
             // output file and one output map).
-#ifdef _OPENMP
-#pragma omp critical (osw_write_out)
-#endif
+            #pragma omp critical (osw_write_out)
```
timosachsenberg: if these are not needed... why did we have them in the first place?

cbielow: they are only needed if an omp function is used, such as `omp_get_num_threads()` (which is not defined if OMP is off). Precompiler directives do not require #idefs since they will be ignored if unknown.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
cbielow: and it seems I've overlooked a function :-/

timosachsenberg:
FYI: @enetz

PR #3777, opened 2018-10-05
[DOC] use x64 link flags for Win64 builds 

cbielow:
to avoid linker error `LNK1210: exceeded internal ILK size limit; link with /INCREMNTAL:NO`
(usually only during repeated incremental links during development)


timosachsenberg:
BTW do we need to add the hacktoberfest label?

timosachsenberg:
ah no: "Timo has completed the Hacktoberfest challenge!" ;)

PR #3778, opened 2018-10-05
[FIX] memleaks

hroest:


hroest:
@cbielow I did some stress tests to check for memory and performance issues (this is using a value of `1e6`):

```
$ /usr/bin/time /home/hr/openmsall/builds/openms_trunk2/src/tests/class_tests/bin/IsoSpec_test
16.62user 0.00system 0:16.63elapsed 99%CPU (0avgtext+0avgdata 22868maxresident)k
$ /usr/bin/time /home/hr/openmsall/builds/openms_trunk2/src/tests/class_tests/bin/FineIsotopeDistribution_test 
28.95user 0.00system 0:28.95elapsed 99%CPU (0avgtext+0avgdata 24184maxresident)k
```

as you can see there are no memory leaks and the speed of `IsoSpec` is comparable to their implementation. However, `FineIsotopeDistribution` is about a factor 2 slower which is due to (i) using our own isotopic tables, (ii) copying of the data from two `std::vector` into a `std::vector<Peak1D>` and (iii) sorting. I think all three steps are necessary to provide a clean interface that is consistent with the other interfaces in OpenMS - even though I am not sure about the sorting as its not documented but I am pretty sure that people / clients would expect a sorted output?

cbielow:
looks good to me, even though I think that many of the new/delete pairs here could be avoided if the objects would be created on stack. I dont see a reason why iso (for example) is created on the heap.
Should even be faster :) but thats not something we should adress. If the code runs fine and fast, lets keep it that way.

+1

PR #3779, opened 2018-10-05
 enable type query for an MSSpectrum 

cbielow:
[FEATURE] enable type query for an MSSpectrum 

minor:
[DOC, PERF] documentation and minor performance improvements for Deisotoper


timosachsenberg:
+1 can you make a brief search for already existing spectrum type deduction? Maybe some code duplication could be removed

PR #3780, opened 2018-10-06
Improvements for IsoSpec integration

michalsta:
Hi, ( @hroest )

This pull request fixes some of the issues you mentioned with copying memory memory around. It works by dropping the tabulator step, and generating straight into the structure that will be used everywhere else: the vector of Peak1D's. It simplifies the code a bit too.

This should hopefully bring the performance of OpenMS' implementation to the original one. There's still the big elephant in the room, that is the sorting - the algorithm itself is a rather fast (low hidden constant) O(n), while sorting is of course O(nlogn), and is always going to dominate, esp for larger molecules. I'd love to rip this out, or at least make it optional - controlled by a bool argument to the function? Any thoughts?

Also, included is a fix for a rather nasty bug I've uncovered while working on that.


src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
```diff
@@ -124,9 +114,8 @@ namespace OpenMS
       IP[i] = isotopeProbabilities[i].data();
     }
 
-    Iso* iso = new Iso(dimNumber, isotopeNr.data(), atomCounts.data(), IM, IP);
-    run_(iso);
-    delete iso;
+    Iso iso(dimNumber, isotopeNr.data(), atomCounts.data(), IM, IP);
+    return run_(iso);
```
hroest: this actually leaks memory (IM and IP are not deleted)

src/openms/thirdparty/IsoSpec/IsoSpec/isoSpec++.cpp
```diff
@@ -437,7 +439,9 @@ Lcutoff(_threshold <= 0.0 ? std::numeric_limits<double>::lowest() : (_absolute ?
             empty = true;
     }
 
-    maxConfsLPSum[0] = marginalResults[0]->getModeLProb();
+    if(dimNumber > 1)
+        maxConfsLPSum[0] = marginalResults[0]->getModeLProb();
```
hroest: great, will this fix our calculations for the C100 molecule and the issue we had with different compilers?

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
hroest: maybe you can change that into `std::vector<double*>` IM, IP;` which would fix the memory leak?

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
```diff
@@ -61,41 +62,30 @@ namespace OpenMS
...
+    return distribution;
   }
 
-  void IsoSpec::run(const std::string& formula)
+  std::vector<Peak1D> IsoSpec::run(const std::string& formula)
   {
-    Iso* iso = new Iso(formula.c_str());
-    run_(iso);
+    Iso iso(formula.c_str());
+    return run_(iso);
```
hroest: thanks, looks better!

src/openms/thirdparty/IsoSpec/IsoSpec/isoSpec++.cpp
michalsta: That definitely looks like it could be the underlying issue causing this. I'll have a look into this.

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
michalsta: *facepalm*

of course. Silly mistake. Strange though, I would have thought the compiler should be smart enough to warn about there being leftover code after the return statement... Oh well.

Actually, all we have to do is put the deletes before the return statement, and that's it. Long-term, yeah, adding methods that accept vectors in place of double** here is definitely landing in our TODO list - we're planning to finalize a 1.9.0 release in a few days, clean up the old/experimental cruft that's scattered around our codebase, and then we'll do a code dump on you ;) and fix this too.

michalsta: Fixed by commit ab8b6133ab02551d02325f5498c03402d24fc6f9

src/openms/thirdparty/IsoSpec/IsoSpec/isoSpec++.cpp
michalsta: okay, it seems to work with GCC, but I can't seem to get the OpenMS to compile with clang, so I can't really test it...

hroest:
@michalsta can you comment on whether this should also fix this issue which came up in our nightly tests: https://github.com/OpenMS/OpenMS/issues/3782

michalsta:
Right. I merged the fix (and my completely identical one too, which I made before I noticed you have done it too...). I have not enabled the tests on C100 molecule, as I can't check whether they work with clang or MSVC - can you have a look at them?

By the way - when you are going to start considering the interface of the IsoSpec class to be stable (and unchangeable)?

michalsta:
Regarding the issue #3782 - yes, that's absolutely it.

hroest:
@michalsta do you have some time to chat on gitter? https://gitter.im/OpenMS/OpenMS

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
```diff
@@ -61,41 +62,30 @@ namespace OpenMS
...
-    IsoThresholdGenerator* generator = new IsoThresholdGenerator(std::move(*iso), threshold_, absolute_, tabSize, hashSize); 
-    Tabulator<IsoThresholdGenerator>* tabulator = new Tabulator<IsoThresholdGenerator>(generator, true, true, true, true); 
-
-    int size = tabulator->confs_no();
+    IsoThresholdGenerator generator(std::move(iso), threshold_, absolute_, tabSize, hashSize); 
 
-    masses_.clear();
-    masses_.reserve(size);
-    copy(&tabulator->masses()[0], &tabulator->masses()[size], back_inserter(masses_));
+    std::vector<Peak1D> distribution;
```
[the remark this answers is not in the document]
hroest: I was thinking about that too, but I dont think we can. Since the loop is over a generator, the configurations are not generated yet and are produced on-the-fly, therefore I dont think we know the size beforehand. Maybe @michalsta can comment here whether I understand this correctly.

hroest:
@michalsta do you have some time to chat on gitter? https://gitter.im/OpenMS/OpenMS

> By the way - when you are going to start considering the interface of the IsoSpec class to be stable (and unchangeable)?

we (un)fortunately dont have a stable API in OpenMS and therefore dont require API/ABI compatibility between versions. However, it would be nice to have a stable interface until the next release end of the year.

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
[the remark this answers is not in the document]
michalsta: Yeah, that boils down to estimating the number of configurations above a given threshold in a cartesian product of multinomial distributions. We (me and Mateusz, the other IsoSpec developer) have given quite a lot of thought about that a while ago, we even got desperate enough to search the literature ;) but came up with nothing sensible. Not even an estimate that could be depended upon not to be off by several orders of magnitude in its pessimistic cases. 

So, for now, we have given that up. I think that solving this particular problem might be worth a publication by itself (not Nature obviously ;)

But for now, the only way we can compute the size is to go through the configuration space and count the configurations.

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
```diff
@@ -61,41 +62,30 @@ namespace OpenMS
...
 
-    probabilities_.clear();
-    probabilities_.reserve(size);
-    copy(&tabulator->probs()[0], &tabulator->probs()[size], back_inserter(probabilities_));
+    while(generator.advanceToNextConfiguration())
+        distribution.emplace_back(Peak1D(generator.mass(), generator.eprob()));
```
hroest: the only issues with this is that in OpenMS `Peak1D` stores the intensity in a float and not a double. So this means that we can only represent probabilities down to 1.17549e-038 with this interface. Maybe something to consider, especially in case somebody actually wanted access to the accurate numbers.

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpec.cpp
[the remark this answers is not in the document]
michalsta: ...aaand I just checked and actually that seems to be about 20% faster. That is, going through the configuration space just counting configurations w/o storing them, then reserving the proper size of the vector, then throwing away the previous iso, constructing a new one and generating into a pre-allocated vector is on my machine faster than just going through the config space once, storing them in vector, and depending on the vector to grow as needed. That would also have the advantage of leaving the backend storage of the vector at the exact size needed, without the up to 2x overhead.

But this is going to be dependant as hell on the machine memory bandwidth, cache sizes, and most importantly, the efficiency of the OS memory manager and malloc implementation, so for now let's leave it as it is. I'll try to do some more involved tests than just "works for me" (like on other machines and OSes), and make use of this knowledge in the next IsoSpec release (there is obviously lot of optimization on our side possible here, like reusing marginals, parsed formula, and so on for counting and computing). So for now, let's leave it as it is, and when we finalize IsoSpec 1.9.0 I'll make a code dump and change it.

michalsta:
Anyway, I really got to go now. We could probably chat on gitter on Tuesday. What timezone are you in? It's  GMT+2 for me.

michalsta:
Actually, regarding commit 835c908f1136f5d2ce298f629ad727e49b2f48c6... Should the IsoSpecWrapper.h file be added to the cmake file in includes?

hroest:
> Actually, regarding commit [835c908](https://github.com/OpenMS/OpenMS/commit/835c908f1136f5d2ce298f629ad727e49b2f48c6)... Should the IsoSpecWrapper.h file be added to the cmake file in includes?

yes, I think it should

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.cpp
```diff
@@ -51,51 +52,39 @@
 #include "IsoSpec/tabulator.cpp"
 
 using namespace std;
```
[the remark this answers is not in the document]
michalsta: Well, probably yes. Tabulators were meant mostly to make it easier to write the C/Python bindings, in C++ it's more convenient to use the generators directly.

hroest:
checks are successful, Windows failures are unrelated (Sirius):

```
    projectroot.src.tests.topp.TOPP_SiriusAdapter_3
    projectroot.src.tests.topp.TOPP_SiriusAdapter_3_out
```

michalsta:
All right, I've gotten IsoSpec main repo (version_1_9 branch) mostly to where I want it for 1.9.0 release functionality-wise, all that's left is testing and a bit of code de-uglification. So a tenative timeline is that I hope to be able to pull in the code to here after the weekend, and then finalize the PR.

hroest:
Does that include the layered code as well? Could we expose that in OpenMS
now?

michalsta:
Yes, the plan is definitely to expose the Layered generator too. I've pulled in newest code which contains it, all that's left is for Mateusz to write some more tests, I'll write the wrapper around the Layered version, and we should be good to go before the weekend. Hopefully.

I've also managed to mess up the commit history of this pull request, I'll just close it and open a new one when we're ready...

src/openms/thirdparty/IsoSpec/IsoSpec/platform.h
```diff
@@ -0,0 +1,81 @@
...
+
+#if !defined(ISOSPEC_BUILDING_PYTHON)
+#define ISOSPEC_BUILDING_PYTHON false
+#endif
+
+#if !defined(ISOSPEC_BUILDING_OPENMS)
+#define ISOSPEC_BUILDING_OPENMS false
+#endif
+
+#if defined(__unix__) || defined(__unix) || \
```
hroest: maybe this should be wrapped in a 
```suggestion
#ifndef ISOSPEC_TEST_GOT_MMAN
#if defined(__unix__) || defined(__unix) || \
```

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.cpp
```diff
@@ -32,70 +32,60 @@
...
+#include <OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h>
 
 #include <OpenMS/CONCEPT/Macros.h>
+#include <OpenMS/KERNEL/Peak1D.h>
 #include <iterator>
 #include <string>
 
+// Override IsoSpec's use of mmap whenever it is available
+#define ISOSPEC_TEST_GOT_SYSTEM_MMAN false
+#define ISOSPEC_TEST_GOT_MMAN false
```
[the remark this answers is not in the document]
michalsta: Yeah, a silly mistake. I actually meant for platform.h to work in such a way that the use of mmap can be overriden by a #define before including platform.h, but then I used wrong variables. This commit fixes it: 97a8d1a

And the reason for doing this in such a way (detection in platform.h whether appropriate variable has been defined, and setting it up only if not) is that I wanted it to be possible to drop-in a new version of IsoSpec verbatim into OpenMS, without the need to rememeber to rip out mmap on every update.

hroest:
Sounds good!

hroest:
It seems on Windows it now looks for the mman file:



8>c:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\openms\thirdparty\isospec\isospec\platform.h(28): fatal error C1083: Cannot open include file: 'mman.h': No such file or directory (compiling source file C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\openms\source\CHEMISTRY\ISOTOPEDISTRIBUTION\IsoSpecWrapper.cpp) [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\openms\OpenMS.vcxproj]

http://cdash.openms.de/viewBuildError.php?buildid=290501

michalsta:
Right. 20f294a hopfully fixes that. Also, I have reworked the wrapper's interface a bit, and plan to expose the Layered version in a similar fashion. The idea is that IsoSpecWrapper should be an abstract class, defining mainly the interface of the run() method, while different algorithms will subclass that and provide their own implementations, as well as constructors that accept the parameters specific for them.

I'm also kind of torn between tearing out the formula-based constructors in IsoSpec wrappers. They use the IsoSpec's element frequencies and formula parser, which is a sure way for the user to shoot himself in the foot if he's expecting the OpenMS ones to be used - on the other hand, ripping that out will make it a royal pain to write tests for the wrapper... Any thoughts on that (and the general direction this is going in?)

michalsta:
Hi and ping @hroest @timosachsenberg !

All right, I think I've done all I wanted to do conding-wise, and documentation-wise too, and the code should be ready for review/inclusion. I still need to write the tests for the two new wrapper classes, but I will need a few days before I can tackle that, .

The main changes are:
- IsoSpec is brought from "very very alpha, in transition between stable releases" state to release quality 
- ...which fixes a few bugs, some of them SIGSEGVs, and among them, issue #3782 
- I have changed the interface a bit, exposed the generator logic (this allows walking through large sets of isotopologues without the need to ever explicitly store them all in the memory). The old single-shot run() method is also available.

Some remaining issues are as follows: 
- I can't test it on all platforms OMS supports (MSVC, MacOSX) as I don't have them. It seems to pass automated tests now though...
- It might be worth to consider deprecating/eliminating the FineIsotopeGenerator class (beyond the doncumentation note). Right now it is a very thin shell included for compatibility, and all it does is pass its arguments to IsoSpecWrapper, and sort the result. It also encourages the use of IsoSpecThresholdWrapper (for compatibility), while most users should probably be using IsoSpecTotalProbWrapper (see docstrings for both for details)
- In theory IsoSpec is still at beta version, and you might be hesistant to pull in that, but we're planning to release the code as-is as 1.9.0 (for final, stable release we're just missing some more usage examples, a bit of cleanup in the Python bindings, etc. but nothing that would touch the C++ core that OMS uses). Still, even is we do end up still finding some bugs the code is much more stable than the alpha version that is in OMS right now. And anyway, it's not like us slapping the 1.9.0 label will magically make the code bug-free ;) 
- It is our plan to do a long-term maintenance of the code, in particular, to do a pull request for every IsoSpec version that we release to upgrade it. Also, feel free to contact us (me and Mateusz) whenever some issue comes up with that piece of code.
- As mentioned, the tests for IsoSpecOrderedWrapper and IsoSpecTotalProbabilityWrapper are not yet done

Anyway, I think it's (mostly) done (except for some more tests), so feel free to have a look at it and suggest changes. I'd love to get this chunk pulled in a relatively short order though, and do any necessary polishing in a subsequent PR, as this one fixes a handful of known SIGSEGVs... This PR should be pretty solid, definitely much better than the code that's in OMS devel branch right now. But of course, your decision. :)

timosachsenberg:
Hi @michalsta . First, thanks a ton for putting so much time and effort into this PR.
It is really appreciated. We are currently struggling a bit with our build infrastructure but I would be happy to merge this soon.

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -0,0 +1,357 @@
...
+      * @brief Run the algorithm
+      *
+      * This method will run the algorithm with parameters as set up by the constructor. It will return an
+      * IsotopeDistribution containing the observed configurations. The configurations are explicitly stored
+      * in memory, which may become a problem when considering some especially large distributions. If this,
+      * or (a rather small) performance overhead is a concern, then the generator methods (below) should be
+      * used instead.
+      *
+      * This method is provided for convience. As calling that method invalidates the object (the method should
+      * not be called again, nor anythong other than destroying the object should be done with it), the most common
```
timosachsenberg: anythong ;)

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -0,0 +1,357 @@
...
+      * usage pattern of IsoSpecWrapper classes with the run method is:
+      *
+      * IsotopeDistribution dist = IsoSpecWrapperSubclass(...).run();
+      * do something with dist;
+      *
+      * @note Calling this method invalidates the object! In future versions this limitation might be removed.
+      *
+      * @note This method should not be mixed with the generator methods - a given object should only ever have
```
timosachsenberg: ok I think this might be a bit surprising for the user but I guess it is well enough documented.

timosachsenberg: if the state handling of the objects get to complex to properly reset them I guess one could also consider two wrapper for generator and the in memory calculator

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -0,0 +1,357 @@
...
+      *
+      **/
+    virtual IsotopeDistribution run() = 0;
+
+    /**
+     * @brief Move the generator to a next isotopologue
+     *
+     * Advance the internal generator to the next isotopologue. The value returned determines whether the
+     * generator has been exhausted (that is, all eligible configurations have already been visited).
+     * It is invalid to call any other generator methods before the first call to nextConf(), as well as
```
timosachsenberg: hmm would then an interface like bool result = iso.getNextConf(Peak1D&) be better?
then one could write:
while(iso.getNextConf(p))
{
  do something with p
}

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -0,0 +1,357 @@
...
+             const std::vector<std::vector<double> >& isotopeMasses,
+             const std::vector<std::vector<double> >& isotopeProbabilities);
+
+    /**
+      * @brief Setup the algorithm to run on an EmpiricalFormula
+      *
+      **/
+  IsoSpecOrderedGeneratorWrapper(const EmpiricalFormula& formula);
+
+  virtual IsotopeDistribution run() override final
```
timosachsenberg: I think its fine but one could consider if it then makes sense to have a common base class

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
michalsta: Resetting is definitely in the pipeline for close future, but not there yet (it's not very complicated per se, but it'll be a pain to test to make sure we're not missing some edge cases). Splitting the generator and the one-shot run() versions into different objects is definitely sensible though, infact it's what we do in our Python bindings, I'l try to do something in this vein this evening. This will also resolve the issue with run() raising exception for Ordered - there just won't be a non-generator version of that.

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
michalsta: Right. Fixed ;)

michalsta:
Right, I've done the API rework and again the code is ready for next round of reviews/merging. 

Speaking about AUTHORS: sure ;) though, if you're not opposed, I think Mateusz Łącki should also be there: although the commits are from me, he's the co-author of the IsoSpec algorithm, and has been doing (behind the scenes) pretty much all of the testing of this wrapper ;)

also, feel free to consider me and Mateusz unofficial maintainers of that part of code and poke us whenever some issue comes up in that region :)

hroest:
looks good, maybe one question is whether the FineIsotopePatternGenerator should now use the total probability instead of the cutoff. Any comments?

timosachsenberg:
FineIsotopePatternGenerator is a very new addition so I guess we are pretty flexible in handling this situation. I currently prefer keeping the class and just adding a warning to the documentation that the other wrapper is more efficient.  I would also prefer to let the user choose a total probability (I guess that is what he usually wants)

michalsta:
OK, I've done it. It still uses the threshold by default, to avoid breaking backward compatibility, but there's an extra parameter now that allows the user to switch.

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -0,0 +1,475 @@
...
+
+#include <OpenMS/CHEMISTRY/EmpiricalFormula.h>
+#include <OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsotopeDistribution.h>
+
+
+// Override IsoSpec's use of mmap whenever it is available
+#define ISOSPEC_GOT_SYSTEM_MMAN false
+#define ISOSPEC_GOT_MMAN false
+
+#include <OpenMS/../../thirdparty/IsoSpec/IsoSpec/isoSpec++.h>
```
hroest:
```suggestion
#include <IsoSpec/isoSpec++.h>
```

this should work directly, right?

hroest:
FineIsotopePatternGenerator is useful because 

- it provides the same interface as the CoarseIsotopePatternGenerator
- it does not include any IsoSpec headers  / macros into the downstream code, so its a way of isolating the implementation from the interface

hroest:
@timosachsenberg @michalsta it looks ready to me, are there any updates from the IsoSpec side that we should we wait for or can we merge this?

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
michalsta: With this, the library compiles fine, but then the compiler can't find that header when compiling tests. There is probably a missing path or something somwhere in the build system, but I'm not really sure how to fix this, not a cmake guru by a long shot ;) so if you could give it a look that would be nice ;)

michalsta:
Yes, it's based on 1.9.1. With 1.9.1 we've basically done all that we wanted for the 1.9.X branch, no big updates pending, no bugs that we know of, so this would be a good point to merge. 

As for the future, barring any bugs found, the next IsoSpec-related pull request I'm planning to make for OMS is going to be when we release 2.0, but that's at least a few months from now...

PR #3783, opened 2018-10-07
[TEST] Fix/bugfix subisotope

hroest:
- currently a ptr to the *current* configuration was stored in "visited"
  which meant the *same* ptr was stored N times instead of storing the
  different configurations. This meant that visited.count(currentConf)
  would not work reliably.
- this PR is to test whether this fixes the issues with the calculations of C100


hroest:
more information and a detailed analysis is here: https://github.com/MatteoLacki/IsoSpec/pull/14

what matters most for us is that we now have the same results on all platforms.

timosachsenberg:
+1

PR #3787, opened 2018-10-08
[FIX] QT copy directory recursively (sirius_workspace)

oliveralka:
The directory was not copied correctly.
Now a QT based method was used. 

Here a StackOverflow solution was modified, not sure if the comment should be removed or if its good practice to point that out.

Note: std::rename will work on directories in C++17


timosachsenberg:
can you add some documentation?
also one might consider moving this to https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/SYSTEM/File.h

src/utils/SiriusAdapter.cpp
```diff
@@ -420,14 +466,8 @@ class TOPPSiriusAdapter :
...
       
-      // try to create directory if not present
-      if (!sw_dir.exists())
-      {
-        sw_dir.mkpath(sirius_workspace_directory.toQString());
-      }
-      
       // move tmp folder to new location
-      std::rename(tmp_dir.toStdString().c_str(), sirius_workspace_directory.c_str());
+      copyDirectoryFiles(tmp_dir, sirius_workspace_directory.toQString(), true);
```
timosachsenberg: Handle false

oliveralka: Please have a look at the new Version? 
Could you specify "handle false"?

src/utils/SiriusAdapter.cpp
```diff
@@ -421,8 +421,15 @@ class TOPPSiriusAdapter :
       sirius_workspace_directory = String(sw_dir.absolutePath());
       
       // move tmp folder to new location
-      File::copyDirRecursively(tmp_dir, sirius_workspace_directory.toQString());
-      LOG_WARN << "Sirius Workspace was moved to " << sirius_workspace_directory << std::endl;
+      bool copy_status = File::copyDirRecursively(tmp_dir, sirius_workspace_directory.toQString());
+      if (copy_status)
+      { 
+        LOG_WARN << "Sirius Workspace was successfully moved to " << sirius_workspace_directory << std::endl;
```
timosachsenberg: maybe LOG_INFO is better than LOG_WARN

src/openms/include/OpenMS/SYSTEM/File.h
```diff
@@ -82,6 +82,18 @@ namespace OpenMS
     */
     static bool rename(const String& from, const String& to, bool overwrite_existing = true, bool verbose = true);
 
+    /**
+       @brief Copy directory recuresively
```
cbielow: Typo, also below

src/openms/source/SYSTEM/File.cpp
```diff
@@ -140,6 +140,48 @@ namespace OpenMS
     return true;
   }
 
+  // https://stackoverflow.com/questions/2536524/copy-directory-using-qt
+  bool File::copyDirRecursively(const QString &fromDir, const QString &toDir, bool overwrite_existing)
+  {
+    QDir sourceDir(fromDir);
+    QDir targetDir(toDir);
+
```
cbielow: check for self-assignment?! i.e. what if fromDir==toDir? (probably requires a canonical path)

src/tests/class_tests/openms/source/File_test.cpp
```diff
@@ -169,6 +169,22 @@ START_SECTION((static bool isDirectory(const String& path)))
   TEST_EQUAL(File::isDirectory(OPENMS_GET_TEST_DATA_PATH("File_test_text.txt")),false)
 END_SECTION
 
+// make source directory and copy it to new location
+// check copy function and if file exists in target path
+START_SECTION(static bool copyDirRecursively(const QString &fromDir, const QString &toDir, bool overwrite_existing = true))
+  String source_name = File::getTempDirectory() + "/" + File::getUniqueName() + "/" + File::getUniqueName() + "/";
+  String target_name = File::getTempDirectory() + "/" + File::getUniqueName() + "/" + File::getUniqueName() + "/"; 
+  QDir sdir;
+  TEST_EQUAL(sdir.mkpath(source_name.toQString()), true);
```
cbielow: test with a real subdirectory. Thats the whole point, right?

src/utils/SiriusAdapter.cpp
cbielow: If you want to move, why copy?

oliveralka: That is a good point.
I thought that it may be a good idea to copy and still have a temporary Version if debug >= 2.
If that is not necessary I probably can use the rename function of QT to move the directory.

Will leave the copyDirRecursively in File class - if that is ok.

src/openms/source/SYSTEM/File.cpp
```diff
@@ -140,6 +140,54 @@ namespace OpenMS
     return true;
   }
 
+  // https://stackoverflow.com/questions/2536524/copy-directory-using-qt
+  bool File::copyDirRecursively(const QString &fromDir, const QString &toDir, bool overwrite_existing)
+  {
+    if (fromDir == toDir)
```
cbielow: you might want to use `QString QDir::canonicalPath()` here, since providing
`fromDir = /tmp/myPath/../` and ` toDir = /tmp/` is actually the same path, but simple string comparison would not detect that.

jpfeuffer:
@cbielow Are you ok with that? Merge?

src/openms/source/SYSTEM/File.cpp
```diff
@@ -140,6 +140,60 @@ namespace OpenMS
     return true;
   }
 
+  // https://stackoverflow.com/questions/2536524/copy-directory-using-qt
+  bool File::copyDirRecursively(const QString &fromDir, const QString &toDir, bool overwrite_existing)
```
cbielow: naming convention. Should be `from_dir` etc...

src/openms/include/OpenMS/SYSTEM/File.h
```diff
@@ -82,6 +82,18 @@ namespace OpenMS
...
     static bool rename(const String& from, const String& to, bool overwrite_existing = true, bool verbose = true);
 
+    /**
+       @brief Copy directory recursively
+       
+       Copies a source directory to a new target directory (recursive).
+       
+       @param fromDir Source directory
+       @param toDir Target directory
+       @param overwrite_existing Delete existing target, before copying
```
cbielow: it does not really delete the whole target, right? it just overwrites files which are also in the source.
Maybe that's worth adding to the docs.

src/openms/source/SYSTEM/File.cpp
```diff
@@ -140,6 +140,60 @@ namespace OpenMS
...
+
+
+    // make directory if not present 
+    if (!targetDir.exists())
+    {
+      targetDir.mkpath(toDir);
+    }
+  
+    // copy folder recurively
+    QFileInfoList fileInfoList = sourceDir.entryInfoList();
```
cbielow: ...

src/openms/source/SYSTEM/File.cpp
```diff
@@ -140,6 +140,60 @@ namespace OpenMS
...
+    foreach(QFileInfo fileInfo, fileInfoList)
```
cbielow: does this also work with standard C++11 now?, i.e. `for (QFileinfo& file_info : fileinfo_list)` ??

oliveralka: @cbielow thank you for your patience - will work on it.

cbielow:
do we actually really want the current "overwrite_existing" behaviour?
if its false, the recursive copy will stop when processing the first file which already exists in the target dir.
Maybe a user just wants to add files from a source_dir to a target_dir without overwriting. Thus we should keep processing...
Any opinions on this?

jpfeuffer:
Yes. I think there could be an enum that signals the behaviour, with entries: OVERWRITE, SKIP, CANCEL

jpfeuffer:
But why close the PR?

cbielow:
whops... the closing was accidental! I will interrogate certain cat's manners ...

timosachsenberg:
@cbielow don't know if the Hacktoberfest guys accept that excuse ;)

oliveralka:
... damn I thought I could open a new one ... :D

src/openms/include/OpenMS/SYSTEM/File.h
```diff
@@ -90,12 +90,18 @@ namespace OpenMS
...
+       CANCEL: Cancel the copy process if file already exists in target directory - return false.
+
        @param from_dir Source directory
        @param to_dir Target directory
-       @param overwrite_existing Overwrites files in target directory if already present
+       @param option Specify the copy option (OVERWRITE, SKIP, CANCEL)
        @return True on success
     */
-    static bool copyDirRecursively(const QString& from_dir, const QString& to_dir, bool overwrite_existing = true);
+    enum copy_options {OVERWRITE,SKIP,CANCEL};
```
timosachsenberg: - use enum class instead of enum
- it's a type so CopyOptions would be better

oliveralka:
@cbielow this should be it (hopefully)

src/openms/include/OpenMS/SYSTEM/File.h
```diff
@@ -82,6 +82,27 @@ namespace OpenMS
     */
     static bool rename(const String& from, const String& to, bool overwrite_existing = true, bool verbose = true);
 
+    /**
+       @brief Copy directory recursively
+       
+       Copies a source directory to a new target directory (recursive).
+       If the target directory already exists, files will be added.
+       If files already exist in the target, the copy will fail when processing the first duplicate (i.e. return false).
```
cbielow:
```suggestion
       If files from the source already exist in the target, @p option allows for the following behaviour:
```

src/openms/include/OpenMS/SYSTEM/File.h
```diff
@@ -82,6 +82,27 @@ namespace OpenMS
...
+       To avoid this, specify @p overwrite_existing = true.
```
cbielow:
```suggestion

```

src/openms/include/OpenMS/SYSTEM/File.h
```diff
@@ -82,6 +82,27 @@ namespace OpenMS
...
+       
+       Copy options:
```
cbielow:
```suggestion

```

src/openms/source/SYSTEM/File.cpp
```diff
@@ -140,6 +140,73 @@ namespace OpenMS
...
+      if (entry.isDir())
+      {
+        if (!copyDirRecursively(entry.filePath(), target_dir.filePath(entry.fileName()), option))
+        {
+          return false;
+        }
+      }
+      else
+      {
+        if (skip_existing && target_dir.exists(entry.fileName()))
```
cbielow: how about:
```
if (target_dir.exists(entry.fileName()))
{
  switch (option)
  {
   case CopyOptions::cancel: 
     return false;
  case CopyOptions::skip:
   continue;
  case CopyOptions::overwrite:
    target_dir.remove(entry.fileName());
  }
}
if (!QFile::copy....
```

cbielow: this way, you can remove the two bools.. less code and a bit cleaner

src/tests/class_tests/openms/source/File_test.cpp
```diff
@@ -169,6 +169,32 @@ START_SECTION((static bool isDirectory(const String& path)))
...
+  TEST_EQUAL(sdir.mkpath(source_name.toQString()), true)
+  // test canonical path
+  TEST_EQUAL(File::copyDirRecursively(source_name.toQString(),source_name.toQString()),false)
+  // test copy overwrite default
+  TEST_EQUAL(File::copyDirRecursively(source_name.toQString(),target_name.toQString()),true)
+  TEST_EQUAL(File::exists(target_name + "/pdata/1/proc"),true);
+  File::removeDirRecursively(target_name);
+  // test copy overwrite 
+  TEST_EQUAL(File::copyDirRecursively(source_name.toQString(),target_name.toQString(), File::CopyOptions::OVERWRITE),true)
+  TEST_EQUAL(File::exists(target_name + "/pdata/1/proc"),true);
```
cbielow: mhh.. how do you know the file was actually overwritten? Could have been skipped as well.
You'd have to check the content e.g. by emptying the content first (e.g. ofstream.open()+close()), then overwrite and check content again (or just size).

src/openms/source/SYSTEM/File.cpp
oliveralka: I have not thought about that - thanks!

oliveralka:
@cbielow thanks again for the corrections! 
Do you know how I am able to read out the LOG_WARN from the SKIP option in File_test;
Could use it to check if the file was skipped (additional to the ofstream for the overwrite option)

cbielow:
certainly possible, but I'd just go with filesize (or content).
If you want to go all the way, see LogStream_test on how to do it (i.e. append a stringstream to Log_warn should do it).

oliveralka:
@cbielow 
Introduced the changes - please let me know if there is an easier way to check the number of lines in a file.

oliveralka:
rebuild jenkins

PR #3788, opened 2018-10-08
Isotope dist

hroest:
- fixes getMin / getMax in the isotope distribution for the general case
- removes  isNormalized and isConvolutionUnit()
- updates docu
- adds tests


jpfeuffer:
Cool. Looks good. But I guess py bindings need to be updated?

PR #3789, opened 2018-10-09
[FEATURE] return transformation statistics

timosachsenberg:
make some basic statistics on deviations before/after applying the transformation model available to down-stream tools

PR #3790, opened 2018-10-11
Feature/td description

hroest:
replaces https://github.com/OpenMS/OpenMS/pull/3789


timosachsenberg:
cool, thanks for also adding the pyOpenMS output

PR #3791, opened 2018-10-12
[TEST] FeatureFinderIdentification: modified test case to trigger error with bracket mods

timosachsenberg:
bracket modifications (in this case I replaced Carbamidomethyl by two delta masses) crashes FFId

### DIFF
--- a/src/tests/topp/FeatureFinderIdentification_1_input.idXML
+++ b/src/tests/topp/FeatureFinderIdentification_1_input.idXML
@@ -78,11 +78,11 @@
 			</PeptideHit>
 		</PeptideIdentification>
 		<PeptideIdentification score_type="q-value" higher_score_better="false" significance_threshold="0" MZ="722.325378417969" RT="1736.66821289062" >
-			<PeptideHit score="0.0344827586206897" sequence="YIC(Carbamidomethyl)DNQDTISSK" charge="2" aa_before="K" aa_after="L" protein_refs="PH_14" >
+			<PeptideHit score="0.0344827586206897" sequence="YIC[47.021464]D[10.0]NQDTISSK" charge="2" aa_before="K" aa_after="L" prot="PH_14" >
 				<UserParam type="float" name="OMSSA_score" value="0.0173429321700537"/>
-				<UserParam type="string" name="target_decoy" value="target"/>
-			</PeptideHit>
-		</PeptideIdentification>
+				<UserParam type="string" name="target_decoy" value="target"/>           
+			</PeptideHit>                                                                   
+		</PeptideIdentification>                                                                
 		<PeptideIdentification score_type="q-value" higher_score_better="false" significance_threshold="0" MZ="443.711242675781" RT="1738.03344726562" >
 			<PeptideHit score="0.0344827586206897" sequence="DDSPDLPK" charge="2" aa_before="K" aa_after="L" protein_refs="PH_14" >
 				<UserParam type="float" name="OMSSA_score" value="0.00395109914742618"/>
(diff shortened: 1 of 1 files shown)


hroest:
if you insert the OpenSwath cout again, you get

```
Will analyse 26 peptides with a total of 52 transitions 
Error: Unexpected internal error (the element 'UniMod:-1' could not be found)
```

hroest:
I did some digging into this which was harder than expected since Hendrik turned off the output from OpenSWATH and the problem is here

hroest:
Btw: what are these mods? it seems like OpenMS cannot map them, when you read in the file this happens and I think from this all problems follow:

```
 String::ConstIterator AASequence::parseModSquareBrackets_(const String::ConstIterator str_it, 
 float mass tol: 1e-05
Warning: unknown modification '47.021464' of residue 'C' - adding it to the database
 String::ConstIterator AASequence::parseModSquareBrackets_(const String::ConstIterator str_it, 
 float mass tol: 1
Warning: unknown modification '10.0' of residue 'D' - adding it to the database
 String::ConstIterator AASequence::parseModSquareBrackets_(const String::ConstIterator str_it, 
```

timosachsenberg:
These are just artificial mods I created from converting Carbamidomethyl ~57 into two bracket mods in the original test file. But the problem is relevant if we want to quantify peptides with unknown modifications but known delta mass

timosachsenberg:
Thanks for looking into this. I guess a workaround would then be to register the mods with a pseudo unimod accession. Is there a cleaner way to do that? Hendrik mentioned in the code comment that it is currently not that straight forward.

hendrikweisser:
Nice work, Hannes - thanks for that!

> Hendrik turned off the output from OpenSWATH

Yes, if I remember correctly there was a lot of irrelevant (for FFId) output that I wanted to suppress. However, losing error messages is of course bad. Looking at the code [here](https://github.com/OpenMS/OpenMS/blob/f6e0426b3dfd8c41b71a8c1feca6e4b00bce809f/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp#L291), I see two options:
1. Make this call dependent on `debug_level_` (probably a good idea anyway).
2. In OpenSWATH, send errors to `LOG_ERROR` (only `LOG_INFO` should be suppressed in FFId).

> Is there a cleaner way to do that?

I guess implementing the proper `AASequence` to `Peptide`-with-mods conversion function would be a solution. I don't know if/how "unknown" mods can even be handled there, though. (@hroest?)
Is something like this perhaps implemented for mzTab already? There the sequence and mods are also stored separately.

hendrikweisser:
> Make this call dependent on debug_level_ (probably a good idea anyway).

Just did that (#3794). Should hopefully make it easier to debug problems next time.

hroest:
> I guess implementing the proper `AASequence` to `Peptide`-with-mods conversion function would be a solution. I don't know if/how "unknown" mods can even be handled there, though. (@hroest?)

The problem is that once we convert it to `LightModification` we only store the unimod id - so if there is no unimod id then there cannot be a mod https://github.com/OpenMS/OpenMS/blob/5a70018d9e03ce32e64fcbb1c985b7a1256efc7a/src/openswathalgo/include/OpenMS/OPENSWATHALGO/DATAACCESS/TransitionExperiment.h#L142-L146

I see these solutions

- simply set unimod id to -1 and skip those mods when converting (loss of data = bad)
- change the struct to
```C++
 struct LightModification 
 { 
   int location;
   int unimod_id;
   double massdiff; // only set if unimod_id == -1
}; 
```

which of course takes more memory (16 bytes instead of 8) but at least wont loose data. Note that a union with `unimod_id` and `massdiff` wont save memory either since we still need to align the first int and will introduce other complications.

hendrikweisser:
Another alternative would be along the lines of Timo's suggestion:
> register the mods with a pseudo unimod accession

This could happen automatically, so the user didn't have to edit the Unimod file.

timosachsenberg:
It is kind of an ugly solution and I would need to check if I need to readd the modification to ResidueDB with the new Unimod id (e.g., -112). By default the bracket mods get all added as Unimod id -1

PR #3795, opened 2018-10-15
[Fix] knime linux package

jpfeuffer:
Fixes linux packaging for KNIME by shipping all non-system shared libraries.


jpfeuffer:
Should go to release, too, otherwise KNIME package is broken.

PR #3799, opened 2018-10-16
[FEATURE] Updated OpenMP inclusion to modern target based CMake to si…

jpfeuffer:
…mplify macOS support for OMP.

Let's see how this goes on Win.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@
 # $Authors: Stephan Aiche, Julianus Pfeuffer $
 # --------------------------------------------------------------------------
 
-cmake_minimum_required(VERSION 3.0.2 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.1.0 FATAL_ERROR)
 project("OpenMS_host")
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jpfeuffer:
```
-- OpenMP support requested: ON
-- Found OpenMP_C: -openmp (found version "2.0") 
-- Found OpenMP_CXX: -openmp (found version "2.0") 
-- Found OpenMP: TRUE (found version "2.0") 
```
Seems to work on win. Any idea how to check if it was executing correctly?

timosachsenberg:
Note that omp_get_num_threads() must be called inside a parallel region. Otherwise, it always returns 1.

hroest:
yes, but maybe we can call a tool with `-threads N` and then in the tool test that `omp_get_max_threads()` returns N ?

timosachsenberg:
maybe in OpenMSInfo?

hroest:
> maybe in OpenMSInfo? We could output that information as text but also fail with non-zero exit code if OpenMS was enabled in the build system but the number is only 1

I think OpenMS info should rather be simply information, why not a separate "Multithreading_test.cpp" for this?

hroest:
looks good, can you replace tabs with 2 spaces?

jpfeuffer:
Yep right. I actually always thought it was enabled on Linux already. Good to know. Will update indents too.

PR #3800, opened 2018-10-17
Added fraction support to Proteinquantifier and MzTab

timosachsenberg:
+ some minor cleanup (removed the ProteinResolver as discussed with @jpfeuffer)

In preparation for the LFQ tool with complete export to PRIDE

ProteinQuantifier now internally stores a mapping: fraction -> charge -> sample -> abundance
before it only stored: charge -> sample -> abundance

- Fraction information is read from experimental design
- Peptide abundance summarization works as before but now uses the fraction and charge state with most quantitative values. Before it only used the charge state with most quantitative values.

The more controversial part might be how quantitative values are transferred to mzTab. Currently, it annotates the quantities in the protein group (added float data array).

TODO: update python bindings


src/openms/include/OpenMS/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.h
```diff
@@ -60,8 +61,8 @@ namespace OpenMS
     /// Quantitative and associated data for a peptide
     struct PeptideData
     {
-      /// mapping: charge -> sample -> abundance
-      std::map<Int, SampleAbundances> abundances;
+      /// mapping: fraction -> charge -> sample -> abundance
+      std::map<Int, std::map<Int, SampleAbundances>> abundances;
```
hroest: this may not wrap well in Python, you may have to do something like this
```suggestion
     struct FractionAbundances { std::map<Int, SampleAbundances> sample_abundances;};
      std::map<Int, FractionAbundances> abundances;
```

src/openms/include/OpenMS/METADATA/ProteinHit.h
```diff
@@ -179,7 +209,7 @@ namespace OpenMS
     String accession_;   ///< the protein identifier
     String sequence_;    ///< the amino acid sequence of the protein hit
     double coverage_;    ///< coverage of the protein based upon the matched peptide sequences
-
+    std::set<std::pair<Size, ResidueModification> > modifications_; ///< modified positions in a protein
```
hroest: wouldnt the "usual" way of doing this be
```suggestion
    std::set<std::pair<Size, const ResidueModification*> > modifications_; ///< modified positions in a protein
```

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.h
timosachsenberg: so far it has not been wrapped: https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/pxds/PeptideAndProteinQuant.pxd#L64
but I can add that.

timosachsenberg:
What is your feeling about the way of transporting the protein (group) quantities to the exporter?
I extended the protein group interface by data arrays:
https://github.com/OpenMS/OpenMS/pull/3800/files#diff-6d2d0be807a4ba165c36e57ad08f195eL76
And use these to store the quantities.

src/openms/include/OpenMS/METADATA/ProteinHit.h
```diff
@@ -171,6 +196,11 @@ namespace OpenMS
     /// sets the coverage (in percent) of the protein hit based upon matched peptides
     void setCoverage(const double coverage);
 
+    /// returns the set of modified protein positions
+    const std::set<std::pair<Size, ResidueModification> >& getModifications() const;
+
+    /// sets the set of modified protein positions
+    void setModifications(std::set<std::pair<Size, ResidueModification> >& mods);
```
hroest: you will also have to change the get/set Modifications fxns otherwise it probably wont compile

timosachsenberg:
merged via proteomicsLFQ branch

PR #3805, opened 2018-10-18
SwathQC output

cbielow:
some QC data cannot be crammed into the current output formats, such as featureXML/tsv.

This PR introduces a new json output file to OpenSwath, where arbitrary metrics can be stored in JSON format. For now, charge distribution in MS1. More metrics coming soon.

[FEATURE] add SwathQC file (Charge Distribution support for now)
[FEATURE] add -out_qc flag to OpenSwathWorkflow and OpenSwathFileSplitter
[FEATURE] add JSON to known filetypes


src/openms/include/OpenMS/FORMAT/DATAACCESS/SwathFileConsumer.h
```diff
@@ -631,7 +634,7 @@ namespace OpenMS
         addMS1Map_();
       }
       ms1_consumer_->consumeSpectrum(s);
-      s.clear(false);
+      ms1_map_->addSpectrum(s); // keep spectrum in memory
```
hroest: you cannot do that, this will keep all MS1 spectra in memory. We have data with MS1 maps > 50 GB and that is not going to work for us. Why dont you use a consumer that keeps data in memory and retrieve the data from there? There should actually be one that does this the `MSDataStoringConsumer.cpp` it will do that for you and you dont have to change this interface here!

src/openms/include/OpenMS/FORMAT/DATAACCESS/SwathFileConsumer.h
```diff
@@ -542,6 +542,9 @@ namespace OpenMS
    * map) objects of MSDataCachedConsumer which can consume the spectra and
    * write them to disk immediately.
    *
+   * Warning: only the MS1 swathmap will be available then calling retrieveSwathMaps()
+   *          for downstream use.
```
hroest: Hmmm I just checked this interface again and its a bit different than the rest as it will simply write out the spectra to mzML files and be done with it. Currently `retrieveSwathMaps` is not going to do what you would expect it to do and maybe the correct interface would be to have additional keep in memory parameters to specify whether MS1 or MS2 or both should be kept in memory.

src/openms/include/OpenMS/FORMAT/DATAACCESS/SwathFileConsumer.h
```diff
@@ -552,7 +555,7 @@ namespace OpenMS
     typedef MapType::SpectrumType SpectrumType;
     typedef MapType::ChromatogramType ChromatogramType;
 
-    MzMLSwathFileConsumer(String cachedir, String basename, Size nr_ms1_spectra, std::vector<int> nr_ms2_spectra) :
+    MzMLSwathFileConsumer(const String& cachedir, const String& basename, Size nr_ms1_spectra, const std::vector<int>& nr_ms2_spectra) :
```
hroest:
```suggestion
    MzMLSwathFileConsumer(const String& cachedir,
    const String& basename,
    Size nr_ms1_spectra, 
    const std::vector<int>& nr_ms2_spectra,
    const bool keep_ms1_for_retrieval,
    const bool keep_ms2_for_retrieval) :
```

maybe like this?

src/openms/include/OpenMS/FORMAT/DATAACCESS/SwathFileConsumer.h
hroest: this is going to be problematic for some cases, e.g. it is not clear that a "dump to disk" consumer will keep all MS1 data in memory. The issue is that we have SWATH data > 200 GB where the MS1 data alone is over 50 GB and this will clearly lead to issues if this is done. We may have to introduce a flag to regulate this behavior, probably both for MS1 and MS2

cbielow:
tests pass except 1 failing test (FineIsotopeDistribution_test.cpp:113:5: error: floating constant truncated to zero [-Werror=overflow]).

Jenkins is still down though... wait for it?

PR #3806, opened 2018-10-18
change warning to info as in precursor correction

timosachsenberg:


src/openms/source/FILTERING/CALIBRATION/PrecursorCorrection.cpp
```diff
@@ -230,7 +230,7 @@ using namespace std;
 
       if (count_error_highest_intenstiy != 0)
       {
-        LOG_WARN << "Error: The method highest_intensity_peak failed " << count_error_highest_intenstiy << "times.";
+        LOG_INFO << "Correction to the highest intensity peak failed " << count_error_highest_intenstiy << " times. Not changes were applied in these cases.";
```
hroest:
```suggestion
        LOG_INFO << "Correction to the highest intensity peak failed " << count_error_highest_intenstiy << " times. No changes were applied in these cases.";
```

?

PR #3808, opened 2018-10-18
[FIX] Error in OpenSwathWorkflow_21 in debug mode

### LINKED ISSUE #3764: OpenSWATHWorkflow fails in debug mode.
Caught in the nightlies.

```
FATAL: uncaught exception!
exception of type Precondition failed occured in line 440, function OpenSwath::SpectrumPtr OpenMS::filterByDrift(OpenSwath::SpectrumPtr, double, double) of /home/jenkins/ws/openms_nightly_supplement/fd57e9df/source/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
error message: input->getDriftTimeArray() != nullptr
```

jpfeuffer:
Fixes #3764 . Prevents successful nightlies. This should be the easy way according to tools suggestion. Not sure if it makes sense. Alternative is probably to add a name everywhere in the transitions.


timosachsenberg:
rebuild jenkins

hroest:
I think #3882 should work...

PR #3819, opened 2018-10-22
[ADD] Param: processors (SiriusAdapter)

oliveralka:
Added recently to SIRIUS.
May come in handy when processing locally.

edit:
Use OpenMP check number of processors.
If OpenMP is not available - SIRIUS default -> all possible CPUs
If OpenMS is available parameter "threads" used for the number of processors.
If threads was set too high -> reset to maximal numbers of processors.

@jpfeuffer  Is it OK, to abuse the threads parameter that way?

### DIFF
--- a/src/utils/SiriusAdapter.cpp
+++ b/src/utils/SiriusAdapter.cpp
@@ -167,7 +167,7 @@ protected:
     registerIntOption_("compound_timeout", "<num>", 10, "Time out in seconds per compound. To disable the timeout set the value to 0", false);
     registerIntOption_("tree_timeout", "<num>", 0, "Time out in seconds per fragmentation tree computation.", false);
     registerIntOption_("top_n_hits", "<num>", 10, "The number of top hits for each compound written to the CSI:FingerID output", false);
-
+    registerIntOption_("processors", "<num>", -1, "Number of cpu cores to use. If set to -1 Sirius uses all available cores.", false);
     registerFlag_("auto_charge", "Use this option if the charge of your compounds is unknown and you do not want to assume [M+H]+ as default. With the auto charge option SIRIUS will not care about charges and allow arbitrary adducts for the precursor peak.", false);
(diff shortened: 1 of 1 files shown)


jpfeuffer:
He @oliveralka . I think you do not need to depend on OpenMP here. It should suffice to just use the threads parameter and pass it to SIRIUS.

src/utils/SiriusAdapter.cpp
```diff
@@ -328,11 +342,15 @@ class TOPPSiriusAdapter :
...
                    << "--ppm-max" << ppm_max
                    << "--compound-timeout" << compound_timeout
-                   << "--tree-timeout" << tree_timeout 
+                   << "--tree-timeout" << tree_timeout
                    << "--quiet"
                    << "--output" << out_dir.toQString(); //internal output folder for temporary SIRIUS output file storage
 
-    // add flags
+    // check OpenMP
+    #ifdef _OPENMP
```
hroest: I dont understand this, this is highly surprising for users. If OpenMS was built without OpenMP support, then you cannot have multithreading in java? I dont really understand the logic of this, why not simply pass forward the `--threads` parameter?

oliveralka: Ok, I will do that.
So it does not matter if more threads are "called" than available - or is that not possible?

oliveralka: Ok, put in the case of SIRIUS you will specify the number of processors with the thread parameter.

oliveralka:
Solution without OpenMP.
Threads are regarded as processors for SIRIUS (Number of CPUs) - is that Ok in this context?

hroest:
>  Threads are regarded as processors for SIRIUS (Number of CPUs) - is that Ok in this context?

I think thats fine and usually the user (should) know best what s/he wants. Note that on a cluster where you have shared loads (e.g. 48 cores but you are assigned 12 of these), it is impossible to infer at runtime how many cores you are assigned and you simply need to *know* in advance and pass the correct parameter. Any changes of this could actually result in problems

Issue #3793, opened 2018-10-12
Identification View: fragmentation indicator only appears on second navigation


hroest:
I tried out the Identification View and noticed something odd: when opening the BSA example for the first time, I got this view: 
![screenshot from 2018-10-12 09-45-00](https://user-images.githubusercontent.com/464014/46873013-a5ff5d00-ce03-11e8-8248-19efd0758947.png)

however when going one peptide up and then *back to the same peptide* I suddenly got a nice fragmentation indicator on the top right that was not there before. I tried this a few times and it seems when opening the PSM for the *first time* its always absent, but always present for subsequent views:

![screenshot from 2018-10-12 09-45-19](https://user-images.githubusercontent.com/464014/46873082-ca5b3980-ce03-11e8-80a8-aec7328069ca.png)

also it seems like the colors are different the second time around. System: Linux 16.04 with Qt 5.5

hendrikweisser:
Two things I've noticed:
1. Initially a second layer will be created (see first screenshot), which is gone when the spectrum is revisited (second screenshot).
2. Clicking in a different cell in the same row in the "Identification view" table already has the update effect. It is not necessary to switch to a different row/spectrum.

hendrikweisser:
Interestingly, this problem doesn't occur on my RNA test data. I believe the difference is that there, peak annotations are available directly in the idXML input file. In contrast, in the BSA example, a theoretical spectrum is first generated from the peptide sequence and matches between peaks and ions are detected. (The second layer that comes up initially shows the whole theoretical spectrum, when activated.)

timosachsenberg:
I think if Fragment annotations are already present it works fine. If not they *should* be generated using the settings in the Id dialog preferences. Maybe just an updating issue...

hendrikweisser:
@timosachsenberg:
> If not they should be generated using the settings in the Id dialog preferences.

Yes, that happens, there's just something weird about how they are displayed (initially).

timosachsenberg:
I think a proper solution is to generate the fragment annotations once (e.g., when the id file is annotated).
Currently its done on the fly

timosachsenberg:
@enetz do you think that creating the fragment annotations in addTheoreticalSpectrumLayer_ in TOPPViewIdentificationViewBehavior is a viable solution?

hendrikweisser:
Looks like this got fixed in #3821.

Issue #3792, opened 2018-10-12
TOPPView: "Identification view" may require extra prodding


hendrikweisser:
The "Identification view" in TOPPView is not working properly any more for me. Running "Annotate with identification" on an mzML file still works and populates the "Identification view" window on the right, but clicking on a row in the "Identification view" table does nothing. (It used to show the corresponding MS2 spectrum or precursor ion.)

This is with the current "develop" version on Ubuntu 18.04.1 with Qt 5.9.5. The problem can be reproduced e.g. with the "BSA" example data.

Possibly related to #3424, where another TOPPView feature stopped working.

hroest:
I have tried this on Ubuntu Linux 16.04 with Qt 5.5 and it seems to work fine for me. I did notice a minor bug #3793 though but apart from that it seems to work.

hendrikweisser:
I tested an older OpenMS revision (b3c4e7 from 9 April) and with that it works. So luckily it seems to be a recent OpenMS bug, not (entirely) a problem with the Qt version.

hendrikweisser:
I'll try to find the commit that caused the problem.

hendrikweisser:
It appears that there's a workaround again: After the annotation, if you click on the "Scan view" and then back to the "Identification view", it works (with the strange caveat that Hannes found in #3793).

hendrikweisser:
Git bisect result (@timosachsenberg):

```
9cf4fdcfa7d9624b54ca6ffb6af4746bbb464a49 is the first bad commit
commit 9cf4fdcfa7d9624b54ca6ffb6af4746bbb464a49
Author: Timo Sachsenberg <sachsenb@informatik.uni-tuebingen.de>
Date:   Fri May 18 12:31:25 2018 +0200

[FIX] don't create extra layer but directly annotate to observed spectrum

:040000 040000 bf0a999a955c165103eed29cb8a358fb06d29947 3d9fd88dbc0c0a95e8835498eb3a89a27fd71f6d M	src
```

Since that commit, TOPPView switches to the "Identification view" tab automatically after the "Annotate with identification" step, but (under Qt 5.9.5) clicking on a row in the table doesn't show the annotated spectrum any more.

PR #3821, opened 2018-10-22
[FIX] TOPPView IDs with theoretical spectra

enetz:
Because of synchronization between peak labels and PeptideHits,
PeakAnnotations were generated after matching to Theoretical Spectra
and these were kept and used after the second time an ID was chosen.
This lead to a changes in the ID view depending on whether you look at an ID for the first time.
Now the labels are synchronized and the theoretical spectrum deleted immediately
and the PeakAnnotations stored in the PeptideHits used right away.
This leads to the sequence coverage being shown immediately and a more consistent representation of IDs.

(Btw. the changes look more severe than they are, the content of `activate1DSpectrum` did not change after line 345. I just removed an `else{...}` statement around a big chunk and changed the indent associated with that)


hroest:
does this fix https://github.com/OpenMS/OpenMS/issues/3793 ? and maybe also https://github.com/OpenMS/OpenMS/issues/3792 ?

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
```diff
@@ -330,112 +330,121 @@ namespace OpenMS
             {
               // if no fragment annotations are stored, create a theoretical spectrum
               addTheoreticalSpectrumLayer_(ph);
+
+              // synchronize PeptideHits with the annotations in the spectrum
+              current_layer.synchronizePeakAnnotations();
+              // remove labels and theoretical spectrum (will be recreated using PH annotations)
+              removeGraphicalPeakAnnotations_(spectrum_index);
```
timosachsenberg: are these changes compatible with user added annotations?

enetz:
yes, this should fix #3793.
I am not sure about #3792, as I am not sure I can reproduce this.
What happens for me is, after reading in ID data and clicking on the first ID, nothing happens.
But changing from one ID to another triggers the annotations consistently now.

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
enetz: this is inside the `if (ph.getPeakAnnotations().empty())` code block, so it will be run only once per PeptideHit using the theoretical spectrum, when there are no PeakAnnotations there yet,
and before the user can add any annotations manually.
Immediately after that, the behaviour should be identical to reading in an ID file with PeakAnnotations.

hroest:
>  What happens for me is, after reading in ID data and clicking on the first ID, nothing happens.

that is still not great, but better than before :-) maybe @hendrikweisser can comment whether this fixes his problem as well on Qt 5.9 -- @enetz which Qt are you running?

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
```diff
@@ -1055,6 +1064,29 @@ namespace OpenMS
     }
   }
 
+  void TOPPViewIdentificationViewBehavior::removeGraphicalPeakAnnotations_(int spectrum_index)
+  {
+    Spectrum1DWidget* widget_1D = tv_->getActive1DWidget();
+    LayerData& current_layer = widget_1D->canvas()->getCurrentLayer();
```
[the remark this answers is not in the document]
enetz: I don't think so. I copied this section from code above in the function, where that is tested. It seems after generating the PeptideHit PeakAnnotations I have to do this to update the currently used PH, but it should already be clear, that it isn't nullptr from an earlier check before the label synchronization. If you think that that is not guaranteed, I can add another check.

enetz:
@hroest I have Qt 5.10.1 on an Ubuntu 14.04.5 LTS system.

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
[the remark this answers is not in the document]
timosachsenberg: I would test it

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
enetz: I experimented around with user annotations. After they are created from the theoretical spectrum, they are treated as peak annotations that were read in from an ID file.
They will be saved when saving a file, together with any user added annotations (or excluding any annotations the user deleted). Deleting all annotations in a spectrum will trigger the automatic annotation again, when the ID is reselected.

timosachsenberg: Ok so this means that user annotations are preserved if spectra are changed

hendrikweisser:
> maybe @hendrikweisser can comment whether this fixes his problem as well on Qt 5.9

Will do, but it may take me some time. I'm currently running time-consuming analyses for a paper and therefore can't rebuild OpenMS at the moment.

src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp
enetz: Yes,
that line `removeGraphicalPeakAnnotations_(spectrum_index)` is only applied to the labels generated by the theoretical spectrum. A second set is recreated afterwards from the PeptideHit annotations, so if I don't delete them here, we would end up with two sets of the same labels.
It is a bit of unnecessary computing: creating labels, making annotations from these labels, deleting labels, recreating them from annotations.
But this way we can use the existing code for annotated IDs and this is only done once per PeptideHit.

timosachsenberg:
works for me now

Issue #3825, opened 2018-10-25
SpectrumLookup::extractScanNumber wiff-format


oliveralka:
```
[Term]
id: MS:1000770
name: WIFF nativeID format
def: "Native format defined by sample=xsd:nonNegativeInteger period=xsd:nonNegativeInteger cycle=xsd:nonNegativeInteger experiment=xsd:nonNegativeInteger." [PSI:MS]
is_a: MS:1000767 ! native spectrum identifier format
```

Right now:
```
    // "experiment=NUMBER"
    else if (native_id_type_accession == "MS:1000770")
    {
      regexp = std::string("experiment=(?<GROUP>\\d+)");
    }
```

It seems that "cycle" would be the right expression (see Fig)

![spectrumlookup_ms1000770_wiff](https://user-images.githubusercontent.com/15729190/47488020-dc959880-d843-11e8-8449-abe3ffd010f7.png)

Affected test:
SpectrumLookup
SiriusAdapter
OpenSWATH (?)

timosachsenberg:
@hroest

hroest:
what is it that you want to extract? why not simply use the scan index? Neither the `cycle` nor `experiment` are unique numbers in my data. My data it looks like this:

```
<spectrum index="430" id="sample=1 period=1 cycle=12 experiment=14" defaultArrayLength="123">
```

so I guess a unique number would be a combination of `cycle` and `experiment` as the experiment counts up to `N` and then rolls over, a new cycle starts and experiment gets set to 1 again. Unless each cycle has the same number of experiments, there is no way to compute a unique number from this. 

Why not use the scan index in each mzML file?

hendrikweisser:
> Maybe @hendrikweisser can explain why this design was chosen in SpectrumLookup and what the use-case is to look up spectra by scan number rather than by index?

It makes a difference once you deal with spectra that may be pre-filtered. The index just applies to the current set of spectra, while the scan number (as part of the native ID) should stay constant even if you filter. (Merging is more problematic and not really considered.)

hroest:
> It makes a difference once you deal with spectra that may be pre-filtered. The index just applies to the current set of spectra, while the scan number (as part of the native ID) should stay constant even if you filter. (Merging is more problematic and not really considered.)

ok, that makes sense. However in this case a single number is not sufficient to describe the spectra and at least 2 numbers are required for SCIEX data. There are 2 ways out of this

- disable query by scan number
- allow query by a string or multiple ints

[...]

PR #3826, opened 2018-10-25
[FIX] SpectrumLookup::extractScanNumber wiff regex

oliveralka:
see issue #3825


oliveralka:
rebuild jenkins

timosachsenberg:
I see. So I guess if a single index is not enough we can't trim down the spectrum identifier in our ID data structures down to an int. Which might contribute to some extent to the size of the id data.

hroest:
@timosachsenberg see here: https://github.com/OpenMS/OpenMS/issues/3825#issuecomment-433759993 -- we basically need 2 ints or a smart way to encode this in a single int ...

hroest:
@oliveralka any progress on this?

oliveralka:
@hroest May you have a look - when you have time?

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -311,11 +319,27 @@ namespace OpenMS
           return -1;
         }
       }
+      else if (matches.size() == 2)
+      {
+        try
+        {
+          int value = String(matches[0]).toInt() * 1000 + String(matches[1]).toInt();
```
hroest: TODO: check that `String(matches[1]).toInt() < 1000` and warn/abort/fail if not.

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -258,16 +258,19 @@ namespace OpenMS
...
+    auto subgroups = {1};
     
     // "scan=NUMBER" 
     if (std::find(scan.begin(), scan.end(), native_id_type_accession) != scan.end())
     {
       regexp = std::string("scan=(?<GROUP>\\d+)");
     }
-    // "experiment=NUMBER"
+    // id="sample=1 period=1 cycle=96 experiment=1" - this will be described by a combination of (cycle * 1000 + experiment)
     else if (native_id_type_accession == "MS:1000770")
```
hroest:
```suggestion
    else if (native_id_type_accession == "MS:1000770") // WIFF nativeID format
```

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -296,13 +299,18 @@ namespace OpenMS
...
     if (!regexp.empty()) 
     {
-      boost::smatch match;
-      bool found = boost::regex_search(native_id, match, regexp);
-      if (found && match["GROUP"].matched)
-      { 
+      vector<string> matches;
+      boost::sregex_token_iterator current_begin(native_id.begin(), native_id.end(), regexp, subgroups);
+      boost::sregex_token_iterator current_end(native_id.end(), native_id.end(), regexp, subgroups);
+      for (auto it = current_begin ; it != current_end; ++it)
```
timosachsenberg: a.insert(a.end(), b.begin(), b.end());

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -296,13 +299,18 @@ namespace OpenMS
...
+      {
+        matches.push_back(*it);
+      }
+      if (matches.size() == 1)
```
timosachsenberg: "default case: one native identifier"

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -311,11 +319,34 @@ namespace OpenMS
           return -1;
         }
       }
+      else if (matches.size() == 2)
```
timosachsenberg: "special case: wiff file with two native identifiers"

src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -311,11 +319,34 @@ namespace OpenMS
...
+      {
+        try
+        {
+          if (String(matches[1]).toInt() < 1000)
```
timosachsenberg: add comment

timosachsenberg:
+1

PR #3827, opened 2018-10-25
release/2.4.0 fixed OpenPepXL parallelization errors

enetz:
race conditions caused crashes, when LOG_DEBUG was written in from different threads.
Now all LOG_DEBUG writing operations are in critical sections and only compiled and run if they are activated using the `#define DEBUG_...`and `#ifdef DEBUG_...`conditions.

### DIFF
--- a/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
+++ b/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
@@ -38,7 +38,7 @@
 #include <OpenMS/CONCEPT/LogStream.h>
 
 // turn on additional debug output
-#define DEBUG_OPXLHELPER
+//#define DEBUG_OPXLHELPER
 
 
 using namespace std;
--- a/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
+++ b/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
@@ -37,13 +37,16 @@
 #include <OpenMS/CHEMISTRY/ModificationsDB.h>
 #include <OpenMS/CONCEPT/LogStream.h>
 
+// turn on additional debug output
+#define DEBUG_OPXLHELPER
+
 
 using namespace std;
 
 namespace OpenMS
 {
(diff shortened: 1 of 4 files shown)


enetz:
so essentially, most of the debug output is turned off now to avoid race conditions or many critical sections

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -37,25 +37,23 @@
...
     // initialize empty vector for the results
     vector<OPXLDataStructs::XLPrecursor> mass_to_candidates;
 
     double min_precursor = spectrum_precursors[0];
     double max_precursor = spectrum_precursors[spectrum_precursors.size()-1];
 
-// Multithreading options: schedule: static, dynamic, guided
-// use OpenMP to run this for-loop on multiple CPU cores
-#ifdef _OPENMP
-#pragma omp parallel for schedule(static)
```
timosachsenberg: Why was that removed?

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -195,16 +193,9 @@ namespace OpenMS
     if (low_it != up_it) // if they are not equal, there are matching precursors in the data
     {
       // found_matching_precursors = true;
-
-// don't access this vector from two processing threads at the same time
-#ifdef _OPENMP
-#pragma omp critical
```
timosachsenberg: Why not needed anymore

enetz: related to the removed nested parallel loop. Since the function itself is not parallel anymore and different calls to this function don't access the same instances of these vectors, there is no need for a critical section

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
timosachsenberg: make sense

timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=290498

PR #3829, opened 2018-10-27
[FIX] Fixes concurrency issues for LogStream on Mac and avoids access…

jpfeuffer:
…ing omp symbols in dtor in case of premature closing.

### DIFF
--- a/src/openms/source/CONCEPT/LogStream.cpp
+++ b/src/openms/source/CONCEPT/LogStream.cpp
@@ -80,10 +80,7 @@ namespace OpenMS
 
     LogStreamBuf::~LogStreamBuf()
     {
-      sync();
-#ifdef _OPENMP
-#pragma omp critical
-#endif
+      syncLF_();
       {
         clearCache();
         if (incomplete_line_.size() > 0)
@@ -173,7 +170,6 @@ namespace OpenMS
     std::string LogStreamBuf::addToCache_(std::string const & line)
     {
       std::string extra_message = "";
-
       if (log_cache_.size() > 1) // check if we need to remove one of the entries
       {
         // get smallest key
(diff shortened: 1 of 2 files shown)


src/openms/source/CONCEPT/LogStream.cpp
```diff
@@ -425,8 +427,9 @@ namespace OpenMS
     void LogStreamNotifier::registerAt(LogStream & log)
     {
       unregister();
-
-      registered_at_ = &log;
+      {
```
timosachsenberg: why the brackets?

jpfeuffer: I probably had a pragma there before.

hroest:
rebuild jenkins

src/openms/source/CONCEPT/LogStream.cpp
```diff
@@ -80,10 +80,7 @@ namespace OpenMS
 
     LogStreamBuf::~LogStreamBuf()
     {
-      sync();
-#ifdef _OPENMP
-#pragma omp critical
-#endif
+      syncLF_();
```
hroest:
```suggestion
      // Prevent issue on OSX with OpenMP: destructors of global objects seem to be called after tearing down the OpenMP context, we therefore cannot use any locks here.
      syncLF_();
```

src/openms/source/CONCEPT/LogStream.cpp
hroest:
```suggestion
```

src/openms/source/CONCEPT/LogStream.cpp
```diff
@@ -425,8 +427,9 @@ namespace OpenMS
     void LogStreamNotifier::registerAt(LogStream & log)
     {
       unregister();
-
-      registered_at_ = &log;
+      {
+        registered_at_ = &log;
+      };
```
hroest:
```suggestion
```

hroest:
looks, good -- would be even better if the tests on OSX would run to test whether it works

jpfeuffer:
Thanks! Will adapt the changes. Jenkins seemed to be  down on the weekend.

jpfeuffer:
rebuild jenkins

timosachsenberg:
Looks good to me

Issue #3687, opened 2018-08-17
OpenMS 2.4 is now available on the Debian repositories


lopippo:
Greetings,

just to let you know that OpenMS is now available in Debian (sid) as version 2.4.

Thanks for such an awesome mass spec package.

Cheers,
Filippo

lopippo:
Update:

the build of OpenMS fails on the armhf platform. If anybody is interested in fixing the build this is the log url:

https://buildd.debian.org/status/fetch.php?pkg=openms&arch=armhf&ver=2.4.0-1&stamp=1534525425&file=log

Cheers,
Filippo

AdrianBunk:
On the armel and armhf ports in Debian (and on the armhf and arm64 ports in Ubuntu) Qt5 is compiled with OpenGL ES.

The only way to fix these builds would be to make (and keep) all OpenGL code in OpenMS compatible with OpenGL ES.

DarioS:
How is it possible if it hasn't been released as a source version on GitHub nor the OpenMS website?

lopippo:
Is not this enough ?

https://github.com/OpenMS/OpenMS/tree/release/2.4.0

DarioS:
I'm not sure but I don't see it in the [release listing](https://github.com/OpenMS/OpenMS/releases) and it appears to be only [28% complete](https://github.com/OpenMS/OpenMS/milestone/13).

lopippo:
What is that "28% complete" metric ? This is odd, since the build went flawlessly. Hmm, can anybody tell if the release/2.4.0 nomenclature actually corresponds to a release that is ready or not ? If not, then that is misleading.

hroest:
Dear Dario and Filippo

The official releases of OpenMS are here:
https://github.com/OpenMS/OpenMS/releases while what Filippo was using
was a "branch" in git which was forked from the main development
branch some weeks ago, however it does not mean that this is a
finished and tested release. I suggest to use the above link to obtain
official releases and monitor the open-ms-announcements list to obtain
information about new releases.

The 28% complete metric is a way for us to track issues and feature
requests for the release that we may want to implement.

Best

Hannes

hroest:
> So, what now ? I'll refrain from packaging any non-Qt5 version.

I suggest you wait until we have a release version from https://github.com/OpenMS/OpenMS/releases 



> It's incredible that with this low value, the whole project builds perfectly fine with the Qt 5 port code enabled.

[...]

PR #3831, opened 2018-10-30
[NOP] release cleanup

jpfeuffer:
see #3687


timosachsenberg:
+1

hroest:
great, but shouldnt we merge these into `master` ?

timosachsenberg:
I squashed the commit so we could apply that to master as well.

PR #3832, opened 2018-10-30
Update ConstRefVector.h

timosachsenberg:


timosachsenberg:
Is it defined twice then? If so the second one should be completely removed

src/openms/include/OpenMS/DATASTRUCTURES/ConstRefVector.h
```diff
@@ -214,7 +214,7 @@ namespace OpenMS
         return tmp;
       }
 
-      reference operator*()
+      reference operator*() const
```
hroest: why do we have 

```
pointer operator->();
pointer operator->() const;
```

but only

```
reference operator*() const;
```

I am not sure that this makes sense, actually it is unclear why we have the same function as const and non-const, there isnt really a great usecase for the non-const function, is there?

timosachsenberg: Doesn’t make sense... is it used anywhere?

timosachsenberg: I removed them... did still compile

PR #3834, opened 2018-11-01
try less compile threads

cbielow:
Travis always fails currently.. testing a bit how to rescue it

according to
https://docs.travis-ci.com/user/reference/overview/
we have 2 CPU cores and 4 GB ram.
i.e. using more than 2 threads might restrict RAM too much (leading to swapping?)


cbielow:
no dice. Runs as slowly as the other ones. Did not even finish building. No tests were run.
...
Options:
 - Optimize/split build -- seems almost a waste of time.
 - Ditch Travis?

hroest:
you can do this, then it will work again: https://github.com/OpenMS/OpenMS/wiki/Developer-FAQ#travis_to

hroest:
the problem is that we rely heavily on caching during the compile and if someone changes too many headers at once, the cache breaks and travis times out.

PR #3836, opened 2018-11-04
new makePeakPositionUnique() method for SpectrumHelper

lars20070:
Spectra and chromatograms might contain some peaks with identical position i.e. m/z or RT.

The new `makePeakPositionUnique()` method combines peaks with identical position to a single one. Its intensity is the median, mean, min or max of the individual peaks. The resulting new spectrum/chromatogram contains only unique positions.

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
+++ b/src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
@@ -32,6 +32,8 @@
 // $Authors: Timo Sachsenberg $
 // --------------------------------------------------------------------------
 
+#include <OpenMS/MATH/STATISTICS/StatisticFunctions.h>
+
 #pragma once
 
 namespace OpenMS
@@ -55,6 +57,7 @@ namespace OpenMS
     }
     return it;
   }
+  
   template <class DataArrayT>
   typename DataArrayT::const_iterator getDataArrayByName(const DataArrayT& a, const String& name)
   {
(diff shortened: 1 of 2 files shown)


timosachsenberg:
Hi Lars. Can you give an example when this happens? Some specific instrument type?

src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
```diff
@@ -133,6 +136,70 @@ namespace OpenMS
...
     // Note: data arrays are not updated
   }
+  
+  /**
+   @brief Make peak positions unique.
+    
+   A peak container may contain multiple peaks with the same position, i.e. either
+   an MSSpectrum containing peaks with the same m/z position, or an MSChromatogram
+   containing peaks with identical RT position. The method combines peaks with the
+   same position to a single one with the intensity dteremined by method m.
```
timosachsenberg: determined

src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
```diff
@@ -133,6 +136,70 @@ namespace OpenMS
...
+
+   @param[in] p The peak container to be manipulated.
+   @param[in] m The method for determining peak intensity from peaks with same position (1 = median, 2 = mean, 3 = min, 4 = max).
+  */
+  template <typename PeakContainerT>
+  void makePeakPositionUnique(PeakContainerT& p, const int m = 1)
```
timosachsenberg: please use an **enum class** instead of an int or a simple enum

cbielow:
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1MSSpectrum.html#a6f5c0fcfb9d669ba928a05355dae38f5 can solve problem 2.
It currently exists for MSSpectrum only. However, the code can easily be templated, such that it works for both.

However, before focussing on how to solve the problem, is it really one? What kind of vendor would write such data?

timosachsenberg:
I agree with @cbielow. The function seems very "special" and a more generic way would be to consider that problem as a border case of a spectrum merge algorithm (e.g., when a merge window is set to 0)

lars20070:
@timosachsenberg @cbielow I need this functionality for constructing better peptide intensities in `FeatureFinderMultiplex`. Instead of hiding it in my classes, I wanted to make it more widely available.

There are of course no vendor data with non-unique positions.

timosachsenberg:
I see. If you point us to the specific problem we might give you some hints how this could be done more efficiently in your code.

src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
```diff
@@ -133,6 +137,82 @@ namespace OpenMS
...
+  /**
+   * @brief Make peak positions unique.
+   * 
+   * A peak container may contain multiple peaks with the same position, i.e. either
+   * an MSSpectrum containing peaks with the same m/z position, or an MSChromatogram
+   * containing peaks with identical RT position. One scenario where this might happen
+   * is when multiple spectra are merged to a single one.
+   * 
+   * The method combines peaks with the same position to a single one with the
+   * intensity deteremined by method m.
```
timosachsenberg: determined

PR #3837, opened 2018-11-05
Update FineIsotopeDistribution_test.cpp

timosachsenberg:


timosachsenberg:
current builds are failing because of an overflow warning. this is fixed in #3780 . This PR provides a hot fix until the main PR is merged

michalsta:
Hi, I think you got the wrong commit. The one that fixes it is: 731f12490545d041694274bee7a3c1c8e7274803

This one is just some deobfuscation of the tests: both 1e-1000 and 1e-350 are smaller than the smallest positive double and get rounded down to zero at compile time, so this just changes the code to explicitly state what happens anyway.

On the other hand, if you just wish to disable the failing tests, then these are the ones involving a monoelemental molecule (C100).

timosachsenberg:
Ok that was just to get rid of the warnings.

PR #3840, opened 2018-11-07
change default params in IDMapper

### LINKED ISSUE #2468: review IDMapper
make advanced:
-ignore_charge

change default:
-mz_reference In my experience, mapping using the 'precursor' instead of the 'peptide' (theoretical mass) is inferior. I would vote for 'peptide' as  default,
- feature:use_centroid_mz should be default in my view as mapping to convex hulls leads to many false positives (also default for consensusXML?)

for reference:
```
Options (mandatory options marked with '*'):
  -id <file>*                  Protein/peptide identifications file (valid formats: 'mzid', 'idXML')
  -in <file>*                  Feature map/consensus map file (valid formats: 'featureXML', 'consensusXML', 'mzq')
  -out <file>*                 Output file (the format depends on the input file format). (valid formats: 'featureXML', 'consensusXML', 'mzq')

  -rt_tolerance <value>        RT tolerance (in seconds) for the matching of peptide identifications and (consensus) features. Tolerance is understood as 'plus or minus x', so the matching range increases by twice the given value. (default: '5' min: '0')
  -mz_tolerance <value>        M/z tolerance (in ppm or Da) for the matching of peptide identifications and (consensus) features. Tolerance is understood as 'plus or minus x', so the matching range increases by twice the given value. (default: '20' min: '0')
  -mz_measure <choice>         Unit of 'mz_tolerance'. (default: 'ppm' valid: 'ppm', 'Da')
  -mz_reference <choice>       Source of m/z values for peptide identifications. If 'precursor', the precursor-m/z from the idXML is used. If 'peptide', masses are computed from the sequences of peptide hits; in this case, an identification matches if any of its hits matches. ('peptide' should be used together with 'feature:use_centroid_mz' to avoid false-positive matches.) (default: 'precursor' valid: 'precursor', 'peptide')
  -ignore_charge               For feature/consensus maps: Assign an ID independently of whether its charge state matches that of the (consensus) feature.


Additional options for featureXML input:
  -feature:use_centroid_rt     Use the RT coordinates of the feature centroids for matching, instead of the RT ranges of the features/mass traces.
  -feature:use_centroid_mz     Use the m/z coordinates of the feature centroids for matching, instead of the m/z ranges of the features/mass traces. (If you choose 'peptide' as 'mz_reference', you should usually set this flag to avoid false-positive matches.)
```

lars20070:
This pull request changes the default parameters of the `IDMapper` tool as follows.
```
ignore_charge = false (advanced)
mz_reference = peptide
use_centroid_mz = true
```
Resolves #2468.


timosachsenberg:
does not resolve the isotopic misassignment (e.g., if the peptide m/z is feature m/z - 1.0033548/z )

lars20070:
IMO, it should. In your example, you do an MS-GF+ search with `isotope_error_range = -1,1`. The -1 error gets corrected and the correct sequence identified. Then you match the masses, not m/z.

hendrikweisser:
I think this is a useful improvement, for the reasons that @timosachsenberg pointed out in #2468. It doesn't solve the problem with mapping spuriously deamidated peptides (see https://github.com/OpenMS/OpenMS/issues/2468#issuecomment-302867094), but the old options that would allow that are still available.

One minor annoyance after this change is that "use_centroid_rt" is now a different type of parameter (flag) than "use_centroid_mz" (string). Maybe change "use_centroid_rt" to string as well for consistency?

timosachsenberg:
@lars20070 as Hendrik also pointed out, your search/correction only corrects for one type of error.
If the MS2 precursor is at the correct m/z but the feature is starting at the mass trace right to the monoisotopic, then it doesn't work.

lars20070:
@hendrikweisser Shall I change `use_centroid_rt` for consistency sake?

hendrikweisser:
@lars20070: Yes, please.

PR #3842, opened 2018-11-08
[FEATURE] MRMFeatureSelector and MRMFeatureScheduler

pasqoo:
`MRMFeatureSelector` provides a Base class (it contains a pure virtual function named `optimize()`) for selection of MRM Features based on:
- relative retention time using a quadratic mixed integer programming formulation
- a linear programming where each possible transition is weighted by a user defined score (most often retention time and peak intensity)

The two versions are implemented as derived classes `MRMFeatureSelectorQMIP` and `MRMFeatureSelectorScore`.

`MRMFeatureScheduler` offers an interface to `MRMFeatureSelector`. It takes a vector of parameters which will be used in multiple calls to the latter class' methods.

PS: Welcome Svetlana to the OpenMS project :smile: @sgalkina


timosachsenberg:
nice addition and welcome @sgalkina to OpenMS !

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -72,6 +72,8 @@ namespace OpenMS
     ~MRMFeatureFilter() override;
     //@}
 
+    void getDefaultParameters(Param& params) const;
```
timosachsenberg: add method documentation

timosachsenberg:
the diff shows some changes to the contrib. Does this point to an older version.?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureScheduler.h
```diff
@@ -0,0 +1,117 @@
...
+        nn_threshold(nn),
+        locality_weight(lw),
+        select_transition_group(stg),
+        segment_window_length(swl),
+        segment_step_length(ssl),
+        variable_type(vt),
+        optimal_threshold(ot),
+        score_weights(sw) {}
+
+      Int    nn_threshold            = 4;
```
timosachsenberg: please add comments to the members ///<

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureScheduler.h
```diff
@@ -0,0 +1,117 @@
...
+    /**
+      Calls `feature_selector.select_MRMFeature()` feeding it the parameters found in `parameters_`.
+      It calls said method `parameters_.size()` times, using the result of each cycle as input
+      for the next cycle.
+
+      @param[in] feature_selector Base class for the feature selector to use
+      @param[in] features Input features
+      @param[out] selected_features Selected features
+    */
+    void schedule_MRMFeatures(MRMFeatureSelector& feature_selector, const FeatureMap& features, FeatureMap& selected_features) const;
```
timosachsenberg: we usually don't use underscore in method names so consider removing them

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+#pragma once
+
+#include <OpenMS/config.h> // OPENMS_DLLAPI
+#include <OpenMS/DATASTRUCTURES/LPWrapper.h>
+#include <OpenMS/DATASTRUCTURES/String.h>
+#include <OpenMS/KERNEL/FeatureMap.h>
+
+namespace OpenMS
+{
+  class OPENMS_DLLAPI MRMFeatureSelector
```
timosachsenberg: please add class documentation

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+      @param[out] result Transitions' names filtered out of the LP problem
+    */
+    virtual void optimize(
+      const std::vector<std::pair<double, String>>& time_to_name,
+      const std::map<String, std::vector<Feature>>& feature_name_map,
+      std::vector<String>& result
+    ) = 0;
+
+    /**
+      Splits the features into time segments. `optimize()` method is run on each of these segments.
```
timosachsenberg: if possible some additional information should be added to the comment. e.g., are there any constraints on the segments? are these non-overlapping? etc.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+
+      @param[in] features Input features
+      @param[out] selected_filtered Output features
+    */
+    void select_MRMFeature(const FeatureMap& features, FeatureMap& selected_filtered);
+
+    void setNNThreshold(const Int nn_threshold);
```
timosachsenberg: please add documentation. 
Is this method generic enough to be put into this base class?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+    void setSelectTransitionGroup(const bool select_transition_group);
+    bool getSelectTransitionGroup() const;
+
+    void setSegmentWindowLength(const Int segment_window_length);
+    Int getSegmentWindowLength() const;
+
+    void setSegmentStepLength(const Int segment_step_length);
+    Int getSegmentStepLength() const;
+
+    void setVariableType(const String& variable_type);
```
timosachsenberg: consider using an enum class instead of string constants

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+    */
+    void select_MRMFeature(const FeatureMap& features, FeatureMap& selected_filtered);
+
+    void setNNThreshold(const Int nn_threshold);
+    Int getNNThreshold() const;
+
+    void setLocalityWeight(const bool locality_weight);
+    bool getLocalityWeight() const;
+
+    void setSelectTransitionGroup(const bool select_transition_group);
```
timosachsenberg: see above: doc and consider if generic enough for base class

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+    void setSegmentStepLength(const Int segment_step_length);
+    Int getSegmentStepLength() const;
+
+    void setVariableType(const String& variable_type);
+    String getVariableType() const;
+
+    void setOptimalThreshold(const double optimal_threshold);
+    double getOptimalThreshold() const;
+
+    void setScoreWeights(const std::map<String, String>& score_weights);
```
timosachsenberg: see above

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+      std::vector<Int> indices,
+      std::vector<double> values,
+      const String& name,
+      const double lb,
+      const double ub,
+      const LPWrapper::Type param
+    ) const;
+
+private:
+    Int    nn_threshold_            = 4;
```
timosachsenberg: these variables look familiar ;) how do they relate to SelectorParameters?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+      will be used to score candidate features based on their deviation from the
+      relative distance between the target transition's or transition group's times
+
+      @param[in] features Input features
+      @param[out] time_to_name Pairs representing a mapping of retention times to transition names
+      @param[out] feature_name_map Transitions' names to their features objects
+    */
+    void constructTargTransList_(
+      const FeatureMap& features,
+      std::vector<std::pair<double, String>>& time_to_name,
```
timosachsenberg: do these need to be sorted? I assume not...

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+      - "lambda score: 1/log10(score)"
+
+      @throw Exception::IllegalArgument When an invalid `lambda_score` is passed
+
+      @param[in] score Value to transform
+      @param[in] lambda_score A string representing the desired transformation
+
+      @return The weighted value
+    */
+    double weightScore_(const double score, const String& lambda_score) const;
```
timosachsenberg: consider using an enum class here

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
```diff
@@ -0,0 +1,265 @@
...
+
+  /**
+    Class used to select MRMFeatures based on a linear programming where each
+    possible transition is weighted by a user defined score (most often retention
+    time and peak intensity). The method is described in [TODO: update when published].
+  */
+  class OPENMS_DLLAPI MRMFeatureSelectorScore : public MRMFeatureSelector
+  {
+public:
+    void optimize(
```
timosachsenberg: please add parameter documentation to optimize methods

src/openms/include/OpenMS/DATASTRUCTURES/LPWrapper.h
```diff
@@ -100,7 +100,7 @@ namespace OpenMS
 
     enum Type
     {
-      UNBOUNDED = 1,
+      UNBOUNDED = 1,
```
timosachsenberg: the most import thing: remove that extra space ;)

timosachsenberg:
Look already very good and I only found minor issues.
In general, more documentation of parameters and potential constraints could be added.
Also, tests for the two different methods could be more extensive on the class level and when run on real data. Also, I think only one class can be accessed using pyOpenMS, right= 
From a design point of view, my only concerns are that the base class might contain too much or too specialized methods. But I would like to hear somebody with experience with MRM scheduling.

src/tests/class_tests/openms/source/MRMFeatureSelector_test.cpp
```diff
@@ -0,0 +1,356 @@
...
+  score = selector.computeScore_(feature);
+  TEST_REAL_SIMILAR(score, 0.20059549093267626)
+
+  selector.setScoreWeights({{"sn_ratio", "lambda score: 1/log(score)"}, {"peak_apices_sum", "lambda score: 1/log10(score)"}});
+  score = selector.computeScore_(feature);
+  TEST_REAL_SIMILAR(score, 0.10427624775717449)
+}
+END_SECTION
+
+// START_SECTION(schedule_MRMFeaturesQMIP() integer) // integer variable type
```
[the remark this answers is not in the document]
pasqoo: It is commented out because the test takes several minutes to complete

pasqoo:
Thank you @timosachsenberg for the review. I'll take care of all the points.
I'll check the `contrib` issue, but those changes can be safely removed from the PR.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
pasqoo: I refactored the parameters. I moved `SelectorParameters` to `MRMFeatureSelector`. The class takes a single objects, and then its members are accessed by the derived classes.

I had to do this, otherwise the scheduler would not find the (moved) setters in the base class.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
pasqoo: The `optimize` method makes use of two indices when accessing the elements of `time_to_name`. That is why I kept it as a `vector`.

pasqoo:
@timosachsenberg The PR is up to date, ready for another check :smiley:

hroest:
@pasqoo @timosachsenberg is this based on the following paper: https://www.ncbi.nlm.nih.gov/pubmed/20201589 ? Was this code by A Bertsch ever integrated into OpenMS ? Is it this class http://ftp.mi.fu-berlin.de/pub/OpenMS/release2.2.0-documentation/html/classOpenMS_1_1PrecursorIonSelection.html ? 

If so, how does the current code integrate with the PrecursorIonSelection?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h
hroest: can you please explain in the documentation what the class does, e.g. what is "optimal" in this context? What is the class selecting based on which criteria?

timosachsenberg:
I think the code never made it into the SVN but I could be wrong. That was before I joined the group.

dmccloskey:
The algorithm that we are proposing is quite a bit different from my understanding.  

Also, the use case for the algorithm is feature selection after peak integration as opposed to optimal choice or scheduling of MRMs.

hroest:
> Also, the use case for the algorithm is feature selection after peak integration as opposed to optimal choice or scheduling of MRMs.

ah ok, then I misunderstood this. Sorry.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureScheduler.h
```diff
@@ -0,0 +1,81 @@
...
+    void scheduleMRMFeaturesQMIP(const FeatureMap& features, FeatureMap& selected_features) const;
+
+    /// Setter for the scheduler's parameters
+    void setSchedulerParameters(const std::vector<MRMFeatureSelector::SelectorParameters>& parameters);
+
+    /// Getter for the scheduler's parameters
+    std::vector<MRMFeatureSelector::SelectorParameters>& getSchedulerParameters(void);
+
+private:
+    /// Parameters for a single call to the scheduler. All elements will be consumed.
```
timosachsenberg: what does "consumed" mean in this context?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureScheduler.h
```diff
@@ -0,0 +1,81 @@
...
+#include <OpenMS/config.h> // OPENMS_DLLAPI
+#include <OpenMS/ANALYSIS/OPENSWATH/MRMFeatureSelector.h>
+#include <OpenMS/KERNEL/FeatureMap.h>
+#include <map>
+#include <vector>
+
+namespace OpenMS
+{
+  /**
+    Class used to schedule multiple calls to `MRMFeatureSelector`
```
timosachsenberg: can you add more documentation here? Just looking at this single line would not tell me much what the class is doing.

pasqoo:
@timosachsenberg updated

pasqoo:
@timosachsenberg PR updated. This should cover all the changes discussed on Gitter:
- rename `MRMFeatureScheduler` to `MRMBatchFeatureSelector`
- refactor `MRMFeatureSelector` to manage parameters through methods' arguments instead of class' members
- make `MRMBatchFeatureSelector`'s methods static

hroest:
can you look at this test failure: http://cdash.openms.de/testDetails.php?test=53359842&build=292805

```

checking batchMRMFeaturesQMIP() continuous ... 
 -  line 276:  TEST_EQUAL(output_selected.size(),82): got 84, expected 82

```

pasqoo:
Thanks for the links.
Those test failures seem tricky. The compiled code gives different results according to the platform.
On my machine I get 82, and those two tests get 84 and 81 elements in output.

Can these differences be caused by the linear programming algorithms? Or do you suspect something different?

timosachsenberg:
@cbielow we are currently struggling a bit with this test. I thought you used the LPWrapper quite a bit in the past. Maybe you have any insights into how we could solve that problem.

cbielow:
mhh.. not sure why this is. Might even be the input (from what I understand). Maybe getting rid of inverse_log and replacing it with linear weights might do the trick (at least for the sake of the test)?

pasqoo:
@cbielow Thank you so much!

What do you think are the reasons for that?

@hroest @timosachsenberg It looks like tests now pass

timosachsenberg:
MRMFeatureSelector_150601_0_BloodProject01_PLT_QC_Broth-1_1.featureXML is rather large test file. Can you slim that down?

timosachsenberg:
One way to preserve the full test is to remove lines not required in this featureXML and whitelist them in the test

PR #3843, opened 2018-11-12
pin cython version to 0.25

timosachsenberg:


timosachsenberg:
@hroest pyOpenMS builds again. linux timed out

mwalzer:
Did you have a chance to check: Does this also solve the issue with building pyOpenMS in docker?

timosachsenberg:
it now builds here but I did not check if it works in the docker container. Will be back at the harbour the next days

PR #3846, opened 2018-11-14
delete valid strings for comet version

Leon-Bichmann:
I realized that the comet version shouldnt be constrained with valid strings as otherwise all other comet versions will fail with this adapter.


timosachsenberg:
I think we should still document somewhere which versions are known to work

PR #3848, opened 2018-11-14
test xenial base image

timosachsenberg:
just a quick check if xenial works out of the box


timosachsenberg:
GCC 7 seems to compile faster for OpenMP code so  this might also be worth a try

jpfeuffer:
and I think on xenial there is a new enough Qt5, too.

hroest:
looks like it almost worked:

```
The following tests FAILED:
	608 - TOPPView_test (Failed)
	2103 - TOPP_MyriMatchAdapter_1 (Failed)
	2104 - TOPP_MyriMatchAdapter_1_out (Failed)
	2187 - UTILS_INIUpdater_1 (Failed)
	2188 - UTILS_INIUpdater_1_out (Failed)
	2190 - UTILS_INIUpdater_3 (Failed)
	2191 - UTILS_INIUpdater_3_out (Failed)
	2259 - TOPP_ExecutePipeline_1 (Failed)
```

jpfeuffer:
Iniupdater, toppview and execute pipeline require an xserver as far as I know.

jpfeuffer:
I think we can also get rid of this line:
https://github.com/OpenMS/OpenMS/blob/develop/tools/travis/lnx-cibuild.sh#L53

see https://benlimmer.com/2019/01/14/travis-ci-xvfb/

and we probably need to add the "--user" flag to the pip install commands
https://github.com/OpenMS/OpenMS/blob/develop/tools/travis/lnx-cibuild.before.sh#L31

and we should probably switch to python 3
https://github.com/OpenMS/OpenMS/blob/develop/tools/travis/lnx-cibuild.before.sh#L22

but let's check step by step

tools/travis/cibuild.cmake
```diff
@@ -27,13 +27,13 @@ PY_NUM_MODULES=4
...
 PY_SINGLE_THREADED=ON
-PY_NO_OUTPUT=On
+PY_NO_OUTPUT=Off
 ENABLE_STYLE_TESTING=$ENV{ENABLE_STYLE_TESTING}
 ENABLE_TOPP_TESTING=$ENV{ENABLE_TOPP_TESTING}
 ENABLE_CLASS_TESTING=$ENV{ENABLE_CLASS_TESTING}
 WITH_GUI=$ENV{WITH_GUI}
 ADDRESS_SANITIZER=$ENV{ADDRESS_SANITIZER}
-WITH_THERMORAWFILEPARSER_TEST=Off"
+WITH_THERMORAWFILEPARSER_TEST=On"
```
hroest: this suppress Python output on travis ...

jpfeuffer:
I assume travis will deprecate trusty soon. So it is worth finishing this one.

jpfeuffer:
ok. impossible. with logs travis exits because of log length, without it thinks the build is stalled. this would need a major restructuring. we should just ditch travis.

PR #3851, opened 2018-11-15
[FIX] replace hostname in mascot query

### LINKED ISSUE #3611: KNIME and MascotAdapterOnline fails for Mascot 2.6
Hej OpenMS Team,

We just updatet our Mascot server to version 2.6.2 (from 2.5) and are running into problems with the stable KNIME nodes (2.3.0.201712221504):
The Mascot node is starting the search and it is finishing on the server, but the transfer back to OpenMS does not work. Errors on the nodes are:

Standard output:
> [...]
> Submitting Mascot query (now: 2018-07-10 10:54:02)...
> Mascot query finished
> An error occurred during the query: MascotRemoteQuery: The server returned an error status code '404': Not Found
> Try accessing the server
>   mascot/mascot
>  from your browser and check if it works fine.
> MascotAdapterOnline took 7.14 s (wall), 1.65 s (CPU), 0.02 s (system), 1.63 s (user).
> Peak Memory Usage: <unknown>

Error output:
> MascotRemoteQuery: An error occurred (requestId=8): Request aborted (QT Error Code: 7)
> MascotRemoteQuery: An error occurred (requestId=10): Request aborted (QT Error Code: 7)

Obviously, accessing the server via the browser does work at the indicated addresses.

I may have pinned the problem down to a change in the export behavior of Mascot, as the call of
> GET /mascot/cgi/export_dat_2.pl?file=..%2F ....
Gets redirected to a caching-site:
> GET /mascot/cgi/nph-cache_families.pl?file=..%2Fdata%
and the node seems to not expect this.

Is this already fixed in the nightlies and maybe coming to the stable branch soon? Or is there another way to get the KNIME nodes working with Mascot 2.5 again?
For some reasons I don't like to use the nightlies branch on the stable workflows ;)

jpfeuffer:
Was replacing all occurrences of the hostname. Wrong in case of equal servername.
This is on top of my head. Changes untested.
fixes #3611 if successful

### DIFF
--- a/src/openms/source/FORMAT/MascotRemoteQuery.cpp
+++ b/src/openms/source/FORMAT/MascotRemoteQuery.cpp
@@ -742,7 +742,11 @@ namespace OpenMS
       endRun_();
       return;
     }
-    url.remove(host_name_.toQString());
+    
+    // makes sure that only the first occurrence in the String is replaced,
+    // in case of an equal server_name_
+    url.replace(url.indexOf(host_name_.toQString()),
+                          host_name_.toQString().size(), QString(""));
 
     // ensure path starts with /
     if (url[0] != '/') url.prepend('/');


jpfeuffer:
rebuild jenkins

PR #3852, opened 2018-11-15
Feature/rework decharger output

FabianAicheler:
Improved output. Fixed removal of new userParams from FFM not wanted in Decharger output. Changed decharged mz of ungrouped monoisotopic features in Decharger consensusXML to the value of the input feature (in addition to flag marking such ungrouped monoisotopic), in accordance with user wishes.

### DIFF
--- a/src/tests/class_tests/openms/data/MetaboliteFeatureDeconvolution_neg_output.consensusXML
+++ b/src/tests/class_tests/openms/data/MetaboliteFeatureDeconvolution_neg_output.consensusXML
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
 <?xml-stylesheet type="text/xsl" href="file:////home/fa/OpenMS/share/OpenMS/XSL/ConsensusXML.xsl"?>
-<consensusXML version="1.7" id="cm_16772366217575462398" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/OpenMS/OpenMS/develop/share/OpenMS/SCHEMAS/ConsensusXML_1_7.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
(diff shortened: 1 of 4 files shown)


FabianAicheler:
@oliveralka Those output changes (see updated test files) should not affect SIRIUS, right?

oliveralka:
@FabianAicheler SiriusAdapter right now works on the featureXML output - if you did not change anything there this should not matter.

src/tests/class_tests/openms/data/MetaboliteFeatureDeconvolution_neg_output.consensusXML
```diff
@@ -22,145 +22,120 @@
 			<UserParam type="intList" name="distinct_charges" value="[-3, -2, -1]"/>
 			<UserParam type="int" name="distinct_charges_size" value="3"/>
 			<UserParam type="int" name="pure_proton_features" value="3"/>
-		</consensusElement>
-		<consensusElement id="e_6880193271103366492" quality="0">
-			<centroid rt="100" mz="0" it="100"/>
+			<UserParam type="string" name="102" value="H-3O-1"/>
```
oliveralka: If you would like to extract the information again it may be better to use a StringList with a set e.g. 
103:H-2Na1 - or something in this direction. What do you think?

oliveralka:
Look good, but I think it would be nicer to have more compressed of storing the adducts of the specific features (e.g StringList with a set 20:M+H)
<UserParam type="StringList" name="CombindedAdducts" value={"103:H-2Na1"}/>

What is your opinion on that?

src/tests/class_tests/openms/data/MetaboliteFeatureDeconvolution_neg_output.consensusXML
FabianAicheler: Mhh, possibly, gotta think about that if improvement for future, currently I just aim to make it the same as after FeatureLinking features with adducts.

FabianAicheler: Currently I think this would provide no gains beside aesthetics: use cases require to read the elems (which contain further information), which provide the IDs to query the corresponding adduct  userParams.

oliveralka: Ok - then leave it as it is for now.

PR #3862, opened 2018-11-22
fix #3424 in TOPPViewBase.cpp

### LINKED ISSUE #3424: FeatureXML and idXML don’t open as new layer anymore in TOPPView
Only new window

timosachsenberg:
Fixes #3424 
Somehow, opening the file dialog invalidates the current active sub window in the enhanced tab bar.
This fix saves and restores it.


timosachsenberg:
maybe someone can confirm that this fixed the issue for him

PR #3864, opened 2018-11-23
[FIX] OpenMS#3860

bretttully:


src/openswathalgo/source/OPENSWATHALGO/ALGO/Scoring.cpp
```diff
@@ -105,7 +105,14 @@ namespace OpenSwath
       x_len = std::sqrt(x_len);
       y_len = std::sqrt(y_len);
 
-      return std::acos(dotprod / (x_len * y_len));
+      // normalise, avoiding a divide by zero. See unit tests for what happens
+      // when one of the vectors has a length of zero.
+      dotprod /= std::max(std::numeric_limits<double>::min(), x_len * y_len);
```
timosachsenberg: imho this changes the sign of the dotproduct and sets it to -1 or +1.
wouldn't it be easier to just do a simple if (y_len == 0) return ... etc.

hroest: I agree, one could do something like this, right?

```suggestion
      if (x_len * y_len == 0) {return pi}; 
      return std::acos(dotprod / (x_len * y_len));
```

bretttully: I agree on the `std::max` call here, it definitely doesn't take into account sign and it should.

```suggestion
      double denominator = x_len * y_len;
      dotprod = (denominator == 0.0) ? 0.0 : dotprod / denominator;
```

But you still need the `clip` below to account for floating point errors that can cause `acos` to blow up

src/tests/class_tests/openswathalgo/Scoring_test.cpp
```diff
@@ -106,25 +106,122 @@ END_SECTION
...
+              0.0
+              >>> spectral_angle((1, 0, 0), (-1, 0, 0))
+              3.141592653589793
+              >>> spectral_angle((0, 0, 0), (0, 0, 0))  # error or pi/2?
+              1.5707963267948966
+      """
+      numer = np.dot(v1, v2)
+      v1_u = np.sqrt(np.dot(v1, v1))
+      v2_u = np.sqrt(np.dot(v2, v2))
+      denom = max(1e-15, v1_u * v2_u)
```
bretttully: Once the above is decided, this should be changed too

timosachsenberg:
sounds good to me. Agree that clipping seems still to be needed.

bretttully:
PR updated

timosachsenberg:
rebuild jenkins

bretttully:
> rebuild jenkins

Was that for me @timosachsenberg?

timosachsenberg:
hehe no, sorry for the confusion  - just keywords to trigger a (re)build on our test server

bretttully:
Thanks for the quick turn-around guys! 

Would you normally tag this as Release2.4.1? Or should we just reference the commit hash in our workflows?

PR #3865, opened 2018-11-26
IDFileConverter - NativeID prior to RT lookup

Leon-Bichmann:
Conversion of idXML files to pepXML files results in reannotation of scan ids.

When providing the original mz_file scan index lookup is carried out using an RT lookup.

Since the RT can change upon transformation a scan_index based lookup should be preferential.

### DIFF
--- a/src/openms/source/FORMAT/PepXMLFile.cpp
+++ b/src/openms/source/FORMAT/PepXMLFile.cpp
@@ -279,7 +279,11 @@ namespace OpenMS
 
         if (lookup.empty())
         {
-          if (it->metaValueExists("RT_index")) // Setting metaValue "RT_index" in XTandemXMLFile in the case of X! Tandem.
+          if (it->metaValueExists("spectrum_reference"))
+          {
+            scan_index = it->getMetaValue("spectrum_reference");
+          }
+          else if (it->metaValueExists("RT_index")) // Setting metaValue "RT_index" in XTandemXMLFile in the case of X! Tandem.
           {
             scan_index = it->getMetaValue("RT_index");
           }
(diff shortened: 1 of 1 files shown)


src/openms/source/FORMAT/PepXMLFile.cpp
```diff
@@ -279,15 +279,26 @@ namespace OpenMS
 
         if (lookup.empty())
         {
-          if (it->metaValueExists("RT_index")) // Setting metaValue "RT_index" in XTandemXMLFile in the case of X! Tandem.
+          if (it->metaValueExists("spectrum_reference"))
+          {
+            scan_index = it->getMetaValue("spectrum_reference");
```
timosachsenberg: I think spectrum reference is a native id in most cases (string) and not just an integer.
See the idXML files in https://github.com/OpenMS/OpenMS/search?q=spectrum_reference&unscoped_q=spectrum_reference
So this could be a bit dangerous and will probably not work in most cases.

src/openms/source/FORMAT/PepXMLFile.cpp
```diff
@@ -279,15 +279,26 @@ namespace OpenMS
...
           {
             scan_index = it->getMetaValue("RT_index");
           }
           scan_nr = scan_index + 1;
         }
         else
         {
+          if (it->metaValueExists("spectrum_reference"))
+          {
+            scan_index = lookup.findByNativeID(it->getMetaValue("spectrum_reference"));
```
timosachsenberg: looks good

timosachsenberg:
@oliveralka do we have some code to automatically detect the native id format and extract the scan number? I think I recall that you implemented something for SIRIUSAdapter

PR #3870, opened 2018-11-27
[FIX] pyOpenMS fix for Win/OSX

hroest:
- unsigned long is apparently not the same as uint64_t on all platforms
  and we need to implement a different solution that is also portable


hroest:
duplicate of https://github.com/OpenMS/OpenMS/pull/3868

PR #3877, opened 2018-11-30
[REFACTOR] SiriusAdapter

oliveralka:
[ADD] class SiriusAdapterAlgorithm 
[ADD] class FragmentAnnotation -edit: renamed to SiriusFragmentAnnotation
[ADD] CompoundInfo and Comments to SiriusMSConverter 
[ADD] Corresponding tests and python bindings

SiriusAdapterAlgorithm: Functionalize SiriusAdapter and add param handling. For easier integration in other tools (such as AssayGeneratorMetabo).

FragmentAnnotation: Extract Fragment Annotation (mass, int, exact_mass and explanation)  from SIRIUS output (consensus spectrum) and save it in MSSpecturm with annotated NativeID. 

CompoundInfo will later be used in the AssayGeneratorMetabo (Mapping) 
and the Comments will be needed for SIRIUS 4.0.4 and its mztab-m integration. 

Parameter Names were changed for the SiriusAdapter which will break the current workflows after the update. This was done to easier distinguish the corresponding parameter for OpenMS preprocessing and SIRIUS. 
e.g. 
feature_only -> preprocessing:feature_only 
profile -> sirius:profile


src/openms/include/OpenMS/FORMAT/DATAACCESS/FragmentAnnotation.h
```diff
@@ -0,0 +1,89 @@
...
+// $Authors: Oliver Alka $
+// --------------------------------------------------------------------------
+
+#pragma once
+
+#include <OpenMS/KERNEL/MSSpectrum.h>
+
+namespace OpenMS
+{
+  class OPENMS_DLLAPI FragmentAnnotation
```
timosachsenberg: hmm we already have a peak annotation  class in OpenMS and this one seems to be a bit specific for SIRIUS.
Also its not clear to me what and how it is annotated. In the DataArrays?

timosachsenberg: maybe put that into SIRIUS namespace ?

oliveralka: or call it SiriusFragmentAnnotation? You are right it is more or less SIRIUS specific. 

Ok, I add a more detailed explanation in the comments. 
It consists of a Peak1D with the mass and int. Additional it has a FloatDataArray for the exact mass and a StringDataArry where the "explanation" is stored. This will be appended to the MSSpectrum. 
It is possible to switch the mass and the exact mass since for assay libraries you would rather use the exact one.

oliveralka: renamed class to SiriusFragmentAnnotation and added further details in the description.

oliveralka:
@timosachsenberg Do you have any other questions or suggestions? If that is not the case - It should be ready for merging. Thank you!

src/openms/include/OpenMS/ANALYSIS/ID/SiriusMSConverter.h
```diff
@@ -45,21 +45,56 @@ namespace OpenMS
   {
 public:
 
+  // struct to store information about accsessions
```
hroest: typo accsessions ?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -0,0 +1,361 @@
...
+          {
+            num_masstrace_filter = 1;
+            LOG_WARN << "Parameter: filter_by_num_masstraces, was set to 1 to retain the adduct information for all MS2 spectra, if available. Please use the masstrace filter in combination with feature_only." << std::endl;
+          }
+
+          // filter feature by number of masstraces
+          auto map_it = remove_if(feature_map.begin(), feature_map.end(),
+                                  [&num_masstrace_filter](const Feature &feat) -> bool
+                                  {
+                                    unsigned int n_masstraces = feat.getMetaValue("num_of_masstraces");
```
hroest: this will be very slow, better store `feat.getMetaValue("num_of_masstraces");` in a cache variable

src/openms/source/ANALYSIS/MAPMATCHING/FeatureMapping.cpp
```diff
@@ -40,13 +40,13 @@ using namespace std;
 namespace OpenMS
 {
   // return map of ms2 to feature and a vector of unassigned ms2
-  FeatureMapping::FeatureToMs2Indices FeatureMapping::assignMS2IndexToFeature(OpenMS::PeakMap& spectra,
+  FeatureMapping::FeatureToMs2Indices FeatureMapping::assignMS2IndexToFeature(OpenMS::MSExperiment& spectra,
```
hroest: spectra could be `const` here

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -0,0 +1,175 @@
...
+      @param featureinfo: Path to featureXML
+      @param spectra: Input of MSExperiment with spectra information
+      @param fp_map_kd: KDTree used for query and match spectra with features
+      @param sirius_algo: Parameters needed for preprocessing
+      @param feature_mapping: Empty FeatureToMs2Indices
+      */
+      static void preprocessingSirius(const String& featureinfo,
+                                      MSExperiment& spectra,
+                                      std::vector<FeatureMap>& v_fp,
+                                      KDTreeFeatureMaps& fp_map_kd,
```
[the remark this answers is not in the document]
oliveralka: The spectra could be const, but the v_fp and fp_map_kd not since they are filled in the function and in the end the pointer from the featrure map are referenced back to the vector of FeatureMap v_fp.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -0,0 +1,175 @@
...
+      @param tmp_ms_file: path to temporary .ms file
+      @param tmp_out_dir: path to temporary output folder
+      @param executable: path to executable
+      @param out_csifingerid: path to CSI:FingerID output (can be empty).
+      @param sirius_algo: Parameters needed for SIRIUS
+
+      */
+      static std::vector<String> callSiriusQProcess(const String& tmp_ms_file,
+                                                    const String& tmp_out_dir,
+                                                    String& executable,
```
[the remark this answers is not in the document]
oliveralka: yes, that should be possible.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -0,0 +1,175 @@
...
+                                      MSExperiment& spectra,
+                                      std::vector<FeatureMap>& v_fp,
+                                      KDTreeFeatureMaps& fp_map_kd,
+                                      const SiriusAdapterAlgorithm& sirius_algo,
+                                      FeatureMapping::FeatureToMs2Indices& feature_mapping);
+
+      /**
+      @brief logs numnber of features and spectra used
+
+      @return numnber of features and spectra used (LOG)
```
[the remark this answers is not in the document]
oliveralka: yes - will print basically the LOG_INFO in the output.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusMSConverter.h
oliveralka: yes

src/openms/source/ANALYSIS/MAPMATCHING/FeatureMapping.cpp
oliveralka: yes you are right.

oliveralka:
@hroest thank you - will make additional changes.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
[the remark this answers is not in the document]
hroest: maybe make this more clear in the `@param` so that its clear what is input and what is output

PR #3879, opened 2018-12-03
fix: LPWrapper destructor frees ctor's allocated memory

pasqoo:
While testing my code with Valgrind I noticed a large amount of memory not freed. The code (_MRMFeatureSelector_) uses _LPWrapper_. I looked into it and it seems that the reason might lie within the destructor.

Hopefully the proposed changes solve the issue. I did not have a chance to do extensive testing on my machine today, therefore I am relying on the CI system you have put in place.


cbielow:
good catch. looks good on first sight.

PR #3883, opened 2018-12-04
[FIX] SiriusAdapter: negative mode feature_only

oliveralka:
Introduces further changes for [REFACTOR] #3877.

[FIX] Will assume feature charge +1 or -1 for features with one masstrace based on acquisition mode. 
Additional feature charge will be set to +1 or -1 instead of 1 depending on the mode. This is used in the .ms file and later in Sirius - to use the appropriate adducts.

[TEST] Added tests for negative mode


oliveralka:
rebuild jenkins

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -188,6 +188,26 @@ namespace OpenMS
           count_to_neg = count_to_neg + 1;
         }
 
+        // set feature_charge value for msfile
+        if (p == IonSource::Polarity::POSITIVE && feature_charge == 1)
```
timosachsenberg: this is not needed

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -188,6 +188,26 @@ namespace OpenMS
...
+          feature_charge = +1;
+        }
+        if (p == IonSource::Polarity::NEGATIVE && feature_charge == 1)
+        {
+          feature_charge = -1;
+        }
+        if (p == IonSource::Polarity::POSITIVE && feature_charge == 0)
+        {
+          feature_charge = +1;
+          count_to_pos = count_to_pos + 1;
```
timosachsenberg: ++count_to_pos ?

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -188,6 +188,26 @@ namespace OpenMS
           count_to_neg = count_to_neg + 1;
         }
 
+        // set feature_charge value for msfile
+        if (p == IonSource::Polarity::POSITIVE && feature_charge == 1)
+        {
+          feature_charge = +1;
+        }
+        if (p == IonSource::Polarity::NEGATIVE && feature_charge == 1)
```
timosachsenberg: the whole part could be written much shorter e.g.:
if (feature_charge == 0) { feature_charge = 1; }
if (p == IonSource::Polarity::NEGATIVE) { -abs(feature_charge) }

timosachsenberg: and would already handle higher charge states

oliveralka: alright will change that - thank you!

hroest: > and would already handle higher charge states

yes, that would be good to have -- do not assume all charges are +1 or -1 ...

oliveralka: That is correct. 
In general: SIRIUS is only able to use charge +1/-1  data

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -165,48 +165,20 @@ namespace OpenMS
...
-        }
-        if (p == IonSource::Polarity::NEGATIVE && feature_charge == 0)
-        {
-          feature_charge = -1;
-          count_to_neg = count_to_neg + 1;
-        }
+        if (p == IonSource::Polarity::POSITIVE && feature_charge == 1) { feature_charge = +1; }
+        if (p == IonSource::Polarity::NEGATIVE && feature_charge == 1) { feature_charge = -1; }
+        if (feature_charge == 0) { feature_charge = +1; ++count_to_pos; }
+        if (p == IonSource::Polarity::NEGATIVE){ -abs(feature_charge); ++count_to_neg; }
```
hroest: why not use `else if` ?

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -165,48 +165,20 @@ namespace OpenMS
...
-        {
-          feature_charge = +1;
-          count_to_pos = count_to_pos + 1;
-        }
-        if (p == IonSource::Polarity::NEGATIVE && feature_charge == 0)
-        {
-          feature_charge = -1;
-          count_to_neg = count_to_neg + 1;
-        }
+        if (p == IonSource::Polarity::POSITIVE && feature_charge == 1) { feature_charge = +1; }
```
hroest: this looks like a NOP, when its 1, then set it to 1 ?

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
oliveralka: thanks!

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
oliveralka: see below.

oliveralka:
rebuild jenkins

timosachsenberg:
http://cdash.openms.de/testDetails.php?test=53619633&build=293977

hroest:
can you undo 477a9581f5972771e6ea1d4dbee347c1d2d3e1ad ?

Either it should be

```
if (x) {foo();}
```

or 

```
if (x)
{
  foo();
}
```

and not something between.

oliveralka:
If I try to do is as follows:
```
        else if (p == IonSource::Polarity::NEGATIVE && feature_charge == 1) {feature_charge = -1;}
        else if (feature_charge == 0) {feature_charge = +1; ++count_to_pos;}
        else if (p == IonSource::Polarity::NEGATIVE) {-abs(feature_charge); ++count_to_neg;}
```
cpplint throws an error:
```
If an else has a brace on one side, it should have it on both  [readability/braces] [5]
```
(That explanes the current style - sorry)
It seems that is a problem with the one liner.

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -169,16 +169,42 @@ namespace OpenMS
...
+        {
+          int_charge = -1;
+        }
+        else if (precursor_charge == 0)
+        {
+          int_charge = +1; ++count_to_pos;
+        }
+        else if (p == IonSource::Polarity::NEGATIVE)
+        {
+          -abs(int_charge); ++count_to_neg;
```
timosachsenberg: please don't put multiple expressions in the same line

timosachsenberg:
why not the short version I proposed?

´´´
if (feature_charge == 0) { feature_charge = 1; }
if (p == IonSource::Polarity::NEGATIVE) { -abs(feature_charge) }
´´´

oliveralka:
Think I got it now ...

oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

hroest:
looks good but the SiriusAdapter seem to fail quite randomly on our test infrastructure...

oliveralka:
Yes, I am not sure about that - For Windows It may be a memory issue. 

In linux I really don't know what is happening with the test - all tests pass on MacOS 14.10 and scientific linux 7.

timosachsenberg:
yes, the machine is unfortunately too weak / has not enough ram for the the many VMs it is running

oliveralka:
As discussed with @hroest.
Added a switch for SIRIUS (SiriusAdapter) tests (default: ON).
Added dependencies for one test at a time (hopefully reduces memory consumption)

hroest:
we discussed this and one solution is to add CMake dependencies so that only one test can run at a time...lets see whether that helps

timosachsenberg:
sounds like a good solution. let's see if that works.

oliveralka:
I am not able to reproduce the error in Ubuntu 18.04. But it seems to be consistent over several build attempts. 

SiriusAdapter_6 is a converter test - so it tests the .ms file output for the negative mode. This means that SIRIUS is not called in this test. It seems that the "checkFeatureSpectraNumber" is called correctly (see output), but then either the store or copy process goes wrong. 

This should be shown:
```
Number of features to be processed: 1
Number of additional MS2 spectra to be processed: 45
No MS1 spectrum for this precursor. Occurred 1 times.
0 spectra were skipped due to precursor charge below -1 and above +1.
Mono charge assumed and set to charge 1 with respect to current polarity 1 times.
0 features were skipped due to feature charge below -1 and above +1.
SiriusAdapter was used in converter mode and is terminated after openms preprocessing.
If you would like to run SIRIUS internally please disable the converter mode.
``` 

This is shown right now:
```
Number of features to be processed: 1
Number of additional MS2 spectra to be processed: 45
```

I is not reproducible on scientific linux 7. I will try to reproduce the problem in a VM (on the weekend).

hroest:
so what happens between 


Number of additional MS2 spectra to be processed: 45

and 

No MS1 spectrum for this precursor. Occurred 1 times.

?

timosachsenberg:
rebuild jenkins

timosachsenberg:
timeout on travis.. windows unreleated

PR #3884, opened 2018-12-05
Feature/osw debug

hroest:
some code cleanup, fix of bug #3764, use `auto` to make code more readable

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
@@ -269,9 +269,9 @@ namespace OpenMS
         std::cout << "line " << i << " " << tmp_line[i] << std::endl;
       }
 
-      for (std::map<std::string, int>::iterator iter = header_dict.begin(); iter != header_dict.end(); ++iter)
+      for (const auto& iter : header_dict)
       {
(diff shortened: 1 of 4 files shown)


src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
```diff
@@ -1038,22 +1038,28 @@ namespace OpenMS
 
     // Try to parse full UniMod string including modifications. If we fail, we
     // can force reading and only parse the "naked" sequence.
+    // Note: If the user did not provide a modified sequence string, we will
+    // fall back to the "naked" sequence by default.
     std::vector<TargetedExperiment::Peptide::Modification> mods;
     AASequence aa_sequence;
+    String sequence = tr_it->FullPeptideName;
+    if (sequence.empty()) sequence = tr_it->PeptideSequence;
```
hroest: missing UniMod peptide sequence will now mean that the naked sequence is used instead by default and no `force_invalid_mods` is required

timosachsenberg:
cool. Would this also fix https://github.com/OpenMS/OpenMS/pull/3791 ?

src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
grosenberger: 👍

PR #3886, opened 2018-12-07
[Feature] OpenPepXL speedup by simplification

enetz:
I started with a data set, that took about one hour to analyze on two CPU cores on my laptop and tried to make the tool as fast as possible without changing the output in any way.
Most of the changes concern the generation of theoretical spectra, but I timed the entire tool. 

Starting conditions:
`OpenPepXLLF took 53:40 m (wall), 01:47:03 h (CPU), 2.48 s (system), 01:47:01 h (user).`

I previously calculated many metrics for every candidate. I changed the main loop to only compute the stuff really necessary for the main score that is used for ranking and filtering the hits and removed String annotations from the theoretical spectra. For this I added a new parameter for adding charges to `TheoreticalSpectrumGeneratorXLMS`, because the charges were included in the parameter for metainfo annotations. The old full workflow is now only applied to the top X hits for each spectrum (X being an input parameter of the tool).
New time:
`OpenPepXLLF took 26:50 m (wall), 53:27 m (CPU), 1.72 s (system), 53:25 m (user).`

I implemented a simpler TSG called SimpleTSGXLMS (Simple TheoreticalSpectrumGenerator for CrossLinking MS). For a start I copied the old TSGXLMS and removed any operations and parameters concerning String annotations. Not much improvement was expected here, since they were turned off anyway, but some String operations and declarations were done in the old code even with annotations turned off.
`OpenPepXLLF took 22:17 m (wall), 44:28 m (CPU), 0.90 s (system), 44:27 m (user).`

The generation of neutral losses was determined to be the most inefficient part of the tool.
I simplified the generation of losses by precomputing a database of the losses I use for each residue.
`OpenPepXLLF took 06:51 m (wall), 13:40 m (CPU), 0.64 s (system), 13:40 m (user).`

Changed the SimpleTSGXLMS to use a vector of SimplePeak structs containing only an mz value and a charge as theoretical spectra. Sorting of spectra is done using a comparator only comparing the mz values. Not as big of an improvement, but mainly the sorting of spectra got more efficient, probably because there is no need to keep track of a separate DataArray of charges and sort it in parallel to the spectrum itself.
`OpenPepXLLF took 06:15 m (wall), 12:29 m (CPU), 1.05 s (system), 12:28 m (user).`

Further simplified the generation of neutral losses. Instead of using sets to keep losses for each peptide fragment unique (I know, was never a good idea in the first place, but was still 10x faster than the old code in the original TheoreticalSpectrumGenerator at the time)
now I precomputed the masses for the two losses I actually use (H20 and NH3) and use a struct of two booleans for each residue or peptide fragment to keep track of when I need to apply each neutral loss.
`OpenPepXLLF took 03:09 m (wall), 06:18 m (CPU), 0.93 s (system), 06:17 m (user).`

In the `Residue` class the full residue mass is precomputed for each residue, but getting the monoisotopic weight using `getMonoWeight()` for each fragment type used stuff like `EmpiricalFormula("H2O")` and calculating the weights from chemical compositions for each call.
I defined constants in the `Residue` class that get computed once, making the `getMonoWeight()`
function of `Residue` a simple addition for each fragment type. This should speed up every code we have, that uses this function without any changes in the results, I did not have to update any tests for this.
Final time:
`OpenPepXLLF took 02:33 m (wall), 05:06 m (CPU), 1.23 s (system), 05:05 m (user).`

So I reduced the runtime of the entire tool by a factor of almost 25.

@timosachsenberg we should implement something similar for the losses in the TheoreticalSpectrumGenerator, but maybe in a more general way if we need more types of neutral losses there.


enetz:
Added a filter that cancels analysis of a candidate if it does not have at least 2 matched linear peaks per peptide. This avoids the generation of more complex theoretical spectra and additional alignments for matches that I would not trust after manual validation of the spectra anyway. This of course does change the results. It reduces the number of completely useless hits in the output and therefore the output file size. The runtime improvement from this will probably be more variable depending on the dataset.
New time:
`OpenPepXLLF took 01:46 m (wall), 03:33 m (CPU), 0.44 s (system), 03:32 m (user).`

timosachsenberg:
That's a really nice speedup. Would it be possible to introduce some templated methods that can deal with both simple and standard spectra?
Have you also tried if you get some speedup if you use two vectors (one for mz and one for charge) instead of a vector of simplepeaks?

enetz:
@timosachsenberg I have not tried using separate vectors for mz and charge, because that would make much of the code more complex again. The point was to avoid having a complex sorting algorithm that has to sort multiple vectors by the values in one of them. The sorting is still one of the slowest parts of the tool. Maybe I will try an experiment to see how much faster it is to sort a `vector < double >` than a `vector < SimplePeak >`, but I am not planning to make that change in my tool anytime soon.
With the current struct I don't think it makes sense to make templated methods. It would be possible if I added functions like `getMZ()` to make the interface more similar, but the only function in OpenPepXL where that would work would be `XQuestScores::matchOddsScore(...)`. I duplicated the alignment algorithm to make it simpler by removing everything intensity related and the collection of mass deviations for each match. But I need the intensity checks for aligning two experimental spectra, so I can't remove the old version either. These are the only two functions where I use SimplePeaks.

enetz:
@timosachsenberg for completeness' sake I did an artificial experiment on sorting `double` and `SimplePeak` vectors. Sorting 50k vectors of length 2k that start off in the reverse order:
`double`: 16 sek
`SimplePeak`: 22 sek

There is a speedup of about 25% for the sorting when using simple `double` vectors. In OpenPepXL, sorting spectra takes about 30% of the total runtime now, so I could reduce the runtime by 25% of 30%, so about 7.5% total, but would have to add the overhead of a more complex sorting to keep two vectors synchronized.

src/openms/include/OpenMS/CHEMISTRY/SimpleTSGXLMS.h
```diff
@@ -0,0 +1,309 @@
...
+
+        This function generates a vector of sets. Each set contains the possible neutral losses for a specific suffix of the peptide.
+
+       * @param peptide The peptide or ion for which to collect possible losses
+       */
+      std::vector< LossIndex > getBackwardLosses_(AASequence & peptide) const;
+
+      // TODO copied from normal TSG, but it is protected over there. Move it to Residue class maybe?
+      /// helper for mapping residue type to letter
+      char residueTypeToIonLetter_(Residue::ResidueType res_type) const;
```
timosachsenberg: hmm sounds as if it could be moved to Residue class

src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp
```diff
@@ -350,7 +364,141 @@ namespace OpenMS
...
+        ++e;
+      }
+      else if (d > 0) // theo. peak is left of exp. peak (outside of tolerance window)
+      {
+        ++t;
+      }
+    }
+  }
+
+  void OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentSimple(
```
timosachsenberg: hmm ok I see your point with the two different interfaces but there seems to be a lot of code duplication.
probably doesn't get much more readable if we would squeeze that into a templated method

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
```diff
@@ -465,10 +479,134 @@ using namespace OpenMS;
...
+        double new_rel_error_weight = -0.03;
+        double new_score = new_match_odds_weight * std::log(1e-7 + match_odds) + new_rel_error_weight * abs(rel_error);
+
+        csm.score = new_score;
+        csm.match_odds = match_odds;
+        csm.match_odds_alpha = match_odds_alpha;
+        csm.match_odds_beta = match_odds_beta;
+        csm.precursor_error_ppm = rel_error;
+
+        mainscore_csms_spectrum.push_back(csm);
```
timosachsenberg: is it thread save to add it here? I did not check the whole context...

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorXLMS.cpp
```diff
@@ -574,15 +595,19 @@ namespace OpenMS
...
-    String ion_name("[M+H]");
+    String ion_name;
 
     // precursor peak
     double mono_pos = precursor_mass + (Constants::PROTON_MASS_U * static_cast<double>(charge));
     p.setMZ(mono_pos / static_cast<double>(charge));
     p.setIntensity(pre_int_);
     if (add_metainfo_)
     {
+      ion_name = "[M+H]";
```
timosachsenberg: do you need to copy the ion name into a variable before pushing it into ion_names? Otherwise, 
ion_names.emplace_back("[M+H]") could be slightly faster.

src/openms/include/OpenMS/CHEMISTRY/SimpleTSGXLMS.h
enetz: Yes, I can do that. And adapt all spectrum generators to use that.

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
enetz: This vector only exists within the context of each thread. It holds the main score of each candidate for one spectrum.

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorXLMS.cpp
enetz: I need to check where I can use that. I think in some cases I build up the strings from multiple components and it might not work there as easily, but for precursor peaks like this one it should work.
I will clean up the class a bit and also implement the faster neutral loss generation into this algorithm, but right now I only use this slower spectrum generation for the top X hits per spectrum. With my standard setting of top 5, this takes less than 0.5% of the total tool runtime on a small test example and will be even less with bigger protein databases.

timosachsenberg: Ok it's not that important right now.

enetz:
The test files were not deleted, they are just much smaller now. That is because of the last optimization step, where I stop analyzing candidates that don't have at least 2 linear ion matches per peptide. Previously the top 5 hits for each spectrum were always reported, no matter how bad they were. Now those that don't pass this low bar are skipped for the more complicated analysis steps and therefore also don't get reported.
I need to implement some kind of quality filter for the reporting of results and this is a first step and a very low bar in my opinion.

src/openms/include/OpenMS/CHEMISTRY/Residue.h
```diff
@@ -390,6 +390,9 @@ namespace OpenMS
     bool isInResidueSet(const String& residue_set);
     //@}
 
+    /// helper for mapping residue types to letters for Text annotations and labels
+    static char residueTypeToIonLetter(ResidueType res_type);
```
timosachsenberg: const ResidueType& might allow more aggressive optimizations

src/openms/include/OpenMS/CHEMISTRY/SimpleTSGXLMS.h
```diff
@@ -0,0 +1,303 @@
...
+
+namespace OpenMS
+{
+  class AASequence;
+
+  /**
+      @brief Generates theoretical spectra for cross-linked peptides
+
+      The spectra this class generates are normal vectors of SimplePeaks.
+
```
hroest: can you add some comments on how this is different from the regular TheoSpectrumGenerator?

hroest:
@enetz are any of your improvements applicable to the regular TheoSpectrumGenerator? Eg the precomputation of certain losses -- could that also speed up the regular class? If not, can you add a note / warning to the docu of the TheoSpectrumGenerator that the class is slow and that developers should look at your implementation on how to make it fast?

enetz:
@hroest my implementation of neutral losses is very limited and only considers H2O and NH3 losses. But the method should be extensible to any number of additional losses if needed. I just don't know which losses, if any, other people use with the regular TSG. We should certainly implement something similar there. Also the changes to Residue::getMonoWeight() should lead to an improvement for the regular TSG already, because it uses that function often as well.

I will add some documentation.

PR #3889, opened 2018-12-10
[STYLE] LPWrapper: avoid pointers, use const, use STL containers

pasqoo:
A few improvements to the class. I'm using it in some code of mine and it was very annoying to have to pass non-`const` variables to functions/methods that actually did not need to change the passed data.

This cleanup does not change the behaviour of the class.

**Advantages**
- the user is not required to passo non-`const` variables
- it better informs the user, because `const`-correctness assures that passed data won't be changed
- in some cases, copying of data is avoided (or delayed, when the right conditions are `true`, i.e. `if (solver_ == SOLVER_GLPK)`)
- avoid usage of `new`, no need of `delete`-ing memory
- it should be easier to spot out-of-range bugs (which I suspect might be present: glpk's routines assume 1-indexed arrays while we commonly assume 0-indexed arrays/vectors. It's easy to mistake some `<= n` / `< n` comparison or similar)

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/LPWrapper.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/LPWrapper.h
@@ -148,11 +148,11 @@ public:
 
     // problem creation/manipulation
     /// adds a row to the LP matrix, returns index
-    Int addRow(std::vector<Int> row_indices, std::vector<double> row_values, const String& name);
+    Int addRow(const std::vector<Int>& row_indices, const std::vector<double>& row_values, const String& name);
     /// adds an empty column to the LP matrix, returns index
     Int addColumn();
     /// adds a column to the LP matrix, returns index
-    Int addColumn(std::vector<Int> column_indices, std::vector<double> column_values, const String& name);
+    Int addColumn(const std::vector<Int>& column_indices, const std::vector<double>& column_values, const String& name);
 
     /**
       @brief Adds a row with boundaries to the LP matrix, returns index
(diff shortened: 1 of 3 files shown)


hroest:
great, this looks like a good start, but there is more to do (german comments, documentation etc) - it might make sense to consistently fix the other methods as well @timosachsenberg ?

hroest:
@timosachsenberg are there any builds where we do not use COINOR in OpenMS? Is there any reason / environment where we want to use GLPK (it seems COINOR is always better and faster, right)? Do we even test both solvers?

pasqoo:
From my side, I am done with this PR. Hopefully in future I'll have some more time to dedicate to _LPWrapper_.
I am trying to figure out what's wrong with _MRMFeatureSelector_, specifically: why do tests get different results between the various platforms (win, mac, linux).
Feel free to merge.

hroest:
overall looks good and if tests run through, we should merge.

I looked through the code and all new/delete were removed but I have one question for @timosachsenberg : 

https://github.com/OpenMS/OpenMS/blob/e1a181fb7216e9a4f786790c90cd258776a9ffe5/src/openms/source/DATASTRUCTURES/LPWrapper.cpp#L594

Doesnt this leak memory, what happens to the old `CoinModel` - it seems that this does not get deleted! Should this be 

```
  delete model_;
  model_ = new CoinModel(filename.c_str()); 
```

or am I missing something?

timosachsenberg:
I think you are right. the model should be deleted here and also in the destructor

hroest:
I have made the edit in 58f0afb0c74c986c3069617968c0ad90e68ae436

pasqoo:
Tests look good. Just a timelimit hit in travis.

PR #3893, opened 2018-12-12
Drift time unit

hroest:
Since a few months, `MS:1002815` has been introduced to store ion mobility in volt-second per square centimeter in addition to milliseconds. This PR introduces code in the mzML handler that properly handles both cases and stores the unit in an additional enum `DriftTimeUnit`. Due to historic reasons, this needs to be stored both in the Precursor and the MSSpectrum object 

- This requires an update of the OBO file


hroest:
it seems the newest obo files is actually defective and I fixed the OpenMS warnings with 38a4811d68b10a000a007193bdc746499a1d93c3

src/openms/source/ANALYSIS/OPENSWATH/DATAACCESS/SpectrumAccessOpenMSCached.cpp
```diff
@@ -77,11 +76,9 @@ namespace OpenMS
         "Error while changing position of input stream pointer.", filename_cached_);
     }
 
-    Internal::CachedMzMLHandler::readSpectrumFast(mz_array, intensity_array, ifs_, ms_level, rt);
-
     OpenSwath::SpectrumPtr sptr(new OpenSwath::Spectrum);
-    sptr->setMZArray(mz_array);
-    sptr->setIntensityArray(intensity_array);
+    sptr->getDataArrays() = Internal::CachedMzMLHandler::readSpectrumFast(ifs_, ms_level, rt);
```
hroest: this fixes a bug where caching of ion-mobility enabled data would not work properly

src/openms/include/OpenMS/KERNEL/MSSpectrum.h
```diff
@@ -489,6 +502,9 @@ namespace OpenMS
     /// Drift time
     double drift_time_;
 
+    /// Drift time unit
+    DriftTimeUnit drift_time_unit_;
```
hroest: @timosachsenberg the question here is whether we want to have `drift_time_` and `drift_time_unit_` as attributes of MSSpectrum or of SpectrumSettings. In one way its more like retention time, so I feel it should be in the spectrum. 

On the other hand, its also unclear why the `native_id_` is in the SpectrumSettings and the `name_` is in the MSSpectrum. Maybe the correct approach is to move all members into SpectrumSettings and only keep data in MSSpectrum?

hroest: PS: why do we even have a `name_` attribute? It is not set in any of the handlers including mzML, mzXML and mzData

src/openms/include/OpenMS/KERNEL/MSSpectrum.h
```diff
@@ -191,10 +194,20 @@ namespace OpenMS
...
     /**
-      @brief Returns the ion mobility drift time in milliseconds
+      @brief Sets the ion mobility drift time
     */
     void setDriftTime(double dt);
 
+    /**
+      @brief Returns the ion mobility drift time unit
+    */
+    DriftTimeUnit getDriftTimeUnit() const;
```
timosachsenberg: return const& ?

src/openms/include/OpenMS/KERNEL/MSSpectrum.h
timosachsenberg: please feel free to clean this up and decide for yourself. I would be fine having it in the spectrum.

src/openms/include/OpenMS/METADATA/Precursor.h
```diff
@@ -80,6 +80,16 @@ namespace OpenMS
       SIZE_OF_ACTIVATIONMETHOD
     };
 
+    ///Drift time unit
+    enum DriftTimeUnit
```
timosachsenberg: if nothing speaks against enum class we could use that in new code

src/openms/include/OpenMS/METADATA/Precursor.h
```diff
@@ -154,6 +164,17 @@ namespace OpenMS
     /// sets the ion mobility drift time in milliseconds
     void setDriftTime(double drift_time);
 
+    /**
+      @brief Returns the ion mobility drift time unit
+    */
+    DriftTimeUnit getDriftTimeUnit() const;
```
timosachsenberg: const& ?

src/openms/include/OpenMS/METADATA/Precursor.h
```diff
@@ -154,6 +164,17 @@ namespace OpenMS
...
+
+    /**
+      @brief Sets the ion mobility drift time unit
+    */
+    void setDriftTimeUnit(DriftTimeUnit dt);
```
timosachsenberg: const DriftTimeUnit& ?

timosachsenberg:
looks good! only minor stuff ... feel free to merge

PR #3897, opened 2018-12-13
[NOP] whitespace changes, remove dead code (exceptions)

hroest:
- remove exceptions that cannot be reached
- use else if for mutually exclusive statements
- add brackets (style)


hroest:
hmm looks like the removal of the exceptions triggers warnings about "control reaches end of non-void function" - so maybe we have to leave them in for now?

cbielow:
The exceptions are meant to protect against forgetting to add code in case a new solver is added.
Why remove them? Its not performance critical code in any way.

PR #3904, opened 2018-12-18
[FEATURE] add "acq_method_name" metavalue to ChromeleonFile parser

pasqoo:
Parse the `Channel` info from the data file and add it to the experiment as a new metavalue named `"acq_method_name"`.

The `Channel` info is found twice within the file (the first time at line 2, and then within the raw information data). Same info repeated. The code would simply overwrite the first "set" value.


timosachsenberg:
+1

PR #3905, opened 2018-12-20
Feature/metabo decharger ppm

FabianAicheler:
Added support for ppm-based mass errors.

Assumes simplification that such error window is symmetric around the observed mz (e.g., +- exp_mz*5*10^-6). (Instead of computing error windows based on putative causative true mzs leading to observed exp_mz).

Analogously to the previous absolute Thomson tolerance case, observed mass shifts then have to match to entries in the mass shift table, +- the sum of the ppm tolerances of both affected observed mzs.


FabianAicheler:
Might have to adapt the whitelist to look at the groups. Currently thinking about how to verify that grouping is correct. Probably use the consensusXML output with the groups instead of the annotated featureXML

PR #3906, opened 2018-12-21
Fix ElementDB and Elements.xml

cbielow:
ASAN reported mem leaks for ElementDB. This fixes it, and some other issues:

[FIX] ElementDB: memory leak (lost pointer, [1], e.g. when atomic number already exists (e.g. Hydrogen vs. Deuterium have both '1') and double free (duplicate pointer to same Element [2])

[FIX] fix mass of Pb208 in Elements.xml (also fixes a mem-leak)

[FIX] ElementDB: fix properties of last element 'Samarium' in File (not constructed correctly)

And to avoid future issues:
[FEATURE] ElementDB now reports inconsistencies in Elements.xml (or future versions of it) to avoid mem-leaks

[FIX] make construction of ElementDB thread-safe (using Myers Singleton -- instead of a pointer)

[1] https://github.com/OpenMS/OpenMS/blob/ac4c8767657504e98572baf453c41dcf1d7fff15/src/openms/source/CHEMISTRY/ElementDB.cpp#L197
[2] https://github.com/OpenMS/OpenMS/blob/ac4c8767657504e98572baf453c41dcf1d7fff15/src/openms/source/CHEMISTRY/ElementDB.cpp#L222

### DIFF
--- a/src/openms/source/CHEMISTRY/ElementDB.cpp
+++ b/src/openms/source/CHEMISTRY/ElementDB.cpp
@@ -200,7 +200,7 @@ namespace OpenMS
       }
       if (symbols_.has(iso_symbol))
       {
-        std::cerr << "Error: ElementDB encountered duplicated names for \n" << *symbols_[iso_symbol] << "\n" << *iso_e << "\nKeeping only the first one!\n";
+        std::cerr << "Error: ElementDB encountered duplicated symbol for \n" << *symbols_[iso_symbol] << "\n" << *iso_e << "\nKeeping only the first one!\n";
         delete e;
         return;
       }
--- a/src/openms/source/CHEMISTRY/ElementDB.cpp
+++ b/src/openms/source/CHEMISTRY/ElementDB.cpp
@@ -161,7 +161,7 @@ namespace OpenMS
     }
     if (atomic_numbers_.has(an))
     {
-      std::cerr << "Error: ElementDB encountered duplicated atomic_numbers_ for \n" << *atomic_numbers_[an] << "\n" << *e << "\nKeeping only the first one!\n";
+      std::cerr << "Error: ElementDB encountered duplicated atomic number for \n" << *atomic_numbers_[an] << "\n" << *e << "\nKeeping only the first one!\n";
       delete e;
       return; // next element
     }
--- a/src/openms/source/CHEMISTRY/ElementDB.cpp
+++ b/src/openms/source/CHEMISTRY/ElementDB.cpp
@@ -110,19 +110,11 @@ namespace OpenMS
   double ElementDB::calculateAvgWeight_(const Map<UInt, double>& Z_to_abundance, const Map<UInt, double>& Z_to_mass)
   {
     double avg = 0;
-    // extract Zs
-    vector<UInt> keys;
+    // calculate weighted average
     for (Map<UInt, double>::const_iterator it = Z_to_abundance.begin(); it != Z_to_abundance.end(); ++it)
     {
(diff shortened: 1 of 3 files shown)


cbielow:
not sure, but the tests run fine, so probably not, since it would throw.
There is a way to allow additional elements, but it will be a bit more work in the D'tor (first collecting all pointers from all Maps, removing duplicates and then delete them).

cbielow:
and why would you explicitly use "D", but normal isotopes for all other elements?

cbielow:
D and T are also not part of the PSE.

Ready for merge, unless there are objections due to these two.

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -141,6 +133,82 @@ namespace OpenMS
...
+    }
+    if (symbols_.has(symbol))
+    {
+      std::cerr << "Error: ElementDB encountered duplicated symbol for \n" << *symbols_[symbol] << "\n" << *e << "\nKeeping only the first one!\n";
+      delete e;
+      return; // next element
+    }
+    if (atomic_numbers_.has(an))
+    {
+      std::cerr << "Error: ElementDB encountered duplicated atomic_numbers_ for \n" << *atomic_numbers_[an] << "\n" << *e << "\nKeeping only the first one!\n";
```
hroest:
```suggestion
      std::cerr << "Error: ElementDB encountered duplicated atomic number for \n" << *atomic_numbers_[an] << "\n" << *e << "\nKeeping only the first one!\n";
```

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -141,6 +133,82 @@ namespace OpenMS
...
+      iso_isotopes.set(iso_container);
+
+      /*
+      // print name, symbol and atomic mass of the current isotope
+      cout << "Isotope Name: " << iso_name << " Symbol: " << iso_symbol << " AtomicMass: " << iso_mono_weight << endl;
+      */
+      Element* iso_e = new Element(iso_name, iso_symbol, an, iso_avg_weight, iso_mono_weight, iso_isotopes);
+      if (names_.has(iso_name))
+      {
+        std::cerr << "Error: ElementDB encountered duplicated names for \n" << *names_[iso_name] << "\n" << *iso_e << "\nKeeping only the first one!\n";
```
hroest:
```suggestion
        std::cerr << "Error: ElementDB encountered duplicated name for \n" << *names_[iso_name] << "\n" << *iso_e << "\nKeeping only the first one!\n";
```

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -141,6 +133,82 @@ namespace OpenMS
...
+        delete e;
+        return;
+      }
+      if (symbols_.has(iso_symbol))
+      {
+        std::cerr << "Error: ElementDB encountered duplicated names for \n" << *symbols_[iso_symbol] << "\n" << *iso_e << "\nKeeping only the first one!\n";
```
hroest:
```suggestion
        std::cerr << "Error: ElementDB encountered duplicated symbol for \n" << *symbols_[iso_symbol] << "\n" << *iso_e << "\nKeeping only the first one!\n";
```

timosachsenberg:
I access only "Hydrogen"
https://github.com/OpenMS/OpenMS/blob/develop/src/utils/MetaProSIP.cpp#L1527
and take the second isotope so it should be fine

PR #3910, opened 2019-01-08
[FIX] MSSimulator thread-safe now

cbielow:
on rare occasions, MSSimulator would crash when using >1 threads, since Boost RNGs are not threadsafe.
This is fixed now.

Also fixes a `bad_alloc` when the user specifies protein intensities manually which are numeric_limits<uint>. This is now resulting in
```
exception of type InvalidValue occured in line 325, function void __cdecl OpenMS::IonizationSimulation::ionizeEsi_(class OpenMS::FeatureMap &,class OpenMS::ConsensusMap &) of C:\dev\openmsqt5\src\openms\source\SIMULATION\IonizationSimulation.cpp
error message: the value '6.67539e+18' was used but is not valid; Protein abundance is too high. Please use values in [0,4294967295]!
```

### DIFF
--- a/src/openms/source/SIMULATION/IonizationSimulation.cpp
+++ b/src/openms/source/SIMULATION/IonizationSimulation.cpp
@@ -268,7 +268,6 @@ public:
                    boost::bind(std::multiplies<double>(), _1, 10));
     for (size_t i = 0; i < weights.size(); ++i)
       std::cout << "weights[" << i << "]: " << weights.at(i) << std::endl;
-    boost::random::discrete_distribution<Size, double> ddist(weights.begin(), weights.end());
 
     try
     {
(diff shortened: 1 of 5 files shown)


timosachsenberg:
looks good

src/openms/source/SIMULATION/IonizationSimulation.cpp
```diff
@@ -288,179 +287,194 @@ namespace OpenMS
...
+
+            if (charge == 0)
+            {
+              continue;
+            }
+
+            /////
+// distribute charges across adduct types
+/////
+Compomer cmp;
```
timosachsenberg: indentation?

timosachsenberg:
please add a test for MSSimulator

src/openms/source/SIMULATION/IonizationSimulation.cpp
cbielow: fixed

PR #3911, opened 2019-01-08
Fix MaRaClusterAdapter Exectuable Check Fail in KNIME

Leon-Bichmann:


src/topp/MaRaClusterAdapter.cpp
```diff
@@ -282,19 +282,6 @@ class MaRaClusterAdapter :
     const String maracluster_executable(getStringOption_("maracluster_executable"));
     writeDebug_(String("Path to the maracluster executable: ") + maracluster_executable, 2);
 
-    if (maracluster_executable.empty())  //TODO? - TOPPBase::findExecutable after registerInputFile_("maracluster_executable"... ???
```
jpfeuffer: I think you can leave the check for empty. This should really not happen.

Leon-Bichmann:
so i get the following exit code:

`Executing maracluster ...
Executing: maracluster
FATAL: External invocation of maracluster failed. Standard output and error were:


255`

jpfeuffer:
Hmm unfortunately this is not very clear/satisfying.
Maybe we have to check QProcess::error() to get a better defined error message. Would be best to do it in the runExternalProcess function.

Leon-Bichmann:
I see.. not sure how to do that. Do you have an example code snippet?

jpfeuffer:
Somewhere after qp.waitForFinished() just add:
```
if (qp.error() == QProcess::FailedToStart)
{
  LOG_ERROR << "Process failed to start. Does it exist? Is it executable?" << std::endl;
}
```

Leon-Bichmann:
`Executing maracluster ...
Process failed to start. Does it exist? Is it executable?`

jpfeuffer:
Great. But ;) : could you do it in the runExternalProcess function, so that it is fixed for all the adapters that use this function?
E.g. here:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/APPLICATIONS/TOPPBase.cpp#L1648

Leon-Bichmann:
argh.. ^^ ok now I understand what you mean

cbielow:
how about:
```
LOG_ERROR << "Process '" << executable << "' failed to start. Does it exist? Is it executable?" << std::endl;
```

jpfeuffer:
I'll merge. Just Sirius and IonizationSim (unrelated).

PR #3913, opened 2019-01-08
[REFACTOR] function to copy spectrum meta data / settings

hroest:
extract single function to copy spectral settings (but not data) from one spectrum to another


src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
```diff
@@ -212,6 +212,28 @@ namespace OpenMS
...
+  /**
+   * @brief Copies only the meta data contained in the input spectrum to the output spectrum.
+   * 
+   * @note Actual data is not copied.
+   *
+   * @param[in] input The input spectrum.
+   * @param[out] output The output spectrum (will be cleared and will contain all metadata of the input spectrum).
+   **/
+  static void copySpectrumMeta(const MSSpectrum & input, MSSpectrum & output, bool clear_spectrum = true)
+  {
```
cbielow: since this is not a template, implementation should go to .cpp

src/openms/include/OpenMS/TRANSFORMATIONS/RAW2PEAK/PeakPickerIterative.h
```diff
@@ -301,15 +302,8 @@ namespace OpenMS
...
       if (input.size() < 3) return;
 
-      // copy meta data of the input spectrum
-      output.clear(true);
-      output.SpectrumSettings::operator=(input);
-      output.MetaInfoInterface::operator=(input);
-      output.setRT(input.getRT());
-      output.setMSLevel(input.getMSLevel());
-      output.setName(input.getName());
-      output.setType(SpectrumSettings::CENTROID);
```
cbielow:
```
output.setType(SpectrumSettings::CENTROID);
output.getFloatDataArrays().clear();
```
are effectively removed now.  But it seems like they are needed.

src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp
```diff
@@ -670,17 +671,9 @@ namespace OpenMS
...
-    out.setRT(old_spectrum.getRT());
-
-    out.setNativeID(old_spectrum.getNativeID());
-    out.setInstrumentSettings(old_spectrum.getInstrumentSettings());
-    out.setAcquisitionInfo(old_spectrum.getAcquisitionInfo());
-    out.setSourceFile(old_spectrum.getSourceFile());
-    out.setDataProcessing(old_spectrum.getDataProcessing());
-    out.setType(old_spectrum.getType());
-    out.setMSLevel(old_spectrum.getMSLevel());
-    out.setName(old_spectrum.getName());
```
cbielow: also here, some information is lost

src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerSH.cpp
```diff
@@ -63,13 +65,7 @@ namespace OpenMS
     Size progress = 0;
     for (Size scan_idx = 0; scan_idx != input.size(); ++scan_idx)
     {
-      output[scan_idx].clear(true);
-      output[scan_idx].SpectrumSettings::operator=(input[scan_idx]);
-      output[scan_idx].MetaInfoInterface::operator=(input[scan_idx]);
-      output[scan_idx].setRT(input[scan_idx].getRT());
-      output[scan_idx].setMSLevel(input[scan_idx].getMSLevel());
-      output[scan_idx].setName(input[scan_idx].getName());
-      output[scan_idx].setType(SpectrumSettings::CENTROID);
```
cbielow: information of type lost.

src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerSH.cpp
```diff
@@ -63,13 +65,7 @@ namespace OpenMS
...
+      copySpectrumMeta(input[scan_idx], output[scan_idx]);
```
hroest:
```suggestion
      copySpectrumMeta(input[scan_idx], output[scan_idx]);
      output[scan_idx].setType(SpectrumSettings::CENTROID);
```

src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
```diff
@@ -212,6 +212,28 @@ namespace OpenMS
...
     std::swap(p_new, p);
   }
+
+  /**
+   * @brief Copies only the meta data contained in the input spectrum to the output spectrum.
+   * 
+   * @note Actual data is not copied.
+   *
+   * @param[in] input The input spectrum.
+   * @param[out] output The output spectrum (will be cleared and will contain all metadata of the input spectrum).
```
timosachsenberg: add doxygen comment for clear_spectrum

timosachsenberg:
otherwise +1

src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp
cbielow: SpectrumSettings::operator= covers it. all good.

cbielow:
what about the `MetaInfoInterface::operator=`? That is not done here, right?

src/openms/include/OpenMS/KERNEL/SpectrumHelper.h
```diff
@@ -33,6 +33,9 @@
 // --------------------------------------------------------------------------
 
 #include <OpenMS/MATH/STATISTICS/StatisticFunctions.h>
+#include <OpenMS/METADATA/DataArrays.h>
+#include <OpenMS/KERNEL/MSSpectrum.h>
+#include <OpenMS/KERNEL/MSChromatogram.h>
```
cbielow: MSChromatogram is not required, right?

and `#pragma once` should come first :)

hroest:
> what about the `MetaInfoInterface::operator=`? That is not done here, right? (i.e. get"s lost)

MSSpectrum does not inherit directly from MetaInfoInterface, so that should be covered by SpectrumSettings:

https://github.com/OpenMS/OpenMS/blob/1c5cf83322b564ccc2953c36301f47aa1d22e4e4/src/openms/include/OpenMS/KERNEL/MSSpectrum.h#L67-L70

timosachsenberg (approved):

timosachsenberg:
tests passed according to cdash

PR #3914, opened 2019-01-08
Feature/pyopenms string

hroest:
proposal for dealing with functions that return OpenMS::String in pyopenms for Python 3.x

this proposal implements a solution where the returned string is either
-  converted to a Python String assuming UTF8 encoding (99% of all cases)
- a `OpenMS::String` object is returned if that is not possible (<1% of all cases)

this is not a great solution as the user does not know which object s/he will get and it could be either an OpenMS String or a Python string. However, it makes code easier to read for the most common use cases. Allows one to write in Python3

```
>>> from pyopenms import *
>>>
>>> methanol = EmpiricalFormula("CH3OH")
>>> water = EmpiricalFormula("H2O")
>>> wm = EmpiricalFormula(str(water) + str(methanol))
>>> print(water, "and", methanol)
H2O1 and C1H4O1
```

and

```
ustr = u"bläh"
s = pyopenms.MSSpectrum()
s.setNativeID(ustr)
s.getNativeID()
assert( isinstance(s.getNativeID(), str) )

s.setNativeID(ustr.encode("iso8859_15"))
r = s.getNativeID()
assert( isinstance(r, pyopenms.String) ) #uglyness, first its a str, now a pyopenms.String
```


hroest:
Note: this will fail the test, this is just as a basis for discussion

another idea is to return a "bytes" object in case of failure instead of a "OpenMS::String" object

oliveralka:
That looks great! I think to return 'OpenMS::String' in case of failure is good, since it will permit to perform basic String operations:  

Whitout your function (in case of "bytes" object):
```
wm = EmpiricalFormula(str(water) + str(methanol))
__str__ returned non-string (type bytes) 
```
in case of OpenMS::String: 
```
wm = EmpiricalFormula(water.toString() + methanol.toString())
```
and you would always be able to decode it manually afterwards:
```
print(wm.toString().decode('utf-8'))
C1H6O2
```

hroest:
> I think to return 'OpenMS::String' in case of failure is good, since it will permit to perform basic String operations:

the question is whether its easier / more "pythony" to differentiate between a `str` or `bytes` object or to differentiate between `str` and `OpenMS::String`

secondly, this case would only occur if we have non-ASCII characters in an encoding that is not UTF8 and there is only a few places where these could come from and in principle we can control these:

- XML parsing (see xerces discussion)
- command line (we had a discussion on that as well, probably not relevant for Python)
- Python itself (e.g. you can do `s.setNativeID(ustr.encode("iso8859_15"))`)

in all these cases we could either tell the user to do not do that or ensure we have all UTF8 inside OpenMS

src/pyOpenMS/addons/ADD_TO_FIRST.pyx
```diff
@@ -9,21 +9,16 @@ cdef inline convOutputString(_String _r):
...
 
     try:
         py_result = c_string[:length].decode('UTF-8')
     except UnicodeDecodeError:
-        # TODO : what should we return? we know its not UTF8 but what the hell
-        # is it? Best option - return an OpenMS String object?
-        res = shared_ptr[_String](new _String(_r))
-        alt_result.inst = res
-        return alt_result
+        py_result = c_string[:length]
```
hroest: @oliveralka I have now changed this from returning an OpenMS String to a `bytes` object, somehow I feel that this is more in line with what Python does and I dont see the advantage of OpenMS::String over `bytes`

src/pyOpenMS/tests/unittests/test000.py
```diff
@@ -467,11 +471,11 @@ def testIdentificationHit():
     assert pyopenms.IdentificationHit().setRank is not None
     assert pyopenms.IdentificationHit().getRank is not None
 
-    f.setId(b"test_id")
-    assert f.getId() == b"test_id"
+    f.setId("test_id")
+    assert f.getId() == "test_id"
```
hroest: most cases now are simple strings, no more bytes

hroest:
One caveat remains: strings inside maps / sets / vectors currently remain bytes which is a bit inconsistent and I am not sure how much of a problem that is. I think there is still a fundamental question

- do we want to be completely consistent but annoying (keep everything in bytes) or do we want to get things "right" most of the time but require the user to check whether they get back `str` or `bytes` every single time?

@timosachsenberg @jpfeuffer some comments?

timosachsenberg:
I think we should definitly be able to write:
    f.setId("test_id")
    assert f.getId() == "test_id"
as you did.

Regarding the container, I guess it is fine if we properly document that in the doc/tutorial.
We sometimes return set<String> or vector<String> so we could think about adding a method to pyOpenMS that converts it to the python set /  lists. Would that make sense?

hroest:
@timosachsenberg @oliveralka looks ready to merge, any further comments?

oliveralka:
Looks good from my side!

PR #3916, opened 2019-01-09
FuzzDiff Performance improvements

cbielow:
## FuzzyDiff speed
comparing "large" files (6MB) using FuzzyDiff takes about 9 seconds on VS2015 Release(!), whereas its rather fast (0.3 sec) on GCC.
This PR speeds up FuzzyDiff by avoiding the  `stringstream>>double operator` which is awfully slow on VS.
Test time on Windows is now also 0.3 secs for this special case.
In general the combined time of TOPP+UTIL tests (1117 tests), many of which use FuzzyDiff, went from 55 sec to 36 sec (10 jobs parallel) on VS. GCC builds also profit, but not as much (102 sec --> 82 sec).

## reduces time for the stopwatch test 
12 sec to 0.5 sec by reducing waiting time

### DIFF
--- a/src/openms/include/OpenMS/SYSTEM/StopWatch.h
+++ b/src/openms/include/OpenMS/SYSTEM/StopWatch.h
@@ -93,6 +93,10 @@ public:
     /** Start the stop watch.
             The stop watch is started. If the stop watch is already running, <b>false</b>
             is returned.
+
+            If the watch holds data from previous measurements, these will be reset before starting up,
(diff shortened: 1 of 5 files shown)


timosachsenberg:
@jpfeuffer  Do you think that decreasing the wait time will cause errors?

jpfeuffer:
Well the StopWatch test is very fragile and can only be tested roughly. I think he adapted the expected results accordingly and therefore I think it should be fine.

PR #3917, opened 2019-01-09
Storing Unicode in OpenMS::String

hroest:
- Allow generation of unicode strings from a `char*` which was not possible before as the conversion would abort at zero bytes, thus truncating data
- Adds tests for the new behavior
- Add pyOpenMS wrappers that utilize the new behavior

This basically makes the two constructors

```
String(const char* s, size_type count)
std::string(const char* s, size_type count)
```

behave the same way as before the `std::string` would not stop at zero bytes while the `OpenMS::String` would. Quite a bit of surprising behaviour there! See

https://en.cppreference.com/w/cpp/string/basic_string/basic_string

"Constructs the string with the first count characters of character string pointed to by s. s can contain null characters. The length of the string is count. The behavior is undefined if [s, s + count) is not a valid range."

Allows the following:

```
  // Unicode test
  char test_utf8[] =  "T\xc3\xbc\x62ingen";
  char test_utf16[] = "\xff\xfeT\x00\xfc\x00b\x00i\x00n\x00g\x00e\x00n\x00";
  char test_iso8859[] = "T\xfc\x62ingen";

  String s_utf8(test_utf8, 9);
  String s_utf16(test_utf16, 18);
  String s_iso8859(test_iso8859, 8);

  TEST_EQUAL(s_utf16.size(), 18);
  TEST_EQUAL(s_utf8.size(), 9);
  TEST_EQUAL(s_iso8859.size(), 8);
```

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
@@ -124,7 +124,7 @@ namespace OpenMS
     {
       String res;
       size_t count = 0;
-      while (count < length && *(s + count) != 0)
+      while (count < length)
       {
         res += *(s + count);
         ++count;
--- a/src/tests/class_tests/openms/source/String_test.cpp
+++ b/src/tests/class_tests/openms/source/String_test.cpp
@@ -62,41 +62,41 @@ START_TEST(String, "$Id$")
 String* s_ptr = nullptr;
 String* s_nullPointer = nullptr;
 START_SECTION((String()))
-	s_ptr = new String;
+  s_ptr = new String;
   TEST_NOT_EQUAL(s_ptr, s_nullPointer)
 END_SECTION
 
 START_SECTION(([EXTRA] ~String()))
-	delete s_ptr;
+  delete s_ptr;
 END_SECTION
 
 START_SECTION((String(const QString &s)))
-	QString qs("bla");
-	String s(qs);
-	TEST_EQUAL(s=="bla",true)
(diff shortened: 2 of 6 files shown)


src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
```diff
@@ -124,7 +124,7 @@ namespace OpenMS
     {
       String res;
       size_t count = 0;
-      while (count < length && *(s + count) != 0)
```
hroest: @cbielow @timosachsenberg I hope you agree with this change, it aligns us with `std::string`

src/tests/class_tests/openms/source/String_test.cpp
```diff
@@ -62,255 +62,352 @@ START_TEST(String, "$Id$")
...
+  TEST_EQUAL(s2,"")
+
+  String s3("abcdedfg",8);
+  TEST_EQUAL(s3,"abcdedfg")
+
+  // This function does *not* check for null bytes as these may be part of a
+  // valid string! If you provide a length that is inaccurate, you are to blame
+  // (we do not test this since we would cause undefined behavior).
+  // String s4("abcdedfg", 15);
+  // TEST_NOT_EQUAL(s4,"abcdedfg")
```
hroest: Note: this wont work any more as the parser will not stop at zero

hroest:
rebuild jenkins

cbielow:
looks good!

timosachsenberg:
looks good

PR #3918, opened 2019-01-10
[ADD] additional scores to LuciPhorAdapter ouptut

oliveralka:
Adds local and global false localization rate to the idXML output.

For an explanation of the available scores see:
http://luciphor.sourceforge.net/output_fields.html


cbielow:
looks good to me +1

timosachsenberg:
I am issing a test that fails after that change ;)

oliveralka:
Haha, it seems that there is no thirdparty test for the LuciPhorAdapter.
I will add one.

Issue #3861, opened 2018-11-22
CometAdapter fixed modifications not properly working


mwalzer:
specifically, _fixed modifications_ seem to be **ignored**:
```
Warning - invalid parameter found: add_X_unspecified/unknown.  Parameter will be ignored.
```
(Comet version "2016.01 rev. 3")

mwalzer:
`CometAdapter -in BSA.mzML -database proteins.fasta -out BSA.idXML -variable_modifications 'Carbamidomethyl (C)' 'Oxidation (M)' -fixed_modifications 'Acetyl (N-term)'`

jpfeuffer:
Weird.. it doesn't find our own modifications? Which OpenMS commit/tag?

mwalzer:
I think it happens during the transition from TOPP parameters to the comet param.txt

PR #3919, opened 2019-01-10
[FIX] CometAdapter terminal modification 

oliveralka:
It seems that the terminal modification was not detected properly which lead to the error

> Warning - invalid parameter found: add_X_unspecified/unknown.  Parameter will be ignored 

and in the end, lead to a failure of CometAdapter.

Example:
```
Value of string option 'fixed_modifications': TMT6plex (K)
Value of string option 'fixed_modifications': TMT6plex (N-term)
```
diff param_before.txt param_after.txt
```
< add_X_unspecified/unknown = 229.163
---
> add_Nterm_peptide = 229.163
```

probably related to #3861 

Further I am not sure if we should write more explicite to set the "clear_mz_range" parameter when using isobaric tags (with the corresponding ranges)- please see: http://comet-ms.sourceforge.net/notes/20171005_isotopiclabeling.php

### DIFF
--- a/src/topp/CometAdapter.cpp
+++ b/src/topp/CometAdapter.cpp
@@ -473,8 +473,10 @@ protected:
     {
       for (vector<ResidueModification>::const_iterator it = fixed_modifications.begin(); it != fixed_modifications.end(); ++it)
       {
-        String AA = it->getOrigin();
-        if ((AA!="N-term") && (AA!="C-term"))
+        // check whether amino acid or terminal modification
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Looks good. I see the error. Yes, I think excluding the range would make sense, if that is how Comet solves the problem of the fragment tags.
Another improvement would be adding binary modification groups to enable single searches for SILAC experiments.

PR #3921, opened 2019-01-11
Circleci Example - NO MERGE

hroest:
This is an example CircleCI file that works for OpenMS and has been tested on my private repository. The nice thing is that this all runs on docker and does not have a time limit. It also seems to work well for pull requests and we could consider enabling it for testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,13 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: debian:stretch
+    steps:
+      - checkout
+      - run:
+          name: Greeting
+          command: echo Hello, world.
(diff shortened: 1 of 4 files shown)


cbielow:
looks good.
Their docs say that we get one VM with 4GB RAM and 2 Cores for free. But it seems we're capped to single-threaded? It will take some time to finish all builds which we currently run on travis (but other than Travis, at least they will finish), correct?

hroest:
>  It will take some time to finish all builds which we currently run on travis (but other than Travis, at least they will finish), correct?

yes, that would be the idea. However, now with travis we have 90 minutes, so it should now finish (hopefully)

PR #3922, opened 2019-01-13
[FIX] fix travis error

hroest:


hroest:
fixes at least the build error ...

PR #3927, opened 2019-01-17
FeatureFinderMultiplex - fix for overlapping, unresolved peptide features

lars20070:
The high resolution of modern mass spectrometers make it unlikely that isotopic peaks from two different peptide features overlap, see first figure. Peaks can be resolved and feature detection is not problematic.

In some special cases, peaks are too close to be distinguished. See second figure. Take for example, labelling with light/heavy dimethyl (UniMod 36 and 199) and consider a 4+ peptide pair. The fifth isotopic peak of the light peptide is at +1.00335 Th, and the mono-isotopic peak of the heavy peptide is at + 1.00628 Th. Both peaks overlap and connot be resolved.

(1) This PR ensures that each peak is attributed to only one peptide feature. (In the above example, the mono-isotopic peak of the heavy peptide. This peak would not fit into the averagine model of the light one.) This ensures that features are correctly detected. Old `TOPP` tests remain unchanged. An example for the special case is added.

(2) A small bug is removed, for the [correction factor](https://github.com/OpenMS/OpenMS/pull/3927/files#diff-a4a0a740e6b1905e206e7f5a0ea51757R315) of peptide intensities in triplets.

40e0f8eb-4b4b-4038-932f-19db3e8837d1


lars20070:
@timosachsenberg Charge deconvolution?
At the moment, I check if fold changes from different charge states are consistent. Will adds this to our [summary reports](https://github.com/ProteomicsScripts/ProteomicsScripts/blob/master/mzTab2report/misc/example_5.pdf). Might then use this to improve fold changes.

lars20070:
Think I get your idea. Might even split the contributions of this peak into peptides 1 and 2.
Currently, the entire peak contributes to 2 only.

timosachsenberg:
One thing to note: If you do the deconvolution during feature finding, you can only assume the averagine composition. If you resolve intensities after ID mapping, I know the exact elemental composition and you can do a better deconvolution.

lars20070:
Good point. But not part of this PR.

PR #3928, opened 2019-01-18
[FEATURE] OpenPepXL: more efficient memory usage

enetz:
- Improved memory efficiency by using pointers: Enumeration of candidates of peptide pairs now points to the list of single peptides instead of copying the AASequences for each pair.

- Reduced memory usage by reworking the parallelization: previously the parallelized loop was looping over the spectra. The lists of candidates and results for each spectrum had to be enumerated and computed for each spectrum in parallel and this increased the memory usage proportional to the number of used threads. Now several parallelized loops loop over the candidate enumeration and result computation for one spectrum. This makes the parallelization slightly less efficient, but it removes the increase in memory usage for multiple threads.

I tested it on a data set with a 256 protein database, where some spectra have more than 10 million peptide pair candidates within a 6 ppm precursor tolerance window. I previously needed somewhere between 75GB and 100GB memory to run it on 50 threads on a cluster (2GB for each thread does not sound much, until you have dozens of them :) ). Now this data set needs around 5GB.

### DIFF
--- a/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
+++ b/src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
@@ -47,7 +47,7 @@ using namespace std;
 namespace OpenMS
 {
   // Enumerate all pairs of peptides from the searched database and calculate their masses (inlcuding mono-links and loop-links)
-  vector<OPXLDataStructs::XLPrecursor> OPXLHelper::enumerateCrossLinksAndMasses(const vector<OPXLDataStructs::AASeqWithMass>&  peptides, double cross_link_mass, const DoubleList& cross_link_mass_mono_link, const StringList& cross_link_residue1, const StringList& cross_link_residue2, const vector< double >& spectrum_precursors, vector< int >& precursor_correction_positions, double precursor_mass_tolerance, bool precursor_mass_tolerance_unit_ppm)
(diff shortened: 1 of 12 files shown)


src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -549,8 +562,9 @@ namespace OpenMS
             cross_link_candidate.cross_linker_mass = cross_link_mass;
             cross_link_candidate.cross_linker_name = cross_link_name;
             cross_link_candidate.precursor_correction = precursor_corrections[i];
-            cross_link_candidates.push_back(cross_link_candidate);
 
+#pragma omp critical (cross_link_candidates_access)
```
cbielow: again, all those criticals can be avoided if using a private vector per thread

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
```diff
@@ -429,9 +429,9 @@ using namespace OpenMS;
     progresslogger.startProgress(0, 1, "Matching to theoretical spectra and scoring...");
     Size spectrum_counter = 0;
 
-#ifdef _OPENMP
-#pragma omp parallel for schedule(guided)
-#endif
+// #ifdef _OPENMP
```
cbielow: the omp pragmas below can go as well since this is a serial region now, right?

cbielow:
another hint: [guided](https://stackoverflow.com/questions/42970700/openmp-dynamic-vs-guided-scheduling) might not always be a good idea. You might be better off using dynamic.

timosachsenberg:
@cbielow thanks for the valuable insights. I'll merge this PR as it passes CI.

enetz:
I will try out the suggestions and open another PR later.

PR #3930, opened 2019-01-21
numeric mass shifts for FeatureFinderMultiplex

lars20070:
The labelling of the samples is specified in the parameter `algorithm:labels`. For example, `[Dimethyl0][Dimethyl4][Dimethyl8]` is dimethyl labelling of three samples with light, medium and heavy. The label string must be one of the labels specified in the advanced/hidden `labels` section. In this section each label string is mapped to a specific mass shift.

This pull request allows to specify mass shifts directly in the `algorithm:labels` parameter, for example `[0][6.03705]`. A further `TOPP` test for this case is added.


src/openms/source/TRANSFORMATIONS/FEATUREFINDER/MultiplexDeltaMassesGenerator.cpp
```diff
@@ -123,38 +123,66 @@ namespace OpenMS
...
+        double mass_shift;
+        try
+        {
+          mass_shift = std::stod(samples_labels_[i][j]);
+          
+          // For numeric mass shifts, long and short label names as well as the numerical mass shift are trivial.
+          // For example, long label name ("3.1415"), short label name ("3.1415") and numerical mass shift (3.1415).
+          label_delta_mass_.insert(make_pair(samples_labels_[i][j], mass_shift));
+          label_short_long_.insert(make_pair(samples_labels_[i][j], samples_labels_[i][j]));
+          label_long_short_.insert(make_pair(samples_labels_[i][j], samples_labels_[i][j]));
```
hroest: will there be any problem if line 147 and 148 go through but 149 throws and the three vectors are not of equal lengths?

lars20070: No. Line 143 throws the exception and 147-49 are never reached. (Just tested with debug output.)
The vectors are always of equal length.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/MultiplexDeltaMassesGenerator.cpp
```diff
@@ -435,48 +481,48 @@ namespace OpenMS
...
       {
-        LOG_INFO << samples_labels_[i][j] << "    ";
+        std::cout << samples_labels_[i][j] << "    ";
       }
-      LOG_INFO << "\n";
+      std::cout << "\n";
     }
   }
   
   void MultiplexDeltaMassesGenerator::printDeltaMassesList() const
```
timosachsenberg: you could pass the output stream as parameter so the user could choose if it is written to std::out, a stringstream, a file, or log_info/debug ...

timosachsenberg: no that's not what I mean.
you add a parameter to your method: print...(ostream& theStream)
then the user can decide which stream get's the output.
you can search for "passing stream as parameter c++" to see some examples

lars20070: Done. Much nicer. Thanks.

src/pyOpenMS/pxds/MultiplexDeltaMassesGenerator.pxd
```diff
@@ -13,8 +13,8 @@ cdef extern from "<OpenMS/TRANSFORMATIONS/FEATUREFINDER/MultiplexDeltaMassesGene
         MultiplexDeltaMassesGenerator(String labels, int missed_cleavages, libcpp_map[ String, double ] label_mass_shift) nogil except +
 
         void generateKnockoutDeltaMasses() nogil except +
-        void printSamplesLabelsList() nogil except +
-        void printDeltaMassesList() nogil except +
+        void printSamplesLabelsList(std::ostream &stream) nogil except +
```
lars20070: @hroest Could you please comment? I am unsure how to fix the `pxd`. Thanks.

timosachsenberg: I think this cannot be wrapped in python (but it should be trivial to just print the array in python).

lars20070: So I remove both methods from the wrapper for now?

PR #3937, opened 2019-01-23
[FIX] ASan error

jpfeuffer:
From cppreference:
Hexadecimal escape sequences have no length limit and terminate at the first character that is not a valid hexadecimal digit. If the value represented by a single hexadecimal escape sequence does not fit the range of values represented by the character type used in this string literal (char, char16_t, char32_t, or wchar_t), the result is unspecified.


hroest:
:+1: thanks, you learn something new every day (specifications of hexadecimal escape sequences in C++ however is not something I really did want to learn about)

PR #3939, opened 2019-01-25
Simple mass tag calculator - NO MERGE

timosachsenberg:
Calculate sequence tags from m/z values

Current restrictions and potential extensions:
  - all peaks are considered charge 1
 - first prefix/suffix ion don't give rise to first character in tags (as currently ion types are not specified)
  - modifications are not supported
  - distance to a missing precursor fragment could be considered
  - gaps are not supported
  - turn recursion into iteration


timosachsenberg:
Speed ~2ms per unfiltered spectrum

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,77 @@
...
+    *  - gaps are not supported
+    **/
+  class OPENMS_DLLAPI Tagger
+  {
+    public:
+      // initalize tagger with minimum/maximum tag length and +- tolerance ppm
+      Tagger(size_t min_tag_length, float ppm, size_t max_tag_length = 65535);
+
+      // generate tags from mass vector @mzs using the standard residues in ResidueDB
+      void getTag(const std::vector<float>& mzs, std::set<std::string>& tags) const;
```
hroest: why use a float here? You loose accuracy and also all our MSSpectra use `double`, so converting from double to float will take additional time.

PR #3946, opened 2019-01-29
Feature/rn pxl log reg

timosachsenberg:
- removed some unneeded variables
- slight speedup of searches with full scoring
- allow to filter peptide hits by metavalue
- score tuned by logistic regression


enetz:
:+1:  for the new IDFilter option. Makes it possible to avoid some ID score changing and filtering by String meta values. I was thinking that I need something like this for a while, too.
Could you add a test for that?

The rest seems fine too, except the Novel Scoring commit, because the diff in GitHub seems to think the whole file was deleted and replaced and I don't see what changed.

timosachsenberg:
rebuild jenkins

PR #3948, opened 2019-01-30
Feature/im pqp

hroest:
- store library ion mobility drift time in PQP file
- better error handling (check SQLite return value)
- better readability (whitespace changes)
- support for old and new format (checking whether sqlite column exists) with correct query for both formats


src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
```diff
@@ -37,6 +37,36 @@
 namespace OpenMS
 {
 
+  bool columnExists(sqlite3 *db, const String& tablename, const String& colname)
```
timosachsenberg: Would it make sense to move these kind of general methods to a SQLite helper class?

hroest: yes, I will do that in a separate PR

hroest: yes, I have done that now

PR #3949, opened 2019-01-31
[FIX] LuciphorAdapter - ConsensusID 

oliveralka:
Workflow using multiple search engines lead to small differences in the retention time (identification vs. spectra). Error in LuciphorAdapter.

Using the ConsensusID - PepXMLFile will use the RT for mapping - here the default rt in SpectrumLookup is 0.01.

To be able to use a higher rt_tolerance:
- the parameter was introduced to LuciphorAdapter
- the variable was added to PepXMLFile::store (with the same default as used in SpectrumLookup) 

e.g.
rt_tolerance: 0.05
spectrum_rt: 1724.67
lower_diff: 0.02492
upper_diff: 0.1117

### DIFF
--- a/src/topp/LuciphorAdapter.cpp
+++ b/src/topp/LuciphorAdapter.cpp
@@ -183,6 +183,7 @@ protected:
     setValidStrings_("run_mode", ListUtils::create<String>("0,1"));
 
     registerDoubleOption_("rt_tolerance", "<num>", 0.01, "Set the retention time tolerance (for the mapping of identifications to spectra in case multiple search engines were used)", false);
+    setMinFloat_("rt_tolerance", 0.0);
     
     registerInputFile_("java_executable", "<file>", "java", "The Java executable. Usually Java is on the system PATH. If Java is not found, use this parameter to specify the full path to Java", false, false, ListUtils::create<String>("skipexists"));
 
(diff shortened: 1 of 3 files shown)


jpfeuffer:
IMHO This is as good as it gets for a quick workaround. In the long term you probably want to adapt
ConsensusID so that it either allows execution without tolerance (i.e. 1:1 mapping by spectrum_reference) or that it also sets a spectrum_reference (e.g. taking the one from the best PSM or a combination of all the spectra involved).

src/topp/LuciphorAdapter.cpp
```diff
@@ -180,7 +180,9 @@ class LuciphorAdapter :
     setMinInt_("num_threads", 0);
 
     registerStringOption_("run_mode", "<choice>", "0", "Determines how Luciphor will run: 0 = calculate FLR then rerun scoring without decoys (two iterations), 1 = Report Decoys: calculate FLR but don't rescore PSMs, all decoy hits will be reported", false);
-    setValidStrings_("run_mode", ListUtils::create<String>("0,1")); 
+    setValidStrings_("run_mode", ListUtils::create<String>("0,1"));
+
+    registerDoubleOption_("rt_tolerance", "<num>", 0.01, "Set the retention time tolerance (for the mapping of identifications to spectra in case multiple search engines were used)", false);
```
jpfeuffer: Can you quickly set a minimum value of 0.0 here?

oliveralka:
@jpfeuffer pyopenms unrelated?

hroest:
> @jpfeuffer pyopenms unrelated?

yes, memoryleak is unrelated

> Since it is a parameter with standard value - I guess I do not have to add it to the python bindings (.pxd).

I advise against that since now it will be impossible to set the parameter from Python

oliveralka:
@hroest 
Ok, so I will add it to the .pxd. (new PR)

PR #3950, opened 2019-02-01
[FIX] LuciphorAdapter 

oliveralka:
Adjust LuciphorAdapter/PepXMLFile to work with "pep" score type.

Workflow for multiple search engines -> FileMerger -> PSMFeatureExtractor -> PeptideIndexer -> PercolatorAdapter -> IDFilter -> LuciPhorAdapter. 

Add previously added parameter "rt_tolerance" to PepXMLFile.pxd


jpfeuffer:
Cool. Maybe we can introduce CVs for the score_types during the dev meeting.

PR #3957, opened 2019-02-05
[FIX] fix for N-terminal modifications

hroest:
- fix for N-terminal modifications that are only allowed on specific
  amino acids, they can now be parsed properly.

- can now parese `C[143]PEPTIDEK` and `Q[111]PEPTIDEK`


timosachsenberg:
looks good.

PR #3963, opened 2019-02-07
update contrib (MSVC14 coinor Debug libs)

cbielow:


jpfeuffer:
I believe it was a hiccup.

timosachsenberg:
rebuild jenkins

PR #3966, opened 2019-02-11
Feature/sql cleanup

hroest:
- move sqlite-specific stuff to its own class
- move code to cpp files
- add tests for  MzMLSqliteHandler_test.cpp MzMLSqliteSwathHandler_test.cpp SqMassFile_test.cpp
- general cleanup


hroest:
its basically a lot about removing

```
rc = sqlite3_prepare_v2(db, "SELECT COUNT(*) FROM TRANSITION;", -1, &cntstmt, nullptr);
if (rc != SQLITE_OK)
```

with a standard way of calling the database executing the associated commands.

Issue #3962, opened 2019-02-06
Faster Number -> String conversion


cbielow:
Similar to switching to Boost:Qi for String->Number some time ago (#692), I did some microbenchmarks for converting a number (int, float, double) to String.
We are currently using std::stringstream, which is about the worst in terms of performance one could imagine.

I only tested converters which are readily available in OpenMS with C++11 and contrib. There are others (FMT lib for example), but I did not want to introduce yet another dependency.

Candidates:

- **stringstream** -- standard stringstream without precision modifiers (everything has precision of 6)
- **stringstreamFP** -- also a stringstream but using precision of 6 (for float) and 15 (for double); 
- **sprintf** -- the good old plain C function
- **std::to_string** -- available from C++11 onwards for all relevant types, but not really a viable candidate since we have no control how the string is formatted; basically, as [printf with format '%f'](http://www.cplusplus.com/reference/string/to_string/). e.g. the double `8.592e307` becomes a string with 308 chars!
- **boost::karma** -- sibling of boost::Qi, very fast, but different notion of precision; we use up to three decimal digits (default) here ***
- **boost::karmaFP** -- full precision variant, using 6 (float) and 15 (double) **_decimal_** digits
- **boost::lexical_cast** -- no control over precision [https://www.boost.org/doc/libs/1_65_0/doc/html/boost_lexical_cast.html)[docs], but included for completeness
- **boost::format** -- similar to sprintf with formatting allowed

In OpenMS we currently use a mixture of stringstream and stringstreamFP, depending on how an outputstream is used. If just operator << is used, precision is 6; however, when using String(double/float) or precisionWrapper(double/float), we get 6/15 precision. Note that precision in terms of stringstream means up to 6 or 15 digits in total (not only fractional(decimal) digits).
For Boost::Karma, we get something sligthly different, i.e. 6/15 fractional digits, no matter how large the integer part is (i.e. potentially larger strings).

Here is the plain benchmark data for converting 100000 numbers in the range of [0,1e6] to strings.
Length at each lines gives the summed length of the generated strings.
```
**Windows, VS14**
Bench for int:
stringstream: 112ms length: 888705
stringstreamFP: 113ms length: 888705
sprintf: 16ms length: 888705
to_string: 3ms length: 888705
karma: 13ms length: 888705
karmaFP: 11ms length: 888705
boostLC: 16ms length: 888705
toBoostFmt: 211ms length: 888705


Bench for float:
stringstream: 178ms length: 1088320
[...]

PR #3969, opened 2019-02-13
[FEATURE] faster number conversion to String using Boost::Karma

cbielow:
solves https://github.com/OpenMS/OpenMS/issues/3962

here are some additional benchmarks for **writing** larger idXML and featureXML files:

## IDFileConverter on 320 MB IdXML file 

**Win:**
old: 20.89 s
new: 12.59 s (-40%, 6% size increase)

**Linux**:
old: 14.21 s
new: 10.63 s (-26%, 6% size increase)

## FileConverter on 192 MB FeatureXML file

**Win**:
old: 31.16 s
new: 25.78 s (-18%, 3.5% size increase)

**Linux**:
old: 27.96 s
new: 25.85 s (-8%, 3.5% size increase)

The increase in featureXML is simply not that large because I did not look at the writer code (i.e. using the new String conversion instead of default `... << double << ...`). All benefits are implicit.

Windows benefits the most, since the stream operators were very inefficient. Its not that bad with GCC, but the speed still increases.

Edit: FuzzyDiff runs about 3x faster (Release build) now as well, after switching to Boost::Qi instead of `stringstream::operator >>` (which was required since we hit an AppleClang compiler bug which makes clang interpret `5e6` as 5 and then stop reading from the stream; this was not a problem before since all our output looked like this. Now with Boost::Karma we write `5.0e6` which clang reads in full; Boost::Qi just reads both input formats just fine).

### After this PR is merged, it is advised to maximize writing speed when outputting files like this:
`os << String(int/float/double) << ...` instead of just the slower `os << int/float/double << ...`.
This will give you full precision by default for float and doubles. If you only need up to three fractional digits use `String(my_number, false)`, i.e. `full_precision = false`.


hroest:
looks great, that should be a real improvement!

Can you look at these failed tests: https://travis-ci.org/OpenMS/OpenMS/jobs/492738641


```
The following tests FAILED:
	142 - ConsensusXMLFile_test (Failed)
	156 - IdXMLFile_test (Failed)
	160 - InspectInfile_test (Failed)
	201 - SequestInfile_test (Failed)
	211 - TraMLFile_test (Failed)
	1877 - TOPP_InspectAdapter_1 (Failed)
	1878 - TOPP_InspectAdapter_1_out1 (Failed)
	1879 - TOPP_InspectAdapter_2 (Failed)
	1880 - TOPP_InspectAdapter_2_out1 (Failed)
	2131 - TOPP_MaRaClusterAdapter_1_out_1 (Failed)
	2133 - TOPP_MaRaClusterAdapter_2_out_1 (Failed)
```

cbielow:
fun fact:
calling `maracluster.exe -p 20ppm   ...` yields a slightly different result than `maracluster.exe -p 20.0ppm   ...`, even though the parameter is specified as double in their help section explicitly. And integers are perfectly representable as doubles (no rounding error)... go figure...

cbielow:
more fun facts:
appleclang cannot parse `5.0e-06` as double using `stringstream >> d` (it will read 5.0 and then stop). All other compilers are just fine (and appleclang is also fine when reading `5.5e-06`.... 
see:
```
in1:  /Users/builder/jenkins/ws/openms_pr_ci/c797036f/build/src/tests/class_tests/openms/MascotXMLFile_test_249.tmp   (line: 646, position/column: 49/78)
				<UserParam type="float" name="EValue" value="!
				                                             5.0e-06"/>

in2:  /Users/builder/jenkins/ws/openms_pr_ci/c797036f/source/src/tests/class_tests/openms/data/MascotXMLFile_test_out_3.idXML   (line: 646, position/column: 49/78)
				<UserParam type="float" name="EValue" value="!
				                                             5e-06"/>
```

src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
```diff
@@ -53,65 +55,162 @@ namespace OpenMS
...
+      return str;
+    }
+
+
+
+    /// high precision (15 fractional digits) conversion to String
+    inline void append(double d, String& target)
+    {
+      std::back_insert_iterator<std::string> sink(target);
+      boost::spirit::karma::generate(sink, BK_PrecPolicyDouble, d);
```
timosachsenberg: Can sink be an rvalue? With direct construction in de method call generate

src/openms/include/OpenMS/DATASTRUCTURES/ListUtilsIO.h
```diff
@@ -55,21 +54,45 @@ namespace OpenMS
...
+      }
+      os << String(v.back());
+    }
+
+    os << "]";
+    return os;
+  }
+
+  template<typename T>
+  struct VecLowPrecision
```
timosachsenberg: Name looks not 100% OpenMS style ... but well ...

cbielow:
ok, this is ready for merge.

timosachsenberg:
I am a bit afraid of hitting the merge button ;) ... 
YOLO

Issue #1827, opened 2016-02-18
workflow tests


lars20070:
Both class tests [1] and TOPP tool tests [2] are essential for the stability of the OpenMS code. Could we introduce one further set of tests that checks the performance of complete workflows?

- class tests
- TOPP tool tests
- **workflow tests**

Often basic workflows such as protein identification and feature detection fail. It is then down to users to report these bugs. Most TOPP tools work well, but are not fully integrated with other tools in a workflow. IMO, workflow tests have several advantages.

- The basic ID and quantification workflows for data from various machines are always working.
- A single workflow test covers multiple TOPP tool tests and is easier to maintain.

On the other hand, it will be difficult to integrate the processing of complete (and possibly large) datasets into the testing. 

[1] https://github.com/OpenMS/OpenMS/tree/develop/src/tests/class_tests
[2] https://github.com/OpenMS/OpenMS/tree/develop/src/tests/topp

timosachsenberg:
This is definitely nice to have and one thing we are working towards. As you point out is a bit difficult to provide automated tests for large files. This means we need to set this up on our own infrastructure and use e.g. nightly tests.

jpfeuffer:
This is already possible with KNIME Testflows (if we manage to setup a KNIME build system on our own that allows us to store/test with big files).
I haven't thought about a solution for our nostalgic TOPPAS users yet.

lars20070:
Sounds good @jpfeuffer. I can convert the TOPPAS workflows from the repository [1] to KNIME and provide `mzML`.

Can we integrate the workflow tests with GitHub? Should workflow tests be able to block pull requests? Let's discuss in Northumberland.

[1] https://github.com/OpenMS/OpenMS/wiki/Workflow-Collection

jpfeuffer:
Well, yes it will be possible to trigger everything from a pull request from here, but I am not sure if we should do this PR based as it might take quite some additional time.
I think the dev meeting is a good place to discuss.

mwalzer:
That would be great.

timosachsenberg:
I close this discussion as this will be done using KNIME testflows

PR #3972, opened 2019-02-16
[UPDATE] unimod.obo and unimod.xml to latest version

### LINKED ISSUE #3960: Update UNIMOD: IDFileConverter cannot process 15N-label
Hi.

I tried converting an idXML file to mzid with IDFileConverter. The idXML came from an original MSGF+ search, where I used 15N-labels. 
All amino acids were labelled, so I used Unimod 994, 995, 996 and 897. The labels were processed so far without any error by several Toppas Tools.

However, IDFileConverter gave the following:
`Error: Unexpected internal error (the value 'UNIMOD:994' was used but is not valid; Invalid CV identifier!)`
The error does not occur, when converting into `pepXML`, though.

To reproduce the error, use this [input file](https://gist.github.com/Stortebecker/38440a267892d81b917c063643932ec2) and run `IDFileConverter -in INPUT_FILE.idXML -out INPUT_FILE.mzid`

used version: 2.3
platform: win10

lukaszimmermann:
Updates `unimod.obo` and `unimod.xml` to latest version. `UNIMOD:994` was not available.

Fixes #3960


timosachsenberg:
Thanks! Does it also contain the empty custom0 .. modifications we added manually?

lukaszimmermann:
No, I forgot this, will add them asap.

timosachsenberg:
rebuild jenkins

hroest:
seems like a whole bunch of tests fail, pleaes fix: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=g++-7,os_label=ubuntu_1804/2975/#showFailuresLink

hroest:
this is worrying, why does changing the file lead to a sefault:

```
	156 - IdXMLFile_test (Failed)
	175 - MzIdentMLFile_test (SEGFAULT)
	219 - XQuestResultXMLFile_test (Failed)
```

jpfeuffer:
@lars20070 can you add your Methyl:2H(2)13C modification and maybe suggest inclusion to Unimod?
@enetz can you add your custom crosslink modifications? Maybe we could store them in a separate file for easier updates of unimod?

enetz:
This is strange, I never added XL-MS mods to the unimod file, e.g. `Xlink:DSS` should be in there natively and I can find it on the unimod website. For those that are missing in unimod we have the XLMOD.obo.
So I don't know why the XL-MS tests suddenly can't find `Xlink:DSS` by name

enetz:
Ok, I think I see the problem now, in the new unimod.xml `Xlink:DSS` is gone and `Xlink:DSS[138]` was added instead. They added this cross-link in different versions with different numbers, but broke backwards compatibility by removing the old version. We could just replace `Xlink:DSS` with `Xlink:DSS[138]` in all test files and tests.

enetz:
On the website they have both versions, they have the same mass and composition.

jpfeuffer:
I see. Thanks for looking into this. But we included the correct file right? Or can one somewhere download an alternative file without the suffixes?
What are the suffixes by the way? Why are they needed?

If unimod keeps it that way, yes I would say updating the tests is the right way.

lars20070:
FYI, custom mods '#99988' and '#99989' no longer needed. They are now in Unimod as '#1826' and '#1827'.

timosachsenberg:
@lars This will break downwards compatibility.

enetz:
Updating my XL-MS test files caused merge conflicts with the current develop branch.
I merged with develop and fixed the test files.
There should be three remaining failing tests, that are unrelated to XL-MS stuff.

enetz:
The number suffixes on `Xlink:DSS` mods are to make it easier to differentiate between the different chemical entities that result from cross-linking using DSS. Unimod does not really have a way to encode cross-links, so they previously only had the two possible mono-links / dead-end links for DSS. One was called `Xlink:DSS` (water-quenched) and the other `Xlink:DSS-NH2` (ammonium-quenched). Now they changed the entries completely. `Xlink:DSS[138]` is the actual cross-link that was not there before (but defined as a normal mod binding to one side chain, because unimod still doesn't have a way to encode cross-links), `Xlink:DSS[156]` and `Xlink:DSS[155]` are the two quenched dead-end links. The website also has a summary entry under the old water-quenched name `Xlink:DSS` that seems to describe all three of those as neutral losses of the same modification, but this one is thankfully missing from the unimod.xml, it would only cause more confusion anyway.

timosachsenberg:
Can someone add tests to AASequence? Could make problems with our parsing of bracket mods.

enetz:
There are still failing tests in the AASequence tests. One test expects an Exception to be thrown, and fails because apparently none is thrown, but then there is an error message because an unexpected exception is thrown in the next test.

It looks like the definition of this mod changed in unimod. The tests imply it can only modify C at the N-term. The new unimod entry says it can modify any C + some other residues and any Protein N-term. It became much less specific and our tests are too strict on it.

So it makes sense that the first of those tests fails:

`-line 1369:  TEST_EXCEPTION(Exception::InvalidValue,AASequence::fromString(".DC(UniMod:1009)CFPIANGER.")): no exception thrown!`

The second test is confusing, because it tests an N-term mod that should be allowed in unimod ("(UniMod:1009).CFPIANGER."), but it can't find the mod. I am not sure what to do here:

`Error: Caught unexpected exception of type 'InvalidValue' thrown in line 941 of file '/home/eugen/Development/OpenMS/src/openms/source/CHEMISTRY/AASequence.cpp' in function 'static String::ConstIterator OpenMS::AASequence::parseModRoundBrackets_(const String::ConstIterator, const OpenMS::String &, OpenMS::AASequence &, const ResidueModification::TermSpecificity &)' - Message: the value 'Cannot convert string to peptide modification. No N-terminal modification matches to term specificity and origin.' was used but is not valid; (UniMod:1009).CFPIANGER.`

jpfeuffer:
This is still relatively high priority.

jpfeuffer:
We might need to make a new PR but it would be great if we manage to update this for the next release.

jpfeuffer:
@enetz This would be really important. Unfortunately, your recent changes in your test data probably overwrote the changes in this PR. Would you be willing to update this PR (with newest Unimod and your new test data)?
I would do it, but I have no idea how to resolve your test data.

Regarding the previously failed tests, just change test nr 1.
Failing test nr 2 needs some small investigation though.

enetz:
This PR is being redone in #4803

Issue #3943, opened 2019-01-27
target-decoy column in PSM section of mzTab


lars20070:
According to mzTab 1.0.0 specs 5.12.5, target-decoy information can/should be stored in a column `opt_global_cv_MS:1002217_decoy_peptide` with boolean values `(0, 1)`. This column is implemented in MaxQuant 1.6.3.3. OpenMS writes this information in a column `opt_global_target_decoy` with values `(target, decoy, target+decoy)` instead.

Can we add a column `opt_global_cv_MS:1002217_decoy_peptide` as suggested in the specs? The column is optional, but our `mzTab` would become more compliant to the standard.

timosachsenberg:
sure if you want you can remove our column and add the opt_global_cv_MS:1002217_decoy_peptide yourself

lars20070:
Our column is not binary. So I intend to map.
```
target --> 0
decoy --> 1
target + decoy --> 0
```
The last case is very unlikely, [0.1%](https://github.com/ProteomicsScripts/ProteomicsScripts/blob/master/mzTab2PSMreport/misc/example.pdf) of all PSMs in many analyses.

lukaszimmermann:
I took the liberty to address this

lars20070:
@lukaszimmermann Thank you! Very helpful.

jpfeuffer:
Should be done, right? Reopen if not.

PR #3973, opened 2019-02-16
MzTabExporter: Rename opt_global_target_decoy to opt_global_cv_MS:1002217_decoy_peptide

lukaszimmermann:
Replaces the `opt_global_target_decoy ` column in `MzTabExporter` by the `opt_global_cv_MS:1002217_decoy_peptide` column by remapping like:
```
target --> 0
decoy --> 1
target+decoy --> 0
```
This addresses #3943 

For consistency, I also renamed the column in the PRT section, but this is not specified by the standard (we could also revert back to the previous name here).

I kept `null` values. The specification is  ambiguous whether this is allowed.

Test files were adapted accordingly using a script which remapped the old test files.


src/topp/MzTabExporter.cpp
```diff
@@ -1096,6 +1102,27 @@ namespace OpenMS
       MzTabFile().store(out, mztab);
       return EXECUTION_OK;
     }
+  private:
+      static void remap_target_decoy(vector<MzTabOptionalColumnEntry> &opt)
```
timosachsenberg: rename to: remapTargetDecoy_

timosachsenberg:
just a quick question. For the commits it seems as if you remap when adding a row.
Would it make sense to just rename the single column before returning the mzTab object?
This would keep the code a bit more generic (aka turning meta values into optional columns) and then just fix it later?

lukaszimmermann:
Yes. Actually I think that approach is better, because then the phases of construcing rows from user values and remapping are clearly separated. (The would then be separated on section level, now they are separated on row level) I will change that.

lukaszimmermann:
Actually I now remember why I did the mapping this way:
The types:
```
MzTabPeptideSectionRows
MzTabProteinSectionRows
MzTabPSMSectionRows
```
are unrelated, but their respective entries share the  `opt_` field.
Now this implies duplicated code if the `opt_` field is changed in all three sections. 

Would it be an option to pull the `opt_` field into a base struct that is extended by 
`MzTabPSMSectionRow` et al. ? This would make the remapping of optional columns more explicit.

lukaszimmermann:
So compared to the previous solution, the loop for the rows of all sections needs to be duplicated.

src/topp/MzTabExporter.cpp
```diff
@@ -369,16 +367,16 @@ namespace OpenMS
         row.mass_to_charge = MzTabDouble(f.getMZ());
         MzTabDoubleList rt_list;
         vector<MzTabDouble> rts;
-        rts.push_back(MzTabDouble(f.getRT()));
+        rts.emplace_back(MzTabDouble(f.getRT()));
```
cbielow: the point of `emplace_back` is to avoid the move, yet you are constructing a temporary (so this code does not actually run faster).
You should be using `rts.emplace_back(f.getRT());` to get the performance benefit. Or did I miss the point here?
Applies also for other replacements below (especially make_pair).

lukaszimmermann: True, is used `emplace_back wrongly here.

timosachsenberg:
rebuild jenkins

lars20070:
Can we merge?

lukaszimmermann:
No, because I have not yet addressed the issue raised by @cbielow

cbielow:
code looks good.
`TOPP_MzTabExporter_*` seems to fail on Travis.

lars20070:
On Travis it's just a time-out as far as I can see. I started a rebuild.

lars20070:
On Travis `MzTabExporter` was not build. Hence the tests did not run and failed.
```
Could not find executable /home/travis/build/OpenMS/OpenMS/_build/bin/MzTabExporter
```
Apparently, fou compiler errors.
```
Error(s) when building project
   4 Compiler errors
   28 Compiler warnings
```
Where can I find the compiler errors? Want to fix the PR.

timosachsenberg:
CDASH. If you rebuild it via travis it should show up there.
Or your own build

lars20070:
FYI. Some warnings.
```
pr-3998-Windows_10_x64-msvc-15-Release-x64: 479>C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\topp\MzTabExporter.cpp(304): warning C4081: expected 'identifier'; found 'string' [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\topp\MzTabExporter.vcxproj]
pr-3998-Windows_10_x64-msvc-15-Release-x64: 479>C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\topp\MzTabExporter.cpp(464): warning C4081: expected 'identifier'; found 'string' [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\topp\MzTabExporter.vcxproj]
pr-3998-Windows_10_x64-msvc-15-Release-x64: 479>C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\topp\MzTabExporter.cpp(689): warning C4267: 'argument': conversion from 'size_t' to 'const int', possible loss of data [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\topp\MzTabExporter.vcxproj]
pr-3998-Windows_10_x64-msvc-15-Release-x64: 479>C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\topp\MzTabExporter.cpp(804): warning C4081: expected 'identifier'; found 'string' [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\topp\MzTabExporter.vcxproj]
pr-3998-Windows_10_x64-msvc-15-Release-x64: 479>C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\topp\MzTabExporter.cpp(852): warning C4081: expected 'identifier'; found 'string' [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\topp\MzTabExporter.vcxproj]
pr-3998-Windows_10_x64-msvc-15-Release-x64: 479>C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\topp\MzTabExporter.cpp(1060): warning C4081: expected 'identifier'; found 'string' [C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\topp\MzTabExporter.vcxproj]
```

lars20070:
Here the error.
```
                 from /.../OpenMS/src/topp/MzTabExporter.cpp:35:
/usr/include/c++/4.8/ext/new_allocator.h: In instantiation of ‘void __gnu_cxx::new_allocator<_Tp>::construct(_Up*, _Args&& ...) [with _Up = std::pair<OpenMS::String, OpenMS::MzTabString>; _Args = {const char (&)[29], OpenMS::String}; _Tp = std::pair<OpenMS::String, OpenMS::MzTabString>]’:
/usr/include/c++/4.8/bits/alloc_traits.h:254:4:   required from ‘static typename std::enable_if<std::allocator_traits<_Alloc>::__construct_helper<_Tp, _Args>::value, void>::type std::allocator_traits<_Alloc>::_S_construct(_Alloc&, _Tp*, _Args&& ...) [with _Tp = std::pair<OpenMS::String, OpenMS::MzTabString>; _Args = {const char (&)[29], OpenMS::String}; _Alloc = std::allocator<std::pair<OpenMS::String, OpenMS::MzTabString> >; typename std::enable_if<std::allocator_traits<_Alloc>::__construct_helper<_Tp, _Args>::value, void>::type = void]’
/usr/include/c++/4.8/bits/alloc_traits.h:393:57:   required from ‘static decltype (_S_construct(__a, __p, (forward<_Args>)(std::allocator_traits::construct::__args)...)) std::allocator_traits<_Alloc>::construct(_Alloc&, _Tp*, _Args&& ...) [with _Tp = std::pair<OpenMS::String, OpenMS::MzTabString>; _Args = {const char (&)[29], OpenMS::String}; _Alloc = std::allocator<std::pair<OpenMS::String, OpenMS::MzTabString> >; decltype (_S_construct(__a, __p, (forward<_Args>)(std::allocator_traits::construct::__args)...)) = <type error>]’
/usr/include/c++/4.8/bits/vector.tcc:97:40:   required from ‘void std::vector<_Tp, _Alloc>::emplace_back(_Args&& ...) [with _Args = {const char (&)[29], OpenMS::String}; _Tp = std::pair<OpenMS::String, OpenMS::MzTabString>; _Alloc = std::allocator<std::pair<OpenMS::String, OpenMS::MzTabString> >]’
/.../OpenMS/src/topp/MzTabExporter.cpp:713:77:   required from here
 /usr/include/c++/4.8/ext/new_allocator.h:120:4: error: no matching function for call to ‘std::pair<OpenMS::String, OpenMS::MzTabString>::pair(const char [29], OpenMS::String)’
   { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }
    ^
```

src/topp/MzTabExporter.cpp
```diff
@@ -706,10 +710,7 @@ namespace OpenMS
         row.calc_mass_to_charge = best_ph.getCharge() != 0 ? MzTabDouble(aas.getMonoWeight(Residue::Full, best_ph.getCharge()) / best_ph.getCharge()) : MzTabDouble();
 
         // add opt_global_modified_sequence in opt_ and set it to the OpenMS amino acid string (easier human readable than unimod accessions)
-        MzTabOptionalColumnEntry opt_entry;
-        opt_entry.first = String("opt_global_modified_sequence");
-        opt_entry.second = MzTabString(aas.toString());
-        row.opt_.push_back(opt_entry);
+        row.opt_.emplace_back("opt_global_modified_sequence", aas.toString());
```
lars20070: @lukaszimmermann Can you please check the `emplace_back`? Seems to cause the build error.

lars20070: @cbielow Can I simply add a `make_pair()`?

cbielow: `emplace_back()` only considers explicit ctors. So using 
`row.opt_.emplace_back(String("opt_global_modified_sequence"), MzTabString(aas.toString()));` should work

lars20070: Lukas needs to switch on `Allow edits from maintainers.`first. Right?

lukaszimmermann: I think you should be able to push to this branch.

lars20070:
rebuild jenkins

lars20070:
Good to merge?

lukaszimmermann:
Not sure, there is a failed test on MSVC

lars20070:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=305824

Just the ini file version number. Right?

cbielow:
the first test ran, but quit with error code (which does not look like an OpenMS return code, but rather this one https://stackoverflow.com/questions/23409809/windows-7-exception-code-0xc0000409).

hroest:
:-1: until the MSVS test is fixed, seems like it accesses memory it is not supposed to

lars20070:
rebuild jenkins

jpfeuffer:
superseded by #4094

PR #3976, opened 2019-02-20
Faster write of (id)XML

cbielow:
even faster writing of idXML (mostly) by avoiding needless copies and using number->string conversion of OpenMS instead of std::ostream.

Also:
[FIX] Fix a bug in ClassTest.h's `TEST_FILE_EQUAL(....)` which would report the line of the previous TEST_ macro as point of failure, see https://github.com/OpenMS/OpenMS/compare/develop...cbielow:fasterWrites?expand=1#diff-b194e44883fffa6c069caae5d0164329R709


timosachsenberg:
looks good to me

src/openms/source/FORMAT/IdXMLFile.cpp
```diff
@@ -647,12 +653,12 @@ namespace OpenMS
...
         for (std::vector<String>::const_iterator it = accessions.begin(); it != accessions.end(); ++it)
         {
-          std::map<String, String>::const_iterator it2 = proteinid_to_accession_.find(*it);
+          const auto it2 = proteinid_to_accession_.find(*it);
           if (it2 != proteinid_to_accession_.end())
           {
             PeptideEvidence pe;
             pe.setProteinAccession(it2->second);
-            peptide_evidences_.push_back(pe);
+            peptide_evidences_.push_back(std::move(pe));
```
hroest: is PeptideEvidence moveable?

cbielow: yes it is. If it were not, this would just make it an r-value and copy assignment would kick in.

PR #3977, opened 2019-02-20
Feature/gene

hroest:
add GeneName to tsv / TraML / pqp file, allowing users to also map peptides to gene loci


hroest:
this is with the idea that we can implement a gene-level FDR downstream...

src/openms/include/OpenMS/FORMAT/SqliteConnector.h
```diff
@@ -162,6 +162,15 @@ namespace OpenMS
 
 
 
+    /**
+      @brief Checkes whether the given table exists
```
cbielow: -> `Checks`

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
```diff
@@ -153,7 +153,8 @@ namespace OpenMS
     // Insert run_id information
     std::stringstream sql_run;
     sql_run << "INSERT INTO RUN (ID, FILENAME) VALUES ("
-            << *(int64_t*)&run_id_ << ", '" // Conversion from UInt64 to int64_t to support SQLite
+            // Conversion from UInt64 to int64_t to support SQLite (and conversion to 63 bits)
+            <<  static_cast<int64_t >(run_id_ & ~(1UL << 63)) << ", '"
```
cbielow: Why do you need this?
It run_id_ potentially larger than 2^63? Otherwise a simple cast would do, right?

src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
```diff
@@ -456,10 +489,18 @@ namespace OpenMS
       int group_set_index = group_map[peptide.id];
       int peptide_set_index = peptide_map[peptide_sequence];
 
-      for (std::vector<String>::iterator it = peptide.protein_refs.begin(); it != peptide.protein_refs.end(); ++it)
+      for (const auto& it : peptide.protein_refs)
+      {
+        peptide_protein_map.push_back(std::make_pair(peptide_set_index, protein_map[it]));
```
cbielow: performance opt: `peptide_protein_map.emplace_back(peptide_set_index, protein_map[it]);`

src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
```diff
@@ -456,10 +489,18 @@ namespace OpenMS
...
+      }
+
+      String gene_name = "NA";
+      if (peptide.metaValueExists("GeneName"))
       {
-        peptide_protein_map.push_back(std::make_pair(peptide_set_index,protein_map[*it]));
+        gene_name = peptide.getMetaValue("GeneName");
       }
+      if (gene_map.find(gene_name) == gene_map.end()) gene_map[gene_name] = gene_map.size();
+      peptide_gene_map.push_back(std::make_pair(peptide_set_index, gene_map[gene_name]));
```
cbielow: you could replace with 
```
auto r = gene_map.emplace(gene_name, gene_map.size()); // only inserts if not present
peptide_gene_map.emplace_back(peptide_set_index, *r.first);
```
should safe two map lookups (one in each line).

src/openms/include/OpenMS/FORMAT/SqliteConnector.h
hroest:
```suggestion
      @brief Checks whether the given table exists
```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
hroest: the issue is that the OpenMS UniqueId is 64 bits unsigned while SQLite interprets all values as signed values, so when you retrieve the values in Python they are negative in half the cases. Its ugly but it means the values are not comparable between C++ and Python, this should solve the problem.

src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
hroest:
```suggestion
        peptide_protein_map.emplace_back(peptide_set_index, protein_map[it]);
```

src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
```diff
@@ -456,10 +489,18 @@ namespace OpenMS
...
+        peptide_protein_map.push_back(std::make_pair(peptide_set_index, protein_map[it]));
+      }
+
+      String gene_name = "NA";
+      if (peptide.metaValueExists("GeneName"))
       {
-        peptide_protein_map.push_back(std::make_pair(peptide_set_index,protein_map[*it]));
+        gene_name = peptide.getMetaValue("GeneName");
       }
+      if (gene_map.find(gene_name) == gene_map.end()) gene_map[gene_name] = gene_map.size();
```
hroest:
```suggestion
     auto r = gene_map.emplace(gene_name, gene_map.size()); // only inserts if not present`
```

src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
hroest:
```suggestion
      peptide_gene_map.emplace_back(peptide_set_index, *r.first);
```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
cbielow: ok, I see.
If you do not care about the actual value, then right-shifting by 1 would do the same thing. Might be easier to reason about.

Issue #980, opened 2014-09-09
C++11 support for OpenMS


hroest:
Currently the main issue with C++11 support for OpenMS is our Python bindings. Specifically, the Python 2.7 binaries for Windows are only built using MSVS 2008 which does not support C++11 [1,2]. RHEL 6.0 is another legacy system that has gcc 4.4 which has partial support for C++11 [5,6].

Python 3.3 and 3.4 seem to use MSVS 2010 [3] which supports a subset of C++11 [4].
Python 3.5 uses MSVS 2015 [3]. Probably Python 3.6 as well, according to [3].

The main obstacle here is Python 2.x on Windows, however people with this configuration will probably be our primary target group for intermediate programmers that want to try out pyOpenMS. We have made progress to support Python 3 #744 and in principle, the technical path to dropping Python 2.x on Windows would be open and replace it with Python 3 support. Other systems such as RHEL also only have partial support for C++11 (gcc 4.4).

Ref:
1. https://mail.python.org/pipermail/python-dev/2012-August/121410.html 
2. http://stackoverflow.com/questions/2676763/what-version-of-visual-studio-is-python-on-my-computer-compiled-with
3. https://docs.python.org/devguide/setup.html#windows
4. http://msdn.microsoft.com/en-us/library/hh567368.aspx
5. https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/6.0_Release_Notes/compiler.html#idp15008256
6. https://gcc.gnu.org/projects/cxx0x.html

hroest:
If we move to C++11 I suggest that the maximal amount of features that we should use is the intersection of MSVS 2010 and gcc 4.4.7 as these are the legacy system that are likely important to continue supporting. gcc because of CentOS systems that at least here in Zurich have a large install base and MSVS2010 because that is the compiler for which Python 3.3 and 3.4 are built with by default on Windows. For the features see this list

- https://gcc.gnu.org/gcc-4.4/cxx0x_status.html
- http://blogs.msdn.com/b/vcblog/archive/2010/04/06/c-0x-core-language-features-in-vc10-the-table.aspx

in short: rvalue yes (partial), auto yes, lambda no, atomic no, threads mostly no, std::move yes (partial)

hendrikweisser:
@hroest:
You mentioned #744 as the main blocker in your first post. You've closed that issue in the meantime - so I suppose Python 3 support in pyOpenMS is in place now?

[...]

PR #3978, opened 2019-02-22
[REFACTOR] Replace boost::regex with std::regex in some parsers

pasqoo:
This PR replaces `boost::regex` usage with `std::regex`.
Also `std::regex_search` with `std::regex_match`, where possible.

The affected classes are:
AbsoluteQuantitationMethodFile
ChromeleonFile
MRMFeatureQCFile
MSPGenericFile


hroest:
I dont think we can currently support this as we still need to support gcc 4.8 which does not have std::regex implemented. See https://github.com/OpenMS/OpenMS/issues/980#issuecomment-198783912 -- one of the main reasons we need to support gcc 4.8 is the manylinux container

cbielow:
sounds like a good PR for the future though.
Has anyone done some benchmarks on boost vs std implementation? I remember performance problems in EnzymaticDigestion (or similar) due to regex, so one needs to be a bit careful...

PR #3980, opened 2019-02-23
[FIX] Bug in MRMFeatureSelector

dmccloskey:
Added check to MRMFeatureSelector::computeScore_ to account for NaN and Inf.


cbielow:
looks good to me.
I think the appleclang error http://cdash.openms.de/viewBuildError.php?buildid=300509 is unrelated.

PR #3981, opened 2019-02-23
[FIX] bugfix

hroest:
fix segfault if AASequence is empty


cbielow:
nice. thanks for taking care of this 👍

PR #3992, opened 2019-03-06
MultiplexResolver now reports "not quantifiable" peptide features 

lars20070:
The combination of `FeatureFinderMultiplex + MultiplexResolver` allows the detection of true *knock-out* quantifications [1]. Take for example a simple light-heavy SILAC experiment. Say `FeatureFinderMultiplex` detects only a single peptide feature in the heavy channel. `MultiplexResolver` then completes the multiplet by adding a light *dummy peptide feature* of intensity `zero`. But this is not always correct.

There might be **two** distinct reasons, why a peptide feature singlet (and not a doublet) was detected.
(1) In the region of the expected light peptide feature is no peptide MS1 signal. In this case, a dummy feature of intensity `zero` is reported.
(2) Some isotopes of the light peptide feature overlap exactly with another unrelated peptide feature. `FeatureFinderMultiplex` blacklists this region and can consequently not detect the light peptide feature. In this case, a dummy feature of intensity `nan` is reported. The feature is *not quantifiable*. In complex, not fractionated samples, this case is increasingly likely [2].

This PR introduces the new (optional) `FeatureFinderMultiplex` output `-out_blacklist`. If this output is passed to the (optional) `MultiplexResolver` input `-in_blacklist`, then dummy features of intensity `nan` are reported. See (2) above.

[1] An absent peptide is reported as true `zero`, not as the sum of the *background noise* in this region. Say a gene/protein is knocked out in one of the two samples.

[2] A large number of `nan` are a sign that the sample is too complex. The sample should be further pre-fractionated or a longer LC gradient be used. Reliable peptide quantification is no longer possible.


src/topp/FeatureFinderMultiplex.cpp
```diff
@@ -195,6 +199,18 @@ class TOPPFeatureFinderMultiplex :
     file.store(filename, map);
   }
   
+  /**
+   * @brief Write blacklist to mzML file.
+   *
+   * @param filename    name of mzML file
+   * @param blacklist    blaclist for output
+   */
+  void writeBlacklist_(const String& filename, MSExperiment& blacklist) const
```
hroest:
```suggestion
  void writeBlacklist_(const String& filename, const MSExperiment& blacklist) const
```

src/topp/FeatureFinderMultiplex.cpp
```diff
@@ -195,6 +199,18 @@ class TOPPFeatureFinderMultiplex :
     file.store(filename, map);
   }
   
+  /**
+   * @brief Write blacklist to mzML file.
+   *
+   * @param filename    name of mzML file
+   * @param blacklist    blaclist for output
```
hroest:
```suggestion
   * @param blacklist    blacklist for output
```

timosachsenberg:
@lars20070 can you clarify what happens to the unrelated feature that was successfully quantified before it led to blacklisting of the second feature?

lars20070:
The unrelated feature is *lucky* and detected first in `FeatureFinderMultiplex`. It simply passes through `MultiplexResolver` unchanged.

timosachsenberg:
the files added to this PR are too large for inclusion

timosachsenberg:
@lars20070 out of interest. how often does this happen in complex samples?

lars20070:
@timosachsenberg More than I had hoped/expected. I run tests on various samples and show you numbers at the retreat.

lars20070:
@timosachsenberg The `mzML` was the huge file which is now reduced. Hope the other file sizes are ok.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderMultiplexAlgorithm.cpp
```diff
@@ -1151,4 +1152,9 @@ namespace OpenMS
   {
     return(consensus_map_);
   }
+  
+  MSExperiment& FeatureFinderMultiplexAlgorithm::getBlacklist()
+  {
+    return(exp_blacklist_);
```
timosachsenberg: you might want to remove these extra braces in return statements:
return exp_blacklist_; is less redundant

hroest:
> @timosachsenberg The `mzML` was the huge file which is now reduced. Hope the other file sizes are ok.

@timosachsenberg to make sure this doesnt appear in the git history, you will need to squash the commits (instead of merging)

timosachsenberg:
rebuild jenkins

timosachsenberg:
Ok I will merge. But imho you should add some extra documentation that this is mainly kockout related

lars20070:
@timosachsenberg In this example dataset, 5% of all reported features are not quantifiable, see [table 3](https://github.com/ProteomicsScripts/ProteomicsScripts/blob/master/mzTab2report/misc/example_7.pdf). This explain (nearly) all outliers in the peptide fold changes for specific proteins, see before/after plots below.

[plot_ProteinsManyPeptides_1_BEFORE.pdf](https://github.com/OpenMS/OpenMS/files/2961604/plot_ProteinsManyPeptides_1_BEFORE.pdf)
[plot_ProteinsManyPeptides_1_AFTER.pdf](https://github.com/OpenMS/OpenMS/files/2961605/plot_ProteinsManyPeptides_1_AFTER.pdf)

PR #3993, opened 2019-03-07
Make ResidueDB and ModificationsDB thread-safe

timosachsenberg:
Currently only implements on lock for reading and writing (=one common critical section).
Main work was in removing exceptions in critical section (or moving them to the caller).
Should be easy to promote to a readers-writer-lock.
I did not do that as the XL search engine (which I consider one of the heaviest use cases) did not show a significant speed drop (~10% on 20 threads).
All tests are passing.

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -76,11 +76,7 @@ public:
     /// Returns a pointer to the modifications DB (singleton)
     inline static ModificationsDB* getInstance(OpenMS::String unimod_file = "CHEMISTRY/unimod.xml", OpenMS::String psimod_file = "CHEMISTRY/PSI-MOD.obo", OpenMS::String xlmod_file = "CHEMISTRY/XLMOD.obo")
     {
-      static ModificationsDB* db_ = nullptr;
-      if (db_ == nullptr)
-      {
-        db_ = new ModificationsDB(unimod_file, psimod_file, xlmod_file);
-      }
+      static ModificationsDB* db_ = new ModificationsDB(unimod_file, psimod_file, xlmod_file);
       return db_;
     }
 
@@ -94,10 +90,8 @@ public:
 
     /**
        @brief Returns the modification with the given index
-
-       @throw Exception::IndexOverflow if the index is too large
     */
-    const ResidueModification& getModification(Size index) const;
+    const ResidueModification* getModification(Size index) const;
 
     /**
        @brief Collects all modifications which have the given name as synonym
@@ -119,10 +113,9 @@ public:
 
        If more than one matching modification is found, the first one is returned with a warning.
 
-       @throw Exception::ElementNotFound if no modification named @p mod_name exists (via searchModifications())
-       @throw Exception::InvalidValue if no matching modification exists
+       Returns nullptr if no modification named @p mod_name exists (via searchModifications()) or if no matching modification exists
     */
-    const ResidueModification& getModification(const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
+    const ResidueModification* getModification(const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
 
     /// Returns true if the modification exists
     bool has(String modification) const;
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -99,7 +99,7 @@ public:
 
        If @p residue is set, only modifications with matching residue of origin are considered.
        If @p term_spec is set, only modifications with matching term specificity are considered.
-       The resulting set of modifications may be empty if no modification exists that fulfills the criteria.
+        The resulting set of modifications will be empty if no modification exists that fulfills the criteria.
     */
     void searchModifications(std::set<const ResidueModification*>& mods, const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
 
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -89,7 +89,8 @@ public:
     Size getNumberOfModifications() const;
 
     /**
-       @brief Returns the modification with the given index
+       @brief Returns the modification with the given index.
+       note: out-of-bounds check is only performed in debug mode.
     */
     const ResidueModification* getModification(Size index) const;
 
@@ -99,7 +100,6 @@ public:
        If @p residue is set, only modifications with matching residue of origin are considered.
        If @p term_spec is set, only modifications with matching term specificity are considered.
        The resulting set of modifications may be empty if no modification exists that fulfills the criteria.
-
     */
     void searchModifications(std::set<const ResidueModification*>& mods, const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
 
@@ -111,7 +111,8 @@ public:
 
        If more than one matching modification is found, the first one is returned with a warning.
 
-       Returns nullptr if no modification named @p mod_name exists (via searchModifications()) or if no matching modification exists
+       @throw Exception::ElementNotFound if no modification named @p mod_name exists (via searchModifications())
+       @throw Exception::InvalidValue if no matching modification exists
     */
     const ResidueModification* getModification(const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
 
@@ -129,6 +130,8 @@ public:
 
        return numeric_limits<Size>::max() if not exactly one matching modification was found
        or no matching residue or modification were found
+
+       @throw Exception::ElementNotFound if not exactly one matching modification was found. 
     */
     Size findModificationIndex(const String& mod_name) const;
 
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -111,7 +111,9 @@ public:
 
        If more than one matching modification is found, the first one is returned with a warning.
 
-       @throw Exception::ElementNotFound if no modification named @p mod_name exists (via searchModifications())
+        @note Will never return a null pointer, instead will throw an exceptions.
+       
+        @throw Exception::ElementNotFound if no modification named @p mod_name exists (via searchModifications())
        @throw Exception::InvalidValue if no matching modification exists
     */
     const ResidueModification* getModification(const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -100,8 +100,6 @@ public:
        If @p term_spec is set, only modifications with matching term specificity are considered.
        The resulting set of modifications may be empty if no modification exists that fulfills the criteria.
 
-       @throw Exception::ElementNotFound if no modification named @p mod_name exists
-       @throw Exception::InvalidValue if no residue named @p residue exists
     */
     void searchModifications(std::set<const ResidueModification*>& mods, const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
 
@@ -122,15 +120,15 @@ public:
 
     /**
        @brief Add a new modification to ModificationsDB.
-       @throw Exception::InvalidValue if modification already exists (based on its fullID)
+       Will skip adding if modification already exists (based on its fullID)
     */
     void addModification(ResidueModification * new_mod);
 
     /**
        @brief Returns the index of the modification in the mods_ vector; a unique name must be given
 
-       @throw Exception::ElementNotFound if not exactly one matching modification was found
-       @throw Exception::InvalidValue if no matching residue or modification were found
+       return numeric_limits<Size>::max() if not exactly one matching modification was found
+       or no matching residue or modification were found
     */
     Size findModificationIndex(const String& mod_name) const;
 
@@ -142,25 +140,6 @@ public:
     */
     void searchModificationsByDiffMonoMass(std::vector<String>& mods, double mass, double max_error, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY);
 
-    /** @brief Returns the best matching modification for the given mass and residue
-
-        Query the modifications DB to get the best matching modification with
-        the given mass at the given residue (NULL pointer means no result,
-        maybe the maximal error tolerance needs to be increased). Possible
-        input for CAM modification would be a mass of 160 and a residue of
-        "C".
-
-        @note If there are multiple possible matches with equal masses, it
-        will choose the _first_ match which defaults to the first matching
-        UniMod entry.
-
-        @param residue The residue at which the modifications occurs
-        @param mass The monoisotopic mass of the residue including the mass of the modification
-        @param max_error The maximal mass error in the modification search
-
-        @return A pointer to the best matching modification (or NULL if none was found)
-    */
-    const ResidueModification* getBestModificationByMonoMass(double mass, double max_error, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY);
 
     /** @brief Returns the best matching modification for the given delta mass and residue
 
--- a/src/openms/source/CHEMISTRY/AASequence.cpp
+++ b/src/openms/source/CHEMISTRY/AASequence.cpp
@@ -912,7 +912,7 @@ namespace OpenMS
       if (m == nullptr)
       {
         throw Exception::InvalidValue(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, str,
-          "Cannot convert string to peptide modification. No N-terminal modification matches to term specificty and origin.");
+          "Cannot convert string to peptide modification. No N-terminal modification matches to term specificity and origin.");
       }
       aas.n_term_mod_ = m;
       return mod_end;
--- a/src/openms/source/CHEMISTRY/AASequence.cpp
+++ b/src/openms/source/CHEMISTRY/AASequence.cpp
@@ -908,12 +908,18 @@ namespace OpenMS
       ++next_aa;
       if (*next_aa == '.') ++next_aa;
 
-      const ResidueModification* m = mod_db->getModification(mod, String(*next_aa), ResidueModification::N_TERM);
-      if (m == nullptr)
+      const ResidueModification* m;
+      try
       {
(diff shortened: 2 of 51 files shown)


timosachsenberg:
http://cdash.openms.de/testDetails.php?test=55088436&build=302015
http://cdash.openms.de/testDetails.php?test=55088830&build=302015

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -119,10 +113,9 @@ namespace OpenMS
 
        If more than one matching modification is found, the first one is returned with a warning.
 
-       @throw Exception::ElementNotFound if no modification named @p mod_name exists (via searchModifications())
-       @throw Exception::InvalidValue if no matching modification exists
+       Returns nullptr if no modification named @p mod_name exists (via searchModifications()) or if no matching modification exists
     */
-    const ResidueModification& getModification(const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
+    const ResidueModification* getModification(const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
```
hroest: @timosachsenberg hmm - so doesnt searchModifications still throw, see here:

https://github.com/timosachsenberg/OpenMS/blob/e1d796d5763446630e3c9f8582144505c6684144/src/openms/source/CHEMISTRY/ModificationsDB.cpp#L119-L123

timosachsenberg: Yes. I will add to this PR. So far only ResidueDB is kind of fixed. If it calls moddb there are still problems

timosachsenberg: In this case I can simply print a warning return without throwing.
Need to make sure the calling code does not use try / catch for any semantics though

timosachsenberg:
Need to fix pyOpenMS though.

hroest:
ok, so this is another stab at #905 which was previously attempted in #3439, maybe we should compare what we was done there before

hroest:
ok, so this is another stab at #905 which was previously attempted in #3439, maybe we should compare what we was done there before

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -908,44 +908,55 @@ namespace OpenMS
...
+    else // no internal mod for this residue
     {
       // TODO: get rid of this code path, its deprecated and is only a hack for
       // C-terminal modifications that don't use the dot notation
       if (std::distance(mod_end, str.end()) == 1) // potentially a C-terminal mod without explicitly declaring it using dot notation?
       {
         // old ambiguous notation: Modification might be at last amino acid or at C-terminus
-        try
+        const ResidueModification* term_mod = mod_db->getModification(mod, res, ResidueModification::C_TERM);
+        aas.c_term_mod_ = term_mod;
```
hroest: @timosachsenberg this is not quite what was there before (before c_term_mod was only set when no exception occured). Not sure whether that makes a difference, but maybe move this line to `if (term_mod != nullptr)`

src/openms/source/CHEMISTRY/ModificationsDB.cpp
```diff
@@ -574,43 +572,45 @@ namespace OpenMS
     // now use the term and all synonyms to build the database
     for (multimap<String, ResidueModification>::const_iterator it = all_mods.begin(); it != all_mods.end(); ++it)
     {
-
-      // check whether a unimod definition already exists, then simply add synonyms to it
-      if (it->second.getUniModRecordId() > 0)
+      #pragma omp critical(OpenMS_ModificationsDB)
```
hroest: this may actually slow down execution if you acquire / release the lock in each loop - why not just put the lock on the top-level?

timosachsenberg: Good point. Will do

src/openms/source/CHEMISTRY/AASequence.cpp
timosachsenberg: it sets c_term_mod_ to nullptr in that case which should be the same as before. but yes it might be easier to read if I put it in an extra if block

timosachsenberg:
> ok, so this is another stab at #905 which was previously attempted in #3439, maybe we should compare what we was done there before

I think it is an attempt to clean up some long-standing issues with exceptions and reduce dependencies between the singletons.
Introducing the critical section is born out of a practical need from my side and for my application, it seems to have no speed impact using one single "mutex" (via critial sections). Eugen is also testing this PR if it impacts speed of his protein-protein XL search engine. I like the readers-writer lock approach a lot. My current concerns are:
- mixing different parallelization techniques (boost mutex + openmp) in OpenMS
- more complex locking schemes might come with a performance impact if most read access in OpenMS is e.g. done via getResidue using a single letter code. There might be ways around in the more complex reader-writer implementation but I would prefer to start with the simple solution, measure and then move the more advanced ones you proposed. Also, I don't consider myself experienced enough on that topic so I am afraid of the more complex one ;) .

cbielow:
I did not look through the whole PR to be honest but from what I understand, we're now locking ResidueDB on every access, right? This is not neccessary IMHO. The better approach is to have a read-only datastructure which corresponds 1:1 with a file that gets loaded once on startup (global static, done) and a protected ModifyableResidueDB which is fully protected and comes with overhead when accessing. This might simply be a second datastructure within ResidueDB which gets queried when the RO part does not deliver a hit.
Much cleaner and also faster. Maybe we can hack that during retreat. Should take a day at most.

cbielow:
This uses the assumption that most queries can be answered from the RO (read-only) part of the DB, which I think is reasonable for most applications.

timosachsenberg:
@cbielow yes this might be even a better solution and now possible as dependencies are now not circular anymore. Most access is done via AASequence by read-accessing a residue by single letter code so this should be fine. 

I took a fresh look at the code and wonder if there is an even easier fix:
Basically, we just need to get rid of the locking at: https://github.com/timosachsenberg/OpenMS/blob/fix/ResidueDB/src/openms/source/CHEMISTRY/ResidueDB.cpp#L77
Which is currently not possible because of:
https://github.com/timosachsenberg/OpenMS/blob/fix/ResidueDB/src/openms/source/CHEMISTRY/ResidueDB.cpp#L163
and calling https://github.com/timosachsenberg/OpenMS/blob/fix/ResidueDB/src/openms/source/CHEMISTRY/ResidueDB.cpp#L446
after adding a residue.
I think the first one (line 163) is a bug and not required as buildresiduenames clears this array anyway (I will try to remove it). The second call in buildresiduenames does not seem necessary to me and we could do this on construction and if setResidues is called. That way we could remove the critical section from  const Residue* ResidueDB::getResidue(const unsigned char& one_letter_code) const

I think this could also solve all practical relevant problems (except adding billions of modifications in parallel - but well).

If this is fine with you I will go forward with this.

timosachsenberg:
rebuild jenkins

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -908,44 +908,59 @@ namespace OpenMS
       ++next_aa;
       if (*next_aa == '.') ++next_aa;
 
-      aas.n_term_mod_ = &(mod_db->getModification(mod, String(*next_aa), ResidueModification::N_TERM));
+      const ResidueModification* m = mod_db->getModification(mod, String(*next_aa), ResidueModification::N_TERM);
+      if (m == nullptr)
+      {
+        throw Exception::InvalidValue(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, str,
+          "Cannot convert string to peptide modification. No N-terminal modification matches to term specificty and origin.");
```
hroest:
```suggestion
          "Cannot convert string to peptide modification. No N-terminal modification matches to term specificity and origin.");
```

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -908,44 +908,59 @@ namespace OpenMS
...
 
     const String& res = aas.peptide_.back()->getOneLetterCode();
     if (specificity == ResidueModification::C_TERM) 
     {
-      aas.c_term_mod_ = &(mod_db->getModification(mod, res, ResidueModification::C_TERM));
+      const ResidueModification* m = mod_db->getModification(mod, res, ResidueModification::C_TERM);
+      if (m == nullptr)
+      {
+        throw Exception::InvalidValue(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, str,
+          "Cannot convert string to peptide modification. No C-terminal modification matches to term specificty and origin.");
```
hroest:
```suggestion
          "Cannot convert string to peptide modification. No C-terminal modification matches to term specificity and origin.");
```

src/openms/source/CHEMISTRY/Residue.cpp
```diff
@@ -506,8 +506,8 @@ namespace OpenMS
   void Residue::setModification(const String& name)
   {
     ModificationsDB* mod_db = ModificationsDB::getInstance();
-    const ResidueModification& mod = mod_db->getModification(name, one_letter_code_, ResidueModification::ANYWHERE);
-    setModification_(mod);
+    const ResidueModification* mod = mod_db->getModification(name, one_letter_code_, ResidueModification::ANYWHERE);
+    setModification_(*mod);
```
hroest: what if the mod does not exist?

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -970,34 +970,34 @@ namespace OpenMS
                     {
                       if (spci->second.front().getAccession() == "MS:1001189")  // nterm
                       {
-                        ResidueModification m = ModificationsDB::getInstance()->getModification(mname, r, ResidueModification::N_TERM);
-                        mod = m.getFullId();
+                        const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname, r, ResidueModification::N_TERM);
+                        mod = m->getFullId();
```
hroest: didnt you just convert an exception to a segfault here? what if we cannot find the mod because its not in our unimod file? seems like a bad way to deal with an error

src/tests/class_tests/openms/source/ModificationsDB_test.cpp
```diff
@@ -233,16 +233,16 @@ END_SECTION
...
+  TEST_EQUAL(ptr->getModification("Phosphorylation", "S", ResidueModification::ANYWHERE)->getId(), "Phospho");
+  TEST_EQUAL(ptr->getModification("Phosphorylation", "S", ResidueModification::ANYWHERE)->getFullId(), "Phospho (S)");
 
   // terminal mod:
-  TEST_EQUAL(ptr->getModification("NIC", "", ResidueModification::N_TERM).getId(), "NIC");
-  TEST_EQUAL(ptr->getModification("NIC", "", ResidueModification::N_TERM).getFullId(), "NIC (N-term)");
-  TEST_EQUAL(ptr->getModification("Acetyl", "", ResidueModification::N_TERM).getFullId(), "Acetyl (N-term)");
+  TEST_EQUAL(ptr->getModification("NIC", "", ResidueModification::N_TERM)->getId(), "NIC");
+  TEST_EQUAL(ptr->getModification("NIC", "", ResidueModification::N_TERM)->getFullId(), "NIC (N-term)");
+  TEST_EQUAL(ptr->getModification("Acetyl", "", ResidueModification::N_TERM)->getFullId(), "Acetyl (N-term)");
```
hroest: can you test for a mod that does not exist and document behavior what happens then (e.g. return of a null ptr)

hroest:
Overall looks good, however I am wondering what happens in all these cases where now the exception is removed and what if somebody provides invalid input. It seems that before we had a nice exception that makes it clear what is wrong and then aborts the program. Now we would terminate with a segfault in many cases - potentially crashing the whole application (e.g. what if you open an invalid file in TOPPView, the segfault would crash all of TOPPView instead of a catch-able exception that you can catch and simply provide an error message that the file is invalid).

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
timosachsenberg: True. I think I will throw an exception here.

timosachsenberg:
Ok I see. That's a valid concern. I think the exception needs to be thrown in the file handler.

src/openms/source/CHEMISTRY/Residue.cpp
timosachsenberg: we could fatally exit the program with an error message 
or return an error code (leaving the mod untouched)

enetz:
I tested the runtime of this branch compared to the current develop branch with OpenPepXL, on 25 threads on a cluster. The tool uses ResidueDB and ModificationsDB moderately, outside the really computationally heavy parts and does not add new Residues or Modifications, only reads from both DBs. At least in this case, I saw no difference in runtime (and no new errors).

timosachsenberg:
If we assign a modification not present in ModificationsDB to a residue during parsing of AASequence strings this works. AASequence internally adds the modification to moddb (e.g., [+1234.234]) and then assigns it to the residue. 

Using directly residue.setModification("[1234.234]") results in a segfault as the mod is not registered.
Question is, if we should expose this String interface at the level of Residue or only allow residue.setModification(ResMod* r)?
Seems a bit cleaner to me (but maybe less convenient ?).

hroest:
I am wondering now whether your approach is really the best one and whether this introduces a whole host or problems and unhandled errors. I also wonder why you were getting segfaults. According to this: 

https://stackoverflow.com/questions/13663231/throwing-a-c-exception-inside-an-omp-critical-section

The OpenMP standard mandates, that if an exception is thrown within most OpenMP constructs (parallel, section, master, single, for, critical, task, etc.), execution must resume within the same construct and that the same thread must catch the exception

maybe this was the problem? Maybe the constructs need to look like this

```
bool exception_thrown = false;
#pragma omp critical
{
try
{
  // do stuff
}  catch 
{
  exception_thrown = true;
  // store message if its an OpenMS exception?
}
}
if (exception_thrown)
{
  throw something;
}
```

as for example done here:

https://github.com/OpenMS/OpenMS/blob/1d629f2d9df41cfa09306210a845f3c3e17dd483/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp#L148-L178

timosachsenberg:
The segmentation fault was because of dereferencing a nullptr.
The constructs in the XML handler need to look like you proposed. I just did not happen to add this to this PR.

hroest:
Well, I remember the timeline for this as follows: 

1. add critical sections
2. notice that there are uncaught exceptions / faults because of this
3. remove all exceptions and replace with returning null ptrs
4. now we have segfaults due to dereference of null ptrs and we dont know which code relied on catching these exceptions

so my suggestion is to go back to (1) and add try/catch blocks into the critical sections in ResidueDB / ModificationsDB to catch those exceptions and see whether (2) doesnt happen any more

src/openms/source/FORMAT/HANDLERS/TraMLHandler.cpp
```diff
@@ -680,8 +680,8 @@ namespace OpenMS
                 {
                   residue = it->sequence[mit->location];
                 }
-                const ResidueModification& rmod = mod_db->getModification("UniMod:" + String(mit->unimod_id), residue, term_spec);
-                String modname = rmod.getId();
+                const ResidueModification* rmod = mod_db->getModification("UniMod:" + String(mit->unimod_id), residue, term_spec);
+                String modname = rmod->getId();
```
hroest: same here, this can segfault now

timosachsenberg:
@hroest I added back the exceptions to most methods.
I did not add them back to the index accessors and searchModifications because I think that this method should just return an empty set if nothing matches the search criteria.
It was used only in a handful of places where I now throw an exception to keep the old behavior.

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
```diff
@@ -1002,8 +1007,8 @@ namespace OpenMS
                         mod = xl_db->getModification(mods[s], "", ResidueModification::C_TERM);
                       }
                     }
-                    acc = mod.getPSIMODAccession();
-                    name = mod.getId();
+                    acc = mod->getPSIMODAccession();
+                    name = mod->getId();
```
hroest: hmm for some reason gcc doesnt like this:



/.../OpenMS/src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp:1011:39: error: ‘mod’ may be used uninitialized in this function [-Werror=maybe-uninitialized]

```suggestion
                    if (mod != nullptr) name = mod->getId();
```

should help?

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
```diff
@@ -986,7 +991,7 @@ namespace OpenMS
                 {
                   if (mods[s].hasSubstring(jt->getMetaValue("xl_mod")))
                   {
-                    ResidueModification mod;
+                    const ResidueModification* mod;
```
hroest:
```suggestion
                    const ResidueModification* mod = nullptr;
```
does that work with `const` ?

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
```diff
@@ -1002,8 +1007,8 @@ namespace OpenMS
                         mod = xl_db->getModification(mods[s], "", ResidueModification::C_TERM);
                       }
                     }
-                    acc = mod.getPSIMODAccession();
-                    name = mod.getId();
+                    acc = mod->getPSIMODAccession();
```
hroest:
```suggestion
                    // mod should never be null, but gcc complains (-Werror=maybe-uninitialized)
                    if (mod != nullptr) acc = mod->getPSIMODAccession();
```

for consistency's sake at least check both dereferences of `mod`?

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -93,21 +89,17 @@ namespace OpenMS
...
     */
-    const ResidueModification& getModification(Size index) const;
+    const ResidueModification* getModification(Size index) const;
 
     /**
        @brief Collects all modifications which have the given name as synonym
 
        If @p residue is set, only modifications with matching residue of origin are considered.
        If @p term_spec is set, only modifications with matching term specificity are considered.
        The resulting set of modifications may be empty if no modification exists that fulfills the criteria.
```
hroest:
```suggestion
       The resulting set of modifications will be empty if no modification exists that fulfills the criteria.
```

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -122,22 +114,24 @@ namespace OpenMS
        @throw Exception::ElementNotFound if no modification named @p mod_name exists (via searchModifications())
```
hroest:
```suggestion
       @note Will never return a null pointer, instead will throw an exceptions.
       
       @throw Exception::ElementNotFound if no modification named @p mod_name exists (via searchModifications())
```

src/openms/source/FORMAT/HANDLERS/MzIdentMLHandler.cpp
timosachsenberg: I always need to look it up https://stackoverflow.com/questions/1143262/what-is-the-difference-between-const-int-const-int-const-and-int-const
works because this means: pointer to a const ResidueModification

hroest:
:+1: from me, looks good.

cbielow:
do the performance measurements (10% speed drop) still hold for this final version?
And is 10% the loss of the WHOLE program, i.e. residueDB itself would be much worse?

timosachsenberg:
I think speed issues are mainly relevant in the context of creating AASequence objects.
For creating AASequences without modifications, it is lock-free.
If you create a modified sequence, it requires a lock because modified residues are added on the first encounter. So if we create an algorithm where we add many (e.g. different mass tags) the result would be much worse (Hannes did some measurements on that - still contained in the tests). Then again, if I would write such an algorithm then I would not use the look-up structures of ResidueDB and ModDB anyways. My personal take on that is, that I would allocate more time if this becomes a problem.

PR #3994, opened 2019-03-08
[Update] MSstatsFile

oliveralka:
Merge MSstatsFile from the proteomic_lfq branch in preparation 
of adding MSstatsTMT support.

@jpfeuffer: Please let me know if further changes are required here: 
Removed following lines (220), since protein inference does not take place here and multiple
ProteinIdentifications may be possible:
```
  // check if we have one single protein identification object (containing the infered proteins and groups)
    if (consensus_map.getProteinIdentifications().size() != 1)
    {
      throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, 
        "Exactly one protein run expected containing the infered protein groups. Found: "
        + String(consensus_map.getProteinIdentifications().size()) + " protein runs"); 
    }
```


src/openms/source/FORMAT/MSstatsFile.cpp
```diff
@@ -214,13 +217,79 @@ void OpenMS::MSstatsFile::store(const OpenMS::String &filename, ConsensusMap &co
   }
   const String delim(",");
 
-  // Keeps track of unique peptides (Size of value set is 1)
+  // We quantify indistinguishable groups with one (corner case) or multiple proteins.
+  // If indistinguishable groups are not annotated (no inference or only trivial inference has been performed) we assume
+  // that all proteins can be indipendently quantified (each forming an indistinguishable group).
+  using IndProtGrp = ProteinIdentification::ProteinGroup;
+  using IndProtGrps = std::vector<IndProtGrp>;
+  IndProtGrps& ind_prots = consensus_map.getProteinIdentifications()[0].getIndistinguishableProteins();
```
hroest: what if `consensus_map.getProteinIdentifications().empty()` ?

oliveralka:
Talked to Julianus, we will work first on the integration of MSstatsTMT, then on adding the inference the right way.

PR #3996, opened 2019-03-11
[FIX] fix AASequence string conversion

hroest:
- fix bug introduced in #3969 which changes meaning of AASequence
  strings
- "PEPTID[160]E" has a different meaning from "PEPTID[160.0]E" with the latter meaning that the difference is *exactly* 160.0 while the former means a mod of 160 +/- 0.5


timosachsenberg:
good catch. maybe add small regression test?

hroest:
sorry, I think there was no bug (I found this in a different branch) in #3969 so no need for alarm. However, the current PR is still more clear about what is going on and does not rely on `String` doing any number conversion so I would like to still merge it.

cbielow:
but it uses String's `operator+` (which is a good thing).

PR #3997, opened 2019-03-11
feat: MRMFeatureFilter failure messages as StringList

pasqoo:
The code would output the failure messages as a contatenation of all messages, delimited by a character.
In this PR I am replacing that logic with `StringList`, so that the end-user can decide how to separate/deal with the messages.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
@@ -53,7 +53,7 @@ namespace OpenMS
   /**
 
     @brief The MRMFeatureFilter either flags components and/or transitions that do not pass the QC criteria or filters out
-      components and/or transitions that do not pass the QC criteria. 
+      components and/or transitions that do not pass the QC criteria.
 
     @htmlinclude OpenMS_MRMFeatureFilter.parameters
 
@@ -89,10 +89,10 @@ public:
       @param filter_criteria MRMFeatureQC class defining QC parameters
       @param transitions transitions from a TargetedExperiment
 
-    */    
+    */
     void FilterFeatureMap(FeatureMap& features, const MRMFeatureQC& filter_criteria,
       const TargetedExperiment & transitions);
-    
+
     /**
       @brief Converts a FeatureMap to a qcMLFile::Attachment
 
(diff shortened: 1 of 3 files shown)


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -160,19 +160,18 @@ namespace OpenMS
...
-      @param str_vec vector of Strings
-      @param delim token to separate Strings in the list
+      @param[in] messages A StringList containing the failure messages
 
-      @return A concatenated string.
-    */ 
-    String uniqueJoin(std::vector<String>& str_vec, String& delim);
+      @return A copy of the input, without duplicates
+    */
+    StringList getUniqueSorted(const StringList& messages) const;
```
cbielow: if you want to optimize this a bit, use pass by value
`StringList getUniqueSorted(StringList messages) const;` and work on `messages` directly. This allows the compiler to use copy-elision if r-values are passed in.
Even more fancy is to actually pass in r-values at the call site using std::move (if that is safe -- i did not check).

Since this is a pretty generic function, you might want to expose it as one and put it in a more prominent place (e.g. StringUtils). But I have no strong feelings about that.

pasqoo: Thanks for the hint, I did not think about possible rvalues.
The fact that the function is currently called without rvalues would give us 0 advantages, correct?
I cannot pass the argument with `std::move()` because the original vector is later used for its size. I would like to rewrite a good portion of this class, in future. I don't like the way it looks.

If people would like to have the utility in `StringUtils`, then I will write the "optimized" version in it. Otherwise feel free to merge as it is.

pasqoo: I actually had changed the code by using a `std::map` (because it sorts and keeps element unique), but then I decided to keep the changes at a minimum, because I'd like to keep the rewrite for the future.
The "cannot std::move" is more of a bad side-effect of the current state of the class, than a good reason.

pasqoo:
tests seem to have failed on `git clone`

jpfeuffer:
True, but it was probably just a slow connection at that time. Since the rest passes (including linux on travis) it should be okay.

PR #3998, opened 2019-03-12
[TEST] create multi-threaded test

hroest:
As reported by  @dmccloskey I have created a true test for multi-threaded logging and indeed: it fails miserably and every single time on my machine


hroest:
some more info

```
(gdb) info threads
  Id   Target Id         Frame 
  1    Thread 0x7ffff7e557c0 (LWP 11464) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:265
  2    Thread 0x7fffe8c9c700 (LWP 11477) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:268
* 3    Thread 0x7fffe849b700 (LWP 11478) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:268
  4    Thread 0x7fffe7c9a700 (LWP 11479) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:265
  5    Thread 0x7fffe7499700 (LWP 11481) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:265
  6    Thread 0x7fffe6c98700 (LWP 11482) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:265
  7    Thread 0x7fffe6497700 (LWP 11483) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:268
  8    Thread 0x7fffe5c96700 (LWP 11484) "LogStream_test" __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:265
(gdb) thread
[Current thread is 3 (Thread 0x7fffe849b700 (LWP 11478))]
(gdb) bt
#0  __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:268
#1  0x00007ffff54cfe5b in std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007ffff54c0ec6 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff54c1237 in std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x0000000000416063 in main._omp_fn.0(void) () at /home/hr/openmsall/source/OpenMS_experimental/src/tests/class_tests/openms/source/LogStream_test.cpp:103
#5  0x00007ffff519f43e in ?? () from /usr/lib/x86_64-linux-gnu/libgomp.so.1
#6  0x00007ffff26366ba in start_thread (arg=0x7fffe849b700) at pthread_create.c:333
#7  0x00007ffff4cb741d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
(gdb) 
```

dmccloskey:
@hroest I have not tried your suggestion yet about setting `nullptrs` but do you still think that could work in the short term?

hroest:
it seems that the main problem here is that we are using `std::streambuf` which is *not threadsafe* when shared across threads [1]. Basically the segfault occurs before it even reaches any of our code and I think the only way to really fix this is to use a thread-safe logger and until then change the macros

https://github.com/OpenMS/OpenMS/blob/88c5833544b45c3bc137220d289329f8134e2d97/src/openms/include/OpenMS/CONCEPT/LogStream.h#L456-L457

to something like 

```
#define STRINGIFY(a) #a

#define OPENMS_THREAD_CRITICAL(name) \
  _Pragma( STRINGIFY( omp critical (name) ) )

#ifdef _OPENMP
#define LOG_INFO \
OPENMS_THREAD_CRITICAL \
Log_info 
#else
#define LOG_INFO \ 
Log_info 
#endif




```

1. https://stackoverflow.com/questions/9963413/istream-and-ostream-with-shared-streambuf-mutually-thread-safe-for-duplex-i-o

src/tests/class_tests/openms/source/LogStream_test.cpp
```diff
@@ -72,6 +72,27 @@ class TestTarget
...
+//  Temporary measure to fix logging
+// -----------------------------------
+#define STRINGIFY(a) #a
+
+#define OPENMS_THREAD_CRITICAL(name) \
+    _Pragma( STRINGIFY( omp critical (name) ) )
+
+#define LOG_INFO \
+  OPENMS_THREAD_CRITICAL(oms_log) \
+  Log_info
```
hroest: @dmccloskey I think you would have to change LogStream.h to include the `OPENMS_THREAD_CRITICAL(oms_log)` before every definition, can you try and do that? That should fix the issue

hroest:
> @hroest I have not tried your suggestion yet about setting `nullptrs` in the logs but do you still think that could work in the short term?

No, now I dont think that is actually sufficient as the segfault can occur before the execution even reaches OpenMS and it occurs when the default `std::operator<< ` is called.

hroest:
It looks like we have tried to fix this issue with 0d882fef4b59a515ba0bb5c7a82781580 and the following code:

https://github.com/OpenMS/OpenMS/blob/88c5833544b45c3bc137220d289329f8134e2d97/src/openms/source/CONCEPT/LogStream.cpp#L554-L562

but that was not sufficient since this overload only seems to work for certain types and it seems that in many circumstances still the regular ` std::operator<<` of the base class will be called [1] (actually I am not even really sure when the above template is called at all since it is not in the header and does not seem to serve any purpose).

Overall, the issue is that its not sufficient to just lock when calling `sync` but we also have to lock much earlier as soon as the underlying STL constructs are involved which are not thread-safe. So it looks like we have 2 options:

1. replace all of our logging with a thread-safe logging mechanism
2. replace our macros with a `critical` pragma as laid out in this PR which fixes the problem in the short run (and hopefully performance penalties are minimal unless logging is a major part of your application)

1. call:
#3  0x00007ffff54c1237 in std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6

0d882fef4b59a515ba0bb5c7a82781580

timosachsenberg:
What about
LOG_WARN
LOG_ERROR
?

src/tests/class_tests/openms/source/LogStream_test.cpp
dmccloskey: I added the following to LogStream.h

```
#define STRINGIFY(a) #a

#define OPENMS_THREAD_CRITICAL(name) \
    _Pragma( STRINGIFY( omp critical (name) ) )

  /// Macro to be used if fatal error are reported (processing stops)
#define LOG_FATAL_ERROR \
  OPENMS_THREAD_CRITICAL(oms_log) \
  Log_fatal << __FILE__ << "(" << __LINE__ << "): "

  /// Macro to be used if non-fatal error are reported (processing continues)
#define LOG_ERROR \
  OPENMS_THREAD_CRITICAL(oms_log) \
  Log_error

  /// Macro if a warning, a piece of information which should be read by the user, should be logged
#define LOG_WARN \
  OPENMS_THREAD_CRITICAL(oms_log) \
  Log_warn

  /// Macro if a information, e.g. a status should be reported
#define LOG_INFO \
  OPENMS_THREAD_CRITICAL(oms_log) \
  Log_info

  /// Macro for general debugging information
#define LOG_DEBUG \
  OPENMS_THREAD_CRITICAL(oms_log) \
  Log_debug << __FILE__ << "(" << __LINE__ << "): "
```

However on windows the build chokes on `_Pragma`.  Where is `_Pragma` coming from?

hroest:
rebuild jenkins

src/tests/class_tests/openms/source/LogStream_test.cpp
hroest: maybe you need to use `__pragma` on Windows? https://docs.microsoft.com/en-us/cpp/preprocessor/pragma-directives-and-the-pragma-keyword?view=vs-2017

dmccloskey: Thanks, that resulted in a successful build.  However, it does not look to fix the segmentation fault problem.  It still faults at line 89 `delete[] pbuf_;` in LogStream.cpp.

timosachsenberg:
rebuild jenkins

hroest:
rebuild jenkins

dmccloskey:
@hroest and @timosachsenberg Perhaps it might be better just to implement a thread safe logging library such as https://github.com/SergiusTheBest/plog.  @pasqoo has been working with this library and it seems to work quite well.

src/openms/source/ANALYSIS/DECHARGING/FeatureDeconvolution.cpp
```diff
@@ -713,7 +713,7 @@ namespace OpenMS
       }
 
     }
-    LOG_INFO << "Agreeing charges: " << agreeing_fcharge << "/" << (aedges * 2) << std::endl;
+    // LOG_INFO << "Agreeing charges: " << agreeing_fcharge << "/" << (aedges * 2) << std::endl;
```
hroest: check if that helps to fix the error?

hroest:
all tests pass except gcc 4.8, see 

- http://cdash.openms.de/index.php?project=OpenMS&date=2019-03-26
- http://cdash.openms.de/viewBuildError.php?buildid=304790

hroest:
rebuild jenkins

hroest:
rebuild jenkins

hroest:
> @hroest and @timosachsenberg Perhaps it might be better just to implement a thread safe logging library such as https://github.com/SergiusTheBest/plog. @pasqoo has been working with this library and it seems to work quite well.

yes, this is something we should probably consider in the long run. @timosachsenberg Maybe something you can discuss at the developer retreat? It is not going to be that easy since TOPP and TOPPAS all use the logging function as far as I am aware. Unfortunately I wont be able to make it.

dmccloskey:
@hroest and @timosachsenberg I should note that many of the macros (e.g., LOG_DEBUG) are named exactly the same in Plog, and utilize the same format as `std::cout` (e.g., `LOGI << "[" << (i + 1) << "/" << n << "] steps in processing sequence";` for LOG_INFO).  It might not be too painful of a transition.  

It also allows one to control whether to log to the console or text/.csv files.  It tooks us maybe 8 hours at most to implement.  Most of the time was just setting up CMAKE to include the necessary header files.

timosachsenberg:
sounds interesting. another alternative would be boost.Log that we are already pulling in via our dependencies

dmccloskey:
In the meantime, is there a way to turn off logging within OpenMS?

hroest:
> In the meantime, is there a way to turn off logging within OpenMS?

not really, it depends where the problem occurs. You can try to comment out the body of `int LogStreamBuf::sync()` and make it into a NOP. If that doesnt work, you could try and redefine all the macros to pipe to something that is a NOP, e.g. 

```
struct nopstream
{
  template < typename T>
  void operator<<(T)
  {
    // NOP
  }
}

#define LOG_INFO nopstream()
...
```

where you may have to either use templates or overloading to get the function arguments right

hroest:
> In the meantime, is there a way to turn off logging within OpenMS?

PS: alternatively, why not just replace in OpenMS/CONCEPT/LogStream.h

```
#define LOG_INFO std::cout
#define LOG_WARN std::cout
... etc ...

```

pasqoo:
Are there updates on the logging side?

pasqoo:
Only failing test: http://cdash.openms.de/testDetails.php?test=58070099&build=319936

hroest:
looks good to me, I would say lets merge

PR #4003, opened 2019-03-13
[ADD] mz, rt and nativeID to SiriusAdapter mzTab output

oliveralka:
This PR adds mass to charge, retention time and the nativeID to the mzTab of the SiriusAdapter.

This information can now be obtained from the corresponding spectrum.ms file. With the help of comments, which are retained when using SIRIUS.

The writer tools will probably be deprecated with the next SIRIUS release (mztab-m), but I wanted to introduce the changes for the sake of completeness and as a simplification for the users.


src/openms/include/OpenMS/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.h
```diff
@@ -83,6 +83,9 @@ namespace OpenMS
 
           struct CsiAdapterIdentification
```
lukaszimmermann: Perhaps introduce a base struct for the current structs `CSIAdapterIdentification` and `SiriusAdapterIdentification` and there declare the fields `mz, rt, native_id, scan_index, scan_number, feature_id`?  Then the derived classes would only add  the vector with the respective hits.

Perhaps also explain the why the id is called "native".

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -56,6 +56,44 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
 
   for (std::vector<String>::const_iterator it = sirius_output_paths.begin(); it != sirius_output_paths.end(); ++it)
```
lukaszimmermann: Use Range-based for loop.

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -56,6 +56,44 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
...
+      const String pmass_prefix = ">parentmass ";
+      String line;
+      while (getline(spectrum_ms_file, line))
+      {
+        if (line.hasPrefix(pmass_prefix))
+        {
+           String pmass = line.erase(line.find(pmass_prefix), pmass_prefix.size());
+           ext_mz = pmass.toDouble();
+        }
+        if (line.hasPrefix(rt_prefix))
```
lukaszimmermann: else if

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -56,6 +56,44 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
...
+        {
+           String rt = line.erase(line.find(rt_prefix), rt_prefix.size());
+           ext_rt = rt.toDouble();
+        }
+        if (line.compare(nid_prefix) == 1)
```
lukaszimmermann: else if

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -56,6 +56,44 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
...
+        {
+           String nid = line.erase(line.find(nid_prefix), nid_prefix.size());
+           ext_nid = nid;
+        }
+        if (line == ">ms1peaks")
```
lukaszimmermann: else if, also use hasprefix like above

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -56,6 +56,44 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
...
+    {
+      const String nid_prefix = "##nid ";
+      const String rt_prefix = "#rt ";
+      const String pmass_prefix = ">parentmass ";
+      String line;
+      while (getline(spectrum_ms_file, line))
+      {
+        if (line.hasPrefix(pmass_prefix))
+        {
+           String pmass = line.erase(line.find(pmass_prefix), pmass_prefix.size());
```
lukaszimmermann: perhaps just inline` toDouble() `call and use one variable (also below)

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -56,6 +56,44 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
...
+    const String sirius_spectrum_ms = *it + "/spectrum.ms";
+    ifstream spectrum_ms_file(sirius_spectrum_ms);
+    if (spectrum_ms_file)
+    {
+      const String nid_prefix = "##nid ";
+      const String rt_prefix = "#rt ";
+      const String pmass_prefix = ">parentmass ";
+      String line;
+      while (getline(spectrum_ms_file, line))
+      {
```
lukaszimmermann: perhaps trim the line here to avoid errors due to white space

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -167,6 +208,14 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
               uri.push_back(MzTabString(hit.links[k]));
             }  
 
+            smsr.exp_mass_to_charge = MzTabDouble(id.mz);
+
+            vector<MzTabDouble> v_rt;
+            MzTabDoubleList rt_list;
+            v_rt.push_back(MzTabDouble(id.rt));
```
lukaszimmermann: maybe use` emplace_back `and remove constructor.

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -183,10 +232,15 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
             featureId.first = "featureId";
             featureId.second = MzTabString(id.feature_id);
 
+            MzTabOptionalColumnEntry native_id;
```
lukaszimmermann: use `make_pair` or value init

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
```diff
@@ -59,6 +59,43 @@ void SiriusMzTabWriter::read(const std::vector<String> & sirius_output_paths,
 
   for (std::vector<String>::const_iterator it = sirius_output_paths.begin(); it != sirius_output_paths.end(); ++it)
```
lukaszimmermann: Same comments as in the CSIFingerID mztab writer

src/openms/include/OpenMS/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.h
oliveralka: native_id -> The nativeID is usually a unique reference for a spectrum in a file based on the vendor. e.g. "sample=1 period=1 cycle=2177 experiment=5"

oliveralka: You are right this could be done using a base struc/class, but I think this is not really necessary at this point.

oliveralka:
@lukaszimmermann Thank you for the review! I added the requested changes.

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
oliveralka: I guess it should not be a problem using the hasPrefix() method.

oliveralka:
Ready for merge @lukaszimmermann

PR #4005, opened 2019-03-14
FIX:  initializeXCorrMatrix

dmccloskey:
Removed default calculation of `initializeXCorrMatrix` when scores using are not computed.


timosachsenberg:
.vs/slnx.sqlite

hroest:
:+1: please remove the sqlite file

PR #4006, opened 2019-03-14
Poshul/hendrikweisser rna ms2

poshul:
Add a "max_size" option for nucleotide lengths. This is useful when searching data that has been nonspecifically digested.


timosachsenberg:
looks good

PR #4007, opened 2019-03-14
[DRAFT] SiriusAdapter: First adaption to new SIRIUS version

lukaszimmermann:
This is my first attempt to adapt `SiriusAdapter` to the new SIRIUS release scheduled to be released soon.

The update of `SiriusAdapter` is designed such that it can be further extended to incorporate other subtools, such as `ZODIAC` and `CANOPUS`.

There are some smaller things missing and of course this update of `SiriusAdapter` needs to be tested with the new version of `SIRIUS` once it is released.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
@@ -2,7 +2,7 @@
 //                   OpenMS -- Open-Source Mass Spectrometry
 // --------------------------------------------------------------------------
 // Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
-// ETH Zurich, and Freie Universitaet Berlin 2002-2018.
+// ETH Zurich, and Freie Universitaet Berlin 2002-2019.
 //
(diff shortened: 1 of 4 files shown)


src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -53,6 +56,38 @@ namespace OpenMS
...
+       *        parameter, but not in this DefaultParamHander, are ignored.
+       */
+      void updateExistingParameter(const Param &param);
+
+      /**
+       * Checks whether this DefaultParamHandler has a ParamEntry with the provided name.
+       * @param name The name of the ParamEntry that should be checked for its existence in this DefaultParamHandler
+       * @return Whether this DefaultParamHandler has an ParamEntry for the provided name.
+       */
+      bool hasFullNameParameter(const String &name) const;
```
hroest: can you check which of these should be exposed to Python? And then add these to the `.pxd` ?

oliveralka:
This PR will be integrated into https://github.com/OpenMS/OpenMS/pull/4725 and discussed there after the build system was upgraded to java 11 to be able to run SIRIUS.

PR #4008, opened 2019-03-15
[Add] Adducts and Annotations for compounds (tsv, traml, pqp)

oliveralka:
Add adducts and fragment annotation to tsv, traml and pqp.
Additionally, add min/max transition filter for MRMAssay (in preparation for refactored AssayGeneratorMetabo)

**Question:** There is still a problem with CompoundName. It is correctly annotated in the tsv file and the TraML file, but after the conversion to PQP or from PQP is performed the CompoundName (Dimethachlor) seems to be changed to the TransistionGroupID (0_Dimethachlor_0) - I am not sure where to look for the Problem. Any ideas? 
**Solved:** CompoundID was used instead of CompoundName in the PQP File line 612


hroest:
looks like you are missing the file `MRMAssay_detectingTransitionsCompound_input`, see:



checking void detectingTransitionsCompound(OpenMS::TargetedExperiment& exp, int min_transitions, int max_transitions) ... 
Error: Caught unexpected exception of type 'FileNotFound' thrown in line 111 of file '/home/travis/build/OpenMS/OpenMS/src/openms/source/FORMAT/XMLFile.cpp' in function 'void OpenMS::Internal::XMLFile::parse_(const OpenMS::String&, OpenMS::Internal::XMLHandler*)' - Message: the file '/home/travis/build/OpenMS/OpenMS/src/tests/class_tests/openms/data/MRMAssay_detectingTransitionsCompound_input.TraML' could not be found
: failed
Warning: no subtests performed in 'void detectingTransitionsCompound(OpenMS::TargetedExperiment& exp, int min_transitions, int max_transitions)' (line 564)!

oliveralka:
Yes, it seems there is something wrong with the MRMAssay_test - the file should be there "src/tests/class_tests/openms/data/MRMAssay_detectingTransistionCompound_input.TraML"
I will have a look.

oliveralka:
rebuild jenkins

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMAssay.h
```diff
@@ -134,7 +136,17 @@ namespace OpenMS
 
     */
     void detectingTransitions(OpenMS::TargetedExperiment& exp, int min_transitions, int max_transitions);
+    
+    /**
+      @brief Filters transistions by number
```
hroest:
```suggestion
      @brief Filters transitions by intensity, only keeping the top N transitions
      
      Filters transitions by intensity, only keeping the top N transitions and removes compounds if they have less than the minimal number of transitions.
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMAssay.h
```diff
@@ -134,7 +136,17 @@ namespace OpenMS
...
+
+      @param exp the input, unfiltered transitions
```
hroest:
```suggestion
      @param exp the transition list which will be filtered
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMAssay.h
```diff
@@ -134,7 +136,17 @@ namespace OpenMS
...
+      @param min_transitions the minimum number of transitions required per assay
+      @param max_transitions the maximum number of transitions required per assay
 
+    */
+    void detectingTransitionsCompound(OpenMS::TargetedExperiment& exp, int min_transitions, int max_transitions);
```
hroest: can you add this to pyOpenMS ?

src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
```diff
@@ -1145,6 +1148,7 @@ namespace OpenMS
     compound.molecular_formula = tr_it->SumFormula;
     compound.smiles_string = tr_it->SMILES;
     compound.setMetaValue("CompoundName", tr_it->CompoundName);
+    compound.setMetaValue("Adducts", tr_it->Adducts);
```
hroest:
```suggestion
    if (!tr_it->Adducts.empty()) compound.setMetaValue("Adducts", tr_it->Adducts);
```
will prevent all the empty adducts tags

oliveralka:
rebuild jenkins

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMAssay.h
oliveralka: Done!

PR #4010, opened 2019-03-21
Feature/flkd charge adduct merging

FabianAicheler:
Adds support to FLKD to treat charges/adducts different ways when considering what to group.

E.g., now it can finally link features with any charge to those without (setting "link:charge_merging" "Zero")


FabianAicheler:
I looked at the output for the 3*3 different combinations using the tutorial metabolomics dataset. Seemed to work correctly.

timosachsenberg:
Can you add a proper test?

FabianAicheler:
If forced to, but adding 9 regression tests doesn't make that much sense.

It doesn't really change the linking algorithm (which is already (successfully) tested by the default settings, which correspond to the settings before the pull request), just prevents some linking combinations.

timosachsenberg:
The code relies on specific meta values so we definitely need a test that ensures that e.g. renaming these meta values does not change the behavior of the linker. 
"Seemed to work correctly" is also not good enough ;) especially the adduct and charge handling should be tested
Please also add to the tool documentation.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
```diff
@@ -66,6 +66,10 @@ namespace OpenMS
     defaults_.setMinFloat("link:rt_tol", 0.0);
     defaults_.setValue("link:mz_tol", 10.0, "m/z tolerance (in ppm or Da)");
     defaults_.setMinFloat("link:mz_tol", 0.0);
+    defaults_.setValue("link:charge_merging","Zero","whether to disallow charge mismatches (None), allow to link charge zero (i.e., unknown charge) with every other charge (Zero), or disregard charges (All).");
+    defaults_.setValidStrings("link:charge_merging", ListUtils::create<String>("None,Zero,All"));
+    defaults_.setValue("link:adduct_merging","All","whether to only allow the same adduct for linking (None), allow linking adduct-free features with other features (Unknowns), or disregard adducts (All).");
```
cbielow: 'None' seems to be an unfortunate name for this (implies no merging), but means 'Identical'?

'All' could be 'Any'?

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
```diff
@@ -66,6 +66,10 @@ namespace OpenMS
...
+    defaults_.setValidStrings("link:adduct_merging", ListUtils::create<String>("None,Unknowns,All"));
```
cbielow: C++11 now allows

`defaults_.setValidStrings("link:adduct_merging", {"None", "Unknowns", "All"});`

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
```diff
@@ -393,18 +397,95 @@ namespace OpenMS
...
+          continue;
+        }
+      }
+      else if (merge_charge == "All")
+      {
+        //we allow to merge all, keep case for parameter check
+      }
+      else
+      {
+        throw Exception::InvalidParameter(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "WARNING! Invalid charge merging parameter! " + String(param_.getValue("link:charge_merging"))  + "\n");
```
cbielow: useless, since restrictions are enforced in TOPPBase anyway?!

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
```diff
@@ -393,18 +397,95 @@ namespace OpenMS
...
+      {
+        if (kd_data.charge(*it) != charge_i)
+        {
+          continue;
+        }
+      }
+      //what to consider for linking with existing features _that have charge_. This ensures that we won't collect different non-zero charges.
+      else if (merge_charge == "Zero")
+      {
+        if (kd_data.charge(*it) != charge_i && kd_data.charge(*it) != 0)
```
cbielow: Can you add brackets ` (a !=b) && (c != d)`

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
```diff
@@ -393,18 +397,95 @@ namespace OpenMS
...
-        points_for_map_index[kd_data.mapIndex(*it)].push_back(*it);
+        //subcase 1: one has adduct, other not
+        if (kd_data.feature(*it)->metaValueExists("dc_charge_adducts") != f_i->metaValueExists("dc_charge_adducts"))
+        {
+          continue;
+        }
+        //subcase 2: both have adduct, but is it the same?
+        if (kd_data.feature(*it)->metaValueExists("dc_charge_adducts"))
+        {
+          if (EmpiricalFormula(kd_data.feature(*it)->getMetaValue("dc_charge_adducts")) != EmpiricalFormula(f_i->getMetaValue("dc_charge_adducts")))
```
cbielow: Do you need to create an EF for this? String comparison is more fragile, but might be sufficient here?

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
```diff
@@ -393,18 +397,95 @@ namespace OpenMS
...
+          }
+        }
+      }
+      else if (merge_adduct == "All")
+      {
+        //we merge all, keep case for parameter check
+      }
+      else
+      {
+        throw Exception::InvalidParameter(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "WARNING! Invalid adduct merging parameter! " + String(param_.getValue("link:adduct_merging"))  + "\n");
```
cbielow: same as above

src/tests/topp/.nfs000000000019cb050001c481
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
cbielow: please remove stale NFS files....

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
FabianAicheler: Theoretically different gains/losses of charged/neutral adducts might result in the same end EF contained atoms. I don't know enough about the EF behavior, is the atom order designed to be deterministic?

FabianAicheler:
Made the test files smaller, and reduced to three files.

The motivation for the use two of the test inputs in duplicate each is to cover interesting cases for linking behavior (for methods using centroiding for the consensus feature/ aggregation by centroid comparison to new and majority votes for centroid updates):
-What happens if the centroid is uncharged/without adduct, can it (incorrectly) in an alternating manner pick up disagreeing charges/adducts (eg start with uncharged centroid from sample 3, link with sample 1, then 2, then 1, then 2)
-What happens if the majority is uncharged/without adduct (e.g., links 2 uncharged, then adds from one sample a charged, then from another sample with other charge because centroid still uncharged/without adduct?)

timosachsenberg:
Thanks!

PR #4012, opened 2019-03-22
[FIX] fix overflow on Windows

### LINKED ISSUE #4009: Indexed mzML overflow issue
On the latest pyOpenMS versions on Windows, the index seems to overflow when a file larger than 2GB is written, this happens on 64 bit builds. We have had a bug report when using pyOpenMS on Windows but not on Linux. This may be a general issue in the OpenMS library.

Reproduce this error (potentially):

```
import pyopenms
e = pyopenms.MSExperiment()
for i in range(int(1.5e5)):
  s  = pyopenms.MSSpectrum()
  s.set_peaks([[k for k in range(1000)], [k for k in range(1000)]])
  e.addSpectrum(s)

pyopenms.MzMLFile().store("/tmp/test.mzML", e)
```

On Linux:

```
$ tail /tmp/test.mzML 
		<offset idRef="spectrum=149995">2613743706</offset>
		<offset idRef="spectrum=149996">2613761133</offset>
		<offset idRef="spectrum=149997">2613778560</offset>
		<offset idRef="spectrum=149998">2613795987</offset>
		<offset idRef="spectrum=149999">2613813414</offset>
	</index>
</indexList>
<indexListOffset>2613830871</indexListOffset>
<fileChecksum>0</fileChecksum>
```

On Windows:

```
$ tail test.mzML
                <offset idRef="spectrum=149995">-1681223590</offset>
                <offset idRef="spectrum=149996">-1681206163</offset>
                <offset idRef="spectrum=149997">-1681188736</offset>
                <offset idRef="spectrum=149998">-1681171309</offset>
                <offset idRef="spectrum=149999">-1681153882</offset>
        </index>
</indexList>
<indexListOffset>-1681136425</indexListOffset>
<fileChecksum>0</fileChecksum>
```

hroest:
Ensure that we actually use 64 bit integers to store spectra / chromatogram offsets on all platforms

I have confirmed this on Windows that it fixes #4009 

There is currently no test for it as it would require to write a 2+ GB file to disk which we probably dont want to do routinely

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/HANDLERS/MzMLHandler.h
+++ b/src/openms/include/OpenMS/FORMAT/HANDLERS/MzMLHandler.h
@@ -487,8 +487,8 @@ protected:
        * <chromatogram> tag is stored and will then be stored at the end of the file.
        **/
       //@{
-      std::vector<std::pair<std::string, long> > spectra_offsets_; ///< Stores binary offsets for each <spectrum> tag
-      std::vector<std::pair<std::string, long> > chromatograms_offsets_; ///< Stores binary offsets for each <chromatogram> tag
+      std::vector<std::pair<std::string, Int64> > spectra_offsets_; ///< Stores binary offsets for each <spectrum> tag
+      std::vector<std::pair<std::string, Int64> > chromatograms_offsets_; ///< Stores binary offsets for each <chromatogram> tag
       //@}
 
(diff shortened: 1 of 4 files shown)


hroest:
should fix #4009 (not tested)

timosachsenberg:
Looks good!

PR #4014, opened 2019-03-22
[PYOPENMS] fix inheritance for MetaInfoInterface / CVTermList

hroest:
[TEST] fix test

- remove overloaded methods from parent class (we anyways wont use some of these)
- allow proper inheritance through Cython of all affected classes
- do this for MetaInfoInterface / CVTermList


hroest:
pyOpenMS only, passes travis

PR #4015, opened 2019-03-22
[FIX] forward declarations

hroest:


hroest:
rebuild jenkins

src/openms/include/OpenMS/METADATA/MetaInfoInterface.h
```diff
@@ -34,11 +34,16 @@
 
 #pragma once
 
-#include <OpenMS/METADATA/MetaInfo.h>
+#include <vector>
+
+#include <OpenMS/CONCEPT/Types.h>
+#include <OpenMS/METADATA/MetaInfoRegistry.h>
+#include <OpenMS/DATASTRUCTURES/DataValue.h>
```
hroest: MetaInfoInterface does not need to include `MetaValue` since it just holds a ptr (PIMPLE principle) and thus we can decouple these two

hroest:
all tests are green, lets merge

PR #4016, opened 2019-03-25
[FIX] make it work for lowercase sqmass files

hroest:


src/openms/include/OpenMS/APPLICATIONS/OpenSwathBase.h
```diff
@@ -259,7 +259,7 @@ class TOPPOpenSwathBase :
   {
     if (!out_chrom.empty())
     {
-      if (out_chrom.hasSuffix(".sqMass"))
+      if (out_chrom.lower().hasSuffix(".sqMass"))
```
hroest:
```suggestion
      if (out_chrom.toLower().hasSuffix(".sqmass"))
```

src/openms/include/OpenMS/APPLICATIONS/OpenSwathBase.h
```diff
@@ -259,7 +259,7 @@ class TOPPOpenSwathBase :
   {
     if (!out_chrom.empty())
     {
-      if (out_chrom.hasSuffix(".sqMass"))
+      if (out_chrom.toLower().hasSuffix(".sqmass"))
```
hroest:
```suggestion
      String tmp = out_chrom;
      if (tmp.toLower().hasSuffix(".sqmass"))
```

PR #4017, opened 2019-03-25
Feature/sql fix

hroest:
- fix for getting a subset of chromatograms from sqMass
- add tests for the case of multiple chromatograms


hroest:
looks good, Windows timed out

Issue #521, opened 2014-01-09
Upgrade contrib libsvm [521]


timosachsenberg:
Submitted by aiche on 2013-04-03 10:56:29

Currently we use a patched version of libsvm 2.91 in our contrib. We should upgrade to the latest version 3.17 and consider using the following libsvm fork that provides a CMake based build system.

https://github.com/aiche/libsvm-buildsystem

---

Commented by aiche on 2013-06-30 00:25:51:
upgraded/finalized in [11442], [11443], and [11445]

---

Commented by hroest on 2013-07-12 14:41:14:
It seems quite a few SVMWrapper tests fail:

http://cdash.openms.de/testSummary.php?project=3&name=SVMWrapper_test&date=2013-07-12

it seems only Linux is affected. With the old libSVM version I can pass all the tests, on two different Linux systems. I will now test the new contrib with the new libSVM.

---

Commented by hroest on 2013-07-13 13:38:49:
The  tests on Linux fail which seems to be a difference in where a label at 0.5 probability is assigned to. So while the following change [1] would fix it on Linux, it might break the tests on other platforms. However, it might also not be the best way to test a label that has exactly probability 0.5 whether it falls in group 1 or 2. On the other hand it seems that different platforms handle this case differently which is also not optimal.

Ref 1:

```
@@ -1050,8 +1050,9 @@ START_SECTION((void getSVCProbabilities(struct svm_problem *problem, std::vector
    TEST_EQUAL(predicted_labels.size() == probabilities.size(), true)
    for (Size i = 0; i < predicted_labels.size(); ++i)
    {
-       TEST_EQUAL((predicted_labels[i] < 0 && probabilities[i] <= 0.5)
-                           || (predicted_labels[i] > 0 && probabilities[i] > 0.5), true)
+       TEST_EQUAL((predicted_labels[i] < 0 && probabilities[i] < 0.5)
+                           || (predicted_labels[i] > 0 && probabilities[i] >= 0.5), true)
    }

 END_SECTION
```

---

Commented by hroest on 2013-07-15 15:45:14:
After some testing, on Linux it seems that the following versions of libsvm still work
- 3.0 Works (test passed)
- 3.12 Works (test passed)
- 3.13 Works (test passed)
- 3.14 Works (test passed)
- 3.15 segfault
- 3.16 segfault
- 3.17 Doesnt work (no segfault but tests not passed) 

old files were retrieved from:
[...]

PR #4020, opened 2019-03-26
NucleicAcidSearchEngine improvements: custom mods, higher speed

hendrikweisser:
There are two major improvements to NucleicAcidSearchEngine in this set of changes:
1. Introduction of "generic" and "ambiguous" modifications. One criticism of NASE was that we could search for e.g. "m1A" or "m6A" as variable mods, and they would appear as such in the results, but we can't actually distinguish between them (based on tandem MS data). To avoid misleading results, I've added "generic" versions of these (groups of) mods, e.g. "mA". The definitions are in a new file ("share/OpenMS/CHEMISTRY/Custom_RNA_modifications.tsv"), to keep them separate from the "official" mod definitions (".../Modomics.tsv"). Users can also add their own mods in this file.
"Ambiguous" mods are for cases where two mods are isobaric, but could in principle be distinguished based on a-B ions, because a methyl group is either attached to the base or to the ribose (and thus lost/not lost in the a-B ion). There are two modes for dealing with these, depending on parameter `modifications:resolve_ambiguities` (see documentation).
2. The other improvement is a massive increase in speed, mainly due to optimization of theoretical spectrum generation. In retrospect, the previous implementation was strikingly inefficient: When generating a spectrum for a nucleotide sequence, peak m/z values were recalculated from scratch based on chemical formulas for every subsequence, ion type and charge. Now I compute an uncharged spectrum for all subsequences and ion types in one go, then simply generate charged versions of that. In addition, calculations are performed with numbers rather than sum formulas early on.
For searches of bigger sequence databases with many mods (e.g. tRNA), this can save >90% of runtime! (I've just redone a search that previously took almost 11 days in less than 4 hours.)

I've also made some changes that unfortunately break backwards compatibility of .ini files. I've removed the support for fixed mods (not useful for RNA) and changed the notation for some multi-methylated mods: Since commas aren't allowed in string list parameter values, commas in mod codes were previously  replaced by underscores (e.g. "m6,6A" -> "m6_6A"), which I always found looked weird. Now the commas are simply omitted (e.g. "m66A"), which looks better and still works (no conflicts).

More comments:
- The new mod codes are currently just added to the list of available mods (parameter `modifications:variable`). I don't know if it would be cleaner to move them to a separate parameter, or if the "indistinguishable" versions of the mods (e.g. "m1A", "m6A") should be removed.
- The switch to calculating with numbers rather than sum formulas during spectrum generation leads to minor numeric differences, which can very slightly affect the search results. Some individual peaks may be matched/not matched, which affects some hyperscores and in turn most q-values. In my tests the new results were not worse and the time savings about 50%, so worth it.
- I haven't touched spectrum generation for peptide sequences, but I believe it could be similarly optimized with analogous effects on runtime. There are however some additional options (isotope clusters, losses) to consider for peptides.

(@poshul, I would have liked to request you as a reviewer as well, but you're not on the list. It seems you're not a member of the OpenMS GitHub project - you probably should be.)


timosachsenberg:
These are nice additions.
Can you split the NA part from the TheoreticalSpectrumGenerator? It's becoming a bit ugly with all that "only applies to ...." Doesn't need a common base class or anything...

hendrikweisser:
> Can you split the NA part from the TheoreticalSpectrumGenerator?

Okay, I've just done that.

hroest:
can you fix these tests: 

https://travis-ci.org/OpenMS/OpenMS/jobs/511595122
The following tests FAILED:
	886 - UTILSWRITEINI_NucleicAcidSearchEngine (Not Run)
	887 - UTILSWRITEINI_NucleicAcidSearchEngine_SectionName (Failed)
	994 - UTILSWRITECTD_NucleicAcidSearchEngine (Not Run)
	2303 - UTILS_NucleicAcidSearchEngine_1 (Not Run)
	2304 - UTILS_NucleicAcidSearchEngine_11_out (Failed)
	2305 - UTILS_NucleicAcidSearchEngine_12_out (Failed)

which are mainly due to this error: http://cdash.openms.de/viewBuildError.php?buildid=305073


/.../OpenMS/src/utils/NucleicAcidSearchEngine.cpp:297:50: error: no matching function for call to ‘regex_replace(OpenMS::String&, std::regex&, const char [4])’

which may not exist with gcc 4.8, can you check that?

src/openms/include/OpenMS/CHEMISTRY/Ribonucleotide.h
```diff
@@ -170,7 +171,13 @@ namespace OpenMS
     /// Return true if this is a modified ribonucleotide and false otherwise
     bool isModified() const;
 
+    /// Return whether this is an "ambiguous" modification (representing isobaric modifications on the base/ribose)
+    bool isAmbiguous() const;
```
hroest: can you add this to the pyopenMS .pxd file?

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
```diff
@@ -100,6 +101,14 @@ namespace OpenMS
     */
     ConstRibonucleotidePtr getRibonucleotidePrefix(const std::string& seq);
 
+    /**
+       @brief Get the alternatives for an ambiguous modification code
+
+       @throw Exception::ElementNotFound if nothing was found
+    */
+    std::pair<ConstRibonucleotidePtr, ConstRibonucleotidePtr> getRibonucleotideAlternatives(const std::string& code);
```
hroest: can you add this to the pyopenMS .pxd file?

src/openms/include/OpenMS/CHEMISTRY/TheoreticalSpectrumGeneratorNA.h
```diff
@@ -0,0 +1,138 @@
...
+  class NASequence;
+
+  /**
+      @brief Generates theoretical spectra for nucleic acid sequences
+
+      @htmlinclude OpenMS_TheoreticalSpectrumGeneratorNA.parameters
+
+      @ingroup Chemistry
+  */
+  class OPENMS_DLLAPI TheoreticalSpectrumGeneratorNA :
```
hroest: please create a new `TheoreticalSpectrumGeneratorNA.pxd` in this folder `src/pyOpenMS/pxds/`

, see also https://pyopenms.readthedocs.io/en/latest/wrap_classes.html

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
```diff
@@ -0,0 +1,554 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Hendrik Weisser $
+// $Authors: Andreas Bertsch, Hendrik Weisser $
```
hroest:
```suggestion
// $Authors: Hendrik Weisser $
```

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
poshul: I can take a shot at this tomorrow if @hendrikweisser doesn't have any objections.

share/OpenMS/CHEMISTRY/Custom_RNA_modifications.tsv
```diff
@@ -0,0 +1,34 @@
+# This file contains OpenMS- and user-defined RNA modifications.
+# It is read in addition to the Modomics data file ("Modomics.tsv") and uses the same format (except for the optional "alternatives" column).
+# For terminal modifications, the "new_nomenclature" value has to end in "N" and contain either "33" (3'-terminal mod.) or "55" (5'-terminal mod.).
```
timosachsenberg: why 33 and 55? is this an official nomenclature?

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
```diff
@@ -0,0 +1,554 @@
...
+    return spectrum;
+  }
+
+
+  void TheoreticalSpectrumGeneratorNA::addChargedSpectrum_(
+    MSSpectrum& spectrum, const MSSpectrum& uncharged_spectrum, Int charge,
+    bool add_precursor) const
+  {
+    // based on very limited testing, it looks like this hydrogen mass gives
+    // better results than using "Constants::PROTON_MASS_U" (why?):
```
timosachsenberg: hmm this is indeed a bit strange and doesn't make sense to me.
I would guess that the reason is something like slight slightly miscalibrated instrument.

src/tests/class_tests/openms/source/TheoreticalSpectrumGeneratorNA_test.cpp
```diff
@@ -0,0 +1,256 @@
+
```
timosachsenberg: remove extra line

share/OpenMS/CHEMISTRY/Custom_RNA_modifications.tsv
hendrikweisser: Yep, Modomics nomenclature.

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
hendrikweisser: @poshul: Thanks, that would be helpful. Are you okay to update the bindings for TheoreticalSpectrumGenerator[NA] as well?

hendrikweisser:
> can you fix these tests:

I've replaced `std::regex` (which GCC 4.8 doesn't have) with `boost::regex` - hopefully that'll solve it.

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
poshul: Yes, I can do that, though it probably won't happen until tomorrow.

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
hendrikweisser: I'm going to try this again on another large example.

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
```diff
@@ -0,0 +1,554 @@
...
+    Size size = uncharged_spectrum.size();
+    if (add_precursor_peaks_ && !add_precursor)
+    {
+      --size; // uncharged spectrum contains precursor peak - exclude it
+    }
+    for (Size i = 0; i < size; ++i)
+    {
+      spectrum.push_back(uncharged_spectrum[i]);
+      double mass = spectrum.back().getMZ() + charge * H_mass;
+      spectrum.back().setMZ(abs(mass / charge));
```
cbielow: its more efficient (less computation) and numerically more stable if you do:
`setMZ( old_mz / charge + H_mass);`

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
```diff
@@ -0,0 +1,554 @@
...
+    }
+    if (add_metainfo_)
+    {
+      auto& ions = spectrum.getStringDataArrays()[0];
+      auto source_it = uncharged_spectrum.getStringDataArrays()[0].begin();
+      ions.insert(ions.end(), source_it, source_it + size);
+      auto& charges = spectrum.getIntegerDataArrays()[0];
+      charges.resize(charges.size() + size, charge);
```
cbielow: how do you know that `[0]` is the charge array?
Why resize based on old value (the number of elements should be exactly 'size')?

src/openms/source/CHEMISTRY/sources.cmake
```diff
@@ -29,12 +29,12 @@ RNaseDigestion.cpp
 Ribonucleotide.cpp
 RibonucleotideDB.cpp
 SpectrumAnnotator.cpp
-TheoreticalSpectrumGeneratorXLMS.cpp
 SimpleTSGXLMS.cpp
 SvmTheoreticalSpectrumGenerator.cpp
 SvmTheoreticalSpectrumGeneratorTrainer.cpp
 SvmTheoreticalSpectrumGeneratorSet.cpp
 TheoreticalSpectrumGenerator.cpp
+TheoreticalSpectrumGeneratorNA.cpp
```
cbielow: `NA` has a weird connotation.
Would `NucAcid` be an option?

hendrikweisser: I understand your objection, but I don't like the "NucAcid" suffix. How about "NucleicAcidSpectrumGenerator"?

poshul:
> > can you fix these tests:
> 
> I've replaced `std::regex` (which GCC 4.8 doesn't have) with `boost::regex` - hopefully that'll solve it.

I'm seeing:
	484 - SVMWrapper_test (Failed)
	1652 - TOPP_TargetedFileConverter_test_1 (Failed)
	1653 - TOPP_TargetedFileConverter_test_1_out1 (Failed)
	1654 - TOPP_TargetedFileConverter_test_2 (Failed)
	1655 - TOPP_TargetedFileConverter_test_2_out1 (Failed)
	1657 - TOPP_TargetedFileConverter_test_3 (Failed)
	1659 - TOPP_TargetedFileConverter_test_4 (Failed)
	1665 - TOPP_TargetedFileConverter_test_9 (Failed)
	1778 - TOPP_OpenSwathWorkflow_13 (Failed)
	1781 - TOPP_OpenSwathWorkflow_14 (Failed)
	1798 - TOPP_OpenSwathWorkflow_17_b (Failed)

On my Ubuntu 18.04 build machine. The SVMWrapper test has never worked on this machine but the others seem to be new (though not necessarily related to this specific PR).

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
hendrikweisser: Using the first array (`[0]`) is a convention from the original `TheoreticalSpectrumGenerator`. I don't really like it, either - see https://github.com/OpenMS/OpenMS/pull/4020/files/dcb360e2cf28073f5906e64fc7cd833aacd26e2b#diff-429a6a69257db40684324599fff0e7bcR390. I think it would have made more sense to use the last array. I didn't change the convention because I don't know what parts of OpenMS depend on it. At least it's documented: https://github.com/OpenMS/OpenMS/pull/4020/files/dcb360e2cf28073f5906e64fc7cd833aacd26e2b#diff-5408e877edfef997d52836ef8df272b7L58

This function *adds* peaks to a spectrum. So the `resize` extends the size of the array by the number of new peaks.

hendrikweisser:
> The SVMWrapper test has never worked on this machine but the others seem to be new (though not necessarily related to this specific PR).

They failed for me as well, but I ignored them because they seemed unrelated.

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
hendrikweisser: Good point. That also occurs for the precursor peak (twice) in `getMultipleSpectra`.

hroest:
regarding SVM, are you using the contrib libsvm or the Ubuntu one? See also this discussion: https://github.com/OpenMS/OpenMS/issues/521#issue-25311239 it seems that libsvm >= 3.14 do not work well with our tests. At some point we should upgrade

hendrikweisser:
> regarding SVM, are you using the contrib libsvm or the Ubuntu one?

Ubuntu (libsvm 3.21).

hroest:
> Ubuntu (libsvm 3.21).

thats probably the issue, if you take the lib from the contrib it should work

poshul:
Same here. I guess thats the issue.
> > Ubuntu (libsvm 3.21).
> 
> thats probably the issue, if you take the lib from the contrib it should work

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
poshul: Okay, dumb Cython question, ConstRibonucleotidePtr is defined: 
`using ConstRibonucleotidePtr = const Ribonucleotide *;`
How do I include that in a pxd? Attempting to use the underlying type ie
`   libcpp_pair[ const Ribonucleotide *, const Ribonucleotide * ] getRibonucleotideAlternatives(libcpp_string & code) nogil except +`
 throws 'arg code wrong type'.

How do I let Cython know the correct type?

src/openms/include/OpenMS/CHEMISTRY/Ribonucleotide.h
poshul: I've done this. I'll submit it in a PR once I figure out the other two pxd related things.

hroest:
currently fails on windows https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-15,os_label=win10_64/2912/testReport/junit/projectroot.src.tests.class_tests/openms/ModifiedNASequenceGenerator_test/
```
checking static void applyVariableModifications(const std::set<ConstRibonucleotidePtr>& var_mods, const NASequence& seq, Size max_variable_mods_per_NASequence, std::vector<NASequence>& all_modified_NASequences, bool keep_original = true) ... 
 +  line 87:  TEST_EQUAL(ams.size(),7): got '7', expected '7'
 +  line 88:  TEST_STRING_EQUAL(ams[0].toString(),NASequence::fromString("AUAUAUA").toString()): got "AUAUAUA", expected "AUAUAUA"
 -  line 89:  TEST_STRING_EQUAL(ams[1].toString(),NASequence::fromString("AUAUA[m3U]A").toString()): got "AUAUA[s4U]A", expected "AUAUA[m3U]A"
 -  line 90:  TEST_STRING_EQUAL(ams[2].toString(),NASequence::fromString("AUAUA[s4U]A").toString()): got "AUAUA[m3U]A", expected "AUAUA[s4U]A"
 -  line 91:  TEST_STRING_EQUAL(ams[3].toString(),NASequence::fromString("AUA[m3U]AUA").toString()): got "AUA[s4U]AUA", expected "AUA[m3U]AUA"
 -  line 92:  TEST_STRING_EQUAL(ams[4].toString(),NASequence::fromString("AUA[s4U]AUA").toString()): got "AUA[m3U]AUA", expected "AUA[s4U]AUA"
 -  line 93:  TEST_STRING_EQUAL(ams[5].toString(),NASequence::fromString("A[m3U]AUAUA").toString()): got "A[s4U]AUAUA", expected "A[m3U]AUAUA"
 -  line 94:  TEST_STRING_EQUAL(ams[6].toString(),NASequence::fromString("A[s4U]AUAUA").toString()): got "A[m3U]AUAUA", expected "A[s4U]AUAUA"
 +  line 100:  TEST_EQUAL(ams.size(),6): got '6', expected '6'
 +  line 105:  TEST_EQUAL(ams.size(),3*3*3): got '27', expected '27'
 +  line 114:  TEST_EQUAL(ams.size(),3*3*3*2*2*2*2): got '432', expected '432'
: failed

: passed

FAILED
Failed lines: 89 90 91 92 93 94 
```

maybe a sorting issue?

src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
hendrikweisser: Apparently none of the other functions of `RibonucleotideDB` have been wrapped so far. Does `const Ribonucleotide *` work as a return type when you wrap e.g. `getRibonucleotide`? Based on how `ResidueDB` is wrapped, I would expect it does.

src/openms/source/CHEMISTRY/sources.cmake
hendrikweisser: I've renamed the class.

src/openms/source/CHEMISTRY/TheoreticalSpectrumGeneratorNA.cpp
hendrikweisser: On this other example, I get slightly *more* hits with `PROTON_MASS_U` (1174 instead of 1148 at 5% FDR). :-/
What now?

PR #4026, opened 2019-03-28
[FEATURE] moved SSE code to class, added documentation

timosachsenberg:
Test are currently done via TOPP tests. I added a stub for the class tests, though.


timosachsenberg:
I just figured I forgot the pxd files...

hendrikweisser:
> I just figured I forgot the pxd files...

Always the pesky pyOpenMS... ;-)

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -0,0 +1,654 @@
...
+      int scan_index = s_it - spectra.begin();
+      vector<Precursor> precursor = s_it->getPrecursors();
+
+      // there should only one precursor and MS2 should contain at least a few peaks to be considered (e.g. at least for every AA in the peptide)
+      if (precursor.size() == 1 && s_it->size() >= peptide_min_size_)
+      {
+        int precursor_charge = precursor[0].getCharge();
+
+        if (precursor_charge < precursor_min_charge_ 
+         || precursor_charge > precursor_max_charge_)
```
hroest: 395 / 396: 

/.../OpenMS/src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp:395:32: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]


the question is whether the charge can actually be negative here - do you support negative charges?

src/tests/class_tests/openms/source/SimpleSearchEngineAlgorithm_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+START_SECTION(~SimpleSearchEngineAlgorithm())
+{
+	delete ptr;
+}
+END_SECTION
+
+START_SECTION((ExitCodes search(const String &in_mzML, const String &in_db, std::vector< ProteinIdentification > &prot_ids, std::vector< PeptideIdentification > &pep_ids) const ))
+{
+  // TODO
```
hendrikweisser: What we usually do in these cases is put `NOT_TESTABLE // tested by the tool test`.

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
timosachsenberg: no I am even not sure if the TheoreticalSpectraGenerators do so this is always >=1

timosachsenberg:
pyopenms/pyopenms_4.cpp:119043:15: error: cannot convert 'OpenMS::SimpleSearchEngineAlgorithm::ExitCodes' to 'int' in assignment
     __pyx_t_9 = __pyx_v_self->inst.get()->search((*__pyx_f_8pyopenms_10pyopenms_1_convString(__pyx_v_in_mzML).get()), (*__pyx_f_8pyopenms_10pyopenms_1_convString(__pyx_v_in_db).get()), (*__pyx_v_v2), (*__pyx_v_v3));
what's again the proper way to deal with enum classes? does this need to be an enum instead?

hroest:
> what's again the proper way to deal with enum classes? does this need to be an enum instead?

yes, see https://github.com/OpenMS/OpenMS/blob/1db1c512204b7345cdd1ba0668b4efd6d9624164/src/pyOpenMS/pxds/CVMappingRule.pxd#L59-L67

src/openms/include/OpenMS/ANALYSIS/ID/SimpleSearchEngineAlgorithm.h
```diff
@@ -0,0 +1,180 @@
...
+
+#include <map>
+#include <algorithm>
+
+#ifdef _OPENMP
+  #include <omp.h>
+  #define NUMBER_OF_THREADS (omp_get_num_threads())
+#else
+  #define NUMBER_OF_THREADS (1)
+#endif
```
hroest: is this required for anything? I would remove that from the header at least...

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -0,0 +1,654 @@
...
+          phs.push_back(ph);
+        }
+        pi.setHits(phs);
+        pi.assignRanks();
+
+#ifdef _OPENMP
+#pragma omp critical (peptide_ids_access)
+#endif
+        {
+          peptide_ids.push_back(pi);
```
hroest: could you move them back here?
```suggestion
          peptide_ids.push_back(std::move(pi));
```

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -0,0 +1,654 @@
...
+    search_parameters.mass_type = mass_type;
+    search_parameters.fixed_modifications = modifications_fixed;
+    search_parameters.variable_modifications = modifications_variable;
+    search_parameters.missed_cleavages = peptide_missed_cleavages;
+    search_parameters.fragment_mass_tolerance = fragment_mass_tolerance;
+    search_parameters.precursor_mass_tolerance = precursor_mass_tolerance;
+    search_parameters.precursor_mass_tolerance_ppm = precursor_mass_tolerance_unit_ppm == "ppm" ? true : false;
+    search_parameters.fragment_mass_tolerance_ppm = fragment_mass_tolerance_unit_ppm == "ppm" ? true : false;
+    search_parameters.digestion_enzyme = *ProteaseDB::getInstance()->getEnzyme(enzyme);
+    protein_ids[0].setSearchParameters(search_parameters);
```
hroest:
```suggestion
    protein_ids[0].setSearchParameters(std::move(search_parameters));
```

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -0,0 +1,654 @@
...
+        }
+
+        // skip peptides that have already been processed
+        if (already_processed) { continue; }
+
+#ifdef _OPENMP
+#pragma omp critical (processed_peptides_access)
+#endif
+        {
+          processed_petides.insert(c);
```
hroest: I think you would want to have that in the same `omp critical` section, otherwise you could have a race condition: two identical peptides start at the top, both find that they are not in `processed_petides` and then they both progress to here, which is exactly what you want to avoid. I think you need to have

```

#ifdef _OPENMP
#pragma omp critical (processed_peptides_access)
#endif
        {
          // peptide (and all modified variants) already processed so skip it
          if (processed_petides.find(c) != processed_petides.end())
          {
            already_processed = true;
          }
       
          // skip peptides that have already been processed
          if (!already_processed) processed_petides.insert(c);
        }
        if (already_processed) { continue; }
```

this will prevent the race condition

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -0,0 +1,654 @@
...
+            ah.score = score;
+
+#ifdef _OPENMP
+            omp_set_lock(&(annotated_hits_lock[scan_index]));
+            {
+#endif
+              annotated_hits[scan_index].push_back(ah);
+
+              // prevent vector from growing indefinitly (memory) but don't shrink the vector every time
+              if (annotated_hits[scan_index].size() >= 2 * report_top_hits_)
```
hroest: this might work well for large values of `report_top_hits_` but for the default value of 1, this will cause shrinking for every second hit:

- add hit, vector.size() = 1
- add hit, vector.size() = 1
- add hit, vector.size() = 2
- add hit, vector.size() = 3 -> shrink to 1
- add hit, vector.size() = 2
- add hit, vector.size() = 3 -> shrink to 1
...

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -0,0 +1,654 @@
...
+    spectrum_generator.setParameters(param);
+
+    // preallocate storage for PSMs
+    vector<vector<AnnotatedHit_> > annotated_hits(spectra.size(), vector<AnnotatedHit_>());
+    for (auto & a : annotated_hits) { a.reserve(2 * report_top_hits_); }
+
+#ifdef _OPENMP
+    // we want to do locking at the spectrum level so we get good parallelisation 
+    vector<omp_lock_t> annotated_hits_lock(annotated_hits.size());
+    for (size_t i = 0; i != annotated_hits_lock.size(); i++) { omp_init_lock(&(annotated_hits_lock[i])); }
```
hroest: is there a performance issue with using many locks in OpenMP? These will be potentially hundreds of thousands of locks...

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
timosachsenberg: you are right. this could lead to the same peptide being scored twice. otherwise I think there should be no harm done. I will change it

timosachsenberg: I changed it to:
```
#pragma omp critical (processed_peptides_access)
        { 
          // peptide (and all modified variants) already processed so skip it
          if (processed_petides.find(c) != processed_petides.end())
          { 
            already_processed = true;
          }
          else
          { 
            processed_petides.insert(c);
          }
        }             
                      
        // skip peptides that have already been processed
        if (already_processed) { continue; }
```
and wondered if putting the continue statement in the critical section is allowed.
@cbielow or @hroest I did not find this in the specification. Do you know if that is allowed?

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
timosachsenberg: I once used a priority queue and it was slightly slower. One reason might be that no reallocation is happening and the only thing happening during the resize is calling the destructor of AnnotatedHit. Maybe this is a noop for this simple struct?

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
timosachsenberg: Good question and I don't know the definite answer to that. For sure, it scales much better if every single spectrum is locked. Also, a critical section is internally also modelled using a lock. As the locks are in the shared memory there might be some speed problems related to loading the locks into the cache. @cbielow what is your take on that?

Issue #1042, opened 2014-10-09
EmpiricalFormula's charge is misleading?


cbielow:
A careful reviewer has come up with this (I'll cite my correspondence with him below).
This is related to #992 and #1041.

To summarize:
The goal here would be to find an interface which allow to write
EmpiricalFormula("Na").setCharge(1)  // get weight of Na with missing electron!

but at the same time allow
EmpiricalFormula ef ("C20H100O10S2").setCharge(2);  // some uncharged peptide formula
to give the expected m/z position without explicitly adding 2H.

Maybe an interface allowing to add an explicit adduct internally:
setCharge(SignedSize charge, Element\* adduct = 0);
With this, you can perfectly model the "Na" case as written above, and
EmpiricalFormula ef ("C20H100O10S2").setCharge(2, "H") for the peptide case.

Mixing adducts (e.g. H+Na+ as +2 would not be possible here though)...

Open for discussion, and suggestions.

> > Am 10/9/2014 12:32 PM, schrieb ????? ?????:
> > 
> > >    Dear Colleague,
> > > 
> > > Excuse me if this question was already asked, but please give explanation of feature of EmpiricalFormula.C module, which looks like a bug. The question is about mass calculation functions.
> > > 
> > > For example:
> > > 
> > > DoubleRealEmpiricalFormula :: getMonoWeight ()  const
> > > {
> > >     DoubleRealweight(0);
> > >     if(charge_>0)
> > >     {
> > >       weight+=Constants::PROTON_MASS_U_charge_;
> > >     }
> > >     Map<constElement_,SignedSize>::ConstIteratorit=formula_.begin();
> > >     for(;it!=formula_.end();++it)
> > >     {
> > >       weight+=it->first->getMonoWeight()*(DoubleReal)it->second;
> > >     }
> > >     return weight;
> > > }
> > > 
> > > Code looks strange, as it is uncertain why to add proton masses to molecule mass.  Another strange thing is why to do this only for positively charged molecules? What is about negatively charged molecules?
> > > 
> > > ```
> > > if(charge_>0)
> > > {
> > >   weight+=Constants::PROTON_MASS_U*charge_;
> > > }
> > > ```
> > > 
> > > That looks like too obvious to be error...
> > > 
> > > Would it be correct to use ELECTRON mass? Just subtract the number of electrons according to charge?
> > >     weight -= Constants::ELECTRON_MASS_U*charge_;
> > > 
[...]

PR #4027, opened 2019-03-28
Support negative charges in 'EmpiricalFormula::getAverageWeight' (fixes #1041)

### LINKED ISSUE #1041: EmpiricalFormula computes incorrect weight for negative charge
Positive charges are currently implicitly modeled as proton adducts (which are not explicitly part of the formula).
When the charge is positive, proton masses are added
if(charge_>0)
{
  weight+=Constants::PROTON_MASS_U*charge_;
}

However, when charge is negative, nothing is done.
Hoewever, even if the weight is small, we should:

if (charge_ < 0)
{
  weight -= Constants::ELECTRON_MASS_U*charge_;
}

hendrikweisser:
This makes `getAverageWeight` consistent with `getMonoWeight` (which was fixed previously as part of  #2874).
In all cases, protons are assumed to be the charge carriers that are gained/lost (which is problematic, but not addressed here - see #1042).

Also added tests for `getAverageWeight` and `getMonoWeight` with charges and removed some unnecessary whitespace.

### DIFF
--- a/src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
+++ b/src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
@@ -83,11 +83,7 @@ namespace OpenMS
 
   double EmpiricalFormula::getAverageWeight() const
   {
-    double weight(0);
-    if (charge_ > 0)
-    {
-      weight += Constants::PROTON_MASS_U * charge_;
-    }
+    double weight = Constants::PROTON_MASS_U * charge_;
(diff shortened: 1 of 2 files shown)


hendrikweisser:
"openms_pr_docker_pyopenms" CI failure is unrelated:
```
-- Could NOT find Boost
CMake Error at src/openswathalgo/CMakeLists.txt:44 (message):
  Boost was not found!
```

hroest:
:+1: but we should really fix / address #1042

PR #4028, opened 2019-03-28
[FEATURE] move MzTabExport to lib and extend export

timosachsenberg:
pulling out current MzTabExport from lfq branch seems to be a bit more involved. @cbielow this is probably most likely what you want to work with. It contains much more information and the output has been validated by PRIDE (at least for consensusXML with IDs and inference information).
I get some failing tests on my side potentially related to moving to unordered_set for the inference code ...


src/openms/include/OpenMS/METADATA/ProteinHit.h
```diff
@@ -171,6 +196,11 @@ namespace OpenMS
     /// sets the coverage (in percent) of the protein hit based upon matched peptides
     void setCoverage(const double coverage);
 
+    /// returns the set of modified protein positions
+    const std::set<std::pair<Size, ResidueModification> >& getModifications() const;
+
+    /// sets the set of modified protein positions
+    void setModifications(std::set<std::pair<Size, ResidueModification> >& mods);
```
hroest: can you add Python bindings for these?

src/openms/include/OpenMS/METADATA/ProteinIdentification.h
```diff
@@ -95,6 +108,100 @@ namespace OpenMS
...
+    private:
+      /// Float data arrays
+      FloatDataArrays float_data_arrays_;
+
+      /// String data arrays
+      StringDataArrays string_data_arrays_;
+
+      /// Integer data arrays
+      IntegerDataArrays integer_data_arrays_;
     };
```
hroest: can you add Python bindings for these?

hroest:
looks good, can you add Python bindings the new library functions?

timosachsenberg:
Quick question because one change is a bit controversial.
To transport quantification at protein group level (=single protein or true group) to mzTab, @jpfeuffer and me currently annotate the quantities in a float data array attached to the protein group.
While this works great some may not like that because:
- some may argue that quant. data is in the id data structure
- float data arrays are not as easy to encode in @hendrikweisser  new ID data structure

Among other alternatives, we thought about (e.g., encoding in a consensus feature) did not play well together with the final result file that contains fraction wise consensus features that map to protein ids (and not groups).

Any thoughts about that?

hendrikweisser:
Without having been part of the previous discussions/deliberations, some thoughts from my side:
- It would be desirable if we could find a solution that is easily adaptable to the new ID data structures.
- Those datastructures can absolutely be amended to support new requirements. If we decide the best solution is to store (some) quant. data directly in there, I'm sure that wouldn't be very difficult to add.
- I'm sceptical whether "hiding" fundamental information, like quantitative data, in meta data structures is a good idea.
- If we don't have a suitable data structure for this type of quantitative data at the moment, maybe we should make one.
- A quantitative data structure could easily reference relevant parts of the new ID data structure (peptides, proteins, protein groups...). A possible design for mzTab, where we want to export ID data with optional quant. data, would be: `export(IdentificationData id_data, QuantData q_data = QuantData())`, where `q_data` contains references into `id_data` and has an empty default.

hroest:
> If we don't have a suitable data structure for this type of quantitative data at the moment, maybe we should make one.

I agree, it should be explicit and not implicit if possible.

cbielow:
this seems all very reasonable. For a current project it would be very neat to have the current functionality as provided by this PR in the library.
I suggest a two step process where we: address the immediate comments (python bindings) and go forward with a merge. In a second step (dev meeting?) we can come up with a solution to the data structure problem.

lars20070:
:+1: for two small steps.

Would be good to have this PR merged before the developer meeting. I would like to see if the problem in  https://github.com/OpenMS/OpenMS/pull/3973 persists after a rebase.

src/openms/include/OpenMS/METADATA/ProteinHit.h
timosachsenberg: so I need an extra struct again for the set I guess

timosachsenberg:
Ok that's fine with me to merge it before the developer meeting if that increases our chance to make some real progress on that.

PR #4029, opened 2019-03-28
Feature/ams used charge adducts

FabianAicheler:
Add option/support to compare AMS candidate adduct with available feature adduct.

Tested whether it leads to correct subset for previous test inputs extended by some adducts, I think it should support all kinds of adducts (including M;+ and M;- via protonation/deprotonation?)

Use old school String lists for possible param values because old school Param class complains otherwise? Too many problems with the abinode nfs today and I wanted to keep possible sources of conflict minimal.


FabianAicheler:
No idea why it shows no float precision changes for the different output test files, were they already all updated server side? Had to at least locally replace them for the recently changed additional digits of floats.. No idea, don't want to think about it more^^

FabianAicheler:
@timosachsenberg , @oliveralka I think the changes should cover all possible adducts. Stuff like 2M should be internally handled by AMS/ is separated from the adduct when parsing the Positive/negative mode adduct files.

I used a EF cast instead of direct String comparison because I'm still not sure how standardized adducts are written out/in across different tools. Anyways, AMS is fast, this should not hurt.

FabianAicheler:
These changes are only designed for featureXML input, with possible adduct annotations.

For MetaboliteAdductDecharger output consensusXML, the centroids of successfully grouped features represent neutral masses anyway (so something like "M;+" as AMS input adduct list), and ungrouped monositopic features should be removed/split of via FileFilter before use of AMS for treatment with a separate adduct list.

For FeatureLinker consensusXML output I didn't know/want to address what to do for consensus features having multiple different adducts linked. -> currently not changed.

hroest:
How do you currently handle non-proton adducts? I think we really need to think about #1042 again and find a better solution that also makes sense for non-proton adducts!

FabianAicheler:
What do you mean with non-Proton adducts? Simply stuff like "M+Na"?

Afaik AMS parses the "Positive adduct" and "Negative adduct" file lists, separates the adduct from the 'M' part, and charge substrings and uses the adduct part itself only for the sumformula.

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -675,7 +690,15 @@ namespace OpenMS
 
     std::vector<AccurateMassSearchResult> results_part;
 
-    queryByMZ(feature.getMZ(), feature.getCharge(), ion_mode, results_part);
+    String use_feature_adducts = param_.getValue("use_feature_adducts").toString();
+    if ((use_feature_adducts == "true") && feature.metaValueExists("dc_charge_adducts"))
```
oliveralka: not sure if this is needed, but you could convert to bool.
bool use_feature_adducts = (param_.getValue("use_feature_adducts").toString() == "true") ? true : false;
if ((use_feature_adducts) && ..

oliveralka:
Nice addition I am happy to test in on my data.

timosachsenberg:
please adapt the pyOpenMS bindings

FabianAicheler:
Hopefully does the pyopenms thing. Couldn't test myself, because I'm fighting with installing the required stuff for pyOpenMS.

hroest:
> Hopefully does the pyopenms thing. Couldn't test myself, because I'm fighting with installing the required stuff for pyOpenMS.

please use the docker container for testing as documented here https://github.com/OpenMS/OpenMS/wiki/pyOpenMS#--testing

FabianAicheler:
Thx will do.

oliveralka:
rebuild jenkins

src/openms/include/OpenMS/ANALYSIS/ID/AccurateMassSearchEngine.h
```diff
@@ -86,6 +86,9 @@ namespace OpenMS
     /// original string used for parsing
     const String& getName() const;
 
+    // EF of adduct itself. Useful for comparison with feature adduct annotation
```
timosachsenberg: /// for doxygen compatible comment

src/openms/include/OpenMS/ANALYSIS/ID/AccurateMassSearchEngine.h
```diff
@@ -267,7 +270,7 @@ namespace OpenMS
       @brief search for a specific observed mass by enumerating all possible adducts and search M+X against database
```
timosachsenberg: please extend documentation including parameters

src/openms/include/OpenMS/ANALYSIS/ID/AccurateMassSearchEngine.h
```diff
@@ -267,7 +270,7 @@ namespace OpenMS
...
 
        */
-    void queryByMZ(const double& observed_mz, const Int& observed_charge, const String& ion_mode, std::vector<AccurateMassSearchResult>& results) const;
+    void queryByMZ(const double& observed_mz, const Int& observed_charge, const String& ion_mode, std::vector<AccurateMassSearchResult>& results, const EmpiricalFormula& observed_adduct=EmpiricalFormula()) const;
```
timosachsenberg: spaces around =

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -581,6 +590,12 @@ namespace OpenMS
         continue;
       }
 
+      if ((observed_adduct != EmpiricalFormula()) && (observed_adduct != it->getEmpiricalFormula()))
```
timosachsenberg: is the comment correct? or is this more like the charge 0 case (not annotated)

FabianAicheler:
> please use the docker container for testing as documented here https://github.com/OpenMS/OpenMS/wiki/pyOpenMS#--testing

Can't successfully use the docker, get unrelated error:

c++: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.8/README.Bugs> for instructions.
make[3]: *** [src/openms/CMakeFiles/OpenMS.dir/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp.o] Error 4
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [src/openms/CMakeFiles/OpenMS.dir/all] Error 2
make[1]: *** [src/pyOpenMS/CMakeFiles/pyopenms.dir/rule] Error 2
make: *** [pyopenms] Error 2

FabianAicheler:
The only failed test with Jenkins pyOpenMS is "projectroot.pyOpenMS.pyopenms_test_memoryleaktests".

Does this mean that my PyOpenMS changes worked out?

FabianAicheler:
Requested changes addressed from my side, failed nightly/pyOpenMS runs assumed to be unrelated.

PR #4036, opened 2019-04-03
lockfree generation of modified peptides

timosachsenberg:
Before:
Generating 29 * 1e5 peptides took:
1 thread: 16s
20 threads: 17s
After:
1 thread 15s
20 threads 0.6s

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/RNPXL/ModifiedPeptideGenerator.h
+++ b/src/openms/include/OpenMS/ANALYSIS/RNPXL/ModifiedPeptideGenerator.h
@@ -52,17 +52,21 @@ namespace OpenMS
   public
 :
     // Applies fixed modifications to a single peptide
-    static void applyFixedModifications(const std::vector<ResidueModification>::const_iterator& fixed_mods_begin, const std::vector<ResidueModification>::const_iterator& fixed_mods_end, AASequence& peptide);
+    static void applyFixedModifications(const std::vector<const ResidueModification*>& fixed_mods, AASequence& peptide);
 
     // Applies variable modifications to a single peptide. If keep_original is set the original (e.g. unmodified version) is also returned
-    static void applyVariableModifications(const std::vector<ResidueModification>::const_iterator& var_mods_begin, const std::vector<ResidueModification>::const_iterator& var_mods_end, const AASequence& peptide, Size max_variable_mods_per_peptide, std::vector<AASequence>& all_modified_peptides, bool keep_original=true);
+    static void applyVariableModifications(const std::vector<const ResidueModification*>& var_mods, 
+     const AASequence& peptide, 
+     Size max_variable_mods_per_peptide, 
+     std::vector<AASequence>& all_modified_peptides, 
+     bool keep_original=true);
 
   protected:
     // Recursively generate all combinatoric placements at compatible sites
-    static void recurseAndGenerateVariableModifiedPeptides_(const std::vector<int>& subset_indices, const std::map<int, std::vector<ResidueModification> >& map_compatibility, int depth, const AASequence& current_peptide, std::vector<AASequence>& modified_peptides);
+    static void recurseAndGenerateVariableModifiedPeptides_(const std::vector<int>& subset_indices, const std::map<int, std::vector<const ResidueModification*> >& map_compatibility, int depth, const AASequence& current_peptide, std::vector<AASequence>& modified_peptides);
 
     // Fast implementation of modification placement. No combinatoric placement is needed in this case - just every site is modified once by each compatible modification. Already modified residues are skipped
-    static void applyAtMostOneVariableModification_(const std::vector<ResidueModification>::const_iterator& var_mods_begin, const std::vector<ResidueModification>::const_iterator& var_mods_end, const AASequence& peptide, std::vector<AASequence>& all_modified_peptides, bool keep_original=true);
+    static void applyAtMostOneVariableModification_(const std::vector<const ResidueModification*>& var_mods, const AASequence& peptide, std::vector<AASequence>& all_modified_peptides, bool keep_original=true);
 
   };
 }
--- a/src/openms/include/OpenMS/ANALYSIS/RNPXL/ModifiedPeptideGenerator.h
+++ b/src/openms/include/OpenMS/ANALYSIS/RNPXL/ModifiedPeptideGenerator.h
@@ -54,7 +54,8 @@ namespace OpenMS
     */
 
   public:
-    using MapToResidueType = boost::container::flat_map<const ResidueModification*, const Residue*>;
+    // struct needed to wrap the template for pyOpenMS
+    struct MapToResidueType { boost::container::flat_map<const ResidueModification*, const Residue*> val; };
 
       /**
       * @brief Retrieve modifications from strings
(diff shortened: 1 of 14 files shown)


timosachsenberg:
To generate modified AASequences in ModifiedPeptideGenerator required to lock and query the modified Residue* via a String for every single modified Sequence.
Now, all modified residues are queried only once into a look-up structure and can be applied lock-free.

timosachsenberg:
maybe also relevant for @hendrikweisser

hroest:
> After:
> 1 thread 15s
> 20 threads 0.6s

hmm that is hard to believe, your total CPU time decreased? 20x 0.6  = 12s :-)

src/openms/include/OpenMS/ANALYSIS/RNPXL/ModifiedPeptideGenerator.h
```diff
@@ -49,25 +53,54 @@ namespace OpenMS
...
     // Applies fixed modifications to a single peptide
-    static void applyFixedModifications(const std::vector<const ResidueModification*>& fixed_mods, AASequence& peptide);
+    static void applyFixedModifications(
+      const MapToResidueType& fixed_mods, 
+      AASequence& peptide);
 
     // Applies variable modifications to a single peptide. If keep_original is set the original (e.g. unmodified version) is also returned
     static void applyVariableModifications(
-     const std::vector<const ResidueModification*>& var_mods, 
+     const MapToResidueType& var_mods,
```
hroest: this wont fly in pyOpenMS I think, either

- use the old interface for Python
- delete the interface
- create a small struct around the type:
```
struct MapToResidue {boost::container::flat_map<const ResidueModification*, const Residue*> rmap;}
```

src/openms/source/ANALYSIS/RNPXL/ModifiedPeptideGenerator.cpp
```diff
@@ -33,32 +33,79 @@
...
+ // static
+ ModifiedPeptideGenerator::MapToResidueType ModifiedPeptideGenerator::getModifications(const StringList& modNames)
+ {   
+   vector<const ResidueModification*> modifications;
+  
+   for (String modification : modNames)
+   {
+     const ResidueModification* rm;
+
+     // TODO: check if this is still required
```
hroest: is it required?

src/openms/source/ANALYSIS/RNPXL/ModifiedPeptideGenerator.cpp
```diff
@@ -281,31 +338,33 @@ namespace OpenMS
...
       }
       else if (current_index == N_TERM_MODIFICATION_INDEX)
       {
-        new_peptide.setNTerminalModification(m->getFullName());
+        new_peptide.setNTerminalModification(m);
       }
       else
       {
-        new_peptide.setModification(current_index, m->getFullName());
+        const Residue* r = var_mods.at(m); // map modification to the modified residue
```
hroest: any reason you use `at` and not direct access? Is it possible that the Residue is not found?

src/openms/source/ANALYSIS/RNPXL/ModifiedPeptideGenerator.cpp
```diff
@@ -281,31 +338,33 @@ namespace OpenMS
...
       // recurse with modified peptide
-      recurseAndGenerateVariableModifiedPeptides_(subset_indices, map_compatibility, depth + 1, new_peptide, modified_peptides);
+      recurseAndGenerateVariableModifiedPeptides_(subset_indices, map_compatibility, var_mods, depth + 1, new_peptide, modified_peptides);
     }
   }
 
   // static
   void ModifiedPeptideGenerator::applyAtMostOneVariableModification_(
-    const vector<const ResidueModification*>& var_mods, 
+    const boost::container::flat_map<const ResidueModification*, const Residue*>& var_mods,
```
hroest:
```suggestion
    const MapToResidueType& var_mods, 
```

hroest:
>  Now, all modified residues are queried only once into a look-up structure and can be applied lock-free.

great! The only concern is that pyOpenMS is broken since it wont play nice with the `boost::flat_map` at the moment. As I indicated, maybe wrap that in its own little class

src/openms/source/ANALYSIS/RNPXL/ModifiedPeptideGenerator.cpp
timosachsenberg: Because .at is const access . Bracket is non-const

timosachsenberg:
I also added some moves which improved performance a bit (that’s probably why it adds up to a lower value). Main concern was scalability though.

src/openms/include/OpenMS/ANALYSIS/RNPXL/ModifiedPeptideGenerator.h
timosachsenberg: I found a way that does not need the map and should be compatible with pyOpenMS

src/openms/source/ANALYSIS/RNPXL/ModifiedPeptideGenerator.cpp
timosachsenberg: I changed the relevant part so it is not required anymore

hroest:
seems like you have some sorting issue?

 -  line 200:  TEST_EQUAL(modified_peptides[0].toString(),"ACAACAAC(Carbamidomethyl)A"): got 'ACAACAAC(Glutathione)A', expected 'ACAACAAC(Carbamidomethyl)A'
 -  line 201:  TEST_EQUAL(modified_peptides[1].toString(),"ACAACAAC(Glutathione)A"): got 'ACAACAAC(Carbamidomethyl)A', expected 'ACAACAAC(Glutathione)A'

src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
```diff
@@ -134,7 +128,11 @@ namespace OpenMS
        * @param c_term_linker True, if the cross-linker can react with the C-terminal of a protein
        * @return A vector of AASeqWithMass containing the peptides, their masses and information about terminal peptides
        */
-      static std::vector<OPXLDataStructs::AASeqWithMass> digestDatabase(std::vector<FASTAFile::FASTAEntry> fasta_db, EnzymaticDigestion digestor, Size min_peptide_length, StringList cross_link_residue1, StringList cross_link_residue2, const std::vector<const ResidueModification*>& fixed_modifications, const std::vector<const ResidueModification*>& variable_modifications, Size max_variable_mods_per_peptide);
+      static std::vector<OPXLDataStructs::AASeqWithMass> digestDatabase(std::vector<FASTAFile::FASTAEntry> fasta_db, 
+        EnzymaticDigestion digestor, Size min_peptide_length, StringList cross_link_residue1, StringList cross_link_residue2,
+        const ModifiedPeptideGenerator::MapToResidueType& fixed_modifications, 
+        const ModifiedPeptideGenerator::MapToResidueType& variable_modifications, 
+        Size max_variable_mods_per_peptide);
```
hroest: pyOpenMS bindings?

timosachsenberg:
tests passed including pyOpenMS

PR #4037, opened 2019-04-03
[UPDATE] AssayGeneratorMetabo

oliveralka:
AssayGeneratorMetabo is now able to use prior feature or identification information and SIRIUS for MS2 spectra fragment annotation. Transistions are then extracted based on the annotation and the intensity. Afterwards they are filtered via MRMAssay (min, max transistions) and an assay library is generated (tsv, TraML, pqp). 

Additional the non fragmentation annotation method was retainend and can use the deisotoper to reduce the peak set of the MS2 spectra, to work with "clean" spectra for transistion extraction. Here either the MS2 spectrum with the highest intensity precursor or a consensus spectrum (cosine similarity check) can be used.  

Both modes are able to use semi-targeted mode, which basically does not need identification information. Here for the unkown features (e.g. FeatureFinderMetabo) the transistions are extracted either with or without fragment annotation. If the fragment annotation via SIRIUS is used, the best rank will be choosen, its sumformula and fragment annotation will be retainend. 

In the future CSI:FingerID could potentially be used here to asses a vaild metabolite identification in addition to the sumformula and fragment annotation. 

Changes:
Add fragment annotation via SIRIUS for assay library generation

Add MetaboTargetedAssay: Stores precursor information and potential transitions for a metabolite (is parsed into TargetedExperiment later on.)

Remove converter_mode from SiriusAdapterAlgorithm (Parameter in the SiriusAdapter)

[FIX] SWATHQC issue, which leads to an exception because of an empty spectrum provided after peak picking.

Use "mid" in SIRIUSMSConverter to resolve multiple possible identifications from AccurateMassSearch, via SIRIUS (experimental feature - not yet clear how well the annotation works in these cases to validate the spectrum). The idea is, that SIRIUS may be ablte to asses the best matching identification based on fragment anntotation. 

Add tests for all cases.


src/openms/include/OpenMS/ANALYSIS/TARGETED/MetaboTargetedAssay.h
```diff
@@ -0,0 +1,93 @@
...
+    double precursor_rt;
+    double precursor_int;
+    int precursor_charge;
+    double transition_quality_score;
+    String compound_name;
+    String compound_adduct;
+    TargetedExperiment::Compound potential_cmp;
+    std::vector<ReactionMonitoringTransition> potential_rmts;
+
+    // method to extract a potential transition based on the ms/ms based of the highest intensity precursor or a consensus spectrum
```
timosachsenberg: adapt comment

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -132,157 +127,110 @@ class TOPPAssayGeneratorMetabo :
...
+  	string key;
+  	compare(string const &i): key(i){}
+  
+  	bool operator()(string const &i)
+  	{
+  		return (i == key);
+  	}
+  };
+
+  static bool compoundNameLess_(const OpenMS::MetaboTargetedAssay& i, const OpenMS::MetaboTargetedAssay& j)
```
timosachsenberg: you could potentiall use std::tie to define the comparison order

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -132,157 +127,110 @@ class TOPPAssayGeneratorMetabo :
...
+    {
+      return false;
+    }
   }
-
-  // precursor correction (highest intensity)
-  Int getHighestIntensityPeakInMZRange(double test_mz, const MSSpectrum& spectrum1, double tolerance, bool ppm)
+ 
+  // use std::unique with this function to retain the unique element with the highest precursor intensity
+  static bool compoundUnique_(const OpenMS::MetaboTargetedAssay& i, const OpenMS::MetaboTargetedAssay& j)
```
timosachsenberg: https://en.cppreference.com/w/cpp/algorithm/unique: Elements are compared using the given binary predicate p. The behavior is undefined if it is not an equivalence relation.

oliveralka: Yes, I will change that.

hroest:
please fix:

- http://cdash.openms.de/viewTest.php?onlyfailed&buildid=306297
- http://cdash.openms.de/viewTest.php?onlyfailed&buildid=306296

e.g.

Invalid parameter: Invalid output file extension 'AccurateMassSearch_1_output.tmp.tsv'. Valid file extensions are: 'mzTab'.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusMSConverter.h
```diff
@@ -71,6 +71,7 @@ namespace OpenMS
     String source_file;
     String source_format;
     std::vector<String> native_ids;
+    std::vector<String> mids;
     std::vector<String> scan_indices;
```
hroest: maybe a good idea to update `src/pyOpenMS/pxds/SiriusMSFile.pxd`

src/openms/include/OpenMS/ANALYSIS/TARGETED/MetaboTargetedAssay.h
```diff
@@ -0,0 +1,93 @@
...
+#include <OpenMS/ANALYSIS/TARGETED/TargetedExperiment.h>
+
+#include <OpenMS/ANALYSIS/ID/SiriusMSConverter.h> //SiriusMSFile
+
+#include <OpenMS/CONCEPT/Exception.h>
+
+namespace OpenMS
+{
+  /**
+  @brief This class provides methods for the extraction of targeted assays for metabolomics.
```
hroest: maybe a bit more comments on how this is supposed to be used would be useful

src/openms/include/OpenMS/ANALYSIS/TARGETED/MetaboTargetedAssay.h
```diff
@@ -0,0 +1,93 @@
...
+  */
+
+  class OPENMS_DLLAPI MetaboTargetedAssay
+  {
+    public:
+
+    double precursor_mz;
+    double precursor_rt;
+    double precursor_int;
+    int precursor_charge;
```
hroest: why is charge/intensity/mz required here if it is also availble in `TargetedExperiment::Compound` ?

src/openms/include/OpenMS/ANALYSIS/TARGETED/MetaboTargetedAssay.h
hroest: Also wrapping in pyOpenMS would be nice

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+                                                                                    const double& cosine_sim_threshold,
+                                                                                    const double& transition_threshold,
+                                                                                    const bool& method_consensus_spectrum,
+                                                                                    const bool& exclude_ms2_precursor,
+                                                                                    const unsigned int& file_counter)
+  {
+    int transition_group_counter = 0;
+    vector <MetaboTargetedAssay> v_mta;
+
+    for (auto it = feature_ms2_spectra_map.begin();
```
hroest: why not 
```suggestion
    for (auto& it : feature_ms2_spectra_map)
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+        String description("UNKNOWN"), sumformula("UNKNOWN"), adduct("UNKNOWN");
+        StringList v_description, v_sumformula, v_adduct;
+
+        double feature_rt;
+        const BaseFeature *min_distance_feature = it->first;
+        feature_rt = min_distance_feature->getRT();
+
+        // extract metadata from featureXML
+        if (!(min_distance_feature-> getPeptideIdentifications().empty()) && !(min_distance_feature->getPeptideIdentifications()[0].getHits().empty()))
```
hroest:
```suggestion
        auto metaboliteIdentifications = min_distance_feature-> getPeptideIdentifications();
        if (!(metaboliteIdentifications.empty()) && !(metaboliteIdentifications[0].getHits().empty()))
```

then its more clear downstream what you are working with and it makes more intuitive sense

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+        {
+          // accurate mass search may provide multiple possible Hits
+          // for heuristics use the identification with the smallest mz error (ppm)
+          double min_id_mz_error = std::numeric_limits<double>::max();
+          for (unsigned int j = 0; j != min_distance_feature->getPeptideIdentifications()[0].getHits().size(); ++j)
```
hroest:
```suggestion
          for (const auto& mhit : min_distance_feature->getPeptideIdentifications()[0].getHits())
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+      vector <size_t> index = it->second;
+
+      for (auto index_it = index.begin(); index_it != index.end(); ++index_it)
+      {
+        const MSSpectrum &spectrum = spectra[*index_it];
+
+        // check if MS2 spectrum is empty
+        if (spectrum.empty())
+        {
+          LOG_WARN << "Empty MSMS spectrum was provided. Please hava a look. Index: " << *index_it << std::endl;
```
hroest:
```suggestion
          LOG_WARN << "Empty MSMS spectrum was provided. Please manually investigate at index: " << *index_it << std::endl;
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+          continue;
+        }
+
+        const vector <Precursor> &precursor = spectrum.getPrecursors();
+
+        // get m/z and intensity of precursor
+        // only spectra with precursors are in the map, therefore no need to check for their presence
```
hroest: how do you know? how can you be sure about your input? what about checking and throw an exception?

```suggestion
        // only spectra with precursors are in the map, therefore no need to check for their presence
        if (precursor.empty()) throw ...
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+          highest_precursor_int = precursor_int;
+          highest_precursor_mz = precursor_mz;
+          highest_precursor_int_spectrum = spectrum;
+          highest_precursor_charge = precursor_charge;
+        }
+          transition_spectrum = highest_precursor_int_spectrum;
+      }
+
+      // if only one MS2 is available and the consensus method is used - jump right to the transition list calculation
+      // fallback: highest intensity precursor
```
hroest: I dont think this comments reflects well what happens below, can you adjust that? It seems you use the highest intensity spectrum, then find otehr spectra with high enough cosine angle and add all those together?

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+          double min_id_mz_error = std::numeric_limits<double>::max();
+          for (unsigned int j = 0; j != min_distance_feature->getPeptideIdentifications()[0].getHits().size(); ++j)
+          {
+            double current_id_mz_error = min_distance_feature->getPeptideIdentifications()[0].getHits()[j].getMetaValue("mz_error_ppm");
+            // compare the absolute error absolute error
+            if (abs(current_id_mz_error) < min_id_mz_error)
+            {
+              description = min_distance_feature->getPeptideIdentifications()[0].getHits()[j].getMetaValue("description");
+              sumformula = min_distance_feature->getPeptideIdentifications()[0].getHits()[j].getMetaValue("chemical_formula");
+              adduct = min_distance_feature->getPeptideIdentifications()[0].getHits()[j].getMetaValue("modifications");
```
hroest: modifications is not a great name for adducts - something to keep in mind...

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+      // sort intensity in MS2 spectrum to extract transitions
+      transition_spectrum.sortByIntensity(true);
+
+      // filter out the precursors if they are in the ms2 spectrum;
+      if (exclude_ms2_precursor)
+      {
+        for (auto spec_it = transition_spectrum.begin(); spec_it != transition_spectrum.end(); ++spec_it)
+        {
+          if (transition_spectrum.getPrecursors()[0].getMZ() == spec_it->getMZ())
```
hroest: why exact matching of floating point number?
```suggestion
          if ( abs(transition_spectrum.getPrecursors()[0].getMZ() - spec_it->getMZ()) < 1e-3)
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+          {
+            transition_spectrum.erase(spec_it);
+            break;
```
hroest: I would remove *all* peaks close to the precursor, not just the first one!

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+          rmt.setCompoundRef (String(transition_group_counter) + "_" + description + "_" + file_counter);
+          rmt.setNativeID (String(transition_group_counter)+ "_" + String(transition_counter)+ "_" + description + "_" + file_counter);
+
+          v_rmt.push_back(rmt);
+          transition_counter += 1;
+        }
+      }
+      transition_group_counter += 1;
+      MetaboTargetedAssay mta;
+      mta.precursor_mz = highest_precursor_mz;
```
hroest: it looks like this line and line 294 just store the same information

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+      double exact_mass_precursor = 0.0;
+      for (auto spec_it = transition_spectrum.begin();
+           spec_it != transition_spectrum.end();
+           ++spec_it)
+      {
+        int spec_index = spec_it - transition_spectrum.begin();
+        OpenMS::DataArrays::StringDataArray explanation_array = transition_spectrum.getStringDataArrays()[0];
+        if (explanation_array.getName() != "explanation")
+        {
+          LOG_WARN << "Fragment explanation was not found. Please check if you annotation works properly." << std::endl;
```
hroest:
```suggestion
          LOG_WARN << "Fragment explanation was not found. Please check if your annotation works properly." << std::endl;
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+      transition_spectrum.sortByIntensity(true);
+
+      // have to remove ms2 precursor peak before min/max
+      double exact_mass_precursor = 0.0;
+      for (auto spec_it = transition_spectrum.begin();
+           spec_it != transition_spectrum.end();
+           ++spec_it)
+      {
+        int spec_index = spec_it - transition_spectrum.begin();
+        OpenMS::DataArrays::StringDataArray explanation_array = transition_spectrum.getStringDataArrays()[0];
```
hroest: can you first check that the array is not empty (otherwise segfault)

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,498 @@
...
+                  .erase(transition_spectrum.getStringDataArrays()[0].begin() + spec_index);
+              transition_spectrum.getFloatDataArrays()[0]
+                  .erase(transition_spectrum.getFloatDataArrays()[0].begin() + spec_index);
+              break; // if last element have to break if not iterator will go out of range
+            }
+          }
+        }
+      }
+
+      // find max and min intensity peak
```
hroest: all of the following code looks duplicated  -> extract function?

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -136,259 +136,281 @@ namespace OpenMS
...
-        // no charge annotated - assume mono-charged
-        if (precursor_charge == 0)
-        {
-          precursor_charge = 1;
-          ++count_assume_mono;
-        }
-        // negative mode - make sure charges are < 0
-        if (p == IonSource::Polarity::NEGATIVE) { precursor_charge = -(std::abs(precursor_charge)); }
+          // sirius supports only single charged ions (+1; -1)
+          // if charge = 0, it will be allocted to +1; -1 depending on Polarity
```
hroest:
```suggestion
          // if charge = 0, it will be changed to +1; -1 depending on Polarity
```

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -136,259 +136,281 @@ namespace OpenMS
...
+            feature_charge = 1;
+            ++count_assume_mono;
+          }
+          // negative mode - make sure charges are < 0
+          if (p == IonSource::Polarity::NEGATIVE) { feature_charge = -(std::abs(feature_charge)); }
 
-        // extract collision energy
-        double collision = precursor[0].getActivationEnergy();
+          // get m/z and intensity of precursor != MS1 spectrum
+          double precursor_mz = precursor[0].getMZ();
```
hroest: can you check for empty precursor?

src/openms/include/OpenMS/ANALYSIS/TARGETED/MetaboTargetedAssay.h
oliveralka: Will add pyOpenMS wrapper and comments are on the way.

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -136,259 +136,281 @@ namespace OpenMS
...
-            os << ">parentmass " << precursor_mz << fixed << "\n";
-            cmpinfo.pmass = precursor_mz;
-          }
-  
-          if (!adducts.empty())
-          {
-            os << ">ionization " << ListUtils::concatenate(adducts, ',') << "\n";
-            cmpinfo.ionization = ListUtils::concatenate(adducts, ',');
-          }
+            // write internal unique .ms data as sirius input
```
hroest: this should maybe go to `FORMAT/DotMSFile.h` ?

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -499,47 +527,59 @@ namespace OpenMS
...
+
+          for (unsigned int j = 0; j != feature->getPeptideIdentifications()[0].getHits().size(); ++j)
+          {
            String adduct;
-           description = feature->getPeptideIdentifications()[0].getHits()[0].getMetaValue("description");
-           sumformula = feature->getPeptideIdentifications()[0].getHits()[0].getMetaValue("chemical_formula");
-           adduct = feature->getPeptideIdentifications()[0].getHits()[0].getMetaValue("modifications");
+           description = feature->getPeptideIdentifications()[0].getHits()[j].getMetaValue("description");
+           sumformula = feature->getPeptideIdentifications()[0].getHits()[j].getMetaValue("chemical_formula");
+           adduct = feature->getPeptideIdentifications()[0].getHits()[j].getMetaValue("modifications");
```
hroest: in the feature adducts are called "adducts", there theyare in the meta bvalue "modifications"

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -309,246 +261,377 @@ class TOPPAssayGeneratorMetabo :
...
-    KDTreeFeatureMaps fp_map_kd;
-    vector<FeatureMap> v_fp;
-    v_fp.push_back(feature_map);
-    fp_map_kd.addMaps(v_fp);
 
-    // read FeatureMap in KDTree for feature-precursor assignment
-    // only spectra with precursors are in the map - no need to check for presence of precursors
-    map<const BaseFeature*, std::vector<size_t> > feature_ms2_spectra_map = extractMetaInformation(spectra, fp_map_kd, precursor_mz_tol, precursor_rt_tol, ppm_prec);
+      // determine type of spectral data (profile or centroided)
+      SpectrumSettings::SpectrumType spectrum_type = spectra[0].getType();
```
hroest: check empty files (no spectra) here, otherwise segfault

hroest:
:-1: looks good overall, see my comments

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
oliveralka: Yes, you are right.
```
      // check if more than one MS/MS spectrum is available to use the consensus method
      // the MS/MS spectrum of the highest intensity precursor is used as reference and compared
      // to the other MS/MS spectrum of a specific feature.
      // if the cosine similarity is over the manually set threshold these are merged via SpectraMerger
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
oliveralka: Yes, this is currently the AccurateMassSearch featureXML implementation:

```
	<PeptideIdentification identification_run_ref="PI_0" score_type="" higher_score_better="true" significance_threshold="0" >
				<PeptideHit score="0" sequence="" charge="1">
					<UserParam type="stringList" name="identifier" value="[76738-62-0]"/>
					<UserParam type="stringList" name="description" value="[Paclobutrazol]"/>
					<UserParam type="string" name="modifications" value="M+H;1+"/>
					<UserParam type="string" name="chemical_formula" value="C15H20ClN3O"/>
					<UserParam type="float" name="mz_error_ppm" value="-1.604379800732327"/>
					<UserParam type="float" name="mz_error_Da" value="-4.719071027921018e-04"/>
				</PeptideHit>
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
oliveralka: That makes sense...

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
oliveralka: > -- or potentially even all peaks above the precursor (e.g. isotopes etc...)

This is done in the AssayGeneratorMetabo (preprocessing) - line 477
```
       // remove peaks form MS2 which are at a higher mz than the precursor + 10 ppm
        for (auto& peakmap_it : spectra)
        {
          MSSpectrum& spectrum = peakmap_it;
          if (spectrum.getMSLevel() == 1) 
          {
            continue;
          }
          else 
          {
            // if peak mz higher than precursor mz set intensity to zero
            double prec_mz = spectrum.getPrecursors()[0].getMZ();
            double mass_diff = Math::ppmToMass(10.0, prec_mz);
            for (auto& spec : spectrum)
            {
              if (spec.getMZ() > prec_mz + mass_diff)
              {
                spec.setIntensity(0);
              }
            }
            spectrum.erase(remove_if(spectrum.begin(),
                                     spectrum.end(),
                                     InIntensityRange<PeakMap::PeakType>(1,
                                                                         numeric_limits<PeakMap::PeakType::IntensityType>::max(),
                                                                         true)), spectrum.end());
          }
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
oliveralka: Yes, the precursor_mz is stored in the ReactionMonitoringTransition as well. 
It is not needed later on - will remove it. 

As far as I see it I could potentially remove precursor_mz, precursor_rt, precursor_charge.
The precursor_intensity, compound_name and the compound_adduct are later on used for sorting and unique filtering. 

What do you think?

oliveralka:
Thanks for taking a look - will fix the issues!

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
hroest: Yes, I suggest to remove duplicate entries since its confusing that its in two places -- and think of   potentially even move some info into Compound itself if you need it often

oliveralka:
@hroest 
I tried to get the python bindings running - it fails in pyopenms_create_cpp with the error:
May you point me in the right direction? 

```
ADDED __str__ method to AASequence
ADDED __str__ method to MassDecomposition
ADDED __str__ method to RANSACParam
ADDED __str__ method to NASequence
ADDED __str__ method to EmpiricalFormula
ADDED __str__ method to DataValue
ADDED __str__ method to MZTrafoModel
ADDED __str__ method to DataFilter
8 8
Autwrap has wrapped 82 classes, 726 methods and 11 enums
  File "create_cpp_extension.py", line 194, in <module>
    autowrap_include_dirs = doCythonCodeGeneration(modname, allDecl_mapping, instance_map, converters)
  File "create_cpp_extension.py", line 158, in doCythonCodeGeneration
    include_boost=False, include_numpy=True, allDecl=allDecl_mapping)
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/__init__.py", line 65, in generate_code
    gen.create_pyx_file(debug)
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/CodeGenerator.py", line 207, in create_pyx_file
    create_for(ResolvedClass, self.create_wrapper_for_class)
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/CodeGenerator.py", line 203, in create_for
    method(resolved)
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/CodeGenerator.py", line 443, in create_wrapper_for_class
    codes = self.create_wrapper_for_method(r_class, name, methods)
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/CodeGenerator.py", line 593, in create_wrapper_for_method
    code = self.create_wrapper_for_nonoverloaded_method(cdcl, py_name, methods[0])
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/CodeGenerator.py", line 757, in create_wrapper_for_nonoverloaded_method
    method)
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/CodeGenerator.py", line 637, in _create_fun_decl_and_input_conversion
    conv_code, call_as, cleanup = converter.input_conversion(t, n, arg_num)
  File "/usr/local/miniconda3/envs/py37/lib/python3.7/site-packages/autowrap/ConversionProvider.py", line 1337, in input_conversion
    "Recursion in std::vector<T> is not implemented for other STL methods and wrapped template arguments")
Exception: Recursion in std::vector<T> is not implemented for other STL methods and wrapped template arguments
make[3]: *** [src/pyOpenMS/CMakeFiles/pyopenms_create_cpp.dir/build.make:57: src/pyOpenMS/CMakeFiles/pyopenms_create_cpp] Error 1
```

timosachsenberg:
can you add the changed pxd here so I can take a look

oliveralka:
I will add structs for the libcpp_map[ x, vector[ y ] ] data structure, then it should work.

hroest:
> Exception: Recursion in std::vector<T> is not implemented for other STL methods and wrapped template arguments

that is the problem :-)

src/pyOpenMS/pxds/MetaboTargetedAssay.pxd
```diff
@@ -0,0 +1,35 @@
...
+
+    cdef cppclass MetaboTargetedAssay "OpenMS::MetaboTargetedAssay":
+
+       MetaboTargetedAssay() nogil except +
+       MetaboTargetedAssay(MetaboTargetedAssay) nogil except + #wrap-ignore
+
+
+
+       libcpp_vector[MetaboTargetedAssay] extractMetaboTargetedAssay(MSExperiment spectra,
+                                                                     libcpp_map[ BaseFeature,libcpp_vector[size_t] ] feature_ms2_spectra_map,
```
hroest: > libcpp_map[ BaseFeature,libcpp_vector[size_t] ]

that probably doesnt work, you may have to remove the wrapper here

oliveralka: tried to set it to FeatureMapping::FeatureToMs2Indices feature_ms2_index, but error persits. 
so it is probably that one: 
libcpp_vector[ libcpp_pair[ SiriusMSFile_CompoundInfo,MSSpectrum ] ]& v_cmp_spec

oliveralka:
rebuild jenkins

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,516 @@
...
+      // no peaks or all peaks have same intensity (single peak / noise)
+      if (min_int >= max_int)
+      {
+        continue;
+      }
+
+      vector <TargetedExperimentHelper::RetentionTime> v_cmp_rt;
+      TargetedExperimentHelper::RetentionTime cmp_rt;
+      cmp_rt.setRT(feature_rt);
+      v_cmp_rt.push_back(cmp_rt);
```
hroest:
```suggestion
      v_cmp_rt.push_back(std::move(cmp_rt));
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,516 @@
...
+      cmp.rts = v_cmp_rt;
```
hroest:
```suggestion
      cmp.rts = std::move(v_cmp_rt);
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,516 @@
...
+          rmt.setProductMZ(current_mz);
+          rmt.setLibraryIntensity(rel_int);
+
+          rmt.setCompoundRef(String(transition_group_counter) + "_" + description + "_" + file_counter);
+          rmt.setNativeID(String(transition_group_counter) + "_" + String(transition_counter) + "_" + description + "_" +
+                          file_counter);
+
+          rmt.setMetaValue("annotation", DataValue(current_explanation));
+
+          v_rmt.push_back(rmt);
```
hroest:
```suggestion
          v_rmt.push_back(std::move(rmt));
```

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -0,0 +1,516 @@
...
+      }
+
+      transition_group_counter += 1;
+      MetaboTargetedAssay mta;
+      mta.precursor_int = 0;
+      mta.compound_name = description;
+      mta.compound_adduct = adduct;
+      mta.potential_cmp = cmp;
+      mta.potential_rmts = v_rmt;
+      v_mta.push_back(mta);
```
hroest:
```suggestion
      v_mta.push_back(std::move(mta));
```

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -309,246 +214,398 @@ class TOPPAssayGeneratorMetabo :
...
-        merger.setParameters(p);
-
-        // all MS spectra should have the same precursor
-        merger.mergeSpectraPrecursors(exp);
-
-        // check if all precursors have been merged if not use highest intensity precursor
-        if (exp.getSpectra().size() < 2)
+
+        // pair compoundInfo and fragment annotation msspectrum (using the mid)
+        for (auto cmp : v_cmpinfo)
```
hroest:
```suggestion
        for (const auto& cmp : v_cmpinfo)
```
unless you want to copy?

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -309,246 +214,398 @@ class TOPPAssayGeneratorMetabo :
...
         {
-          transition_spectrum = exp.getSpectra()[0];
+          for (auto spec_fa : annotated_spectra)
```
hroest:
```suggestion
          for (const auto& spec_fa : annotated_spectra)
```
unless you want to copy?

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -309,246 +214,398 @@ class TOPPAssayGeneratorMetabo :
...
+      }
+      
+      // append potential transitions of one file to vector of all files
+      v_mta.insert(v_mta.end(), tmp_mta.begin(), tmp_mta.end());
+      
+    } // end iteration over all files
+
+    // use first rank based on precursor intensity
+    std::map< std::pair <String,String>, MetaboTargetedAssay > map_mta;
+    for (auto it : v_mta)
```
hroest:
```suggestion
    for (const auto& it : v_mta)
```

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -309,246 +214,398 @@ class TOPPAssayGeneratorMetabo :
...
+                                                             out_csifingerid,
+                                                             sirius_algo);
+  
+        // sort vector path list
+        std::sort(subdirs.begin(), subdirs.end(), extractAndCompareScanIndexLess_);
+        LOG_DEBUG << subdirs.size() << " spectra were annotated using SIRIUS." << std::endl;
+  
+        // get Sirius FragmentAnnotion from subdirs
+        vector<MSSpectrum> annotated_spectra;
+        for (auto subdir : subdirs)
```
hroest:
```suggestion
        for (const auto& subdir : subdirs)
```

PR #4038, opened 2019-04-03
Addendum to PR #4020: use proton mass for charges

hendrikweisser:
Additional testing confirmed that using the theoretically correct mass value (of a proton, not intact hydrogen) didn't lead to worse results after all - so I've made the corresponding change here.


hroest:
related to #4020

PR #4039, opened 2019-04-04
[ADD] MSstatsTMT support in MSstatsConverter

oliveralka:
Split export function to LFQ and ISO. 
Add tests.


src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -65,16 +63,89 @@ namespace OpenMS
...
+        static void checkConditionLFQ_(const ExperimentalDesign::SampleSection& sampleSection, const String& bioreplicate, const String& condition);
+
+        /*
+         *  @Brief: Internal function to check if MSstats_BioReplicate, MSstats_Condition and MSstats_Mixture in Experimental Design
+         */
+        static void checkConditionISO_(const ExperimentalDesign::SampleSection sampleSection, const String& bioreplicate, const String& condition, const String& mixture);
+
+        /*
+         *  MSstats treats runs differently than OpenMS. In MSstats, runs are an enumeration of (SpectraFilePath, Fraction)
+         *  In OpenMS, a run is split into multiple fractions.
```
timosachsenberg: run or fraction group?

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -65,16 +63,89 @@ namespace OpenMS
...
+              run_map[tpl] = run_counter++;
+            }
+          }
+        }
+
+        bool checkUnorderedContent_(const std::vector< String> &first, const std::vector< String > &second)
+        {
+          const std::set< String > lhs(first.begin(), first.end());
+          const std::set< String > rhs(second.begin(), second.end());
+          return lhs.size() == rhs.size()
```
timosachsenberg: would a simple 
return lhs == rhs 
be enough?

hroest:
do you need a 2 MB testfile?

oliveralka:
I can try to make the testfile it smaller.

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
oliveralka: Should be enough but I am not sure what the intention of Lukas was here.

PR #4041, opened 2019-04-05
[FIX] move DPosition

jpfeuffer:
Fixes segfault on windows nightlies (tested on the machine itself).
Change in test unrelated. will change back.

Not sure if this is the best solution and why it occurred at all. Shouldn't this be what the
default move is doing?
I'll have another look at it later.

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/DPosition.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/DPosition.h
@@ -109,7 +109,10 @@ public:
     }
 
     /// Move constructor
-    DPosition(DPosition&& rhs) noexcept = default;
+	DPosition(DPosition&& rhs) noexcept
+	{
+		std::move(std::begin(rhs.coordinate_), std::end(rhs.coordinate_), &coordinate_[0]);
+	}
 
     /// Constructor only for DPosition<2> that takes two Coordinates.
     DPosition(CoordinateType x, CoordinateType y)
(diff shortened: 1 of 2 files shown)


cbielow:
maybe replacing `CoordinateType coordinate_[D];` with std::array will resolve it. Would also safe some of the custom operator implementations.
Why do we implement move here anyways? there is no benefit (for the types we use it for).

src/openms/include/OpenMS/DATASTRUCTURES/DPosition.h
```diff
@@ -109,7 +109,10 @@ namespace OpenMS
     }
 
     /// Move constructor
-    DPosition(DPosition&& rhs) noexcept = default;
+	DPosition(DPosition&& rhs) noexcept
+	{
+		std::move(std::begin(rhs.coordinate_), std::end(rhs.coordinate_), &coordinate_[0]);
```
hroest:
```suggestion
		// NOTE: do not change this before testing with nightly Windows builds ( = default causes segfault)
		std::move(std::begin(rhs.coordinate_), std::end(rhs.coordinate_), &coordinate_[0]);
```

hroest:
>  Why do we implement move here anyways? there is no benefit (for the types we use it for).

its implemented so that we can move derived types for which it actually does matter. But I agree, we might as well copy, but there *needs* to be a move operator, otherwise derived types wont have one by default. See #3758

PR #4044, opened 2019-04-08
[FIX] OPXL: some refactoring to remove broken meta values from parsed xquest.xml files

enetz:


src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -997,5 +997,6 @@ using namespace OpenMS;
     pep_indexing.run(fasta_db, protein_ids, peptide_ids);
 
     OPXLHelper::addProteinPositionMetaValues(peptide_ids);
+    OPXLHelper::addXLTargetDecoyMV(peptide_ids);
```
hroest: no addBetaAccessions here?

enetz: right now addProteinPositionMetaValues also adds beta accessions. If I remove that functionality from there and call both functions, I will have the same result but will have to loop over all IDs twice. Not sure if it is worth it currently.

src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
```diff
@@ -195,6 +195,18 @@ namespace OpenMS
        */
       static void addProteinPositionMetaValues(std::vector< PeptideIdentification > & peptide_ids);
 
+      /**
+       * @brief adds xl_target_decoy MetvaValue that combines alpha and beta target_decoy info
```
timosachsenberg: MetvaValue typo

src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
timosachsenberg: Hmm sounds a bit confusing. Is this properly documented?

enetz:
rebuild jenkins

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1070,7 +1052,64 @@ namespace OpenMS
...
         }
+      }
+    }
+  }
+
+  void OPXLHelper::addBetaAccessions(std::vector< PeptideIdentification > & peptide_ids)
+  {
+    for (PeptideIdentification& id : peptide_ids)
+    {
+      PeptideHit& ph_alpha = id.getHits()[0];
```
hroest: what happens if  its empty?

```suggestion
      if (id.getHits().empty()) continue;
      PeptideHit& ph_alpha = id.getHits()[0];
```

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1070,7 +1052,64 @@ namespace OpenMS
...
+        ph_alpha.setMetaValue("accessions_beta", "-");
+      }
+    }
+  }
+
+  void OPXLHelper::addXLTargetDecoyMV(std::vector< PeptideIdentification > & peptide_ids)
+  {
+    for (PeptideIdentification& id : peptide_ids)
+    {
+      PeptideHit& ph_alpha = id.getHits()[0];
```
hroest: what happens if  its empty?

```suggestion
      if (id.getHits().empty()) continue;
      PeptideHit& ph_alpha = id.getHits()[0];
```

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1070,7 +1052,64 @@ namespace OpenMS
...
+        PeptideHit& ph_beta = id.getHits()[1];
+        String prot2_accessions;
+
+        const std::vector<PeptideEvidence> pevs_beta = ph_beta.getPeptideEvidences();
+        for (std::vector<PeptideEvidence>::const_iterator pev = pevs_beta.begin(); pev != pevs_beta.end(); ++pev)
+        {
+          prot2_accessions = prot2_accessions + "," + pev->getProteinAccession();
+        }
 
+        prot2_accessions = prot2_accessions.suffix(prot2_accessions.size()-1);
```
hroest: this will throw an exception if the string is empty (e.g. no pevs_beta) - maybe check for that?

hroest:
looks good, please add some unit tests and also check for the edge cases wher the hits or the protein accessions are empty

hroest:
looks good, please add some unit tests and also check for the edge cases wher the hits or the protein accessions are empty

enetz:
I added tests and checks for empty vectors with the latest commits. The failed tests look unrelated to me.
This should be ready to merge.

hroest:
> The failed tests look unrelated to me.

I agree

PR #4045, opened 2019-04-08
[FIX] Fixed area under the curve for single specturm chromatogram in …

DennisGoldfarb:
Problem: When an MSExperiment with a single spectrum was the input into FeatureFinderMultiplex, a chromatogram with a single point was created and would fail to get the area under the curve. At least 2 points were required to calculate the area under the curve using the PeakIntegrator class.

Solution: Added a check for a single point chromatogram. If there's only 1, then we just use the intensity of that peak instead of integrating.


hroest:
- can you remove the contrib changes?
- can you add a short test somewhere?

lars20070:
+1 for the changes in TOPP test 9. It's a real, new feature pair that is found.

DennisGoldfarb:
Actually, Lars should the line be intensity_sum = the peak, or should it be +=? It looks like we're summing over something

DennisGoldfarb:
@lars20070

PR #4046, opened 2019-04-09
[FIX] Silently install Redists

jpfeuffer:
Seems like Microsoft loves to change their parameters for silent installs:
http://asawicki.info/news_1597_installing_visual_c_redistributable_package_from_command_line.html


jpfeuffer:
Merge and check in the nightlies? @timosachsenberg

PR #4047, opened 2019-04-09
columns for mz and rt for individual feature handles to mzTab

### LINKED ISSUE #3705: add optional individual mz and RT columns in MzTabExporter output
In addition to standard `mzTab`, the `MzTabExporter` tool writes a number of very useful optional `opt_global_*` columns. Hopefully, some of them will become part of the `mzTab` standard in the future.

Individual peptide intensities are reported as `peptide_abundance_study_variable[1-n]`. Could we also report individual `mz` and `RT` values as `opt_global_*` columns in the `PEP` section [here](https://github.com/OpenMS/OpenMS/blob/develop/src/topp/MzTabExporter.cpp#L888)?

```
opt_global_mass_to_charge_study_variable[1-n]
opt_global_retention_time_study_variable[1-n]
```

One possible use case are `mzTab` based [QC reports](https://github.com/ProteomicsScripts/ProteomicsScripts/tree/master/mzTab2report). For example, in SILAC experiments one expects the heavy-labelled peptide to elude slightly earlier than its light partner. The

```opt_global_retention_time_study_variable[1] - opt_global_retention_time_study_variable[2]```

distribution should peak at negative retention times and would be a helpful QC plot.

lars20070:
When exporting `consensusXML` to `mzTab`, the individual peptide intensities for each study variable are exported. With this PR, the individual `m/z` and `RT` are now exported as well.

Resolves #3705.

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/MzTab.h
+++ b/src/openms/include/OpenMS/FORMAT/MzTab.h
@@ -956,6 +956,7 @@ public:
       const String & filename,
       const bool export_unidentified_features,
       const bool export_unassigned_ids,
+      const bool export_subfeatures,
       String title = "ConsensusMap export from OpenMS");
 
   protected:
--- a/src/openms/source/FORMAT/MzTab.cpp
+++ b/src/openms/source/FORMAT/MzTab.cpp
@@ -2615,6 +2615,7 @@ Not sure how to handle these:
     const String& filename, 
     const bool export_unidentified_features,
     const bool export_unassigned_ids,
+    const bool export_subfeatures,
     String title)
   {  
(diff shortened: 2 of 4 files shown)


timosachsenberg:
can you add an option so adding these optional columns can be switched on/off?

lars20070:
@timosachsenberg As in `fixed_modifications` in our search engine adapters. Correct?

hroest:
@timosachsenberg @lars20070 this PR broke our tests, please do not merge when the tests fail. Breaking travis will lead to all other PR breaking and leads to uncertainty whether the PR itself is at fault or because the current develop is broken.

Issue #4049, opened 2019-04-09
Unintended copies due to C++ auto loop


hroest:
a recent PR got me thinking that its very easy to write

```
for (auto var : arr)
```

instead of the intended

```
for (const auto& var : arr)
```

which leads C++ to copy every element of the array `arr` instead of an efficient iteration. I checked in OpenMS and indeed there is a bunch of instances. We should

- fix these instances
- add a style check so that new code gets a warning if it attempts to use `for (auto var : arr)`

hroest:
we should probably run this script regularly (nightly) or even better as part of our travis run with the style checking. @cbielow @jpfeuffer would it be easy to add a regex like this one and ensure that it returns zero results:

```
$ grep -R for src/openms | grep auto | grep -v begin | grep ":" | grep -v '&' | grep -v '//' | grep "for[ ]*(" 
```

of course we would have to first weed out and annotate false positives

jpfeuffer:
Yes that would certainly be possible. There are multiple possibilities:
1) We report every single file as part of our style build (on travis). Basically we create a shell script and create a test for each of the files.
2) We report a single test as part of the test suites. Then we can report it in Ctest.xml format or plaintext somewhere in the Jenkins/Travis console or CDash.
3) We check if one of our style checkers (cpplint, codacy) (potentially in newer versions) support checking for that.
4) We use a new style checker since I know that my Clang-Tidy complains about it.

hroest:
I would prefer 3/4 since my regex is not perfect and relies on some assumptions (e.g. if `begin` is on the same line its more likely a construct of `for (auto it = v.begin(); it != v.end(); it++)` which should not be `const auto&` so anything that does some real static analysis would be much better. However in the meantime, I think number 2 is the best compromise, just run the regex once over all the files and report that in travis. What do you think?

jpfeuffer:
I started on a job for clang-tidy. It is relatively slow and takes 7h with also clang-analyzer checks enabled. So it is nothing for PRs right now. We could export a compilation database from CMake, cache it and then use the clang-tidy-diff tool, that only checks the files that actually diff'ed in a PR. But that will be some work.
So, yes maybe for now we could include at least your regex on travis as a single test. I will look at that next. In the meanwhile maybe the results of clang-tidy are interesting:
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_CI_jobs/job/openms_ci_clangtidy/6/clang-tidy/#categoryContent

[...]

PR #4050, opened 2019-04-09
[FIX] fix unintended copy of loop variables

hroest:
fixes several instances of 

```
for (auto var : arr)
```

instead of the intended

```
for (const auto& var : arr)
```

as discussed in #4049 but only for now. Basically replacing the copy with a const-ref should have no side effects since it ensures that the const-ref is not modified. 

Found with
```
$ grep -R for src/openms | grep auto | grep -v begin | grep ":" | grep -v '&' | grep -v '//' | grep "for[ ]*(" 
```

### DIFF
--- a/src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
+++ b/src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
@@ -33,6 +33,7 @@
 // --------------------------------------------------------------------------
 
 #include <OpenMS/ANALYSIS/ID/FalseDiscoveryRate.h>
+
 #include <OpenMS/CONCEPT/LogStream.h>
 
 // #define FALSE_DISCOVERY_RATE_DEBUG
@@ -485,17 +486,17 @@ namespace OpenMS
       it->setHigherScoreBetter(false);
       const vector<ProteinHit>& old_hits = it->getHits();
       vector<ProteinHit> new_hits;
-      for (auto hit : old_hits)
+      for (auto hit : old_hits) // NOTE: performs copy
       {
         // Add decoy proteins only if add_decoy_proteins is set
         if (add_decoy_proteins || hit.getMetaValue("target_decoy") != "decoy")
         {
           hit.setMetaValue(score_type, hit.getScore());
           hit.setScore(score_to_fdr[hit.getScore()]);
-          new_hits.push_back(hit);
+          new_hits.push_back(std::move(hit));
         }
       }
-      it->setHits(new_hits);
+      it->setHits(std::move(new_hits));
     }
 
     return;
@@ -625,7 +626,7 @@ namespace OpenMS
     {
       vector<IdentificationData::QueryMatchRef> best_matches =
         id_data.getBestMatchPerQuery(score_ref);
-      for (auto match_ref : best_matches)
+      for (auto match_ref : best_matches) // NOTE: performs copy, should not be necessary?
       {
         handleQueryMatch_(match_ref, score_ref, target_scores, decoy_scores,
                           molecule_to_decoy, match_to_score);
--- a/src/openms/source/ANALYSIS/RNPXL/RNPxlReport.cpp
+++ b/src/openms/source/ANALYSIS/RNPXL/RNPxlReport.cpp
@@ -126,7 +126,7 @@ namespace OpenMS
     for (PeakMap::ConstIterator s_it = spectra.begin(); s_it != spectra.end(); ++s_it)
     {
       int scan_index = s_it - spectra.begin();
-     std::vector<Precursor> precursor = s_it->getPrecursors();
+      std::vector<Precursor> precursor = s_it->getPrecursors();
 
       // there should only one precursor and MS2 should contain at least a few peaks to be considered (e.g. at least for every AA in the peptide)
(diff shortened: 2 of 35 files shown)


src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -485,17 +486,17 @@ namespace OpenMS
       it->setHigherScoreBetter(false);
       const vector<ProteinHit>& old_hits = it->getHits();
       vector<ProteinHit> new_hits;
-      for (auto hit : old_hits)
+      for (auto hit : old_hits) // NOTE: performs copy
```
hroest: I think here the copy is intended since we dont want to change old_hits? @cbielow

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -625,7 +626,7 @@ namespace OpenMS
     {
       vector<IdentificationData::QueryMatchRef> best_matches =
         id_data.getBestMatchPerQuery(score_ref);
-      for (auto match_ref : best_matches)
+      for (auto match_ref : best_matches) // NOTE: performs copy, should not be necessary?
```
hroest: @cbielow looking at `handleQueryMatch_` I think that `match_ref` should actually be `const` and that we should not have a copy here. what do you think?

src/openms/source/ANALYSIS/RNPXL/RNPxlReport.cpp
```diff
@@ -270,7 +270,7 @@ namespace OpenMS
             }
             else // place it anywhere
             {
-              for (auto a : aa)
+              for (auto a : aa) // NOTE: performs copy, this cannot possibly work!
```
hroest: @timosachsenberg I dont think this can possibly work, can you confirm?  I think this should be a non-const ref, right?

```suggestion
              for (auto& a : aa) // NOTE: performs copy, this cannot possibly work!
```

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -644,7 +644,7 @@ namespace OpenMS
         s.substitute("ms_run[","").substitute("]","");
         vector<String> ms_run;
         s.split(',', ms_run);
-        for (auto a : ms_run)
+        for (auto& a : ms_run)
         {
           a.trim();
           mz_tab_metadata.assay[n].ms_run_ref.push_back(a.toInt());
```
hroest: @timosachsenberg this seems reasonable since `ms_run` is a temporary variable, so modifying it should not have any side effects.

hroest:
preventing unnecessary copies could speedup some code...

src/utils/RNPxlSearch.cpp
```diff
@@ -1639,7 +1639,7 @@ class RNPxlSearch :
     // read list of nucleotides that can directly cross-link
     // these are responsible for shifted fragment ions. Their fragment adducts thus determine which shifts will be observed on b-,a-,y-ions
     String can_cross_link = getStringOption_("RNPxl:can_cross_link");
-    for (auto c : can_cross_link) { can_xl_.insert(c); }
+    for (const auto& c : can_cross_link) { can_xl_.insert(c); }
```
cbielow: using a reference for a char is probably slower than just copying the char, but who knows what the compiler will do :)

cbielow:
great catch! if possible, we should run the grep-line regularly, or did you find a sanitizer which can do this?

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -485,17 +486,17 @@ namespace OpenMS
...
         if (add_decoy_proteins || hit.getMetaValue("target_decoy") != "decoy")
         {
           hit.setMetaValue(score_type, hit.getScore());
           hit.setScore(score_to_fdr[hit.getScore()]);
-          new_hits.push_back(hit);
+          new_hits.push_back(std::move(hit));
         }
       }
-      it->setHits(new_hits);
+      it->setHits(std::move(new_hits));
```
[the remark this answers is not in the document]
hroest: I agree, it does not help right now and doesnt hurt, but it will be better once we have an implementation of `void setHits(std::vector<ProteinHit>&& hits);`. Also I think it makes the meaning clear but I agree your solution is better.

src/utils/RNPxlSearch.cpp
hroest: yes, but again the meaning is clearer and the compiler can figure it out. There is probably anyways a better way of doing this with `insert` and iterators.

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
hroest:
```suggestion
      for (auto match_ref : best_matches) // NOTE: performs copy, should not be necessary? TODO: change!
```

hroest:
> great catch! if possible, we should run the grep-line regularly, or did you find a sanitizer which can do this?

lets discuss long term solutions over here: https://github.com/OpenMS/OpenMS/issues/4049

hroest:
merge?

PR #4052, opened 2019-04-10
update gradient only for last layer

### LINKED ISSUE #3012: TOPPView resets colours when new layer is opened
`TOPPView` allows to change the colour of `featureXML` and `consensusXML` layers. When an additional layer is opened, the colours of the *previous layers are reset to default*.

Previously, this was not the case and the colours of the layers remained unchanged. Can we go back to the old behaviour?

lars20070:
Update the gradient only for last layer.
Fixes #3012

### DIFF
--- a/src/openms_gui/source/VISUAL/Spectrum2DCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum2DCanvas.cpp
@@ -1072,9 +1072,13 @@ namespace OpenMS
     {
       gradient_str = linear_gradient_.toString();
     }
+    if (layers_.empty()) return;
     layers_[layers_.size()-1].param.setValue("dot:gradient", gradient_str);
-    recalculateDotGradient_(layers_.size()-1);
-    
+    for (Size i = 0; i < layers_.size(); ++i)
+    {
+      recalculateDotGradient_(i);
+    }
+
     SpectrumCanvas::intensityModeChange_();
   }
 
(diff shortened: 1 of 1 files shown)


src/openms_gui/source/VISUAL/Spectrum2DCanvas.cpp
```diff
@@ -1072,11 +1072,9 @@ namespace OpenMS
     {
       gradient_str = linear_gradient_.toString();
     }
-    for (Size i = 0; i < layers_.size(); ++i)
-    {
-      layers_[i].param.setValue("dot:gradient", gradient_str);
-      recalculateDotGradient_(i);
-    }
+    layers_[layers_.size()-1].param.setValue("dot:gradient", gradient_str);
+    recalculateDotGradient_(layers_.size()-1);
```
enetz: @lars20070 the `recalculateDotGradient_` function updates the gradients to a common maximal and minimal value and ensures that the same color for the same gradient has the same interpretation.
You could try to put this one line back in the loop. This will update the gradient ranges and might change the colors a bit. But it should keep the selected color gradients for each layer as long as you don't reset the `dot:gradient` param.

src/openms_gui/source/VISUAL/Spectrum2DCanvas.cpp
```diff
@@ -1072,11 +1072,9 @@ namespace OpenMS
     {
       gradient_str = linear_gradient_.toString();
     }
-    for (Size i = 0; i < layers_.size(); ++i)
-    {
-      layers_[i].param.setValue("dot:gradient", gradient_str);
-      recalculateDotGradient_(i);
-    }
+    layers_[layers_.size()-1].param.setValue("dot:gradient", gradient_str);
```
hroest: @lars20070 please add a check for empty layers, otherwise we have a segfault here
```suggestion
    if (layers_.empty()) return;
    layers_[layers_.size()-1].param.setValue("dot:gradient", gradient_str);
```

enetz:
:+1:

PR #4053, opened 2019-04-11
minor

cbielow:
some doc fixes and minor stuff


src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -462,35 +454,28 @@ void SimpleSearchEngineAlgorithm::postProcessHits_(const PeakMap& exp,
...
+          }
         }
 
         // skip peptides that have already been processed
         if (already_processed) { continue; }
 
-#ifdef _OPENMP
-#pragma omp atomic
-#endif
         ++count_peptides;
```
hroest: for sanity maybe we can annote that this is atomic?

```suggestion
        ++count_peptides; // atomic update
```

hroest:
:+1:

PR #4054, opened 2019-04-11
spectra_ref column for PEP section in mzTab (MzTabExporter)

lars20070:
When exporting `consensusXML` to `mzTab` with `MzTabExporter`, the `spectra_ref` column (containing `mzML` file name and spectral index) was missing. This PR fills in this information.


src/openms/include/OpenMS/FORMAT/MzTab.h
```diff
@@ -944,7 +944,9 @@ namespace OpenMS
         const std::vector<ProteinIdentification>& prot_ids,
```
hroest: please add docu that explains the different parameters, especially what the maps do

hroest:
please merge into `develop` and update with documentation

src/openms/include/OpenMS/FORMAT/MzTab.h
lars20070: Just FYI, the helper maps are not new. But now just exposed.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/MzTab.cpp#L1775
I will add further docu.

lars20070:
@timosachsenberg Can you please check the extended docu?

lars20070:
For clarification, only identified consensus features report `spectra_ref` in the `PEP` section, i.e. the file and spectrum index corresponding to that PSM. Consensus features without ID report `null` in the `spectra_ref` column.

PR #4055, opened 2019-04-11
ropenms script

Leon-Bichmann:
We tested the new R library "reticulate" and found it very usefull to applying pyopenms tools within R.

Here is an R script that shows some basic usage of these functionalitys in R.

### DIFF
--- a/share/OpenMS/SCRIPTS/ropenms.R
+++ b/share/OpenMS/SCRIPTS/ropenms.R
@@ -5,6 +5,8 @@
 ### if not installed:
 ### - install pyopenms (https://pyopenms.readthedocs.io/en/latest/installation.html)
 ###   make sure R is using the same python environment as your pyopenms installation
+###   eg. use_python("/usr/local/miniconda3/envs/py37/bin/python")
+###   or  Sys.setenv(RETICULATE_PYTHON = "/usr/local/miniconda3/envs/py37/bin/python")
 ### - install reticulate (https://rstudio.github.io/reticulate/)
 ###   eg. install.packages('reticulate')
 
@@ -17,7 +19,7 @@ ropenms=import("pyopenms", convert = FALSE)
 
 f="/OpenMS/OpenMS/share/OpenMS/examples/BSA/BSA1_OMSSA.idXML"
 
-idXML=ropenms$idXMLFile()
+idXML=ropenms$IdXMLFile()
 
 pepids=r_to_py(list())
 protids=r_to_py(list())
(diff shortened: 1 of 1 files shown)


hroest:
This looks great and really useful - would you be willing to contribute this also to the PyOpenMS documentation over here https://github.com/OpenMS/pyopenms-extra/tree/master/docs ? Maybe we can add a page about R in there?

Leon-Bichmann:
Ok sure, I'll give it a try!

share/OpenMS/SCRIPTS/ropenms.R
```diff
@@ -0,0 +1,91 @@
+library("reticulate")
+ropenms=import("pyopenms", convert = FALSE)
```
cbielow: can you document on how to fulfill this requirement? i.e. what needs to be build and what additional dependencies do you need?

Leon-Bichmann:
@oliveralka, can you add a line how to deal with the python path problem which @tjeerdijk experienced

Leon-Bichmann:
@oliveralka, can you add a line how to deal with the python path problem which @tjeerdijk experienced

share/OpenMS/SCRIPTS/ropenms.R
Leon-Bichmann: @oliveralka, can you add a line how to deal with the python path problem which @tjeerdijk experienced

hroest: maybe it would make sense to test for PyOpenMS presence and if not present, output a clear error message that includes Python interpreter used and how the user can install pyOpenMS?

oliveralka:
@Leon-Bichmann 
For me this worked:
```
library(reticulate)
# specify path to python
use_python("/usr/local/miniconda3/envs/py37/bin/python")
```
For Tjeerd we had to set the PATH first and then call the library:
```
# set PATH to python
Sys.setenv(RETICULATE_PYTHON = PATH)
library(reticulate)
```

share/OpenMS/SCRIPTS/ropenms.R
```diff
@@ -5,6 +5,8 @@
 ### if not installed:
 ### - install pyopenms (https://pyopenms.readthedocs.io/en/latest/installation.html)
 ###   make sure R is using the same python environment as your pyopenms installation
+###   eg. use_python("/usr/local/miniconda3/envs/py37/bin/python")
+###   or  Sys.setenv(RETICULATE_PYTHON = "/usr/local/miniconda3/envs/py37/bin/python")
```
oliveralka: Here it is important to set the PATH before calling the library(reticulate) - if it is done otherwise it won't work.

use_python can be used after calling the library (but does not seem to work in some cases)

PR #4060, opened 2019-04-12
[Feature] CometAdapter search for Protein N- and C-term mods

enetz:
includes

- reading in Protein N- and C-term mods from Unimod database

- setting up Comet parameter file to search for Protein N- and C-term mods

- correctly parsing Protein N- and C-term modifications from pepXML and writing that out in idXML

- a `CometAdapter` test searching for variable Protein and Peptide N-term mods

Except one `IDFileConverter_16` test, everything went well.
That test was previously reading in a Protein N-term mod as a Peptide N-term mod.


timosachsenberg:
@cbielow would make sense to query for terminal modified peptides if the modification is protein terminal and discard internal matches

src/openms/include/OpenMS/FORMAT/PepXMLFile.h
```diff
@@ -159,6 +159,7 @@ namespace OpenMS
       bool variable;
       String description;
       String terminus;
+      String protein_terminus; // "y" if protein terminus, "n" if peptide terminus
```
hroest: why not bool?

```suggestion
      bool protein_terminus; 
```

src/openms/source/FORMAT/PepXMLFile.cpp
```diff
@@ -1202,14 +1202,24 @@ namespace OpenMS
...
         aa_mod.terminus = String(attributeAsString_(attributes, "terminus")).toLower();
-        if (aa_mod.terminus == "n")
+        // "y" if protein terminus, "n" if peptide terminus
+        aa_mod.protein_terminus = String(attributeAsString_(attributes, "protein_terminus")).toLower() == "y";
+        if (aa_mod.terminus == "n" && !aa_mod.protein_terminus)
         {
           term_spec = ResidueModification::N_TERM;
         }
-        else if (aa_mod.terminus == "c")
+        else if (aa_mod.terminus == "n" && aa_mod.protein_terminus)
```
cbielow: `aa_mod.terminus == "n"` is duplicated. Can you make the structure a bit clearer by using an internal if/else. same for C-terminus.

PR #4065, opened 2019-04-16
[FIX] PQP & OSW SQLite bugs

grosenberger:
This PR tries to fix two small bugs with PQP and OSW files.

PQP: For some reason, String MetaValues could not be converted to integers anymore.
OSW: ``left join`` resulted in a SQL error. If the gene table is present, we should always link them, right?

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
@@ -343,7 +343,7 @@ namespace OpenMS
 
         if ((String)feature_it.getMetaValue("id_target_num_transitions") != "")
         {
-          int id_target_num_transitions = feature_it.getMetaValue("id_target_num_transitions").toString().toInt();
+          int id_target_num_transitions = std::stoi(feature_it.getMetaValue("id_target_num_transitions").toString(),nullptr);
 
           for (int i = 0; i < id_target_num_transitions; ++i)
           {
@@ -393,7 +393,7 @@ namespace OpenMS
 
         if ((String)feature_it.getMetaValue("id_decoy_num_transitions") != "")
         {
-          int id_decoy_num_transitions = feature_it.getMetaValue("id_decoy_num_transitions").toString().toInt();
+          int id_decoy_num_transitions = std::stoi(feature_it.getMetaValue("id_decoy_num_transitions").toString(),nullptr);
 
           for (int i = 0; i < id_decoy_num_transitions; ++i)
           {
--- a/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
@@ -174,23 +174,11 @@ namespace OpenMS
 
   std::vector<String> OpenSwathOSWWriter::getSeparateScore(const Feature& feature, std::string score_name) const
   {
-    std::vector<String> tmp_separated_scores, separated_scores;
+    std::vector<String> separated_scores;
 
     if (!feature.getMetaValue(score_name).isEmpty())
     {
-      tmp_separated_scores = ListUtils::create<String>((String)feature.getMetaValue(score_name),';');
-    }
-
-    for (Size i = 0; i < tmp_separated_scores.size(); ++i)
-    {
-      if (tmp_separated_scores[i] == "")
-      {
-        separated_scores.push_back("NULL");
-      }
-      else
-      {
-        separated_scores.push_back(tmp_separated_scores[i]);
-      }
+      separated_scores = feature.getMetaValue(score_name).toStringList();
     }
 
     return separated_scores;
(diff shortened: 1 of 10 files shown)


src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
```diff
@@ -343,7 +343,7 @@ namespace OpenMS
 
         if ((String)feature_it.getMetaValue("id_target_num_transitions") != "")
         {
-          int id_target_num_transitions = feature_it.getMetaValue("id_target_num_transitions").toString().toInt();
+          int id_target_num_transitions = std::stoi(feature_it.getMetaValue("id_target_num_transitions").toString(),nullptr);
```
hroest: @cbielow @timosachsenberg can you maybe comment on this, this looks really ugly and I was under the impression / hope that `toInt()` would "just work"

grosenberger: I think it did before, but this seems to be broken now. Were there any changes recently?

hroest: not in `toInt()` but in how floats and integers are streamed to strings I think

cbielow:
the recent changes did not affect string->int parsing.
Do you have an example string which is broken?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
timosachsenberg: seems relatively well tested: https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/String_test.cpp#L541

grosenberger: Ok, I found the issue. ``num_transitions`` is added as ``Score`` (double) and not as ``MetaValue`` and thus returned ``double``.

hroest:
ok, this looks like a substantial improvement now, nice work!

src/openswathalgo/source/OPENSWATHALGO/ALGO/MRMScoring.cpp
```diff
@@ -493,15 +491,13 @@ namespace OpenSwath
       intensities.push_back(intensities_id / xcorr_contrast_matrix_[0].size());
     }
 
-    std::stringstream ss;
+    std::vector<double> scores;
```
cbielow: you could just return `intensities`, its already a `vector<double>` ?

src/openswathalgo/source/OPENSWATHALGO/ALGO/MRMScoring.cpp
```diff
@@ -638,38 +634,35 @@ namespace OpenSwath
...
-  std::string MRMScoring::calcSeparateSNScore(OpenSwath::IMRMFeature* mrmfeature, std::vector<OpenSwath::ISignalToNoisePtr>& signal_noise_estimators)
+  std::vector<double> MRMScoring::calcSeparateSNScore(OpenSwath::IMRMFeature* mrmfeature, std::vector<OpenSwath::ISignalToNoisePtr>& signal_noise_estimators)
   {
     OPENSWATH_PRECONDITION(signal_noise_estimators.size() > 0, "Input S/N estimators needs to be larger than 0");
 
     std::vector<double> sn_score;
     if (signal_noise_estimators.size() == 0)
     {
-      return std::string();
+      return std::vector<double>();
```
cbielow: `return {}` should do?

src/openswathalgo/source/OPENSWATHALGO/ALGO/MRMScoring.cpp
```diff
@@ -638,38 +634,35 @@ namespace OpenSwath
...
     }
 
     for (std::size_t k = 0; k < signal_noise_estimators.size(); k++)
     {
       sn_score.push_back(signal_noise_estimators[k]->getValueAtRT(mrmfeature->getRT()));
     }
 
-    std::stringstream ss;
+    std::vector<double> scores;
```
cbielow: return `sn_score` directly (and do the above filtering when filling it)?

src/tests/class_tests/openms/source/MRMFeatureFinderScoring_test.cpp
```diff
@@ -301,15 +301,20 @@ START_SECTION(void pickExperiment(OpenSwath::SpectrumAccessPtr input, FeatureMap
...
+  TEST_REAL_SIMILAR(feature.getMetaValue("id_target_ind_log_intensity").toDoubleList()[1], 7.92704);
   TEST_REAL_SIMILAR(feature.getMetaValue("id_target_num_transitions"), 2);
-  TEST_EQUAL(feature.getMetaValue("id_target_ind_xcorr_coelution"), "1;1.66667");
-  TEST_EQUAL(feature.getMetaValue("id_target_ind_xcorr_shape"), "0.68631;0.690494");
-  TEST_EQUAL(feature.getMetaValue("id_target_ind_log_sn_score"), "1.16692;4.45008");
-  TEST_EQUAL(feature.getMetaValue("id_target_ind_isotope_correlation"), "");
-  TEST_EQUAL(feature.getMetaValue("id_target_ind_isotope_overlap"), "");
-  TEST_EQUAL(feature.getMetaValue("id_target_ind_massdev_score"), "");
+  TEST_REAL_SIMILAR(feature.getMetaValue("id_target_ind_xcorr_coelution").toDoubleList()[0], 1);
+  TEST_REAL_SIMILAR(feature.getMetaValue("id_target_ind_xcorr_coelution").toDoubleList()[1], 1.66667);
```
cbielow: maybe we should come up with TEST_REAL_SIMILAR version for vectors and avoid this workaround.

hroest:
seems like it all worked, on Windows only SIRIUS failed  (unrelated, see https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-15,os_label=win10_64/2998/testReport/junit/projectroot.src.tests/topp/UTILS_AssayGeneratorMetabo_8/ )

PR #4066, opened 2019-04-17
[ADD] AGM precursor intensity information (fragment annotation)

oliveralka:
The sorting function relies on precursor intensity information. 

Adding monoisotopic precursor intensity information to CompoundInfo solves the issue for fragment annotated spectra. 

Now assay library generation picks consensus spectrum from the highest intensity precursor (as expected).


src/openms/include/OpenMS/ANALYSIS/ID/SiriusMSConverter.h
```diff
@@ -64,6 +64,7 @@ namespace OpenMS
     double pmass;
```
timosachsenberg: I think these structs should at least have some //< comments for the individual fields

timosachsenberg:
can you add a test for that?

oliveralka:
@timosachsenberg 
I will add the comments and try to find a small test case.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusMSConverter.h
```diff
@@ -50,32 +50,34 @@ namespace OpenMS
...
-    String sf_path;
-    String sf_type;
-    String sf_accession;
-    String native_id_accession;
-    String native_id_type;
+    String sf_path; //< sourcefile path for mztab-m
+    String sf_type; //< sourcefile type for mztab-m
+    String sf_accession; //< sourcefile accessions for mztab-m
+    String native_id_accession; //< nativeID accession for mztab-m
+    String native_id_type; //< nativeID type for mztab-m
```
hroest: that should be '///<`, see http://www.doxygen.nl/manual/docblocks.html#memberdoc

oliveralka:
pyopenms_test_memoryleaktests seems unrelated

PR #4067, opened 2019-04-18
mzTab charge to int

timosachsenberg:
charge was wrong type in mzTab file


timosachsenberg:
@jpfeuffer do you see any problems in mzTabReader if we correct that?

jpfeuffer:
No, I see fixed problems instead ;)

jpfeuffer:
Done? Merge?

PR #4068, opened 2019-04-18
[FIX] SiriusAdapter mzTab adherence to standard 

oliveralka:
Add "opt_global_" to optional/custom entries in adherence to the mztab standard.


oliveralka:
pyopenms_test_memoryleaktests seems unrelated

PR #4071, opened 2019-04-18
Add an additional style test for potential copies in loops

jpfeuffer:
Related to https://github.com/OpenMS/OpenMS/issues/4049

We should probably remove the thirdparty folder from the checks, so wait for merge.
And we could also remove all false positives first.


jpfeuffer:
See http://cdash.openms.de/viewTest.php?onlyfailed&buildid=308588

hroest:
removal of false positives with `OMS_CODING_TEST_EXCLUDE`

hroest:
down to 84 reports ...

hroest:
down to 33 reports ...

hroest:
down to 0 reports ...

PR #4074, opened 2019-04-23
[FEATURE] Update rpm package.

jpfeuffer:
I was recently building on a CentOS machine and thought I would bring the rpms back to life.
Initial attempt without GUI tools worked great (also the autodependency resolution).
Could be refactored to join duplicate variables with the Debian package.


jpfeuffer:
Failed builds have to be false positives as it only touches unreachable code.
I'll merge to test more on the build machines.

PR #4076, opened 2019-04-26
[FEATURE] MatchedIterator

cbielow:
when traversing two sorted arrays and trying to find matching pairs based on some distance metric, this new Iterator is better than MSSpectrum::findNearest(). Its also more general, since you can plug in any distance metric and container you like.

I've changed the most obvious candidates (AScore, PScore, Hyperscore) to the new method.

little Benchmark:
Iterating over 1e6 spectra pairs using the old HyperScore implementation vs. the new one:
before: 12.41 s (wall), 12.41 s (CPU), 0.00 s (system), 12.41 s (user)
after: 5.70 s (wall), 5.69 s (CPU), 0.00 s (system), 5.69 s (user)
So >2x speedup. Since Hyperscore does some other things as well, the actual speedup is probably larger.

SimpleSearchEngine does not show a speedup (on Windows at least)... maybe Im doing something wrong. @timosachsenberg maybe you can give it a spin?


src/openms/include/OpenMS/DATASTRUCTURES/MatchedIterator.h
```diff
@@ -0,0 +1,326 @@
...
+
+      /**
+        @brief Advances to the next valid pair
+
+        @exception Exception::InvalidIterator If iterator is already at end
+      */
+      MatchedIterator& operator++()
+      {
+        // are we at end already? --> wrong usage
+        if (isEnd_()) throw Exception::InvalidIterator(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
```
hroest: this should never happen and may introduce an unnecessary check. 

```suggestion
        OPENMS_PRECONDITION(!isEnd_(), "Try to advance beyond end iterator");
```

src/openms/include/OpenMS/DATASTRUCTURES/MatchedIterator.h
```diff
@@ -0,0 +1,326 @@
...
+
+
+    The iterator always choses the closest matching peak in the target container, if more than one candidate is found in the
+    match-window. If two peaks have equal distance, the smaller value is preferred.
+    If no peak is found within the given tolerance (distance), the reference peak does not yield a result and the next reference peak is tested.
+    This means the operator++ can be called at most(!) ref.size()-1 times before it == is.end().
+    
+    The TRAIT template argument (e.g., ValueTrait, DaTrait or PpmTrait) encodes the distance metric (on the value directly, or a member of the value_type, e.g. ppm or Da for m/z, or RT or any other metric you like).
+    The simplest use case would be a vector<float> or similar.
+    The TRAIT struct just defines two functions which return some distance metrics and accept elements of the container CONT as arguments.
```
hroest: maybe be more specific that the TRAIT structy for Da / ppm assumes that the container element has a function called `getMZ()`

src/openms/include/OpenMS/DATASTRUCTURES/MatchedIterator.h
```diff
@@ -0,0 +1,326 @@
...
+#include <OpenMS/MATH/MISC/MathFunctions.h>
+
+#include <iterator>
+#include <type_traits>
+
+namespace OpenMS
+{
+  
+  /**
+    @brief For each element in the reference container the closest peak in the target will be searched. If no match is found within the tolerance window, the peak will be skipped over.
```
hroest: maybe in addition to the theory also discuss a potential application, e.g. iterating through matching peaks in two spectra?
```suggestion
    @brief For each element in the reference container the closest peak in the target will be searched. If no match is found within the tolerance window, the peak will be skipped over.
    
    This class can be used for example to iterate through the matching peaks in two spectra (e.g. experimental spectrum and reference spectrum) that are within a given m/z tolerance.
```

hroest:
> Iterating over 1e6 spectra pairs using the old HyperScore implementation vs. the new one:
> before: 12.41 s (wall), 12.41 s (CPU), 0.00 s (system), 12.41 s (user)
> after: 5.70 s (wall), 5.69 s (CPU), 0.00 s (system), 5.69 s (user)

very nice indeed!

hroest:
@cbielow did you forget a file?


-- Configuring done
CMake Error at src/tests/class_tests/openms/CMakeLists.txt:82 (add_executable):
  Cannot find source file:

    source/MatchedIterator_test.cpp

cbielow:
Thanks Hannes for your suggestions. I updated the docs etc.

Did a small run for SimpleSearchEngine in the meantime:
the new hyperscore implementation using matched iterators shaves of a few seconds (from 1:05min to 1:00min).

src/openms/include/OpenMS/COMPARISON/SPECTRA/SpectrumAlignment.h
```diff
@@ -301,27 +303,15 @@ namespace OpenMS
...
-        Size j = s2.findNearest(theo_mz);
-        double exp_mz = s2[j].getMZ();
-
-        // found peak match
-        if (std::abs(theo_mz - exp_mz) < max_dist_dalton)
-        {
-          alignment.push_back(std::make_pair(i, j));
-        }
+      #endif
+      #endif
```
timosachsenberg: two endifs?

cbielow: yep. Pairs with 
```
#ifdef ALIGNMENT_DEBUG
#if 0
```
in the code. I did not touch that part though

src/openms/include/OpenMS/DATASTRUCTURES/MatchedIterator.h
```diff
@@ -0,0 +1,332 @@
...
+      /// index into target container
+      size_t tgtIdx() const
+      {
+        return it_tgt_ - tgt_begin_;
+      }
+
+      /**
+        @brief Advances to the next valid pair
+
+        @exception Exception::InvalidIterator If iterator is already at end
```
hroest:
```suggestion
```

src/tests/class_tests/openms/source/MatchedIterator_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+START_SECTION(~MatchedIterator())
+{
+  delete ptr;
+}
+END_SECTION
+
+vector<double> ref = { 0, 1, 2, 3, 4, 5, 6, 7, 13 };
+vector<double> target = { -0.01, 2.5, 3.5, 7, 11 };
+vector<double> empty;
```
hroest: @cbielow can you test the ppm option as well?

cbielow: sure

cbielow:
rebuild jenkins

cbielow:
VS2015 jenkins has some unrelated problem *config.h not found*. And py memtest fails.
How do I retry jeninks?

PR #4080, opened 2019-04-29
Feature/pyopenms2

hroest:
Mostly changes to pyOpenMS

- introduces `SVMWrapper::predictPeptideRT` for RT prediction which can be wrapped by Python
- introduces `encode32/64` and `decode32/64` for Base64 coding in Python (before only 64 bit floats could be en/decoded)
- adds a range check for getChromatogram / getSpectrum so that Python doesnt segfault any more
- adds plotting of spectra/chromatograms


hroest:
looks like all checks pass, merge?

PR #4082, opened 2019-04-30
style: rename LogStream macros to avoid name clashes

pasqoo:
This PR renames all occurrences of the logging macros prefixing them with `OPENMS_`.

For example: `LOG_INFO` -> `OPENMS_LOG_INFO`

I made heavy use of find/replace on multiple files (case sensitive, word-match, i.e. `MyString` does not match `ThisIsMyString`) at the same time. I did not check every single change.

This would solve clashes with libraries like plog.

### DIFF
--- a/src/openms/source/FORMAT/MSPGenericFile.cpp
+++ b/src/openms/source/FORMAT/MSPGenericFile.cpp
@@ -93,7 +93,7 @@ namespace OpenMS
     boost::regex re_cas_nist("^CAS#: ([\\d-]+);  NIST#: (\\d+)"); // specific to NIST db
     boost::regex re_metadatum("^(.+): (.+)", boost::regex::no_mod_s);
 
-    LOG_INFO << "\nLoading spectra from .msp file. Please wait." << std::endl;
+    OPENMS_LOG_INFO << "\nLoading spectra from .msp file. Please wait." << std::endl;
 
     while (!ifs.eof())
     {
@@ -101,11 +101,11 @@ namespace OpenMS
       // Peaks
       if (boost::regex_search(line, m, re_points_line))
       {
-        // LOG_DEBUG << "re_points_line\n";
+        // OPENMS_LOG_DEBUG << "re_points_line\n";
         boost::regex_search(line, m, re_point);
         do
         {
-          // LOG_DEBUG << "{" << m[1] << "} {" << m[2] << "}; ";
+          // OPENMS_LOG_DEBUG << "{" << m[1] << "} {" << m[2] << "}; ";
           const double position { std::stod(m[1]) };
           const double intensity { std::stod(m[2]) };
           spectrum.push_back( Peak1D(position, intensity) );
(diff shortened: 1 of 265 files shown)


hroest:
:+1: I am in favor of this so that we can potentially have multiple logging backends or replace them easily. I also agree that we should not be doing logging ourselves if we can avoid it as its tricky to get right in a multi-threaded context

hroest:
> This would solve clashes with libraries like plog.

that sounds like an issue as well if these macros also have global names and are not prefixed with PLOG_WARN / PLOG_DEBUG etc...

src/openms/source/FORMAT/MSPGenericFile.cpp
```diff
@@ -130,22 +130,22 @@ namespace OpenMS
...
       }
     }
     // To make sure a spectrum is added even if no empty line is present before EOF
     addSpectrumToLibrary(spectrum, library);
-    ifs.close();
-    LOG_INFO << "Loading spectra from .msp file completed." << std::endl;
-    std::cout << "PARSE TIME: " << ((std::clock() - start) / (double)CLOCKS_PER_SEC) << std::endl;
+    OPENMS_LOG_INFO << "Loading spectra from .msp file completed." << std::endl;
+    // DEBUG
+    // std::cout << "PARSE TIME: " << ((std::clock() - start) / (double)CLOCKS_PER_SEC) << std::endl;
```
hroest: looks like this leads to an error because you are not using the variable `start` any more.  Also, why are you not closing `ifs` any more?
```suggestion
    OPENMS_LOG_INFO << "PARSE TIME: " << ((std::clock() - start) / (double)CLOCKS_PER_SEC) << std::endl;
```

pasqoo: `.close()` is called automatically by the object's destructor, hence I preferred to remove the line because unnecessary.

I will comment out the unused variable's declaration.

pasqoo:
I plan to also propose an analogous PR to the plog project

hroest:
rebuild jenkins

hroest:
I am all for changing this, @jpfeuffer @timosachsenberg can you see any other places where this needs to be changed (Tutorial, documentation?) I have changed it now here as well: https://github.com/OpenMS/OpenMS/pull/4082/commits/50097001f7df1ac6a3b965fe57aa26a1339d5569 but I am not sure that this is everything...

pasqoo:
My search and replace was done only on *.h and *.cpp files

pasqoo:
I have just done a search of the five old macros without any restriction on file types, but still with case-sensitive and full-word-match enabled.
I did not find anything except for a line in the _CHANGELOG_, line 2343.

pasqoo:
@hroest Changes to plog macros have already been applied: https://github.com/SergiusTheBest/plog/commit/12bfdf0fd91382d7d8ebb4a439e75462b931f3ea
One only needs to make sure that `#PLOG_OMIT_LOG_DEFINES` is defined.

hroest:
Ok great. I guess we could now in theory replace the logging backend?

PR #4085, opened 2019-05-01
[STYLE/TEST] AbsoluteQuantitationMethod: rule of zero, more tests

pasqoo:
### Changes
In class `AbsoluteQuantitationMethod`:
- remove unnecessary code by applying the rule of zero and in-class initializers
- add tests for: default constructor, copy constructor and copy assignment operator

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
+++ b/src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
@@ -127,18 +127,18 @@ public:
     Param getTransformationModelParams() const; ///< Transformation model parameters getter
 
 private:
+    Param transformation_model_params_; ///< transformation model parameters
     String component_name_; ///< id of the component
     String feature_name_; ///< name of the feature (i.e., peak_apex_int or peak_area)
     String IS_name_; ///< the internal standard (IS) name for the transition
+    String concentration_units_; ///< concentration units of the component's concentration
+    String transformation_model_; ///< transformation model
     double llod_ { 0.0 }; ///< lower limit of detection (LLOD) of the transition
     double ulod_ { 0.0 }; ///< upper limit of detection (ULOD) of the transition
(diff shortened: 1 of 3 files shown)


pasqoo:
Things seem fine, except for an exception thrown by the tooling: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-15,os_label=win10_64/3042/console

```
ERROR: Step ‘Publish xUnit test result report’ failed: no workspace for openms_pr_ci/compiler=msvc-15,os_label=win10_64 #3042
ERROR: Step ‘GitHub PR: set PR status’ failed: no workspace for openms_pr_ci/compiler=msvc-15,os_label=win10_64 #3042
Finished: FAILURE
```

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h
```diff
@@ -133,12 +130,12 @@ namespace OpenMS
...
-    double ulod_; ///< upper limit of detection (ULOD) of the transition
-    double lloq_; ///< lower limit of quantitation (LLOQ) of the transition
-    double uloq_; ///< upper limit of quantitation (ULOQ) of the transition
-    Int n_points_; ///< number of points used in a calibration curve
-    double correlation_coefficient_; ///< the Pearson R value for the correlation coefficient of the calibration curve
+    double llod_ { 0.0 }; ///< lower limit of detection (LLOD) of the transition
+    double ulod_ { 0.0 }; ///< upper limit of detection (ULOD) of the transition
+    double lloq_ { 0.0 }; ///< lower limit of quantitation (LLOQ) of the transition
+    double uloq_ { 0.0 }; ///< upper limit of quantitation (ULOQ) of the transition
+    Int n_points_ { 0 }; ///< number of points used in a calibration curve
```
cbielow: maybe order members by size?! `n_points_` incurs 4 bytes of padding to make the following double align on 8 byte boundaries. Reordering might not prevent that in this instance because there is only a single integer, but its good practice :)

PR #4092, opened 2019-05-06
[FIX] SiriusAdapterAlgorithm - precursor_mz_tolerance_unit

oliveralka:
precursor_mz_tolerance_unit was always set to false.

Here, always Da was used for FeatureMapping instead of ppm, which may lead to errors in the mapping of a feature and the dependent MS2 spectra. In the end, SIRIUS may use false MS2 spectra for calculation/fragment annotation. 

This should be fixed with this PR.


src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -197,7 +197,7 @@ namespace OpenMS
           unsigned int num_masstrace_filter = sirius_algo.filter_by_num_masstraces_;
           double precursor_mz_tol = sirius_algo.precursor_mz_tolerance_;
           double precursor_rt_tol = sirius_algo.precursor_rt_tolerance_;
-          bool ppm_prec = (sirius_algo.precursor_mz_tolerance_unit_ == "true") ? true : false; 
+          bool ppm_prec = (sirius_algo.precursor_mz_tolerance_unit_ == "ppm") ? true : false;
```
hroest:
```suggestion
          bool ppm_prec;
          if (sirius_algo.precursor_mz_tolerance_unit_ == "ppm") ppm_prec = true;
          else if (sirius_algo.precursor_mz_tolerance_unit_ == "Da") ppm_prec = false;
          throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Precursor m/z tolerance unit is either ppm or Da");
```

hroest: prevent this from happening in the future - either by doing this or by adding regression tests? also for `feature_only_` ?

oliveralka: Will add that for feature_only_ as well. 
The test should be fine now since the older ones use Da and the newer (SIRIUS) use ppm.

hroest:
> Also: why are most member functions of the class static? You pass in an instance of the class, which somewhat turns the world upside down :)

I agree, you would then have access to the members such as `feature_only_`. But maybe lets do that in the next PR?

oliveralka:
Ok, will fix the typo and will do the other stuff in the next PR (which is not as critical as this one.)

PR #4095, opened 2019-05-08
[FEATURE] ThermoRawFileParser support in FileConverter

timosachsenberg:


jpfeuffer:
Cool addition! Some points should probably be addressed before:
- Add license in the thirdparty so it will be shown in the installer
- Add docu on how to get the required executables
- Add that it only works for thermo raw files somewhere very prominent
- Probably better if you use Windows/Unix specific macros instead of relying on users to not type anything in the executable fields

jpfeuffer:
And, can you create a PR for THIRDPARTY so we can point to master later, instead of a separate branch?

timosachsenberg:
https://github.com/OpenMS/THIRDPARTY/pull/53

timosachsenberg:
Will add: https://github.com/ypriverol/ThermoRawFileParser/blob/master/LICENSE
Which executables? 
The mono runtime? Or where to update the reader to check if a new version exists?

jpfeuffer:
Yes, add a howto for the mono runtime in the tool docu.
And add a howto for maintainers in the Readme.md of the folder in the thirdparty repository.
Add the license to the license.txt of THIRDPARTY. It will be auto-appended to our license.

timosachsenberg:
I think I made all changes. Please feel free to add additional information if you think they are necessary.

timosachsenberg:
do we also need to add them to the path in KNIME?

hroest:
@timosachsenberg any tests? maybe a very small file?

timosachsenberg:
Hmm don’t know if there are super small raw files? Anyone?

timosachsenberg:
https://github.com/HegemanLab/small-thermo-raw-file-examples?files=1

jpfeuffer:
@timosachsenberg Yes right, the executables should be added to the path in the binaries.ini

timosachsenberg:
ok what about mono? IIRC paths are not propagated so do we have access to that?

timosachsenberg:
I added a small test file and it runs on my linux with mono compiled from source.
@jpfeuffer if you find time can you help with the remaining KNIME / packaging stuff

timosachsenberg:
rebuild jenkins

timosachsenberg:
seems mono is missing on our build machines http://cdash.openms.de/testDetails.php?test=56776291&build=311552

jpfeuffer:
should be installed on mac now

jpfeuffer:
yes, will be hard to set this right in KNIME on mac. I think we have to force the user to give a path in the executable parameter.

timosachsenberg:
rebuild jenkins

timosachsenberg:
Seems to fail on Mac http://cdash.openms.de/viewTest.php?onlyfailed&buildid=311691 file not created

jpfeuffer:
Yes. You would have to look into the 130MB test log.

jpfeuffer:
The executables execute succesfully with mono on Mac. Outside of the adapter I also see the file being converted. Can not test it with the adapter since the build does not exist anymore. There has to be something with the moving of the newfile. Maybe it was written to a wrong folder? I would suggest to add more debug output anyway. What if the moving fails?

timosachsenberg:
Did you try this also on your Mac? Running with debug option should output the temp folder

jpfeuffer:
Nope, I don't have a build of it. Just tried the executables.

jpfeuffer:
There seem to be locale related problems on Windows. Probably from the rawparser.

hroest:
thats kind of interesting:

```

in1:  C:\jenkins\ws\openms_pr_ci\979a8a2f\build\src\tests\topp\ginkgotoxin-ms-switching_out_tmp.mzML   (line: 72, position/column: 64/65)
              <cvParam cvRef="MS" accession="MS:1000927" value="!
                                                                170945" name="ion injection time" unitAccession="UO:0000028" unitName="millisecond" unitCvRef="UO" />

in2:  C:\jenkins\ws\openms_pr_ci\979a8a2f\source\src\tests\topp\THIRDPARTY\ginkgotoxin-ms-switching_out.mzML   (line: 72, position/column: 64/65)
              <cvParam cvRef="MS" accession="MS:1000927" value="!
                                                                170.945" name="ion injection time" unitAccession="UO:0000028" unitName="millisecond" unitCvRef="UO" />
```

its off by a factor of 1000 ?

timosachsenberg:
https://github.com/compomics/ThermoRawFileParser/issues/45

jpfeuffer:
Don't you need to disable the raw picking in the test now?

jpfeuffer:
Ah no the opposite. It should be centroided to match the test file. Unfortunately you really don't set the flag. Something is wrong.

hroest:
@timosachsenberg is there any chance we can directly hook into the API and compile against the dlls / so and read the raw files directly? Have you tried that ?

timosachsenberg:
No this happens in the .net / mono vm. Currently there is no way to share memory between host system and net . Might change in newer versions. This uses the new net assembly as opposed to the windows dlls in pwiz.

timosachsenberg:
can't get it to work on travis. any ideas?

hroest:
> can't get it to work on travis. any ideas?

disable it on travis?

src/topp/FileConverter.cpp
```diff
@@ -323,7 +328,7 @@ class TOPPFileConverter :
   {
     registerInputFile_("in", "<file>", "", "Input file to convert.");
     registerStringOption_("in_type", "<type>", "", "Input file type -- default: determined from file extension or content\n", false, true); // for TOPPAS
-    String formats("mzData,mzXML,mzML,cachedMzML,dta,dta2d,mgf,featureXML,consensusXML,ms2,fid,tsv,peplist,kroenik,edta");
+    String formats("mzData,mzXML,mzML,cachedMzML,dta,dta2d,mgf,featureXML,consensusXML,ms2,fid,tsv,peplist,kroenik,edta,raw");
```
cbielow: Can you move .raw more to the front since its probably a common use case

src/topp/FileConverter.cpp
```diff
@@ -351,6 +356,14 @@ class TOPPFileConverter :
     registerDoubleOption_("lossy_mass_accuracy", "<error>", -1.0, "Desired (absolute) m/z accuracy for lossy compression (e.g. use 0.0001 for a mass accuracy of 0.2 ppm at 500 m/z, default uses -1.0 for maximal accuracy).", false, true);
 
     registerFlag_("process_lowmemory", "Whether to process the file on the fly without loading the whole file into memory first (only for conversions of mzXML/mzML to mzML).\nNote: this flag will prevent conversion from spectra to chromatograms.", true);
+#ifdef OPENMS_WINDOWSPLATFORM
+    String net_executable = "";
+#else
+    String net_executable = "mono";
+#endif
+    registerInputFile_("NET_executable", "<executable>", net_executable, "The .NET framework executable. Only required on linux and mac.", false, true, ListUtils::create<String>("skipexists"));
```
cbielow: Not sure this is a good idea. Workflow defaults will break across platforms.
How about: by default the param value is emtpy which gets interpreted as Mono on non-win systems internally.

src/topp/FileConverter.cpp
```diff
@@ -368,6 +381,7 @@ class TOPPFileConverter :
     bool convert_to_chromatograms = getFlag_("convert_to_chromatograms");
     bool lossy_compression = getFlag_("lossy_compression");
     double mass_acc = getDoubleOption_("lossy_mass_accuracy");
+    bool no_peak_picking = getFlag_("noRawPeakPicking");
```
cbielow: Shouldbe no_raw_p(eak)p(icking)??

jpfeuffer:
Concerning calling C# functions from C++, I recently read some blog posts that claim that it should be working. It is a lot of effort since you manually have to map all the functions that you need and it would introduce a big dependency with the full mono runtime but in theory it should work.

https://tzertzelos1.wordpress.com/tag/embedding-mono/
https://www.mono-project.com/docs/advanced/embedding/

timosachsenberg:
passed except memory leaks test

PR #4097, opened 2019-05-10
[FIX] OpenSWATHOSWWriter: Separate score fix

grosenberger:
This PR fixes an issue in OpenSwathOSWWriter, where some types of ``MetaValue`` were not correctly converted.

This PR further removes support for UIS scores from OpenSWATHTSVWriter, as the downstream tools (PyProphet & Percolator) now only support OSW and parsing of concatenated scores with different length was always suboptimal.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathTSVWriter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathTSVWriter.h
@@ -60,15 +60,13 @@ namespace OpenMS
     bool doWrite_;
     bool use_ms1_traces_;
     bool sonar_;
-    bool enable_uis_scoring_;
 
   public:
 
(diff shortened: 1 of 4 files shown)


hroest:
> This PR further removes support for UIS scores from OpenSWATHTSVWriter, as the downstream tools (PyProphet & Percolator) now only support OSW and parsing of concatenated scores with different length was always suboptimal.

is there still a way to convert to TSV through pyProphet export? If so, then I am fine with this

grosenberger:
> > This PR further removes support for UIS scores from OpenSWATHTSVWriter, as the downstream tools (PyProphet & Percolator) now only support OSW and parsing of concatenated scores with different length was always suboptimal.
> 
> is there still a way to convert to TSV through pyProphet export? If so, then I am fine with this

Right now we don't export the scores by default to save space, but that would be easy to add.

PR #4098, opened 2019-05-10
[FIX] some compile warnings on GCC 81 about catching exceptions by value

timosachsenberg:


hroest:
:+1: but lets wait for OSX and Windows

PR #4100, opened 2019-05-13
[FIX] prevent TOPPAS crash on Windows when copying output files 

cbielow:
if the upstream node of an output node only has a single allowed output type which at the same time has lower/uppercase differences to the default types in OpenMS, e.g. ConsensusXML vs. consensusXML, TOPPAS will try to rename the file, which fails, because it first deletes the target (if it exists).
Since this is the same file, we would be deleting the source file... bad...
This is only a problem on Windows since the file system is not case sensitive (and hence regards these files as equal).

### DIFF
--- a/src/openms/source/APPLICATIONS/TOPPBase.cpp
+++ b/src/openms/source/APPLICATIONS/TOPPBase.cpp
@@ -979,13 +979,14 @@ namespace OpenMS
     //check if formats are known
     if (force_OpenMS_format)
     {
-      for (Size i = 0; i < formats.size(); ++i)
+      for (const auto& f : formats)
       {
-        if (formats[i] != "fid")
+        if (f != "fid")
         {
-          if (FileHandler::getTypeByFileName(String(".") + formats[i]) == FileTypes::UNKNOWN)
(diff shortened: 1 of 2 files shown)


jpfeuffer:
Path extracted from Executable Path does not exist! Returning empty string!
No OpenMS version information found in file /home/jenkins/ws/openms_pr_ci/a9154670/source/src/tests/topp/INIUpdater_3_old.toppas! Assuming OpenMS 1.8 and below.
```Update for file /home/jenkins/ws/openms_pr_ci/a9154670/source/src/tests/topp/INIUpdater_3_old.toppas failed because the tool 'FeatureFinderCentroided' returned with an error! Check if the tool works properly.
The following INI/TOPPAS files could not be updated:
  /home/jenkins/ws/openms_pr_ci/a9154670/source/src/tests/topp/INIUpdater_3_old.toppas
INIUpdater took 0.42 s (wall), 0.00 s (CPU), 0.00 s (system), 0.00 s (user).```
Most likely related

cbielow:
rebuild jenkins

cbielow:
all tests pass (the jenkins' icon shows fail, but once you click on it, its all good).
The initial failure was a hiccup.

PR #4103, opened 2019-05-14
[FIX] do not use underscore replacement for mzTab cells

cbielow:
when writing optional column values which contain spaces, they should not be replaced with underscores (as done for column header names)

Example for DoubleLists
old: `"[0.5,_1.4,_-2.0,_0.1]"`
new `"[0.5, 1.4, -2.0, 0.1]"`


timosachsenberg:
I checked against the specification and Chris is right.
+1

cbielow:
rebuild jenkins

cbielow:
jenkins fails spuriously (last time it was ubuntu, this time its windows). Overall - all tests pass :)

PR #4106, opened 2019-05-14
Feature/prm

hroest:
Support for PRM in OpenSwathWorkflow

- allow multiple, overlapping DIA windows at different retention times
- select closest DIA window based on center m/z and assay m/z (select only one)

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
@@ -819,7 +819,7 @@ namespace OpenMS
       if (ms1_map_ == nullptr) 
       {
         throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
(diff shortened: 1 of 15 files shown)


src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -742,6 +816,11 @@ namespace OpenMS
     // share a single filestream and call seek on it, chaos will ensue).
     if (use_ms1_traces_)
     {
+      if (ms1_map_ == nullptr) 
+      {
+        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
+            "Error, you attempted to use MS1 trances but no MS1 map provided." );
```
hroest:
```suggestion
            "Error, attempted to use MS1 traces, but no MS1 map was provided." );
```

hroest:
rebuild jenkins

hroest:
looks like all tests pass, merge? @timosachsenberg @jpfeuffer

PR #4111, opened 2019-05-17
Fix protein identifiers associated to PepIDs

### LINKED ISSUE #70: idXML/featureXML should store identifier linking proteins and peptides [70]
Submitted by hendrikweisser on 2009-12-08 16:47:50

The ProteinIdentification and PeptideIdentification classes each have a member "id_" that links a Peptide-ID to the corresponding Protein-ID. This identifier gets lost when the data is stored in an idXML or annotated featureXML file.
(By default, the identifier is "searchengine_datetime", and it is recreated based on these two fields when an idXML/featureXML file is loaded.)

The problem occurs when, in a set of related protein identification runs, neither the search engine nor date/time are unique. This happens in a pepXML file that bundles search results for different mz(X)ML files, but from the same search engine (there is only one date/time at the beginning of the file). We can easily handle this case in the library by assigning unique identifiers, but these are lost when results are written to a file. This can lead to jumbled identifications later in the pipeline.

I propose adding an optional "id" attribute to the "IdentificationRun" element in idXML and changing the usage of the existing "id" attribute of "IdentificationRun" in featureXML (which is now basically a counter) to store the Protein-/PeptideIdentification "id_" member.

---

Commented by cbielow on 2012-09-17 12:26:45:
- should be postponed to new mzIdentML implementation .. id field must be read/written accordingly

cbielow:
fixes #70 by internally adding a UID to the current identifier (searchengine + datetime).

The fix solves the problem as early as possible, i.e. when ProteinIDs are created in our XML files.

This PR also removed ineffective hacks which increase the DateTime by 1 second if a collision is detected. This is very unreliable since its done at individual file level, i.e. works for multiple ProtIDs in a single file, but fails when a FeatureLinker carelessly merges ProtIDs from different featureXMLs.

((also fixes a segfault in an unrelated file causing multiple tests to fail on VS2015 Debug: 
2bcb5b3)


src/openms/source/FORMAT/ConsensusXMLFile.cpp
```diff
@@ -342,22 +342,24 @@ namespace OpenMS
...
-      String identifier = prot_id_.getSearchEngine() + '_' + attributeAsString_(attributes, "date");
+      // set identifier
+      // always generate a unique id to link a ProteinIdentification and the corresponding PeptideIdentifications
+      // , since any FeatureLinker might just carelessly concatenate PepIDs from different FeatureMaps.
+      // If these FeatureMaps have identical identifiers (SearchEngine time + type match exactly), then ALL PepIDs would be falsely attributed
+      // to a single ProtID...
+
       String id = attributeAsString_(attributes, "id");
+      while (true)
+      { // loop until the identifier is unique (should be on the first iteration -- very(!) unlikely it will not be unique)
```
hroest: why do you need the loop? the uid is 64 bit which gives a space of 2 E19 which should always be unique ? Are you concerned about multiple threads?

hroest: PS: also, why dou still need the search engine and the date - shouldnt the UID be enough?

hroest: PS: the probability of any two ids not being unique with 64 bits and assuming that we have a hundred thousand to a few million spectra is still 1e-9 to 1e-6: https://en.wikipedia.org/wiki/Birthday_attack#Mathematics

src/topp/ProteinQuantifier.cpp
```diff
@@ -710,7 +710,7 @@ class TOPPProteinQuantifier :
       IdXMLFile().load(in, proteins, peptides);
       for (Size i = 0; i < proteins.size(); ++i)
       {
-        files_[i].filename = proteins[i].getIdentifier();
+        files_[i].filename = proteins[i].getSearchEngine() + "_" + proteins[i].getDateTime().toString(Qt::ISODate);
```
hroest: why not : `SE + date + UID` ?

hroest:
looks good, needs some test fixes:

```
Test Result (2 failures / +2)

    projectroot.src.tests.class_tests.openms.IdentificationDataConverter_test
    projectroot.src.tests.topp.UTILS_XFDR_2_out22
```

src/openms/source/FORMAT/FeatureXMLFile.cpp
```diff
@@ -523,22 +523,24 @@ namespace OpenMS
...
-      //set identifier
-      String identifier = prot_id_.getSearchEngine() + '_' + attributeAsString_(attributes, "date");
+      // set identifier
+      // always generate a unique id to link a ProteinIdentification and the corresponding PeptideIdentifications
+      // , since any FeatureLinker might just carelessly concatenate PepIDs from different FeatureMaps.
+      // If these FeatureMaps have identical identifiers (SearchEngine time + type match exactly), then ALL PepIDs would be falsely attributed
+      // to a single ProtID...
+
       String id = attributeAsString_(attributes, "id");
+      while (true)
```
hroest: put this into a function to avoid code duplication?

src/openms/source/FORMAT/ConsensusXMLFile.cpp
cbielow: the loop is just there for MurphysLaw... thats all.
In general, there will be a single ProtID, in rare cases a handful. But why test our luck?
This is still not avoiding the problem of merging protIDs at a later stage.. but its a start.

src/topp/ProteinQuantifier.cpp
cbielow: The output files of proteinquantifier looked like the UID does not belong there, so I treated the UID as an internal mapping and did not expose it

cbielow:
tests pass .. codacy is actually fine (if you click on the link). Not sure what the problem is.
Ready for merge

Issue #4093, opened 2019-05-07
PSI extended fasta format support


jpfeuffer:
Since Comet supports it, we could start thinking about extending PeptideIndexer to include
annotated variations and give it an option to do a modification-aware mapping.

Info:
http://www.psidev.info/peff

PR #4112, opened 2019-05-19
[FIX] skip fasta header files, test with PEFF

hroest:
partially addresses #4093 allowing us to read PEFF data without an error, but does not store/parse any of the meta data. What is still missing

- add PEFF to the list of available file endings
- parse the header properly
- parse the key-value pairs properly

Since all parsing is done in seqan, we may have to either ask them to add PEFF support, do it ourselves or write a new parser for PEFF files.


hroest:
>  Maybe you can switch to that?!

hmm there is still quite a few places where we would have to implement this:
```
$ grep -R getline src/openms | wc
     66     438    6886
```

hroest:
@cbielow @jpfeuffer merge?

cbielow:
fine with me. 
The appleclang test passed but is shown in red.. any idea why?

PR #4114, opened 2019-05-19
[DOC] topp/utils docu

hroest:
fix some issues in the TOPP / UTILS documentation - this is to find tools that are not listed in the documentation anywhere:

```
cat doc/doxygen/public/TOPP.doxygen doc/doxygen/public/UTILS.doxygen  > /tmp/both
for f in src/topp/*.cpp src/utils/*.cpp; do 
  bn=`basename $f .cpp`; cnt=`grep -c $bn /tmp/both`; 
  if [ "$cnt" -eq 0 ];then echo $f;fi ;
done 
```

### DIFF
--- a/src/utils/RNPxlSearch.cpp
+++ b/src/utils/RNPxlSearch.cpp
@@ -92,6 +92,25 @@ using namespace OpenMS;
 using namespace OpenMS::Internal;
 using namespace std;
 
+
+//-------------------------------------------------------------
+// Doxygen docu
+//-------------------------------------------------------------
+
+/**
+    @page UTILS_RNPxlSearch RNPxlSearch 
+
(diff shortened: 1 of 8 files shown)


src/utils/RNPxlSearch.cpp
```diff
@@ -92,6 +92,25 @@ using namespace OpenMS;
...
 
+
+//-------------------------------------------------------------
+// Doxygen docu
+//-------------------------------------------------------------
+
+/**
+    @page UTILS_RNPxlSearch RNPxlSearch 
+
+    @brief Annotate RNA to peptide crosslinks in MS/MS spectra.
```
hroest: @timosachsenberg can you expand the docu here at some point?

timosachsenberg: will do

PR #4116, opened 2019-05-20
[FEATURE] annotate ion index and charge as subscript

timosachsenberg:
Added sub/superscripts to TOPPView Peakannotations.
Also added vertical + diagonal line drawing of peak annotations to (manually) reduce overlaps between annotations.


timosachsenberg:
In TOPPView: prints ion index as subscript, charge as superscript
Common losses e.g. H2O are now also with 2 as subscript

timosachsenberg:
@enetz and @hendrikweisser can you check if this change impacts visualization of your fragment ion annotations?

timosachsenberg:
test

src/openms_gui/source/VISUAL/ANNOTATION/Annotation1DPeakItem.cpp
```diff
@@ -196,20 +196,78 @@ namespace OpenMS
...
       delete(closest_ip);
     }
 
-    painter.drawText(bounding_box_, Qt::AlignCenter, text_);
+    // some pretty printing
+    QString text = text_;
+    if (!text.contains("<\\")) // don't process HTML strings again
+    {
+      // extract ion index
+      QRegExp reg_exp("[a|b|c|x|y|z](\\d+)");
```
enetz: according to the QRegExp documentation:
`[ABCD] will match an A or a B or a C or a D`
So you should not need the `|` bars, the `[]` brackets already imply `or` between the letters, I think.
Looks like you would also be matching strings like `|3+` right now.

src/openms_gui/source/VISUAL/ANNOTATION/Annotation1DPeakItem.cpp
```diff
@@ -196,20 +196,78 @@ namespace OpenMS
...
+      text.replace(QRegExp("3\\+$"), "<sup>3+</sup>");
+      text.replace(QRegExp("\\+\\+$"), "<sup>2+</sup>");
+      text.replace(QRegExp("\\+$"), "");
+      text.replace(QRegExp("3\\-$"), "<sup>3-</sup>");
+      text.replace(QRegExp("\\-\\-$"), "<sup>2-</sup>");
+      text.replace(QRegExp("\\-$"), "");
+
+      // common losses
+      text.replace("H2O","H<sub>2</sub>O");
+      text.replace("H2O1","H<sub>2</sub>O");
```
enetz: The order is wrong, `H2O` is replaced first, and then `H<sub>2</sub>O1` will not be seen as `H2O1` and the `1` stays. This happens to my labels.
For `NH3` the two versions are sufficiently different, so this does not happen there.

timosachsenberg: aaah great catch! Thanks!

enetz:
@timosachsenberg my labels are changed substantially by these changes, but since I like the idea and look in general, I adapted the code to work with my labels (PR to your branch).
Right now I have these issues:

- the colors of the labels were lost, my previously green and red labels are all black now. The peak colors are still there. I am actually not sure why this happened and whether this PR did this. The label color should not be affected by the changes here.
- I am not sure I agree, that just removing the single charge `+` and `-` is a good idea.

timosachsenberg:
can you rebase your PR on my recent changes?

timosachsenberg:
@enetz I tried to be closer to what is currently considered quasi-standard (in at least one subfield ;). 
See e.g., https://www.mcponline.org/content/early/2019/05/14/mcp.TIR118.001209 .
It is also the way they do it in Hennings lab.

enetz:
@timosachsenberg I just noticed another case in the RNPxl test files:
e.g. `[M+H]-H2O++ U'+`, here the `++` is not at the end and isn't turned into a `2+` superscript.

Also in some cases, if the largest peak is labeled, the Y scale is normalized by that peak and the label is above the peak, outside of the spectrum. I noticed that by trying to remove all labels, can't see or select this one in some spectra.

I added a check for match position of the ion type again to exclude the cases without matches. It

timosachsenberg:
Yes the old test files need an update. the string was not properly generated in the first place.

enetz:
A similar issue with the automatically generated labels:
e.g. `b6+3\nYTPYLG`
the subscript for the ion index works, but the charge is at the end of the first line, not the whole string.
If we constrain the charge search to the first line, this would work.

timosachsenberg:
Good point. So Eiter the first newline or end of string- whatever comes first. Can you add that?

timosachsenberg:
k I think I fixed "Also in some cases, if the largest peak is labeled, the Y scale is normalized by that peak and the label is above the peak, outside of the spectrum. I noticed that by trying to remove all labels, can't see or select this one in some spectra."
as well as NASE annotations

timosachsenberg:
let's merge and add the support for the automatically generated ones later

PR #4118, opened 2019-05-20
[ADD] Output SIRIUS workspace (AssayGeneratorMetabo)

oliveralka:
Outputs SIRIUS workspace to allow easier handling of fragmentation trees for decoy generation with passatuto. (Current solution.)

### DIFF
--- a/src/utils/AssayGeneratorMetabo.cpp
+++ b/src/utils/AssayGeneratorMetabo.cpp
@@ -56,6 +56,8 @@
 #include <OpenMS/FORMAT/DATAACCESS/SiriusMzTabWriter.h> 
 #include <OpenMS/FORMAT/DATAACCESS/SiriusFragmentAnnotation.h>
 
+#include <QDir>
+
 using namespace OpenMS;
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


oliveralka:
rebuild jenkins

oliveralka:
@jpfeuffer I am not sure what happened here - rebuild again? 
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=appleclang-9.0.0,os_label=highsierra/3102/console

jpfeuffer:
No, consider it as passed. The tests ran but CDash was not responsive. Since there are no test errors, it means it is good to go. What about pyopenms? The usual memory leaks?

oliveralka:
> What about pyopenms? The usual memory leaks?
Yes

jpfeuffer:
+1 from me. But in the future, could one directly create the files in this workspace or can't you pass it to Sirius?

oliveralka:
> +1 from me. But in the future, could one directly create the files in this workspace or can't you pass it to Sirius? 

Ok lets do it like this for now and I will think about an alternative e.g. passing the workspace to SIRIUS - the problem here is that sometimes it is not able to "merge" the workspace. Basically, if the same workspace is selected multiple times (e.g. multiple files), then SIRIUS might fail. 

Another alternative would be just to move instead of copying the directory.

jpfeuffer:
I see. Ok

PR #4120, opened 2019-05-21
[Fix] clang tidy fixes

jpfeuffer:
Attempt.


src/openms/include/OpenMS/METADATA/PeptideHit.h
```diff
@@ -233,7 +233,7 @@ namespace OpenMS
     PeptideHit(double score,
                UInt rank,
                Int charge,
-               const AASequence& sequence);
+               AASequence sequence);
```
timosachsenberg: is there a move inside?

src/openms/source/METADATA/PeptideHit.cpp
```diff
@@ -85,10 +85,10 @@ namespace OpenMS
   PeptideHit::PeptideHit(PeptideHit&& source) noexcept :
     MetaInfoInterface(std::move(source)), // NOTE: rhs itself is an lvalue
     sequence_(std::move(source.sequence_)),
-    score_(std::move(source.score_)),
-    analysis_results_(std::move(source.analysis_results_)),
-    rank_(std::move(source.rank_)),
-    charge_(std::move(source.charge_)),
+    score_(source.score_),
+    analysis_results_(source.analysis_results_),
```
timosachsenberg: why no move here?

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
```diff
@@ -783,7 +783,7 @@ namespace OpenMS
     const map<IdentificationData::ScoreTypeRef, Size>& scores,
     map<Size, MzTabParameter>& output)
   {
-    for (const pair<IdentificationData::ScoreTypeRef, Size>& score_pair :
+    for (const pair<const IdentificationData::ScoreTypeRef, Size>& score_pair :
```
timosachsenberg: I would be finde with const auto&

src/openms/source/METADATA/PeptideHit.cpp
jpfeuffer: Discouraged since it's a built-in type.

src/openms/source/METADATA/ID/IdentificationDataConverter.cpp
jpfeuffer: Yes true. I did not address "autos" yet since it is hard to decide automatically when it helps readability and when it doesnt.

src/openms/source/METADATA/PeptideHit.cpp
timosachsenberg: ah yeah right

src/openms/include/OpenMS/METADATA/PeptideHit.h
jpfeuffer: Yes. It uses the default move constructor of AASequence.

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmSpectrumAlignment.cpp
```diff
@@ -831,14 +831,7 @@ namespace OpenMS
...
     String tmp = (String)param_.getValue("debug");
-    if (tmp == "true")
-    {
-      debug_ = true;
-    }
-    else
-    {
-      debug_ = false;
-    }
+    debug_ = tmp == "true";
```
hroest:
```suggestion
    debug_ = (tmp == "true");
```

src/openms/source/METADATA/PeptideHit.cpp
hroest: hmm I understand but also it seems more readable to use `std::move` and it could lead to problems if we decide to change the type later on and then there is no move...

src/openms/include/OpenMS/METADATA/PeptideHit.h
hroest: I am not sure whether it is really better to 

1. make a copy of the object and then 2. move the object into another object 

compared to 

3. construct an object with a const-ref

I think constructing with a const-ref should always be faster since (1) and (3) should take the same amount of time (copy object vs construct from object) and the first solution takes extra time to then move. Comments?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -524,7 +524,9 @@ namespace OpenMS
       }
       if (drift_upper > 0) 
       {
+        //@todo Why can't the following be done "in-place"?
         std::vector<OpenSwath::SpectrumPtr> tmp;
+        tmp.reserve(all_spectra.size());
         for (const auto& s: all_spectra) tmp.push_back( filterByDrift(s, drift_lower, drift_upper) );
```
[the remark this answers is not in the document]
jpfeuffer: From what I saw, I think so. I will try then.

src/openms/source/COMPARISON/SPECTRA/SpectraSTSimilarityScore.cpp
```diff
@@ -95,6 +95,7 @@ namespace OpenMS
     double min_high_intensity = 0;
     if (!spec.empty())
     {
+      //@todo timosachsenberg Is this a dirty undocumented hack or just wrong?
```
jpfeuffer: @timosachsenberg Can you have a look at that? Since cut_peaks_below is a UInt, I think this will always be either inf, 0 or 1

src/openms/include/OpenMS/METADATA/PeptideHit.h
cbielow: depends on the use-case:
if we pass in rvalues, the copy will be elided, i.e. it only costs a move.
If we always pass in l-values, const-ref is better.

jpfeuffer: True, I decided on that because: In 1) we "only" pay an additional move if the developer does not use an r-value. In 3) we cannot use r-values at all. Did I get that right?

src/openms/source/COMPARISON/SPECTRA/SpectraSTSimilarityScore.cpp
timosachsenberg: Yes this looks wrong. Should be doubles (and probably not use any sorting - just pick max)

jpfeuffer:
Oh no.. It collides with the changes for exceptions by @timosachsenberg 
So where do we put the reference? In the TEST macro or when passing it to the macro?

timosachsenberg:
In the test macro would look nicer I guess

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmSpectrumAlignment.cpp
```diff
@@ -572,7 +572,7 @@ namespace OpenMS
       */
       std::sort(temp.begin(), temp.end(), Compare(false));
       //Int anchor=(Int)(size*anchorPoints_/100);
-      float anchor = (temp.size() * anchorPoints_ / 100);
+      float anchor = temp.size() * anchorPoints_ / 100.0;
```
cbielow: should be `100.0f` ? (otherwise conversion error from double to float)

cbielow:
> 
> 
> Oh no.. It collides with the changes for exceptions by @timosachsenberg
> So where do we put the reference? In the TEST macro or when passing it to the macro?

where exactly is the code youre referring to?

nice PR btw :)

src/openms/include/OpenMS/METADATA/PeptideHit.h
hroest: > In 3) we cannot use r-values at all (at least we dont gain from it). Did I get that right?

I think for 3 it does not help but also does not hurt to use r-values. So I think we should always be better off with (3) since its better for l-values but neither better or worse for r-values. Right?

src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp
```diff
@@ -109,7 +109,7 @@ void CsiFingerIdMzTabWriter::read(const std::vector<String> & sirius_output_path
         const UInt top_n_hits_cor = (top_n_hits >= rowcount) ? rowcount-header : top_n_hits;
 
         // fill identification structure containing all candidate hits for a single spectrum
-        CsiFingerIdMzTabWriter::CsiAdapterIdentification csi_id;
+        CsiFingerIdMzTabWriter::CsiAdapterIdentification csi_id{};
```
hroest: fails with 

```


/.../OpenMS/src/openms/source/FORMAT/DATAACCESS/CsiFingerIdMzTabWriter.cpp:112:65: error: missing initializer for member ‘OpenMS::CsiFingerIdMzTabWriter::CsiAdapterIdentification::mz’ [-Werror=missing-field-initializers]

         CsiFingerIdMzTabWriter::CsiAdapterIdentification csi_id{};
                                                                 ^
```

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
```diff
@@ -111,7 +111,7 @@ void SiriusMzTabWriter::read(const std::vector<String> & sirius_output_paths,
         const UInt top_n_hits_cor = (top_n_hits >= rowcount) ? rowcount-header : top_n_hits;
         
         // fill identification structure containing all candidate hits for a single spectrum
-        SiriusMzTabWriter::SiriusAdapterIdentification sirius_id;
+        SiriusMzTabWriter::SiriusAdapterIdentification sirius_id{};
```
hroest: same issue here - breaks the compilation

hroest:
@jpfeuffer I suggest to fix the issue and merge at least part of this sooner than later, otherwise it will become harder and harder to merge.

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
jpfeuffer: That's a gcc bug but I will fix.

src/openms/include/OpenMS/METADATA/PeptideHit.h
jpfeuffer: I just did two overloads now. It was done the same already with the setter, so why not in the constructor? It was actually used with r-values e.g. in MzIdentML.

src/openms/source/METADATA/PeptideHit.cpp
jpfeuffer: readded the move.

jpfeuffer:
rebuild jenkins

jpfeuffer:
I don't know why, but I seem to have discovered a big error in MzTabExport:
If you look at the failing test:
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=appleclang-9.0.0,os_label=highsierra/3150/testReport/junit/projectroot.src.tests/topp/TOPP_MzTabExporter_1_out/

It complains about the scan number (11412) for a certain ID .(TMT6plex)VNEVNQFAAK(TMT6plex)
However in the input consensusXML, this combination does not exist. Only the combination of the new code with scan number 11531.

If you agree, I will change the test output.

jpfeuffer:
Instead I introduced a bug in MzTabExporter.

jpfeuffer:
rebuild jenkins

PR #4125, opened 2019-05-23
Fix/tv minor

timosachsenberg:
- using m/z instead of M/Z with unit Th
- using retention time [s] instead of RT [sec]
as axis labels

### DIFF
--- a/src/openms_gui/source/VISUAL/Spectrum1DWidget.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum1DWidget.cpp
@@ -57,7 +57,7 @@ namespace OpenMS
     //set the label mode for the axes  - side effect
     setCanvas_(new Spectrum1DCanvas(preferences, this));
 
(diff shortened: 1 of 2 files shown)


hroest:
generally seems good, maybe we should put those strings into a central location / make them user-defined. Especially as the axis labels are not always the same, they could be m/z, seconds or ion mobility see 

https://github.com/OpenMS/OpenMS/blob/c6f9e0d9c70ae99d46be5f47b612f75f48e30f40/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp#L3577

timosachsenberg:
Hmm, we could put this in the SpectrumWidget base class. Or would you put it in the core library?
By the way, @everyone feel free to add or propose changes (e.g., like using more established terms instead of historical/dated terms that were introduced long ago in TOPPView). I think this could also lower the barrier for new users.

cbielow:
I like the idea of a central place to put these labels. Maybe `GUILabels.h`?
Ideally, each data type knows its associated label and there is no need to manually set it.. but maybe thats a bit too much to ask.

timosachsenberg:
I would not make the label a property of the data type but construct the label depending on the type of data.

PR #4126, opened 2019-05-24
Feature/tv mzannot

timosachsenberg:
for discussion and one of the most frequently requested features:
automated peak m/z and charge annotations of "interesting" peaks.
Bildschirmfoto vom 2019-05-24 16-29-34

I think it is a good start but could be improved by:
- recalculating interesting peaks based on the currently visible area
- read tolerances from config dialog


timosachsenberg:
I will move this to visible area

Issue #3418, opened 2018-04-30
Show all the `peak annotation m/z` at once in MS/MS spectra 


wenjiebi:
Until now, TOPPView does not support adding `peak annotation mz` for all peaks in an MS/MS spectra. It would be really nice if someone can add this function in the newer version.

timosachsenberg:
I think I might be addressing something similar in https://github.com/OpenMS/OpenMS/pull/4126 
I don't annotate all peaks but exclude isotopic peaks and low intensity peaks for now.

PR #4128, opened 2019-05-27
Feature/tv annot2

timosachsenberg:
Currently disabled by default but can be enabled via context menu
grafik

There is a lot of room for improvements but I think it already addresses the main request by most users.

### DIFF
--- a/src/openms_gui/include/OpenMS/VISUAL/APPLICATIONS/TOPPViewBase.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/APPLICATIONS/TOPPViewBase.h
@@ -336,6 +336,8 @@ public slots:
     void toggleGridLines();
     /// Toggles the axis legends
     void toggleAxisLegends();
+    /// Toggles drawing of interesting MZs
+    void toggleInterestingMZs();
     /// Shows current layer preferences
     void showPreferences();
     /// dialog for inspecting database meta data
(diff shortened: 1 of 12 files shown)


timosachsenberg:
Addresses https://github.com/OpenMS/OpenMS/issues/3418 and similar user requests

timosachsenberg:
@enetz I think this one is finished.
- Interesting peaks can now automatically annotated (enable via context menu)
- I readded colors to annotations
- I scale SNAP and PERCENTAGE mode so a small margin remains on the top -> annotations are always visible and 100% in axis in percentrage mode doesn't get cropped anymore.

timosachsenberg:
- also added the axis strings to the base class @cbielow  and @hroest

timosachsenberg:
I merge as this fixes a bug I introduced in the last PR

enetz:
@timosachsenberg something went wrong here, I just tried to load my XL-MS idXML file to annotate an mzML file and this error popped up:
`Caught exception: 'NotImplemented' with message 'this method has not been implemented yet. Feel free to complain about it!'`
on the current develop branch.
The automatic annotations are cool though.

enetz:
@timosachsenberg the ID table works fine after the error, the peak annotatition pop-up table looks fine as well, but selecting or double-clicking an ID does not open up the corresponding MS2 spectrum.
The error message can pop up multiple times, e.g. every time I change tabs between scan view and ID view

enetz:
@timosachsenberg reproducible with these test files in tests/topp/:
OpenPepXL_input.mzML
OpenPepXL_output.idXML

timosachsenberg:
hmm works for me...

timosachsenberg:
k found the issues ... preparing a fix

timosachsenberg:
pushed directly...

enetz:
:+1:  solved
thanks!

PR #4131, opened 2019-05-28
Percolatoradapter (AA before and after do not neccessarily exist)

### LINKED ISSUE #4129: PercolatorAdapter OSW tests fail on Windows debug
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=314366

Two of them have a stack_buffer_overrun after executing the Percolator.exe successfully.
The other probably depends on the output of them.

I suspect it could be related to the changes to the getSeparateScore function, that somehow the sizes of the generated vectors are not big enough and during writing it accesses unallocated memory. But this is just a guess.
(Maybe check the sizes of the vectors before this loop https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp#L359)

cbielow:
fixes #4129 

I'm not sure what the origin of the problem was though. Just fixing symptoms.
Maybe someone with recent insight can double-check that is what we want.

### DIFF
--- a/src/topp/PercolatorAdapter.cpp
+++ b/src/topp/PercolatorAdapter.cpp
@@ -144,9 +144,9 @@ protected:
       
         OPENMS_PRECONDITION(left_dot < right_dot, "Peptide sequence encoding must have dot notation (e.g., A.PEPTIDER.C).")
  
-        // retrieve pre and post AA, e.g., A and C in A.PEPTIDE.C
-        preAA = (row[4][left_dot - 1] == '-') ? '[' : row[4][left_dot - 1];  // const char PeptideEvidence::N_TERMINAL_AA = '[';
-        postAA = (row[4][right_dot + 1] == '-') ? ']' : row[4][right_dot + 1]; // const char PeptideEvidence::C_TERMINAL_AA = ']';
+        // retrieve pre and post AA, e.g., A and C in "A.PEPTIDE.C" or ".PEPTIDE."
+        preAA = (left_dot == 0 || row[4][left_dot - 1] == '-') ? '[' : row[4][left_dot - 1];  // const char PeptideEvidence::N_TERMINAL_AA = '[';
+        postAA = (right_dot + 1 < row[4].size() || row[4][right_dot + 1] == '-') ? ']' : row[4][right_dot + 1]; // const char PeptideEvidence::C_TERMINAL_AA = ']';
 
         // retrieve sequence between dots, e.g., PEPTIDE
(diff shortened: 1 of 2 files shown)


jpfeuffer:
And what is the difference between -.peptide.- and .peptide. ?

cbielow:
> 
> 
> And what is the difference between -.peptide.- and .peptide. ?

The crash occurred because the input string was `.PEPTIDE.`, i.e. `left_dot` is 0, and hence `left_dot - 1` is a segfault.

jpfeuffer:
I don't doubt the usefulness of the check but it would be interesting if both versions occur and if there is a difference between them.

jpfeuffer:
Or if the notation changed in Percolator versions maybe?

cbielow:
The bug was probably not present before (I run debug builds from time to time and would have caught it probably...). So it might be percolator versions, which would mean we need to account for both variants.

hroest:
:+1: I am not sure with OpenSWATH we always have a previous / next AA stored somewhere, so this information may simple be missing.

src/topp/PercolatorAdapter.cpp
```diff
@@ -144,9 +144,9 @@ class PercolatorAdapter :
       
         OPENMS_PRECONDITION(left_dot < right_dot, "Peptide sequence encoding must have dot notation (e.g., A.PEPTIDER.C).")
  
-        // retrieve pre and post AA, e.g., A and C in A.PEPTIDE.C
-        preAA = (row[4][left_dot - 1] == '-') ? '[' : row[4][left_dot - 1];  // const char PeptideEvidence::N_TERMINAL_AA = '[';
-        postAA = (row[4][right_dot + 1] == '-') ? ']' : row[4][right_dot + 1]; // const char PeptideEvidence::C_TERMINAL_AA = ']';
+        // retrieve pre and post AA, e.g., A and C in "A.PEPTIDE.C" or ".PEPTIDE."
+        preAA = (left_dot == 0 || row[4][left_dot - 1] == '-') ? '[' : row[4][left_dot - 1];  // const char PeptideEvidence::N_TERMINAL_AA = '[';
+        postAA = (right_dot + 1 < row[4].size() || row[4][right_dot + 1] == '-') ? ']' : row[4][right_dot + 1]; // const char PeptideEvidence::C_TERMINAL_AA = ']';
```
timosachsenberg: does this work for EDTGWY(Phospho)WCGIQRDFAR.[324.035869904700007]

PR #4136, opened 2019-05-29
Add icons to visually distinguish spectra, features, consensusfeatures in 2D and 3D view

timosachsenberg:
image


timosachsenberg:
Should we remove the file extension here?
IMHO would look cleaner. Opinions?
Hovering over the layer still shows the full filename (including full path).

hroest:
>  Hovering over the layer still shows the full filename (including full path).

:+1:

PR #4137, opened 2019-05-29
Feature/deisotoper

hroest:
currently the Deisotoper only works if the mono-isotopic peak is the most abundant peak. This assumption is made implicitly and not transparent to the user. I propose the following changes:

- make assumption about decreasing isotopic abundance explicit in the Deisotoper
- add option to remove assumption
- add tests for full peptide spectra, include test for highly charge long peptide where mono-isotopic peak is not the highest

ultimately, a better algorithm for de-isotoping individual spectra would be beneficial in OpenMS


src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -204,6 +204,7 @@ namespace OpenMS
       // deisotope
       Deisotoper::deisotopeAndSingleCharge(exp[exp_index], 
         fragment_mass_tolerance, fragment_mass_tolerance_unit_ppm, 
+        "decreasing", // model
```
timosachsenberg: can we use an enum - not really fan of string parameters

timosachsenberg:
AFAIK it also works if the first or second peak is the highest one.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -113,7 +118,7 @@ void Deisotoper::deisotopeAndSingleCharge(MSSpectrum& spec,
           {
             // Possible improvement: include proper averagine model filtering. for now start at the second peak to test hypothesis
             // Note: this is a common approach used in several other search engines
-            if (old_spectrum[p].getIntensity() > old_spectrum[extensions.back()].getIntensity())
+            if (model_decreasing && old_spectrum[p].getIntensity() > old_spectrum[extensions.back()].getIntensity())
```
hroest: @timosachsenberg how would this work if the second peak is the highest one? Wouldnt it just abort?

hroest:
> AFAIK it also works if the first or second peak is the highest one.

not if the second peak is the highest one:

```
from pyopenms import *
z = 5
seq = AASequence.fromString("DFPIANGERDFPIANGERDFPI")
print(seq.getMonoWeight(Residue.ResidueType.Full, 1)) # weight of M+1
seq_formula = seq.getFormula() + EmpiricalFormula("H" + str(z))
isotopes = seq_formula.getIsotopeDistribution( CoarseIsotopePatternGenerator(8) )

s = MSSpectrum()
for iso in isotopes.getContainer():
  print ("Isotope", iso.getMZ() / z, ":", iso.getIntensity())
  iso.setMZ( iso.getMZ() / z )
  s.push_back(iso)

Deisotoper.deisotopeAndSingleCharge(s, 10, True, 1, 5, True, 2, 10, True, True)
print(s.size())
for p in s:
    print(p.getMZ(), p.getIntensity() )
```

produces the following output:
```
('Isotope', 498.84733507822, ':', 0.24178795516490936)
('Isotope', 499.04800604578, ':', 0.3276079297065735)
('Isotope', 499.24867701334006, ':', 0.23770062625408173)
('Isotope', 499.4493479809, ':', 0.12159404903650284)
('Isotope', 499.65001894846, ':', 0.048920124769210815)
('Isotope', 499.85068991602003, ':', 0.016409706324338913)
('Isotope', 500.05136088358006, ':', 0.004757835529744625)
('Isotope', 500.25203185114003, ':', 0.0012217717012390494)
2
(498.84733507822, 0.24178795516490936)
(2491.210924361816, 0.3276079297065735)
```

and produces 2 peaks instead of a single peak

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
timosachsenberg: hmm, maybe @cbielow changed that?

timosachsenberg: this would clearly be a regression (actually it states in the comment "start at the second peak to test hypothesis")

timosachsenberg: I think it should read:
if (n_extensions >=2)

hroest: Yes I see but I think this is already present in a very early version: 016a63d62ed9ec5440d6ec8ab2897344efc11ba1

Also I think @enetz had similar issues in his code and commented out this part: https://github.com/OpenMS/OpenMS/blob/5c874d295aaa958f67b81f54ee101ce48b4c9394/src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp#L565-L578

maybe it would make more sense to use the `Deisotoper` in his code and just disable this part of the code? Any comment @enetz

hroest:
there is currently multiple deisotoping algorithms in OpenMS, which all seem to attempt similar things, either in 1D or in 2D. I think this is quite confusing: 

- https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp#L98
- https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp#L516
- https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp#L46

as well as DECHARGING algorithms:

- FeatureDeconvolution
- MetaboliteFeatureDeconvolution

and the FEATUREFINDING algorithms (Picked, etc)

hroest:
I think we should attempt to provide the user with options, probably a combination of a simple greedy approach (starting with the first m/z value) and the averagine model for large compounds should probably work for most organic molecules as long as they dont have too crazy patterns (e.g. chlorine etc).

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
enetz: @hroest I have been tinkering with my version of the Deisotoper a bit, e.g. my function reports the number of isotopic peaks found for each pattern and adds up the intensities of all isotopic peaks into the monoisotopic peak. I can try to combine the deisotopers and maybe add two more flags for this additional functionality.

hroest: > @hroest I have been tinkering with my version of the Deisotoper a bit, e.g. my function reports the number of isotopic peaks found for each pattern and adds up the intensities of all isotopic peaks into the monoisotopic peak. I can try to combine the deisotopers and maybe add two more flags for this additional functionality.

yes, this would be great as an option. Thanks!

timosachsenberg:
We could also prevent assigning charge > 1 for fragments with m/z > precursor_mz / charge

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
hroest:
```suggestion
            if (model_decreasing && extensions.size > 2 && old_spectrum[p].getIntensity() > old_spectrum[extensions.back()].getIntensity())
```

@timosachsenberg is that what you had in mind?

timosachsenberg:
closing the old PR

PR #4138, opened 2019-05-29
[FIX] fix drawing of projections if mz is mapped to x-axis

timosachsenberg:
TICs are now always drawn with DM_CONNECTEDLINES


timosachsenberg:
I merge as it fixes an obvious bug

Issue #128, opened 2014-01-09
Auto-Sort layers by type in TOPPView [128]


timosachsenberg:
Submitted by cbielow on 2010-03-03 15:09:32

when adding raw data as a layer, it should be displayed below any feature, consensus or idXML layer, as these will be hard to read otherwise.

timosachsenberg:
could be easily implemented in Spectrum2DCanvas paint event. Instead of iterating over each layer in a for loop we could first determine a sensible order (peakmap, features, ids) and paint int that order.

PR #4139, opened 2019-05-29
[FIX] always keep order: spectra, features, consensusfeatures, ids in…

timosachsenberg:
… layer manager

It doesn't make sense to draw features e.g., below spectra so we keep a fixed order:
spectra, features, consensus features, ids

@hroest or @oliveralka I did not test it on chromatograms. Would be cool is someone could give it a try

### DIFF
--- a/src/openms_gui/source/VISUAL/SpectrumCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/SpectrumCanvas.cpp
@@ -398,15 +398,17 @@ namespace OpenMS
     new_layer.setPeakData(map);
     new_layer.setOnDiscPeakData(od_map);
 
-    if (new_layer.getPeakData()->getChromatograms().size() != 0 
-        && new_layer.getPeakData()->size() != 0)
+    // both empty
+    if (!new_layer.getPeakData()->getChromatograms().empty() 
+     && !new_layer.getPeakData()->empty())
     {
       // TODO : handle this case better
       OPENMS_LOG_WARN << "Your input data contains chromatograms and spectra, falling back to display spectra only." << std::endl;
     }
 
-    if (new_layer.getPeakData()->getChromatograms().size() != 0 
-        && new_layer.getPeakData()->size() == 0)
+    // check which one is empty
+    if (!new_layer.getPeakData()->getChromatograms().empty() 
+      && new_layer.getPeakData()->empty())
     {
       new_layer.type = LayerData::DT_CHROMATOGRAM;
     }
@@ -415,19 +417,14 @@ namespace OpenMS
       new_layer.type = LayerData::DT_PEAK;
     }
 
-    // insert after last peak or chomatogram layer
-    for (auto it = layers_.rbegin(); it != layers_.rend(); ++it)
-    {
-      if (it->type == LayerData::DT_PEAK 
-        || it->type == LayerData::DT_CHROMATOGRAM)
-      {
-        layers_.insert(it.base(), new_layer);
-        return finishAdding_();
-      }
-    }
+    // insert after last layer of same type, 
+    // if there is no such layer after last layer of previous types, 
+    // if there are no layers at all put at front
+    auto it = std::find_if(layers_.rbegin(), layers_.rend(), [](const LayerData& l) 
+      { return l.type <= LayerData::DT_PEAK; });
+      
+    layers_.insert(it.base(), std::move(new_layer));
 
-    // insert at front
-    layers_.insert(layers_.begin(), new_layer);
     return finishAdding_();
   }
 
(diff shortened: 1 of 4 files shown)


timosachsenberg:
also removes the file extension from the layer view (we will have icons from the other PR)

timosachsenberg:
@cbielow @hroest the code works but I think its efficiency heavily depends on whether feature maps, vector of peptideids and consensus map can be moved as I insert with: layers_.insert(it.base(), new_layer);
These are not movable right now iirc.

timosachsenberg:
addresses https://github.com/OpenMS/OpenMS/issues/128

src/openms_gui/source/VISUAL/SpectrumCanvas.cpp
```diff
@@ -391,60 +391,120 @@ namespace OpenMS
...
+      new_layer.type = LayerData::DT_PEAK;
+    }
+
+    // insert after last peak or chomatogram layer
+    for (auto it = layers_.rbegin(); it != layers_.rend(); ++it)
+    {
+      if (it->type == LayerData::DT_PEAK 
+        || it->type == LayerData::DT_CHROMATOGRAM)
+      {
+        layers_.insert(it.base(), new_layer);
```
hroest: in order to get the move insert , e.g. https://en.cppreference.com/w/cpp/container/vector/insert (2) you need to do this to at least prepare for this option:  
```suggestion
        layers_.insert(it.base(), std::move(new_layer));
```

hroest: PS: T must meet the requirements of MoveAssignable and MoveInsertable in order to use overload (2).

hroest: > PS: T must meet the requirements of MoveAssignable and MoveInsertable in order to use overload (2).

so this means that `LayerData` should have a move assignment operator and either a move constructor or a copy constructor that is `const -ref`. It currently has neither. Another idea would be that the layer would keep track of `shared_ptr` of the data and no data ever gets copied when the layer gets copied

hroest:
> @cbielow @hroest the code works but I think its efficiency heavily depends on whether feature maps, vector of peptideids and consensus map can be moved as I insert with: layers_.insert(it.base(), new_layer);
> These are not movable right now iirc.

currently you are copying an object of type `LayerData` which contains shared-ptrs to the real data, the only data you really copy is `Size current_spectrum_` and `ExperimentType::SpectrumType cached_spectrum_` so not the whole experiment. I think you should be fine.

src/openms_gui/source/VISUAL/SpectrumCanvas.cpp
```diff
@@ -391,60 +391,120 @@ namespace OpenMS
...
-    if (layers_.back().getPeakData()->getChromatograms().size() != 0 
-        && layers_.back().getPeakData()->size() != 0)
+    LayerData new_layer;
+    new_layer.param = param_;
+    new_layer.filename = filename;
+    new_layer.name = QFileInfo(filename.toQString()).baseName();
+    new_layer.setPeakData(map);
+    new_layer.setOnDiscPeakData(od_map);
+
+    if (new_layer.getPeakData()->getChromatograms().size() != 0
```
cbielow: use `if (!new_layer.getPeakData()->getChromatograms().empty() && !new_layer.getPeakData().getSpectra()->empty()` for clarity?

src/openms_gui/source/VISUAL/SpectrumCanvas.cpp
```diff
@@ -391,60 +391,120 @@ namespace OpenMS
...
+    {
+      if (it->type == LayerData::DT_PEAK 
+        || it->type == LayerData::DT_FEATURE)
+      {
+        layers_.insert(it.base(), new_layer);
+        return finishAdding_();
+      }
+    }
+    // insert at front
+    layers_.insert(layers_.begin(), new_layer);
```
cbielow: The need to check for all previous layer types seems very brittle and easy to get wrong.
How about we introduce an ordering here. The enum `LayerData::DataType` would be good, you just need to reorder the elements to the order we want and document that the order has meaning.
Then its very short to write (untested code ahead):
```
auto it = std::find_if(layers_.rbegin(), layers_.rend(), [](const LayerData& l) 
      { return l.type <= LayerData::DT_FEATURE; };
layers_.insert(it, std::move(new_layer);
return finishAdding_();
```

src/openms_gui/source/VISUAL/SpectrumCanvas.cpp
cbielow: The bigger problem here is the `insert()`, which will copy all layers behind the insertion point to their new destination (or copy the whole vector upon resize). All because we lack the move assignment. So this would be really nice to have.

src/openms_gui/source/VISUAL/SpectrumCanvas.cpp
timosachsenberg: Yeah I though about that and will probably implement this.

timosachsenberg:
I think I addressed all comments. Having movable  PepIds would be nice in the future but is currently not an urgent  issue

PR #4144, opened 2019-06-05
Init isotopes only when needed and not yet init.

### LINKED ISSUE #4072: CompNovoCID write_ini test takes over a minute
In debug mode. Therefore the test timeouts on our build machines.
This should be impossible for simply writing the ini file, there has to be something wrong.
The same happens for CompNovo itself and I think also ConsensusID but less pronounced.
I do not have admin rights on the build machines, so I can't do performance analysis. Someone with windows has to check this. @cbielow ? You like to play around with Perf anyway right? Maybe you could have a short look next time you are in a Debug build.

jpfeuffer:
fixes #4072


jpfeuffer:
mac only failed during upload to cdash.

timosachsenberg:
+1

PR #4147, opened 2019-06-10
[FIX-#4141] Fixed PercolatorAdapter missing columns bug in DOC mode. …

### LINKED ISSUE #4141: rt and dm columns for percolator
Hello, when using PercolatorAdapter with the option for "description of correct features", the following error message appears: ERROR: Could not find column with name "rt"/"retentiontime" and "dm"/"deltamass" necessary for the calculation of DOC features.

How can these columns be added? I tried IDFileConverter with an additional mzML input, but to no avail.

Thanks for any support!

MatthewThe:
…Fixes #4141

Adds columns for delta mass and retention time which are needed if the "description of correct" feature is used in percolator.

### DIFF
--- a/src/topp/PercolatorAdapter.cpp
+++ b/src/topp/PercolatorAdapter.cpp
@@ -409,6 +409,7 @@ protected:
       Int scan_number = getScanNumber_(scan_identifier);
       
       double exp_mass = it->getMZ();
+      double retention_time = it->getRT();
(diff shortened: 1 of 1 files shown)


timosachsenberg:
Thanks!

oliverschilling:
Thanks a lot! Will this be accessible in a "nightly build"?

PR #4150, opened 2019-06-13
[FIX-#4140] Fixed PSMFeatureExtractor unknown output format issue. 

### LINKED ISSUE #4140: PSMFeatureExtractor crashes
Hello; PSMFeatureExtractor crashes for a search with msgf+; no ,modifications other iodoacteamide. Based on some other reports, I am have also tried to include IDFilter and other measures to prevent that non-mapped peptides from PeptideIndexer are piped to PSMFeatureExtractor. Output / error message is:
Loading input file: /tmp/20190531_152416_12646_1/msgf-percolator_idxml-input/005_IDFilter/out/java.idXML
MS-GF+ PSM with missing NumMatchedMainIons skipped.
Error: File not found (the file '/tmp/20190531_152416_12646_1/msgf-percolator_idxml-input/006_PSMFeatureExtractor/out/java.unknown' could not be found)
<MS-GF+ PSM with missing NumMatchedMainIons skipped.> occurred 4 times

Thanks for support!

MatthewThe:
Fixes #4140

Adds out_type option to PSMFeatureExtractor to resolve ambiguity for idXML or mzid output, specifically in the interface between PSMFeatureExtractor and PercolatorAdapter.

### DIFF
--- a/src/utils/PSMFeatureExtractor.cpp
+++ b/src/utils/PSMFeatureExtractor.cpp
@@ -114,7 +114,9 @@ protected:
     registerInputFileList_("in", "<files>", StringList(), "Input file(s)", true);
     setValidFormats_("in", ListUtils::create<String>("mzid,idXML"));
     registerOutputFile_("out", "<file>", "", "Output file in mzid or idXML format", true);
-    setValidFormats_("out", ListUtils::create<String>("mzid,idXML"));
+    setValidFormats_("out", ListUtils::create<String>("mzid,idXML"));    
+    registerStringOption_("out_type", "<type>", "", "Output file type -- default: determined from file extension or content.", false);
+    setValidStrings_("out_type", ListUtils::create<String>("mzid,idXML"));
     registerStringList_("extra", "<MetaData parameter>", vector<String>(), "List of the MetaData parameters to be included in a feature set for precolator.", false, false);
     // setValidStrings_("extra", ?);
     // TODO: add this MHC feature back in with TopPerc::hasMHCEnd_()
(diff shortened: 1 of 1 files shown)


src/utils/PSMFeatureExtractor.cpp
```diff
@@ -114,7 +114,9 @@ class PSMFeatureExtractor :
     registerInputFileList_("in", "<files>", StringList(), "Input file(s)", true);
     setValidFormats_("in", ListUtils::create<String>("mzid,idXML"));
     registerOutputFile_("out", "<file>", "", "Output file in mzid or idXML format", true);
-    setValidFormats_("out", ListUtils::create<String>("mzid,idXML"));
+    setValidFormats_("out", ListUtils::create<String>("mzid,idXML"));    
+    registerStringOption_("out_type", "<type>", "", "Output file type -- default: determined from file extension or content.", false);
+    setValidStrings_("out_type", ListUtils::create<String>("mzid,idXML"));
```
timosachsenberg: @jpfeuffer is this automatically removed for KNIME nodes or do we need to add that?

jpfeuffer: In KNIME this is automatically added with the setValidFormats above. This appears
in the OutputTypes Tab of the Dialog and adds the appropriate ending to the filename that is generated. Then the tool should be able to infer the correct type. If I am not mistaken, this should have worked for KNIME all the time already. The out_type parameters though does no harm except that you could potentially now give an idXML filename and write mzID.

timosachsenberg: ah ok I thought we filter these params from the config dialog inKNIME

PR #4153, opened 2019-06-17
New OpenSWATH  parameters

hroest:


src/utils/OpenSwathWorkflow.cpp
```diff
@@ -467,8 +467,11 @@ class TOPPOpenSwathWorkflow
     registerInputFile_("swath_windows_file", "<file>", "", "Optional, tab-separated file containing the SWATH windows for extraction: lower_offset upper_offset. Note that the first line is a header and will be skipped.", false, true);
     registerFlag_("sort_swath_maps", "Sort input SWATH files when matching to SWATH windows from swath_windows_file", true);
 
-    registerFlag_("use_ms1_traces", "Extract the precursor ion trace(s) and use for scoring", true);
-    registerFlag_("enable_uis_scoring", "Enable additional scoring of identification assays", true);
+    registerStringOption_("use_ms1_traces", "<name>", "true", "Extract the precursor ion trace(s) and use for scoring if present", false, true);
+    setValidStrings_("use_ms1_traces", ListUtils::create<String>("true,false"));
```
hroest: @grosenberger can you look at these parameter settings and cross-check with your settings please

src/utils/OpenSwathWorkflow.cpp
```diff
@@ -467,8 +467,11 @@ class TOPPOpenSwathWorkflow
...
+
+    registerStringOption_("enable_uis_scoring", "<name>", "true", "Enable additional scoring of identification assays using IPF (see online documentation)", false, true);
+    setValidStrings_("enable_uis_scoring", ListUtils::create<String>("true,false"));
```
hroest: @grosenberger this had some interesting side effects, see https://github.com/OpenMS/OpenMS/issues/4268

grosenberger: Since we are changing the default parameters, we could consider to further unify the flag names. For example, we could replace ``-enable_uis_scoring`` with ``-enable_ipf`` to be consistent with OpenSwathAssayGenerator. We could also consider replace ``-use_ms1_traces`` with ``-enable_ms1``, but that is of course only cosmetics and with the new defaults, it shouldn't be required frequently anyway.

timosachsenberg:
@hroest good to merge?

timosachsenberg:
what about the codeship stuff?

hroest:
@grosenberger are we ready to merge. to me it looks like we are close

grosenberger (approved):
Looks great, thanks for the changes!

timosachsenberg:
I can clean up the circle-ci part

hroest:
> I can clean up the circle-ci part or is this intended to get docker images?

yes, you can clean it up. This doesnt need to be here.

PR #4159, opened 2019-06-21
[FIX] Fixed Percolator PTM notation and default svm score

MatthewThe:
* Changed the peptide sequence PTM notation to use square brackets instead of round brackets, which is the notation used in Percolator.
* Changed the default SVM score for unmatched PSMs to -100.0, as the previous value of 0.0 is the score for PSMs at the FDR threshold. Typically the lowest scoring PSM has a SVM score of about -4 or -10, depending on the search engine.
* Changed the log message of unmatched PeptideHits into a debug message. In a regular percolator run this message pops up thousands of times when multiple PSMs are reported per spectrum, even though nothing is wrong.

### DIFF
--- a/src/topp/PercolatorAdapter.cpp
+++ b/src/topp/PercolatorAdapter.cpp
@@ -497,7 +497,8 @@ protected:
 
         sequence += aa_before;
         sequence += "."; 
-        sequence += hit.getSequence().toString();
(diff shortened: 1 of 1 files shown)


timosachsenberg:
Thanks! will test on a larger set of data

PR #4160, opened 2019-06-24
[Feature] Print Stdout/Err of external commands when debug mode is on

jpfeuffer:
Just had a hard time finding the cause of an issue without it.

### DIFF
--- a/src/openms/source/APPLICATIONS/TOPPBase.cpp
+++ b/src/openms/source/APPLICATIONS/TOPPBase.cpp
@@ -1652,16 +1652,27 @@ namespace OpenMS
       return EXTERNAL_PROGRAM_ERROR;
     } 
 
-    if (success == false || qp.exitStatus() != 0 || qp.exitCode() != 0)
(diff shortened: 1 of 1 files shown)


src/openms/source/APPLICATIONS/TOPPBase.cpp
```diff
@@ -1652,16 +1652,27 @@ namespace OpenMS
       return EXTERNAL_PROGRAM_ERROR;
     } 
 
-    if (success == false || qp.exitStatus() != 0 || qp.exitCode() != 0)
+    bool any_failure = success == false || qp.exitStatus() != 0 || qp.exitCode() != 0;
```
hroest:
```suggestion
    bool any_failure = (success == false || qp.exitStatus() != 0 || qp.exitCode() != 0);
```

hroest:
:+1: feel free to merge if tests pass

hroest:
travis passes, I assume this is fine

PR #4163, opened 2019-06-25
[Feature] Param 1.7.0 support

jpfeuffer:
Mainly a cosmetic change since there is a bool parameter type now.
Internally I left it to a String representation since it is quite deeply anchored in our tools like that.


jpfeuffer:
Don't merge though. I only want flags to get the bool notation.

jpfeuffer:
OK I changed it to our current default. Designate it a flag, when it is a Param object of type String with restrictions true/false and default value false.
This won't work if you want to write current (non-default) settings where you e.g. have set the value of such a "flag" to true already. In this case the ini/ctd will not be a correct representation of the equivalent command line. In this case an ini has to be passed back to the tool to correctly represent its state. But before we just ignored any differences between those two types in the ini anyway, so nothing changed.
This could be solved with an additional tag of the parameter but requires more work and adaption of many tools, so I will not do it here.

jpfeuffer:
Merge? @timosachsenberg

jpfeuffer:
Mac was uploading error only and the new Win machine currently has problems due to boost and probably a virus scan making it slow.

Issue #4156, opened 2019-06-19
Clean-up IDFilter params


oliveralka:
Hello everyone, 

I had some problems running the IDFilter due to its abundance on parameters with their respective sometimes cryptic explanation. 

For example, the "unique" option, which I think should always mean "unique per protein". Seems to remove duplicate peptide hits for a peptide identification (not sure how often that happens).

Additional the "thresh" and "score" possibility, where usually only "score" is used to filter based on an FDR threshold. 

It would be great if we could collect some ideas, which params are usually used and which should be renamed and marked as advanced or even be deprecated. Please let me know what you think.   

Related see #2464, #4089

hroest:
there is clearly too many options in the regular output: https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/TOPP_IDFilter.html

also the online documentation could be improved using a better description or even potentially a few example use cases of how this could be used to achieve a certain result

hroest:
Btw: I just opened https://github.com/OpenMS/OpenMS/pull/4157 so that now a link to the online docu should appear to help people with the plethora of options

hroest:
@oliveralka I started some cleanup here: https://github.com/OpenMS/OpenMS/tree/feature/idfilter_params

This boils down the parameters a lot, I think we could also have the whitelist/blacklist parameters go. What do you think?

jpfeuffer:
I'm all up for clean up but I would probably keep a white/black listing functionality. If not in Idfilter, then in a separate tool. I've seen quite some people using it.
We could rename it to "match_reference" and have a bool for include or exclude to trim it down a little.

PR #4166, opened 2019-06-26
Clean-Up IDFilter parameters 

oliveralka:
Based on the branch from Hannes, with some minor changes. 
Please let me know if you think further changes should be introduced. 

I think this makes it cleaner than before and the important parameters are all easily accessible. 

Related to / fixes: #4156


src/topp/IDFilter.cpp
```diff
@@ -227,19 +229,16 @@ class TOPPIDFilter :
...
     registerStringOption_("best:n_to_m_peptide_hits", "[min]:[max]", ":", "Peptide hit rank range to extracts", false, true);
 
-    registerStringOption_("length", "[min]:[max]", ":", "Keep only peptide hits with a sequence length in this range.", false);
-
-    registerStringOption_("charge", "[min]:[max]", ":", "Keep only peptide hits with charge states in this range.", false);
 
     registerFlag_("var_mods", "Keep only peptide hits with variable modifications (as defined in the 'SearchParameters' section of the input file).", false);
 
-    registerFlag_("unique", "If a peptide hit occurs more than once per peptide ID, only one instance is kept.");
+    registerFlag_("unique_per_spectrum", "If a peptide-spectrum match occurs more than once for a spectrum, only one instance is kept.", true);
```
jpfeuffer: Which one? First? Best by score?

src/topp/IDFilter.cpp
```diff
@@ -180,29 +182,29 @@ class TOPPIDFilter :
...
     registerInputFile_("blacklist:peptides", "<file>", "", "Peptides with the same sequence and modification assignment as any peptide in this file are filtered out. Use with 'blacklist:ignore_modifications' to only compare by sequence.\n", false);
     setValidFormats_("blacklist:peptides", ListUtils::create<String>("idXML"));
-    registerFlag_("blacklist:ignore_modifications", "Compare blacklisted peptides by sequence only.", false);
-    registerStringList_("blacklist:modifications", "<selection>", vector<String>(), "Remove all peptides with sequences that contain (any of) the selected modification(s)", false);
+    registerFlag_("blacklist:ignore_modifications", "Compare blacklisted peptides by sequence only.", true);
+    registerStringList_("blacklist:modifications", "<selection>", vector<String>(), "Remove all peptides with sequences that contain (any of) the selected modification(s)", false, true);
     setValidStrings_("blacklist:modifications", all_mods);
 
     registerTOPPSubsection_("in_silico_digestion", "This filter option removes peptide hits which are not in the list of in silico peptides generated by the rules specified below");
```
jpfeuffer: What is in_silico_digestion filter doing? Isn't it the same as running PeptideIndexer and then filter the unmatched ones?

src/topp/IDFilter.cpp
```diff
@@ -227,19 +229,16 @@ class TOPPIDFilter :
...
 
-    registerStringOption_("length", "[min]:[max]", ":", "Keep only peptide hits with a sequence length in this range.", false);
-
-    registerStringOption_("charge", "[min]:[max]", ":", "Keep only peptide hits with charge states in this range.", false);
 
     registerFlag_("var_mods", "Keep only peptide hits with variable modifications (as defined in the 'SearchParameters' section of the input file).", false);
 
-    registerFlag_("unique", "If a peptide hit occurs more than once per peptide ID, only one instance is kept.");
+    registerFlag_("unique_per_spectrum", "If a peptide-spectrum match occurs more than once for a spectrum, only one instance is kept.", true);
     registerFlag_("unique_per_protein", "Only peptides matching exactly one protein are kept. Remember that isoforms count as different proteins!");
```
jpfeuffer: I am not a big fan of unique_per_protein, too.
Wouldn't "remove_shared_peptides" or "remove_ambiguous_peptides" be easier to understand?

jpfeuffer:
What about the "thresh" functions. What are they actually doing?

src/topp/IDFilter.cpp
oliveralka: "remove_ambiguous_peptides" - this maybe.

But "remove_shared_peptides" is usually called "unique" - I am not sure what would be the best option.

oliveralka:
I think Timo can give a short explanation of why the "thresh" was introduced. I am not sure if we need to keep it. @timosachsenberg

src/topp/IDFilter.cpp
oliveralka: first occurrence (will add that see IDFilter.h comment).

src/topp/IDFilter.cpp
oliveralka: Good question.

src/topp/IDFilter.cpp
jpfeuffer: Because unique_per_protein (especially with the meaning of the option right above it) makes me definitely think about removing all but one peptide per protein.

timosachsenberg:
I think the thresh parameter was introduced a long time ago and the intention was to
store the score threshold while keeping all data.
Kind of what mzIdentML also can do.
I personally would get rid of it.

jpfeuffer:
Isn't that implicit once you use a score that can be thresholded like an FDR or q-value?
Totally in favour of getting rid of it.

src/topp/IDFilter.cpp
timosachsenberg: agree with Julianus.
What is the problem with remove_shared_peptides - seems best fit to me?

oliveralka: There is no problem per se - we can go with it. The only thing I see is that in a lot of other software it is just called "unique" and this may confuse some users.

timosachsenberg: Yes, I see but this is dangerous if the user opens an old workflow where "unique" had a different meaning. Don't know if that will be an issue but if it is, it will be super hard to debug for users.

oliveralka: Ok, then we should go with the suggestion of Julianus.

src/topp/IDFilter.cpp
oliveralka: I think in general there should be no difference, but these two options are not supported by the PeptideIdexer and filtering.

```
  -in_silico_digestion:missed_cleavages <integer>                 Range of allowed missed cleavages in the peptide sequences
                                                                  By default missed cleavages are ignored (default: '-1' min: '-1')
  -in_silico_digestion:methionine_cleavage                        Allow methionine cleavage at the N-terminus of the protein
```

src/topp/IDFilter.cpp
```diff
@@ -149,16 +146,15 @@ class TOPPIDFilter :
     registerOutputFile_("out", "<file>", "", "output file ");
     setValidFormats_("out", ListUtils::create<String>("idXML"));
 
-    registerTOPPSubsection_("precursor", "Filtering by precursor RT or m/z");
+    registerTOPPSubsection_("precursor", "Filtering by precursor attributes (RT, m/z, charge, length)");
     registerStringOption_("precursor:rt", "[min]:[max]", ":", "Retention time range to extract.", false);
     registerStringOption_("precursor:mz", "[min]:[max]", ":", "Mass-to-charge range to extract.", false);
+    registerStringOption_("length", "[min]:[max]", ":", "Keep only peptide hits with a sequence length in this range.", false);
+    registerStringOption_("charge", "[min]:[max]", ":", "Keep only peptide hits with charge states in this range.", false);
```
hroest: this should probably change to `precursor:charge` and `precursor:length` since it will display correctly (and make sense?)

src/topp/IDFilter.cpp
```diff
@@ -227,19 +223,16 @@ class TOPPIDFilter :
...
 
-    registerStringOption_("length", "[min]:[max]", ":", "Keep only peptide hits with a sequence length in this range.", false);
-
-    registerStringOption_("charge", "[min]:[max]", ":", "Keep only peptide hits with charge states in this range.", false);
 
     registerFlag_("var_mods", "Keep only peptide hits with variable modifications (as defined in the 'SearchParameters' section of the input file).", false);
 
-    registerFlag_("unique", "If a peptide hit occurs more than once per peptide ID, only one instance is kept.");
-    registerFlag_("unique_per_protein", "Only peptides matching exactly one protein are kept. Remember that isoforms count as different proteins!");
+    registerFlag_("remove_ambiguous_peptides", "If a peptide-spectrum match occurs more than once for a spectrum, only the first instance is kept.", true);
```
hroest:
```suggestion
    registerFlag_("remove_ambiguous_peptides", "Only the best PSM (peptide spectrum match) is kept per peptide in case there are multiple matching spectra.", true);
```

is that correct? It is unclear wha the "first instance" is - is it the best hit? Also we should be clear whether we keep the best peptide per PSM *or* the best PSM per peptide id? which one is it?

hroest:
```suggestion
    registerFlag_("remove_ambiguous_peptides", "Keep only the best PSM (peptide spectrum match) per spectrum and remove lower scoring matches.", true);
```

I think that is correct? actually the docu of the fxn is ambigous even...

oliveralka: `@brief Removes duplicate peptide hits from each peptide identification, keeping only unique hits (per ID). By default, hits are considered duplicate if they compare as equal using PeptideHit::operator==. However, if @p seq_only is set, only the sequences (incl     . modifications) are compared. In both cases, the first occurrence of each hit in a peptide ID is kept, later ones are removed.`

So it seems it will only remove duplicated PSMs for a spectrum - and leaves the others. So you may still have a list of PSMs for the spectrum. - Is that correct? 

"Removes duplicated PSMs per spectrum and retrains the one with the higher score." (The one with the higher score is the one which occurs first - at least that is what I think.) 


There is another option which is called "keepNBestHits" which can be used to only use the best PSM per Spectrum.

hroest: ok I see, so this is quite confusing especially with a flag called "ambigous peptides" - what is ambigous about that? Why are there even duplicates? maybe call the flag "remove_duplicate_psm" ?

src/topp/IDFilter.cpp
```diff
@@ -227,19 +223,16 @@ class TOPPIDFilter :
...
 
-    registerStringOption_("length", "[min]:[max]", ":", "Keep only peptide hits with a sequence length in this range.", false);
-
-    registerStringOption_("charge", "[min]:[max]", ":", "Keep only peptide hits with charge states in this range.", false);
 
     registerFlag_("var_mods", "Keep only peptide hits with variable modifications (as defined in the 'SearchParameters' section of the input file).", false);
 
-    registerFlag_("unique", "If a peptide hit occurs more than once per peptide ID, only one instance is kept.");
-    registerFlag_("unique_per_protein", "Only peptides matching exactly one protein are kept. Remember that isoforms count as different proteins!");
+    registerFlag_("remove_ambiguous_peptides", "Removes duplicated PSMs per spectrum and retrains the one with the higher score.", true);
```
hroest:
```suggestion
    registerFlag_("remove_ambiguous_peptides", "Removes duplicated PSMs per spectrum and retains the one with the higher score.", true);
```

hroest: small typo

oliveralka:
@jpfeuffer Should be alright, or?

hroest:
looks good :+1:

jpfeuffer:
Yes much better. Thanks!

PR #4170, opened 2019-07-04
[BUILD] add nested parallelism to cmake interface

hroest:
allow setting of nested OMP using CMake


timosachsenberg:
@jpfeuffer the failure on ubuntu is unrelated I guess

jpfeuffer:
yes I think so

PR #4172, opened 2019-07-04
Feature/osw speed

hroest:
several improvements for OpenSwath under highly parallel loads

- memory improvements: instead of loading all MS1 chromatograms into memory, only load those of the current batch
- speed improvements: store scores in MRMFeature not in a map any more but as meta-values
- speed improvements: store paramters as members (e.g. `sn_win_len_`)
- allow compilation option `MT_ENABLE_NESTED_OPENMP` to utilize nested parallelism in OpenSWATH

speed improvements on a heavy parallel load with 40 threads:

before:

```
 OpenSwathWorkflow took 19:14 m (wall), 07:52:12 h (CPU), 02:39 m (system), 07:49:32 h (user).
 28172.61user 160.30system 19:19.99elapsed 2442%CPU (0avgtext+0avgdata 11329876maxresident)k
 0inputs+0outputs (0major+36875216minor)pagefaults 0swaps
```

after:

```
OpenSwathWorkflow took 05:38 m (wall), 02:03:15 h (CPU), 02:14 m (system), 02:01:00 h (user).
7260.90user 135.00system 5:44.10elapsed 2149%CPU (0avgtext+0avgdata 6870080maxresident)k
0inputs+0outputs (0major+21793570minor)pagefaults 0swaps
```

* speed is improved 3-fold
* memory requirements are reduced by 60%

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,11 @@ option(HAS_XSERVER "Indicates if an X server is available. If set to Off it will
 option(ENABLE_DOCS "Indicates whether documentation should be built." ON)
 option(ENABLE_TUTORIALS "Indicates whether tutorials should be build. Note that this also depends on the availability of (pdf)latex." ON)
 option(WITH_GUI "Build GUI parts of OpenMS (TOPPView&Co). This requires QtGui." ON)
+if(MSVC)
+  option(MT_ENABLE_NESTED_OPENMP "Enable nested parallelism." OFF)
+else()
+  option(MT_ENABLE_NESTED_OPENMP "Enable nested parallelism." ON)
+endif()
 
 #------------------------------------------------------------------------------
 # Extend module path with our modules
--- a/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
@@ -570,7 +570,7 @@ namespace OpenMS
         // Call the scoring for MS1 only
         ///////////////////////////////////
 
-        OpenSwath_Scores scores;
+        OpenSwath_Scores& scores = mrmfeature->getScores();
         precursor_mz = mrmfeature->getMZ();
 
         // S/N scores
@@ -663,7 +663,7 @@ namespace OpenMS
           precursor_ids.push_back(precursor_id);
         }
 
-        OpenSwath_Scores scores;
+        OpenSwath_Scores& scores = mrmfeature->getScores();
         scorer.calculateChromatographicScores(imrmfeature, native_ids_detection, precursor_ids, normalized_library_intensity,
                                               signal_noise_estimators, scores);
 
@@ -935,7 +935,7 @@ namespace OpenMS
       pep_hit_.setScore(xx_lda_prescore);
       if (swath_present)
       {
-        pep_hit_.setScore(mrmfeature->getScore("xx_swath_prelim_score"));
+        // pep_hit_.setScore(mrmfeature->getMetaValue("xx_swath_prelim_score"));
       }
 
       if (pep->isPeptide())
@@ -994,6 +994,7 @@ namespace OpenMS
       mrmfeature->setMetaValue("peak_apices_sum", total_peak_apices);
       mrmfeature->setMetaValue("ms1_area_intensity", ms1_total_intensity);
       mrmfeature->setMetaValue("ms1_apex_intensity", ms1_total_peak_apices);
+      mrmfeature->setMetaValue("xx_swath_prelim_score", 0.0);
       feature_list.push_back((*mrmfeature));
 
       delete imrmfeature;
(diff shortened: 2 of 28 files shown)


src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -935,7 +932,7 @@ namespace OpenMS
       pep_hit_.setScore(xx_lda_prescore);
       if (swath_present)
       {
-        pep_hit_.setScore(mrmfeature->getScore("xx_swath_prelim_score"));
+        // pep_hit_.setScore(mrmfeature->getMetaValue("xx_swath_prelim_score"));
```
timosachsenberg: what about this one?

hroest: good point, is fixed now (this caused issues when the meta value was not present)

hroest:
failures are unrelated, @timosachsenberg merge?

CMakeLists.txt
```diff
@@ -81,6 +81,11 @@ option(HAS_XSERVER "Indicates if an X server is available. If set to Off it will
 option(ENABLE_DOCS "Indicates whether documentation should be built." ON)
 option(ENABLE_TUTORIALS "Indicates whether tutorials should be build. Note that this also depends on the availability of (pdf)latex." ON)
 option(WITH_GUI "Build GUI parts of OpenMS (TOPPView&Co). This requires QtGui." ON)
+if(MSVC)
+  option(MT_ENABLE_NESTED_OPENMP "Enable nested parallelism." OFF)
```
cbielow: MSVC does not support it at all, if I remember correctly?

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -312,7 +312,7 @@ namespace OpenMS
                                                                  const unsigned int sn_bin_count_,
                                                                  const double det_intensity_ratio_score,
                                                                  const double det_mi_ratio_score,
-                                                                 bool write_log_messages,
+                                                                 bool write_log_messages_,
```
cbielow: what does the naming imply here? Its not a member in the traditional sense.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -1026,6 +1024,10 @@ namespace OpenMS
     uis_threshold_peak_area_ = param_.getValue("uis_threshold_peak_area");
     scoring_model_ = param_.getValue("scoring_model");
 
+    sn_win_len_ = (double)param_.getValue("TransitionGroupPicker:PeakPickerMRM:sn_win_len");
```
cbielow: why the extra cast?
There are operator()'s for DataValue which do this implicitly.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -34,12 +34,32 @@
...
-
 // OpenSwathCalibrationWorkflow
 namespace OpenMS
 {
 
+  OpenSwath::SpectrumAccessPtr loadMS1Map(const std::vector< OpenSwath::SwathMap > & swath_maps, bool load_into_memory)
+  {
+    OpenSwath::SpectrumAccessPtr ms1_map;
+    // store reference to MS1 map for later -> note that this is *not* threadsafe!
+    for (SignedSize i = 0; i < boost::numeric_cast<SignedSize>(swath_maps.size()); ++i)
```
cbielow: replace with ranged based for
```
for (auto& map : swath_maps) 
{ ...
}
```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -781,12 +802,17 @@ namespace OpenMS
                   swath_maps[i].lower < coordinates[j].mz && swath_maps[i].upper > coordinates[j].mz)
                 )
             {
-              // write MS1 chromatograms to disk
-              chromConsumer->consumeChromatogram( ms1_chromatograms[j] );
+#ifdef _OPENMP
```
cbielow: no need to ifdef an OpenMP pragma. This is only required for non-preprocessor OpenMP functionality.
Removing the ifdef improves readability of code.

cbielow:
The results seem to be different. Is this expected (and are they still correct?)

CMakeLists.txt
hroest: exactly, they only support quite old versions of OpenMP

src/tests/topp/OpenSwathWorkflow_22_output.chrom.mzML
```diff
@@ -425,11 +425,11 @@
...
-		<offset idRef="tr2">14767</offset>
-		<offset idRef="3">17082</offset>
+		<offset idRef="tr1">7860</offset>
+		<offset idRef="tr2">10175</offset>
+		<offset idRef="3">12490</offset>
+		<offset idRef="PEPTIDEB_Precursor_i0">14803</offset>
+		<offset idRef="PEPTIDEB_Precursor_i0">17099</offset>
 		<offset idRef="4">19395</offset>
 		<offset idRef="5">21708</offset>
 		<offset idRef="6">24021</offset>
```
hroest: @cbielow the output is the same its just a re-ordering of the MS1 chromatograms. They should be the same because the offsets map to exactly the same byte positions in the file, e.g. the chromatogram 4 is at byte 19395 before and after

PR #4178, opened 2019-07-09
[FEATURE] 2% performance increase reading mzML 

cbielow:
by getting rid of static init check (lock) and heap alloc! 

whhhoohaaa! :)

Reading speed: 27 sec vs 26.5 sec (reproducible) for 1.8 GB file

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -667,32 +667,31 @@ namespace OpenMS
 
     void MzMLHandler::startElement(const XMLCh* const /*uri*/, const XMLCh* const /*local_name*/, const XMLCh* const qname, const xercesc::Attributes& attributes)
     {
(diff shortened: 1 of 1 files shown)


hroest:
:+1: nice, but this does not seem to work:



/.../source/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp:670:35: error: int-array initialized from incompatible wide string

       constexpr XMLCh s_count[] = L"count";

cbielow:
seems that XMLCh is a uint16 on Linux and wchar_t on Windows, hence the `L literal` approach only works on Windows (wchar_t on Linux is 32 bit, and probably the reason why xerces is not using it).
There are no ways I could find using C++11 to convert a char[] to int[] during compile time (and even in C14 it looks very ugly). So we use implicit conversion using char initializer lists.

btw: performance increase on GCC 7.2 is also ~2% (nothing to write home about, but hey)

hroest:
rebuild jenkins

PR #4179, opened 2019-07-10
Feature/osw cleanup

hroest:


src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -93,6 +93,8 @@ namespace OpenMS
     defaults_.setMinFloat("quantification_cutoff", 0.0);
     defaults_.setValue("write_convex_hull", "false", "Whether to write out all points of all features into the featureXML", ListUtils::create<String>("advanced"));
     defaults_.setValidStrings("write_convex_hull", ListUtils::create<String>("true,false"));
+    defaults_.setValue("spectrum_addition_method", "simple", "For spectrum addition, either use simple concatenation or use peak resampling", ListUtils::create<String>("advanced"));
+    defaults_.setValidStrings("spectrum_addition_method", ListUtils::create<String>("simple,resample"));
```
cbielow: you can use `{"simple", "resample"}` instead of `ListUtils::create`

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -46,13 +49,55 @@
...
+{
+  void sortSpectrumByMZ(OpenSwath::SpectrumPtr spec)
+  {
+    //sort index list
+    std::vector<std::pair<double, Size> > sorted_indices;
+    sorted_indices.reserve(spec->getMZArray()->data.size());
+    auto mz_it = spec->getMZArray()->data.begin();
+    for (Size i = 0; i < spec->getMZArray()->data.size(); ++i)
+    {
+      sorted_indices.push_back(std::make_pair(*mz_it, i));
```
cbielow: `emplace_back(*mz_it, i);` ?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -46,13 +49,55 @@
...
+      ++mz_it;
+    }
+    std::stable_sort(sorted_indices.begin(), sorted_indices.end(), PairComparatorFirstElement<std::pair<double, Size> >());
+
+    // extract list of indices
+    std::vector<Size> select_indices;
+    select_indices.reserve(sorted_indices.size());
+    for (Size i = 0; i < sorted_indices.size(); ++i)
```
cbielow: prefer range based, since you do not need the index?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -46,13 +49,55 @@
 
 // basic file operations
 
+namespace OpenMS
+{
+  void sortSpectrumByMZ(OpenSwath::SpectrumPtr spec)
```
cbielow: IMHO passing a smart pointer is overkill here.
It should be sufficient to pass `OpenSwath::Spectrum& spec`, right?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -87,7 +88,8 @@ namespace OpenMS
     void initialize(double rt_normalization_factor,
                     int add_up_spectra,
                     double spacing_for_spectra_resampling,
-                    const OpenSwath_Scores_Usage & su);
+                    const OpenSwath_Scores_Usage & su,
```
cbielow: `const OpenSwath_Scores_Usage& su`

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
timosachsenberg: Don’t know if relevant but nice to know https://stackoverflow.com/questions/1577475/c-sorting-and-keeping-track-of-indexes

cbielow:
Good point.

We are doing something similarly non-C11 style in https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/KERNEL/MSSpectrum.cpp#L301

Might be worth changing.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
hroest: well, this is how we handle all access to spectra internally and it makes the data handling and the interfaces easier.

cbielow: what is the advantage of this unnecessary indirection? 
I can only see disadvantages:
 - worse performance (passing a shared pointer by copy, which means atomic count increase / lock)
 - user needs to known about the SharedPtr, when everything he really needs is the thing pointed to

This is a bad interface choice and its being discussed in various lectures as "don't do it".

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
```diff
@@ -60,13 +62,26 @@ namespace OpenMS
...
     {
       if (all_spectra[i]->getMZArray()->data.empty() )
       {
         continue;
       }
 
+      // estimate sampling rate
+      for (Size k = 0; k < all_spectra[i]->getMZArray()->data.size() && sampling_rate < 0; k++)
+      {
+        if (k > 0)
```
cbielow: why not init `k = 1` and omit the `if`?

cbielow:
looks good to me, maybe remove the empty line at https://github.com/OpenMS/OpenMS/blob/19cfc9357bce518af9f5207850e22b5ff3222363/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp#L582
to reduce the number of warnings for CodeFactor (the other two warnings are somewhat abstract and would probably require some more refactoring -- not required IMHO).

PR #4180, opened 2019-07-11
Feature/visual fix

hroest:
fixes two bugs encountered in the wild

- chromatogams now also display meta data
- filtering does not crash if meta data arrays are empty
- substantially speeds up loading of data if no index is present in TOPPView

### DIFF
--- a/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
+++ b/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
@@ -1239,13 +1239,15 @@ namespace OpenMS
         bool parsing_success = false;
         if (type == FileTypes::MZML)
         {
-          MzMLFile f;
-          on_disc_peaks->openFile(filename);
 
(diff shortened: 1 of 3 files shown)


src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -1239,13 +1239,15 @@ namespace OpenMS
...
-          on_disc_peaks->openFile(filename);
 
+          // Load index only and check success (is it indexed?)
+          MzMLFile f;
           Internal::IndexedMzMLHandler indexed_mzml_file_;
           indexed_mzml_file_.openFile(filename);
           if ( indexed_mzml_file_.getParsingSuccess() && cache_ms2_on_disc)
           {
+            // If it has an index, now load index and meta data
+            on_disc_peaks->openFile(filename, false);
```
hroest: if no index is present, fail fast and move on (instead of trying to parse through the whole file)

PR #4181, opened 2019-07-12
[Update] MSstatsConverter (TMT)

oliveralka:
Some minor updates for the MSstatsConverter, due to changes in the next R package version (MSstatsTMT). 

Question: 
The class MSstatsFile, due to the addition of the TMT-method has a lot of code duplication, or at least code overlap, with small changes e.g. MSstatsFile::MSstatsLine and MSstatsFile::MSstatsTMTLine; storeLFQ and storeISO.

Here, I tried to do some refactoring, but I am a little bit lost - where do I start and what would be best to make the code a bit cleaner and more functional. 

I would be very happy if you could give me some pointers! 
This will then probably go to a new PR.

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/MSstatsFile.h
+++ b/src/openms/include/OpenMS/FORMAT/MSstatsFile.h
@@ -92,8 +92,8 @@ namespace OpenMS
 
 
         /*
-         *  @brief: Class to aggregate intermediate information from ConsensusFeature and ConsusneMap,
-         *  such as filenames, intensities, retention times, labes and features (for futher processing)
+         *  @brief: Struct to aggregate intermediate information from ConsensusFeature and ConsensusMap,
+         *  such as filenames, intensities, retention times, labes and features (for further processing)
          */
         struct AggregatedConsensusInfo
         {
@@ -301,7 +301,7 @@ namespace OpenMS
       };
 
         /*
-         *  @brief Constructs the lines and add them to the TextFile
+         *  @brief Constructs the lines and adds them to the TextFile
          */
         template <class LineTemplate>
         void constructFile(const String& retention_time_summarization_method,
--- a/src/openms/include/OpenMS/FORMAT/MSstatsFile.h
+++ b/src/openms/include/OpenMS/FORMAT/MSstatsFile.h
@@ -82,7 +82,7 @@ namespace OpenMS
                       const String& condition,
                       const String& mixture,
                       const String& retention_time_summarization_method);
-    
+
     private:
 
         typedef OpenMS::Peak2D::IntensityType Intensity;
@@ -90,21 +90,26 @@ namespace OpenMS
 
         const String na_string = "NA";
 
+
+        /*
+         *  @brief: Class to aggregate intermediate information from ConsensusFeature and ConsusneMap,
+         *  such as filenames, intensities, retention times, labes and features (for futher processing)
+         */
         struct AggregatedConsensusInfo
         {
           std::vector< std::vector< String > > consensus_feature_filenames;           // Filenames of ConsensusFeature
-          std::vector< std::vector< Intensity > > consensus_feature_intensites;       // Intensites of ConsensusFeature
+          std::vector< std::vector< Intensity > > consensus_feature_intensities;       // Intensities of ConsensusFeature
           std::vector< std::vector< Coordinate > > consensus_feature_retention_times; // Retention times of ConsensusFeature
           std::vector< std::vector< unsigned > > consensus_feature_labels;          // Labels of ConsensusFeature
+          std::vector<BaseFeature> features;                                        // Features of ConsensusMap
         };
 
-        MSstatsFile::AggregatedConsensusInfo aggregateInfo(ConsensusMap const &consensus_map,
-                                                           std::vector<BaseFeature> &features,
-                                                           std::vector<String> const &spectra_paths);
+        MSstatsFile::AggregatedConsensusInfo aggregateInfo(const ConsensusMap& consensus_map,
+                                                           const std::vector<String>& spectra_paths);
 
         /*
-        *  @brief: Internal function to check if MSstats_BioReplicate and MSstats_Condition in Experimental Design
-        */
+         *  @brief: Internal function to check if MSstats_BioReplicate and MSstats_Condition in Experimental Design
+         */
         static void checkConditionLFQ_(const ExperimentalDesign::SampleSection& sampleSection, const String& bioreplicate, const String& condition);
 
         /*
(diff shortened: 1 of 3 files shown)


src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -63,6 +63,21 @@ namespace OpenMS
...
+        struct AggregatedConsensusInfo
+        {
+          std::vector< std::vector< String > > consensus_feature_filenames;           // Filenames of ConsensusFeature
+          std::vector< std::vector< Intensity > > consensus_feature_intensites;       // Intensites of ConsensusFeature
+          std::vector< std::vector< Coordinate > > consensus_feature_retention_times; // Retention times of ConsensusFeature
+          std::vector< std::vector< unsigned > > consensus_feature_labels;          // Labels of ConsensusFeature
+        };
+
+        MSstatsFile::AggregatedConsensusInfo aggregateInfo(ConsensusMap const &consensus_map,
+                                                           std::vector<BaseFeature> &features,
```
timosachsenberg: I find it a bit confusing that there are two "results" of this function.
The column vectors (=AggregatedConsensusInfo) and &features

src/openms/source/FORMAT/MSstatsFile.cpp
```diff
@@ -569,13 +539,13 @@ void OpenMS::MSstatsFile::storeISO(const OpenMS::String &filename, const Consens
           for (const OpenMS::PeptideEvidence &pep_ev : peptide_evidences)
           {
             // Write new line for each run
-            for (Size j = 0; j < consensus_feature_filenames[i].size(); j++)
+            for (Size j = 0; j < AggregatedInfo.consensus_feature_filenames[i].size(); j++)
```
timosachsenberg: I find it a bit confusing that you are iterating over the features in the outer loop but also take the information from the AggregatedInfo datastructure.

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
timosachsenberg: could this function and struct be private?

oliveralka: The function could and struct could be private. 

And you are right - there should not be two "results". I was not paying enough attention to that while extracting the method. I actually wanted to store all relevant information in AggregatedConsensusInfo.

src/openms/source/FORMAT/MSstatsFile.cpp
oliveralka: yes - see comment above.

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -82,14 +82,34 @@ namespace OpenMS
...
+        typedef OpenMS::Peak2D::CoordinateType Coordinate;
+
         const String na_string = "NA";
 
+
         /*
-        *  @brief: Internal function to check if MSstats_BioReplicate and MSstats_Condition in Experimental Design
-        */
+         *  @brief: Struct to aggregate intermediate information from ConsensusFeature and ConsensusMap,
+         *  such as filenames, intensities, retention times, labes and features (for further processing)
```
timosachsenberg: typo: labels

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -82,14 +82,34 @@ namespace OpenMS
...
+         */
+        struct AggregatedConsensusInfo
+        {
+          std::vector< std::vector< String > > consensus_feature_filenames;           // Filenames of ConsensusFeature
```
timosachsenberg: use proper doxygen commnts //<

timosachsenberg: add information what the entries correspond to: e.g. consensus_feature_filenames[0] corresponds to the the filenames stored in the first ConsensusMap

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -82,14 +82,34 @@ namespace OpenMS
...
+          std::vector< std::vector< Intensity > > consensus_feature_intensities;       // Intensities of ConsensusFeature
+          std::vector< std::vector< Coordinate > > consensus_feature_retention_times; // Retention times of ConsensusFeature
+          std::vector< std::vector< unsigned > > consensus_feature_labels;          // Labels of ConsensusFeature
+          std::vector<BaseFeature> features;                                        // Features of ConsensusMap
+        };
+
+        MSstatsFile::AggregatedConsensusInfo aggregateInfo(const ConsensusMap& consensus_map,
```
timosachsenberg: docuement or make private

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -82,14 +82,34 @@ namespace OpenMS
...
         const String na_string = "NA";
 
+
         /*
-        *  @brief: Internal function to check if MSstats_BioReplicate and MSstats_Condition in Experimental Design
-        */
+         *  @brief: Struct to aggregate intermediate information from ConsensusFeature and ConsensusMap,
+         *  such as filenames, intensities, retention times, labes and features (for further processing)
+         */
+        struct AggregatedConsensusInfo
```
timosachsenberg: can this struct be private? AggregatedConsensusInfo_

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -82,14 +82,34 @@ namespace OpenMS
...
+          std::vector< std::vector< Coordinate > > consensus_feature_retention_times; // Retention times of ConsensusFeature
+          std::vector< std::vector< unsigned > > consensus_feature_labels;          // Labels of ConsensusFeature
+          std::vector<BaseFeature> features;                                        // Features of ConsensusMap
+        };
+
+        MSstatsFile::AggregatedConsensusInfo aggregateInfo(const ConsensusMap& consensus_map,
+                                                           const std::vector<String>& spectra_paths);
+
+        /*
+         *  @brief: Internal function to check if MSstats_BioReplicate and MSstats_Condition in Experimental Design
```
timosachsenberg: *exists* in

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -144,142 +163,154 @@ namespace OpenMS
...
+        String fragment_ion_;
+        String frag_charge_;
+        String isotope_label_type_;
+        String condition_;
+        String bioreplicate_;
+        String run_;
+        String fraction_;
+      };
+
         class MSstatsTMTLine
```
timosachsenberg: different intendation? see line below

src/openms/include/OpenMS/FORMAT/MSstatsFile.h
```diff
@@ -144,142 +163,154 @@ namespace OpenMS
...
+        String mixture_;
+        String techrepmixture_;
+        String fraction_;
+      };
+
+        /*
+         *  @brief Constructs the lines and adds them to the TextFile
+         */
+        template <class LineTemplate>
+        void constructFile(const String& retention_time_summarization_method,
```
timosachsenberg: private? _

PR #4185, opened 2019-07-16
minor wording fixes

bernt-matthias:
I thing that 'Either' is wrong here if its possible to specify both.


timosachsenberg:
Thanks!

Issue #4168, opened 2019-06-27
Misleading MS2 precursor m/z information


hendrikweisser:
It's a commonly observed problem in DDA experiments that MS2 precursor m/z values can differ from the monoisotopic m/z values of the corresponding peptides (or RNA oligonucleotides) because the precursors can get selected from higher-intensity isotopologue peaks. So it's quite normal to see precursor m/z values that are higher than the monoisotopic m/z values, and there are methods to correct for that (e.g. HighResPrecursorMassCorrector or the "precursor:isotopes" parameter in NucleicAcidSearchEngine).

However, I've now observed cases where the precursor m/z values (as represented in OpenMS) are *lower* than the monoisotopic m/z values, and are thus moved outside the boundaries of the chromatographic peak (feature). See the screenshot from TOPPView below, where the black MS2 indicators are below the feature in the main window; some precursor locations don't even have a data point (peak). The data is from a Thermo Q Exactive HF, centroided and converted to mzML with ProteoWizard.

![TOPPView_precursor_problem](https://user-images.githubusercontent.com/2251068/60273649-7b502880-98ee-11e9-882a-d26677863a8c.png)

Looking in the mzML file, the problem seems to be due to conflicting/misleading information in the precursor section:
```
          <precursorList count="1">
            <precursor spectrumRef="controllerType=0 controllerNumber=1 scan=5762">
              <isolationWindow>
                <cvParam cvRef="MS" accession="MS:1000827" name="isolation window target m/z" value="763.83056640625" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
                <cvParam cvRef="MS" accession="MS:1000828" name="isolation window lower offset" value="0.699999988079" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
                <cvParam cvRef="MS" accession="MS:1000829" name="isolation window upper offset" value="0.699999988079" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
              </isolationWindow>
              <selectedIonList count="1">
                <selectedIon>
                  <cvParam cvRef="MS" accession="MS:1000744" name="selected ion m/z" value="763.45484085234" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
                  <cvParam cvRef="MS" accession="MS:1000041" name="charge state" value="8"/>
[...]

PR #4187, opened 2019-07-18
NucleicAcidSearchEngine improvements

hendrikweisser:
- Allow negative values for parameter "precursor:isotopes" (see issue #4168)
- Take precursor intensities into account when estimating target RTs for label-free quantification (see parameter "lfq_out") - this improves feature detection using FeatureFinderMetaboIdent
- Add log messages with number of identified spectra (before/after FDR filtering)

### DIFF
--- a/src/utils/NucleicAcidSearchEngine.cpp
+++ b/src/utils/NucleicAcidSearchEngine.cpp
@@ -850,7 +850,9 @@ protected:
     {
       IDFilter::filterQueryMatchesByScore(id_data, fdr_ref, fdr_cutoff);
       OPENMS_LOG_INFO << "Search hits after FDR filtering: "
-               << id_data.getMoleculeQueryMatches().size() << endl;
+                      << id_data.getMoleculeQueryMatches().size()
+                      << "\nIdentified spectra after FDR filtering: "
+                      << id_data.getDataQueries().size() << endl;
     }
   }
 
@@ -1297,10 +1299,11 @@ protected:
     }
     progresslogger.endProgress();
 
-    OPENMS_LOG_INFO << "Undigested nucleic acids: " << fasta_db.size() << endl;
-    OPENMS_LOG_INFO << "Oligonucleotides: " << id_data.getIdentifiedOligos().size()
-             << endl;
-    OPENMS_LOG_INFO << "Search hits: " << hit_counter << endl;
+    OPENMS_LOG_INFO << "Undigested nucleic acids: " << fasta_db.size()
+                    << "\nOligonucleotides: "
+                    << id_data.getIdentifiedOligos().size()
+                    << "\nSearch hits (spectrum matches): " << hit_counter
+                    << endl;
 
     if (!exp_ms2_out.empty())
     {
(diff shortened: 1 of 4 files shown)


src/utils/NucleicAcidSearchEngine.cpp
```diff
@@ -846,21 +858,27 @@ class NucleicAcidSearchEngine :
...
   }
 
 
   void generateLFQInput_(IdentificationData& id_data, const String& out_file)
   {
-    // mapping: (oligo, adduct) -> charge -> RTs
-    map<pair<NASequence, String>, map<Int, vector<double>>> rt_info;
+    // mapping: (oligo, adduct) -> charge -> (precursor intensity, RT)
+    map<pair<NASequence, String>,
+        map<Int, vector<pair<double, double>>>> rt_info;
```
timosachsenberg: Hard to read :) typedefs?

cbielow: ehh.. maps within a map? Are you sure?
is not `map<a, map<b, c>>` better represented as map `< (a,b), c>` in terms of efficiency?
depends a bit on the use case I guess (do you want to count charges, for example?)

hendrikweisser: @cbielow: I don't think efficiency is a concern at that point. The code only gets executed once per tool run. In `map<pair<a, b>, c>`, the `a`s are stored multiple times, which may make it more or less efficient.

timosachsenberg:
rebuild jenkins

src/utils/NucleicAcidSearchEngine.cpp
```diff
@@ -879,13 +899,20 @@ class NucleicAcidSearchEngine :
...
+        for (const auto& rt_pair : charge_pair.second)
+        {
+          weighted_rt += rt_pair.first * rt_pair.second;
+          total_weight += rt_pair.first;
+        }
+        rts.push_back(weighted_rt / total_weight);
       }
       tsv << name << ef << 0 << ListUtils::concatenate(charges, ",");
-      // @TODO: do something more sophisticated for the RT?
+      // overall target RT is median over all charge states:
```
timosachsenberg: without seeing the full picture this sounds a bit dangerous

hendrikweisser: My reasoning is that LC happens before ionisation, so in principle all charge states must have the same RT. I'm trying to alleviate problems from patchy MS2 sampling by combining data from multiple charge states.

timosachsenberg: I don’t know much about that but I thought that this Depends on the adduct.

hendrikweisser: Not sure if that's what you mean, but the RT calculations are done separately for every combination of oligonucleotide and adduct. So you can get different RTs for the same oligo with different adducts.

timosachsenberg: so every oligo + adduct pair get 1 target RT. If several charge states of the same oligo + adduct pair exist then also the median of all RTs is returned?

hendrikweisser: Yes. One final RT per oligo/adduct pair, which is calculated from all charge states of that oligo/adduct pair.

timosachsenberg: Ok fine with me. I only see problems with e.g. deuterium as charge carrier but this sounds kind of artificial.

timosachsenberg:
rebuild jenkins

timosachsenberg:
rebuild jenkins

PR #4188, opened 2019-07-18
ImageCreator improvements

hendrikweisser:
- Add parameters to specify RT and m/z ranges to visualise only a part of the input data
- Update the gradients to the newer defaults used in TOPPView

The "normal" default gradient looks better than before, but the "log-scale" one is greyscale. Not sure if that's intentional or was supposed to be only a temporary solution.

One limitation of ImageCreator that has become apparent when visualising (small) data ranges is that all data points are shown as single pixels. When there are fewer spectra or m/z points in the range than the width/height of the image, much of the image may stay white (or whatever is chosen as the background colour). A workaround is to lower the resolution (i.e. width/height) of the image.


hroest:
please fix these tests:

```
The following tests FAILED:
	485 - SimpleSearchEngineAlgorithm_test (Failed)
	927 - UTILSWRITEINI_SimpleSearchEngine (OTHER_FAULT)
	928 - UTILSWRITEINI_SimpleSearchEngine_SectionName (Failed)
	1016 - UTILSWRITECTD_SimpleSearchEngine (OTHER_FAULT)
	2236 - UTILS_ImageCreator_1_out1 (Failed)
	2238 - UTILS_ImageCreator_2_out1 (Failed)
	2283 - UTILS_SimpleSearchEngine_1 (OTHER_FAULT)
	2284 - UTILS_SimpleSearchEngine_1_out (Failed)
```

src/openms/source/KERNEL/MSExperiment.cpp
```diff
@@ -311,7 +307,7 @@ namespace OpenMS
       if (it->getMZ() < RangeManagerType::pos_range_.minY()) RangeManagerType::pos_range_.setMinY(it->getMZ());
       if (it->getMZ() > RangeManagerType::pos_range_.maxY()) RangeManagerType::pos_range_.setMaxY(it->getMZ());
 
-      // do not update RT and in if the spectrum is empty
+      // do not update RT and intensity if the spectrum is empty
```
hroest:
```suggestion
      // do not update RT and intensity if the chromatogram is empty
```

src/utils/ImageCreator.cpp
```diff
@@ -274,34 +277,34 @@ class TOPPImageCreator :
         return ILLEGAL_PARAMETERS;
       }
     }
+    const double init = numeric_limits<double>::max();
+    double rt_min = -init, rt_max = init, mz_min = -init, mz_max = init;
+    bool filter_rt = parseRange_(getStringOption_("rt"), rt_min, rt_max);
+    if (rt_min > rt_max) swap(rt_min, rt_max);
```
hroest: @hendrikweisser: make this part of `parseRange_` to fix min/max if user provides it incorrectly?

hendrikweisser:
@hroest: I fixed the comment and the ImageCreator tests. The other test failures are unrelated to my changes.

Issue #4189, opened 2019-07-18
TOPPView: hiding of automated m/z annotations doesn't always work


hendrikweisser:
In TOPPView's 1D view, the context menu option "Settings > Show/hide automated m/z annotations" doesn't always work to hide the annotations. It seems that:
- If a 1D view is opened with the annotations already switched on, they cannot be hidden.
- If a 1D view is opened without the annotations, they can be switched on and back off.

This is particularly a problem for the "Identification view", where the m/z annotations overlap with fragment ion annotations, creating a very bad visual appearance.

To reproduce:
- start TOPPView
- open the "BSA1.mzML" example file in 2D mode (default)
- double-click on the first spectrum in the "Scans" window to open it in 1D view
- note the m/z annotations on the peaks
- try switching the annotations off through the context menu
- nothing changes
- close the 1D view tab
- double-click on the first spectrum again
- it opens in 1D view without annotations
- now annotations can be enabled and disabled again through the context menu

hendrikweisser:
> This is particularly a problem for the "Identification view", where the m/z annotations overlap with fragment ion annotations, creating a very bad visual appearance.

Forgot to add: For this reason I think the annotations should be disabled by default in "Identification view".

PR #4190, opened 2019-07-18
Disable automated m/z annotations in TOPPView by default

hendrikweisser:
Temporary (?) fix for #4189.


timosachsenberg:
rebuild jenkins

PR #4192, opened 2019-07-22
[FEATURE/FIX] ChromeleonFile handles raw data header

pasqoo:
### Additions
- handle new `Chromatogram Data:` raw data header (tests included)
- handle possible string `n.a.` in the `Step` column
- `line` is now a dynamically allocated string instead of a fixed-size array
- using `regex_match` instead of `regex_search`
- do not include character `\r` in sub-match

Note: the first commit is part of another PR: https://github.com/OpenMS/OpenMS/pull/4191 I branched out of that otherwise I could not compile and test this PR's code locally.


cbielow:
you can omit the checks for `\r` if you use
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1TextFile.html#ab12061f3ba3ea49afa5f0ba6b377a077

PR #4198, opened 2019-07-29
[FIX] OpenSWATH: Patch to prevent NaN from spectral angle score

grosenberger:
In very rare cases (e.g. with ultra-short gradients), the spectral angle score of OpenSWATH results in NaN values. This patch ensures that always numerical values are reported.


src/openswathalgo/source/OPENSWATHALGO/ALGO/MRMScoring.cpp
```diff
@@ -593,6 +593,11 @@ namespace OpenSwath
 
     spectral_angle = Scoring::SpectralAngle(&experimental_intensity[0], &library_intensity[0], boost::numeric_cast<unsigned int>(transitions.size()));
 
+    if (boost::math::isnan(spectral_angle))
```
timosachsenberg: does std::isnan also work?

timosachsenberg:
feel free to merge

PR #4199, opened 2019-07-30
Improve handling of precursor m/z when reading/writing mzML

hendrikweisser:
This addresses issue #4168 using the approach mentioned here: https://github.com/OpenMS/OpenMS/issues/4168#issuecomment-508520131

What remains is checking for the meta value "isolation window target m/z" in the isolation window visualisation code for TOPPView. Can anyone point me to where this should be changed?

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -3852,7 +3852,9 @@ namespace OpenMS
       {
         os << "\t\t\t\t\t\t\t<cvParam cvRef=\"MS\" accession=\"MS:1000044\" name=\"dissociation method\" />\n";
       }
-      //as "precursor" has no own user param its userParam is stored here
+      // as "precursor" has no own user param its userParam is stored here;
+      // don't write out parameters that are used internally to distinguish
+      // between precursor m/z values from different sources:
       writeUserParam_(os, precursor, 7, "/mzML/run/spectrumList/spectrum/precursorList/precursor/activation/cvParam/@accession", validator, {"isolation window target m/z", "selected ion m/z"});
       os << "\t\t\t\t\t\t</activation>\n";
       os << "\t\t\t\t\t</precursor>\n";
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -121,7 +121,7 @@ namespace OpenMS
     void MzMLHandler::getCounts(Size& spectra_counts, Size& chromatogram_counts)
     {
       if (load_detail_ == XMLHandler::LD_RAWCOUNTS)
-      { 
+      {
         spectra_counts = std::max(scan_count_total_, 0); // default is -1; if no specs were found, report 0
         chromatogram_counts = std::max(chrom_count_total_, 0);
       }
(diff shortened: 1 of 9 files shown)


src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -3836,7 +3853,7 @@ namespace OpenMS
         os << "\t\t\t\t\t\t\t<cvParam cvRef=\"MS\" accession=\"MS:1000044\" name=\"dissociation method\" />\n";
       }
       //as "precursor" has no own user param its userParam is stored here
-      writeUserParam_(os, precursor, 7, "/mzML/run/spectrumList/spectrum/precursorList/precursor/activation/cvParam/@accession", validator);
+      writeUserParam_(os, precursor, 7, "/mzML/run/spectrumList/spectrum/precursorList/precursor/activation/cvParam/@accession", validator, {"isolation window target m/z", "selected ion m/z"});
```
timosachsenberg: please add some extra documentation why these are not written out

src/tests/topp/FileMerger_6_output.mzML
```diff
@@ -199,9 +199,6 @@
 				</scanList>
 				<precursorList count="1">
 					<precursor>
-						<isolationWindow>
-							<cvParam cvRef="MS" accession="MS:1000827" name="isolation window target m/z" value="445.34" unitAccession="MS:1000040" unitName="m/z" unitCvRef="MS" />
-						</isolationWindow>
 						<selectedIonList count="1">
```
hroest: @hendrikweisser is this a good idea to not write this out any more?

hendrikweisser: Yes. The element actually isn't present in the input file that gets merged, so the new behaviour is correct.

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
hendrikweisser: Done.

timosachsenberg:
rebuild jenkins

timosachsenberg:
Do we have any experience if the stripped information is used by third-party tools?

hendrikweisser:
@timosachsenberg:
What do you mean by "stripped information"?

timosachsenberg:
If some thid-party tools rely on these meta values being present.

hendrikweisser:
I'm still not sure I understand. There would have to be a meta value with one of the names used internally for it to be misinterpreted and removed. Possible in principle, but seems unlikely, and in any case an external tool (that accepts "standard" mzML) could not rely on such a meta value being present.

timosachsenberg:
ok can you merge against develop. Would like to get this included for testing in the release candidate.

hendrikweisser:
Can do. Which option should I use (normal, squash, rebase)?

Do you know where changes would need to be made in TOPPView to fix the isolation window visualisation?

timosachsenberg:
should be in https://github.com/OpenMS/OpenMS/blob/develop/src/openms_gui/source/VISUAL/TOPPViewIdentificationViewBehavior.cpp#L827
you can just merge and we can squash and merge later

hendrikweisser:
Thanks for the TOPPView pointer, Timo. I've made the small change and it seems to have the desired effect - see screenshot (left: before, right: after).

TOPPView_precursor_fix

Still okay to merge?

timosachsenberg:
I think we can merge and see how/if it negatively affects the visualization of other data.

PR #4200, opened 2019-07-31
Update MapAlignmentAlgorithmIdentification.h

timosachsenberg:
use_feature_rt was uninitialized.
Now the proper reference RTs should be extracted


hroest:
why not add it to the update_members_ function?

PR #4201, opened 2019-07-31
Simplification of the XL-MS PeptideIdentification / PeptideHit ID structures

enetz:
Previously OpenPepXL output had two PeptideHits under one PeptideIdentification for each Cross-Link ID.
This was mainly because we needed separate `PeptideHits` for the two peptides to run `PeptideIndexer` and it made sense for `MzIdentML` output as well.
But this caused a lot of redundancy and incompatibilities with other tools.
Now `OpenPepXL` still generates the old ID structure, then runs the `PeptideIndexer`, then processes the necessary info from the two `PeptideHits` and stores all info as MetaValues in one hit and throws away the second hit.
These `PeptideHits` are then summarized under one `PeptideIdentification` for each MS2 spectrum. This makes the data structure more in sync with all other tools in OpenMS, e.g. the `PercolatorAdapter` now works for `OpenPepXL` output as well.
Writing and reading of all related output formats has been adjusted accordingly.
I also asked for new CV terms for `OpenPepXL` in the `psi-ms.obo`. I updated the `psi-ms.obo` to use the new CV terms in `MzIdentML` and had to change a few test files because of a change in a CV term name.

- Updated psi-ms.obo
- Updated `MzMLHandler` and test files: The CV term: `accession="MS:1000810" name="mass chromatogram"` was changed to: `accession="MS:1000810" name="ion current chromatogram"`
- Changed the internal ID data structures of `OpenPepXL/LF` as well as idXML output
- Adapted input/output of XQuest.xml and mzIdentML files to new internal data structures
- Adapted `XFDR` and `TOPPView` to use the new datastructures
- Added an `extra_features` MetaValue to make `OpenPepXL/LF` or `XFDR` output compatible with `PercolatorAdapter`


src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -883,79 +887,83 @@ namespace OpenMS
...
+
         ph_beta.setSequence(*top_csms_spectrum[i].cross_link.beta);
         ph_beta.setCharge(precursor_charge);
         ph_beta.setScore(top_csms_spectrum[i].score);
-        ph_beta.setRank(DataValue(i+1));
-        ph_alpha.setMetaValue("beta_sequence", (*top_csms_spectrum[i].cross_link.beta).toString());
-        ph_beta.setMetaValue("beta_sequence", (*top_csms_spectrum[i].cross_link.beta).toString());
+        // ph_beta.setRank(DataValue(i+1));
+
+        // TODO when beta PeptideHits are not used anymore, remove most of these, that are not used by PeptideInderer
```
timosachsenberg: there are a lot of commented out stuff
can you remove these (maybe mark them with a commit message so you find them again if needed)

src/openms/source/CONCEPT/Constants.cpp
```diff
@@ -155,9 +155,90 @@ namespace OpenMS
     // User parameter name for precursor mz error in ppm
     const std::string PRECURSOR_ERROR_PPM_USERPARAM = "OMS:precursor_mz_error_ppm";
 
+    // User parameter name for monoisotopic peak misassignment correction of a precursor (usually an integer x with the correction being x times C13C12_MASSDIFF_U)
```
timosachsenberg: please state in which direction the values should be used (e.g, 1 = subtract (1/charge)*C13C12_MASSDIFF_U from uncorrected precursor_mz

src/openms/include/OpenMS/CONCEPT/Constants.h
```diff
@@ -258,13 +258,148 @@ namespace OpenMS
     */
     extern OPENMS_DLLAPI const std::string   PRECURSOR_ERROR_PPM_USERPARAM;
 
+
+    /** User parameter name for monoisotopic peak misassignment correction of a precursor (usually an integer x with the correction being x times C13C12_MASSDIFF_U)
+            String
+    */
+    extern OPENMS_DLLAPI const std::string   PRECURSOR_ASSIGNMENT_CORRECTION;
```
timosachsenberg: also add _USERPARAM suffix for all of them?
Another idea would be to put all USER_PARAM constants into their own namespace
Constants::UserParams::FRAGMENT_ANNOTATION

enetz: Should I make an additional class for these altogether? They do not really fit with the other constants.

enetz: something like Names.h or UserParamNames.h

timosachsenberg: hmm I think it is fine to just put the extra namespace

namespace OpenMS
{
  namespace Constants
  {
   namespace UserParams
    {

in the same file.
@cbielow or @jpfeuffer what is your opinion on that?

enetz:
Stuff still missing:
- I have not looked at python bindings yet. I did not change many function declarations, but will add python bindings for a few new functions.

hroest:
just a quick question: why is there a need to run the peptide indexer each time? why not incorporate that directly into the tool and make the workflow much easier?

enetz:
@hroest that is kind of what I am doing now. After the OpenPeXL algorithm is finished, the PeptideIndexing class is used at the end automatically before any data is written out. I just added another data transformation step after that, that changes the ID data structures and idXML output.
This newer ID data structure is now the "default" and neither the user nor other OpenMS tools should have to deal with the old one. It is only temporarily present in memory to run the PeptideIndexing algorithm (that is why I adapted the inout/output and XFDR to only use the new structure).
The only downside is that now the type of data structure that the PeptideIndexer tool would need to run outside of OpenPepXL is not written out anymore, but that should not be necessary anyway.

timosachsenberg:
As a proper tool I think it should moved into an algorithm class and provide python binding.

enetz:
OpenPepXL and OpenPepXLLF already have OpenPepXL/LFAlgorithm classes with `run(...)` functions and python bindings.
I am now working on the same for XFDR. Should be finished today.

enetz:
This PR is finished now, content wise.
If the tests pass and there are no new issues, this can be merged.

src/openms/include/OpenMS/ANALYSIS/XLMS/XFDRAlgorithm.h
```diff
@@ -0,0 +1,246 @@
...
+
+    // unique top hits
+    std::vector<String> unique_ids_;
+    std::vector<double> unique_id_scores_;
+
+    // maps index of peptide id all_pep_ids_ to vector of cross link class
+    std::map<String, std::vector<String>> cross_link_classes_;
+
+    // Whether input comes from xQuest
+    bool is_xquest_input_;
```
timosachsenberg: is this needed?  codacy complains about it not being initialized

enetz: I just noticed that this variable is not used much anymore. I was able to get rid of it.

timosachsenberg:
Feel free to merge if test passes

PR #4204, opened 2019-08-06
[FIX] avoid mixing hashes

timosachsenberg:
fixes the problem in MRMIonSeries where unordered_map["test"] was mapped to a different bucket than unordered_map[String("test")]


src/openms/source/ANALYSIS/OPENSWATH/MRMIonSeries.cpp
```diff
@@ -528,6 +529,7 @@ namespace OpenMS
                 "\" with precursor charge +" + String(precursor_charge) + " could not be generated.");
           }
 
+          std::cout << *ft_it + String(i) + "^" + String(charge) << ":" << Math::roundDecimal(pos, round_decPow) << std::endl;
```
timosachsenberg: Need to remove that

jpfeuffer:
nice, let's check nightlies with debug mode

src/openms/source/DATASTRUCTURES/String.cpp
```diff
@@ -526,13 +528,11 @@ namespace OpenMS
     this->append(s);
     return *this;
   }
+  
+  ::size_t hash_value(String const& s)
+  {
+    boost::hash<std::string> hasher;
```
timosachsenberg: Is this costly?

jpfeuffer: dont know. but this is recommended from boost. https://www.boost.org/doc/libs/1_62_0/doc/html/hash/custom.html

jpfeuffer: it has nothing but the operator()

jpfeuffer: which returns hash_value. It is probably much safer to use it this way instead of trying to save one layer of abstraction.

timosachsenberg:
Yep construction etc. is optimized away already at -O1

PR #4205, opened 2019-08-06
[FIX] simplified OPXLHelper::computeDeltaScores

enetz:
This should fix a complaint by the  AddressSanitizer in the nightly build

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
+++ b/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
@@ -218,7 +218,7 @@ namespace OpenMS
       static void addPercolatorFeatureList(ProteinIdentification& prot_id);
 
       /**
-       * @brief adds the delta score as a MetaValue
+       * @brief adds the delta score as a MetaValue, assumes the hits under each PeptideIdentification are ordered by score rank
        * @param peptide_ids The vector of peptide_ids containing XL-MS search results without beta PeptideHits
        */
       static void computeDeltaScores(std::vector< PeptideIdentification >& peptide_ids);
--- a/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
+++ b/src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
@@ -218,7 +218,7 @@ namespace OpenMS
       static void addPercolatorFeatureList(ProteinIdentification& prot_id);
 
       /**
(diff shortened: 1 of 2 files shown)


src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLHelper.h
```diff
@@ -218,7 +218,7 @@ namespace OpenMS
       static void addPercolatorFeatureList(ProteinIdentification& prot_id);
 
       /**
-       * @brief adds the delta score as a MetaValue
+       * @brief adds the delta score as a MetaValue, assumes the hits under each PeptideIdentification are ordered by score rank
```
timosachsenberg: can you add an OPENMS_PRECONDITION that checks if it is sorted?

enetz: is there a quick way to check if PeptideHits are sorted by score?

enetz: ah Ok, PeptideIdentification has a function to sort Hits by score. I guess I can just run that before calculating the delta score to make sure as well.

jpfeuffer:
This is good! But I assume there is also something wrong during reading or writing mzIdentML with XL mods (see Windows machines).

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1188,27 +1188,14 @@ namespace OpenMS
...
-        PeptideHit& first = pep_id.getHits()[rank_indices[rank + 1]];
-        PeptideHit& second = pep_id.getHits()[rank_indices[rank]];
-
-        double delta_score = first.getScore()
-                             / second.getScore();
-
-        second.setMetaValue(Constants::UserParam::DELTA_SCORE, delta_score);
-      }
-      pep_id.getHits()[rank_indices.back()].setMetaValue(Constants::UserParam::DELTA_SCORE, 0.0);
+      phs[phs.size()-1].setMetaValue(Constants::UserParam::DELTA_SCORE, 0.0);
```
hroest: what if phs is empty?

enetz:
@jpfeuffer the MzIdentML test seemed to fail because of the same function. It is also called when reading in mzid if it does not have a delta score.

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
enetz: you are right, I will add a check

PR #4206, opened 2019-08-06
 [FIX] OPXLHelper::computeDeltaScores, added checks for the number of PeptideHits

enetz:


timosachsenberg:
failing test was a timeout

PR #4207, opened 2019-08-07
[FEATURE] Added Reference column to MSstats output

jpfeuffer:
Especially important without internal aggregation across different RTs since MSstatsTMT is doing
that by itself (hopefully now based on that reference column which is the filename+nativeID of the spectrum used for that quantity)


src/openms/source/FORMAT/MSstatsFile.cpp
```diff
@@ -497,11 +508,14 @@ void OpenMS::MSstatsFile::storeISO(const String& filename,
...
 
   // The Retention Time is additionally written to the output as soon as the user wants to resolve multiple peptides manually
-  const bool rt_summarization_manual(retention_time_summarization_method == "manual");
+  bool rt_summarization_manual(retention_time_summarization_method == "manual");
 
-  if (rt_summarization_manual)
+  if (!rt_summarization_manual)
   {
-    cout << "WARNING: One feature might appear at multiple retention times in the output file. This is invalid input for MSstats. Combining of features over retention times is needed!" << endl;
+    OPENMS_LOG_WARN << "WARNING: rt_summarization set to something else thatn 'manual' but MSstatsTMT does aggregation of"
```
timosachsenberg: thatn

timosachsenberg:
looks good to me. the only thing I am concerned are special characters in the filnames. These can be paths etc.

src/openms/source/FORMAT/MSstatsFile.cpp
```diff
@@ -497,11 +508,14 @@ void OpenMS::MSstatsFile::storeISO(const String& filename,
...
+            " intensities of peptide-chargestate combinations in the same file itself."
+            " Reverting to 'manual'" << endl;
+    rt_summarization_manual = true;
```
oliveralka: should we add the possibility to force a different aggregation method, when using the MSstatsConverter with TMT data? E.g. if someone wants to use it for aggregation and work with his own code. Or should that not be intended?

timosachsenberg: hmm don't know if that is a relevant use case. It implies that you don't use MSstats and I guess you would just work with the raw data and could do the max / min etc. calculations in the scripting language yourself.

oliveralka: Yes, seems fair! Let's leave it at that.

jpfeuffer:
> looks good to me. the only thing I am concerned are special characters in the filnames. These can be paths etc. Maybe contain , ?

Ok sure. Should I quote or just replace the delimiter?

src/openms/source/FORMAT/MSstatsFile.cpp
jpfeuffer: will fix

timosachsenberg:
whatever works ;)

PR #4208, opened 2019-08-08
fix random segfaults if identified spectra were browsed

timosachsenberg:


jpfeuffer:
I usually prefer the empty() method (also because of clang-tidy)

Issue #3785, opened 2018-10-08
Display TOPPView & TOPPAS macOS 10.14 (Mojave) dark mode


oliveralka:
macOS 10.14 has a light and dark mode feature. 

TOPPView and TOPPAS are not able to handle the switch in terms of font colour. 

edit: The tool nodes are shown correctly (TOPPAS - dark mode)

![toppview_light](https://user-images.githubusercontent.com/15729190/46612051-c8e1f680-cb0f-11e8-841e-74eb60f28e9d.png)
![toppview_dark](https://user-images.githubusercontent.com/15729190/46612052-c8e1f680-cb0f-11e8-8205-9a6c22b538ea.png)
![toppas_light](https://user-images.githubusercontent.com/15729190/46612060-cc757d80-cb0f-11e8-8c80-12d27bb7a9b0.png)
![toppas_dark](https://user-images.githubusercontent.com/15729190/46612061-cc757d80-cb0f-11e8-93f8-b9bd8d0eaad0.png)

timosachsenberg:
see https://bugreports.qt.io/browse/QTBUG-68850 for details

jpfeuffer:
@oliveralka Can you recheck with the newest QT in brew? I could update the build machines if that helps.

jpfeuffer:
According to their bug tracker this is fixed in 5.12.

oliveralka:
It works (develop, QT 5.12.2), but the labels are unfortunately dark on dark - any idea how to change that depending on the systems dark mode?

oliveralka:
As an alternative, we could opt-out of dark mode by adding:
```
<key>NSRequiresAquaSystemAppearance</key>
<string>true</string>
```
to OpenMS/src/openms_gui/source/VISUAL/APPLICATIONS/GUITOOLS/TOPPView-resources/TOPPView.plist.in

see https://developer.apple.com/documentation/appkit/nsappearancecustomization/choosing_a_specific_appearance_for_your_macos_app

jpfeuffer:
Ah cool. Hmm well.. I wouldn't need it. I guess for now an opt-out is better. I think we should change a lot of stuff to make it look nice (maybe plot bg, icons, axis)

timosachsenberg:
addressed un https://github.com/OpenMS/OpenMS/pull/4209

PR #4209, opened 2019-08-08
[FIX] Temporarily opt-out from dark mode feature (MacOS)

oliveralka:
Apps TOPPView and TOPPAS will be started in "aqua/light" mode instead of dark mode independent of system settings (MacOS).

I think if we want to support dark more work has to go into optimization - see #3785


timosachsenberg:
let's merge and get someone test the nightlies :P

PR #4210, opened 2019-08-09
[FIX] set primary MS run support in more tools

timosachsenberg:


timosachsenberg:
The current behavior is now:
- If an MSExperiment is passed with properly annotated mzML (that exists on the file system). Then it is carried over as primary MS run. This basically means that an intermediate spectrum filtering step like centroiding can hand over the original filename to the subsequent tool.
- if the file does not exist, then the annotation is probably invalid or refers to a file on a different file system. In that case the fallback to the mzML filename as passed to a tool is used.

timosachsenberg:
and these are the fixes applied while loading a mzML file (from an old PR - but maybe good for reference or rereview here): 
```
else if (tag == "sourceFile")
      {
        current_id_ = attributeAsString_(attributes, s_id);
        // Name of the source file, without reference to location (either URI or local path). e.g. "control.mzML"
        String name_of_file = attributeAsString_(attributes, s_name);

        //URI-formatted location where the file was retrieved.
        String path_to_file = attributeAsString_(attributes, s_location);

        // mzML files often deviate from the specification by storing e.g. the full path in the name attribute etc.
        // error: whole path is stored in file name. fix: split into path and file name
        if (path_to_file.empty() && !name_of_file.empty())
        {
          path_to_file = File::path(name_of_file);
          name_of_file = File::basename(name_of_file);
          if (path_to_file == ".") path_to_file = "file://./";
        }

        // format URI prefix as in mzML spec.
        if (path_to_file.hasPrefix("File://")) path_to_file.substitute("File://", "file://");
        if (path_to_file.hasPrefix("FILE://")) path_to_file.substitute("FILE://", "file://");
        if (path_to_file.hasPrefix("file:///.")) path_to_file.substitute("file:///.", "file://./");

        bool is_relative_path = path_to_file.hasPrefix("file://./") || path_to_file.hasPrefix("file://../");

        // ill formed absolute or relative path
        if (!is_relative_path && path_to_file.hasPrefix("file://") && !path_to_file.hasPrefix("file:///"))
        {
          warning(LOAD, "Ill formed absolute or relative sourceFile path: " + path_to_file);
        }

        // if possible convert relative path to absolute path
        if (is_relative_path && File::isDirectory(path_to_file))
        {
          String normal_path = String(path_to_file).substitute("file://", ""); // remove URI prefix
          path_to_file = String("file://") + File::absolutePath(normal_path); // on linux this e.g. file:///home... on win: file://C:/...
        }

        // absolute path to the root: remove additional / otherwise we will get file://// on concatenation
        if (!is_relative_path && path_to_file == "file:///")
        {
          path_to_file = "file://";
        }

        source_files_[current_id_].setNameOfFile(name_of_file);
        source_files_[current_id_].setPathToFile(path_to_file);
      }
```

timosachsenberg:
@jpfeuffer merge?

src/openms/source/KERNEL/ConsensusMap.cpp
```diff
@@ -559,11 +561,31 @@ namespace OpenMS
...
       column_description_[i].filename = p;
       ++i;
     }
   }
 
+  void ConsensusMap::setPrimaryMSRunPath(const StringList& s, MSExperiment & e)
+  {
+    StringList ms_path;
+    e.getPrimaryMSRunPath(ms_path);
+    if (ms_path.size() == 1 && ms_path[0].hasSuffix("mzML") && File::exists(ms_path[0]))
```
hroest: @timosachsenberg what if you run this on a different system, e.g. on a cluster where the original file path is invalid? I am not sure that `File::exists(ms_path[0])` is a always a good idea...

timosachsenberg: The reasoning is that we carry over a file that exists on the file system. If not stored in the mzML, then from the input to the current tool execution.
We could in principle remove the check if the file exists but then we face the following problem:
If a user converted a raw file or mzML file at a different place/machine (which I think might be quite common) this filename is carried over to the end of the workflow and it is unlikely that we can open it e.g., in TOPPView.

timosachsenberg: E.g., in many workflow systems, the input files are accessible on all nodes but the intermediate files might be on the nodes temporary folder. This scenario works with current implementation (it will set the proper path to the mzML in the input folder in the first tool) but fails if the check is ommited (and the mzML contains an inaccessible location e.g. from the instrument machine).

hroest: The question is whether we care about keeping track of the file or whether we want to be able to open the file itself later on (seems like a much harder task). I agree that under most conditions it should still work, I just feel that checking whether the file exists on that system is not robust and will thus produce different results when run on test/development systems etc. Seems like a way to introduce hard to track / reproduce bugs

>  If a user converted a raw file or mzML file at a different place/machine (which I think might be quite common) this filename is carried over to the end of the workflow and it is unlikely that we can open it e.g., in TOPPView.

I assumed that is the point, to track where the file came from? So that I can track which input file produced which output file

> E.g., in many workflow systems, the input files are accessible on all nodes but the intermediate files might be on the nodes temporary folder. This scenario works with current implementation (it will set the proper path to the mzML in the input folder in the first tool) but fails if the check is ommited (and the mzML contains an inaccessible location e.g. from the instrument machine).

this will not work with docker where the file paths are not global for the system

PR #4213, opened 2019-08-14
Fixes bugs encountered in a full TMT pipeline:

jpfeuffer:
- When consensusXML is merged rowwise instead of colwise, files in the
ID structures might not exist in the quant structures anymore. Throw
better error.
- Allow first_run_inference_only also for consensusID and make it an
advanced flag for the manual export in MzTabExporter
- Make mzTab the default file suffix in ProteinQuantifier mzTab output.
Nonetheless allow tsv when storing mzTab files.


timosachsenberg:
+1 if tests pass

PR #4214, opened 2019-08-14
Optional default value for "MetaInfo[Interface]::get[Meta]Value"

hendrikweisser:
Adds an argument for a default return value to `MetaInfo::getValue` and `MetaInfoInterface::getMetaValue`, for cases when a meta value doesn't exist. This solves the inefficiency/annoyance of having to do a separate check for existence before reading a meta value:
```
int x = 0;
if (bla.metaValueExists("foo")) x = bla.getMetaValue("foo");
```
This now becomes:
`int x = bla.getMetaValue("foo", 0);`

It's fully backwards compatible because the return value defaults to the previous `DataValue::EMPTY`.

@hroest: Python bindings still need updating. How do I deal with default values for function arguments?


timosachsenberg:
just for discussion as I think they don't go well with python:
 
would something like
if (boost::optional<std::string> name = m.getMetaValue("key")) {...}
be more natural than having different defaults?
or something more C++ like that is the potentially more efficient:
iterator_type it = m.getMetaValue("key")?

I think that constructs like:
if (auto it = m.find("key"); it != m.end()) {...}
would be C17 style and probably what I would prefer (would require a change to iterators).

timosachsenberg:
Currently, we have e.g.:
```
      return this->getPeakData()->size() > 0 &&
             this->getPeakData()->metaValueExists("is_ion_mobility") &&
             this->getPeakData()->getMetaValue("is_ion_mobility").toBool();
```
would we prefer to write constructs like these as:
```
      return this->getPeakData()->size() > 0 &&
                this->getPeakData()->getMetaValue("is_ion_mobility", false).toBool();
```

hendrikweisser:
There's got to be some way to handle this in pyOpenMS. In the worst case there have to be two different functions with one/two arguments.

I think having a default value is exactly what you want in some cases. In other cases you want the "optional" behaviour (where your code paths are different depending on whether you have a value or not).
The `get` method on Python dictionaries has an optional default value; I see this as an equivalent case.

timosachsenberg:
yes I think having two functions in the pxd file - one with / one without should work

hroest:
> yes I think having two functions in the pxd file - one with / one without should work

yes, that is the way to do it. Then two different functions will be called, depending on the number of arguments you provide.

hendrikweisser:
Hm. I tried making the changes, but it looks like pyOpenMS doesn't build. Can't test locally.

There's a comment in "MetaInfoInterface.pxd" saying: `Cython has a problem with inheritance of overloaded methods, so we can only declare one of the two methods here (most people will want to use the String-based methods).`
Is that also a problem if I have two methods with different signatures (number of parameters)?

src/pyOpenMS/pxds/MetaInfo.pxd
```diff
@@ -17,6 +17,11 @@ cdef extern from "<OpenMS/METADATA/MetaInfo.h>" namespace "OpenMS":
         # returns the value corresponding to an index
         DataValue getValue(UInt index) nogil except +
 
+        # returns the value corresponding to a string, with default
+        DataValue getValue(String name, DataValue default_value) nogil except +
+        # returns the value corresponding to an index, with default
+        DataValue getValue(UInt index, DataValue default_value) nogil except +
```
hroest: this will probably not work since Cython cannot handle overloaded functions that are inherited. That is the reason we did not implement them

hroest:
>  Is that also a problem if I have two methods with different signatures (number of parameters)?

yes, as the comment said you cannot have overloaded inherited methods. Either you need a different name for the method - either you have to rename the method or you have to manually copy-paste it to every class that inherits

hendrikweisser:
> Either you need a different name for the method

How easy is it to do that just for Python? I.e. I want to keep the name in C++, but define an alias for the two-arguments variant for the Python binding.

hroest:
> How easy is it to do that just for Python? I.e. I want to keep the name in C++, but define an alias for the two-arguments variant for the Python binding.

you could try the "wrap-as" approach, e.g. 

```
libcpp_map[libcpp_string, int] toMap() nogil except + #wrap-as:getElementalComposition 
```

however, that will just deal with autowrap and re-name it, I am not sure whether it will solve the Cython issue. The other approach is to use the `src/pyOpenMS/addons/` folder to do the C++ function call yourself.

hendrikweisser:
> you could try the "wrap-as" approach, e.g.
> `libcpp_map[libcpp_string, int] toMap() nogil except + #wrap-as:getElementalComposition`

It looks to me like there's some confusion about where the C++ name and where the Python name goes in some of the "wrap-as" usages. In your example above, "toMap" is defined in C++ and gets renamed to "getElementalComposition" in Python. However, there's also this:

`EmpiricalFormula iadd(EmpiricalFormula)   nogil except + # wrap-as:operator+=`

Where "operator+=" is defined in C++ and you want to rename it to "iadd" in Python.
Similarly in "AASequence.pxd" with "iadd" and "fromString".

Or do I misunderstand something?

hendrikweisser:
I've removed the problematic Python binding. The Jenkins failure seems to be due to a problem with the build itself - see [log](https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/3384/compiler=g++-7,os_label=ubuntu_1804/console).

What's the magic command for restarting the Jenkins build?

cbielow:
rebuild jenkins

hendrikweisser:
Thanks, Chris.

It looks to me like the failing Jenkins builds are broken. The nightly build hasn't run for two days. The Mac build log contains:
```
Starting class and or tool tests...
No tests were found!!!
```
The Ubuntu log has:
```
ERROR: Step ?Publish xUnit test result report? failed: [CTest-Version 3.x (default)] - No test report file(s) were found with the pattern 'build/**/Testing/**/*.xml' relative to '/home/jenkins/ws/openms_pr_ci/a9154670' for the testing framework 'CTest-Version 3.x (default)'.
```

Issue #4215, opened 2019-08-15
OpenMS assumes strict hierarchy of MS levels in mzML?


jpfeuffer:
It seems to me that IsobaricAnalyzer ignores precursor annotations in mzMLs and
only works with the last spectrum up one level. At least in my data I don't think this is true. There is exactly one MS3 scan for every MS2 scan but they are not in a hierarchical/descending order.

jpfeuffer:
![Bildschirmfoto 2019-08-15 um 14 34 26](https://user-images.githubusercontent.com/8102638/63094861-d525b400-bf69-11e9-8601-f7083e7973d1.png)

This actually does not only affect IsobaricAnalyzer but the whole library, since everything assumes that this hierarchy exists.

jpfeuffer:
Data can be found here:
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Tutorials/Data/latest/Example_Data/IsobaricLabellingTMT/dataset/

jpfeuffer:
Is is Orbitrap Fusion data and we might see results of MultiNotch MS3.
https://www.dcbiosciences.com/news/raw-proteomics-data-overview/

jpfeuffer:
So I think this should be solved for the getPrecursor function and the IsobaricAnalyzer but might need fixing for the display in TOPPView.

timosachsenberg:
ok so I close this blocker

jpfeuffer:
I reopen as this is still an issue (last checked with 3.0). I fixed the getPrecursor some time ago but the display of the hierarchy is still wrong. Toppview therefore probably makes other assumptions.

jpfeuffer:
I think this is pretty serious since it makes it really hard to check for errors in SPS-MS3-TMT data.

PR #4217, opened 2019-08-15
[Fix] Precursor lookup and cxml default labelled experimental design

jpfeuffer:
partly addresses #4215


timosachsenberg:
looks good. thanks!

PR #4218, opened 2019-08-16
[Tests] interleaved ms3

jpfeuffer:
Adds to #4217


src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -3714,7 +3714,17 @@ namespace OpenMS
 
     void MzMLHandler::writePrecursor_(std::ostream& os, const Precursor& precursor, const Internal::MzMLValidator& validator)
     {
-      os << "\t\t\t\t\t<precursor>\n";
+      // optional attributes
+      String external_spectrum_id =
+          precursor.metaValueExists("external_spectrum_id") ?
+          " externalSpectrumID=\"" + precursor.getMetaValue("external_spectrum_id").toString() + "\"" :
```
timosachsenberg: never read about externalSpectrumID is this used?

jpfeuffer: yes, but don't ask me how. there is even another one (which is unused, so I did not include).

timosachsenberg:
looks good

PR #4221, opened 2019-08-21
[FIX] More fixes to correctly export more complex labelled experimental designs to mzTab

jpfeuffer:


jpfeuffer:
could be merged.

PR #4223, opened 2019-08-22
[Fix] reading of optional parameter groups in XTandem output files.

jpfeuffer:


jpfeuffer:
https://www.thegpm.org/docs/X_series_output_form.pdf

That is the best XTandem schema description that I could find.

jpfeuffer:
I dont know why CodeFactor still complains. I think this is as good as needed now.

PR #4224, opened 2019-08-22
[FIX] deal with boolean values

hroest:
- improve conversion from text to boolean values
- ensure that Skyline output can be parsed directly

add tests for the changes in the Skyline library output format


jpfeuffer:
Windows unrelated. It ran on a new (not yet stable) machine.

PR #4225, opened 2019-08-22
[FEATURE] handle protein N/C terminal modifications

hroest:
- properly parse protein C and N terminal modifications on amino acids
- allow protein C and N terminal modifications on AA sequences

select proper protein modification if available (default case will pick a peptide modification but if not is available, it will search for a potential protein modification that may fit)


src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -233,6 +234,49 @@ START_SECTION(AASequence fromString(const String& s, bool permissive = true))
...
+    TEST_EQUAL(seq2.hasCTerminalModification(), true);
+    TEST_EQUAL(seq2.getCTerminalModification() == nullptr, false);
+    TEST_EQUAL(seq2.isModified(), true);
+    TEST_STRING_EQUAL(seq2.getCTerminalModification()->getFullId(), "Met->Hse (C-term M)");
+    // no terminal dot
+    seq2 = AASequence::fromString("PEPTIDEM(UniMod:10)");
+    TEST_STRING_EQUAL(seq2.getCTerminalModification()->getFullId(), "Met->Hse (C-term M)");
+
+    // test protein N-term modification
+    AASequence seq3 = AASequence::fromString("(UniMod:51)CPEPTIDE"); // <umod:mod title="Tripalmitate" full_name="N-acyl diglyceride cysteine"
```
hroest: this failed before and works now

hroest:
of course there is a question of whether we should allow any protein N/C terminal modification on a AminoAcid, however if not other information is given, it seems that at least we should be permissive and allow for it.

hroest:
merge?

PR #4231, opened 2019-08-30
Try to make the calls to Sirius more reliable and extract their output

jpfeuffer:
Debugging is otherwise almost impossible.
Only tested on Mac.
Main idea is to skip this horrible script they have (only way to reduce weird java errors and swallowed output).
Also check for empty output since Sirius gives you a exit code of 0 when you use an older version and it crashes with a fatal error because they weirdly don't support calls to their CSI server from this version anymore.
Also allow non-quiet Sirius, etc.


jpfeuffer:
Actually we do not need an executable since we just call the main java class. So I use the executable as a reference to the lib folder (which we need for the classpath).
I still need to include the reading of the SIRIUS_PATH which we introduced.
I don't know why a workspace is needed in the checkSiriusExecutablePath function.

oliveralka:
@jpfeuffer look good so far. 
Added:
- java_memory
- library path (THIRDPARTY)
- getenv (Gurobi, Cplex)

It worked on MacOS, but I am not sure about Windows. Let's see.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -275,12 +282,50 @@ namespace OpenMS
...
+        sirius_libpath = sirius_libpath + ":" + gurobi_home + "/lib/gurobi.jar";
+      }
+      if (getenv("CPLEX_HOME") != nullptr)
+      {
+        String cplex_home = getenv("CPLEX_HOME");
+        sirius_libpath = sirius_libpath + ":" +  cplex_home + "/lib/cplex.jar";
+      }
+
+      // library path depends if original or THIRDPARTY version is used.
+      // assemble SIRIUS parameterj
```
hroest:
```suggestion
      // assemble SIRIUS parameters
```

oliveralka:
Should work with:
THIRDPARTY OpenMS -> e.g. THIRDPARTY/All/Sirius/lib/sirius_openms_plugin-4.0.1.jar 
Stand-alone -> e.g. sirius-osx64-4.0.1/lib/sirius_cli-4.0.1.jar

By adding the parameters:
`-sirius:quiet false -debug 2`
Additional SIRIUS output is provided:

```
INFO  09:57:41 - Sirius Workspace Successfull initialized at: /Users/alka/.sirius
INFO  09:57:41 - You run SIRIUS 4.0.1 (build 9)
INFO  09:57:41 - Sirius was compiled with the following ILP solvers: GLPK-v1.7.0 (included), Gurobi-v7.5.2, CPLEX-v12.8
INFO  09:57:41 - Treebuilder priorities are: [GUROBI, CPLEX, GLPK]
INFO  09:57:41 - CPU check done. 2 cores that handle 4 threads were found.
INFO  09:57:41 - Bug reporter initialized!
WARNING09:57:41 - Could not load GrbSolver! 
```

oliveralka:
rebuild jenkins

oliveralka:
@jpfeuffer win_test unrelated?

hroest:
> @jpfeuffer win_test unrelated?

yes, unrelated. See #4297

src/tests/topp/THIRDPARTY/third_party_tests.cmake
```diff
@@ -79,7 +79,7 @@ OPENMS_FINDBINARY(FIDOCHOOSEPARAMS_BINARY "FidoChooseParameters" "FidoChoosePara
 
 #------------------------------------------------------------------------------
 # Sirius
-OPENMS_FINDBINARY(SIRIUS_BINARY "sirius;sirius-console-64.exe" "Sirius")
```
jpfeuffer: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/3440/compiler=msvc-15,os_label=win10_64/artifact/build/General_Testing/Testing/Temporary/LastConfigure_20191011-1349.log/*view*/

I think this makes it not find sirius on windows. Where do we have a sirius.jar? The executable is still needed but only to find the right directory.

jpfeuffer: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=appleclang-9.0.0,os_label=highsierra/ws/SEARCHENGINES/Sirius/

This is how the structure looks like

jpfeuffer: And we only add the Sirius directory to the PATH environment. I can change it to Sirius/lib or we can restructure the  Sirius folder. Or we revert that line.

oliveralka: See below, was changed to the .jar - so it should not look for the "binary" at all.

oliveralka: The sirius.jar - should be for all at 
THIRDPARTY/All/Sirius/lib 
and the glpk - is at the THIRDPARTY/Windows/64bit/Sirius/lib 

see "lib_class_path" - SiriusAdapterAlgorithm

oliveralka: Ok, now I get it. It will not be tested since the "binary" is not found. 

The problem here is that it depends on what kind of config you are using. 
If you are using the THIRDPARTY module - the jar will be called: sirius_openms_plugin_4.0.1.jar
Using the "normal" CLT: sirius_cli_4.0.1.jar

We could set the "sirius_openms_plugin-4.0.1.jar" as a default. 
-> THIRDPARTY/All/Sirius/lib/sirius_openms_plugin-4.0.1.jar

jpfeuffer: Ok. Yes we could do that as a default but:
1) why is it different? Should we change our third-party module?
2) do we need to point to (i.e. put into the PATH) the lib folder then? Because I think the jar is in the lib folder right?

oliveralka: 1. That was a decision we made, when SIRIUS was integrated in the THIRDPARTY module. To reduce the size of SIRIUS for all the three paltforms, all libraries were placed in to "All".  The glpk are OS dependent, that's why they are in the specific OS dirs. To keep calling SIRIUS consistent, there is a bash-script ("binary") in each OS dir to call the jar file. 

2. I think everything should be in the path as long as the .jar file is set correctly. Please check the "lib_class_path" in SiriusAdapterAlgorithm. 
libpath + "/*" - should actually point to the lib dir, since the .jar file is located there as well. 

How are the binaries usually set in jenkins?

oliveralka:
@jpfeuffer 
Still does not find sirius....jar.
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-15,os_label=win10_64/3445/artifact/build/General_Testing/Testing/Temporary/LastConfigure_20191013-1806.log
What has to be done here? Is that more a jenkins config issue?

oliveralka:
Current procedure (for later reference):

The initial SIRIUS executable is used as a reference point (e.g. sirius; sirius-console-64.exe). From there the path to the libraries and jars are collected. With the information, the library- and classpath is created. Then SIRIUS can be called using java, the classpath and the SIRIUS CLI class. 

This is done to gain access to additional debug information when calling SIRIUS fails. SiriusAdapter can now be called using the parameter `-sirius:quiet false -debug 3` to gain additional debug information directly from SIRIUS information.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -312,11 +312,13 @@ namespace OpenMS
       String lib_original = libpath + "/../lib";
       String lib_merged = libpath + "/lib";
       String lib_thirdparty = libpath + "/../../../All/Sirius/lib";
-      String lib_glpk_thirdparty= libpath + "/../../../MacOS/64bit/Sirius/lib" + sep + libpath + "/../../../Linux/64bit/Sirius/lib" + sep +  libpath + "/../../../Windows/64bit/Sirius";
+      String lib_glpk_thirdparty_mac = libpath + "/../../../MacOS/64bit/Sirius/lib"
+      String lib_glpk_thirdparty_lnx = libpath + "/../../../Linux/64bit/Sirius/lib"
+      String lib_glpk_thirdparty_win = libpath;
```
oliveralka: I think that does not make sense since the glpk lib is in the Windows/64bit/Sirius directory. And the jar in the All/Sirius/lib path. 

-> libpath + "/../../../Windows/64bit/Sirius"

oliveralka: ... but it seems to have worked at least for the Adapter. Not sure why the AssayGeneratorMetabo is not working. 
There seems to be something wrong with the -Djava.library.path in this case...

jpfeuffer: I think it works because we are looking for it relative to the executable and not the jar, right? And the executable Sirius...exe is in the Windows subfolder as well.
Yes I also don't have an idea why AGM would work only on macs. I think we would need to debug it on those platforms.

oliveralka: Yes, you are right! 
I will try to debug on Unix later today.

oliveralka:
Debug/Error output is reported if SIRUS invocation fails - see #4725

PR #4236, opened 2019-09-03
Q: Why are we sorting all spectra in TOPPView?

hroest:
I noticed a significant delay with my data when loading them in TOPPView and then opening the 1D view, I found that we are sorting all spectra when adding a new layer. This seems like redundant work and adds a noticeable delay. I wonder

- can we sort the spectra while reading them from mzML (using parallel OpenMP here)?
- can we sort them before display, so only sort those we need (instead of all)?
- can we sort them only once instead of each time they are used for a new layer (maybe keep a flag)?

PS: do not merge this

### DIFF
--- a/src/openms/source/KERNEL/MSSpectrum.cpp
+++ b/src/openms/source/KERNEL/MSSpectrum.cpp
@@ -264,6 +264,8 @@ namespace OpenMS
 
   void MSSpectrum::sortByPosition()
   {
+    if (std::is_sorted(ContainerType::begin(), ContainerType::end(), PeakType::PositionLess())) return;
+
     if (float_data_arrays_.empty() && string_data_arrays_.empty() && integer_data_arrays_.empty())
     {
       std::stable_sort(ContainerType::begin(), ContainerType::end(), PeakType::PositionLess());
@@ -292,6 +294,9 @@ namespace OpenMS
 
   void MSSpectrum::sortByIntensity(bool reverse)
   {
+    if (reverse && std::is_sorted(ContainerType::begin(), ContainerType::end(), reverseComparator(PeakType::IntensityLess()))) return;
+    else if (!reverse && std::is_sorted(ContainerType::begin(), ContainerType::end(), PeakType::IntensityLess())) return;
+
     if (float_data_arrays_.empty() && string_data_arrays_.empty() && integer_data_arrays_.empty())
     {
       if (reverse)
--- a/src/openms/source/KERNEL/MSSpectrum.cpp
+++ b/src/openms/source/KERNEL/MSSpectrum.cpp
@@ -264,7 +264,7 @@ namespace OpenMS
 
   void MSSpectrum::sortByPosition()
   {
-    if (std::is_sorted(ContainerType::begin(), ContainerType::end(), PeakType::PositionLess())) return;
+    if (isSorted()) return;
 
     if (float_data_arrays_.empty() && string_data_arrays_.empty() && integer_data_arrays_.empty())
     {
@@ -340,13 +340,7 @@ namespace OpenMS
 
   bool MSSpectrum::isSorted() const
   {
-    if (this->size() < 2) return true;
-
-    for (Size i = 1; i < this->size(); ++i)
-    {
-      if (this->operator[](i - 1).getMZ() > this->operator[](i).getMZ()) return false;
-    }
-    return true;
+    return std::is_sorted(ContainerType::begin(), ContainerType::end(), PeakType::PositionLess());
   }
 
   bool MSSpectrum::operator==(const MSSpectrum &rhs) const
--- a/src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
+++ b/src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
@@ -1199,6 +1199,10 @@ namespace OpenMS
     }
 
     // sort spectra in ascending order of position (ensure that we sort all spectra as well as the currently
+    for (Size i = 0; i < getCurrentLayer_().getPeakData()->size(); ++i)
+    {
+      (*getCurrentLayer_().getPeakDataMuteable())[i].sortByPosition();
+    }
     getCurrentLayer_().sortCurrentSpectrumByPosition();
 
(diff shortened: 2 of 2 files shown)


src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
```diff
@@ -1199,10 +1199,6 @@ namespace OpenMS
     }
 
     // sort spectra in ascending order of position (ensure that we sort all spectra as well as the currently
-    for (Size i = 0; i < getCurrentLayer_().getPeakData()->size(); ++i)
-    {
-      (*getCurrentLayer_().getPeakDataMuteable())[i].sortByPosition();
-    }
     getCurrentLayer_().sortCurrentSpectrumByPosition();
```
hroest: I get why we need to sort the current spectrum, but why are we sorting all spectra?

timosachsenberg: I think it is a good idea to have the spectra sorted by m/z. Doing this during loading seems as the best option to me. I guess that code did not have much of a performance impact for older instruments.

hroest: I think the issue is that I have ion mobility data and that means sorting involves sorting the float data array as well. First, each spectrum has many more data points and second the code for sorting is much more involved. We can easily have 30k or more data points per spectrum

hroest:
another solution that works very well is to test for `is_sorted` before sorting. There is a performance penalty, but for the branch with float data array it is definitely worthwhile to check (it is guaranteed to be `O(N)` to check while all sorting is `O(N logN)`). 

This patch fixes the issue on my machine, still we may want to ensure sorting when reading in TOPPView since we always want sorted by m/z

src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
hroest: I have now replaced this with an `is_sorted` check which also solves the issue for me

src/openms/source/KERNEL/MSSpectrum.cpp
```diff
@@ -264,6 +264,8 @@ namespace OpenMS
 
   void MSSpectrum::sortByPosition()
   {
+    if (std::is_sorted(ContainerType::begin(), ContainerType::end(), PeakType::PositionLess())) return;
```
hroest: we could move the test down to the `else` clause where it for sure pays off

timosachsenberg:
I think we could make sorting default in the mzML file reader. we could simply track if the data is already sorted and only sort of required. Probably doesn’t come with any penalty on standard files where data is usually already sorted.

hroest:
> I think it is fine to move it to the else block.

it looks like we can leave is_sorted for all data present, see https://stackoverflow.com/questions/6567326/does-stdsort-check-if-a-vector-is-already-sorted as sorting a sorted array always takes longer than check whether its sorted. Also, the extra check if its already sorted (in our case probably > 99% of the time) will take a small amount of time for an unsorted array.

hroest:
PS: turns out we *are* sorting all spectra by mz by default: 

https://github.com/OpenMS/OpenMS/blob/0f1e2b8fd0ae8274e606a16890ee754fa24c22b6/src/openms/source/FORMAT/OPTIONS/PeakFileOptions.cpp#L58

and even more interesting, there is no single point in the code where we set this to `false`: 

```
$ grep -R setSortSpectraByMZ src/
src/pyOpenMS/pxds/PeakFileOptions.pxd:        void setSortSpectraByMZ(bool doSort) nogil except +
src/openms/source/FORMAT/OPTIONS/PeakFileOptions.cpp:  void PeakFileOptions::setSortSpectraByMZ(bool sort)
src/openms/include/OpenMS/FORMAT/OPTIONS/PeakFileOptions.h:    void setSortSpectraByMZ(bool sort);
```

so sorting is basically guaranteed and it happens inside the OpenMP loop, so it is done in parallel:

https://github.com/OpenMS/OpenMS/blob/0f1e2b8fd0ae8274e606a16890ee754fa24c22b6/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp#L148-L166

hroest:
> On loading we could simply track if the data is already sorted and only sort of required. Probably doesn’t come with any penalty on standard files where data is usually already sorted.

hmm but the user could manipulate the underling data, e.g. `spectrum[0].setMZ(99999)` and then its not sorted any more -> I dont think that this really works unless we "freeze" the data and make it read-only

timosachsenberg:
I meant to sort only if needed in the reader. But as this is already implemented it‘s fine ;)

timosachsenberg:
I think some filter tools callable from TOPPView - e.g. intensity based ones sort by intensity. Maybe the code in TOPPView has been added to deal with those? But I was assuming that we also sort on writing the output files? Would be great to get rid of the sorting in TOPPView and replace it by an assert

hroest:
> I think some filter tools callable from TOPPView - e.g. intensity based ones sort by intensity. Maybe the code in TOPPView has been added to deal with those?

or maybe someone wanted just to be safe? In either case, I dont want to go through all code paths and make sure that TOPPView does not *unsort* the spectra at some point (that may indeed be possible). I think the current PR addresses is ok with that, speeds up all sorting (especially with float data arrays) without any chance of a side effect. When I load spectra with ion mobility data in float data arrays, I notice a substantial improvement in speed (e.g. a few seconds wait time down to milliseconds).

merge?

hroest:
I just checked, TOPPView not seem to sort by intensity:

```
$ grep -R sortByIntens src/openms_gui/ | wc
      0       0       0
hr@hr-Precision-5520:~/openmsall/source/OpenMS_dev$ grep -R sortByIntens src/openms/ | wc
     32     124    3469
hr@hr-Precision-5520:~/openmsall/source/OpenMS_dev$ 
```

since all data in TOPPView goes through the mzML reader (I think), any transformation applied by a filter or similar should be read in as mzML and should be then sorted by m/z. So it may actually be safe to remove the sorting completely here.

timosachsenberg:
Fell free to merge

PR #4238, opened 2019-09-04
[FIX] use VersionInfo instead of Macro

hroest:
fixes a small bug where BayesianProteinInferenceAlgorithm.cpp will re-compile every time a git commit is made (due to a change in `openms_package_version.h`. Instead it now uses `VersionInfo::getVersion()` -> much faster compile time. (Also, BayesianProteinInferenceAlgorithm takes really long to compile probably due to all the templates).

### DIFF
--- a/src/openms/source/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.cpp
+++ b/src/openms/source/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.cpp
@@ -31,6 +31,7 @@
 // $Maintainer: Julianus Pfeuffer $
 // $Authors: Julianus Pfeuffer $
 // --------------------------------------------------------------------------
+
 #include <OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h>
 #include <OpenMS/ANALYSIS/ID/MessagePasserFactory.h>
 #include <OpenMS/ANALYSIS/ID/FalseDiscoveryRate.h>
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Thanks. I guess it was the first way that came to my mind. I think I did the same for BasicProteinInference.

Yes, since some of the operations on the tensors are loop-unrolled based on the number of dimensions it can take a bit more time. I think limiting the max number of dimensions might help, I can test that.

hroest:
> Thanks. I guess it was the first way that came to my mind. I think I did the same for BasicProteinInference.
> 

no I think this was the only instance:

```
$ grep -R package_version src/openms
src/openms/configh.cmake:# Create openms_package_version.h
src/openms/configh.cmake:# replace any variables in openms_package_version.h.in with current values
src/openms/configh.cmake:set (CONFIGURED_OPENMS_PACKAGE_VERSION_H ${PROJECT_BINARY_DIR}/include/OpenMS/openms_package_version.h)
src/openms/configh.cmake:configure_file(${PROJECT_SOURCE_DIR}/include/OpenMS/openms_package_version.h.in ${CONFIGURED_OPENMS_PACKAGE_VERSION_H})
src/openms/source/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.cpp:#include <OpenMS/openms_package_version.h>
src/openms/source/CONCEPT/VersionInfo.cpp:#include <OpenMS/openms_package_version.h>
src/openms/include/OpenMS/openms_package_version.h.in:// This is openms_package_version.h / openms_package_version.h.in
src/openms/include/OpenMS/openms_package_version.h.in:// Please do ONLY change openms_package_version.h.in, as openms_package_version.h is automagically
src/openms/include/OpenMS/openms_package_version.h.in:// created by CMAKE from openms_package_version.h.in
src/openms/include/OpenMS/openms_package_version.h.in:// Use appropriate options to configure instead of changing openms_package_version.h.
src/openms/include/OpenMS/openms_package_version.h.in:// Changes made in openms_package_version.h will be lost as soon as you call CMAKE again.
```


merge?

Issue #4239, opened 2019-09-09
Compare SourceFile mzML and featureXML (setPrimaryMSRunPath)


oliveralka:
I check the accordance of the source file names in the AssayGeneratorMetabo. 

```
      StringList mzml_primary_path;
      StringList featurexml_primary_path;
      spectra.getPrimaryMSRunPath(mzml_primary_path);
      feature_map.getPrimaryMSRunPath(featurexml_primary_path);

      if (mzml_primary_path != featurexml_primary_path)
      {
        throw Exception::MissingInformation(__FILE__,
                                            __LINE__,
                                            OPENMS_PRETTY_FUNCTION,
                                            "Path of the original input file do not match in the .mzML and .featureXML files. \n Please check and provide the corresponding files.");
      }
```

With the change in the function "setPrimaryMSRunPath", this does not work anymore. 
The issue is that FFM sets its StringList in as new "PrimaryMSRunPath" in the featureXML. 

So the path from the source file in the mzML (path/File.wiff) does not correspond to the spectra_data in the featureXML (path/File.mzML) anymore. 

One fix would be to use the "in" parameter of the AssayGeneratorMetabo and see if the "PrimaryMSRunPath" of the feautreXML matches. This works as long as everything is run in the same workflow. If the mzML is copied to another location and used with the previous featureXML it does not work anymore. 

I am actually not sure what to do about that since the current setPrimaryMSRunPath would like to have an mzML and an existing source file.

I would need the reference to the original file. Any suggestions?   

```
  void FeatureMap::setPrimaryMSRunPath(const StringList& s, MSExperiment & e)
  {
    StringList ms_path;
    e.getPrimaryMSRunPath(ms_path);
    if (ms_path.size() == 1 && ms_path[0].hasSuffix("mzML") && File::exists(ms_path[0]))
    {
      setPrimaryMSRunPath(ms_path);
    }
    else
    {
      setPrimaryMSRunPath(s);
    }        
  }
```

timosachsenberg:
maybe just turn the exception into a warning if featureXML primaryMSRunPath and in are different?

oliveralka:
Ok, I will check accordance of the input and the feautreXML primaryMSRunPATH and give a warning if it does not match - this should work in my case.  - Thanks!

PR #4241, opened 2019-09-09
[FIX] AGM Warning PrimaryMSRunPath

oliveralka:
Compare input with PrimaryMSRunPath and introduce warning instead of an error.
see #4239

### DIFF
--- a/src/utils/AssayGeneratorMetabo.cpp
+++ b/src/utils/AssayGeneratorMetabo.cpp
@@ -280,19 +280,19 @@ protected:
         OPENMS_LOG_WARN << "Warning: Original paths of the mzML files do not correspond to the featureXML files. Please check and provide the corresponding files." << std::endl;
 
         OPENMS_LOG_WARN << "Input MzML: " << std::endl;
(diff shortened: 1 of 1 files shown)


src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -271,18 +271,29 @@ class TOPPAssayGeneratorMetabo :
...
+      if (in != featurexml_primary_path)
       {
-        throw Exception::MissingInformation(__FILE__,
-                                            __LINE__,
-                                            OPENMS_PRETTY_FUNCTION,
-                                            "Path of the original input file do not match in the .mzML and .featureXML files. \n Please check and provide the corresponding files.");
+        OPENMS_LOG_WARN << "Warning: Original paths of the mzML files do not correspond to the featureXML files. Please check and provide the corresponding files." << std::endl;
+
+        OPENMS_LOG_WARN << "Input MzML: " << std::endl;
+                        for (const auto& it_mzml : in)
```
timosachsenberg: maybe less auto here and below

PR #4243, opened 2019-09-11
[FIX] ChromeleonFile can handle commas as thousands separator

pasqoo:


src/openms/source/FORMAT/ChromeleonFile.cpp
```diff
@@ -113,11 +113,11 @@ namespace OpenMS
     while (!ifs.eof())
     {
       TextFile::getLine(ifs, line);
-      double rt, intensity;
-      int ret = std::sscanf(line.c_str(), "%lf\t%*s\t%lf", &rt, &intensity);
+      char rt[128], intensity[128];
+      int ret = std::sscanf(line.c_str(), "%s\t%*s\t%s", rt, intensity);
```
cbielow: since the buffer is limited to 128, you should use `%128s\t%*s\t%128s` to avoid buffer overflows.

src/openms/source/FORMAT/ChromeleonFile.cpp
```diff
@@ -131,4 +131,9 @@ namespace OpenMS
     ifs.close();
     experiment.addChromatogram(chromatogram);
   }
+
+  double ChromeleonFile::parseDouble(String number) const
```
cbielow: could you rename that function to `parseStringWithComma` or something along those lines?
Because from the signature the reader wonders why you are not simply using the vanilla String::toDouble()

pasqoo: I'm going with `removeCommasAndParseDouble()`

cbielow:
+1 (if pyopenms succeeds)

src/openms/include/OpenMS/FORMAT/ChromeleonFile.h
```diff
@@ -67,6 +67,15 @@ namespace OpenMS
...
     void load(const String& filename, MSExperiment& experiment) const;
+
+    /**
+      @brief Remove commas from the string (used as thousands separators) and
+      parse its value
+
+      @param[in] number A string representing a floating-point number
+      @return The value converted to `double`
+    */
+    double removeCommasAndParseDouble(String number) const;
```
hroest:
```suggestion
    double removeCommasAndParseDouble(const String& number) const;
```

src/openms/source/FORMAT/ChromeleonFile.cpp
```diff
@@ -113,11 +113,11 @@ namespace OpenMS
     while (!ifs.eof())
     {
       TextFile::getLine(ifs, line);
-      double rt, intensity;
-      int ret = std::sscanf(line.c_str(), "%lf\t%*s\t%lf", &rt, &intensity);
+      char rt[128], intensity[128];
+      int ret = std::sscanf(line.c_str(), "%127s\t%*s\t%127s", rt, intensity);
```
hroest: wouldnt it be easier to use String tools instead of error-prone `sscanf`, e.g. 
```
line.split('\t', result);
```

hroest:
looks good to me

PR #4244, opened 2019-09-11
[FEATURE] createTransition_() adds the metavalue LabelType to the transition

pasqoo:
Would you recommend a different way to add the `LabelType` information to a transition?
I went for a metavalue.

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/TransitionTSVFile.cpp
@@ -792,6 +792,7 @@ namespace OpenMS
   {
     // the following attributes will be stored as meta values (userParam):
     // - annotation (as by SpectraST)
+    // - label type (e.g. "heavy" or "light")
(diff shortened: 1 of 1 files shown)


pasqoo:
Some CIs crashed, but TOPP tests data files need to be updated.
I'll do it once it's confirmed that using the `MetaInfoInterface` is the way to go for `LabelType`.

hroest:
this is already added to the peptide/compound - is there a need to add it to the transition as well?

pasqoo:
I tried fetching the info from the peptides, but I couldn't find a way to "locate" the correct peptide, given a `transition_name` from the traml.csv file (we briefly discussed this in February, on Gitter).

hroest:
> I tried fetching the info from the peptides, but I couldn't find a way to "locate" the correct peptide, given a `transition_name` from the traml.csv file (we briefly discussed this in February, on Gitter).

I suggest to do 
```
auto pep = exp.getPeptideByRef(transition.getPeptideRef())
```

PR #4245, opened 2019-09-11
QualityControl TOPP tool

cbielow:
supporting label-free pipelines (for now)

The output is mzTab which can be used to create QC reports (e.g. using PTXQC -- an updated version which supports mzTab will be released soon -- currently available as a branch)

The PDF report then somewhat looks like this (on a minimal example, thus not very pretty)
[report_v0.92.6_outmztab.pdf](https://github.com/OpenMS/OpenMS/files/3601479/report_v0.92.6_outmztab.pdf)

There is a Doxygen Tutorial Section (`TOPP Tutorial -> Quality Control`) and a example pipeline in `OpenMS/share/OpenMS/examples/TOPPAS/QualityControl.toppas`


hroest:
please fix these two issues:


```
src/openms/source/FORMAT/MzTab.cpp:2575:22: error: redeclaration of ‘std::vector<OpenMS::String> pid_keys’


src/openms/source/FORMAT/MzTab.cpp:2584:31: error: redeclaration of ‘std::set<OpenMS::String> pid_key_set’

```

src/openms/CMakeLists.txt
```diff
@@ -93,7 +93,7 @@ set(OPENMS_DEP_LIBRARIES  ${COIN_LIBRARIES}
                           ${BZIP2_LIBRARIES}
                           ${ZLIB_LIBRARIES}
                           ${SQLite_LIBRARY}
-                          ${HDF5_CXX_LIBRARIES}
+                          #${HDF5_CXX_LIBRARIES}
```
hroest: @cbielow please undo this change

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/WindowMower.h
```diff
@@ -176,45 +176,26 @@ namespace OpenMS
...
       }
 
-      if (peaks_in_window.empty()) // last window is empty -> no special handling needed
+      if (!peaks_in_window.empty()) // last window is not empty
       {
-        // select peaks that were retained
-        std::vector<Size> indices;
-        for (typename SpectrumType::ConstIterator it = spectrum.begin(); it != spectrum.end(); ++it)
+        // Note that the last window might be much smaller than windowsize.
+        // Therefor the number of peaks copied from this window should be adapted accordingly.
```
hroest:
```suggestion
        // Therefore the number of peaks copied from this window should be adapted accordingly.
```

src/tests/class_tests/openms/source/FragmentMassError_test.cpp
```diff
@@ -0,0 +1,291 @@
...
+    PeakSpectrum ms_spec_2_alabama = createMSSpectrum(2, 2, "XTandem::2", Precursor::ActivationMethod::ECD);
+    TheoreticalSpectrumGenerator theo_gen_al;
+    Param theo_gen_settings_al = theo_gen_al.getParameters();
+    theo_gen_settings_al.setValue("add_c_ions", "true");
+    theo_gen_settings_al.setValue("add_z_ions", "true");
+    theo_gen_settings_al.setValue("add_b_ions", "false");
+    theo_gen_settings_al.setValue("add_y_ions", "false");
+    theo_gen_al.setParameters(theo_gen_settings_al);
+    theo_gen_al.getSpectrum(ms_spec_2_alabama, AASequence::fromString("ALABAMA"), 2, 2);
+    for(Peak1D& peak : ms_spec_2_alabama) peak.setMZ(peak.getMZ() + 0.001);
```
timosachsenberg: I think it would be easier to verify tests if instead of 0.001 e.g. ground_truth_ppm * peak.getMZ() was used.

src/tests/class_tests/openms/source/TIC_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+START_SECTION(const String& getName() const override)
+  TEST_EQUAL(tic.getName(), "TIC")
+END_SECTION
+
+START_SECTION(Status requires() const override)
+  TEST_EQUAL((tic.requires() == QCBase::Status(QCBase::Requires::RAWMZML)),true);
+END_SECTION
+
+START_SECTION(void compute(const MSExperiment &exp, float bin_size))
+  NOT_TESTABLE
```
timosachsenberg: why not testable? Do we have test mzML file with TIC already in our test files?

hroest:
looks good, @cbielow any plans to add pyOpenMS bindings? I think these should be relatively easy for all the classes you added.

cbielow:
ok, tests look good (just the upload to cdash seems to be broken).
Python bindings are on the agenda for an upcoming PR (there is new stuff already in prep) and I dont want to make this PR even more complicated.

ready for merge.

jpfeuffer:
Also this broke our documentation. Is doxygen minimum 1.3.16 needed now?

jpfeuffer:
Hmm I started to upgrade on MacOS but it doesn't solve the problem. Any ideas?
It might be the table in the new section.

cbielow:
solved it. PR coming up.

PR #4249, opened 2019-09-13
[FEATURE] New features in Deisotoper

enetz:
The Deisotoper can now annotate the number of peaks in each isotopic pattern. The averagine model filtering can be turned off. The intensities of each isotopic pattern can be summed up on top of the monoisotopic peak.
These options were added as boolean flags.

I merged the functionality of my own deisotoping function that was used exclusively by OpenPepXL into the Deisotoper. OpenPepXL also uses the Deisotoper now and I removed the redundant function.


src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
```diff
@@ -64,6 +64,9 @@ class OPENMS_DLLAPI Deisotoper
    * @param [max_isopeaks] The maximum number of isotopic peaks (at least 2) considered for an isotopic cluster
    * @param [make_single_charged] Convert deisotoped monoisotopic peak to single charge
    * @param [annotate_charge] Annotate the charge to the peaks in the IntegerDataArray: "charge" (0 for unknown charge)
+   * @param [annotate_iso_peak_count] Annotate the number of isotopic peaks in a pattern for each monoisotopic peak in the IntegerDataArray: "iso_peak_count"
+   * @param [use_averagine] Use a simple averagine model that expects heavier isotopes to have less intensity. If false, no intensity checks are applied.
```
hroest: I would not call that averagine since it really is a much simpler model that assumes you have less than 60 or so Carbon molecules in your analyte. I would argue that using averagine here is actually misleading, maybe lets try `use_decreasing_model` or something that is more explicit

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -113,13 +135,17 @@ void Deisotoper::deisotopeAndSingleCharge(MSSpectrum& spec,
           {
             // Possible improvement: include proper averagine model filtering. for now start at the second peak to test hypothesis
             // Note: this is a common approach used in several other search engines
-            if (old_spectrum[p].getIntensity() > old_spectrum[extensions.back()].getIntensity())
+            if (use_averagine && (old_spectrum[p].getIntensity() > old_spectrum[extensions.back()].getIntensity()))
```
hroest: @timosachsenberg again, the commend says "start at the second peak" but the code doesnt do that... maybe add something like `extensions.size() > 1` here?

timosachsenberg: @netz can you add this or change the comment?

hroest: I think adding this would actually make sense, to have a legacy, decrasing and none model here

enetz: Should I add an additional flag then?

Also I think there might be different interpretations of the comment.
The compared intensities are from the peaks `old_spectrum[extensions.back()]`
and `old_spectrum[p]`. In the first iteration of this loop, `old_spectrum[extensions.back()]` is the monoisotopic peak (it is added before this loop starts by default) and `old_spectrum[p]` is the closest peak to the monoisotopic mass + 1 . It tests whether the monoisotopic peak has a higher intensity than the second one, so it starts at the second peak, right?

What you are suggesting is to add a mode where the monoisotopic peak can have a lower intensity than the second one. Skipping this first comparison and starting with comparing the second to the third peak. I would consider that "starting with the third peak", because at this point the second peak is already accepted and the third is the one that is tested.
Did I get that right?

AlexanderMoysa:
Dear Eugen Netz, Is it possible to use Deisotoper for obtaining intensities of all peaks from the isotope pattern of peptide?

PR #4250, opened 2019-09-13
Feature: Ion mobility scoring

hroest:
- add ion mobility scoring to OpenSWATH


src/openms/source/KERNEL/MRMFeature.cpp
```diff
@@ -147,5 +147,29 @@ namespace OpenMS
     return precursor_features_.at(precursor_feature_map_.at(key));
   }
 
+  void MRMFeature::IDScoresAsMetaValue(bool decoy, const OpenSwath_Ind_Scores& idscores)
+  {
+    String id = "id_target_";
+    if (decoy) id = "id_decoy_";
+
+    this->setMetaValue(id + "transition_names", idscores.ind_transition_names);
```
timosachsenberg: remove this-> ?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/IonMobilityScoring.h
```diff
@@ -0,0 +1,158 @@
...
+      @param spectrum The DIA MS2 spectrum found at the peak apex
+      @param transitions The transitions used for scoring
+      @param scores The output scores
+      @param drift_lower Ion Mobility extraction start
+      @param drift_upper Ion Mobility extraction end
+      @param drift_target Ion Mobility extraction target
+      @param dia_extraction_window_ m/z extraction width
+      @param dia_extraction_ppm_ Whether m/z extraction width is in ppm
+      @param use_spline Whether to use spline for fitting
+      @param drift_extra Extra extraction to use for drift time (in percent)
```
timosachsenberg: from the comment, it's not clear to me what drift_extra is

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -66,6 +66,151 @@ namespace OpenMS
...
+        {
+          integratedWindowsIntensity.push_back(intensity);
+          integratedWindowsMZ.push_back(mz);
+        }
+        else if (!remZero)
+        {
+          integratedWindowsIntensity.push_back(0.);
+          integratedWindowsMZ.push_back(*beg);
+        }
+        else
```
timosachsenberg: why empty else block?

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -66,6 +66,151 @@ namespace OpenMS
...
+        itType mz_arr_end = spectrum->getMZArray()->data.end();
+        itType int_it = spectrum->getIntensityArray()->data.begin();
+
+        // this assumes that the spectra are sorted!
+        itType mz_it = std::lower_bound(spectrum->getMZArray()->data.begin(),
+          spectrum->getMZArray()->data.end(), mz_start);
+        itType mz_it_end = std::lower_bound(mz_it, mz_arr_end, mz_end);
+
+        // also advance intensity iterator now
+        std::iterator_traits< itType >::difference_type iterator_pos = std::distance((itType)spectrum->getMZArray()->data.begin(), mz_it);
```
timosachsenberg: should we use auto instead of std::iterator_traits< itType >::difference_type here?

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
```diff
@@ -0,0 +1,476 @@
...
+#include <OpenMS/MATH/STATISTICS/StatisticFunctions.h>
+#include <OpenMS/ANALYSIS/OPENSWATH/SpectrumAddition.h>
+#include <OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h>
+
+// #define DEBUG_IMSCORING
+
+namespace OpenMS
+{
+
+  typedef std::pair<double, double> MobilityPeak;
```
timosachsenberg: would a struct be better? I prefer .mz over .first

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
```diff
@@ -0,0 +1,476 @@
...
+  {
+    auto pr_it = profile.begin();
+    max_peak_idx = 0;
+    double max_int = 0;
+    for (Size k = 0; k < im_grid.size(); k++)
+    {
+      // In each iteration, the IM value of pr_it should be equal to or
+      // larger than the master container. If it is equal, we add the current
+      // data point, if it is larger we add zero and advance the counter k.
+      if (pr_it != profile.end() && fabs(pr_it->first - im_grid[k] ) < 1e-4 )
```
timosachsenberg: is the 1e-4 a check for floating-point equality?

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
```diff
@@ -0,0 +1,476 @@
...
+      }
+      else
+      {
+        al_int_values.push_back(0.0);
+        al_im_values.push_back( im_grid[k] );
+      }
+      // OPENMS_LOG_DEBUG << "grid position " << im_grid[k] << " profile position " << pr_it->first << std::endl;
+
+      // check that we did not advance past 
+      if (pr_it != profile.end() && (im_grid[k] - pr_it->first) > 1e-3)
```
timosachsenberg: same here for the 1e-3?

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
```diff
@@ -0,0 +1,476 @@
...
+                              double & intensity,
+                              IonMobilogram& res, 
+                              double drift_start,
+                              double drift_end)
+  {
+    OPENMS_PRECONDITION(spectrum->getDriftTimeArray() != nullptr, "Cannot filter by drift time if no drift time is available.");
+
+    // rounding multiplier for the ion mobility value
+    // TODO: how to improve this -- will work up to 42949.67296
+    double IM_IDX_MULT = 10e5;
```
timosachsenberg: maybe add more comment for projection to ints?
Are there problems with very similar values to spread over bins? i.e. for binned spectra some offsets can be used to reduce the effect. does something similar apply here?

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
```diff
@@ -0,0 +1,476 @@
...
+    OPENMS_PRECONDITION(spectrum != nullptr, "Spectrum cannot be null");
+    OPENMS_PRECONDITION(spectrum->getDriftTimeArray() != nullptr, "Cannot score drift time if no drift time is available.");
+
+    if (spectrum->getDriftTimeArray() == nullptr)
+    {
+      OPENMS_LOG_DEBUG << " ERROR: Drift time is missing in ion mobility spectrum!" << std::endl;
+      return;
+    }
+
+    double eps = 1e-5; // eps for two grid cells to be considered equal
```
timosachsenberg: another constant for equality here

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
```diff
@@ -168,6 +178,8 @@ namespace OpenMS
     {
       score = feature.getMetaValue(score_name).toString();
     }
+    if (score == "nan") score = "NULL";
```
timosachsenberg: is the string "nan" on all platforms (e.g., not NaN)?

src/openms/source/ANALYSIS/OPENSWATH/SwathMapMassCorrection.cpp
```diff
@@ -42,19 +42,295 @@
...
+    std::vector<double> exp_im;
+    std::vector<double> theo_im;
+#ifdef _OPENMP
+#pragma omp parallel for 
+#endif
+    for (SignedSize k = 0; k < (SignedSize)trgr_ids.size(); k++)
+    {
+      // we need at least one feature to find the best one
+      auto transition_group = transition_group_map.at(trgr_ids[k]);
+      if (transition_group->getFeatures().size() == 0)
```
timosachsenberg: .empty() ?

src/openms/source/ANALYSIS/OPENSWATH/SwathMapMassCorrection.cpp
```diff
@@ -42,19 +42,295 @@
...
+      for (const auto& tr : transition_group->getTransitions())
+      {
+        if (ms1_im_) {continue;}
+        double intensity(0), im(0), left(tr.product_mz), right(tr.product_mz);
+
+        auto pepref = tr.getPeptideRef();
+        // get drift time upper/lower offset (this assumes that all chromatograms
+        // are derived from the same precursor with the same drift time)
+        double drift_target = pep_im_map[pepref];
+        double drift_lower_used = drift_target - im_extraction_win;
```
timosachsenberg: just to double check. is -/+ im_extraction_win correct? I think I recall that somewhere half of a window was used (but might have been mz window).

timosachsenberg:
PR looks good to me. If @grosenberger has more experience with IM data it would be great if he could give it a quick look as I did not dig into the details of e.g. projection to the grid. Feel free to merge if tests pass.

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
hroest: yes that is basically the idea, we assume that the ion mobility values are discrete and use an epsilon cutoff to counter rounding errors

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp
hroest: fixed it by using `toLower()`

src/openms/source/ANALYSIS/OPENSWATH/SwathMapMassCorrection.cpp
hroest: yes, of course. You are right.

cbielow:
nice PR! how much better does it get with IM in terms of #quantifyable peptides? do you have benchmark data somewhere?

hroest:
> nice PR! how much better does it get with IM in terms of #quantifyable peptides? do you have benchmark data somewhere?

yes, we increase up to 3x in peptide identifications, see https://www.biorxiv.org/content/10.1101/656207v1

hroest:
@timosachsenberg @cbielow ready to merge?

cbielow:
there are a few leftover comments.
What bugs me most is passing SpectrumPtr's .. I know its used throughout OSW but its still wrong IMHO. But if you really really want it that way... feel free to merge.

timosachsenberg:
Same here. Feel free to merge but consider an MSSpectrum based interface if there are no good reasons against it. It would make some of the functions easier to use. Some seem generic enough to be used outside of the SWATH context.

hroest:
> Feel free to merge but consider an MSSpectrum based interface if there are no good reasons against it.

that would be a major undertaking and actually not that easy at the current moment. All of our abstractions are based on the OpenSwath::Spectrum and while we can convert back and forth we do all processing on the lightweight spectrum implementation.

PR #4251, opened 2019-09-16
[STYLE] Improve EmgGradientDescent's methods names

pasqoo:
Also move said methods from `protected` to `public`.


timosachsenberg:
+1

PR #4255, opened 2019-09-18
[REFACTOR] some cleanup of PeptideIndexing...

cbielow:
purely refactoring out some clutter which has accumulated ...
one minor fix: https://github.com/OpenMS/OpenMS/compare/develop...cbielow:develop#diff-bf0e280335f966eff190a79052d2921bR237

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
@@ -192,24 +192,25 @@ public:
       // no decoy string provided? try to deduce from data
       if (decoy_string_.empty())
       {
-        bool is_decoy_string_auto_successful = findDecoyString_(proteins);
+        auto r = DecoyHelper::findDecoyString(proteins);
+        proteins.reset();
 
-        if (!is_decoy_string_auto_successful && contains_decoys_)
+        if (!r.success && r.has_decoys)
         {
           return DECOYSTRING_EMPTY;
         }
-        else if (!is_decoy_string_auto_successful && !contains_decoys_)
+        else if (!r.success && !r.has_decoys)
         {
           OPENMS_LOG_WARN << "Unable to determine decoy string automatically, not enough decoys were detected! Using default " << (prefix_ ? "prefix" : "suffix") << " decoy string '" << decoy_string_ << "\n"
-                   << "If you think that this is false, please provide a decoy_string and its position manually!" << std::endl;
+                          << "If you think that this is false, please provide a decoy_string and its position manually!" << std::endl;
         }
         else
         {
-          // decoy string and position was extracted successfully
-          OPENMS_LOG_INFO << "Using " << (prefix_ ? "prefix" : "suffix") << " decoy string '" << decoy_string_ << "'" << std::endl;
+          prefix_ = r.is_prefix;
+          decoy_string_ = r.name;
         }
-
-        proteins.reset();
+        // decoy string and position was extracted successfully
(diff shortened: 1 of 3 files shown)


src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -744,220 +740,39 @@ namespace OpenMS
...
-      /// the amino acid before the peptide in the protein
-      char AAAfter;
+    struct PeptideProteinMatchInformation
+    {
+      OpenMS::Size protein_index; //< index of the protein the peptide is contained in
+      OpenMS::Int position; //< the position of the peptide in the protein
+      char AABefore; //< the amino acid after the peptide in the protein
+      char AAAfter; //< the amino acid before the peptide in the protein
 
+      const auto tie() const
```
timosachsenberg: smells like > x11

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
```diff
@@ -228,20 +229,15 @@ namespace OpenMS
...
       // determine if search engine is solely xtandem or MSGFPlus
       for (const auto& prot_id : prot_ids)
       {
-        if (!msgfplus_fix_parameters && !xtandem_fix_parameters) { break; }
-        String se = prot_id.getSearchEngine();
-        std::string search_engine = StringUtils::toUpper(se);
+        std::string search_engine = StringUtils::toUpper(String(prot_id.getSearchEngine()));
         if (search_engine != "XTANDEM") { xtandem_fix_parameters = false; }
-        if (search_engine != "MSGFPLUS" || "MS-GF+") { msgfplus_fix_parameters = false; }
+        if (!(search_engine == "MSGFPLUS" || search_engine == "MS-GF+")) { msgfplus_fix_parameters = false; }
```
timosachsenberg: nice find

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
hroest: https://en.cppreference.com/w/cpp/utility/tuple/tie -- since C++11

src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
```diff
@@ -329,5 +333,158 @@ class FASTAContainer<TFI_Vector>
...
+  struct Result
+  {
+    bool success;
+    bool has_decoys;
+    String name;
+    bool is_prefix;
+  };
+
+  template<typename T>
+  static Result findDecoyString(FASTAContainer<T>& proteins)
```
hroest: can you add some documentation what this does?

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
timosachsenberg: But also Auto Return value?

hroest: true ... I dont think that works yet

cbielow:
tests look good.
MSVC fails early and unrelated (http://cdash.openms.de/viewBuildError.php?buildid=332854)

ready for merge.

src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
cbielow: it works on MSVC2015.. and I thought I give it a try. Well...

PR #4259, opened 2019-09-19
[FEATURE] some memory fixes for ProteomicsLFQ

hroest:
this reduces the featureXML produced (debug output) by about 10x, from > 250 MB to less than 25 MB and allows me to run iPRG 2015 study with about 2GB of system memory (measurement currently ongoing).


hroest:
```
ProteomicsLFQ took 51:32 m (wall), 51:17 m (CPU), 8.28 s (system), 51:09 m (user).      
3069.44user 8.38system 51:32.34elapsed 99%CPU (0avgtext+0avgdata 3250856maxresident)k                                                              9722136inputs+1428424outputs (12major+1518145minor)pagefaults 0swaps 
```

PR #4262, opened 2019-09-20
[FEATURE] Deisotoper precursor filter and decreasing model fine tuning

enetz:
For every peak masses and charges that in combination would result in
MS2-peak-mass > precursor-mass are skipped now, since we don't expect to see fragments with more mass than the precursor. The hope is that this will avoid some wrong charge assignments and deisotoping of some noise peaks.

Very few fragment annotations in the test files of RNPxlSearch were removed and some scores changed slightly, so it looks like there were cases of this that are now filtered out successfully.

This is only applied to MS Spectra with Precursor data.
I don't know whether there might be special cases where fragment spectra would contain stuff more massive than the precursor mass + tolerance. Or should I consider the upper bound of the isolation window instead? Do we need a flag to be able to turn this off?

Also added a parameter `start_intensity_check` to adjust the averagine / decreasing intensity mode, as mentioned in #4249:
you can set a number and the intensity check will be performed only from that peak on,
e.g. for `start_intensity_check=2` the monoisotopic peak can be less intense than the second peak and for `start_intensity_check=3` both the monoisotopic and the second peak can be less intense than the third.
I set the default to `2` and updated some test files accordingly.
Alternatively we can change the deisotoper settings in each tool that uses it.


enetz:
An example for I and L combinations:

```
HI
HII
HIIK
HIK
HIKI
HIKL
HIL
HILK
HL
HLI
HLIK
HLK
HLKI
HLKL
HLL
HLLK
```

And for modifications:

```
EV
EVC(Carbamidomethyl)
EVC(Carbamidomethyl)V
```

hroest:
@enetz please fix the SwathQC test:

```
checking SwathQC() ... 
 +  line 72:  TEST_NOT_EQUAL(ptr,nullPointer): got '0x557c0b4c98b0', forbidden is '0'
: passed

checking ~SwathQC() ... : passed

checking (static ChargeDistribution getChargeDistribution(const std::vector<SwathMap>& swath_maps, const size_t nr_samples, const double mz_tol)) ... 
 -  line 95:  TEST_EQUAL(cd.size(),cde.size()): got '7', expected '5'
 -  line 96:  TEST_EQUAL(cd == cde,true): got '0', expected '1'
: failed

checking (static void storeJSON(const OpenMS::String& filename)) ... FAILED: 'ratio of numbers is too large'

  input:	in1	in2
  line:	5	5
  pos/col:	6/7	6/7
 --------------------------------
  is_number:	true	true
  numbers:	17	13
  is_space:	false	false
  is_letter:	false	false
  letters:	"1"	"1"
  char_codes:	49	49
 --------------------------------
  relative_max:        1.30769
  relative_acceptable: 1
 --------------------------------
  absolute_max:        4
  absolute_acceptable: 0

Offending lines:			(tab_width = 8, first_column = 1)

in1:  /tmp/20190920_161548_350cc65e0c55_9095_1   (line: 5, position/column: 6/7)
      !
      17

in2:  /tmp/20190920_161548_350cc65e0c55_9095_2   (line: 5, position/column: 6/7)
      !
      13


Easy Access:
/tmp/20190920_161548_350cc65e0c55_9095_1:5:7:
/tmp/20190920_161548_350cc65e0c55_9095_2:5:7:

diff /tmp/20190920_161548_350cc65e0c55_9095_1 /tmp/20190920_161548_350cc65e0c55_9095_2

 -  line 189:  TEST_EQUAL(FuzzyStringComparator().compareFiles(tmp_json, tmp_expected),true): got 'false', expected 'true'
: failed

: passed

FAILED
Failed lines: 95 96 189 
```

hroest:
@enetz adding additional default values to a function **in the middle** is *very dangerous* since you did not adjust other functions and parameters may now be completely out of whack!

please check and fix:
```
$ grep -R deisotopeAndSingleCharge src/openms
src/openms/source/ANALYSIS/OPENSWATH/SwathQC.cpp:        tmp = spec; // make a copy, since deisotopeAndSingleCharge() will modify 
src/openms/source/ANALYSIS/OPENSWATH/SwathQC.cpp:      Deisotoper::deisotopeAndSingleCharge(tmp, this->decon_ms1_mz_tol_, false, 1, 10, true, 3, 10, false, true);
src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp:      Deisotoper::deisotopeAndSingleCharge(exp[exp_index], 
src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp:        Deisotoper::deisotopeAndSingleCharge(deisotoped,
src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp:void Deisotoper::deisotopeAndSingleCharge(MSSpectrum& spec,
src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h:  static void deisotopeAndSingleCharge(MSSpectrum& spectrum,
```

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
```diff
@@ -94,6 +98,7 @@ class OPENMS_DLLAPI Deisotoper
             bool annotate_charge = false,
             bool annotate_iso_peak_count = false,
             bool use_decreasing_model = true,
+            unsigned int start_intensity_check = 2,
             bool add_up_intensity = false);
```
hroest: @enetz if you add new default arguments, please always add at the end to prevent the true/false of another fxn to be interpreted as `start_intensity_check`

enetz:
@hroest thanks for pointing out the parameter position issue, but the other parameters around this new one were also added just a short while ago with my last PR and the only tool that changes these defaults so far is OpenPepXL. There shouldn't be any issue this time. But I will recheck this to make sure.

On the same note I would like to get some feedback about the defaults from @hroest @timosachsenberg and @cbielow :
1. Is it fine to change the default behaviour to only check the intensity from the second peak and ignore the intensity of the monoisotopic peak? Is this the default we want? Or should I change it back to the old one?
2. related to this, if we keep the new default, should I update the tests to the new default, or change the parameters of the tools that currently use the Deisotoper to the previous behaviour? This should be decided on a tool by tool basis. I Updated the tests for OpenPepXL, SimpleSearchEngine and RNPxlSearch to the new default. @cbielow should we do that for SwathQC as well or would you prefer to use the old Deisotoper behaviour?

hroest:
The issue is that this time a boolean value is interpreted as int leading
to unintended behavior. Also, why are we not using the Default Param object
here? Is it because of speed? Because a function with so many parameters is
ugly...

I think the current defaults are ok. They make sense for proteomics so
let's keep them

enetz:
@hroest The argument against `DefaultParamHandler` here would be that this is only one static function and can be called with `Deisotoper::deisotopeAndSingleCharge(...)` and there is no need to initialize a `Deisotoper` object. But you are right that with this many parameters this is starting to get ugly.
@timosachsenberg should we refactor `Deisotoper` to use `DefaultParamHandler`?

Otherwise this PR should be finished content wise, and ready to merge if the tests pass.

enetz:
@hroest @timosachsenberg the pyopenms memleak test is the only one that failed, I think it is unrelated to the changes here.
Can we merge this?

src/tests/topp/SimpleSearchEngine_1_out.idXML
```diff
@@ -18,14 +18,14 @@
 			<UserParam type="stringList" name="spectra_data" value="[file://SimpleSearchEngine_1.mzML]"/>
 		</ProteinIdentification>
 		<PeptideIdentification score_type="hyperscore" higher_score_better="true" significance_threshold="0.0" MZ="520.262817382812045" RT="2655.095703125" >
-			<PeptideHit score="20.34238715336474" sequence="DFASSGGYVLHLHR" charge="3" aa_before="[" aa_after="E" start="0" end="13" protein_refs="PH_0" >
+			<PeptideHit score="12.094764845630944" sequence="DFASSGGYVLHLHR" charge="3" aa_before="[" aa_after="E" start="0" end="13" protein_refs="PH_0" >
```
hroest: why are these results different @timosachsenberg is that expected ?

hroest:
> please check and fix:
> 
> ```
> $ grep -R deisotopeAndSingleCharge src/openms
> src/openms/source/ANALYSIS/OPENSWATH/SwathQC.cpp:        tmp = spec; // make a copy, since deisotopeAndSingleCharge() will modify 
> src/openms/source/ANALYSIS/OPENSWATH/SwathQC.cpp:      Deisotoper::deisotopeAndSingleCharge(tmp, this->decon_ms1_mz_tol_, false, 1, 10, true, 3, 10, false, true);
> src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp:      Deisotoper::deisotopeAndSingleCharge(exp[exp_index], 
> src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp:        Deisotoper::deisotopeAndSingleCharge(deisotoped,
> src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp:void Deisotoper::deisotopeAndSingleCharge(MSSpectrum& spec,
> src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h:  static void deisotopeAndSingleCharge(MSSpectrum& spectrum,
> ```

@enetz I suggest you need to fix all of the above or move the default argument to the end. Otherwise I am not comfortable to merge.

enetz:
@hroest I did check all of them. They all use only up to the first 10 parameters of the function. It now has 14 and the newest one I added is now the 13th. So for the last 4 parameters including the new one in this PR they all use the default values. So far the only exception is in `OPXLSpectrumProcessingAlgorithms.cpp` and I fixed and tested that.

There are still some tools, like NASE, that use their own deisotoping functions. After this PR is merged we should be able to replace them with this generalized function.

src/tests/topp/SimpleSearchEngine_1_out.idXML
timosachsenberg: @enetz this is indeed a bit strange that the hyper score decreases that much. Can you check

timosachsenberg: @enetz can you check. means that significantly less b/y ions have been matched

enetz: @timosachsenberg I just ran this test with `start_intensity_check = 1`, which is effectively the old default averagine model. The score is then `19.24...`. This smaller left over difference should be due the precursor charge / mass filter.
So the score dropped by loosening the decreasing intensity requirement. This probably lead to accepting some lower intensity and lower mass peaks as new monoisotopic peaks for previously matched masses.
Since the SimpleSearchEngine does not write out peak annotations or numbers of matching peaks, it is difficult to confirm what is happening there and which score is more correct. Do you have a ground truth for these spectra? We could adjust the parameter to the old behaviour, if you think it is more correct.

timosachsenberg: Ok, I see. Maybe we should give this a test on a larger dataset just to confirm that the new behavior does not lead to a large drop in ids at the same e.g. 1% PSM FDR.

timosachsenberg:
rebuild jenkins

src/tests/topp/SimpleSearchEngine_1_out.idXML
timosachsenberg: Eugen did some additional tests and the ID rate did not chagne on the iPRG2015.
So I guess it is fine to merge.

PR #4263, opened 2019-09-20
[Feature] more sensitive Tagger

enetz:
used Tagger (#3939) by @timosachsenberg  and @cbielow  as a basis and started adding several features:
- Multiple charges can be considered now
- Fixed and Variable Modifications (as explicit Strings, e.g. M(Oxidation) )
- Some non-standard residues were removed from the residue list
- The equal mass residues I and L are handled explicitly and redundantly by making tags from all combinations

Also added a simple, brute force function to filter XLMS search candidates by sequence tags and added sequence tag filtering as an optional advanced parameter to OpenPepXLLF as an example use case.

I turned the simple previous structures of chars and strings into strings and StringLists to be able to handle modified residues explicitly.
An alternative would be to add them with the modified masses but only the origin char. This might make the Tagger more efficient, but make the filtering of candidates less specific by using unmodified sequence strings. I have not compared how much of a difference there would be in runtime.


src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -405,6 +415,20 @@ using namespace OpenMS;
       vector< OPXLDataStructs::CrossLinkSpectrumMatch > top_csms_spectrum;
       vector< OPXLDataStructs::ProteinProteinCrossLink > cross_link_candidates = OPXLHelper::collectPrecursorCandidates(precursor_correction_steps_, precursor_mass, precursor_mass_tolerance_, precursor_mass_tolerance_unit_ppm_, filtered_peptide_masses, cross_link_mass_, cross_link_mass_mono_link_, cross_link_residue1_, cross_link_residue2_, cross_link_name_);
 
+      if (use_sequence_tags_)
+      {
+        tagger.setMaxCharge(precursor_charge);
+        std::set<std::string> tags;
+        tagger.getTag(spectrum, tags);
+        std::cout << "TEST TAG NEW SPECTRUM: " << scan_index << std::endl;
```
timosachsenberg: remove or put to debug output

enetz:
An example for I and L combinations:

```
HI
HII
HIIK
HIK
HIKI
HIKL
HIL
HILK
HL
HLI
HLIK
HLK
HLKI
HLKL
HLL
HLLK
```

And for modifications:

```
EV
EVC(Carbamidomethyl)
EVC(Carbamidomethyl)V
```

timosachsenberg:
Nice. Some comments:
- Can you check how much of a difference the switch from char to string makes in terms of performance? As this *could* be orders of magnitude it would be great to check.
- I would just use either I or L and handle ambiguities in the comparison (I think there is a residue set for that)

enetz:
@timosachsenberg 
- I will make a test class for the tagger next and include some repetitive tagging of the same spectrum to see performance changes
- There is a residue J that means either I or L I think. We can use that when we have a better function for the comparison.

The best solution for the comparison / filter would be a function that takes a string and a set of tags and checks whether any of the tags, in the given order or reversed, are contained in the string. That would make it easy to apply for linear searches and my XLMS searches.
I guess this is a good chance to start learning to use SeqAn.

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+  Tagger::Tagger(size_t min_tag_length, float ppm, size_t max_tag_length, size_t min_charge, size_t max_charge, StringList fixedModNames, StringList varModNames)
+  {
+    ppm_ = ppm;
+    min_tag_length_ = min_tag_length;
+    max_tag_length_ = max_tag_length;
+    min_charge_ = min_charge;
+    max_charge_ = max_charge;
+
+    const std::set<const Residue*> aas = ResidueDB::getInstance()->getResidues();
```
timosachsenberg: probably better to choose a residue set here (check the parameters of getResidues) that is not ambigious

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+      // skip non-standard residues, could be made into a parameter (TODO?)
+      // "I" is skipped here and handled explicitly as an alternative to "L" in getTag_
+      if (letter == "B" || letter == "X" || letter == "Z" || letter == "O" || letter == "J" || letter == "I")
+      {
+        continue;
+      }
+      const float mass = r->getMonoWeight(Residue::Internal);
+      mass2aa[mass] = letter;
+      std::cout << "TEST TAGGER added residue: " << letter << " | mass: " << mass << std::endl;
```
timosachsenberg: move to debug output

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+     const float delta = Math::ppmToMass(ppm_, m);
+     auto left = mass2aa.lower_bound(m - delta);
+     //if (left == mass2aa.end()) return " "; // cannot happen, since we checked boundaries above
+
+     // Check for multiple charges, prioritize lower charges (returns with the lowest charge with a match).
+     // This assumes both fragments had the same charge,
+     // but ignores whether ions in the longer sequence all have the same charge.
+     // This could lead to inconsistencies where an ion is considered to have different charges when compared to its left and right neighbors in the sequence (TODO?).
+     // This looks at the difference between two charged ions, so the proton mass is the same on both ions and can be ignored here.
+     for (size_t charge = min_charge_; charge <= max_charge_; ++charge)
```
timosachsenberg: I don't get this part. Don't we already check for a m/charge mass difference above?

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,80 @@
...
+  /**
+    *  @brief Calculate sequence tags from m/z values
+    *
+    *  Current restrictions and potential extensions:
+    *  - all peaks are considered charge 1
+    *  - first prefix/suffix ion don't give rise to first character in tags (as currently ion types are not specified)
+    *  - modifications are not supported
+    *  - gaps are not supported
+    **/
+  class OPENMS_DLLAPI Tagger
```
hroest: please add pyOpenMS bindings for this

src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -405,6 +415,20 @@ using namespace OpenMS;
...
+        for (std::string tag : tags)
```
hroest:
```suggestion
        for (const auto & tag : tags)
```

hroest:
please add tests and pyOpenMS bindings for this

timosachsenberg:
Does it also include termini?

timosachsenberg:
I think no SeqAn is required if you just check for the presence of a fixed length tag (e.g. minimal tag length). Linear scan over each sequence will do.

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,80 @@
...
+  {
+    public:
+      // initalize tagger with minimum/maximum tag length and +- tolerance ppm
+      Tagger(size_t min_tag_length, float ppm, size_t max_tag_length = 65535, size_t min_charge = 1, size_t max_charge = 1, StringList fixedModNames = StringList(), StringList varModNames = StringList());
+
+      // generate tags from mass vector @mzs using the standard residues in ResidueDB
```
cbielow: `@p mzs` ?

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,80 @@
...
+      void getTag(const std::vector<float>& mzs, std::set<std::string>& tags) const;
```
cbielow: for performance reasons: why use a set and not a vector as return type? 
and why not return directly? This makes usage much clearer.

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1367,4 +1367,41 @@ namespace OpenMS
...
+      OPXLDataStructs::ProteinProteinCrossLink candidate = candidates[i];
+      String alpha = candidate.alpha->toUnmodifiedString();
+      String beta;
+      if (candidate.beta)
+      {
+        beta = candidate.beta->toUnmodifiedString();
+      }
+      for (std::string tag : tags)
+      {
+        if (candidate_added) continue;
```
cbielow: I dont get this: why not break the loop whenever you found a hit?

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1367,4 +1367,41 @@ namespace OpenMS
...
+
+        if (alpha.hasSubstring(tag) || beta.hasSubstring(tag))
+        {
+          filtered_candidates.push_back(candidate);
+          candidate_added = true;
+          continue;
+        }
+        std::reverse(tag.begin(), tag.end());
+        if (alpha.hasSubstring(tag) || beta.hasSubstring(tag))
```
cbielow: would it make sense to reverse alpha and beta once and check for unchanged tags? (depends on length of proteins and number of tags I guess)...
This would allow to avoid the copy of the tag...

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1367,4 +1367,41 @@ namespace OpenMS
     }
   }
 
+  void OPXLHelper::filterCandidatesByTags(std::vector <OPXLDataStructs::ProteinProteinCrossLink>& candidates, std::set<std::string>& tags)
+  {
+    std::vector <OPXLDataStructs::ProteinProteinCrossLink> filtered_candidates;
+    std::vector<bool> selection(candidates.size(), false);
```
cbielow: not used? remove?

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1367,4 +1367,41 @@ namespace OpenMS
     }
   }
 
+  void OPXLHelper::filterCandidatesByTags(std::vector <OPXLDataStructs::ProteinProteinCrossLink>& candidates, std::set<std::string>& tags)
```
cbielow: `tags` can be const here?

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1367,4 +1367,41 @@ namespace OpenMS
...
+  {
+    std::vector <OPXLDataStructs::ProteinProteinCrossLink> filtered_candidates;
+    std::vector<bool> selection(candidates.size(), false);
+    
+    // brute force string comparisons for now
+    for (Size i = 0; i < candidates.size(); ++i)
```
cbielow: range based for? `i` is not being used...

src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -174,6 +178,8 @@ using namespace OpenMS;
 
     number_top_hits_ = static_cast<Int>(param_.getValue("algorithm:number_top_hits"));
     deisotope_mode_ = static_cast<String>(param_.getValue("algorithm:deisotope"));
+    use_sequence_tags_ = (param_.getValue("algorithm:use_sequence_tags") == "true" ? true : false);
```
cbielow: `use_sequence_tags_ = (param_.getValue("algorithm:use_sequence_tags") == "true")` is enough

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,80 @@
...
+      void setMaxCharge(size_t max_charge);
+    private:
+      float min_gap_; // will be set to smallest residue mass in ResidueDB
+      float max_gap_; // will be set to highest residue mass in ResidueDB
+      float ppm_; // < tolerance
+      size_t min_tag_length_; // < minimum tag length
+      size_t max_tag_length_; // < maximum tag length
+      size_t min_charge_; // < minimal fragment charge
+      size_t max_charge_; // < maximal fragment charge
+      std::map<float, std::string> mass2aa;
```
cbielow: `mass2aa_`

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <OpenMS/CHEMISTRY/Residue.h>
+#include <OpenMS/CHEMISTRY/ModificationsDB.h>
+#include <OpenMS/MATH/MISC/MathFunctions.h>
+
+namespace OpenMS
+{
+  std::string Tagger::getAAByMass_(float m) const
+  {
+     // fast check for border cases
+     if (m < min_gap_ || m > max_gap_) return " ";
```
cbielow: why not empty?

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+      if (gap > max_gap_) { return; } // already too far away - continue with next parent
+      const std::string aa = getAAByMass_(gap);
+      if (aa == " ") { ++j; continue; } // can't extend tag
+      tag.push_back(aa);
+      getTag_(tag, mzs, j, tags);
+      // each tag is build as a StringList, so that push_back(), pop_back() and size() work as intended even with modified residues, e.g. "M(Oxidation)"
+      // and then the tag is reported as a string by collapsing it here using concatenate()
+      if (tag.size() >= min_tag_length_) tags.insert(ListUtils::concatenate(tag));
+
+      // if aa is "L", then also add "I" as an alternative residue and extend the tag again
```
cbielow: why not use "J" in tags and convert all occurences of I/L in proteins to J before comparison?

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+        tag.push_back("I");
+        getTag_(tag, mzs, j, tags);
+      }
+      if (tag.size() >= min_tag_length_) tags.insert(ListUtils::concatenate(tag));
+      tag.pop_back();  // remove last string
+      ++j;
+    }
+  }
+
+  Tagger::Tagger(size_t min_tag_length, float ppm, size_t max_tag_length, size_t min_charge, size_t max_charge, StringList fixedModNames, StringList varModNames)
```
cbielow: use intilializer list

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+    max_tag_length_ = max_tag_length;
+    min_charge_ = min_charge;
+    max_charge_ = max_charge;
+
+    const std::set<const Residue*> aas = ResidueDB::getInstance()->getResidues();
+
+    for (const auto& r : aas)
+    {
+      std::string letter;
+      letter += r->getOneLetterCode()[0];
```
cbielow: make one line 
`string letter = r-> ...`

src/openms/source/CHEMISTRY/Tagger.cpp
cbielow: `fixed_modnames` ....

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+    for (String mod : fixedModNames)
+    {
+      const ResidueModification* rm = ModificationsDB::getInstance()->getModification(mod);
+      std::cout << "TEST TAGGER rm.getOrigin(): " << rm->getOrigin() << " | rm.getName(): " << rm->getName() << " | rm.getId(): " << rm->getId() << std::endl;
+      Residue r = *(ResidueDB::getInstance()->getResidue(rm->getOrigin()));
+      std::cout << "TEST TAGGER unmodified residue mass: " << r.getMonoWeight(Residue::Internal) << std::endl;
+      r.setModification(rm->getId());
+      std::cout << "TEST TAGGER modified residue mass: " << r.getMonoWeight(Residue::Internal) << std::endl;
+
+      for (std::map<float, std::string>::iterator it = mass2aa.begin(); it != mass2aa.end(); ++it)
```
cbielow: this is an inverse map.. and ineffective. But does not hurt I guess, because the number of items is small?

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+      {
+        continue;
+      }
+      const float mass = r->getMonoWeight(Residue::Internal);
+      mass2aa[mass] = letter;
+      std::cout << "TEST TAGGER added residue: " << letter << " | mass: " << mass << std::endl;
+    }
+
+    // for fixed modifications, replace the unmodified residue with the modified one
+    for (String mod : fixedModNames)
```
cbielow: `const auto& mod : ...`?

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,201 @@
...
+        }
+      }
+      const std::string name = rm->getOrigin() + std::string("(") + rm->getId() + std::string(")");
+      const float mass = r.getMonoWeight(Residue::Internal);
+      std::cout << "TEST TAGGER added residue: " << name << " | mass: " << mass << std::endl;
+      mass2aa[mass] = name;
+    }
+
+    // for variable modifications, add the modified residue to the list
+    for (String mod : varModNames)
```
cbielow: `const auto& mod : ...` ?

src/openms/source/ANALYSIS/XLMS/OPXLHelper.cpp
```diff
@@ -1367,4 +1367,41 @@ namespace OpenMS
...
+    {
+      bool candidate_added(false);
+      OPXLDataStructs::ProteinProteinCrossLink candidate = candidates[i];
+      String alpha = candidate.alpha->toUnmodifiedString();
+      String beta;
+      if (candidate.beta)
+      {
+        beta = candidate.beta->toUnmodifiedString();
+      }
+      for (std::string tag : tags)
```
cbielow: you are just searching unmodified tags against unmodified proteins/peptides, right?
Could be faster to build a search-trie for the tags and push the PPs across them and stop at the first hit -- should be very fast. Building the trie is also very fast.

enetz:
@timosachsenberg @cbielow 
I added benchmark tests for the generation of tags and for filtering of candidates based on tags and tried a few different versions of both algorithms.

- For the generation of tags, using chars instead of strings reduced runtime in one test from 80s to 50s, so it is faster, but not more than a factor of 2.
- I looked up how the Aho-Corasick algorothm is used in PeptideIndexing and tried to use it for candidate filtering, but this only had a small impact: in one worst case test of many candidates and a few tags without any overlaps, runtime was reduced from 87s to 77s. In a real world example, by running my XLMS tool on a dataset that runs for 50min it did not make any difference.
@cbielow the AC Trie filter is in `OPXLHelper::filterCandidatesByTagTrie`, do you think there is something I can do to make it more efficient for this specific application?
- changing the set of tags to a vector of tags also did not change the runtime

I will run some more accurate performance estimation next to see which parts of the algorithms are the slowest. I have a hunch that `AASequence::toUnmodifiedSequence` or another part of the filtering algorithm is the bottleneck rather than the string matching, so whether it uses Tries or `hasSubstring` does not make a big difference. Similarly for the tag generation.

cbielow:
code looks good. it can probably be optimized (e.g. avoid the reverse test by adding reverse tags to the trie -- should halve the search time) but as you say, first find out what the actual bottleneck is before optmizing the wrong part of the code.

timosachsenberg:
please fix the tests http://cdash.openms.de/index.php?project=OpenMS&date=2019-11-19

timosachsenberg:
can you comment on the unresolved points in this discussion and mark them as resolved if addressed

enetz:
@timosachsenberg this PR is not ready for merging yet.
I got the tagger to work as expected for my use case, but several legitimate issues were raised that need to be handled.

PR #4265, opened 2019-09-24
[FIX] Add map_index for Unassigned PeptideIdentifications for linking

jpfeuffer:
and fix indices for assigned PepIDs when linking consensusMaps.

### DIFF
--- a/src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
+++ b/src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
@@ -81,7 +81,7 @@ namespace OpenMS
     // accumulate file descriptions from the input maps:
     // cout << "Updating file descriptions..." << endl;
     out.getColumnHeaders().clear();
-    // mapping: (map index, original id) -> new id
+    // mapping: (input file index / map index assigned by the linkers, old map index) -> new map index
     map<pair<Size, UInt64>, Size> mapid_table;
     for (Size i = 0; i < maps.size(); ++i)
     {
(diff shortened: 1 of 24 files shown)


jpfeuffer:
might need some more work.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
```diff
@@ -129,11 +129,32 @@ namespace OpenMS
         }
       }
       *cons_it = adjusted;
+
+      for (auto& id : cons_it->getPeptideIdentifications())
+      {
+        Size file_index = id.getMetaValue("map_index");
```
timosachsenberg: can we be sure we have a map_index here?

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmQT.cpp
```diff
@@ -83,18 +86,25 @@ namespace OpenMS
         map_it->getProteinIdentifications().begin(),
         map_it->getProteinIdentifications().end());
 
-      // add unassigned peptide identifications to result map:
-      out.getUnassignedPeptideIdentifications().insert(
-        out.getUnassignedPeptideIdentifications().end(),
-        map_it->getUnassignedPeptideIdentifications().begin(),
-        map_it->getUnassignedPeptideIdentifications().end());
+      for (const PeptideIdentification& pepID : map_it->getUnassignedPeptideIdentifications())
```
timosachsenberg: is there a base class we could put this in?

src/topp/FeatureLinkerBase.cpp
```diff
@@ -292,12 +295,21 @@ class TOPPFeatureLinkerBase :
         StringList ms_runs;
         maps[i].getPrimaryMSRunPath(ms_runs);
         ms_run_locations.insert(ms_run_locations.end(), ms_runs.begin(), ms_runs.end());
+        if (keep_subelements)
+        {
+          std::function<void (PeptideIdentification &)> saveOldMapIndex = [](PeptideIdentification& p)
```
timosachsenberg: for these cases I prefer auto

timosachsenberg:
why do we need to keep the old_index?

src/topp/FeatureLinkerBase.cpp
jpfeuffer: It somehow doesn't work with the std::function wrapper.

jpfeuffer:
If you link two consensus with multiple submaps and want to transfer subelements, you need to know where they came from.
Especially with IDs because they are just dangling around.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
jpfeuffer: That is what I am trying to enforce. ConsensusXMLs without map_index in the IDs do not make sense.

jpfeuffer: I will add a warning but I dont know how to proceed with unannotated IDs.

timosachsenberg:
lools good!

PR #4269, opened 2019-09-27
[FIX] MAD set primary output optional

oliveralka:
Minor fix. 

A lot of metabolomics workflows use the featureXML output of the MetaboliteAdductDecharger, but an output folder has to be specified every time since the consensusXML output was mandatory. 
-> set all outputs to optional.


cbielow:
+1 
aapleclang failed due to missing boost contrib.

Issue #4229, opened 2019-08-29
building pyOpenMS with python3 fails


mwalzer:
Builds fail if I want to switch to py3. Are there any additional configuration steps necessary or is this a new issue to 2.4.0?
Consider some dockerfile like:
```
FROM openms/library

ENV DEBIAN_FRONTEND noninteractive

WORKDIR /openms-build

RUN apt-get update -y 
RUN apt-get install -y python3 python3-pip \
				python3-dev \
				python3-numpy
RUN pip3 install nose
RUN pip3 install -U setuptools
RUN pip3 install Cython
RUN pip3 install autowrap
RUN pip3 install wheel

RUN cmake  -DOPENMS_CONTRIB_LIBS="/contrib-build" -DCMAKE_PREFIX_PATH="/usr/;/usr/local" -DBOOST_USE_STATIC=OFF -DHAS_XSERVER=Off -DWITH_GUI=OFF -DBUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF -DPYOPENMS=On -DPYTHON_EXECUTABLE:FILEPATH=/usr/bin/python3 /OpenMS

RUN make pyopenms
```
Is there maybe some incompatibility with the cmake instructions? Maybe @hroest knows?
The error (actually, myriads like it) is 
```
created pyopenms.cpp

Error compiling Cython file:
------------------------------------------------------------
...
    
    def _hasElement_0(self,  name ):
        """Cython signature: bool hasElement(const String & name)"""
        assert (isinstance(name, str) or isinstance(name, unicode) or isinstance(name, bytes) or isinstance(name, String)), 'arg name wrong type'
    
        cdef bool _r = self.inst.get().hasElement(deref((convString(name)).get()))
                                                 ^
------------------------------------------------------------

pyopenms/pyopenms_8.pyx:9297:50: Invalid operand type for '*' (Python object)

Error compiling Cython file:
------------------------------------------------------------
...
    def hasElement(self, *args ):
        """
          - Cython signature: bool hasElement(const String & name)
          - Cython signature: bool hasElement(unsigned int atomic_number)
"""
        if (len(args)==1) and ((isinstance(args[0], str) or isinstance(args[0], unicode) or isinstance(args[0], bytes) or isinstance(args[0], String))):
                                                                                                                                             ^
------------------------------------------------------------

pyopenms/pyopenms_8.pyx:9314:142: 'String' is not a constant, variable or function identifier
```

hroest:
[...]

PR #4273, opened 2019-10-01
[Fix] Epifany and gcc4.8

### LINKED ISSUE #4267: Epifany Error
Trying to replace Fido with Epifany from a nighly

Version: 2.4.0-HEAD-HASH-NOTFOUND-HEAD-HASH-NOTFOUND Sep  9 2019, 13:03:04, Revision: a5903f78752457d371d499ce6b9342ff83ef88cd

When I ran the following command I get an error:

`Epifany -in work/comet/proteinid/e_idf_1.idXML -out work/comet/proteinid/epifany.idXML -threads 6 -algorithm:min_psms_extreme_probability 0 -debug 5`

```
Debug level: 5
 >> Epifany -in work/comet/proteinid/e_idf_1.idXML -out work/comet/proteinid/epifany.idXML -threads 6 -algorithm:min_psms_extreme_probability 0 -debug 5
Value of string option 'test': 0
Debug level (after ini file): 5
Value of string option 'no_progress': 0
/OpenMS/src/openms/source/APPLICATIONS/TOPPBase.cpp(1590):  - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
2019-09-25 19:57:54 Epifany:1: Parameters passed to Epifany
"psm_probability_cutoff" -> "0.001" (Remove PSMs with probabilities less than or equal this cutoff)
"min_psms_extreme_probability" -> "0.0" (Set PSMs with probability lower than this to this minimum probability.)
"max_psms_extreme_probability" -> "1.0" (Set PSMs with probability higher than this to this maximum probability.)
"top_PSMs" -> "1" (Consider only top X PSMs per spectrum. 0 considers all.)
"update_PSM_probabilities" -> "true" ((Experimental:) Update PSM probabilities with their posteriors under consideration of the protein probabilities.)
"user_defined_priors" -> "false" ((Experimental:) Uses the current protein scores as user-defined priors.)
"annotate_group_probabilities" -> "true" (Annotates group probabilities for indistinguishable protein groups (indistinguishable by experimentally observed PSMs).)
"use_ids_outside_features" -> "false" ((Only consensusXML) Also use IDs without associated features for inference?)
"model_parameters|prot_prior" -> "-1.0" (Protein prior probability ('gamma' parameter). Negative values enable grid search for this param.)
"model_parameters|pep_emission" -> "-1.0" (Peptide emission probability ('alpha' parameter). Negative values enable grid search for this param.)
"model_parameters|pep_spurious_emission" -> "-1.0" (Spurious peptide identification probability ('beta' parameter). Usually much smaller than emission from proteins. Negative values enable grid search for this param.)
"model_parameters|pep_prior" -> "0.1" (Peptide prior probability (experimental, should be covered by combinations of the other params).)
"model_parameters|regularize" -> "false" (Regularize the number of proteins that produce a peptide together (experimental, should be activated when using higher p-norms).)
"model_parameters|extended_model" -> "false" (Uses information from different peptidoforms also across runs (automatically activated if an experimental design is given!))
"loopy_belief_propagation|scheduling_type" -> "priority" ((Not used yet) How to pick the next message: priority = based on difference to last message (higher = more important). fifo = first in first out. random_spanning_tree = message passing follows a random spanning tree in each iteration)
"loopy_belief_propagation|convergence_threshold" -> "1.0e-05" (Initial threshold under which MSE difference a message is considered to be converged.)
"loopy_belief_propagation|dampening_lambda" -> "0.001" (Initial value for how strongly should messages be updated in each step. 0 = new message overwrites old completely (no dampening),1 = old message stays (no convergence, don't do that)In-between it will be a convex combination of both. Prevents oscillations but hinders convergence.)
"loopy_belief_propagation|max_nr_iterations" -> "2147483647" ((Unused, autodetermined) If not all messages converge, how many iterations should be done at max?)
"loopy_belief_propagation|p_norm_inference" -> "1.0" (P-norm used for marginalization of multidimensional factors. 1 == sum-product inference (all configurations vote equally) (default),<= 0 == infinity = max-product inference (only best configurations propagate)The higher the value the more important high probability configurations get.)
"param_optimize|aucweight" -> "0.2" (How important is AUC vs calibration of the posteriors? 0 = maximize calibration only, 1 = maximize AUC only, between = convex combination.)
"param_optimize|conservative_fdr" -> "true" (Use (D+1)/(T) instead of (D+1)/(T+D) for parameter estimation.)
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
Value of string option 'in': work/comet/proteinid/e_idf_1.idXML
Checking input file 'work/comet/proteinid/e_idf_1.idXML'
Internal error: Request for unregistered parameter 'min_psms_extreme_probability'
Error occurred in line 1212 of file /OpenMS/src/openms/source/APPLICATIONS/TOPPBase.cpp (in function: const OpenMS::ParameterInformation& OpenMS::TOPPBase::findEntry_(const OpenMS::String&) const) !

```

jpfeuffer:
Fixes #4267 and should make the code work with gcc 4.8 again.


hroest:
can you check the Epifany output here: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-15,os_label=win10_64/3406/testReport/junit/projectroot.src.tests/topp/UTILS_Epifany_1_out/

hroest:
also I dont think you quite solved the problem, there is still such a template at `src/openms/include/OpenMS/ANALYSIS/ID/IDScoreGetterSetter.h:566`

jpfeuffer:
I'm fixing the test right now. Oh, yes, you are right. It shouldn't work yet because of the second occurrence. How did you get it to work for your python builds?

Fixing both now..

hroest:
for Python builds, see over here: https://github.com/OpenMS/OpenMS/issues/4229 basically you can just run the following docker file detailed here https://github.com/OpenMS/OpenMS/wiki/pyOpenMS#test and I used my own branch on top of yours where I deleted the offending function :-) 

```
sudo docker build --build-arg branch=fix/gcc48 --build-arg repo=hroest/OpenMS .
```

jpfeuffer:
Only unrelated errors (upload, py memory leak)

CMakeLists.txt
```diff
@@ -47,13 +47,6 @@ SET(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS true)
 SET(CMAKE_CXX_STANDARD 11)
 SET(CMAKE_CXX_STANDARD_REQUIRED ON)
 
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
cbielow: you could leave that in, but change the version to `>=4.7` (which is what we require now, if I remember correctly)

We should also think of a Tag, such as `GCC-OPT 4.X` which we mention right here for documentation and whenever we increase this version, we can go and check for this tag in our codebase.
You could, for example, tag the current workaround for var.templates and this would remind us to revert back to the nicer var.templ. code in the future.

src/openms/source/METADATA/ProteinHit.cpp
```diff
@@ -90,7 +90,7 @@ namespace OpenMS
 
   //TODO WHAAAAAT why float???
```
cbielow: remove comment?

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreGetterSetter.h
```diff
@@ -240,14 +240,76 @@ namespace OpenMS
      * @brief Helper for getting scores in ConsensusMaps
      * @todo allow FeatureMap?
      */
-    template<class ...Args>
     static void getPeptideScoresFromMap_(
         ScoreToTgtDecLabelPairs &scores_labels,
-        const ConsensusMap &cmap, bool include_unassigned_peptides, Args &&... args)
+        const ConsensusMap &cmap, bool include_unassigned_peptides)
```
cbielow: tag here?
`// GCC-OPT 4.8 -- can be replaced by a variadic template, see #4273`

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreGetterSetter.h
```diff
@@ -488,20 +550,64 @@ namespace OpenMS
...
     static void setPeptideScoresForMap_(const std::map<double, double> &scores_to_FDR,
                                  ConsensusMap &cmap,
                                  bool include_unassigned_peptides,
                                  const std::string &score_type,
                                  bool higher_better,
-                                 bool keep_decoy,
-                                 Args &&... args)
+                                 bool keep_decoy)
+    {
+      //Note: Gcc4.8 cannot handle variadic templates in lambdas
```
cbielow: tag here?

PR #4276, opened 2019-10-01
Feature: Move constructors

hroest:
add more move constructors and assignment operators

### DIFF
--- a/src/openms/include/OpenMS/CONCEPT/UniqueIdInterface.h
+++ b/src/openms/include/OpenMS/CONCEPT/UniqueIdInterface.h
@@ -40,14 +40,14 @@
 namespace OpenMS
 {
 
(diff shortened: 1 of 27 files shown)


jpfeuffer:
This might have broken tests on Windows. There seem to be problems checking the noexcept now:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=334920
Sorry I had the windows machines busy while debugging something else (that is why it didn't build on Windows here in this PR)

hroest:
> This might have broken tests on Windows. There seem to be problems checking the noexcept now:
> http://cdash.openms.de/viewTest.php?onlyfailed&buildid=334920
> Sorry I had the windows machines busy while debugging something else (that is why it didn't build on Windows here in this PR)

this is probably 908757074e0d6e621bdc062a6315649faed5c0b1 -- can you test without this?

PR #4278, opened 2019-10-01
[FIX] write all chromatograms to disk

hroest:
- fix bug where only chromatograms were written to disk within the same
  SWATH window, preventing higher isotopes to be written out


timosachsenberg:
rebuild jenkins

timosachsenberg:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=334639

timosachsenberg:
errors on windows related to noexcept

PR #4280, opened 2019-10-02
Improvements to SqliteConnector

hendrikweisser:
Three main things:
1. Changed name of function `executePreparedStatement` to `prepareStatement`, because (in contrast to the other `execute...` functions) it doesn't actually run the statement against the database. I think that's quite an important distinction that should be reflected in the name.
2. Added a constructor from `unsigned char*` to String to simplify reading text results from SQLite.
3. Simplified the implementation of the `tableExists` function.

### DIFF
--- a/src/openms/source/DATASTRUCTURES/String.cpp
+++ b/src/openms/source/DATASTRUCTURES/String.cpp
@@ -60,6 +60,11 @@ namespace OpenMS
   {
   }
 
+  String::String(const unsigned char* s) :
+    string(reinterpret_cast<const char*>(s))
+  {
+  }
+
   String::String(const QString& s) :
     string(s.toStdString())
   {
--- a/src/openms/source/FORMAT/SqliteConnector.cpp
+++ b/src/openms/source/FORMAT/SqliteConnector.cpp
@@ -60,15 +60,19 @@ namespace OpenMS
   {
     bool found = false;
 
-    sqlite3_stmt * xcntstmt;
-    SqliteConnector::executePreparedStatement(db, &xcntstmt, "PRAGMA table_info(" + tablename + ")");
+    sqlite3_stmt* xcntstmt;
+    prepareStatement(db, &xcntstmt, "PRAGMA table_info(" + tablename + ")");
 
     // Go through all columns and check whether the required column exists
     sqlite3_step(xcntstmt);
-    while (sqlite3_column_type( xcntstmt, 0 ) != SQLITE_NULL)
+    while (sqlite3_column_type(xcntstmt, 0) != SQLITE_NULL)
     {
-      String name = String(reinterpret_cast<const char*>(sqlite3_column_text( xcntstmt, 1 )));
-      if (colname == name) {found = true;}
+      String name = String(reinterpret_cast<const char*>(sqlite3_column_text(xcntstmt, 1)));
+      if (colname == name)
+      {
+        found = true;
+        break;
+      }
       sqlite3_step(xcntstmt);
     }
     sqlite3_finalize(xcntstmt);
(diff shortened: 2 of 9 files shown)


src/openms/source/FORMAT/SqliteConnector.cpp
```diff
@@ -60,15 +60,19 @@ namespace OpenMS
...
+    prepareStatement(db, &xcntstmt, "PRAGMA table_info(" + tablename + ")");
 
     // Go through all columns and check whether the required column exists
     sqlite3_step(xcntstmt);
-    while (sqlite3_column_type( xcntstmt, 0 ) != SQLITE_NULL)
+    while (sqlite3_column_type(xcntstmt, 0) != SQLITE_NULL)
     {
-      String name = String(reinterpret_cast<const char*>(sqlite3_column_text( xcntstmt, 1 )));
-      if (colname == name) {found = true;}
+      String name = String(reinterpret_cast<const char*>(sqlite3_column_text(xcntstmt, 1)));
```
timosachsenberg: reinterpret_cast required here?

src/openms/source/DATASTRUCTURES/String.cpp
```diff
@@ -60,6 +60,11 @@ namespace OpenMS
   {
   }
 
+  String::String(const unsigned char* s) :
+    string(reinterpret_cast<const char*>(s))
```
timosachsenberg: This sounds dangerous. At least the explicit keyword could save us from trouble.

hendrikweisser: I'm happy to add the `explicit`, but none of the other String constructors is declared with it. Do you want it just for this one?

src/openms/source/FORMAT/SqliteConnector.cpp
hendrikweisser: Shouldn't be - well spotted! I'm removing that and another one in the file.

hendrikweisser:
Updated. There's no class test for SqliteConnector, by the way.

src/openms/source/DATASTRUCTURES/String.cpp
timosachsenberg: I would love to have it for every single argument constructor - but requesting that sounds like asking for a lot ;)

hendrikweisser: So I tried that - making every single-argument c'tor `explicit` - for String. First, for other string types it actually makes sense to allow implicit conversion - otherwise you need lots of overloads for the assignment operator, or you can't even assign e.g. a string literal. For numeric types, it would probably make sense to have `explicit` c'tors, but it leads to many test errors, so making the change now is impractical.

I've declared the new c'tor as `explicit`, though.

timosachsenberg:
rebuild jenkins

hroest:
Making these explicit would really go overboard, see the underlying
std::string ctors that are not explicit either
http://www.cplusplus.com/reference/string/string/string/

hroest:
I think for the numbers we should make it explicit but not for string/char
to String conversions

src/openms/source/DATASTRUCTURES/String.cpp
hroest: this is definitely not something we should do routinely, in my mind `unsigned char` is not generally a string but raw binary data. Only if you are really sure that this is a string should you do something like this and use `reinterpret_cast`.  Therefore doing this by default does also seem dangerous to me, even with `explicit` in place...

However, I also learned that `char` is not specified to be signed or unsigned, but to me it seems clear that the intention of `char` and `signed char` would be to process ASCII data while `unsigned char` is raw data of unknown interpretation (a bit like Python string and byte arrays).

hendrikweisser: > in my mind unsigned char is not generally a string but raw binary data

I don't think that's generally true. E. g. the SQLite C API returns `unsigned char*` for a text column, but `void*` for a BLOB, i.e. raw binary data.

Pragmatically, I think it's clearer to use the (explicit) String constructor than to litter the code with `reinterpret_cast` calls.

hroest: > I don't think that's generally true. E. g. the SQLite C API returns `unsigned char*` for a text column, but `void*` for a BLOB, i.e. raw binary data.

the reason here being that this allows UTF8 support: https://stackoverflow.com/questions/319790/sqlite-c-c-api-unsigned-char (where UTF8 *is* a raw binary data that needs interpretation). Basically the conceptual problem here is that you get raw binary data and it makes sense to *explicitly* say how you want to interpret that raw data, in this case it is a choice to interpret that data as ASCII

hendrikweisser: Fair point, but in my opinion an *explicit* String constructor also says that.

cbielow:
how about a function which converts from unsigned char to string and does the magic inside.
This way, we do not 'litter' the string interface yet encapsulate the functionality.

hendrikweisser:
> how about a function which converts from unsigned char to string and does the magic inside.

You mean like a converting constructor? ;-)

(In all seriousness, given all the converting c'tors that String already has, it would be inconsistent to add a non-c'tor function for this purpose to the String interface. Someone could add a static function to SqliteConnector, but I don't actually think that's the better place for it.)

timosachsenberg:
"Someone could add a static function to SqliteConnector, but I don't actually think that's the better place for it.)"

Hmm, why not to SqliteConnector? The encoding is pretty specific to SQLite. 
The constructor would also be fine with me but then I guess one should add a comment to every line to indicate that something not obvious (and with potential side effects) is happening.

hendrikweisser:
I don't want to spend forever discussing a minor issue like this. Someone (Hannes?) just tell me whether to go ahead with or without the String changes, so we can merge.

hendrikweisser:
> one should add a comment to every line to indicate that something not obvious (and with potential side effects) is happening

I think the fundamental disagreement is about whether this is actually the case or not. In my opinion it's an implementation detail that should be hidden, while Hannes thinks there are caveats that should be made obvious.

PR #4283, opened 2019-10-02
[FEATURE] Added notarization when building packages on macOS

jpfeuffer:


jpfeuffer:
I want to still remove the files and maybe use the signing environment variable in another way.

timosachsenberg:
failing tests related to noexcept

Issue #4268, opened 2019-09-26
empty FEATURE_TRANSITION in OSW


hroest:
When using the `-enable_uis_scoring` parameter, the `FEATURE_TRANSITION` table is not populated when running OpenSWATH in non-IPF mode. This is because the corresponding SQL string never gets written out:

https://github.com/OpenMS/OpenMS/blob/61219b2c48abf09d480950fda86d06c4fb30f3ba/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.cpp#L440-L447

hroest:
@grosenberger maybe we should change that line to 
```
if (enable_uis_scoring_ && !sql_feature_uis_transition.empty())
```

PR #4286, opened 2019-10-04
Update OpenSwathOSWWriter.cpp

hroest:
- Fixes https://github.com/OpenMS/OpenMS/issues/4268
- Removes a side effect of UIS scoring (no transition level output in OSW when enabled)

@grosenberger we should test this and investigate what other side effects UIS scoring has


grosenberger:
Thank you, @hroest, that looks like a good solution! 👍

timosachsenberg:
rebuild jenkins

PR #4296, opened 2019-10-11
Add citation for FeatureFinderIdentification to documentation

hendrikweisser:


jpfeuffer:
You can additionally do it in the TOPPTool constructor. Then it will also show up on the command line and in the tool description (i.e. KNIME).

hendrikweisser:
> You can additionally do it in the TOPPTool constructor.

Are there predecents/guidelines for that? So far I've seen and added references only in the main documentation.

hroest:
> > You can additionally do it in the TOPPTool constructor.
> 
> Are there predecents/guidelines for that? So far I've seen and added references only in the main documentation.

https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/a07762.html#ab8287673d1342334ccb7519a5c04a7d7

cbielow:
I was about to send a link to the docs, but doing it seemed almost as fast --> see #4304

PR #4298, opened 2019-10-11
Minor changes to allow compiling with Visual Studio 2019

nateglasser:
In VS2019 the standard libraries were refactored to not define std::string, which led to some compilation errors where string wasn't explicitly included. A comparison operator in two of the tests also needed to be defined const.


hroest:
:+1: if it compiles and thanks a lot @nateglasser

Issue #4073, opened 2019-04-23
FileFilter: add option to remove all meta values


timosachsenberg:
would be a useful option to delete all meta data in id and quant files (e.g., to trim down size)

poshul:
I've implemented this in a pull request. It should be noted that the size of files doesn't change substantially in my tests. I think this is still a worthwhile option for sanitizing data from proprietary workflows.

timosachsenberg:
@poshul in the linked PR you said you would give it another go. Any update there?

PR #4303, opened 2019-10-15
Remove all Metavalues

poshul:
Implements a remove all metavalue option in filefilter as requested in #4073. Also documents the use of remove_meta in mzml files.


src/topp/EICExtractor.cpp
```diff
@@ -2,7 +2,7 @@
 //                   OpenMS -- Open-Source Mass Spectrometry
 // --------------------------------------------------------------------------
 // Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
-// ETH Zurich, and Freie Universitaet Berlin 2002-2018.
+// ETH Zurich, and Freie Universitaet Berlin 2002-2017.
```
timosachsenberg: 2019 or leave as 2018 so we change it with a script

src/topp/FileFilter.cpp
```diff
@@ -408,7 +409,12 @@ class TOPPFileFilter :
...
-    registerStringList_("f_and_c:remove_meta", "<name> 'lt|eq|gt' <value>", StringList(), "Expects a 3-tuple (=3 entries in the list), i.e. <name> 'lt|eq|gt' <value>; the first is the name of meta value, followed by the comparison operator (equal, less or greater) and the value to compare to. All comparisons are done after converting the given value to the corresponding data value type of the meta value (for lists, this simply compares length, not content!)!", false);
+    registerStringList_("f_and_c:remove_meta", "<name> 'lt|eq|gt' <value>", StringList(), "WARNING: THIS FLAG IS DEPRECATED. USE all:remove_meta instead.   Expects a 3-tuple (=3 entries in the list), i.e. <name> 'lt|eq|gt' <value>; the first is the name of meta value, followed by the comparison operator (equal, less or greater) and the value to compare to. All comparisons are done after converting the given value to the corresponding data value type of the meta value (for lists, this simply compares length, not content!)!", false);
+
+    addEmptyLine_();
+    registerTOPPSubsection_("all","Options that can be applied to all file types");
```
timosachsenberg: I think we can put this into "all" or just move it out of subsections (like the other general stuff e.g. sort).

src/topp/EICExtractor.cpp
poshul: Set to 2018.

src/topp/FileFilter.cpp
poshul: I've added equivalent functionality to all:remove_meta. I left this in for backwards compatibility (it throws a warning when used). If I am okay to remove it I can do so.

timosachsenberg: I meant not to use "all" at all ;)

poshul: better now?

src/topp/FileFilter.cpp
```diff
@@ -408,7 +412,7 @@ class TOPPFileFilter :
...
     registerStringOption_("f_and_c:size", "[min]:[max]", ":", "Size range to extract", false);
-    registerStringList_("f_and_c:remove_meta", "<name> 'lt|eq|gt' <value>", StringList(), "Expects a 3-tuple (=3 entries in the list), i.e. <name> 'lt|eq|gt' <value>; the first is the name of meta value, followed by the comparison operator (equal, less or greater) and the value to compare to. All comparisons are done after converting the given value to the corresponding data value type of the meta value (for lists, this simply compares length, not content!)!", false);
+    registerStringList_("f_and_c:remove_meta", "<name> 'lt|eq|gt' <value>", StringList(), "WARNING: THIS FLAG IS DEPRECATED. USE remove_meta instead.   Expects a 3-tuple (=3 entries in the list), i.e. <name> 'lt|eq|gt' <value>; the first is the name of meta value, followed by the comparison operator (equal, less or greater) and the value to compare to. All comparisons are done after converting the given value to the corresponding data value type of the meta value (for lists, this simply compares length, not content!)!", false);
```
cbielow: why move the flag out of `f_and_c`? It effectively only works for these data types

src/topp/FileFilter.cpp
```diff
@@ -306,6 +307,9 @@ class TOPPFileFilter :
     registerStringOption_("int", "[min]:[max]", ":", "Intensity range to extract", false);
 
     registerFlag_("sort", "Sorts the output according to RT and m/z.");
+    registerStringList_("remove_meta", "<name> 'lt|eq|gt' <value>", StringList(), "Expects a 3-tuple (=3 entries in the list), i.e. <name> 'lt|eq|gt' <value>; the first is the name of meta value, followed by the comparison operator (equal, less or greater) and the value to compare to. All comparisons are done after converting the given value to the corresponding data value type of the meta value (for lists, this simply compares length, not content!)!", false);
+    registerFlag_("remove_all_meta","Remove ALL metadata. This happens after filtering by remove_meta");
```
cbielow: misleading name IMHO.
This flag removed metadata ARRAYS from MSExperiments and has nothing to do with metadata as userparams.

src/topp/FileFilter.cpp
poshul: there is extant code applying it to MzML files as well. I can either move it out of f_and_c or remove the code.

src/topp/FileFilter.cpp
poshul: I'm not especially attached to the name. Do you have a suggestion that'd fit better?

hroest:
@poshul can you add a test for `FileFilter`?

poshul:
Added TOPP tests.

src/tests/topp/CMakeLists.txt
```diff
@@ -642,6 +642,17 @@ add_test("TOPP_FileFilter_49" ${TOPP_BIN_PATH}/FileFilter -test -in ${DATA_DIR_T
 add_test("TOPP_FileFilter_49_out" ${DIFF} -in1 FileFilter_49_1.tmp -in2 ${DATA_DIR_TOPP}/FileFilter_49_output.mzML)
 set_tests_properties("TOPP_FileFilter_49_out" PROPERTIES DEPENDS "TOPP_FileFilter_49")
 
+# Test Metadata removal
+add_test("TOPP_FileFilter_50" ${TOPP_BIN_PATH}/FileFilter -test -in ${DATA_DIR_TOPP}/FileFilter_49_input.mzML -peak_options:zlib_compression true -out FileFilter_50_1.tmp)
```
hroest:
```suggestion
add_test("TOPP_FileFilter_50" ${TOPP_BIN_PATH}/FileFilter -test -in ${DATA_DIR_TOPP}/FileFilter_49_input.mzML -remove_all_meta -out FileFilter_50_1.tmp)
```

otherwise, no meta data is removed?

src/tests/topp/FileFilter_50_output.mzML
```diff
@@ -0,0 +1,3867 @@
...
+							<cvParam cvRef="MS" accession="MS:1000827" name="isolation window target m/z" value="412.5" unitAccession="MS:1000040" unitName="m/z" unitCvRef="MS" />
+							<cvParam cvRef="MS" accession="MS:1000828" name="isolation window lower offset" value="12.5" unitAccession="MS:1000040" unitName="m/z" unitCvRef="MS" />
+							<cvParam cvRef="MS" accession="MS:1000829" name="isolation window upper offset" value="12.5" unitAccession="MS:1000040" unitName="m/z" unitCvRef="MS" />
+						</isolationWindow>
+						<activation>
+							<cvParam cvRef="MS" accession="MS:1000044" name="dissociation method" />
+						</activation>
+					</precursor>
+				</precursorList>
+				<binaryDataArrayList count="3">
```
hroest: this does not seem to do what it is supposed to do?

src/topp/FileFilter.cpp
hroest: Also it simply calls `clearMetaInfo` on the maps, which only removes meta info that is directly related to the Feature/ConsensusMap itself but not inside the Features. I dont think that is really what I expect (in the case of mzML it pertains to removing actual data that is `not` mz or intensity, in the case of Feature/Consensus Maps it pertains to removing meta data associated with the while file and not with features). I think this is overall quite confusing.

src/topp/FileFilter.cpp
hroest: I dont see how `remove_meta` currently applies to mzML files, did you add that code somewhere?

src/tests/topp/FileFilter_52_output.consensusXML
```diff
@@ -0,0 +1,13344 @@
...
+			<groupedElementList>
+				<element map="0" id="6722364803272170291" rt="224.964699999999994" mz="1577.545300249219963" it="35056.601563" charge="31"/>
+				<element map="0" id="10887204495569992262" rt="224.119200000000006" mz="1397.248683427450032" it="27952.300781" charge="35"/>
+				<element map="0" id="11369039682247725242" rt="224.964699999999994" mz="1746.748917190310067" it="24485.300781" charge="28"/>
+				<element map="0" id="11551657342509961930" rt="224.964699999999994" mz="1630.010463770880051" it="36577.5" charge="30"/>
+				<element map="0" id="13949842513044340834" rt="221.561399999999992" mz="1358.42698779363991" it="21285.900391" charge="36"/>
+				<element map="0" id="14853527260641856532" rt="224.119200000000006" mz="1686.197995845770038" it="31099.400391" charge="29"/>
+				<element map="0" id="15057571094454825526" rt="226.647600000000011" mz="1528.188969366020046" it="23426.099609" charge="32"/>
+			</groupedElementList>
+			<UserParam type="string" name="label" value="T944"/>
```
hroest: again, I dont think this is doing what https://github.com/OpenMS/OpenMS/issues/4073 intended: the meta data is still there!

hroest:
I think this needs quite a bit more work, it is currently unclear which metadata gets removed from which files and what "meta data" even is. For mzML files, the extra data arrays are clearly not meta data but can be actual important measurement data so removing them when using a tool that does meta data removal is surprising. On the other hand, the proposal does *not* remove actual metadata as suggested in https://github.com/OpenMS/OpenMS/issues/4073

poshul:
> I think this needs quite a bit more work, it is currently unclear which metadata gets removed from which files and what "meta data" even is. For mzML files, the extra data arrays are clearly not meta data but can be actual important measurement data so removing them when using a tool that does meta data removal is surprising. On the other hand, the proposal does _not_ remove actual metadata as suggested in #4073

Okay, I will take another look at this and bring it back when I've figured it out.

PR #4305, opened 2019-10-15
Sqlite connector

cbielow:
amendment to  #4280
which reverses the explicit ctor in `String` for `unsigned char*`.
Reasoning: its a bad place to put this functionality, since its only used in very few places (related to Sql) and std::string does not support this conversion either (and there is probably a very good reason for it). 

During refactoring, some other issues related to our Sql code popped up (e.g. potential mem-leaks), but thats for another PR (where we should make use of RAII to get rid of manual memory management for prepared statements, which currently require calling `sqlite3_finalize(stmt);` to free memory)

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
@@ -226,10 +226,8 @@ namespace OpenMS
       Sql::extractValue<std::string>(&mytransition.group_id, stmt, 6);
       Sql::extractValue<int>((int*)&mytransition.decoy, stmt, 7);
       Sql::extractValue<std::string>(&mytransition.PeptideSequence, stmt, 8);
-      if (sqlite3_column_type( stmt, 9 ) != SQLITE_NULL)
-      {
-        String(sqlite3_column_text(stmt, 9)).split(';', mytransition.ProteinName);
-      }
+      String tmp_field;
+      if (Sql::extractValue<std::string>(&tmp_field, stmt, 9)) tmp_field.split(';', mytransition.ProteinName);
       Sql::extractValue<std::string>(&mytransition.Annotation, stmt, 10);
       Sql::extractValue<std::string>(&mytransition.FullPeptideName, stmt, 11);
       Sql::extractValue<std::string>(&mytransition.CompoundName, stmt, 12);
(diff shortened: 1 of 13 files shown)


hroest:
> (and there is probably a very good reason for it).

I pointed this out before, the reason is that `unsigned char*` are likely bytes of an unknown encoding (UTF8, UTF16, ???) and so converting that to std::string would mean you make implicit assumptions about your data for the gain of syntactic sugar instead of making this explicit. This will work here for SQL but maybe not in other places. Maybe a compromise would be to not have a constructor but have a static function `OpenMS::String::fromASCIIBytestream` where you are explicit that you are expecting only ASCII in your byte stream (maybe?) -- or your solution of doing this only for Sqlite.

src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp
```diff
@@ -244,17 +242,11 @@ namespace OpenMS
...
-      }
+      if (Sql::extractValue<std::string>(&tmp_field, stmt, 24)) tmp_field.split(';', mytransition.uniprot_id);
       Sql::extractValue<int>((int*)&mytransition.detecting_transition, stmt, 25);
       Sql::extractValue<int>((int*)&mytransition.identifying_transition, stmt, 26);
       Sql::extractValue<int>((int*)&mytransition.quantifying_transition, stmt, 27);
-      if (sqlite3_column_type( stmt, 28 ) != SQLITE_NULL)
-      {
-        String(reinterpret_cast<const char*>(sqlite3_column_text( stmt, 28 ))).split('|', mytransition.peptidoforms);
-      }
+      if (Sql::extractValue<std::string>(&tmp_field, stmt, 28)) tmp_field.split(';', mytransition.peptidoforms);
```
hroest: copy-paste error
```suggestion
      if (Sql::extractValue<std::string>(&tmp_field, stmt, 28)) tmp_field.split('|', mytransition.peptidoforms);
```

cbielow: whops. nice catch!

cbielow:
rebuild jenkins

jpfeuffer:
strcomp not declared in this scope. on gcc7. probably related.

cbielow:
rebuild jenkins

src/openms/source/FORMAT/SqliteConnector.cpp
```diff
@@ -131,26 +128,28 @@ namespace OpenMS
...
-      } 
+      }
     }
 
     rc = sqlite3_step(stmt);
     if (rc != SQLITE_DONE)
     {
       std::cerr << "SQL error after sqlite3_step" << std::endl;
       std::cerr << "Prepared statement " << prepare_statement << std::endl;
+      // TODO this is a mem-leak (missing sqlite3_finalize())
```
timosachsenberg: Why not call it?

cbielow: because that only fixes the problem locally. We should think about a more thorough solution?!

timosachsenberg: Something like unique_ptr + lambda deleter?

cbielow: yes. baked into `prepareStatement()` or some other place appropriate...

cbielow:
MSVC has git issues, but since I've build it using VS15 im reasonably sure it works :)
will merge

src/openms/source/FORMAT/SqliteConnector.cpp
hroest: If you are here and throw an exception in the SQL code (e.g. invalid SQL), you probably have much bigger problems than a memory leak?

Issue #3505, opened 2018-05-30
Sort spectra in OnDisc experiment


timosachsenberg:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp#L310 and here https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp#L343

hroest:
also if we assume spectra are sorted, we should sort them in all other readers (mzData, mzXML etc) before use

PR #4306, opened 2019-10-15
Sort spectra

poshul:
Sort spectra for MzML. Addresses #3505


src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp
```diff
@@ -306,6 +306,7 @@ namespace OpenMS
 
     if (data.size() > 2) 
     {
+      spectrum.sortByPosition();
```
hroest: I dont understand how this can work, first you fill in the data in the correct order, then you sort the m/z data only, then you add the meta data? This seems all wrong.

timosachsenberg:
close for now

PR #4307, opened 2019-10-15
[FIX] TOPPView when running TOPP tool 

### LINKED ISSUE #4164: Windows: TOPPView fails to execute TOPP tool if username / tmp file path has space in it
Trying to execute TOPP tools directly in TOPPView fails on Windows if the path to the temp folder has a space in it. This mostly happens if the user name has a space since the new temp paths are located in "C:\Users\Firstname Lastname\Appdata\Local\Temp" or similar.

cbielow:
fixes bug when running a TOPP tool from TOPPView and the current temporary path contains spaces 
(which yields a message box saying that the TOPP tool could not be executed (when writing the default ini file)

fixes  #4164

### DIFF
--- a/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
+++ b/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
@@ -3000,7 +3000,6 @@ namespace OpenMS
       }
       else
       {
-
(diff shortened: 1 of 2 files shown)


cbielow:
rebuild jenkins

timosachsenberg:
Looks good to me. Feel free to merge

PR #4310, opened 2019-10-16
[FIX] MSstatsConverter MSstatTMT fractions

oliveralka:
For consistency reasons, always report fractions for MSstatsTMT.  

As far as I know, the Experimental Design has always a "Fraction" column, which should be mandatory.


oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

jpfeuffer:
There are no tests to change? We should add some at some point.

oliveralka:
> There are no tests to change? We should add some at some point.

Yes, let me have a look.

oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

timosachsenberg:
looks good

src/tests/topp/MSstatsConverter_3_design.tsv
```diff
@@ -0,0 +1,43 @@
...
+1	1	QExactiveHF02_03904.mzML	6	6
+1	1	QExactiveHF02_03904.mzML	7	7
+1	1	QExactiveHF02_03904.mzML	8	8
+1	1	QExactiveHF02_03904.mzML	9	9
+1	1	QExactiveHF02_03904.mzML	10	10
+2	1	QExactiveHF02_03905.mzML	1	11
+2	1	QExactiveHF02_03905.mzML	2	12
+2	1	QExactiveHF02_03905.mzML	3	13
+2	1	QExactiveHF02_03905.mzML	4	14
+2	1	QExactiveHF02_03905.mzML	5	15
```
hroest: it may be helpful for testing this to actually use multiple fractions?

oliveralka: Yes, a test with multiple fractions already exists (MSstatsConverter_2_design.tsv):
https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/CMakeLists.txt#L1171

The additional test in this PR covers only the special case if one fraction is used, to have a uniform output.

PR #4316, opened 2019-10-21
Remove redundant field from IdentificationData::ScoreType

hendrikweisser:
A score type (e.g. from a database search engine) is identified by a CV term, which includes a name field. For simplicity, just use that field even if there's no full CV term for a score type, only a name.

Also clean up some whitespace issues in "CVTerm.h".

### DIFF
--- a/src/tests/topp/NucleicAcidSearchEngine_11_out.idXML
+++ b/src/tests/topp/NucleicAcidSearchEngine_11_out.idXML
@@ -16,7 +16,7 @@
 				<UserParam type="string" name="label" value="UGAGGUAGUAGGUUGUAUAGU"/>
 				<UserParam type="float" name="OMS:precursor_mz_error_ppm" value="-1.851164946914477"/>
 				<UserParam type="int" name="isotope_offset" value="2"/>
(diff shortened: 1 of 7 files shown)


src/tests/topp/NucleicAcidSearchEngine_11_out.idXML
```diff
@@ -16,7 +16,7 @@
 				<UserParam type="string" name="label" value="UGAGGUAGUAGGUUGUAUAGU"/>
 				<UserParam type="float" name="OMS:precursor_mz_error_ppm" value="-1.851164946914477"/>
 				<UserParam type="int" name="isotope_offset" value="2"/>
-				<UserParam type="float" name="q-value" value="0.0"/>
+				<UserParam type="float" name="PSM-level q-value" value="0.0"/>
```
hroest: so why is this stored as user param and not as CV term (`"MS:1002354", "PSM-level q-value",`) ?

hendrikweisser: Because PeptideHit is derived from MetaInfoInterface, but has no facility for storing CV terms.

PR #4317, opened 2019-10-21
Fix issues with (missing) protein coverage in IdentificationData

hendrikweisser:
The (old) ProteinHit class uses "-1" to indicate missing protein sequence coverage. This was not considered during conversion to/from the (new) IdentificationData class, which uses coverage values between 0 and 1, where 0 is equivalent to missing. Fixed now.

Also clean up some whitespace issues in "ProteinHit.cpp".


timosachsenberg:
@jpfeuffer do you see any problem that missing coverage information is the same as a protein that is not covered by a peptide? e.g.: missing vs no coverage

hendrikweisser:
> do you see any problem that missing coverage information is the same as a protein that is not covered by a peptide? e.g.: missing vs no coverage

I'm not Julianus, but my reasoning was that if the protein really has zero coverage, it wasn't identified, so it shouldn't be in the results.

timosachsenberg:
Agree but in the past we had to call removeUnreferencedProteins to get rid of those explicitly. Not sure if there were cases where they made sense...

jpfeuffer:
I think every part of the code that needs only matched proteins will use the removeUnreferencedProteins function which actually looks at the protein-peptide associations, so I think this will not affect anything. We should try to avoid unreferenced proteins anyhow (especially at the end of a tool).
But how do you distinguish between "no coverage" and "coverage not calculated yet"? Can this happen in this structure?

hendrikweisser:
> But how do you distinguish between "no coverage" and "coverage not calculated yet"? Can this happen in this structure?

I'd say "0" means "coverage not calculated yet". When you calculate the coverage and it comes out as zero, the protein should be removed.

(To be honest, at this point I'm not absolutely sure that there won't be a need for a separate "missing" value. But I think it's easier to add when we're sure it's needed, than to remove one that's not needed.)

timosachsenberg:
I tend to agree with Hendrik. In the worst case we would need to add a dataprocessing item

PR #4320, opened 2019-10-22
IDFilter output

timosachsenberg:


jpfeuffer:
Thanks. Much better.

hendrikweisser:
I'm late to the party, but I would suggest using "peptide matches" (or "peptide-spectrum matches") instead of "spectrum matches". Otherwise the difference between "spectra identified" and "spectrum matches" is less clear, and there's no mention of anything about identified peptides.

Issue #4272, opened 2019-09-30
Failing Preconditions in OpenSwath debug mode


jpfeuffer:
I think after the recent changes to OpenSwath there are some preconditions failing in debug mode.
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=334085

@hroest Could you have a look at that?

PR #4323, opened 2019-10-22
Fix swath

cbielow:
some Swath code currently breaks the nightlies...
(not sure how this got merged in the first place)

this fixes the issue, but not the bug per se (the drift-time code seems very experimental to me....)
This cannot have possibly worked on any platform.. yet some compilers dont seem to realize...

### DIFF
--- a/src/tests/class_tests/openms/source/IonMobilityScoring_test.cpp
+++ b/src/tests/class_tests/openms/source/IonMobilityScoring_test.cpp
@@ -444,7 +444,7 @@ START_SECTION(([EXTRA]
   TEST_EQUAL(drift_spectrum->getMZArray()->data.size(), 24)
   TEST_EQUAL(drift_spectrum->getMZArray()->data.size(), drift_spectrum->getIntensityArray()->data.size())
   TEST_EQUAL(drift_spectrum->getMZArray()->data.size(), drift_spectrum->getDriftTimeArray()->data.size())
-
+  /*
   IonMobilityScoring::driftScoringMS1Contrast(drift_spectrum, drift_spectrum_ms1, transitions, scores,
                                    drift_lower, drift_upper,
                                    dia_extract_window_, dia_extraction_ppm_,
@@ -497,6 +497,7 @@ START_SECTION(([EXTRA]
   // deal with all-zero transitions
   drift_lower = 2.5;
   drift_upper = 3.5;
+  
   IonMobilityScoring::driftScoringMS1Contrast(drift_spectrum, drift_spectrum_ms1, transitions, scores,
                                    drift_lower, drift_upper,
                                    dia_extract_window_, dia_extraction_ppm_,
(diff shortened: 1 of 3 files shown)


oliveralka:
May be related: #4324

cbielow:
> 
> 
> May be related: #4324

No, #4324 is a compile issue. The nightlies break at runtime (but compile fine)

cbielow:
this is ready to merge (pyopenms memory leak tests fails)

src/tests/class_tests/openms/source/IonMobilityScoring_test.cpp
```diff
@@ -444,7 +444,7 @@ START_SECTION(([EXTRA]
   TEST_EQUAL(drift_spectrum->getMZArray()->data.size(), 24)
   TEST_EQUAL(drift_spectrum->getMZArray()->data.size(), drift_spectrum->getIntensityArray()->data.size())
   TEST_EQUAL(drift_spectrum->getMZArray()->data.size(), drift_spectrum->getDriftTimeArray()->data.size())
-
+  /*
   IonMobilityScoring::driftScoringMS1Contrast(drift_spectrum, drift_spectrum_ms1, transitions, scores,
```
hroest: why is this commented out?

cbielow: all the tests I disabled for now raise an OPENSWATH_PRECONDITION error in Debug mode.
Here is the stack trace:
```
>	OpenSwathAlgod.dll!OpenSwath::MRMScoring::calcXcorrPrecursorContrastShapeScore() Line 546	C++
 	OpenMSd.dll!OpenMS::IonMobilityScoring::driftScoringMS1Contrast(boost::shared_ptr<OpenSwath::OSSpectrum> spectrum, boost::shared_ptr<OpenSwath::OSSpectrum> ms1spectrum, const std::vector<OpenSwath::LightTransition,std::allocator<OpenSwath::LightTransition> > & transitions, OpenMS::OpenSwath_Scores & scores, const double drift_lower, const double drift_upper, const double dia_extract_window_, const bool dia_extraction_ppm_, const double drift_extra) Line 335	C++
 	IonMobilityScoring_test.exe!main(int argc, char * * argv) Line 453	C++
```

where `xcorr_precursor_contrast_matrix_` is a 1x1 matrix (this is how its initialized in line 334 of IonMobilityScoring). I'm still baffled why our CI did not catch it...

jpfeuffer: We do not have CI in debug mode. I filed an issue some weeks ago because of the failing nightlies. https://github.com/OpenMS/OpenMS/issues/4272

cbielow: @jpfeuffer  is it feasible to run a debug build for CI on travis (probably gcc/clang, since VS is awfully slow yet much more sensitive)

jpfeuffer: Ok. I try it here: https://github.com/OpenMS/OpenMS/pull/4328
Although it will definitely only work with caching due to the time limit on travis.

cbielow:
since the debug build to detect the issue is still underway, I'd still like to merge this PR to get the nightlies going again

cbielow:
we should open a ticket which reminds to fix the underlying issue in the IM scoring code

jpfeuffer:
Fine with me personally

PR #4326, opened 2019-10-23
[NO MERGE] expose accessor to raw data

timosachsenberg:


timosachsenberg:
Following https://stackoverflow.com/questions/16065183/convert-a-stdvector-to-a-numpy-array-without-copying-data something like this might work.

``` 
cimport numpy as np
import numpy as np

    def get_matrix_unmanaged(self):
        """Cython signature: numpy_array, numpy_array get_matrix_unmanaged()
        
        Will return a numpy matrix corresponding to the matrix entries for faster access.
        """
        cdef _Matrix * mat_ = self.inst.get()
        cdef int rows = mat_.rows()
        cdef int cols = mat_.cols()
        double * vec_ptr= mat_.data();
        cdef double[::1] vec_view = <double[:vec.size()]>vec_ptr    # cast to typed memory view
        vec_npr = np.asarray(vec_view)    # get numpy array from memory view
        vec_npr.reshape( (rows, cols) )
        return vec_npr
```

timosachsenberg:
@hroest I can also copy the array but this might take a bit for e.g., 2000x2000 matrices

hroest:
>         vec_npr = np.asarray(vec_view)    # get numpy array from memory view

@timosachsenberg are you sure this is not copying it? can you compare pointers, are they pointing to the same memory or is new memory allocated? -- I would assume that if you call `np.asarray` that it will copy the input data, no matter whether its a memory view or something else?

timosachsenberg:
will open updated one

PR #4327, opened 2019-10-23
Feature/chrom by

hroest:
- allow extraction of spectra/chromatogram by id
- fix some issues when no meta data was present in OnDiscMSSpectrum
- add tests


timosachsenberg:
can we use this from pyOpenMS as well?
Would be nice to make our visualization widgets in pyopenms-extra compatible with on-disc experiments.

hroest:
> can we use this from pyOpenMS as well?

done

> Would be nice to make our visualization widgets in pyopenms-extra compatible with on-disc experiments.

you can already, but access is by integer id and not be native id (string)

timosachsenberg:
rebuild jenkins

hroest:
looks like its ready to merge, before only CodeFactor was complaining so all tests pass. @timosachsenberg merge?

PR #4328, opened 2019-10-24
[FEATURE] Add debug build to travis

jpfeuffer:
Let's see if this quick addition runs without error.


hroest:
looks like it timed out, I will re-start

cbielow:
I restarted it yesterday as well... and yet another time now after Hannes' restart.

The last log printed 
```
uild project

   Each symbol represents 1024 bytes of output.

   '!' represents an error and '*' a warning.

    ....................!.............................  Size: 49K

    ......................!!.!.!!!!!!!!!!!!!!!!!!!!!!!  Size: 99K

    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!  Size: 149K

    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!  Size: 199K

    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!Killed build prematurely to store whatever was already built in the cache. Please restart the travis job.
```

there are an awful lot of errors....

hroest:
lets see if at least the library builds

jpfeuffer:
Seems like it "worked" and we see the error:
```
/.../OpenMS/src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp: In member function ‘void OpenMS::NucleicAcidSpectrumGenerator::getSpectrum(OpenMS::MSSpectrum&, const OpenMS::NASequence&, OpenMS::Int, OpenMS::Int) const’:
src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp:406:38: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
```

cbielow:
this does not seem to run to completion... any ideas?
Can we use jenkinsCI for this?

Zethson:
You may also use Github CI, since it allows for jobs to run longer than Travis.

jpfeuffer:
yeah but they didn't have caching last time I checked. This would be relatively important.

Zethson:
https://github.com/actions/cache

?

jpfeuffer:
Ok nice, must have been added in the last weeks.
Do you also know if it is possible to copy artefacts from workflows of other dependent repositories
https://github.com/OpenMS/contrib/pull/93
?
Those two things let me stop on this "project".

Zethson:
No idea, sorry. Would have to google around myself :)

PR #4334, opened 2019-10-29
[FIX] Python 3 fix

hroest:
in preparation for newer autowrap versions, we need to fix relative imports


timosachsenberg:
I guess it's fine to merge this now. Only the memory test failed.

PR #4336, opened 2019-10-30
[FEATURE] TOPP tools can use input files which are looked up on PATH …

cbielow:
Feature: `TOPPBase::registerInputFile()` now allows to specify executables (which includes .jar files, or basically any other file), which are searched on the system PATH.
The advantage to the previous behaviour is that upon requesting the value of that option (using `getStringOption_()`) the existance is immediately checked via TOPPBase, in contrast to only when the tool is actually called (which could be much later during runtime).
This means that the tag `skipexists` can now be replaced by `is_executable` for most cases.
This has been done for all major Adapters. For some of them, extra code can be removed (especially the KNIME workaround using specific env-variables in the KNIME binaries.ini).
`skipexists` is still useful for specifying filenames in OpenMS/share, for example, which should not be checked for by TOPPBase, but are completed within the tool itself.


src/utils/SiriusAdapter.cpp
```diff
@@ -124,7 +124,14 @@ class TOPPSiriusAdapter :
 
   void registerOptionsAndFlags_() override
   {
-    registerInputFile_("executable", "<executable>", "", "sirius executable e.g. sirius", false, false, ListUtils::create<String>("skipexists"));
+    registerInputFile_("executable", "<executable>", 
+      // choose the default value according to the platform where it will be executed
+#ifdef OPENMS_WINDOWSPLATFORM
```
timosachsenberg: @jpfeuffer is this problematic if a workflow created on Linux is used on windows? and vice versa

cbielow:
tests seem to work, some unexplicable python fails. Does anyone have a clue why?

src/utils/SiriusAdapter.cpp
hroest: cant we just check for both?

cbielow: we could but this requires some extra coding... not worth the effort IMHO
ready for merge?

jpfeuffer:
```
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/source/SYSTEM/File.cpp: In static member function 'static OpenMS::StringList OpenMS::File::getPathLocations(const OpenMS::String&)':
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/source/SYSTEM/File.cpp:667:16: error: 'listSeparator' is not a member of 'QDir'
     path.split(QDir::listSeparator().toLatin1(), paths);
```
It runs on a docker container made from https://github.com/OpenMS/dockerfiles/blob/master/contrib/Dockerfile_min which has Ubuntu 16.04 and QT from apt -> v5.5. This function was introduced in 5.6.
Any objections to upgrade the minimum QT version?
Otherwise I think this is quite a small convenience function that can be avoided.

cbielow:
> 
> 
> ```
> /home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/source/SYSTEM/File.cpp: In static member function 'static OpenMS::StringList OpenMS::File::getPathLocations(const OpenMS::String&)':
> /home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/source/SYSTEM/File.cpp:667:16: error: 'listSeparator' is not a member of 'QDir'
>      path.split(QDir::listSeparator().toLatin1(), paths);
> ```
> 
> It runs on a docker container made from https://github.com/OpenMS/dockerfiles/blob/master/contrib/Dockerfile_min which has Ubuntu 16.04 and QT from apt -> v5.5. This function was introduced in 5.6.
> Any objections to upgrade the minimum QT version?
> Otherwise I think this is quite a small convenience function that can be avoided.

I have removed the function. Should work now.

cbielow:
all tests pass

timosachsenberg:
Regarding the Qt Version I would have had no objections

Issue #4337, opened 2019-10-31
SiriusAdapter keeps temporary files in -converter_mode


alexeigurevich:
Background: `-converter_mode` is added to SiriusAdapter in the `developer` branch of OpenMS. It is not available in 2.4.0 release and in the `master` branch. The goal of the option is just to create `.ms` file and reset further execution of the adapter. This `.ms` can be later processed with `sirius` binary.

Problem: whenever you run SiriusAdapter in `-converter_mode` you will get your temporary directory messed up with a new `.ms` which would never be removed. On Linux, it is just under `/tmp/`, on macOS, it is a bit more complicated, for myself they were stored under `/private/var/folders/b_/b_vxbtn56h7ffb8jssmbb1qc0000gn/T/`.

Reason: temporary file is copied [here](https://github.com/OpenMS/OpenMS/blob/develop/src/utils/SiriusAdapter.cpp#L226) but not removed afterward since the adapter finishes execution just [few lines after](https://github.com/OpenMS/OpenMS/blob/develop/src/utils/SiriusAdapter.cpp#L229) this.
Note: removing of temp files and dirs in SiriusAdapter is [here](https://github.com/OpenMS/OpenMS/blob/develop/src/utils/SiriusAdapter.cpp#L295) but this section never reached in `-converter_mode`.

@oliveralka please take a look!

oliveralka:
I will take a look, thanks for pointing that out!

jpfeuffer:
Seems to be fixed.

PR #4338, opened 2019-10-31
[FIX] Remove Sirius Tmp structure in converter mode

oliveralka:
Fixes the lack of removal of SiriusAdapterAlgorithm temporary directory and ms file: #4337


src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -68,6 +68,16 @@ namespace OpenMS
...
 
+      /**
+      @brief Removes temporary directory and file structure
+
+      @param debug_level Debug level
+      @param tmp_dir_path Path to temporary directory
+      @param tmp_file_path Path to temporary file
+      */
+
+      static void removeSiriusTmp(const int& debug_level, const String& tmp_dir_path, const String& tmp_file_path);
```
cbielow: suggestion: pass ints by value

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -144,6 +144,28 @@ namespace OpenMS
...
+    void SiriusAdapterAlgorithm::removeSiriusTmp(const int& debug_level, const String& tmp_dir_path, const String& tmp_file_path)
+    {
+      // clean tmp directory if debug level < 2 
+      if (debug_level >= 2)
+      {
+        OPENMS_LOG_DEBUG << "Keeping temporary files in directory " << tmp_dir_path << " and msfile at this location "<< tmp_file_path << ". Set debug level to 1 or lower to remove them." << std::endl;
+      }
+      else
+      {
+        if (tmp_dir_path.empty() == false)
```
cbielow: `if (!tmp_dir_path.empty())` .... ?

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -453,26 +453,10 @@ class TOPPAssayGeneratorMetabo :
...
-          }
-          if (tmp_ms_file.empty() == false)
-          {
-            writeDebug_("Deleting temporary msfile '" + tmp_ms_file + "'. Set debug level to 2 or higher to keep it.", 0);
-            File::remove(tmp_ms_file); // remove msfile
-          }
-        }
 
+        // remove temporary SiriusAdapter file structure
+        SiriusAdapterAlgorithm::removeSiriusTmp(debug_level_, tmp_dir, tmp_ms_file);
```
cbielow: seems like we need to remember to call this function at every premature exit point?!
This sounds like it should be delegated to the compiler using RAII?! (i.e. create a small class, which calls this function in its destructor)

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
oliveralka: I guess it is faster to use pass by value (also for primitive types such as bool and byte) since no dereferencing operation is needed - is that correct?

src/utils/AssayGeneratorMetabo.cpp
oliveralka: Ah, did not know about that one - I will have a look. Any suggestions for a somewhat easily digestible (c++) programming concept read?

oliveralka:
@cbielow I think the current commit goes in the right direction in terms of RAII, but is not quite there yet. 

Question: Probably the `SiriusTmpStruct` should be refactored as a small class (with the `removeSiriusTmp` and `constructSiriusTmpStruct` methods). Do you have further suggestions? E.g. call the  `constructSiriusTmpStruct` via the constructor of the new class? Is there a better way to integrate the debug_level? 

Thanks!

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
cbielow: yes, but the compiler might actually do other fancy things if it works on a copy (which is cheap in this case).

cbielow:
yes, almost there:
move `static SiriusAdapterAlgorithm::SiriusTmpStruct constructSiriusTmpStruct(int debug_level);` into the Ctor of `SiriusTmpStruct`. Make all existing members private (data and functions).
Maybe rename the class to something more descriptive.

oliveralka:
rebuild jenkins

oliveralka:
@jpfeuffer Not sure what is wrong with ubuntu? Build system related?

oliveralka:
rebuild jenkins

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -43,6 +43,8 @@
 #include <OpenMS/SYSTEM/File.h>
 #include <OpenMS/FORMAT/FeatureXMLFile.h>
 
+#include <QDir>
```
cbielow: you can move most includes to the .cpp and just forward declare arguments to the major functions

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -54,19 +56,84 @@ namespace OpenMS
...
       {
-        String tmp_dir;
-        String tmp_ms_file;
-        String tmp_out_dir;
-      };
+      public:
 
-      /**
-        @brief Construct temporary folder structure for SIRIUS (SiriusTmpStruct)
+        SiriusTemporaryFileSystemObjects(int debug_level) : removeable_(false) { constructSiriusTemporaryFileSystemObjects_(debug_level); }
```
cbielow: you should move the implementation of all functions to the .cpp

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -54,19 +56,84 @@ namespace OpenMS
...
+          return tmp_out_dir_;
+        } 
 
+        String getTmpMsFile()
+        {
+          return tmp_ms_file_;
+        }
+      
+      private:
+        bool removeable_;
```
cbielow: this bool is always true. Why not toss it?

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -54,19 +56,84 @@ namespace OpenMS
...
+
+        /**
+        @brief Construct temporary folder structure for SIRIUS (SiriusTmpStruct)
+        @param debug_level Debug level
         @return SiriusTmpStruct
-      */
-      static SiriusAdapterAlgorithm::SiriusTmpStruct constructSiriusTmpStruct();
+        */
+
+        void constructSiriusTemporaryFileSystemObjects_(int debug_level)
```
cbielow: this function is only called once. Just move its content to the ctor and get rid of it.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -54,19 +56,84 @@ namespace OpenMS
...
+
+        /**
+        @brief Removes temporary directory and file structure
+
+        @param debug_level Debug level
+        @param tmp_dir Path to temporary directory
+        @param tmp_ms_file Path to temporary ms file
+        */
+
+        void removeSiriusTemporaryFileSystemObjects_(int debug_level, const String& tmp_dir, const String& tmp_ms_file)
```
cbielow: also only called once. move to dtor

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -54,19 +56,84 @@ namespace OpenMS
...
-      };
+      public:
 
-      /**
-        @brief Construct temporary folder structure for SIRIUS (SiriusTmpStruct)
+        SiriusTemporaryFileSystemObjects(int debug_level) : removeable_(false) { constructSiriusTemporaryFileSystemObjects_(debug_level); }
+
+        ~SiriusTemporaryFileSystemObjects(){ if (removeable_) removeSiriusTemporaryFileSystemObjects_(debug_level_, tmp_dir_, tmp_ms_file_); }
+
+        String getTmpDir()
```
cbielow: should be `const String& getTmpDir() const;`

cbielow: similarly for the other accessors

src/utils/SiriusAdapter.cpp
```diff
@@ -192,10 +192,10 @@ class TOPPSiriusAdapter :
     f.load(in, spectra);
 
     // make temporary files
-    SiriusAdapterAlgorithm::SiriusTmpStruct sirius_tmp = SiriusAdapterAlgorithm::constructSiriusTmpStruct();
-    String tmp_dir = sirius_tmp.tmp_dir;
-    String tmp_ms_file = sirius_tmp.tmp_ms_file;
-    String tmp_out_dir = sirius_tmp.tmp_out_dir;
+    SiriusAdapterAlgorithm::SiriusTemporaryFileSystemObjects sirius_tmp(debug_level_);
+    String tmp_dir = sirius_tmp.getTmpDir();
```
cbielow: why use extra variables and not just call the get...() functions where needed?

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -54,19 +56,84 @@ namespace OpenMS
...
+        @brief Removes temporary directory and file structure
+
+        @param debug_level Debug level
+        @param tmp_dir Path to temporary directory
+        @param tmp_ms_file Path to temporary ms file
+        */
+
+        void removeSiriusTemporaryFileSystemObjects_(int debug_level, const String& tmp_dir, const String& tmp_ms_file)
+        {
+          // clean tmp directory if debug level < 2
```
cbielow: extra:
maybe add a 
`constexpr int debug_threshold = 2;`
and use it throughout the function to have a single place of reference for the required debug level (avoiding mentioning it multiple times which could go out of sync if we decide to change it at some point)

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
oliveralka: Thanks for all your suggestions! 

Could you have a third look on the forward declarations - if I did that correctly?

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
oliveralka: Just for my understanding: The second const (after the function) will throw a compiler error for this class function if a member variable of the class will be changed. 

So it is basically a safety mechanism, or does it have further additional functionality?

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -35,38 +35,45 @@
...
-#include <OpenMS/KERNEL/MSSpectrum.h>
-#include <OpenMS/KERNEL/FeatureMap.h>
-#include <OpenMS/ANALYSIS/QUANTITATION/KDTreeFeatureMaps.h>
 #include <OpenMS/ANALYSIS/MAPMATCHING/FeatureMapping.h>
 
-#include <OpenMS/SYSTEM/File.h>
-#include <OpenMS/FORMAT/FeatureXMLFile.h>
-
-#include <QString>
+class QDir;
```
cbielow: this is not needed?!

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -134,16 +139,53 @@ namespace OpenMS
...
+      // clean tmp directory if debug level < debug threshold
+      if (debug_level_ >= debug_threshold)
+      {
+        OPENMS_LOG_DEBUG << "Keeping temporary files in directory " << tmp_dir_ << " and msfile at this location "<< tmp_ms_file_ << ". Set debug level to 1 or lower to remove them." << std::endl;
+      }
+      else
+      {
+        if (!tmp_dir_.empty())
+        {
+          OPENMS_LOG_DEBUG << "Deleting temporary directory " << tmp_dir_ << ". Set debug level to 2 or higher to keep it." << std::endl;
```
cbielow: there is still mentions of the hardcoded debug level. replace by the variable

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
oliveralka: Yes, you are right. That is a remnant - from me playing around with the forward declaration - forgot to remove it in the end...

oliveralka:
rebuild jenkins

src/utils/SiriusAdapter.cpp
```diff
@@ -290,29 +290,10 @@ class TOPPSiriusAdapter :
     // should the ms file be retained (non-converter mode)
     if (!out_ms.empty())
     {  
-      QFile::copy(tmp_ms_file.toQString(), out_ms.toQString());
+      QFile::copy(sirius_tmp.getTmpMsFile().toQString(), out_ms.toQString());
```
cbielow: just realizing: could you also move the file (instead of copying)?
i.e. would the deletion of the (non-existent) `sirius_tmp.getTmpMsFile()` be a problem?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -33,11 +33,16 @@
 // --------------------------------------------------------------------------
 
 #include <OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h>
-
+#include <OpenMS/FORMAT/FeatureXMLFile.h>
```
cbielow: optional (knowing that Im getting picky): sorting includes lexicographically has proven to be better maintainable (avoids duplicate includes -- which are not a performance problem, but noise nevertheless)

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -443,7 +441,7 @@ class TOPPAssayGeneratorMetabo :
           sirius_workspace_directory = String(sw_dir.absolutePath());
 
           // move tmp folder to new location
-          bool copy_status = File::copyDirRecursively(tmp_dir.toQString(), sirius_workspace_directory.toQString());
+          bool copy_status = File::copyDirRecursively(sirius_tmp.getTmpDir().toQString(), sirius_workspace_directory.toQString());
```
cbielow: same here: moving possible?
If yes, you could set the `sirius_tmp.getTmpDir()` to empty so the DTor does not invoke deletion of a non-existant directory. Or am I missing something?

src/utils/SiriusAdapter.cpp
oliveralka: I don't think it would be a problem - I can try it out. 

I thought that if debug > 3 - it would be nice to have everything in one place (the directory and the .ms file. If not the tmp_ms will be deleted anyway later on. 

Does that make sense?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
oliveralka: No worries! At least something to learn here ;)

src/utils/AssayGeneratorMetabo.cpp
oliveralka: Same as above. I think moving would be possible, but debug would then reference a non-existing path since the output had been moved previously. 

I copied to have the debug output in the same place. 

If the debug output is not needed it is removed afterwards after the copy step. 

If "-debug 3" would be used, the path before the move would be referenced see d'tor SiriusAdapterAlgorihm.

oliveralka:
@cbielow I would leave it at the copy for now since the path to the temporary directory and ms file are used for the debug mode. I will have to restructure some code further for the next SIRIUS release and will think about a better integration regarding the move and debug method. Would that be alright?

cbielow:
yep. feel free to merge when ready.

Issue #4332, opened 2019-10-28
Warning about primary MS run format


hendrikweisser:
Since #4210 (@timosachsenberg), I get warnings like this when I run NucleicAcidSearchEngine:
```
To ensure tracability of results please prefer mzML files as primary MS run.
Filename: 'file:///X:\win\path/file.raw'
```
(Also note the mismatched path separators.)

The code for this is here: https://github.com/OpenMS/OpenMS/blob/5d961be216823625676053b6782fc385ec413c2c/src/openms/source/METADATA/ProteinIdentification.cpp#L376-L380

From our perspective, the .raw file *is* the primary MS file. We may delete converted mzML files, but will definitely keep the .raw files. I would expect that to be quite a common practice, too (among labs that use mzML files).

What is the point of this warning?
Shouldn't we be happy with .raw files as primary MS files?

timosachsenberg:
There are a few reasons so I only list some of them here:
- some formats and downstream tools want to trace down a file that actually exists on the file system (e.g., for visualization) form the information contained in a e.g. mzTab file. This was the preferred way to carry readable spectra information through the other file formats.
- entry for the raw file is often malformed or not filled out - conversion from Thermo is kind of an exception
- folder often only exists on the instrument machine etc.

TL;DR to do meaningful visualize a result file (e.g., quant + id) at the end of a processing workflow we need the first "sensible" mzML files (e.g., after conversion or peak picking)

One could think about carrying the raw file name (if it exists - see comment above) in another meta data field.

hendrikweisser:
Thanks for the quick reply, Timo, but I still find this problematic. Imagine an analysis pipeline that produces several intermediary files (e.g. filtering, signal processing):
.raw -> .mzML -> ... -> .mzML -> idXML/featureXML

Generally, I think it's most important to store the identity of the very first file (the .raw file), because all others are derived from that and it may provide the link to information about the experiment/sample.

For visualisation, you probably want the most recent .mzML file, not the first one. But this depends a lot on the situation/workflow, so I'm not sure there is even a general solution.

Accessibility of the file should not play a role, in my opinion. In any particular processing step, the current input file is guaranteed to be accessible; whether any earlier file is still accessible should not affect the meta data at the current step. Otherwise it's just confusing.

Where (by which tools) is the "spectra_data" information currently actually used?

timosachsenberg:
[...]

PR #4342, opened 2019-11-06
Support for raw files as primary MS runs (solves #4332)

hendrikweisser:
These changes unify the handling of primary MS run information in class `ProteinIdentification` and extend it to support keeping track of raw files (in addition to mzMLs).
This also avoids the warning in NucleicAcidSearchEngine mentioned in #4332.

### DIFF
--- a/src/openms/source/KERNEL/MSExperiment.cpp
+++ b/src/openms/source/KERNEL/MSExperiment.cpp
@@ -503,7 +503,10 @@ namespace OpenMS
       }
       else
       {
-        String ms_run_location = path + "/" + filename;
+        // use Windows or UNIX path separator?
+        String actual_path = path.hasPrefix("file:///") ? path.substr(8) : path;
+        String sep = (actual_path.has('\\') && !actual_path.has('/')) ? "\\" : "/";
+        String ms_run_location = path + sep + filename;
         toFill.push_back(ms_run_location);
       }
     }
--- a/src/openms/include/OpenMS/METADATA/ID/IdentificationDataConverter.h
+++ b/src/openms/include/OpenMS/METADATA/ID/IdentificationDataConverter.h
@@ -202,6 +202,5 @@ namespace OpenMS
     static void exportMSRunInformation_(
       IdentificationData::ProcessingStepRef step_ref,
       ProteinIdentification& protein);
-
   };
 }
--- a/src/openms/include/OpenMS/METADATA/ID/IdentificationDataConverter.h
+++ b/src/openms/include/OpenMS/METADATA/ID/IdentificationDataConverter.h
@@ -197,5 +197,11 @@ namespace OpenMS
     /// Helper function to export DB search parameters to legacy format
     static ProteinIdentification::SearchParameters exportDBSearchParameters_(
       IdentificationData::SearchParamRef ref);
+
+    /// Helper function to export (primary) MS run information to legacy format
+    static void exportMSRunInformation_(
+      IdentificationData::ProcessingStepRef step_ref,
+      ProteinIdentification& protein);
+
   };
 }
--- a/src/openms/include/OpenMS/METADATA/ProteinIdentification.h
+++ b/src/openms/include/OpenMS/METADATA/ProteinIdentification.h
@@ -237,16 +237,16 @@ public:
 
       SearchParameters();
       /// Copy constructor
-      SearchParameters(const SearchParameters &) = default;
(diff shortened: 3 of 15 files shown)


src/openms/source/KERNEL/MSExperiment.cpp
```diff
@@ -503,7 +503,10 @@ namespace OpenMS
       }
       else
       {
-        String ms_run_location = path + "/" + filename;
+        // use Windows or UNIX path separator?
+        String actual_path = path.hasPrefix("file:///") ? path.substr(8) : path;
```
timosachsenberg: I would probably prefer "/" as Qt does
https://doc.qt.io/qt-5/qdir.html#cleanPath

timosachsenberg:
Having the raw flag seems a bit counterintuitive and harder to wrap.
What is your opinion on splitting that up into something like getPrimaryMSRun... and getRawFile.. ?

src/openms/source/KERNEL/MSExperiment.cpp
hendrikweisser: I like the idea, but when I tried it just now, `QDir::cleanPath` didn't actually convert the backslashes - presumably because the "file:///" part makes it recognise the whole thing as a UNIX path. I could just go ahead and convert all "\" to "/", but I think technically "\" is allowed in UNIX paths.

hendrikweisser:
For pyOpenMS, I think you can just add an overload with the additional bool parameter (although I forgot to do that).

I agree that the flag makes for a less intuitive interface, but I don't like the alternative - extending the `ProteinIdentification` API with for four marginally useful functions - either. In addition, I still believe that the term "primary MS run" should more accurately refer to the raw file - that's what you'd typically get from `MSExperiment::getPrimaryMSRunPath`. In that sense, the API reflects that we can't agree on what the "primary MS run" should actually be. ;-)

timosachsenberg:
> I agree that the flag makes for a less intuitive interface, but I don't like the alternative - extending the ProteinIdentification API with for four marginally useful functions - either.

The thing is that the primaryMSRun part is not marginally useful but essential to get automated processing working that uses the experimental design. IMHO it should definitely be part of the interface. Regarding the name I am open for suggestions. I am not 100% happy with the name either.

So maybe we should include more developers to decide on that.  

@OpenMS/developers 

So basically it is:
having a getPrimaryMSRunPath(, bool raw=false)
vs.
keeping the getPrimaryMSRunPath()
and adding getRawPath()

with potential suggestions for better names.

hendrikweisser:
> The thing is that the primaryMSRun part is not marginally useful

Yes, and I didn't mean to imply otherwise. But the variants for raw files are essentially only for bookkeeping, and it looks like at the moment I'm the only one who sees a need for them.

I'm happy to keep the discussion open, but would appreciate it if we could merge the current version for now - I hope we can agree that it's at least better than what we had before.

hroest:
I also believe conceptually that the primary file we refer to is the raw file which is what the machine produced and I think we should track that if at all possible. Why not simply keep a list of files including raw and mzML files and if the code needs to refer back to the raw data, it can simply go through the list and check which files are present and in a readable format (mzML / mzXML) which will then be used for the experimental design? @timosachsenberg would that work?

timosachsenberg:
@hroest One can get this to work in principle but make the reference to the filename in the experimental design much less explicit as I prefer to have it.
In the "list" solution, one has to deal with existing and potentially non-existing .raw files. which makes the code on the callers side a bit more ugly.
IIRC there also some parts in OpenMS that need to merge these file lists and I imagine that this might also get more complex.
But yes - it's also possible and I would be fine if the majority prefers this.

hendrikweisser:
> Why not simply keep a list of files

Because the semantics of what's in the list wouldn't be clear. You can either have multiple files from a single processing step or single files from multiple steps, but you can't distinguish these with only one list.

In my opinion `ProteinIdentification` is a legacy format and there's not much point investing a lot of work in it when it's to be replaced by `IdentificationData` - where each piece of data can track which steps were applied to it and what files were associated with that.

timosachsenberg:
rebuild jenkins

timosachsenberg:
Ok, I merged it as is as my opinion is not too strong on that and there was not that much feedback. Regarding IdentificationData structure I think we need to get to the point where we have an easy and intuitive API for most of the common tasks. Both from the C++ side as well as from pyOpenMS. This is still a bit to go.

PR #4351, opened 2019-11-12
[FIX] FileConverter: changes default output to mzML in KNIME

timosachsenberg:


src/topp/FileConverter.cpp
```diff
@@ -336,7 +336,7 @@ class TOPPFileConverter :
     String method("none,ensure,reassign");
     setValidStrings_("UID_postprocessing", ListUtils::create<String>(method));
 
-    formats = "mzData,mzXML,mzML,cachedMzML,dta2d,mgf,featureXML,consensusXML,edta,csv";
+    formats = "mzML,mzXML,cachedMzML,mgf,featureXML,consensusXML,edta,mzData,dta2d,csv";
```
cbielow: we could gradually move to C11 throughout by using
`vector<String> formats = {"mzML", "mzXML", "cachedMzML", "mgf", "featureXML", "consensusXML", "edta", "mzData", "dta2d", "csv"};`

PR #4355, opened 2019-11-13
[FIX] add exe file format to thermo raw file parser.exe parameter

timosachsenberg:


jpfeuffer:
I guess we did not add exe to our valid file formats yet.

PR #4356, opened 2019-11-13
PeptideIndexer does not match all peptides to DB

cbielow:
this PR fixes two issues:
1) removing restriction on x!tandem special cutting rule: cutting sites are now also valid in semi-specific and unspecific digest settings, since peptides like
D.PTDDIYNRQDFGWSFAGYGAPIMEDD.P would otherwise not be found in semi-tryptic (since both ends require the special cutting rule).

2) our regular expressions for all enzymes do not consider ambiguous amino acids (B, Z, J and X), thus a peptide
`K.ACDX.A` would not be considered fully tryptic (since the `X` at the c-term is not recognized as `K|R`).
This PR fixes all the regexes.

### DIFF
--- a/share/OpenMS/CHEMISTRY/Enzymes.xml
+++ b/share/OpenMS/CHEMISTRY/Enzymes.xml
@@ -3,7 +3,7 @@
   <NODE name="Enzymes">
     <NODE name="Trypsin">
       <ITEM name="Name" value="Trypsin" type="string" />
-      <ITEM name="RegEx" value="(?&lt;=[KR])(?!P)" type="string" />
+      <ITEM name="RegEx" value="(?&lt;=[KRX])(?!P)" type="string" />
       <ITEM name="RegExDescription" value="Trypsin cleaves following a K or R residue unless the next residue is P." type="string" />
       <ITEM name="NTermGain" value="H" type="string" />
       <ITEM name="CTermGain" value="OH" type="string" />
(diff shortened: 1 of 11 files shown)


cbielow:
follow-up on #2497

cbielow:
rebuild jenkins

cbielow:
the Windows-CI failed to build two tests... for unknown reasons (http://cdash.openms.de/viewBuildError.php?buildid=344947), so ready to merge I think.

oliveralka:
Seems that the issue can have a lot of different reasons, which are probably more related to cmake / build system (https://stackoverflow.com/questions/13118947/error-msb6006-cmd-exe-exited-with-code-1/20391803).

We can try to rebuild once more and then merge if everything is ok - looks good ;)

oliveralka:
rebuild jenkins

cbielow:
ok, Win succeeds now, but Ubuntu fails with other tests (unrelated).
Any comments on the code or merge?

hendrikweisser:
I'm getting failed tests on my Ubuntu (18.04.3) that look related to this:

	228 - PepXMLFile_test (Failed)
	390 - ProteaseDB_test (Failed)
	501 - OPXLHelper_test (Failed)
	2201 - TOPP_PeptideIndexer_7_out (Failed)

@cbielow: Are you able to look into this?

cbielow:
the test all pass on my Debian. can you execute with -VV and pinpoint whats wrong?

jpfeuffer:
You can do e.g. `ctest -R PepXMLFile -VV`

hendrikweisser:
Sorry, my mistake! The problem was caused by the fact that an older "share" directory was used on my system. (It's that annoying situation that if you ever `make install`, afterwards even development builds will use the "share" from the installation directory, not the source one.)

PR #4360, opened 2019-11-15
[FEATURE] InternalCalibration fallback when no calibration points

cbielow:
allow -force for IC when no calibration curve could be found - emitting unchanged mzML

### DIFF
--- a/src/topp/InternalCalibration.cpp
+++ b/src/topp/InternalCalibration.cpp
@@ -171,10 +171,10 @@ protected:
   {
     // data
     registerInputFile_("in", "<file>", "", "Input peak file");
-    setValidFormats_("in", ListUtils::create<String>("mzML"));
+    setValidFormats_("in", {"mzML"});
     registerOutputFile_("out", "<file>", "", "Output file ");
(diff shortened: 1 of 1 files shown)


src/topp/InternalCalibration.cpp
```diff
@@ -337,8 +337,17 @@ class TOPPInternalCalibration :
...
-      return UNEXPECTED_RESULT;
+      OPENMS_LOG_ERROR << "No calibration points found! Check your Raw data and calibration masses." << std::endl;
+      if (!getFlag_("force"))
+      {
+        OPENMS_LOG_ERROR << "Set the 'force' flag to true if you want to continue with uncalibrated data." << std::endl;
+        return UNEXPECTED_RESULT;
+      }
+      OPENMS_LOG_ERROR << "The 'force' flag was set to true. Storing uncalibrated data to '-out'." << std::endl;
+      // do not calibrate
+      addDataProcessing_(exp, getProcessingInfo_(DataProcessing::CALIBRATION));
```
hroest: do you want to label this uncalibrated data as "calibrated" ?

Issue #4364, opened 2019-11-18
PrecursorIonSelector string arguments that are actually files


bernt-matthias:
In `PrecursorIonSelector` the arguments `-sim_results`, `-rt_model`, and `-dt_model` are shown as string but are actually files. 

While this seem no problem on the command line it is for parsing (and automatically converting) the CTD files where the options are also shown with `type="string"`.

PR #4372, opened 2019-11-21
make file type explicit in PIS

timosachsenberg:
might fix https://github.com/OpenMS/OpenMS/issues/4364


jpfeuffer:
And can you check other occurrences of rt/pt/dt models? I have the feeling that they are not always file options.
This breaks old workflows however, just saying. But it might be clearer in the end.

timosachsenberg:
I think they are files written by e.g.
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/PTModel.cpp
which already have the .txt extension

jpfeuffer:
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/RTPredict.cpp#L130

I mean occurrences like this. But it seems like they are files already. So it's ok.

Issue #4366, opened 2019-11-19
pyOpenMS macOS Catalina 10.15.1


oliveralka:
Building pyOpenMS fails, due to SDK changes in macOS Catalina (10.15.1).

`xcrun --show-sdk-path:  /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk`
Headers can be found here: 
`/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include`

General: Check for specific MacOS Version from 10.15.* 
```
import platform
platform.mac_ver()[0] 
```
-> `'10.15.1'`

There are multiple possibilities to fix the SDK in src/pyOpenMS/setup.py:

1) add `-isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk ` to compile.args

The issue here is that this might be different in other versions or if the Xcode.app is installed, then the SDK seems to be in a different path:
`/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk`

2) add python package e.g. subprocess, call `xcrun --sdk macosx --show-sdk-path` and add it to the `-isysroot`

Which one would be preferable or is there a better option available?

jpfeuffer:
Which compiler binary is it using?
AFAIK the wrapper at /usr/bin/clang++ should have the current sysroot as default header path.
This is not true if you are using the `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/.../clang``

We can also just configure the setup.py with CMake (since CMake automatically determines the default sysroot of the used compiler executable), then we do not have to rely and parse the output of xcrun manually.

oliveralka:
hm, I am using: 

`/Library/Developer/CommandLineTools/usr/bin/c++ ` which points to clang in the same directoy. 
`/usr/bin/clang++` also refers to InstalledDir: `/Library/Developer/CommandLineTools/usr/bin`

You are right, in the OpenMS cmake the sysroot was set correctly to `/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk`   

As far as I know, env.py is created by our current cmake (environment variables for pyopenms setup.py).
We could add a variable for the sysroot there and add it to the compiler variables in setup.py if macOS is used, this should then work independent of the macOS version, or?

jpfeuffer:
Yes that is a good idea.
[...]

PR #4377, opened 2019-11-25
[FIX] pyopenms macOS catalina

oliveralka:
ADD sysroot variable to env.py using cmake.
Use variable depending on macOS version (setup.py)

Increased minimal CMake version to 3.1 (similar to OpenMS; CMAKE_OSX_SYSROOT available since 3.0.2)

Refers to #4366


jpfeuffer:
Maybe add a check for empty somewhere, otherwise it looks good.

jpfeuffer:
And I would maybe add it for mojave already, since the headers are not installed globablly by default there either.
(Yes, you can install manually but not everyone does it)

oliveralka:
`mac_fix_automation_mojave` is now copied to the pyopenms build directory and can be used after 
`python setup.py bdist_wheel` to automatically fix all the library dependencies (including the frameworks).

oliveralka:
mac_fix_automation_mojave.sh does currently not work, since the QT libraries are not copied over into the whl - have to check.

oliveralka:
@jpfeuffer Since there is currently no straight forward way to fix the libraries using the fix_dependencies.rb script internally, since the build/lib folder is not automatically specified in our cmake, should we leave it at that?

oliveralka:
Closed due to merge with #4398

PR #4379, opened 2019-11-25
[FEATURE] use compile feature instead of fixed CXX standard to require only a minimum

jpfeuffer:
This should be tested on multiple compiler versions first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@
 # $Authors: Stephan Aiche, Julianus Pfeuffer $
 # --------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


hroest:
would this mean that 4.8 does still work?

jpfeuffer:
Good question: If CMake defined 4.8 as fully C11 compatible, yes. I'll check

jpfeuffer:
Otherwise, we might reduce the features needed, such that 4.8 is included.

jpfeuffer:
rebuild jenkins

timosachsenberg:
builds on Trusty: gcc (Ubuntu 4.8.4-2ubuntu1~14.04.4) 4.8.4

PR #4381, opened 2019-11-25
[FIX] use is_Executable for thermoraw and mono

jpfeuffer:


timosachsenberg:
yep think this is correct

cbielow:
the annotation is not checked, since the parameter is not required and empty by default.
Only if the user sets a non-default value, the presence of the executable will be checked (this might be a flaw in how this is implemented in TOPPBase though)

jpfeuffer:
If I added a default for at least ThermoRaw (since we ship it). Would it then be checked?

jpfeuffer:
Hmm but then it would complain even if it is not needed (since it is not required).
Hmm..

cbielow:
the annotation is not checked, since the parameter is not required and empty by default.
Only if the user sets a non-default value, the presence of the executable will be checked (this might be a flaw in how this is implemented in TOPPBase though)

> 
> 
> If I added a default for at least ThermoRaw (since we ship it). Would it then be checked?

No, unless the user changes the default (this is not really desirable though).
Shall we change the internal logic to:
Check the file for existance/executability iff either:
 - required == true
 - or non-empty value (which might be the default) 
is given

jpfeuffer:
Sounds good. But we should make sure, that those defaults then are usually found such that the tools only complain when something is really wrong. (In that case they can still remove the default).

cbielow:
with #4384, you now get better checks.
But the `net_executable` is empty by default and not required, which means, there will be no checking.
Its not a big deal because the execution happens (or fails) immediately afterwards -- so no loss of time.
but one might do the defaults using `#ifdef` when registering the string based on platform.

jpfeuffer:
exactly, dont worry about this. We cannot ship NET or mono so I am not sure if we should provide defaults at all.

PR #4382, opened 2019-11-26
tests make RTPredict parameters explicit

bernt-matthias:
explicitly specify `-in_oligo_params` and `-in_oligo_trainset` instead of implicit by appending `_additional_parameters` and `_samples` to the value given by `-svm_model`.

I need this for the automatic test generation for Galaxy tools.


jpfeuffer:
Where do you need this fix to be applied? Does it suffice to be in develop for the upcoming release or is it urgent such that it really needs to be backported to 2.4?

bernt-matthias:
We can do this as it is most convenient for you, since I can work locally on my branch.

Out of interest is there already a plan / schedule for the next release?

jpfeuffer:
oh. I tried to change the base to develop, since that would be much more convenient. Seems like there are too many diffs now.
Could you maybe reopen a PR against develop?

We are working on the last fixes for installers and deployment (e.g. for pyopenms), then we want to release 2.5.

bernt-matthias:
I just cherry picked the commit .. seemed to work.

jpfeuffer:
Yes, looks better. Let's see.

PR #4383, opened 2019-11-26
remove 1 myrimatchadapter test

bernt-matthias:
because the -min_precursor_charge 4 -max_precursor_charge 3 parameters
do not exist it seems not useful to test for them to fail


jpfeuffer:
While I wouldn't say that this test is useless, I can see that it is maybe not the right place to test the parameter checking and why it would be impossible to test for your use-case.

I would be okay with removing it. Maybe someone else has an opinion on it, too.

cbielow:
this test seems a simple copy&paste bug from https://github.com/OpenMS/OpenMS/commit/67d9f5706516d36031d5002679345ff9d579e6ea

just kick it

jpfeuffer:
failures unrelated. Merge?

PR #4384, opened 2019-11-26
Opt include

cbielow:
augments #4381

fix behaviour of TOPPBase InputFile tags ('is_executable' and 'skipexists').
They are checked now more frequently, i.e. if the value is required or non-empty.
The checks happens when `getStringOption_()` is called.

### DIFF
--- a/src/utils/MetaProSIP.cpp
+++ b/src/utils/MetaProSIP.cpp
@@ -1931,7 +1931,7 @@ public:
 class RIntegration
 {
 public:
-  // Perform a simple check if R and all R dependencies are thereget
+  // Perform a simple check if R and all R dependencies are there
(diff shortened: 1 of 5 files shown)


src/utils/MetaProSIP.cpp
```diff
@@ -3331,33 +3331,21 @@ class TOPPMetaProSIP :
       {
         if (isotopic_intensities[i] < 1e-4) continue;
         Size consecutive_isotopes = 0;
-        Int j = i;
+        Size j = i;
 
-        while (j >= 0)
+        while (j != -1)
```
jpfeuffer: this fails I think.

cbielow: yeah... compiler warnings... fixed.

cbielow:
rebuild jenkins

cbielow:
checks pass. merge?

PR #4385, opened 2019-11-26
SQLite-based file format for identification data

hendrikweisser:
This adds a new file format ".oms" to store peptide/protein (or RNA, small molecule) identification results based on the internal `IdentificationData` class, using an SQLite database for storage. For the moment the new format is only exposed in IDFileConverter to test conversion between idXML and .oms. (My plan is to add support for .oms and `IdentificationData` in MapAlignerIdentification next.)

I'm opening this pull request now for several reasons:
1. To get feedback on the design and implementation.
2. To get others (i.e. you!) to test the data conversion (to see if `IdentificationData` supports all use cases).
3. Before too long, to merge the changes to avoid the branches going out of sync (again)

The implementation of storing and loading to/from .oms files (in class `OMSFile`) should be almost complete. Support for "molecule-query match groups" (e.g. groups of PSMs - for crosslinking, chimeric spectra etc.) is missing because that feature itself may need a redesign. @enetz, it would be good to get some input on use cases/requirements for this.

If you want to look at the data in an .oms file (database), I found the "DB Browser for SQLite" tool very helpful: http://sqlitebrowser.org, Ubuntu package "sqlitebrowser".


timosachsenberg:
Hi @hendrikweisser and also @oliveralka and @jpfeuffer.
First thanks for this long-awaited addition that helps to solve the representability problem of non-proteomic data and hopefully our scalability problems.
Speaking of scalability... I think we should discuss how and what kind of subsets we can query from the oms files. 
Possible use cases are e.g.:
- if we need only information for one single raw file or all of its fractions etc.
- querying a subset of ids (e.g. all ids that have been linked to features over potentially thousands of runs)
- retrieving these queries as chunks if these are still too large to fit into memory

@jpfeuffer what kind of queries would you need to e.g. make sure that you can do inference on thousands of raw files?

cbielow:
some stupid questions:
 - is there an advantage for using QtSql vs. sqlite (which we already have), e.g. interface, speed...
 - any reasons why its called ".oms"? what about "odb" or "oql"? :)

hendrikweisser:
@timosachsenberg:
> Speaking of scalability... I think we should discuss how and what kind of subsets we can query from the oms files.

Filtered queries would be possible in principle, but a problem may be that you have to end up replicating a lot of the logic in `IdentificationData` or `IDFilter` that keeps data consistent. E.g. if you only want certain peptides, you also want only the corresponding proteins, PSMs, data processing steps, etc.
So before implementing these queries, check first whether loading all the data is still prohibitive with the new data structure, since that removes the redundancy.

hendrikweisser:
@cbielow:
> is there an advantage for using QtSql vs. sqlite (which we already have), e.g. interface, speed...

QtSql is just a wrapper. I found it more convenient and expressive than the SQLite C interface (and I'm against implementing another wrapper ourselves).

> any reasons why its called ".oms"? what about "odb" or "oql"? :)

Not that I know. It's what Julianus and Oliver A. chose at the start. I'd be to open to renaming it if we find a significantly better name.

timosachsenberg:
> Filtered queries would be possible in principle, but a problem may be that you have to end up replicating a lot of the logic in `IdentificationData` or `IDFilter` that keeps data consistent. E.g. if you only want certain peptides, you also want only the corresponding proteins, PSMs, data processing steps, etc.
> So before implementing these queries, check first whether loading all the data is still prohibitive with the new data structure, since that removes the redundancy.

There is a trend that studies increase in size and >1000 raw files are quite common nowadays.
So it is very unlikely that these will fit into memory and we need to figure out how to handle those.
Otherwise, we might end up with yet another file format and datastructure ;)

I am not really an expert in these fields so I think we should get some additional input / do some research.
Apparently, there are some "patterns" that try to cope which that issue
see e.g. the reference to read-models in
http://gorodinski.com/blog/2012/06/16/orm-lazy-loading-pitfalls/

jpfeuffer:
Yes, regarding filtered queries my initial approach would have been to use a graph aware database backend (that automatically gets you parents etc.) but others thought that it might be too specific for general queries.

oliveralka:
> * any reasons why its called ".oms"? what about "odb" or "oql"? :)

We named it .oms (OpenMS) in dependence on .osw (OpenSWATH), but as Hendrik already stated we could rename it, if needed.

hendrikweisser:
> There is a trend that studies increase in size and >1000 raw files are quite common nowadays.
So it is very unlikely that these will fit into memory and we need to figure out how to handle those.
Otherwise, we might end up with yet another file format and datastructure ;)

I don't think this is so common that people want to run these analyses on their laptops. If you have a beefy server, I'm not sure you couldn't fit the data into memory - there will be lot of PSMs, but those should have a small memory footprint, while the identified peptides and proteins just saturate.

For these specialized use cases, a better approach may be for the application to operate directly on the SQLite database and load only the data it needs, rather than have `OMSFile` populate an `IdentificationData` structure with filtered data (which may well include meta data etc. the application actually doesn't care about).

timosachsenberg:
So basically that would mean that the data format, as well as basically all tools, won't scale if one has to resort to direct queries on the database. Is this really what we want? This will likely lead to a situation where more and more work is done directly on the database making the ID data structure of less use. 

From experience with idXML files, the major memory issue will be PSMs, which easily reach 100gb or more for such studies of that size. Peptides and proteins have not been a huge issue memory-wise in the files I looked at.
Quick and easy workarounds could be, e.g.:
- passing a filter that skips metadata on the different levels during loading -> might already make a big difference and we don't introduce inconsistencies into the ID data structure
- passing a filter that only extracts an IdentificationData structure with entries associated with a given raw file or list of raw files. (use case: query from sql the list of raw files, then iterate over them and retrieve only information from these. Or in the case of a list: extract all of IDs over all fractions of a sample.). One would need to check how or if this interferes with inference results over multiple files

And more controversial:
- passing a list of PSM references (e.g., referenced from a feature) will subset only these PSMs but and only keep referenced peptides and proteins (should be possible with something like https://www.sqlite.org/lang_with.html). Would then also be consistent regarding the references (but maybe not regarding the inference)

hendrikweisser:
> So basically that would mean that the data format, as well as basically all tools, won't scale if one has to resort to direct queries on the database.

If you have too much data to keep in memory at once, no data structure is going to avoid the need to query selectively directly from the database/file.

> From experience with idXML files, the major memory issue will be PSMs

idXML (and internally, `PeptideIdentification`/`PeptideHit`) is highly inefficient, because every PSM stores the sequence of the matched peptide and the evidence connecting it to proteins (including, in `PeptideHit`, the protein accessions). In `IdentificationData`, all this is replaced by an iterator to the peptide, where the information is stored only once.

All the filters that you mention could be implemented. However, at least in my opinion we want to avoid implementing all these filtering operations twice: Once when filtering an existing `IdentificationData` instance, and once when loading data from SQLite into `IdentificationData`.
(I'm not going to stop you or anyone else from doing this, though.)

timosachsenberg:
> > So basically that would mean that the data format, as well as basically all tools, won't scale if one has to resort to direct queries on the database.
> 
> If you have too much data to keep in memory at once, no data structure is going to avoid the need to query selectively directly from the database/file.

I think I did not make my point clear. My main concern is that you will lose the interface **and** guarantees of the IDDS when people will move to direct manipulation of the database. I also agree that no data structure is going to avoid the need to query selectively directly from the database. That's the reason why I suggested the additions to query selectively while maintaining the guarantees of the ID structure.

> > From experience with idXML files, the major memory issue will be PSMs
> 
> idXML (and internally, `PeptideIdentification`/`PeptideHit`) is highly inefficient, because every PSM stores the sequence of the matched peptide and the evidence connecting it to proteins (including, in `PeptideHit`, the protein accessions). In `IdentificationData`, all this is replaced by an iterator to the peptide, where the information is stored only once.

In the 100gb example, the majority of memory (approx. 10 times for MSGF+) is spent on the meta values and not the sequence. The current (unfiltered) approach will just push the limit by let's say one order of magnitude (because you made it more memory efficient) until it will hit a similar limit. I argue that we probably don't need to scale indefinitely but this is currently not enough.

> All the filters that you mention could be implemented. However, at least in my opinion we want to avoid implementing all these filtering operations twice: Once when filtering an existing `IdentificationData` instance, and once when loading data from SQLite into `IdentificationData`.

I am not sure if we can avoid implementing at least the metadata filter during loading. This might give us another order of magnitude larger studies we can keep in memory.

See 
https://github.com/PRIDE-Archive/pride-metadata-standard/blob/master/experimental-design/examples/PXD010154/sdrf.tsv
for an idea on the size we want to be dealing with (taken from an ELIXIR implementation study together with the guys at the EBI).

hendrikweisser:
> I am not sure if we can avoid implementing at least the metadata filter during loading.

I agree that such a filter would be useful. In general, I think it should be easier to avoid loading certain types of data altogether (e.g. meta data, or PSMs if you just want peptides, etc.) than to load subsets covering all data types.

timosachsenberg:
do you find time to add some of these (e.g.,  without meta data).
You say that loading subsets covering all data types is tricky... dpes this also apply to loading only data associated with a raw/primaryMSRun ?

hendrikweisser:
> do you find time to add some of these (e.g., without meta data).

It's not a priority at the moment. And in general I don't want to develop something just because it "seems like a good idea" without having a use case for it - because it's actually not so straight-forward: Where do you draw the line between data and meta data? Do you need more fine-grained control over exactly what (meta) data you load?

> You say that loading subsets covering all data types is tricky... dpes this also apply to loading only data associated with a raw/primaryMSRun ?

In principle yes, because you want to filter out all peptides, proteins etc. that were not identified in that subset of runs.
Without having tried it, I can't say how tricky it would really be. Maybe it could be solved directly in the database with some smart DELETE statements and "ON DELETE CASCADE" added to the foreign key constraints?

hendrikweisser:
As a proof of concept, integration into MapAlignerIdentification is here:
https://github.com/hendrikweisser/OpenMS/compare/id-file-format...hendrikweisser:mapalignerident-idfile
(Changes are shown relative to the branch used in this pull request.)

Most of the overhead comes from the fact that `IdentificationData` supports multiple scores, so in order to select what the "best" PSM for each spectrum is, or to filter PSMs based on score, there needs to be a selection of which score to use.

timosachsenberg:
I think found a bug in the current IDData design.
It currently needs to declare friends of the class to modify the content:
    friend class IDFilter;
    friend class MapAlignmentTransformer;

hendrikweisser:
> I think found a bug in the current IDData design.
> It currently needs to declare friends of the class to modify the content:

Not a bug, but I agree that it's worth having a discussion about this.
One of my design goals was to ensure data consistency. For example, if a peptide references a protein, that protein needs to be stored in `IdentificationData` before you can store the peptide. Consistency is checked in the `register...` functions, and these are (by design) the only way to store data in `IdentificationData`. This also means that (generally) data that is stored needs to be immutable - otherwise there's no guarantee that changes won't break the consistency.

However, there are of course situations where you do want to change existing data. There are two mechanisms for this:
1. You can re-register an existing data item (e.g. peptide) and any new information should be merged into the existing instance. (The data classes implement `operator+=` for this purpose.) This is safe, but somewhat inefficient. No `friend` declaration needed, though.
2. As you've seen, you can declare `friend` classes that are allowed to modify the data, but should guarantee that consistency is preserved.

If it turns out that these mechanisms are not sufficient, it would be possible to expose additional functions like this: `IdentificationData::modifyX(XRef ref, Functor func)`, for every data type X. You'd need to pass in a reference (iterator) to the item to change, and a functor that gets called with the existing item and updates it (see https://www.boost.org/doc/libs/1_71_0/libs/multi_index/doc/reference/ord_indices.html#modify). The major downside is that consistency checking would be difficult to impossible in this case.
(In principle you could also expose a `replace` function or the whole container as mutable, but then you may as well give up on data consistency altogether.)

My preferred solution at the moment would be to use a combination of friend classes and specific functions in `IdentificationData` itself to modify data (see e.g. `IdentificationData::addScore`) for recurring use cases, and cover niche cases with the re-registering approach. My reasoning is that only certain modifications (e.g. filtering, adding scores or meta values) are used often, while "arbitrary" modifications are actually rare.

timosachsenberg:
Can you define how consistency is defined? Is it just that every reference must exist?
If so, why not expose a modify function that gives full access to the whole datastructure,
takes e.g. a lambda/functor and always calls a checkConsistence_ afterwards?

hendrikweisser:
> Can you define how consistency is defined? Is it just that every reference must exist?

With "consistency" I mean that all the assumptions about "correct" data are satisfied. This means primarily that all references must exist (in the data structure), but also that other constraints are satisfied - e.g. certain fields (score type name, data query ID) can't be empty, coverage values have to be between 0 and 1.

> If so, why not expose a modify function that gives full access to the whole datastructure,
takes e.g. a lambda/functor and always calls a checkConsistence_ afterwards?

It is expensive to check the consistency of the whole data structure if you don't know what could have changed (compared to just checking one item when you register it). I doubt this would be more efficient than re-registering modified data.

timosachsenberg:
I think given that this stuff needs to be in memory it should not take more a fraction of a second to check the constraints you listed.
By the way, is there a plan to add the copy constructor?

hendrikweisser:
> I think given that this stuff needs to be in memory it should not take more a fraction of a second to check the constraints you listed.

Yes, but typically you would make updates to every instance of a data type. So you'd have to be careful to put the whole loop into your "modify" function, or you'd be checking many times. But then all your updates fail if one of them fails.

I'm not saying we can't do this, but I'd rather consider the alternative options (re-registering, friend class, dedicated function) for concrete use cases before doing this.

hendrikweisser:
> By the way, is there a plan to add the copy constructor?

Sure, if we need one. So far that hasn't been the case.

hendrikweisser:
@timosachsenberg:
I think I have a good idea for how to allow updating of data items. We could pass an optional reference to the register function that points to the "old" version of the element:
`IdentificationData::registerX(const X& x, XRef ref = x_container_.end())`
We would do all the usual checks on the "new" element (no code duplication!), and also check that the "key" parts of the "new" element (e.g. peptide sequence, protein accession) are the same as the "old" element. If so, we can replace the element without wasting a look-up, and all references to it will stay intact! (However, there would still be some overhead from copying compared to the `modify` approach.)

timosachsenberg:
Another possibility would be to expose a proxy object through a modify function that has only getters for the key parts and both getter and setter for the other parts.

hendrikweisser:
I just added an output parameter for the database file (.oms) to NucleicAcidSearchEngine (NASE) and did a little bit of testing. The result is that I'm quite concerned about how long it takes to write the output.
In my small example, NASE takes 3.3 seconds to run the analysis and write mzTab (120 KB) and idXML (1.8 MB) output files. It then takes 4.5 minutes (!) to write the SQLite output (1.6 MB). Most of that time (over 3.5 minutes) is spent writing 25,000 rows of peak annotations (c/y ions etc.). I guess it'll be necessary to massively reduce this, e.g. by writing one blob with all the annotations for a spectrum, rather than one row per ion annotation. But even without the peak annotations, it takes at least an order of magnitude longer to write the SQLite output than idXML.
I tested whether enabling foreign key constraints (off by default) or disabling checked constraints (on by default) had an impact on performance, but neither made a noticeable difference.

hroest:
>  In my small example, NASE takes 3.3 seconds to run the analysis and write mzTab (120 KB) and idXML (1.8 MB) output files. It then takes 4.5 minutes (!) to write the SQLite output (1.6 MB). Most of that time (over 3.5 minutes) is spent writing 25,000 rows of peak annotations (c/y ions etc.). I guess it'll be necessary to massively reduce this, e.g. by writing one blob with all the annotations for a spectrum match, rather than one row per ion annotation. But even without the peak annotations, it takes at least an order of magnitude longer to write the SQLite output than idXML.

yes, you definitely want to bundle SQL queries together. One solution is to use 

```
    conn.executeStatement("BEGIN TRANSACTION");
    for (Size i = 0; i < to_osw_output.size(); i++)
    {   
      conn.executeStatement(to_osw_output[i]);
    }   
    conn.executeStatement("END TRANSACTION");
```

or simply creating one very long string which you throw into SQL as we do here https://github.com/OpenMS/OpenMS/blob/1fd89f2ef4532ff4ab3f15d48a23f4ff4abc98d0/src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp#L892

I dont know which is faster, honestly

src/openms/source/FORMAT/OMSFile.cpp
```diff
@@ -0,0 +1,2044 @@
...
+                  ":id, "                                   \
+                  ":cv_term_id, "                           \
+                  ":higher_better)");
+    for (const ID::ScoreType& score_type : id_data_.getScoreTypes())
+    {
+      Key cv_id = storeCVTerm_(score_type.cv_term);
+      query.bindValue(":id", Key(&score_type)); // use address as primary key
+      query.bindValue(":cv_term_id", cv_id);
+      query.bindValue(":higher_better", int(score_type.higher_better));
+      if (!query.exec())
```
hroest: @hendrikweisser you dont want to execute a statement here with every loop iteration, you should be able to  execute once at the end of the loop.

hendrikweisser:
@hroest: Thanks for the hints. I've found [`QSqlQuery::execBatch`](https://doc.qt.io/qt-5/qsqlquery.html#execBatch) and will try that.

timosachsenberg:
idXML and sql file sizes are pretty similar.
Is there a specific reason for that? I had the impression that the new one is more space-efficient.

hendrikweisser:
@timosachsenberg:
Well, the file is still smaller. There are multiple possible reasons for why the difference isn't bigger:
- I don't know what the overhead of encoding data in an SQLite file is compared to XML.
- idXML gets blown up by redundant storage of data, but the example is small, so there isn't much of that.
- `IdentificationData`/the database file can store information that isn't present in idXML (e.g. about applied processing steps).

hroest:
> idXML gets blown up by redundant storage of data, but the example is small, so there isn't much of that.

I dont think you can make any conclusions based on a small dataset, I would go at least for a medium size dataset before making any conclusions. SQL will be much more efficient for pure tabular data (imagine in xml we write "<tag key='keyname' value='6.0'/> to store a single value whereas in SQL you could have a table that simply uses 4-8 bytes to store the value per entry). We do that for sqmass for example to store precursor mz information: each spectrum has a precursor so there is a spot for precursor mz info that takes 8 bytes whereas in XML it will take a full line. If the database is properly normalized, then you should have eliminated all duplicate storage and I am pretty convinced that you should be more space efficient.

hendrikweisser:
Performance testing update:
- Replacing repeated `QSqlQuery::exec` calls with `QSqlQuery::execBatch` didn't make a difference (for storing peak annotations).
- Disabling either synchronous filesystem access or the rollback journal via SQLite pragmas had a huge effect, reducing runtime to a few seconds.
- Disabling both gave the best results (runtime: 1 s).

Adding transactions would have been another way to speed things up, but since we don't need any ACID properties for writing an output file, changing the configuration seems simpler/cleaner.

More information here: https://www.sqlite.org/faq.html#q19

timosachsenberg:
Ok, that's great news! Using transaction is cleaner (aka standard db way to go) and more compatible with other databases so I would prefer that if feasible.

hendrikweisser:
@timosachsenberg: Once you go down that path you have to make non-trivial decisions about what should and shouldn't go into one transaction, and ideally test different options to look at performance impacts. Frankly not a priority at the moment.

timosachsenberg:
why not use one transaction that covers all inserts to the database?

hendrikweisser:
> why not use one transaction that covers all inserts to the database?

I'm pretty sure that's not considered "best practice" in the "standard db way to go". ;-)
(I've seen a lot of arguments for keeping your transactions as small as possible.)

timosachsenberg:
Sure but some of the concerns don't apply here. I am just suggesting that it is likely better to do it via transaction as it otherwise blocks the possibility to use other database backends.

hroest:
I agree with Timo here, this is also what we do elsewhere in the code: simply pack all statements into one transaction and that is also in the code I showed above.

timosachsenberg:
@hendrikweisser I made some small experiments on a feature file format a bachelor student is developing and there I get the same speed using one transaction vs the SQLite specific pragmas.
Using both did not increase the speed further.

hendrikweisser:
Good to know, Timo. I still don't agree that wrapping everything in one transaction is a great idea, though. I think if you want to use transactions, you should use them properly - if not, disable them explicitly.

timosachsenberg:
Why is it not a great idea for our use case - maybe you can elaborate? It speeds things up and keeps compatibility with other databases. What is the advantage of resorting to custom pragmas? Sorry, I don't get it but DBs are not my strongest side. About the other stuff like ACID etc. I don't even care too much right now.

hendrikweisser:
It doesn't actually "keep compatibility with other databases". Transactions add overhead - in exchange for the ACID guarantees (see e.g. here: https://logicalread.com/2015/09/28/optimize-mysql-transaction-perf-part-1-mc13/). They may lock access to tables or rows while they're in effect, which can slow down concurrent connections to the same database. The point is, if we would want to use a "persistent" database (rather than a one-off SQLite file) we would need to think carefully about how we do the transactions. I'd rather be clear that the implementatino is SQLite-only at this point.

timosachsenberg:
Should we close this in favor for the branch you are working on?

hendrikweisser:
> Should we close this in favor for the branch you are working on?

Not sure. It's easier to keep only one branch up-to-date (resolve conflicts), but there are more thematically different changes in the integration branch. Here it's just the new file format.

timosachsenberg:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=433716

timosachsenberg:
TODO: travis CI

timosachsenberg:
I quickly checked but did not find a simple way to add the Qt5 Sql module on Travis.

jpfeuffer:
also needs to be added to external code tests, and probably pyopenms and packaging and knime etc.

jpfeuffer:
rebuild jenkins

jpfeuffer:
randomness errors?

jpfeuffer:
rebuild jenkins

jpfeuffer:
Memo:
```
CMake Warning (dev) at /usr/local/Cellar/cmake/3.17.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (SQLite)
  does not match the name of the calling package (SQLITE).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/modules/FindSQLITE.cmake:201 (find_package_handle_standard_args)
  src/openms/cmake_findExternalLibs.cmake:134 (find_package)
  src/openms/CMakeLists.txt:70 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

jpfeuffer:
@hendrikweisser How can we move forward with this? I fixed all the build system issues. Now there only seem to be platform-dependent problems (hopefully just ordering). Do you have a Linux at hand to double-check? Also, a conflict appeared with your recent changes.

hendrikweisser:
Thanks, @jpfeuffer! I'll have a look at this again soon.

timosachsenberg:
When converting from an idXML file to oms and back, I get
```
Error: Unable to read file (Invalid date-time string in: 0000-00-00 00:00:00)
```
on loading. 
Presumably, because the idXML already contained:
```
<IdentificationRun date="0000-00-00T00:00:00" search_engine="Percolator" search_engine_version="3.02" search_parameters_ref="SP_0" >
```

hendrikweisser:
@timosachsenberg:
Do you have more information about the idXML conversion error? Which step fails? Do you get more details with "-debug"?

Regarding conversion time, we'd need to benchmark. It should be easy to at least distinguish between the time it takes to convert between the internal data formats and the time it takes to generate the SQL.

hendrikweisser:
@timosachsenberg: Thanks. I don't know when I'll have time to look at the problem, so if you could try yourself, that would be great. If not, can you make a minimal example that has this error during date conversion?

timosachsenberg:
this is superseded by a new PR

PR #4390, opened 2019-11-28
fix TICCalculator

### LINKED ISSUE #4389: TICCalculator: wrong results when running multi-threaded
## Context

- OpenMS Version: 2.4.0 Oct 29 2018, 15:59:05, Revision: 9690d06
- Windows 10
- Binary Installer

## Issue

`TICCalculator` returns wrong numbers for *number of peaks* and *total ion current* (TIC) when being called with `-read_method index_parallel -threads N` with N>1. 

Compare:

Numbers are correct when using e.g. `regular` method or `-threads 1`:

```
> TICCalculator -in QX20180816-01.mzML -read_method regular
Read method: regular
[..]
There are 39240 spectra and 23693271 peaks in the input file.
The total ion current is 1.42115e+12
[..]

> TICCalculator -in QX20180816-01.mzML -read_method indexed_parallel -threads 1
[..]
Read method: indexed (parallel)
There are 39240 spectra and 23693271 peaks in the input file.
The total ion current is 1.42115e+12
[..]
```

Numbers are wrong when using `indexed_parallel` method and multiple threads:

```
> TICCalculator -in QX20180816-01.mzML -read_method indexed_parallel -threads 2
[..]
Read method: indexed (parallel)
There are 39240 spectra and 14760886 peaks in the input file.
The total ion current is 1.1904e+12
[..]

> TICCalculator -in QX20180816-01.mzML -read_method indexed_parallel -threads 3
[..]
Read method: indexed (parallel)
There are 39240 spectra and 13019297 peaks in the input file.
The total ion current is 1.12768e+12
[..]

> TICCalculator -in QX20180816-01.mzML -read_method indexed_parallel -threads 4
[..]
Read method: indexed (parallel)
There are 39240 spectra and 11272790 peaks in the input file.
The total ion current is 1.07903e+12
[..]
```

### Observations
- Resulting numbers decrease with increasing numbers of threads
- Numbers seem to be stable for the same number of threads (tested with 10 iterations)

cbielow:
this fixes #4389 (confirmed and fixed on VS2015).
by replacing the `omp critical` with `reduction(+,...`.
ATM I have **no** idea why the critical solution was broken.
The current solution works, as well as switching to `std::atomic` for TIC and nr_peaks.
Still puzzled why critical section breaks (since the final sums are lower, it seems that the addition is not atomic).
Somewhat confirming that this might be a VS bug: g++7.1 does not need this fix. It works as expected in both variants.

### DIFF
--- a/src/utils/TICCalculator.cpp
+++ b/src/utils/TICCalculator.cpp
@@ -283,30 +283,19 @@ protected:
       map.setSkipXMLChecks(true);
 
       double TIC = 0.0;
-      long int nr_peaks = 0;
+      long nr_peaks = 0;
 
       if (load_data)
       {
-
         // firstprivate means that each thread has its own instance of the
         // variable, each copy initialized with the initial value 
-#ifdef _OPENMP
-#pragma omp parallel for firstprivate(map) 
-#endif
+#pragma omp parallel for firstprivate(map) reduction(+: TIC, nr_peaks)
         for (SignedSize i =0; i < (SignedSize)map.getNrSpectra(); i++)
         {
           OpenMS::Interfaces::SpectrumPtr sptr = map.getSpectrumById(i);
-          double nr_peaks_l = sptr->getIntensityArray()->data.size();
-          double TIC_l = std::accumulate(sptr->getIntensityArray()->data.begin(), sptr->getIntensityArray()->data.end(), 0.0);
-#ifdef _OPENMP
-#pragma omp critical (indexed)
-#endif
-          {
-            TIC += TIC_l;
-            nr_peaks += nr_peaks_l;
-          }
+          nr_peaks += sptr->getIntensityArray()->data.size();
+          TIC += std::accumulate(sptr->getIntensityArray()->data.begin(), sptr->getIntensityArray()->data.end(), 0.0);
         }
-
       }
 
       std::cout << "There are " << map.getNrSpectra() << " spectra and " << nr_peaks << " peaks in the input file." << std::endl;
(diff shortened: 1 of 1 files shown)


timosachsenberg:
Strange. And I guess we can be sure that the map has no global state.

cbielow:
> 
> 
> Strange. And I guess we can be sure that the map has no global state.

No, thats fine.
I debugged a bit further and printed all the individual numbers from each iteration using different number of threads. They are all identical, irrespective of thread count, i.e. the threading is not the culprit per se. But summing them up is broken.
Also switching to (3rd solution)
```
          #pragma omp atomic
            TIC += TIC_l;
#pragma omp atomic
            nr_peaks += nr_peaks_l;
```
fixes the code.
Also: removing the `omp critical` gives the same result as when using it. So it seems the pragma is ineffective in this case. However, I know that it works in general in other cases.
Lets call it a weird compiler bug.

elrubio:
First of all, thanks for the quick fix!

I was also trying to wrap my head around the *how* and *why* since I couldn't see any obvious code issues. I found some reports related to the Visual C++ compiler where enabling optimizations breaks the intrinsic *flush* when entering *critical* section. These where related to older compiler and OpenMP versions but might still be lurking around in VS2015. See for example [this post on SO](https://stackoverflow.com/questions/47160261/why-does-my-openmp-2-0-critical-directive-not-flush).

@cbielow Does disabling optimizations fixes the issue for you, too?

Now, from the perspective of a cautious user, this is actually a MAJOR issue since it potentially affects all results coming from multi-threaded code in Windows builds. You will understand that, unless you find measures to address (or rule out) the problem project-wide, I now see myself obligated to *not* trust any results coming from OpenMS on Windows with enabled multi-threading. :-/

In any case, I'm happy to provide further assistance, if needed.

cbielow:
Indeed, after some more testing I can confirm that on Visual Studio 2015:
 - using /Od (no optimization) solves the problem
 - using 
```
#pragma omp flush(TIC, nr_peaks)
#pragma omp critical
```
also solves the problem.
So I can totally understand you err on the side of caution.

jpfeuffer:
What if you specifically specify those variables as shared?

timosachsenberg:
haha that's what I also asked chris on skype ;)

cbielow:
> What if you specifically specify those variables as shared?

does not help. still broken.

cbielow:
can someone try using VS2017?!

Issue #4371, opened 2019-11-20
FeatureFinderMetabo - Using the peak height instead of peak area 


lfnothias:
I want to use the LC-MS peak height instead of the peak area in FeatureFinderMetabo.
There is a "quant _method"=median parameter but it is not clear from the description if this would actually do this ? Anyone has insight in this ?

timosachsenberg:
It is used for direct injection data only where no elution profile is present.
Peak height is not directly supported but it wouldn't be too difficult to add if you need it.

lfnothias:
That would be great ! Note that both methods are commonly used in metabolomics data analysis. In some application, peak height is sometime more robust than peak area, like in HILIC mode.

timosachsenberg:
Probably add new quant method and string around here:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/KERNEL/MassTrace.h#L66
And adapt the MassTrace method to retrieve the "Area"

poshul:
Merged into develop.

PR #4394, opened 2019-12-03
Peak height mass trace

poshul:
Add the ability to quantify mass traces based on the highest peak. Addresses https://github.com/OpenMS/OpenMS/issues/4371

### DIFF
--- a/src/openms/source/KERNEL/MassTrace.cpp
+++ b/src/openms/source/KERNEL/MassTrace.cpp
@@ -39,7 +39,7 @@
 namespace OpenMS
 {
     // must match MassTrace::MT_QUANTMETHOD enum!
-    const std::string MassTrace::names_of_quantmethod[] = {"area", "median"};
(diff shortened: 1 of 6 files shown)


timosachsenberg:
Hi looks good so far. Can you check if / how this integrates into FEatureFinderMetabo. Having a topp test for FFM would be great

poshul:
TOPP test added.

src/openms/source/KERNEL/MassTrace.cpp
```diff
@@ -421,6 +423,8 @@ namespace OpenMS
             return computeFwhmArea();
           case MT_QUANT_MEDIAN:
             return computeMedianIntensity_();
+          case MT_QUANT_HEIGHT:
+            return getMaxIntensity(false);
```
oliveralka: Should that not work with non_smoothed data as well?

poshul:
rebuild jenkins

timosachsenberg:
Thanks! looks good

PR #4397, opened 2019-12-05
[FIX] findExecutable should skip directories in Env with same name

jpfeuffer:
Important on Unices.

### DIFF
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -679,7 +679,7 @@ namespace OpenMS
 
   bool File::findExecutable(OpenMS::String& exe_filename)
   {
-    if (exists(exe_filename)) return true;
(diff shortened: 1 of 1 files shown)


src/openms/source/SYSTEM/File.cpp
```diff
@@ -697,7 +697,7 @@ namespace OpenMS
     {
       for (const String& fn : exe_filenames)
       {
-        if (exists(p + fn))
+        if (exists(p + fn) && !isDirectory(p + fn))
```
cbielow: line https://github.com/OpenMS/OpenMS/pull/4397/files#diff-d39cf05fce87e1511bd1ca97fcee03d6R682
should also get this check.

jpfeuffer: True! thanks

timosachsenberg:
looks good

Issue #4358, opened 2019-11-15
Fix Ascore score calculation and improve performance


AnthonyOfSeattle:
Ascore is getting on in its years but it is still used in some pipelines (our lab's included) and as a baseline for new localization algorithms. So it is nice that OpenMS has a common implementation that can be drawn upon. There are just a couple issues that I would like to address.

Score calculation error:
 - The probability of a match in AScore::calculatePermutationPeptideScores_ and in AScore::compute should be multiplied by 2 * fragment_mass_tolerance to make the scores correct for a specific precision.
https://github.com/OpenMS/OpenMS/blob/3bf32aec5510faed51846a4673d33b7ea8d9cc18/src/openms/source/ANALYSIS/ID/AScore.cpp#L572-L573
https://github.com/OpenMS/OpenMS/blob/3bf32aec5510faed51846a4673d33b7ea8d9cc18/src/openms/source/ANALYSIS/ID/AScore.cpp#L167
 - The AScore::peptideScore_ function should be a weighted average but is not. This is does not affect final results but would be good for consistency.
https://github.com/OpenMS/OpenMS/blob/3bf32aec5510faed51846a4673d33b7ea8d9cc18/src/openms/source/ANALYSIS/ID/AScore.cpp#L383-L397

General enhancements:
 - In the algorithm, AScore::computeCumulativeScore_ computes the cumulative binomial probability every time it is called. This leads to a lot of slow downs, especially for high precision instruments since you spend more time only having a couple of matches (not 0 matches, because this automatically returns 1 and so is fast). We solved this before in our internal version of the algorithm by caching the probability calculations in an unordered_map. If this kind of idea works with the style of the code, it may be a good way to get fairly large speed up (like at least an order of magnitude or more).
 - Purely from a style standpoint, it seems that the ambiguity score portion of the AScore::compute could just be made its own function and AScore::getSpectrumDifference_ might be better placed in the .cpp file

I work a lot with this algorithm and would be happy to put in some pull requests to fix these, if you like.

timosachsenberg:
Sounds great! We would be happy to integrate those fixes via a PR.

AnthonyOfSeattle:
[...]

PR #4403, opened 2019-12-06
Fix AScore Scoring

AnthonyOfSeattle:
This PR addresses #4358 and changes some scoring constants within the AScore class to help it fall more in line with Beausoleil et al. 

**Weighted average:**  
Within the weighted average step of the PepScore calculation, the algorithm was dividing by the wrong constant to make the dot product a weighted average at original line 396. I fixed this to be the true constant and added a couple tests to prove that this is a weighted average.

**Match Probability:**  
The interpretation I always had of Beausoleil et al. was that the probability of a match that you use for calculating the PepScore and AScore is based on the window size for matching. For mass units in mz, I made the probability based on 2 times the mass window. When the mass units are in ppm, the issue for this algorithm is that there is no straight forward way to go from counts of matched peaks to a score that takes into account the difference in window sizes for each theoretical peak. So I mitigate this issue by basing the probability on a very conservative estimate, the largest mz measured for a spectra. Tests are changed to address both situations.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/AScore.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/AScore.h
@@ -167,6 +167,9 @@ namespace OpenMS
     */
     void determineHighestScoringPermutations_(const std::vector<std::vector<double>>& peptide_site_scores, std::vector<ProbablePhosphoSites>& sites, const std::vector<std::vector<Size>>& permutations, std::multimap<double, Size>& ranking) const;
 
+    /// Computes probability for a peak depth of one given spectra and mass_tolerance variables
+    double computeBaseProbability_(double ppm_reference_mz) const;
+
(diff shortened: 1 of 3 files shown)


AnthonyOfSeattle:
Huh, seems when I build from the develop branch, I get a similar `pyopenms_test_memoryleaktests (Failed)` from `make test` as well as other pyOpenMS errors that also seem unrelated to AScore. So I'm guessing this PR isn't introducing anything that would cause the error, correct? `ctest -R AScore_ -V` passes fine on the PR branch.

jpfeuffer:
Don't worry. Memory leak tests for pyOpenMS may fail from time to time. Looks good from the test side! Very nice!

timosachsenberg:
Looks good! Thanks for your contribution. I think this PR is good to go.

PR #4409, opened 2019-12-11
[FEATURE] parallelized PrecursorPurity and spectrum preprocessing in OpenPepXL

enetz:
also includes an internal change in RNPxlSearch to use the new PrecursorPurity class
(most of the changes there are removed trailing whitespaces)

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PrecursorPurity.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PrecursorPurity.h
@@ -65,7 +65,6 @@ namespace OpenMS
     {
       double total_intensity = 0.0;
       double target_intensity = 0.0;
-      double residual_intensity = 0.0;
       double signal_proportion = 0.0;
       Size target_peak_count = 0;
       Size residual_peak_count = 0;
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PrecursorPurity.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PrecursorPurity.h
@@ -78,7 +78,7 @@ namespace OpenMS
      * @param precursor_mass_tolerance The precursor tolerance. Is used for determining the targeted peak and deisotoping.
      * @param precursor_mass_tolerance_unit_ppm The unit of the precursor tolerance
     */
-    static std::vector<PurityScores> computePrecursorPurities(const PeakMap& spectra, double precursor_mass_tolerance, bool precursor_mass_tolerance_unit_ppm);
+    static std::map<String, PurityScores> computePrecursorPurities(const PeakMap& spectra, double precursor_mass_tolerance, bool precursor_mass_tolerance_unit_ppm);
 
     /** @brief compute precursor purity metrics for one MS2 precursor
 
--- a/src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
+++ b/src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
@@ -119,7 +119,6 @@ namespace OpenMS
 
     score.total_intensity = total_intensity;
     score.target_intensity = target_intensity;
-    score.residual_intensity = noise_intensity;
     score.signal_proportion = rel_sig;
     score.target_peak_count = target_peak_count;
     score.residual_peak_count = isolated_window.size();
@@ -132,7 +131,6 @@ namespace OpenMS
     PrecursorPurity::PurityScores score;
     score.total_intensity = score1.total_intensity + score2.total_intensity;
     score.target_intensity = score1.target_intensity + score2.target_intensity;
-    score.residual_intensity = score1.residual_intensity + score2.residual_intensity;
     if (score.target_intensity > 0.0) // otherwise default value of 0 is used
     {
       score.signal_proportion = score.target_intensity / score.total_intensity;
--- a/src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
+++ b/src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
@@ -144,7 +144,7 @@ namespace OpenMS
   {
     std::map<String, PrecursorPurity::PurityScores> purityscores;
 
-    if (spectra[0].getMSLevel() == 2)
+    if (spectra[0].getMSLevel() != 1)
     {
       OPENMS_LOG_WARN << "Warning: Input data not suitable for Precursor Purity computation. Will be skipped!\n";
       return purityscores;
@@ -154,9 +154,10 @@ namespace OpenMS
     int current_parent_index = 0;
     int next_parent_index = 0;
     bool lastMS1(false);
+    int spectra_size = static_cast<int>(spectra.size());
 
 #pragma omp parallel for schedule(guided)
-    for (int i = 0; i < static_cast<int>(spectra.size()); ++i)
+    for (int i = 0; i < spectra_size; ++i)
     {
       // change current parent index if a new MS1 spectrum is reached
       if (spectra[i].getMSLevel() == 1)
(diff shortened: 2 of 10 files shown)


src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp
```diff
@@ -120,15 +127,11 @@ namespace OpenMS
...
-    NLargest nlargest_filter = NLargest(500);
-
     PeakMap filtered_spectra;
-    Size MS2_counter(0);
 
-    // TODO does not work multithreaded because of the MS2_counter, find an alternative or keep single threaded
-// #ifdef _OPENMP
-// #pragma omp parallel for
-// #endif
+#ifdef _OPENMP
```
cbielow: no need to ifdef a pragma. Only omp-functions require this.

src/openms/source/ANALYSIS/XLMS/OPXLSpectrumProcessingAlgorithms.cpp
```diff
@@ -183,30 +177,21 @@ namespace OpenMS
         if (deisotoped.size() > peptide_min_size * 2 || labeled)
         {
           window_mower_filter.filterPeakSpectrum(deisotoped);
-          nlargest_filter.filterPeakSpectrum(deisotoped);
           deisotoped.sortByPosition();
 
 #ifdef _OPENMP
-#pragma omp critical
+#pragma omp critical (filtered_spectra_access)
```
cbielow: no ifdef here

src/openms/include/OpenMS/ANALYSIS/ID/PrecursorPurity.h
```diff
@@ -78,7 +78,7 @@ namespace OpenMS
      * @param precursor_mass_tolerance The precursor tolerance. Is used for determining the targeted peak and deisotoping.
      * @param precursor_mass_tolerance_unit_ppm The unit of the precursor tolerance
     */
-    static std::vector<PurityScores> computePrecursorPurities(const PeakMap& spectra, double precursor_mass_tolerance, bool precursor_mass_tolerance_unit_ppm);
+    static std::map<String, PurityScores> computePrecursorPurities(const PeakMap& spectra, double precursor_mass_tolerance, bool precursor_mass_tolerance_unit_ppm);
```
cbielow: instead of switching to a map (slow), could you also preallocate  the vector outside the parallel region and use `operator[]` to access for each thread. This means that indices for spectra which are skipped over remain untouched. If this is not a problem, then you can get rid of the `critical` section (which might be broken on Windows...)

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -36,6 +36,13 @@
 #include <OpenMS/CONCEPT/Constants.h>
 #include <OpenMS/CONCEPT/LogStream.h>
 
+#ifdef _OPENMP
+#include <omp.h>
+#define NUMBER_OF_THREADS (omp_get_num_threads())
```
cbielow: Is this ever used? Using a macro is code smell IMHO...

src/openms/include/OpenMS/ANALYSIS/ID/PrecursorPurity.h
enetz: @cbielow one of the issues I want to avoid with this, is the dependency between the order of the spectra in the peak map and this list The peak map is processed further in my tool and some of the spectra get filtered out. Using a map is much simpler and less error prone, than what I did before: saving the indices of the removed spectra and then removing the same ones from this list. Also this map only has the same number of items as there are MS2 spectra in the peak map and won't become a performance bottleneck anytime soon. The parallelization of PrecursorPurity is just a bonus that I added, because it was simple to do now. Using a map makes handling this class much easier and made the parallelization of my spectrum deisotoping and filtering also simpler.

src/utils/RNPxlSearch.cpp
```diff
@@ -1391,7 +1391,15 @@ class RNPxlSearch :
...
           {
-            ph.setMetaValue("precursor_purity", purities[scan_index].signal_proportion);
+            map<String, PrecursorPurity::PurityScores>::const_iterator it = purities.find(exp[scan_index].getNativeID());
+            if (it != purities.end())
+            {
+              ph.setMetaValue("precursor_purity", it->second.signal_proportion);
+            }
+            else // probably not necessary
+            {
+              ph.setMetaValue("precursor_purity", 0);
```
timosachsenberg: would assuming a perfect purity (1.0?) be better?

enetz: I don't think so. This is the case where we don't have any information about the precursor. 1 would  imply prefect purity and happens quite often with real data, while 0 would be extremely rare and a weird glitch in a case where we have data. I would just write the code using these values in a way that ignores values of 0.

timosachsenberg:
Regarding PrecursorPurity:
it seems that some of the returned values are redundant / can be easily calculated from others.
Would it make more sense to keep this minimal or should we not bother here as this will likely never be a memory issue?

enetz:
Two of the six saved parameters are redundant and can be computed from the others. I think two additional `double` values per MS2 spectrum should not cause an issue for any dataset. `score.residual_intensity` is the least important and I would not mind removing that.

enetz:
this can be merged, if all tests succeed

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -164,23 +140,23 @@ namespace OpenMS
...
 
-  std::vector<PrecursorPurity::PurityScores> PrecursorPurity::computePrecursorPurities(const PeakMap& spectra, double precursor_mass_tolerance, bool precursor_mass_tolerance_unit_ppm)
+  std::map<String, PrecursorPurity::PurityScores> PrecursorPurity::computePrecursorPurities(const PeakMap& spectra, double precursor_mass_tolerance, bool precursor_mass_tolerance_unit_ppm)
   {
-    std::vector<PrecursorPurity::PurityScores> purityscores;
+    std::map<String, PrecursorPurity::PurityScores> purityscores;
 
-    // TODO throw an exception or at least a warning?
-    // if there is no MS1 before the first MS2, the spectra datastructure is not suitable for this function
     if (spectra[0].getMSLevel() == 2)
```
timosachsenberg: maybe extend test

enetz:
The Ubuntu 1804 test that failed previously now succeeds,
the Windows 10 and highsierra tests show a Java error message, so I assume that is a build system failure,
the pyopenms test has a memory leak.
All these tests had already passed a few commits ago.

Would it make sense to restart the jenkins builds? They haven't all finished without build system failures even once today.

timosachsenberg:
rebuild jenkins

timosachsenberg:
This will produce strange results if there are non unique native ids (e.g., empty native ids).
Can you add at least a precondition that checks if the precursor purity has already been calculated?

enetz:
@timosachsenberg what would be the best option to do, if it already has been calculated? only keep the old value, or replace it with the new one?

enetz:
@timosachsenberg I am using the `[]` operator to add new scores. If the key already exists, then its value will be overwritten by the new value. So the last value for each native ID is kept. I could check for empty native IDs and skip a spectra that do not have it.

timosachsenberg:
I would probably fail in that case as it would lead to unexpected behavior

enetz:
Alright, I will fail it with a Warning when:
- the first spectrum does not contain a native ID (if one is missing, there is probably something else wrong with the data as well)
- after I find the first case of a duplicated ID

enetz:
PyOpenMS memory leak and a fail of the build system. All failing tests since I fixed the python bindings a few commits ago have been unrelated.
And I think we made it robust enough for the applications that currently use the PrecursorPurity class.
I think this is ready to merge.

PR #4413, opened 2019-12-12
Feature/parallel tagger

enetz:
another version of #4263

I addressed all of the small code efficiency and style suggestions from that PR.
Additionally I parallellized the tagging and filtering. Python bindings and tests are also included.
Another feature included here is a simplified and more efficient enumeration of peptide pair candidates for an MS2 spectrum. Splitting that into two PRs would cause a major merge conflict afterwards, so I think this way is simpler.

Previously I had to call `AASequence::toUnmodifiedString` for every candidate sequence, but I changed some parts of the candidate enumeration to use the function once and then keep and propagate the sequence as a `String`. Now I don't have to generate or even copy the candidate sequence Strings. This made the filtering much faster.

After comparing the simple, brute force approach of using `String::hasSubstring` with using AhoCorasick for the string matching, it turned out the first, simple approach is at least an order of magnitude faster. It seems the overhead of passing longer sequences through a very shallow trie (I only use sequence tags of length up to 3) makes it very inefficient in this case.

The tagger now, using a tag length of 2, halves the search space for OpenPepXL, while the sensitivity stays roughly the same. Since many target and decoy sequences are filtered out, it is difficult to estimate how the FDR will change for any one dataset, so you may get less or more hits depending on the searched database. The really confident high scoring hits are always identified with the Tagger as well.

Some comments from the previous PR, that I did not address yet:
- I and L handling: the tagger still generates tags with all combinations of I and L. This is done in a naive and brute force way and causes some searches to be repeated.
- There are much more efficient ways to do the filtering as well.

The runtime for tagging and filtering could still be further improved, but is already negligible compared to generating theoretical spectra. Although the AhoCorasick filtering approach is an order of magnitude slower, which filtering function I use in OpenPepXL does not make a difference for the total tool runtime.

For now the explicit generation of all I/L combinations combined with the simple `hasSubstring` approach is simple to use and works well enough.
I think this is a prototype that works well and a good starting point for future optimization.


src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,78 @@
...
+      // initalize tagger with minimum/maximum tag length, +- tolerance ppm, min/max charge, fixed and variable modifications
+      Tagger(size_t min_tag_length, double ppm, size_t max_tag_length = 65535, size_t min_charge = 1, size_t max_charge = 1, const StringList& fixed_mods = StringList(), const StringList& var_mods = StringList());
+
+      // generate tags from mass vector @p mzs using the standard residues in ResidueDB
+      void getTag(const std::vector<double>& mzs, std::vector<std::string>& tags) const;
+      void getTag(const MSSpectrum& spec, std::vector<std::string>& tags) const;
+      void setMaxCharge(size_t max_charge);
+
+    private:
+      double min_gap_; // will be set to smallest residue mass in ResidueDB
```
hroest:
```suggestion
      double min_gap_; //< will be set to smallest residue mass in ResidueDB
```
please change docu style to default style everywhere (doxygen) -- see https://github.com/OpenMS/OpenMS/wiki/Coding-conventions#doxygen

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,78 @@
...
+    *
+    *  Current restrictions and potential extensions:
+    *  - first prefix/suffix ion don't give rise to first character in tags (as currently ion types are not specified)
+    *  - gaps are not supported
+    **/
+  class OPENMS_DLLAPI Tagger
+  {
+    public:
+      // initalize tagger with minimum/maximum tag length, +- tolerance ppm, min/max charge, fixed and variable modifications
+      Tagger(size_t min_tag_length, double ppm, size_t max_tag_length = 65535, size_t min_charge = 1, size_t max_charge = 1, const StringList& fixed_mods = StringList(), const StringList& var_mods = StringList());
```
hroest: please add full documentation here: https://github.com/OpenMS/OpenMS/wiki/Coding-conventions#doxygen

CMakeLists.txt
```diff
@@ -221,6 +221,13 @@ option(ENABLE_SIRIUS_TEST "Enable test for fingerid (SiriusAdapter)" ON)
 #------------------------------------------------------------------------------
 option(ENABLE_FINGERID_TEST "Enable test for fingerid (SiriusAdapter)" OFF)
 
+#------------------------------------------------------------------------------
```
cbielow: better place for this would be `cmake/compiler_flags.cmake`

src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -350,6 +356,20 @@ using namespace OpenMS;
     specGenParams_mainscore.setValue("add_k_linked_ions", "true");
     specGen_mainscore.setParameters(specGenParams_mainscore);
 
+    // use a less stringent tolerance for the tagger
+    double tagger_tol;
+    if (fragment_mass_tolerance_unit_ppm_) // ppm: increase tolerance by 50%
+    {
+      tagger_tol = fragment_mass_tolerance_xlinks_ + (fragment_mass_tolerance_xlinks_ / 2.0);
```
cbielow: how about `fragment_mass_tolerance_xlinks_  * 1.5` ?

src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -350,6 +356,20 @@ using namespace OpenMS;
...
+    }
+    else // dalton: turn into ppm at 500 mz
+    {
+      tagger_tol = fragment_mass_tolerance_xlinks_ / 1e-6 / 500;
```
cbielow: dividing by 1e-6 is somewhat confusing.
How about using `Math::getPPM(500, 500+fragment_mass_tolerance_xlinks_)` ?

src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -350,6 +356,20 @@ using namespace OpenMS;
...
+    }
+
+    Tagger tagger = Tagger(sequence_tag_min_length_, tagger_tol, sequence_tag_min_length_, 1, max_precursor_charge_, fixedModNames_, varModNames_);
```
cbielow: why construct and then assign?

src/openms/source/ANALYSIS/XLMS/OpenPepXLLFAlgorithm.cpp
```diff
@@ -404,19 +424,28 @@ using namespace OpenMS;
...
       OPENMS_LOG_DEBUG << "Size of enumerated candidates: " << double(cross_link_candidates.size()) * sizeof(OPXLDataStructs::ProteinProteinCrossLink) / 1024.0 / 1024.0 << " mb" << endl;
 #endif
 
-
-      spectrum_counter++;
-      cout << "Processing spectrum " << spectrum_counter << " / " << spectra.size() << " |\tSpectrum index: " << scan_index << "\t| at: " << DateTime::now().getTime() << endl;
+      cout << "Processing spectrum " << spectrum_counter << " / " << spectra.size() << " |\tSpectrum ID: " << spectrum.getNativeID() << "\t| at: " << DateTime::now().getTime() << endl;
       cout << "Number of peaks: " << spectrum.size() << " |\tNumber of candidates: " << cross_link_candidates.size() << endl;
 
+      if (cross_link_candidates.size() < 1)
```
cbielow: == empty() ?

src/tests/class_tests/openms/source/Tagger_test.cpp
```diff
@@ -0,0 +1,262 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2018.
```
hroest:
```suggestion
// ETH Zurich, and Freie Universitaet Berlin 2002-2019.
```

can you fix that everywhere?

CMakeLists.txt
enetz: oh, I didn't mean to include this change here. I will remove that.

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,78 @@
...
+
+    private:
+      double min_gap_; // will be set to smallest residue mass in ResidueDB
+      double max_gap_; // will be set to highest residue mass in ResidueDB
+      double ppm_; // < tolerance
+      size_t min_tag_length_; // < minimum tag length
+      size_t max_tag_length_; // < maximum tag length
+      size_t min_charge_; // < minimal fragment charge
+      size_t max_charge_; // < maximal fragment charge
+      std::map<double, char> mass2aa;
```
cbielow: should be `mass2aa_`

src/openms/include/OpenMS/CHEMISTRY/Tagger.h
```diff
@@ -0,0 +1,78 @@
...
+// --------------------------------------------------------------------------
+// $Maintainer: Eugen Netz $
+// $Authors: Timo Sachsenberg, Eugen Netz $
+// --------------------------------------------------------------------------
+
+#pragma once
+
+#include <OpenMS/CONCEPT/Types.h>
+#include <OpenMS/OpenMSConfig.h>
+#include <OpenMS/KERNEL/MSSpectrum.h>
```
cbielow: no need to include MSSpectrum. Just forward declare it.

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+namespace OpenMS
+{
+  char Tagger::getAAByMass_(double m) const
+  {
+    // fast check for border cases
+    if (m < min_gap_ || m > max_gap_) return ' ';
+
+    const double delta = Math::ppmToMass(ppm_, m);
+    auto left = mass2aa.lower_bound(m - delta);
```
cbielow: does the border case above guarantee that this call never returns `mass2aa.end()`?

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,195 @@
...
+        getTag_(tag, mzs, j, tags, charge);
+      }
+      tag.pop_back();  // remove last string
+      ++j;
+    }
+  }
+
+  Tagger::Tagger(size_t min_tag_length, double ppm, size_t max_tag_length, size_t min_charge, size_t max_charge, const StringList& fixed_mods, const StringList& var_mods)
+  {
+    ppm_ = ppm;
```
cbielow: prefer initializer list for this initialization

src/openms/source/CHEMISTRY/Tagger.cpp
enetz: `max_gap_` is the mass of the heaviest residue, so the last element in `mass2aa`. `lower_bound` is used with `m - delta` and `m` is guaranteed to be `<= max_gap_`. So as long as `delta` is a positive number, it will always at least return the last element, not `mass2aa.end()`. I guess it is possible to cause an error by using a negative ppm tolerance. Should I check for that? I could use abs(ppm) in the constructor.

src/openms/source/CHEMISTRY/Tagger.cpp
```diff
@@ -0,0 +1,195 @@
...
+    min_gap_ = mass2aa.begin()->first - Math::ppmToMass(ppm, mass2aa.begin()->first);
+    max_gap_ = mass2aa.rbegin()->first + Math::ppmToMass(ppm, mass2aa.rbegin()->first);
+  }
+
+  void Tagger::getTag(const std::vector<double>& mzs, std::vector<std::string>& tags) const
+  {
+    // start peak
+    if (min_tag_length_ > mzs.size()) return; // avoid segfault
+
+#pragma omp parallel for schedule(guided)
```
cbielow: using `omp critical` within a `getTag_(...)` is somewhat unexpected (and not neccessary).
You can simply create a separate vector for each thread and join them later.
This is cleaner and potentially also faster (more scalable in any case).

```
#pragma omp parallel
{
  std::vector<std::string> tags_local;
  #pragma omp for schedule(guided)
    ....
    getTag_(tags_local, ...)
    ...
  #pragma omp critical (join_tags)
  {
    tags.insert(tags.end(), tags_local.begin(), tags_local.end());
  }
}
  

}
```

src/tests/class_tests/openms/source/OPXLHelper_test.cpp
```diff
@@ -485,4 +488,50 @@ START_SECTION(static void OPXLHelper::isoPeakMeans(OPXLDataStructs::CrossLinkSpe
...
 END_SECTION
 
+START_SECTION(filterPrecursorsByTags(std::vector <OPXLDataStructs::XLPrecursor>& candidates, std::vector<std::string>& tags))
+
+  std::cout << std::endl;
+  std::vector< int > spectrum_precursor_correction_positions;
+  std::vector<OPXLDataStructs::XLPrecursor> precursors = OPXLHelper::enumerateCrossLinksAndMasses(peptides, cross_link_mass, cross_link_mass_mono_link, cross_link_residue1, cross_link_residue2, spectrum_precursors, spectrum_precursor_correction_positions, precursor_mass_tolerance, precursor_mass_tolerance_unit_ppm);
+
+  // set of tags
+  std::vector<std::string> tags;
```
cbielow: `std::vector<std::string> tags = { "DE", "PP", .... };`

src/openms/source/CHEMISTRY/Tagger.cpp
cbielow: yes, that would be perfect

src/tests/class_tests/openms/source/Tagger_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+  TEST_EQUAL(std::find(tags.begin(), tags.end(), "HISTA") != tags.end(), true)
+  TEST_EQUAL(std::find(tags.begin(), tags.end(), "ISTAG") != tags.end(), true)
+  TEST_EQUAL(std::find(tags.begin(), tags.end(), "STAGG") != tags.end(), true)
+  TEST_EQUAL(std::find(tags.begin(), tags.end(), "TAGGE") != tags.end(), true)
+  TEST_EQUAL(std::find(tags.begin(), tags.end(), "AGGER") != tags.end(), true)
+  TEST_EQUAL(std::find(tags.begin(), tags.end(), "GGER") != tags.end(), true)
+  TEST_EQUAL(std::find(tags.begin(), tags.end(), "GER") != tags.end(), true)
+
+  // tagger searching for charge +2 with variable modification
+  Tagger tagger5 = Tagger(2, 10, 5, 2, 2, StringList(), ListUtils::create<String>("Oxidation (D)"));
```
timosachsenberg: would it make sense to distinguish "unmodified" tags as well as 
modified ones? There are probably several use cases where one might 
want to distinguish the Tag: PEPD from PEPD(Oxidiation)

src/tests/class_tests/openms/source/Tagger_test.cpp
enetz: do we have a script to do this globally? We can change that to 2020 in a few weeks again anyway.

src/tests/class_tests/openms/source/Tagger_test.cpp
enetz: This would be easy to do, I would need Strings for each residue, e.g. "D" and "D(Oxidation)" and StringLists for the list of strings. The filtering would also change from 2 or 3 letters to matching strings of an undefined length. I don't think I can squeeze both of them into the same class without copying much of the code

timosachsenberg: yeah probably one would then use Residue* instead of char to be fully generic.

enetz:
The MSVC-15 build error seems to be build machine related: `ERROR: [WS-CLEANUP] Cannot delete workspace: Unable to delete 'c:\jenkins\ws\openms_pr_ci\979a8a2f'.`
Not sure about the g++-7 ubuntu test.

PR #4415, opened 2019-12-13
[FEATURE] OpenPepXL score and runtime optimization

enetz:
- Fixed a numerical issue with the scores in `XQuestScores`.
- Removed a few peak types from OpenPepXL theoretical spectra. These were determined to be mostly useless after further analysis.
- Optimization of `SimpleTSGXLMS`, the main bottleneck for OpenPepXL runtime
- Some additional small fixes, e.g. using references in a comparator function

I tested several sorting algorithms for my theoretical spectra, and `boost::pdqsort` turned out to be the most efficient. My spectra have an organ shape, many sections are already partially sorted in descending order. Reversing the spectra also speeds up the sorting by a bit. I also reordered the adding of peaks a bit to make the pre-sorted-ness of the spectra more prevalent. 

`boost::pdqsort` is not available in some older boost versions, that we still support. Is it possible to use it, when available and fall back to std::sort when it is not?


src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLDataStructs.h
```diff
@@ -184,7 +184,7 @@ namespace OpenMS
        */
       struct CLSMScoreComparator
       {
-        bool operator() (const CrossLinkSpectrumMatch& a, const CrossLinkSpectrumMatch& b)
+        inline bool operator() (const CrossLinkSpectrumMatch& a, const CrossLinkSpectrumMatch& b)
```
cbielow: no need to specify `inline` here, since you provide the implementation within the class declaration (implicit inline)

cbielow: see https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
```diff
@@ -184,7 +184,8 @@ namespace OpenMS
       }
     }
 
-    std::sort(spectrum.begin(), spectrum.end(), SimpleTSGXLMS::SimplePeakComparator());
+    boost::sort::pdqsort(spectrum.begin(), spectrum.end(), SimpleTSGXLMS::SimplePeakComparator());
```
cbielow: would a lambda function also work? potentially faster...

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
```diff
@@ -404,77 +408,55 @@ namespace OpenMS
...
   {
-    if (pos < 0)
-    {
-      return;
-    }
-    SimpleTSGXLMS::SimplePeak p;
-    p.mz = pos;
-    p.charge = charge;
-    spectrum.push_back(p);
+    spectrum.resize(spectrum.size() + 1);
```
cbielow: `spectrum.emplace_back(pos, charge);`?

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
```diff
@@ -404,77 +408,55 @@ namespace OpenMS
...
     double mono_pos = precursor_mass + (Constants::PROTON_MASS_U * static_cast<double>(charge));
-    p.mz = mono_pos / static_cast<double>(charge);
-    spectrum.push_back(p);
-
     if (add_isotopes_ && max_isotope_ >= 2) // add second isotopic peak with fast method, if two or more peaks are asked for
     {
-      double pos = mono_pos + (Constants::C13C12_MASSDIFF_U / static_cast<double>(charge));
-      p.mz = pos;
-      spectrum.push_back(p);
+      addPeak_(spectrum, (mono_pos + Constants::C13C12_MASSDIFF_U) / static_cast<double>(charge), charge);
```
cbielow: explicit casting to double is not required if one of the operands is a double already IMHO. Is there a warning otherwise?

cbielow:
> `boost::pdqsort` is not available in some older boost versions, that we still support. Is it possible to use it, when available and fall back to std::sort when it is not?

Some parts of OpenMS use
`#if OPENMS_BOOST_VERSION_MINOR >= 64 ...` ...
You should also check for MAJOR ==1 version IMHO, because it breaks with Boost 2.0 otherwise...

src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLDataStructs.h
enetz: does it hurt to make the intent explicit? in case these functions end up in the cpp file after future refactoring.

enetz:
The failing tests seem to be unrelated. XTandemAdapter on MSVC-15 and the other failure seems build machine related (e.g. `cp: cannot stat '/home/jenkins/ws/openms_pr_ci/a9154670/source/THIRDPARTY/All/*': No such file or directory`)
There are no changes related to THIRDPARTY or XTandemAdapter in this PR.

src/openms/include/OpenMS/ANALYSIS/XLMS/OPXLDataStructs.h
timosachsenberg: https://stackoverflow.com/questions/5808498/function-inlining-what-are-examples-where-it-hurt-performance

enetz: I removed the redundant inlines and some from functions that have more than two lines of code. Did not affect the runtime of my benchmark test, so it should be fine this way.

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
```diff
@@ -321,7 +326,13 @@ namespace OpenMS
       addPrecursorPeaks_(spectrum, precursor_mass, maxcharge);
     }
 
-    std::sort(spectrum.begin(), spectrum.end(), SimpleTSGXLMS::SimplePeakComparator());
+#ifdef OPENMS_USE_PDQSORT
+    std::reverse(spectrum.begin(), spectrum.end());
+    boost::sort::pdqsort(spectrum.begin(), spectrum.end(), [](const SimplePeak& a, const SimplePeak& b) {return a.mz < b.mz;});
```
timosachsenberg: would spectrum.rbegin() work instead of using std::reverse?

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
```diff
@@ -631,7 +589,13 @@ namespace OpenMS
       addPrecursorPeaks_(spectrum, precursor_mass, maxcharge);
     }
 
-    std::sort(spectrum.begin(), spectrum.end(), SimpleTSGXLMS::SimplePeakComparator());
+#ifdef OPENMS_USE_PDQSORT
+    std::reverse(spectrum.begin(), spectrum.end());
```
timosachsenberg: same here - reverse iterator

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
enetz: then I would sort them in reverse order, descending, and would have to reverse them again afterwards. although it might actually work if I reverse the comparator...  I will try something

timosachsenberg: yep comp needs to be reversed

enetz:
@jpfeuffer the OpenMS build system defines some boost modules that we use, e.g. in `OpenMS/contrib/libraries.cmake/boost.cmake`
I used a function from boost/sort here, which is not included there. Will that cause problems in the future?
Boost/sort is also not available on older versions of boost that we support, so I added this:
```
#if OPENMS_BOOST_VERSION_MINOR >= 67 && OPENMS_BOOST_VERSION_MAJOR == 1
#define OPENMS_USE_PDQSORT
#include <boost/sort/pdqsort/pdqsort.hpp>
#endif
```

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
cbielow: since the comparator is branchless, calling `boost::sort::pdqsort_branchless` should help as well.
https://www.boost.org/doc/libs/1_67_0/libs/sort/doc/html/sort/single_thread/pdqsort/pdqsort_usage.html

jpfeuffer:
Depends. Is Boost::sort a compiled library module or just another header-only part of certain boost versions?

enetz:
@jpfeuffer since I did not change the build system and it works without building Boost::sort, I guess it must be header-only

jpfeuffer:
Then I think your define should be enough.

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
enetz: ok, it works, but for some reason is slower. benchmark went from 34s to 43s by using reverse iterators and a reversed comparator. I think using rbegin and rend with a reversed comparator essentially does the same operation, just starting from a different end. So the pre-sortedness of the spectra still only helps, if you reverse it first.

src/openms/source/CHEMISTRY/SimpleTSGXLMS.cpp
enetz: using a benchmark run that takes about 35s, I don't see a difference using the branchless version. If there is one, it is smaller than the difference between repeat trials. It reached a lower minimal runtime by about 0.3s after 5 trials, but on average the difference is very small.
But the doc says it will not be slower than normal pdqsort in my case, so I will keep it in.

enetz:
I tested out and implemented all the optimization suggestions that worked. All failing tests since a few commits ago were unrelated to the changes here.
I also tested it for thread safety.
I think this is ready to be merged.

PR #4416, opened 2019-12-15
FIX: filename and sample name mismatch in AbsoluteQuantitationStandards

dmccloskey:
### Bug
- "sample names" (i.e., filenames) obtained from `getPrimaryMSRunPath` sometimes do not match the sample name provided in the StandardsConcentration as a result of ".mzML" or ".txt" suffixes added to the filenames reported by `getPrimaryMSRunPath`

### Fixes
- pre-emptively remove all ".mzML" and ".txt" substrings prior to comparing sample names
- modified the tests to test for feature::primaryMSRunPaths with ".mzML" and ".txt" suffixes


src/openms/source/METADATA/AbsoluteQuantitationStandards.cpp
```diff
@@ -71,9 +71,11 @@ namespace OpenMS
...
       for (const FeatureMap& fmap : feature_maps) // not all elements are necessarily processed (break; is present inside the loop)
       {
-        StringList sample_name;
-        fmap.getPrimaryMSRunPath(sample_name);
-        if (!sample_name.size() || sample_name[0] != run.sample_name) // if the FeatureMap doesn't have a sample_name, or if it is not the one we're looking for: skip.
+        StringList filename;
+        fmap.getPrimaryMSRunPath(filename);
+        String sample_name_1 = filename[0].substitute(".mzML", "");
+        String sample_name_2 = filename[0].substitute(".txt", "");
+        if (!filename.size() || !(sample_name_1 == run.sample_name || sample_name_2 == run.sample_name)) // if the FeatureMap doesn't have a sample_name, or if it is not the one we're looking for: skip.
```
cbielow: Better imho: filename.empty() ...

src/openms/source/METADATA/AbsoluteQuantitationStandards.cpp
```diff
@@ -71,9 +71,11 @@ namespace OpenMS
       }
       for (const FeatureMap& fmap : feature_maps) // not all elements are necessarily processed (break; is present inside the loop)
       {
-        StringList sample_name;
-        fmap.getPrimaryMSRunPath(sample_name);
-        if (!sample_name.size() || sample_name[0] != run.sample_name) // if the FeatureMap doesn't have a sample_name, or if it is not the one we're looking for: skip.
+        StringList filename;
+        fmap.getPrimaryMSRunPath(filename);
+        String sample_name_1 = filename[0].substitute(".mzML", "");
```
cbielow: Global replace is dangerous? Maybe just check suffix?

timosachsenberg:
some .vs stuff slipped in

src/openms/source/METADATA/AbsoluteQuantitationStandards.cpp
dmccloskey: I like that.  Is there a way to easily suffix?

timosachsenberg:
You could use something like: 
if (filename.hasSuffix(".mzML")) filename.resize(filename.size() - 5)

dmccloskey:
Any idea what went wrong with the nightly?

jpfeuffer:
Dont worry about this. It's just that people prioritize fixing nightlies before merging new PRs that introduce more stuff. It's only packaging issues in the nightlies. Code is fine.

PR #4418, opened 2019-12-16
new feature: MapAlignerTreeGuided

jthueringer:
MapAlignerTreeGuided without tests


src/tests/class_tests/openms/data/MzMLFile_1_test.tmp
```diff
@@ -0,0 +1,522 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+<indexedmzML xmlns="http://psi.hupo.org/ms/mzml" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://psi.hupo.org/ms/mzml http://psidev.info/files/ms/mzML/xsd/mzML1.1.0_idx.xsd">
```
hroest: @jthueringer do you need this file? seems like its in here by accident?

hroest:
Thanks a lot for your contribution. Since this is a new tool (or two new tools), could you please write black box / regression tests as described here: https://github.com/OpenMS/OpenMS/wiki/Write-tests (TOPP Tests). This will allow you to verify that your tool functions as you expect it to work and helps us to avoid regressions in future changes to the code.

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,580 @@
...
+        </tr>
+        <tr>
+            <td VALIGN="middle" ALIGN = "center" ROWSPAN=2> @ref TOPP_FeatureLinkerUnlabeledKD or @n @ref TOPP_FeatureLinkerUnlabeledQT </td>
+        </tr>
+    </table>
+</CENTER>
+
+    This tool provides an algorithm to align the retention time scales of multiple input files, correcting shifts and distortions between them. Retention time adjustment may be necessary to correct for chromatography differences e.g. before data from multiple LC-MS runs can be combined (feature grouping), or when one run should be annotated with peptide identifications obtained in a different run.
+
+    All map alignment tools (MapAligner...) collect retention time data from the input files and - by fitting a model to this data - compute transformations that map all runs to a common retention time scale. They can apply the transformations right away and return output files with aligned time scales (parameter @p out), and/or return descriptions of the transformations in trafoXML format (parameter @p trafo_out). Transformations stored as trafoXML can be applied to arbitrary files with the @ref TOPP_MapRTTransformer tool.
```
hroest: `map all runs to a common retention time scale` -> Is that still true here?

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,580 @@
...
+        ++union_size;
+      }
+      Size intercept_size = intercept_rts1.size();
+
+      // pearsonCorrelationCoefficient(rt_map_i, rt_map_j)
+      float pearson_val;
+      pearson_val = static_cast<float>(pearsonCorrelationCoefficient(intercept_rts1.begin(), intercept_rts1.end(),
+              intercept_rts2.begin(), intercept_rts2.end()));
+      if (pearson_val > 1)
+        throw Exception::InvalidRange(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
```
hroest:
```suggestion
      {
        throw Exception::InvalidRange(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
      }
```

src/topp/SequenceRemover.cpp
```diff
@@ -0,0 +1,135 @@
...
+using namespace OpenMS;
+using namespace std;
+using namespace boost;
+
+//-------------------------------------------------------------
+//Doxygen docu
+//-------------------------------------------------------------
+
+/**
+    @page TOPP_SequenceRemover SequenceRemover
```
hroest: can you give some motivation when this would be useful? maybe move this to the UTIL folder instead of TOOL ?

src/topp/MapAlignerTreeGuided.cpp
jthueringer: It's still true: It uses the methods align() and computeTransformation() of the MapAlignmentAlgorithmIdentification class, but determines the order of the alignments of multiple maps based on distances between them and is therefore tree-guided. I will add this to the documentation and also write tests in January.

src/topp/SequenceRemover.cpp
timosachsenberg: The functionality could be move to FileFilter if we want to keep it in a tool.
Alternatively, this could be done with pyopenms in a few lines of code - would not require a full tool

jthueringer: I've used it to calculate recall and precision of the aligners and remove it because I don't think it needs to be integrated.

.gitignore
```diff
@@ -26,4 +26,6 @@ openms-build
 *.swp
 cmake-build-debug
 cmake-build-release
-
+src/topp/MapAlignerTree.cpp
```
cbielow: please undo changes to this file

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+    <B>The command line parameters of this tool are:</B> @n
+    @verbinclude TOPP_MapAlignerTreeGuided.cli
+    <B>INI file documentation of this tool:</B>
+    @htmlinclude TOPP_MapAlignerTreeGuided.html
+*/
+
+// We do not want this class to show up in the docu:
+/// @cond TOPPCLASSES
+
+class TOPPMapAlignerTreeGuided :
```
cbielow: There seems to be a mixed usage of `MapAlignerTreeGuided` and `MapAlignerTree` throughout.
Please pick one of them and stick to it for internal class names, filenames, INI files, documentation etc...

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+  TOPPMapAlignerTreeGuided() :
+          TOPPMapAlignerBase("MapAlignerTreeGuided", "Tree guided correction of retention time distortions between maps.")
+  {
+  }
+
+private:
+  /// Type to store retention times given for individual peptide sequence
+  typedef std::map<String, DoubleList> SeqAndRTList;
+
+  class PeptideIdentificationsPearsonDistance
```
cbielow: should be `PeptideIdentificationsPearsonDistance_` (since its a private class)

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+      // pearsonCorrelationCoefficient(rt_map_i, rt_map_j)
+      float pearson_val;
+      pearson_val = static_cast<float>(pearsonCorrelationCoefficient(intercept_rts1.begin(), intercept_rts1.end(),
+              intercept_rts2.begin(), intercept_rts2.end()));
+      if (pearson_val > 1)
+      {
+        throw Exception::InvalidRange(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
+      }
+
+      return (1 - (pearson_val * intercept_size / union_size));
```
cbielow: can you add some documentation of why this is useful

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+  progresslogger.startProgress(0, ins.size(), "loading input files");
+  for (Size i = 0; i < ins.size(); ++i)
+  {
+    progresslogger.setProgress(i);
+    fxml_file.load(ins[i], maps[i]);
+  }
+  progresslogger.endProgress();
+}
+
+void TOPPMapAlignerTreeGuided::getPeptideSequences_(const vector<PeptideIdentification>& peptides,
```
cbielow: since the output vectors are not cleared, 'addPeptideSequences_' would be a better name?

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+    }
+  }
+}
+
+void TOPPMapAlignerTreeGuided::extractSeqAndRt_(const vector<FeatureMap>& feature_maps,
+        vector<SeqAndRTList>& maps_seq_and_rt, vector<vector<double>>& maps_ranges)
+{
+  for (Size i = 0; i < feature_maps.size(); ++i)
+  {
+    vector<double> rts_tmp(feature_maps[i].size());
```
cbielow: ouch!
This creates a vector with a lot of 0's, into which feature RTs are appended...
This seems wrong...

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+    {
+      if (!feature_it->getPeptideIdentifications().empty())
+      {
+        getPeptideSequences_(feature_it->getPeptideIdentifications(), maps_seq_and_rt[i], rts_tmp);
+      }
+    }
+    sort(rts_tmp.begin(), rts_tmp.end());
+
+    maps_ranges[i] = rts_tmp;
+    rts_tmp.clear();
```
cbielow: no need to clear. Only if `rts_tmp` is declared outside the loop (which might increase performance)

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+  }
+}
+
+void TOPPMapAlignerTreeGuided::buildTree_(vector<FeatureMap>& feature_maps, std::vector<BinaryTreeNode>& tree, vector<vector<double>>& maps_ranges)
+{
+  vector<SeqAndRTList> maps_seq_and_rt(feature_maps.size());
+  extractSeqAndRt_(feature_maps, maps_seq_and_rt, maps_ranges);
+  PeptideIdentificationsPearsonDistance pep_dist;
+
+  DistanceMatrix<float> dist_matrix;
```
cbielow: can be condensed to 
```
DistanceMatrix<float> dist_matrix(feature_maps.size(), 0.0);
```

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+  while (original_distance(min.first, min.second) < threshold)
+  {
+    //grow the tree
+    cluster_tree.emplace_back(*(clusters[min.second].begin()), *(clusters[min.first].begin()), original_distance(min.first, min.second)/2);
+    if (cluster_tree.back().left_child > cluster_tree.back().right_child)
+    {
+      std::swap(cluster_tree.back().left_child, cluster_tree.back().right_child);
+    }
+
+    if (original_distance.dimensionsize() > 2)
```
cbielow: for better readability you could just
```
if (original_distance.dimensionsize() <= 2) break;
```
and omit the else

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+    fxml_file.setOptions(param);
+
+    vector<FeatureMap> feature_maps(in_files_size);
+    loadInputMaps_(feature_maps, in_files, fxml_file);
+
+    //-------------------------------------------------------------
+    // calculations
+    //-------------------------------------------------------------
+
+    vector<vector<double>> maps_ranges(in_files_size);  // to save ranges for alignment (larger rt_range -> reference)
```
cbielow: can you move functions `buildTree_`, `treeGuidedAlignment_`, `computeTransformationsByOrigInFeature_` and `computeTransformedFeatureMaps_` to a class inside the library, so everyone can use it

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -0,0 +1,581 @@
...
+        point.first = fit->getRT();
+      }
+      point.second = fit->getRT();
+      point.note = fit->getUniqueId();
+      trafo_data_tmp.push_back(point);
+      ++fit;
+    }
+    transformations[map_idx] = TransformationDescription(trafo_data_tmp);
+    transformations[map_idx].fitModel(model_type, model_params);
+    trafo_data_tmp.clear();
```
cbielow: no need to clear, unless you move `trafo_data_tmp` outside the loop

src/topp/MapAlignerTreeGuided.cpp
```diff
@@ -300,14 +300,15 @@ void TOPPMapAlignerTreeGuided::getPeptideSequences_(const vector<PeptideIdentifi
 void TOPPMapAlignerTreeGuided::extractSeqAndRt_(const vector<FeatureMap>& feature_maps,
         vector<SeqAndRTList>& maps_seq_and_rt, vector<vector<double>>& maps_ranges)
 {
+  vector<double> rts_tmp;
```
cbielow: looking at this again, you could just pass `maps_ranges[i]` to `addPeptideSequences_()` like you do with the `maps_seq_and_rt[i]` and omit the temporary vector.
The `reserve` still makes sense probably, but then also call `.shrink_to_fit()` instead of `clear()`, just to make sure we are not wasting big amounts of memory.

jthueringer: `shrink_to_fit` I did not know at all. But after using `push_back()`  shouldn't you always use `shrink_to_fit` to adjust the capacity, then? 
I would leave out `reserve` now, unless you recommend it. But then I would have to use it in several other places as well?!

cbielow: `shrink_to_fit` only makes sense if you called `reserve` and you know that you overestimated capacity (i.e. the final size is smaller).
The only way to know if reserve is worth it (if you know upfront the capacity will be larger than size!), is to benchmark.

If you already know the exact number of items upfront, reserve or resize definitely make sense. you can `reserve + push_back` or `resize + [i]`... the latter might be faster (for less obvious reasons). No shrink required in both cases.

timosachsenberg: shrink_to_fit reallocates the memory of the vector and needs to internally copy all data to the new memory location (http://www.cplusplus.com/reference/vector/vector/shrink_to_fit/). For large vectors it might make sense. Shrinking many small ones and reallocating them might lead to more memory fragmentation so I think it is generally not advisable to shrink to fit every vector.

cbielow: no, `shrink_to_fit` CAN realloc, but may very well do nothing (e.g. if the capacity is below the minimum block size that the allocator would use anyway).
You just have to trust the implementation to do 'sensible' things.

jthueringer:
Many thanks to all contributors for their comments and suggestions. Because the requested changes, especially the movement of certain functions to the library, require a major refactoring, I am closing the Pull Request now and will open a new one with the changes later. Until then, best regards.

src/topp/MapAlignerTreeGuided.cpp
cbielow: these should all be public (and hence tested in the class test)
`computeTransformationsByOrigInFeature_ ` could be renamed to `computeTrafosByOriginalRT_ `

Issue #4407, opened 2019-12-11
Suggestion: rename csv -> tsv


bernt-matthias:
The command line help often mentions csv as file format but actually means tsv. I guess for backward compatibility .csv files need still to be allowed (since I guess that csv and tsv files are a case where the extension is actually used to determine the file type).

bernt-matthias:
The PercolatorAdapter has .tab format which probably also could be tsv.

jpfeuffer:
Any left?

jpfeuffer:
Since every PR references is merged, I believe it is done. Reopen if not.

jpfeuffer:
Actually @timosachsenberg just introduced some new "tab" files.

PR #4419, opened 2019-12-16
PercolatorAdapter: rename tab to tsv

bernt-matthias:
fixes part of https://github.com/OpenMS/OpenMS/issues/4407

### DIFF
--- a/src/tests/topp/THIRDPARTY/third_party_tests.cmake
+++ b/src/tests/topp/THIRDPARTY/third_party_tests.cmake
@@ -191,7 +191,10 @@ if (NOT (${PERCOLATOR_BINARY} STREQUAL "PERCOLATOR_BINARY-NOTFOUND"))
   set_tests_properties("TOPP_PercolatorAdapter_3" PROPERTIES DEPENDS "TOPP_PercolatorAdapter_2")
   add_test("TOPP_PercolatorAdapter_4" ${TOPP_BIN_PATH}/PercolatorAdapter -test -osw_level transition -in_osw PercolatorAdapter_3_out1.osw -out PercolatorAdapter_4_out1.osw -out_type osw -percolator_executable "${PERCOLATOR_BINARY}")
   set_tests_properties("TOPP_PercolatorAdapter_4" PROPERTIES DEPENDS "TOPP_PercolatorAdapter_3")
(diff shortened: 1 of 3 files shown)


timosachsenberg:
looks good - is this tested?

bernt-matthias:
There are tests for both adapters

https://github.com/OpenMS/OpenMS/blob/ce53d13da1fc6f499f4cb0eb7de28a0bdb6e8740/src/tests/topp/THIRDPARTY/third_party_tests.cmake#L184

https://github.com/OpenMS/OpenMS/blob/ce53d13da1fc6f499f4cb0eb7de28a0bdb6e8740/src/tests/topp/THIRDPARTY/third_party_tests.cmake#L351

Not sure if the corresponding arguments are involved in the test.

timosachsenberg:
I don't know if these changes will work.
At least form the documentation it seems that spectraST is expecting xls as an extension.
see http://tools.proteomecenter.org/wiki/index.php?title=Software:SpectraST

bernt-matthias:
See http://tools.proteomecenter.org/wiki/index.php?title=Software:SpectraST#SpectraST_Search_Mode

> .xls format, a tab-delimited column text format

timosachsenberg:
maybe I am missing something here  but if you pass .tab instead of .xls how does spectrast know how to handle it? It doesn't know about .tab and I think it only looks at the extension

bernt-matthias:
Ok. So you suggest to leave everything in temp_directory_body .tab?

timosachsenberg:
Percolator is probably not that picky regarding the input file extension (but we would need to check). For SpectraST they base the semantics on how to handle a file on the extension so we are basically doomed to write out the file in the extension it expects.
Do you need to know all e.g. MIME types of files that are created (even if they are only temporary files?) for technical reasons? We might come up with workarounds (e.g., writing out an xls file  but moving it later to .tab if needed) if really required.

bernt-matthias:
> Do you need to know all e.g. MIME types of files that are created (even if they are only temporary files?) for technical reasons? We might come up with workarounds (e.g., writing out an xls file but moving it later to .tab if needed) if really required.

I don't care for temporary files at all. It's just that consistent semantics of the file extensions are really important for automatic processing (e.g. in workflow engines). Otherwise one is doomed to create tool specific special cases etc. 

So a file should have the extension `csv` if and only if it is a comma separated tabular file. Similarly for tsv <-> tab separated, xls <-> microsoft excel proprietary format. 

This is really important, in particular for formats that can't be distinguished by their content and should be easy for formats for which the extensions are widely accepted.

timosachsenberg:
I think we need to undo the change to spectrast. For percolator it should be fine

jpfeuffer:
what is the status here?

bernt-matthias:
I guess we can only find out with tests .. which I just added.

jpfeuffer:
you meant test #5 right?

bernt-matthias:
Jep. My fault.

PR #4424, opened 2019-12-18
Fix mz xml for max quant

cbielow:
minor follow up on #4423


timosachsenberg:
feel free to merge if relevant tests pass

PR #4425, opened 2019-12-18
CVInspector: the mapping_file argument is mandatory

bernt-matthias:
While working on a test I realized that the mapping argument is not optional.

Speaking of a test... there is currently none. Does anyone has test data to formulate one?

### DIFF
--- a/src/utils/CVInspector.cpp
+++ b/src/utils/CVInspector.cpp
@@ -84,7 +84,7 @@ protected:
 
     registerStringList_("cv_names", "<names>", StringList(), "List of identifiers (one for each ontology file).");
 
-    registerInputFile_("mapping_file", "<file>", "", "Mapping file in CVMapping (XML) format.", false);
+    registerInputFile_("mapping_file", "<file>", "", "Mapping file in CVMapping (XML) format.");
     setValidFormats_("mapping_file", ListUtils::create<String>("XML"));
 
     registerStringList_("ignore_cv", "<list>", ListUtils::create<String>("UO,PATO,BTO"), "A list of CV identifiers which should be ignored.", false);


timosachsenberg:
The mapping files test additional constraints one can not express in the schema itself.
The MAPPING folder e.g., https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/MAPPING/ms-mapping.xml for mzML can be used AFAIK

bernt-matthias:
Superb. Together with https://raw.githubusercontent.com/HUPO-PSI/mzIdentML/master/cv/XLMOD.obo  got it running. 

No idea if this combination of inputs is useful. 

Should I contribute a PR making this argument mandatory and maybe adding a test? Unfortunately the full path to the mapping file has to be specified and `MAPPING/ms-mapping.xml`

timosachsenberg:
Ok I didn't know that it is required but in that case we should make it mandatory.
@mwalzer you are more familiar with CV + mapping. What's your take on that?

hroest:
@mwalzer I was just looking at this tool and I could not get it to work


```
$ CVInspector -cv_files share/OpenMS/CV/psi-ms.obo  -cv_names ms -mapping_file share/OpenMS/MAPPING/ms-mapping.xml 
...
MS:1002966 chrom format
MS:1002996 Andromeda:apl file format
MS:1003009 Shimadzu Biotech LCD format
Error: Unexpected internal error (the value 'GO:0005575' was used but is not valid; Invalid CV identifier!)
```

is that expected? Did I do something wrong? for some reason `-ignore_cv "UO PATO BTO GO"` did also not help

timosachsenberg:
@hroest does -ignore_cv "UO,PATO,BTO,GO" work?

mwalzer:
@timosachsenberg so the one really important use for mapping files is to check deeper levels of semantic validity of *ML files. (Are specific cv terms (or families) present/absent if certain conditions are met, how often a cv_term is present, etc.)
So in that respect, the important tests would be in the Validators. And there, the only meaningful combination of inputs would imo be: mapping.xml, *ML file (obo and schema implicitly)
If you'd want to see what is mandated for cv terms in specific formats, I believe it would make the most sense to combine cv term list (and obo) with a file type/use case and let that implicitly be associated with the given mapping files. 
I put the obo files in parenthesis before twice, bcs. I think each term should resolve to the correct obo file via the prefix key. However versions might be an issue, this adds a lot more interdependence issues that stem from the way CVs are maintained.

mwalzer:
@hroest if `-ignore_cv "UO PATO BTO GO"` didn't help does the GO term appear in your 'cv_files' entries? I believe the tool has to look up the details for the cv terms it looks at from somewhere. - See my comment before, imo resolving obo file from accession should be automatic, but I don't know if that would be feasible with huge '3rd' party ontologies such as GO.

timosachsenberg:
what was the conclusion? accept as is?

bernt-matthias:
From my side, yes. The PR just fixes that the parameter is required. 

As @hroest and @mwalzer pointed out there are more problems of the tool (or the test .. unsure). Not sure if we should solve them in this PR or in a separate PR (or just save them as issue).

Issue #4428, opened 2019-12-18
RTEvaluation apparently missing argument -sequences_file


bernt-matthias:
RTEvaluation aborts with the message: `Internal error: Request for unregistered parameter 'sequences_file'`.

PR #4429, opened 2019-12-18
RTEvaluation add missing argument

bernt-matthias:
might fix: https://github.com/OpenMS/OpenMS/issues/4428

Currently there is no test for this tool. Suggestions?

### DIFF
--- a/src/utils/RTEvaluation.cpp
+++ b/src/utils/RTEvaluation.cpp
@@ -86,6 +86,8 @@ protected:
   {
     registerInputFile_("in", "<file>", "", "Input file");
     setValidFormats_("in", ListUtils::create<String>("idXML"));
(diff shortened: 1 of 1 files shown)


src/utils/RTEvaluation.cpp
```diff
@@ -86,6 +86,8 @@ class TOPPIdXMLEvaluation :
   {
     registerInputFile_("in", "<file>", "", "Input file");
     setValidFormats_("in", ListUtils::create<String>("idXML"));
+    registerInputFile_("sequences_file", "<file>", "", "Fasta File", false);
+    setValidFormats_("in", ListUtils::create<String>("fasta"));
```
timosachsenberg: I think the valid formats should be set on sequence_file parameter and not on “in”

bernt-matthias: sure

PR #4431, opened 2019-12-19
[FIX] PrecursorPurity not thread safe yet, keep it single threaded

enetz:
After some more testing and thinking this through, this is not thread safe yet, so we should keep it single threaded for now.

The variables `current_parent_index` and `next_parent_index` are defined outside of the parallel loop and are updated within the loop. This can invalidate them.


timosachsenberg:
can you check out https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/KERNEL/MSExperiment.cpp#L523
this is the correct implementation for getting the precursor spectrum

enetz:
@timosachsenberg I can use that for the precursor, the previous MS1.
I also adapted the method to look for the next one. They are now done independently for each MS2 spectrum without common variables.
I am testing this for thread safety as well as I can again now.

enetz:
No problems on 4 cores so far, seems to be working.

cbielow:
this extra `break` code is really required?
What happens if I slice mzMLs in half, then this would most likely happen, right? (even if I only broke the first MS2 specs purity computation?).
And if it happens, right now, we're in big trouble anyway, right? So why not just let the computation finish inside the loop (will take longer, but it should be the rare exception and the code will be cleaner).

enetz:
@cbielow I am not sure how to handle this. Right now essentially anytime we detect something wrong or unexpected with the data, we just give a warning and return an empty vector. Because we can't really be sure what the cause of the unexpected data is and it might lead to unexpected behaviour.
Not having an MS1 in the front could also mean, that all the spectra are just scrambled and out of order. Then all the data returned here would be simply wrong.
You are right about splitting mzML files, but in that case I would prefer to make an mzML file splitter that prefers to put MS2 spectra in the same file as their precursor spectrum. Otherwise you simply have a couple of spectra without this data, and it will be a different subset depending on how you split files.

enetz:
I need some variable to tell me if I have to return data or an empty vector at the end anyway, so the only code I could remove would be 
```
      if (break_loop)
      {
        continue;
      }
```
and I don't think this one piece of code makes it much more complicated

timosachsenberg:
Could you check all in advance? Would make the code less complex

enetz:
I now check for existing parent spectra, empty native IDs, and non-unique non-empty IDs before the parallel loop. Non-unique non-empty IDs are found by inserting initialized values into the map. These are then replaced in the parallel loop.

src/openms/source/ANALYSIS/ID/PrecursorPurity.cpp
```diff
@@ -144,64 +144,63 @@ namespace OpenMS
...
           }
         }
 
         PrecursorPurity::PurityScores score;
-        PrecursorPurity::PurityScores score1 = PrecursorPurity::computePrecursorPurity(spectra[current_parent_index], spectra[i].getPrecursors()[0], precursor_mass_tolerance, precursor_mass_tolerance_unit_ppm);
+        PrecursorPurity::PurityScores score1 = PrecursorPurity::computePrecursorPurity((*parent_spectrum_it), spectra[i].getPrecursors()[0], precursor_mass_tolerance, precursor_mass_tolerance_unit_ppm);
         // use default values of 0, if there is an MS2 spectrum without an MS1 after it (may happen for the last group of MS2 spectra)
         PrecursorPurity::PurityScores score2;
-        if (!lastMS1) // there is an MS1 after this MS2, combine values from previous and next MS1
+        if (next_parent_index > i) // there is an MS1 after this MS2, combine values from previous and next MS1
```
timosachsenberg: why is this done?

enetz: I got rid of the `lastMS1` variable as well, since it was also shared between threads.
If both flanking MS1 spectra are available, I combine the values from both. Otherwise I only use the parent MS1.

timosachsenberg: I see. I would assume that this could go wrong sometimes (e.g., maybe polarity is changed or whatever). Would only use the parent

enetz: @timosachsenberg If I remember correctly, this was originally your suggestion :)
I agree that it is not applicable in some cases. I will only use the parent then.

timosachsenberg: hehe ok well then happy to review own suggestions

enetz: @timosachsenberg  I also had to update some OpenPepXL test files this time, because most of the values PrecursorPurity returns now have changed. But no RNPxlSearch tests failed. Are there no MS1 spectra in you test data? That would be fine then. Otherwise this is unexpected.

timosachsenberg: yes, they are plain MS2 spectra

PR #4432, opened 2019-12-19
mzXML: add more attributes which are optional in mzXML but required for MaxQuant

cbielow:
e.g. then converting data with empty spectra

augments #4423 and #4424


cbielow:
test pass

PR #4441, opened 2019-12-28
FileMerger: allow fasta output

bernt-matthias:
Currently merging works for fasta only if the outfile is named .tmp. 

Documentation on the possible combinations of input types and output types would be nice. I would need help for this.


timosachsenberg:
I agree that the documentation needs some improvement. I will open an issue and let's hope we will find some volunteers :)

timosachsenberg:
nice catch! thanks

PR #4443, opened 2019-12-29
OpenSwathDIAPreScoring: fix outputs

bernt-matthias:
- the `-out` argument was unused
- output is only generated if swath files are given
  - make it mandatory
  - let the user specify the outfiles explicitly


timosachsenberg:
@hroest good to merge?

PR #4451, opened 2020-01-07
OpenPepXL, OpenPepXLLF: remove xml from valid extensions

bernt-matthias:
since https://github.com/OpenMS/OpenMS/blob/c19b545636e76a5770f095f944e0b196b9b56c82/src/openms/source/FORMAT/XQuestResultXMLFile.cpp#L97 will raise an error when a .xml file is given.

I suggest to remove xml also for spec.xml

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/FileTypes.h
+++ b/src/openms/include/OpenMS/FORMAT/FileTypes.h
@@ -110,6 +110,7 @@ namespace OpenMS
       SPLIB,              ///< SpectraST binary spectral library file (sptxt is the equivalent text-based format, similar to the MSP format)
       NOVOR,              ///< Novor custom parameter file
       XQUESTXML,          ///< xQuest XML file format for protein-protein cross-link identifications (.xquest.xml)
+      SPECXML,            ///< xQuest XML file format for matched spectra for spectra visualization in the xQuest results manager (.spec.xml)
       JSON,               ///< JavaScript Object Notation file (.json)
       RAW,                ///< Thermo Raw File (.raw)
       EXE,                ///< Executable (.exe)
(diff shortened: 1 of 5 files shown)


enetz:
There is a problem with the `spec.xml` file ending. Without allowing `xml` as a valid format, I get this when I try to write out a file:
`Invalid parameter: Invalid output file extension for file 'path/to/filename.spec.xml'. Valid file extensions are: 'spec.xml'.`
It works with `xml` as a valid file extension.

It may have something to do with having a dot as part of the file ending, if only the part after the last dot is considered as a file ending. But somehow it works for `filename.xquest.xml`. I haven't figured out yet why they are treated differently.

The tests run through, because they write out `.tmp` files. Those are always allowed. Currently `xml` is necessary as a valid format for `spec.xml`. I would have renamed the formats to `xquestXML` and `xquestSpecXML` or something similar, but then they wouldn't be compatible with the xQuest pipeline anymore.

jpfeuffer:
You might need to register spec.xml in the FileTypes.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/FileTypes.cpp#L142

bernt-matthias:
Thanks @jpfeuffer for the tip. @enetz can you try with my latest commit. Maybe we can/should adapt the test accordingly?

enetz:
@jpfeuffer do you know whether there are any more places it needs to be registered? adding `spec.xml` to `FileTypes.cpp` and `FileTypes.h` did not fix the error for me.

enetz:
@bernt-matthias adding it to just the cpp should cause an error, you also need to add it to the `Type` enum in `FileTypes.h`.  But that did not fix the original Invalid Parameter error for me.

jpfeuffer:
Ah good catch. And the ".h" File as Eugen mentioned.

bernt-matthias:
Done. Attention .. I force pushed

enetz:
I overlooked the two lines in `FileHandler.cpp` in my tests, will test this branch again now

PR #4452, opened 2020-01-07
Feature/aasequence speed

hroest:
Several performance improvements for `AASequence` and parsing from string, especially for modified sequences

- use `std::unordered_map` to store modification names
- reserve space for new residues
- faster check if residues match a given modification or not
- remove unnecessary double lookups in the map

# Benchmarks

Benchmarking was done with 

```
  int nr_mol = 2e7; // for AASequence
  nr_mol = 1e6; // for Oxidation
  int nr_fragments = 0;
  for (Size i = 0; i < nr_mol;  i++) 
  {
    auto pep = AASequence::fromString("DFPMIAMNGMER"); 
    nr_fragments += pep.size();
  }
```

## Unmodified AASequence

This is quite fast, the current PR reduces execution time from 4.5 s to 2s for 2e7 iterations, basically reducing time from 220 nsec to 95 nsec for parsing. This is mainly due to reserving space in the vector

## Modified AASequence

Modifications slow down parsing significantly, taking about 3.2 usec before optimization. Therefore a modified sequence would take about 15x longer to parse (or 32x longer after the optimization above) which means that speed will always be dominated by modified AA strings. A modification with 3 oxidized residues now takes only 1.5 usec to parse (down from 3.2 usec) with each modification taking about 400 nsec to parse:

- 1.5 usec for three mods
- 1.1 usec for two mods
- 700 nsec for one modification

Most of the time is spent looking up modifications names in `searchModifications` and in identifying the correct one among the results (e.g. for Oxidation there are 13 results in the hash, so that means the code needs to iterate through all 13 and find the correct one. It seems a bit excessive since we have a different modification for each oxidized amino acid).

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -39,6 +39,7 @@
 #include <OpenMS/CHEMISTRY/ResidueModification.h>
 
 #include <set>
+#include <unordered_map>
 
 namespace OpenMS
 {
@@ -175,7 +176,7 @@ protected:
     std::vector<ResidueModification*> mods_;
 
     /// Stores the mappings of (unique) names to the modifications
-    Map<String, std::set<const ResidueModification*> > modification_names_;
+    std::unordered_map<String, std::set<const ResidueModification*> > modification_names_;
 
     /// Helper function to check if a residue matches the origin for a modification
     bool residuesMatch_(const String& residue, const ResidueModification* origin) const;
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -100,6 +100,12 @@ public:
     */
     void searchModifications(std::set<const ResidueModification*>& mods, const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
 
+    const ResidueModification* searchModifications_fast(const String& mod_name, 
+bool& multiple_matches,
+                                                        const String& residue = "",
+                                                        ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY
+                                                        ) const;
+
     /**
        @brief Returns the modification with the given name
 
--- a/src/openms/source/CHEMISTRY/AASequence.cpp
+++ b/src/openms/source/CHEMISTRY/AASequence.cpp
@@ -1001,12 +1001,11 @@ namespace OpenMS
           "Cannot convert string to peptide modification: missing ')'");
     }
 
-    ModificationsDB* mod_db = ModificationsDB::getInstance();
-
     // First search for N or C terminal modifications (start of peptide indicates N-terminal modification as well)
     if (aas.peptide_.empty() || specificity == ResidueModification::N_TERM ||
                                 specificity == ResidueModification::PROTEIN_N_TERM)
     {
+      ModificationsDB* mod_db = ModificationsDB::getInstance();
       // Advance iterator one or two positions (we may or may not have a dot
       // after the closing bracket) to point to the first AA of the peptide.
       String::ConstIterator next_aa = mod_end;
@@ -1028,11 +1027,13 @@ namespace OpenMS
     const String& res = aas.peptide_.back()->getOneLetterCode();
     if (specificity == ResidueModification::PROTEIN_C_TERM)
     {
+      ModificationsDB* mod_db = ModificationsDB::getInstance();
       aas.c_term_mod_ = proteinTerminalResidueHelper(mod_db, 'c', str, mod, res);
       return mod_end;
     }
     else if (specificity == ResidueModification::C_TERM)
     {
+      ModificationsDB* mod_db = ModificationsDB::getInstance();
       aas.c_term_mod_ = terminalResidueHelper(mod_db, 'c', true, str, mod, res);
       return mod_end;
     }
@@ -1044,6 +1045,8 @@ namespace OpenMS
     }
     catch(...)  // no internal mod for this residue
     {
+      ModificationsDB* mod_db = ModificationsDB::getInstance();
+
       // TODO: get rid of this code path, its deprecated and is only a hack for
       // C/N-terminal modifications that don't use the dot notation
       if (std::distance(str_it, str.begin()) == -1)
--- a/src/openms/source/CHEMISTRY/AASequence.cpp
+++ b/src/openms/source/CHEMISTRY/AASequence.cpp
@@ -1380,7 +1380,12 @@ namespace OpenMS
   void AASequence::parseString_(const String& pep, AASequence& aas,
                                 bool permissive)
   {
+    // Reserving space, populate it and then shrink again (since we probably
+    // over-allocate due to modifications). This substantially speeds up the
+    // function for unmodified sequences (3x speedup).
     aas.peptide_.clear();
+    aas.peptide_.reserve(pep.size());
+
     String peptide(pep);
     peptide.trim();
 
@@ -1482,6 +1487,8 @@ namespace OpenMS
     // since the user might just want to represent the sequence (including modifications on other AA's),
     // e.g. when digesting a peptide
     // We check for 'weightless' X in places where a mass is needed, e.g. during getMonoMass()
+
+    aas.peptide_.shrink_to_fit();
   }
 
   void AASequence::getAAFrequencies(Map<String, Size>& frequency_table) const
--- a/src/openms/source/CHEMISTRY/AASequence.cpp
+++ b/src/openms/source/CHEMISTRY/AASequence.cpp
@@ -1384,10 +1384,10 @@ namespace OpenMS
     // over-allocate due to modifications). This substantially speeds up the
     // function for unmodified sequences (3x speedup).
     aas.peptide_.clear();
-    aas.peptide_.reserve(pep.size());
 
     String peptide(pep);
     peptide.trim();
+    aas.peptide_.reserve(peptide.size());
 
     if (peptide.empty()) return;
 
--- a/src/openms/source/CHEMISTRY/ModificationsDB.cpp
+++ b/src/openms/source/CHEMISTRY/ModificationsDB.cpp
@@ -119,7 +119,8 @@ namespace OpenMS
     #pragma omp critical(OpenMS_ModificationsDB)
     {
       bool found = true;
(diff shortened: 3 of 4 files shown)


hroest:
Another issue is that we have almost 10k different modification names that we store which also makes this highly inefficient. This is probably why the lookup is so slow.

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -175,7 +176,7 @@ namespace OpenMS
     std::vector<ResidueModification*> mods_;
 
     /// Stores the mappings of (unique) names to the modifications
-    Map<String, std::set<const ResidueModification*> > modification_names_;
+    std::unordered_map<String, std::set<const ResidueModification*> > modification_names_;
 
     /// Helper function to check if a residue matches the origin for a modification
     bool residuesMatch_(const String& residue, const ResidueModification* origin) const;
```
cbielow: this function is not defined anymore. Can be removed.

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1377,7 +1380,12 @@ namespace OpenMS
   void AASequence::parseString_(const String& pep, AASequence& aas,
```
cbielow: you could pass `pep` by value (since its copied internally anyways)

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1377,7 +1380,12 @@ namespace OpenMS
...
                                 bool permissive)
   {
+    // Reserving space, populate it and then shrink again (since we probably
+    // over-allocate due to modifications). This substantially speeds up the
+    // function for unmodified sequences (3x speedup).
     aas.peptide_.clear();
+    aas.peptide_.reserve(pep.size());
```
cbielow: only reserve after trimming?! (line 1390)

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -175,7 +176,7 @@ namespace OpenMS
     std::vector<ResidueModification*> mods_;
 
     /// Stores the mappings of (unique) names to the modifications
-    Map<String, std::set<const ResidueModification*> > modification_names_;
+    std::unordered_map<String, std::set<const ResidueModification*> > modification_names_;
```
cbielow: have you tried using a plain `vector<String>` for searching (you'd need a second vector for the keys)? This might be even faster, especially if you put the most common modifications first

src/openms/source/CHEMISTRY/ModificationsDB.cpp
```diff
@@ -257,6 +289,8 @@ namespace OpenMS
   {
     double min_error = max_error;
     const ResidueModification* mod = nullptr;
+    char res = '?'; // empty
+    if (!residue.empty()) res = residue[0];
```
cbielow: not benchmarked at all, but to avoid the (costly) branch you could do:
```
char candidates[2] = { residue[0], '?' };
const char res = candidates[residue.empty()];
```

`residue[0]` should not be UB since the empty string is guaranteed to end in `\0' as of C++11.

hroest: I did the benchmarking on godbolt: https://godbolt.org/z/zFppCc and it looks like on (gcc): 5 instructions for my solution, 8 instructions for yours. I suggest we stick to the one there is

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
hroest: you mean a sorted vector and using binary search in the vector? No I have not tried. In principle this should be O(1) and binary search is O(log(N)) so it should be slower. I think actually the way to make this faster would be to not have 10k entries in the map in the first place. We cant possibly need that many modifications.

src/openms/source/CHEMISTRY/AASequence.cpp
hroest: I agree, that could be more efficient (not here though): https://stackoverflow.com/questions/270408/is-it-better-in-c-to-pass-by-value-or-pass-by-constant-reference

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
cbielow: No binary search, but plain front-to-back search (making use of cache and prefetcher). Especially when frequently used entries (Oxidation,...) are upfront, this should always beat binary search.
Vs. HashMaps it may or may not be faster... but requires less memory in any case.

src/openms/source/CHEMISTRY/ResidueDB.cpp
```diff
@@ -535,7 +535,8 @@ namespace OpenMS
         try
         {
           // terminal modifications don't apply to residues (side chain), so only consider internal ones
-          mod = ModificationsDB::getInstance()->getModification(modification, residue->getOneLetterCode(), ResidueModification::ANYWHERE);
+          static const ModificationsDB* mdb = ModificationsDB::getInstance();
+          mod = mdb->getModification(modification, residue->getOneLetterCode(), ResidueModification::ANYWHERE);
```
hroest: honestly, I dont understand why that is the case, but it really saves some time. I did not see this effect for `ResidueDB` and other singletons, its strange. I think keeping a single static ptr is safe, as the database can be modified but the ptr location is constant. Still, I cant understand why this makes a difference - but we should also look in other locations for this!

hroest:
speed is now down from initially 3.2 usec to previously 1.5 usec to now 750 nsec for the triple modification peptide. The time it takes to parse the modified peptide `DFPM(Oxidation)IAM(Oxidation)NGM(Oxidation)ER` is as follows:

90 nsec if it has any bracket + 190 * x nanoseconds (x = nr mods)

or

```
105 + has_mod * 90 + nr_mods * 190
```

So overall from the starting point of 3.2 usec, this is a speedup of 4-fold for parsing modified peptides. It also means that peptides with a single modification will take around 400 nsec to parse which is in the same order of magnitude than unmodified peptides (100 nsec) and actually similarly long as an unmodified peptide took before the optimization.

src/openms/source/CHEMISTRY/ModificationsDB.cpp
```diff
@@ -234,33 +234,34 @@ namespace OpenMS
...
     {
       OPENMS_LOG_WARN << "Warning (ModificationsDB::getModification): more than one modification with name '" + mod_name + "', residue '" + residue + "', specificity '" + String(Int(term_spec)) << "' found, picking the first one of:";
-      for (auto it = mods.begin(); it != mods.end(); ++it)
-      {
-        OPENMS_LOG_WARN << " " << (*it)->getFullId();
-      }
+      // for (auto it = mods.begin(); it != mods.end(); ++it)
+      // {
+      //   OPENMS_LOG_WARN << " " << (*it)->getFullId();
+      // }
```
hroest: note that this changes the behavior a bit: if multiple mods are found, we cannot report them any more. We can only say that we did find multiple mods.

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
hroest: Turns out that lookup (if you only do it once) is not that bad, maybe 10% of the total time. However, iterating overall results is bad and takes much more time around 30% of time since for each lookup of "Oxidation" we have to evaluate 13 `ResidueModification*` objects which takes time. These two things together explain most of the time spent in `ModificationsDB`

src/openms/source/CHEMISTRY/ModificationsDB.cpp
timosachsenberg: fine for me

src/openms/source/CHEMISTRY/ResidueDB.cpp
timosachsenberg: is this also in -O3 ? I toyed around a bit on godbolt and the default parameters seem to generate quite some different code (also depending on the optimization level)

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -99,6 +100,12 @@ namespace OpenMS
     */
     void searchModifications(std::set<const ResidueModification*>& mods, const String& mod_name, const String& residue = "", ResidueModification::TermSpecificity term_spec = ResidueModification::NUMBER_OF_TERM_SPECIFICITY) const;
 
+    const ResidueModification* searchModifications_fast(const String& mod_name,
```
timosachsenberg: more style conform would be searchModificationsFast
and some docs would need to be added

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1377,9 +1380,14 @@ namespace OpenMS
   void AASequence::parseString_(const String& pep, AASequence& aas,
                                 bool permissive)
   {
+    // Reserving space, populate it and then shrink again (since we probably
+    // over-allocate due to modifications). This substantially speeds up the
+    // function for unmodified sequences (3x speedup).
     aas.peptide_.clear();
+
     String peptide(pep);
     peptide.trim();
```
timosachsenberg: copying the string and trimming it seems also kind of a waste here. but probably not too much time spend here

timosachsenberg:
I did not check all details but I guess everything is still thread-safe?

src/openms/source/CHEMISTRY/ModificationsDB.cpp
```diff
@@ -48,6 +48,33 @@ using namespace std;
 
 namespace OpenMS
 {
+
+  bool inline residuesMatch_fast_(const char residue, const ResidueModification* curr_mod)
```
enetz: maybe residuesMatchFast(... ?

hroest:
> I did not check all details but I guess everything is still thread-safe?

yes, we are actually testing that using a multithreaded test and if we remove the openmp `pragma` then it fails. The new function also has the pragmas

timosachsenberg:
looks good from my side. maybe add some comments and consider to fix the style suggestions

src/openms/source/CHEMISTRY/ResidueDB.cpp
hroest: Yes, this is clearly the case in -O3. 

I looked a bit further into this and it seems to only happen to `ModificationDB`, see https://github.com/OpenMS/OpenMS/pull/4469 and the only difference that I found was the default parameters which does make a big difference in the assembly (see https://github.com/OpenMS/OpenMS/pull/4469#issuecomment-589467386), basically the compiler cannot optimize away the function call and needs to set up all default string arguments which is very costly.

PR #4453, opened 2020-01-08
Feature/matrix py

hroest:
Faster access to data in the matrix class

# Benchmarks

Code for benchmark:

```
import pyopenms
import numpy as np
m = pyopenms.MatrixDouble()
N = 9000
m.resize(N+1, N+1, 5.0)
```
## Naive
```
rows = N
cols = N
data = np.ndarray(shape=(int(rows), int(cols)), dtype=np.float64)
for i in range(int(rows)):
            for j in range(int(cols)):
                data[i][j] = m.getValue(j,i)

print sum(sum(data))
```

-> takes 28 seconds

# Store in Numpy
```
data = m.get_matrix()
print sum(sum(data))
```

for some reason the access to numpy `data[i][j] =` is slow. It still takes 12 seconds

# Store in temporary vector

```
data = m.get_matrix_as_vec()
print sum((data))
```

storing data in a vector and then copying it was faster, taking 6 seconds

# Memory view
```
data = m.get_matrix_as_view()
print sum(sum(data))
```
however, the fastest approach was the memory view approach, which only takes 0.5 seconds. This of course makes some assumptions, namely that the `Matrix` object is a C++ vector of length col*row with the ptr to the object being the start of the data

-> Overall we get a 60x increase in speed when using the memory view

### DIFF
--- a/src/pyOpenMS/addons/MatrixDouble.pyx
+++ b/src/pyOpenMS/addons/MatrixDouble.pyx
@@ -1,0 +1,95 @@
+from Matrix cimport *
+
+
+# continue with extra code if needed
+    def get_matrix(self):
+        """Cython signature: numpy_matrix get_matrix()
+        """
+
+        cdef _Matrix[double] * mat_ = self.inst.get()
+
+        cdef unsigned int rows = mat_.rows()
+        cdef unsigned int cols = mat_.cols()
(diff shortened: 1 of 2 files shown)


timosachsenberg:
Wow that sounds great! I will give it a try soon.

PR #4454, opened 2020-01-08
Update Windows installation instructions for VS2019

cbielow:
* Warn about VS2015 and VS2017 due to OpenMP bug
 * Some small fixes


timosachsenberg:
looks good

Issue #2350, opened 2017-01-13
Inconsistencies in flags versus string options with true/false


blankclemens:
Hell everyone,

it seems like the CTD entry when using `-write_ctd` is wrong. It allows for two values `true/false` which, when applied, lead to errors like this:

    Trailing text argument(s) '[false]' given. Aborting!```

I assume these are just flags, without any argument. The CTD however speaks of two restrictions, true and false:

    <ITEM name="add_decoys" value="false" type="string" description="Create decoy proteins (reversed sequences) and append them to the database for the search (MS-GF+ parameter &apos;-tda&apos;). This allows the calculation of FDRs, but should only be used if the database does not already contain decoys." required="false" advanced="false" restrictions="true,false" />

and

    <ITEM name="add_features" value="false" type="string" description="Output additional features - needed e.g. by Percolator (default: basic scores only; MS-GF+ parameter &apos;-addFeatures&apos;)" required="false" advanced="false" restrictions="true,false" />

This is important since there is no way to further decide automatically how to handle this flag - e.g. the `precursor_error_units` flag looks machine wise exactly the same, but needs the text arguments:

    <ITEM name="precursor_error_units" value="ppm" type="string" description="Unit of precursor mass tolerance (MS-GF+ parameter &apos;-t&apos;)" required="false" advanced="false" restrictions="Da,ppm" />

Did I get this right, or am I missing something?

Kind regards
Clemens

jpfeuffer:
Hi Clemens,

I can see your confusion and also had to look up the exact handling of flags in our TOPP tools.
I think it's important to emphasize on the difference in calling a tool with options from the command line `./FileInfo -in /bla/foo.mzML -debug` and calling a tool by giving it a path to an ini file in ParamXML format `./FileInfo -ini /bar/baz.ctd`.
The problem here is, that in the ParamXML schema version used by us (1.6.2) there was no support for boolean parameters and these so called flags have to be represented as string parameters with restrictions true/false.

Since the boolean type was added to parameter ITEMs in ParamXML 1.7.0 this might change in the future but we would have to check for downstream implications in our supported workflow managers first.

Long story short: The way of writing CTDs is correct since it also uses ParamXML for its parameter section. It is not optimal and imo should be changed in the future if there are no complications.

For users: On the command line use (boolean) flags without argument. In a CTD or ini it has to be described as a string argument which can be true or false.

Fellow developers, please correct me if I am wrong.

Cheers
Julianus

[...]

PR #4455, opened 2020-01-08
masstrace_snr_filtering: is flag

bernt-matthias:
according to the discussion in https://github.com/OpenMS/OpenMS/issues/2350
this should be represented by the restrictions being true,false

Would be nice to have a method to check if there are more such cases.


jpfeuffer:
I guess we could easily grep it. Which ones do you want to make flags? Everything that is true,false or false,true and has default of false?
I think it would make sense but I did not spend much time thinking about all implications.
Btw, this could also be changed in the CTD creating part of the code, such that the order in each TOPP tool would not matter.

bernt-matthias:
@jpfeuffer good idea, but since I autocreate a commandline for the Galaxy tools from the CTDs the only thing that is imoortant is consistency.

jpfeuffer:
Ah hmm.. just to let you know:
from 2.4 to 2.5 the CTD version will change. You will be able to identify flags then.
They then have the bool type. No more true false magic.

bernt-matthias:
> from 2.4 to 2.5 the CTD version will change. You will be able to identify flags then.
> They then have the bool type. No more true false magic.

Good to know (that there will be more work :) ). In all utils and topp tools?

PR #4457, opened 2020-01-08
Feature/matrix py

hroest:
follow up from #4453 

@timosachsenberg any reason why you used

```
data[i][j] = m.getValue(j,i)
```

instead of 
```
data[i][j] = m.getValue(i,j) 
```
in my testing I noticed that the first one is actually incorrect when using a non-square matrix. I added tests now and add a fast accessor method to the C++ class


src/pyOpenMS/addons/MatrixDouble.pyx
```diff
@@ -89,7 +56,7 @@ from Matrix cimport *
         cdef int j = 0
         for i in range(int(rows)):
             for j in range(int(cols)):
-                mat_.setValue(j,i,data[i][j])
+                mat_.setValue(i,j,data[i][j])
```
hroest: @timosachsenberg in your example this was switched. Any reason?

timosachsenberg: Hmm don’t recall. Maybe I mixed it up with how eigen indexes a matrix

hroest: hmm I think I copied your example and its currently incorrect. That is why I wrote the test, according to the test it is correct now :-)

timosachsenberg:
For reference only, this is the example application where I currently use it
https://github.com/OpenMS/pyopenms-extra/blob/master/src/examples/GUI_MS1MapWidget.py

Issue #4461, opened 2020-01-09
FeatureFinderCentroided tests 


bernt-matthias:
The test defined here: https://github.com/OpenMS/OpenMS/blob/735441decb37d3d4501b682e45ac83d767b803f4/src/tests/topp/CMakeLists.txt#L267

fails with:

```
Abort reasons during feature construction:
- Invalid fit: Fitted model is bigger than 'max_rt_span': 1
```

the ini file actually defines 'max_rt_span': 2.5

Seems to be again due to an outdated ini file:

```
Warning: Parameters file version (1.8.0) does not match the version of this tool (2.4.0-HEAD-2018-10-26).
Your current parameters are still valid, but there might be new valid values or even new parameters. Upgrading the INI might be useful.
```

Little bit of background. For the CTD -> Galaxy tool converter I auto-extract tests for the Galaxy tools from the command line in the CMakeFile and the used ini-files. I understand that OpenMS tries to locate the parameters even if the CTD structure is wrong, my test generation scheme can not do this. Hence I observe differences in the results (in this case in some minor digits of some floating point numbers).

jpfeuffer:
Can you elaborate? Why exactky do you think it does not honor max_rt_span?

bernt-matthias:
Because of the output `Invalid fit: Fitted model is bigger than 'max_rt_span': 1`.. But the ini file sets 2.5.

bernt-matthias:
Odd. `-debug 3` shows `"feature|max_rt_span" -> "2.5" (Maximum RT span in relation to extended area that the model is allowed to have.)` .. then I guess I just mis-interpret the output.

cbielow:
the `: 1` just says that this happened once. This number simply depends on the dataset (and parameters).

cbielow:
> fails with:

It does not fail, it merely reports which features did not pass QC. So its more meant as an information for the user, than an error message.
If there were an error, the return code of the tool would be non-zero.

> Hence I observe differences in the results (in this case in some minor digits of some floating point numbers).

This probably not caused by parameters, but simply by platform. To compare results we use the FuzzyDiff tool (and so should you, I guess).

cbielow:
closing, feel free to open if anything remains unclear.

PR #4466, opened 2020-01-10
a bit clearer warning message

bernt-matthias:
follow up to https://github.com/OpenMS/OpenMS/issues/4461


hroest:
maybe also the keyword "abort" is not great to output during execution, maybe there is a better way to write this?

hroest:
e.g. "Failure to extend features due to the following reasons:"

PR #4467, opened 2020-01-10
IDMAssAccuracy: explicitly specify gnuplot script

bernt-matthias:
therefor combine data and gnuplot script in one file

### DIFF
--- a/src/utils/IDMassAccuracy.cpp
+++ b/src/utils/IDMassAccuracy.cpp
@@ -113,16 +113,11 @@ protected:
     setValidFormats_("id_in", ListUtils::create<String>("idXML"));
 
     registerOutputFile_("out_precursor", "<file>", "", "Output file which contains the deviations from the precursors", false, false);
-    setValidFormats_("out_precursor", ListUtils::create<String>("csv"));
-
-    registerStringList_("precursor_columns", "<columns>", ListUtils::create<String>("MassDifference"), "Columns which will be written to the output file", false);
-    setValidStrings_("precursor_columns", ListUtils::create<String>("MassDifference"));
+    setValidFormats_("out_precursor", ListUtils::create<String>("tsv"));
     registerFlag_("precursor_error_ppm", "If this flag is used, the precursor mass tolerances are estimated in ppm instead of Da.");
 
     registerOutputFile_("out_fragment", "<file>", "", "Output file which contains the fragment ion m/z deviations", false, false);
-    setValidFormats_("out_fragment", ListUtils::create<String>("csv"));
-    registerStringList_("fragment_columns", "<columns>", ListUtils::create<String>("MassDifference"), "Columns which will be written to the output file", false);
-    setValidStrings_("fragment_columns", ListUtils::create<String>("MassDifference"));
+    setValidFormats_("out_fragment", ListUtils::create<String>("tsv"));
     registerFlag_("fragment_error_ppm", "If this flag is used, the fragment mass tolerances are estimated in ppm instead of Da.");
 
     registerDoubleOption_("fragment_mass_tolerance", "<tolerance>", 0.5, "Maximal fragment mass tolerance which is allowed for MS/MS spectra, used for the calculation of matching ions.", false, false);
--- a/src/utils/IDMassAccuracy.cpp
+++ b/src/utils/IDMassAccuracy.cpp
@@ -363,7 +363,7 @@ protected:
         String gnuplot_out_file(getStringOption_("gnuplot_script"));
         if (gnuplot_out_file != "")
         {
-          ofstream out(gnuplot_out_file.c_str());
+          ofstream gpl_out(gnuplot_out_file.c_str());
           gpl_out << "set terminal png" << endl;
           gpl_out << "set output \"" << precursor_out_file  << "_gnuplot.png\"" << endl;
           if (precursor_error_ppm)
@@ -379,10 +379,8 @@ protected:
 
 	  for (vector<DPosition<2> >::const_iterator it = values.begin(); it != values.end(); ++it)
           {
-            out << it->getX() << " " << it->getY() << endl;
+            gpl_out << it->getX() << " " << it->getY() << endl;
           }
-          out.close();
-
           gpl_out.close();
         }
 
--- a/src/utils/IDMassAccuracy.cpp
+++ b/src/utils/IDMassAccuracy.cpp
@@ -131,8 +131,8 @@ protected:
     registerIntOption_("number_of_bins", "<#bins>", 100, "Number of bins that should be used to calculate the histograms for the fitting.", false, true);
     setMinInt_("number_of_bins", 10);
 
-    registerStringOption_("generate_gnuplot_scripts", "<false>", "false", "If this option is set to true, the distributions and the fits are used to generate a gnuplot script, that can be used to generate plots. The options 'precursor_out' and 'fragment_out' must be set to take this effect.", false, true);
-    setValidStrings_("generate_gnuplot_scripts", ListUtils::create<String>("true,false"));
+    registerOutputFile_("gnuplot_script", "<file>", "", "Gnuplot output file. The distributions and the fits are used to generate a gnuplot script, that can be used to generate plots. The options 'precursor_out' and 'fragment_out' must be set to take this effect.", false, true);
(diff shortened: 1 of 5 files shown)


bernt-matthias:
Since there is currently only one column in the output files one might also remove `-separator <character>`

timosachsenberg:
as this falls into the realms of QC, @cbielow and @mwalzer what is your take on that?

src/utils/IDMassAccuracy.cpp
```diff
@@ -361,16 +360,10 @@ class TOPPIDMassAccuracy :
...
+        if (gnuplot_out_file != "")
         {
-          ofstream out(String(precursor_out_file + "_gnuplot.dat").c_str());
-          for (vector<DPosition<2> >::const_iterator it = values.begin(); it != values.end(); ++it)
-          {
-            out << it->getX() << " " << it->getY() << endl;
-          }
-          out.close();
-
-          ofstream gpl_out(String(precursor_out_file + "_gnuplot.gpl").c_str());
```
cbielow: this is missing now... should be re-added

src/utils/IDMassAccuracy.cpp
```diff
@@ -382,7 +375,14 @@ class TOPPIDMassAccuracy :
             gpl_out << "set xlabel \"error in Da\"" << endl;
           }
           gpl_out << "set ylabel \"frequency\"" << endl;
-          gpl_out << "plot '" << precursor_out_file << "_gnuplot.dat' title 'Precursor mass error distribution' w boxes, f(x) w lp title 'Gaussian fit of the error distribution'" << endl;
+          gpl_out << "plot '-' using 1:2 title 'Precursor mass error distribution' w boxes, f(x) w lp title 'Gaussian fit of the error distribution'" << endl;
+
+	  for (vector<DPosition<2> >::const_iterator it = values.begin(); it != values.end(); ++it)
```
cbielow: can you move that back to where `out` is opened?

src/utils/IDMassAccuracy.cpp
```diff
@@ -382,7 +375,14 @@ class TOPPIDMassAccuracy :
...
+          {
+            out << it->getX() << " " << it->getY() << endl;
```
cbielow: `endl` is really bad for write speed. If you touch this, 100 charma points for switching to `'\n'`

bernt-matthias:
Ahh.. there are two places where gnuplot scripts are generated, once for precursors and once for fragments. I can change the code accordingly. 

My main intention of the PR is to output just a single file (currently it was a gnuplot script and the data file), i.e., the tabular data, which is explicitly addressed by a parameter. Then the user can decide herself how to plot it (R, gnuplot, or even excel .. just joking). Maybe a single plot script can be distributed.

timosachsenberg:
by the way ... feel free to add you to the AUTHORS file

bernt-matthias:
> Computing fits and histograms in C++ always seems a bit too static to me

Even more radical would be to just drop these files entirely. Could be computed with any statistics software (R) from the data in the primary output.

src/utils/IDMassAccuracy.cpp
```diff
@@ -131,8 +131,11 @@ class TOPPIDMassAccuracy :
     registerIntOption_("number_of_bins", "<#bins>", 100, "Number of bins that should be used to calculate the histograms for the fitting.", false, true);
     setMinInt_("number_of_bins", 10);
 
-    registerStringOption_("generate_gnuplot_scripts", "<false>", "false", "If this option is set to true, the distributions and the fits are used to generate a gnuplot script, that can be used to generate plots. The options 'precursor_out' and 'fragment_out' must be set to take this effect.", false, true);
-    setValidStrings_("generate_gnuplot_scripts", ListUtils::create<String>("true,false"));
+    registerOutputFile_("precursor_fit_out", "<file>", "", "Gaussian fit of the frequency of the deviations from the precursors. The options 'precursor_out' and 'fragment_out' must be set to take this effect.", false, true);
```
cbielow: canonically, this should be called `out_precursor_fit`.
The description should say `"Gaussian fit to the histogram of mass deviations from the precursors. The option 'precursor_out' must be set for this file to be generated."

The current behaviour is a bit unfortunate, because no file gets written if `precursor_out` (should be `out_precursor` but hey...) is not written. Usually TOPP tools do not have this weird dependency.
Solution: compute internal data if any of those two flags is requested.

src/utils/IDMassAccuracy.cpp
```diff
@@ -131,8 +131,11 @@ class TOPPIDMassAccuracy :
...
+    setValidFormats_("precursor_fit_out", ListUtils::create<String>("tsv"));
+
+    registerOutputFile_("fragment_fit_out", "<file>", "", "Gaussian fit of the frequency of fthe fragment ion m/z deviation. The options 'precursor_out' and 'fragment_out' must be set to take this effect.", false, true);
```
cbielow: same issue here

src/utils/IDMassAccuracy.cpp
```diff
@@ -459,31 +457,26 @@ class TOPPIDMassAccuracy :
       {
         gf.fit(values);
 
-
-        // write gnuplot script
-        if (generate_gnuplot_scripts)
+        // write gnuplot scripts
```
cbielow: not anymore :)

jpfeuffer:
Is this ready?

bernt-matthias:
Seems that IDMassAccuracy.cpp_cpplint_test failed. What does it mean?

jpfeuffer:
```
/home/travis/build/OpenMS/OpenMS/src/utils/IDMassAccuracy.cpp:320:  Missing space before {  [whitespace/braces] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/IDMassAccuracy.cpp:322:  Missing space before {  [whitespace/braces] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/IDMassAccuracy.cpp:417:  Missing space before {  [whitespace/braces] [5]
/home/travis/build/OpenMS/OpenMS/src/utils/IDMassAccuracy.cpp:419:  Missing space before {  [whitespace/braces] [5]
Done processing /home/travis/build/OpenMS/OpenMS/src/utils/IDMassAccuracy.cpp
Total errors found: 4
```

Code style

jpfeuffer:
Put the curly braces of the ifs onto the next line. And add whitespaces around binary operators (!=) in your case.

bernt-matthias:
Done. Note that there is no test for IDMassAccuracy and I also dit not test locally. Should be easy to add one (I have manual "curated" some basic tests for OpenMS binaries that currently have none in https://github.com/OpenMS/OpenMS/tree/08a521e5605b2b9c52fcb610475c4ef1a9d38451/src/tests/topp).

jpfeuffer:
Hmm yes. To be honest I agree with others that unless that tool uses the fitted curves to improve another algorithm, it is pretty useless on its own and can be done better with R scripts.
However we could keep it in case the general idea might be used for e.g. additional ID filtering or confidence adaption. But then tests are not urgent.
@timosachsenberg weren't we recently talking about it?

bernt-matthias:
I meant a test that the tool works at all :)

bernt-matthias:
For a test something along the lines of: 

```
IDMassAccuracy -in spectra.mzML -id_in MSGFPlusAdapter_1_out.idXML -out_precursor /tmp/p.csv -out_fragment /tmp/f.csv 
```

might work. But it raises `Unable to fit a Gaussian distribution to the precursor mass errors`. This we cant test the new outputs.

jpfeuffer:
Hmm that is a bit disappointing concerning the algorithm. 

Also the tool does not seem to consider the ability of some search engines to annotate precursors with isotope errors.

I guess it needs serious rework. I am fine with any kind of test. I am afraid it won't see much usage anyway.

timosachsenberg:
Yeah I think it would make sense to move the algorithm part to the QC modules Chris wrote and then have just a pyOpenMS script that calls the algorithm, writes output and optionally plots them.

jpfeuffer:
@bernt-matthias Can you run it in debug mode 1000 to see what the algorithm prints?

bernt-matthias:
Ahh.. with `-number_of_bins 10` it works. Probably the default of 100 bins makes no sense here.

jpfeuffer:
Yes probably too few data. Well, let's use this. Great.

Issue #2078, opened 2016-07-28
Metabolite spectral database file is missing from OpenMS KNIME extension


iprotsyuk:
MetaboliteSpectralMatcher always reports the following error message:

ERROR MetaboliteSpectralMatcher 0:388:394  Failing process stdout: [Progress of 'loading spectra list':, ,   59.62 %               , -- done [took 0.83 s (CPU), 0.82 s (Wall)] -- , Error: File not found (the file 'CHEMISTRY/MetaboliteSpectralDB.mzML' could not be found)]
ERROR MetaboliteSpectralMatcher 0:388:394  Failing process stderr: []
ERROR MetaboliteSpectralMatcher 0:388:394  Execute failed: Failed to execute node MetaboliteSpectralMatcher

Apparently, the file "CHEMISTRY/MetaboliteSpectralDB.mzML" is missing from the KNIME extension. After I downloaded from this repository and put to the OpenMS plugin folder, everything worked well.

timosachsenberg:
Hi,
the file is rather large so we have - so far - restrained from including it in the release.

iprotsyuk:
Hi Timo,
In this case, I'd suggest adding an additional parameter to the node: path to a file with a spectral library. The link to the missing HMDB library can be added in the node documentation so users can download it.
This apporach would also allow one to use their custom spectral library as it's fairly easy to combine a set of MS/MS spectra into an mzML file.

hroest:
I agree, I never liked the approach with the spectral db. We also do not ship out fasta files of different organisms in the `./share/OpenMS` folder. It should be the users responsibility to download these files and of course we could put them in a convenient location -- also note that giving a PATH would make it easier to add your own metabolites / update the file.

timosachsenberg:
adding a path  would be a good addition.
In the future, we could also think about a simple util that - on execution - pulls extra files from the internet.

timosachsenberg:
Adding the path should now be possible with this PR: https://github.com/OpenMS/OpenMS/pull/2526

hendrikweisser:
Fixed in #2526.

bernt-matthias:
Just stumbled over this again, since the default value of the `-database`  refers to a non-existent file. Would make more sense to have a mandatory argument without a default?

@iprotsyuk 

> After I downloaded from this repository and put to the OpenMS plugin folder, everything worked well.

What do you mean with "this repository"? I could not find anything here.

iprotsyuk:
[...]

PR #4468, opened 2020-01-10
MetaboliteSpectralMatcher: make database mandatory

bernt-matthias:
since the file is not included in the distribution
I guess its better to make the argument mandatory

see https://github.com/OpenMS/OpenMS/issues/2078

if at some point the data is available somewhere
(internet, submodule, ...) then it might be better
to have this info in the description?


oliveralka:
Just for your information, the file should be available in the "Example_Data":
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Tutorials/Data/latest/Example_Data/Metabolomics/databases/

PR #4469, opened 2020-01-10
ModificationsDB::getInstance performance fix

hroest:
this is a substantial improvement for performance on my machine, almost 100x fold for `static ModificationsDB* getInstance`. Basically on my Linux system the time before the fix was almost 100 ns and now is around 1-2ns -- it turns out this matters because people assumed `ModificationsDB::getInstance` is basically free and its all over the place. 

I am not sure why that is the case, there are no such issues with other singletons. It may be because of the default function arguments.

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -66,15 +66,23 @@ namespace OpenMS
       to search engines, e.g. Mascot.
 
       In some scenarios, it might be useful to define different modification
-      databases. This can be done by providing a path when initializing
-      ModificationsDB.
+      databases. This can be done by providing a path through
+      initializeModificationsDB(), however it is important that this is done
+      *before* the first call to getInstance().
   */
   class OPENMS_DLLAPI ModificationsDB
   {
 public:
 
     /// Returns a pointer to the modifications DB (singleton)
-    static ModificationsDB* getInstance(OpenMS::String unimod_file = "CHEMISTRY/unimod.xml", OpenMS::String psimod_file = "CHEMISTRY/PSI-MOD.obo", OpenMS::String xlmod_file = "CHEMISTRY/XLMOD.obo");
+    static ModificationsDB* getInstance()
+    {
+      static ModificationsDB* db_ = ModificationsDB::initializeModificationDB();
+      return db_;
+    }
+
+    /// Initializes the modification DB with non-default modification files (can only be done once)
+    static ModificationsDB* initializeModificationsDB(OpenMS::String unimod_file = "CHEMISTRY/unimod.xml", OpenMS::String psimod_file = "CHEMISTRY/PSI-MOD.obo", OpenMS::String xlmod_file = "CHEMISTRY/XLMOD.obo");
 
     /// Check whether ModificationsDB was instantiated before
     static bool isInstantiated();
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -75,11 +75,7 @@ namespace OpenMS
 public:
 
     /// Returns a pointer to the modifications DB (singleton)
-    static ModificationsDB* getInstance()
-    {
-      static ModificationsDB* db_ = ModificationsDB::initializeModificationsDB();
-      return db_;
-    }
+    static ModificationsDB* getInstance();
 
     /// Initializes the modification DB with non-default modification files (can only be done once)
     static ModificationsDB* initializeModificationsDB(OpenMS::String unimod_file = "CHEMISTRY/unimod.xml", OpenMS::String psimod_file = "CHEMISTRY/PSI-MOD.obo", OpenMS::String xlmod_file = "CHEMISTRY/XLMOD.obo");
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -77,7 +77,7 @@ public:
     /// Returns a pointer to the modifications DB (singleton)
     static ModificationsDB* getInstance()
     {
-      static ModificationsDB* db_ = ModificationsDB::initializeModificationDB();
+      static ModificationsDB* db_ = ModificationsDB::initializeModificationsDB();
       return db_;
     }
 
--- a/src/openms/source/CHEMISTRY/ModificationsDB.cpp
+++ b/src/openms/source/CHEMISTRY/ModificationsDB.cpp
@@ -50,8 +50,14 @@ namespace OpenMS
 {
   bool ModificationsDB::is_instantiated_ = false;
 
-  ModificationsDB* ModificationsDB::getInstance(OpenMS::String unimod_file, OpenMS::String psimod_file, OpenMS::String xlmod_file)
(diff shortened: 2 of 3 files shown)


jpfeuffer:
I think you have to move the implementation of the initialization to the cpp file though.
Based on my research during the last fix for the singletons, static local variables are only unique per _file_. This means if the header is included in multiple compilation units, you actually have two instances which leads to really weird errors.
I hope I understood and explained correctly. This insight helped me last time though.
See #4398 and maybe https://stackoverflow.com/questions/92546/variable-declarations-in-header-files-static-or-not/92641#92641

hroest:
>  Based on my research during the last fix for the singletons, static local variables are only unique per _file_. This means if the header is included in multiple compilation units, you actually have two instances which leads to really weird errors.

interesting, maybe we should move it to the cpp file. On the other hand, we do test this somewhat in the `OpenSwathAssayGenerator` where we use a different set of files for initialization. Also if that is true, we should look at these as well:

```
$ grep -R getInstance OpenMS/ | grep -v ';' | grep -v '//'
OpenMS/CONCEPT/GlobalExceptionHandler.h:      static GlobalExceptionHandler & getInstance()
OpenMS/TRANSFORMATIONS/FEATUREFINDER/IsotopeWavelet.h:    static IsotopeWavelet * getInstance()
OpenMS/CHEMISTRY/CrossLinksDB.h:    inline static CrossLinksDB* getInstance()
OpenMS/CHEMISTRY/ModificationsDB.h:      *before* the first call to getInstance().
OpenMS/CHEMISTRY/DigestionEnzymeDB.h:    static InstanceType* getInstance()
```

or maybe even these:

```
$ grep -R static OpenMS/ | grep -v ';' | grep -v '//' | grep '\*.*(' | grep -v create
OpenMS/ANALYSIS/ID/ConsensusIDAlgorithmPEPMatrix.h:    static inline const int* getData()
OpenMS/ANALYSIS/ID/ConsensusIDAlgorithmPEPMatrix.h:    static inline const int* getData()
OpenMS/CONCEPT/Factory.h:    static Factory * instance_()
OpenMS/CONCEPT/SingletonRegistry.h:    static SingletonRegistry * instance_()
OpenMS/CONCEPT/SingletonRegistry.h:    static FactoryBase * getFactory(const String & name)
OpenMS/TRANSFORMATIONS/FEATUREFINDER/IsotopeWavelet.h:    static IsotopeWavelet * getInstance()
OpenMS/CHEMISTRY/CrossLinksDB.h:    inline static CrossLinksDB* getInstance()
OpenMS/CHEMISTRY/DigestionEnzymeDB.h:    static InstanceType* getInstance()
```

jpfeuffer:
You are right. And I think actually the initialization step helped avoiding this problem already. Still I guess it is cleaner in the cpp.
Thanks!

hroest:
So the main reason that this is rather slow is the default parameters. I compared the following code on godbolt:

## Code 1

```
// Type your code here, or load an example.
#include <string>
#include <utility>
#include <iostream>
#include <sstream>
#include <cmath>
#include <vector>

void doWork();

class ModificationsDB
{
    public:
    ModificationsDB() {}
    ModificationsDB(std::string a, std::string b, std::string c)
    {
        if (!a.empty())
        {
            doWork();
        }
    }

    static ModificationsDB* getInstance_(std::string unimod_file = "CHEMISTRY/unimod.xml", std::string psimod_file = "CHEMISTRY/PSI-MOD.obo", std::string xlmod_file = "CHEMISTRY/XLMOD.obo")
    {
      static ModificationsDB* db_ = new ModificationsDB(unimod_file, psimod_file, xlmod_file);
      return db_;
    }
};


ModificationsDB* test(std::string residue)
{
    //return ModificationsDB::getInstance();
}

ModificationsDB* test2(std::string residue)
{
    return ModificationsDB::getInstance_();
}

int main()
{
    test("test");
    test2("test");
}

```

which produces something similar to

```
_ZN15ModificationsDB11getInstanceEv:
        movzx   eax, BYTE PTR _ZGVZN15ModificationsDB11getInstanceEvE3db_[rip]
        test    al, al
        je      .L19
        mov     rax, QWORD PTR _ZZN15ModificationsDB11getInstanceEvE3db_[rip]
        ret
```
which is rather simple and basically moves the db_ ptr to eax and then checks it for zero and if its not zero, moves it to `rax` the return register and returns it. As simple as it gets (of course if it is zero, then it will jump to `.L19` and do the work required there). That is basically what we want.


## Code 2
On the other hand, with the following code 

```
// Type your code here, or load an example.
#include <string>
#include <utility>
#include <iostream>
#include <sstream>
#include <cmath>
#include <vector>

void doWork();

class ModificationsDB
{
    public:
    ModificationsDB() {}
    ModificationsDB(std::string a, std::string b, std::string c)
    {
        if (!a.empty())
        {
            doWork();
        }
    }

    static ModificationsDB* getInstance_(std::string unimod_file = "CHEMISTRY/unimod.xml", std::string psimod_file = "CHEMISTRY/PSI-MOD.obo", std::string xlmod_file = "CHEMISTRY/XLMOD.obo")
    {
      static ModificationsDB* db_ = new ModificationsDB(unimod_file, psimod_file, xlmod_file);
      return db_;
    }
};


ModificationsDB* test(std::string residue)
{
    //return ModificationsDB::getInstance();
}

ModificationsDB* test2(std::string residue)
{
    return ModificationsDB::getInstance_();
}

int main()
{
    test("test");
    test2("test");
}
```

this produces quite a lot of code especially when calling test2:

```
test2(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >):
        pushq   %r15
        xorl    %edx, %edx
        pushq   %r14
        pushq   %r13
        pushq   %r12
        pushq   %rbp
        pushq   %rbx
        subq    $216, %rsp
        leaq    176(%rsp), %rsi
        leaq    80(%rsp), %rdi
        movq    $19, 176(%rsp)
        leaq    96(%rsp), %rbx
        leaq    64(%rsp), %rbp
        movq    %rbx, 80(%rsp)
        call    std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_create(unsigned long&, unsigned long)
        movq    176(%rsp), %rdx
[ ... ]
```
a total of 66 instructions with 3 allocations of strings which is required to set up the three strings in any case and then an explicit function call at the end to `ModificationsDB::getInstance_`

```
ModificationsDB::getInstance_(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >): # @ModificationsDB::getInstance_(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)
        pushq   %rbp
        pushq   %r15
        pushq   %r14
        pushq   %r13
        pushq   %r12
        pushq   %rbx
        subq    $120, %rsp
        movb    guard variable for ModificationsDB::getInstance_(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)::db_(%rip), %al
        testb   %al, %al
        je      .LBB2_1
.LBB2_45:
```

## Results

I found that the more simple Code 1 produced a very straight forward machine code with about 5 instructions to return the static ptr whereas Code 2 produced a lot of code that included setting up all three strings even if they were not required

PR #4470, opened 2020-01-12
make diff parameter order consistent in tests

bernt-matthias:
I guess this won't hurt .. but help me a bit


cbielow:
This was actually a small bug. Creating to files in the Src tree is not intended.

PR #4473, opened 2020-01-14
[Fix] bug/TODO for reading full TITLE from mgf

jpfeuffer:


hroest:
@timosachsenberg any tests?

Issue #4478, opened 2020-01-15
suffix _REVERSED not automatically recognized by PeptideIndexer as Decoy position and string


timosachsenberg:
_REVERSED

jpfeuffer:
add to:
https://github.com/OpenMS/OpenMS/blob/b1bde1639b16b2298fd363c0ff4897e2f2f9c490/src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h#L363

but is it really a suffix? Sometimes external tools put it after the database part but before the annotation
e.g. "sp|foo_rev|baz"
I remember that we currently do not handle this.

timosachsenberg:
Just came up in one example file. I will add it because it might not hurt

PR #4480, opened 2020-01-15
Update FASTAContainer.h

timosachsenberg:
as it came up in one of our fasta files at the hackathon


timosachsenberg:
https://github.com/OpenMS/OpenMS/issues/4478

PR #4482, opened 2020-01-16
[Improvement] FeatureFindingMetabo improved isotope detection

marcus-ludwig:
Isotope detection for metabolites is currently suboptimal. I want to propose a different way to group isotopes.

**The problem**: Especially for cases in which the isotope pattern is very informative, such as for sulfur-containing compounds, the FeatureFindingMetabo does not group all isotope peaks.
This is because the assumed m/z distances for isotopes are not precise. 
Currently there are 2 options:
1. a C13-C12 difference. Clearly a oversimplification, since all elements have different isotope m/z distances.
2. m/z distances computed based on known molecules. I think the main problem is the assumption that the distance of _monoisotopic to +2 isotope_ peak is 2 times the distance of _monoisotopic to +1 isotope peak_; and also that compounds with uncommon elements are too different from the used average.

**My suggestion:** 
Plausible ranges for isotope peaks can be computed based on the set of expected elements for the specific dataset. This is what we are doing for isotope detection in SIRIUS.

I currently implemented it the following way:
* for a given monoisotopic m/z, the set of elements and the isotope position an expected m/z range is computed
* scoring:
  * if the assessed mass trace lies within this range, it is scored with 1.0 (best possible)
  * if it lies outside this range, the distance to the boundary is scored based on the expected m/z error the same way it is currently done with the other methods.

This means as long as the m/z value lies in the expected range the mass trace is only scored by the peak shape similarity over retention time  (the cosine score).

**I want to put this to discussion:**
Is it sufficient to score mass traces based on retention time only as long as the m/z looks reasonable? I know OpenMS does not allow for multiple hypothesis, such as "this mass trace is either a monoisotopic peak or the isotope peak of another compound" and hence needs a way to decide. But peak shape should hopefully be informative enough.

Isotope detection is also related to the computation of mass trace overlap which is discussed here: #4483.


timosachsenberg:
Cool. Sounds good and probably works well on high-res data. We will ultimately know only if we use it on several real life datasets. @oliveralka do you have some metabolomics data for that?

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
```diff
@@ -139,6 +140,21 @@ namespace OpenMS
 
   };
 
+  /**
+   * @brief Internal class to store a lower and upper bound of an m/z range
+   */
+  class OPENMS_DLLAPI  Range
```
timosachsenberg: I would use a simple struct: 
struct OPENMS_DLLAPI Range
{
  double low = 0;
  double high = 0;
};

oliveralka:
@marcus-ludwig Thanks a lot! I will take a look and test it with some metabolomics data!

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
```diff
@@ -182,6 +198,22 @@ namespace OpenMS
     void updateMembers_() override;
 
 private:
+    /**
+     * @brief parses a string of element symbols into a vector of Elements
+     * @param elements_string string of element symbols without whitespaces or commas. e.g. CHNOPSCl
+     * @return vector of Elements
+     */
+    std::vector<const Element*> elements_from_string_(const std::string& elements_string) const;
```
timosachsenberg: elementsFromString_ would be more style conform

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
```diff
@@ -182,6 +198,22 @@ namespace OpenMS
...
+    /**
+     * Calculate for a given alphabet the maximal and minimal mass defects of isotopes.
+     *
+     * @param alphabet   chemical alphabet (elements which are expected to be present)
+     * @param monomz     m/z of monoisotopic peak
+     * @param peakOffset integer distance between isotope peak and monoisotopic peak (minimum: 1)
+     * @return an interval which should contain the isotopic peak
+     */
+    Range getTheoreticIsotopicMassWindow_(std::vector<const Element*> alphabet, double monomz, int peakOffset) const;
```
timosachsenberg: pass alphabet as const reference

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
```diff
@@ -227,8 +259,12 @@ namespace OpenMS
      *
      * Reference: Kenar et al., doi: 10.1074/mcp.M113.031278
      *
+     * an alternative scoring was added which considers all possible m/z
+     * differences of all isotopes of a set of elements. The mass trace
+     * is positively scored if it could be an isotope based on the given elements.
+     *
     */
-    double scoreMZ_(const MassTrace &, const MassTrace &, Size isotopic_position, Size charge) const;
+    double scoreMZ_(const MassTrace &, const MassTrace &, Size isotopic_position, Size charge, std::vector<const Element*> elements) const;
```
timosachsenberg: maybe use separate functions for the different scoring approaches?

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
```diff
@@ -282,10 +318,12 @@ e conditions are fulfilled. Mainly the
     bool use_smoothed_intensities_;
     
     bool use_mz_scoring_C13_;
+    bool use_mz_scoring_by_element_range;
     bool report_convex_hulls_;
     bool report_chromatograms_;
 
     bool remove_single_traces_;
+    std::vector<const Element*> elements_;
```
timosachsenberg: how are these set?

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -285,6 +287,23 @@ namespace OpenMS
     return feat_score_;
   }
 
+  Range::Range(double left_boundary, double right_boundary)
```
timosachsenberg: implementation not needed if struct used

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -346,6 +371,42 @@ namespace OpenMS
...
 
     remove_single_traces_ = param_.getValue("remove_single_traces").toBool();
+
+    use_mz_scoring_by_element_range = param_.getValue("mz_scoring_by_elements").toBool();
+    std::string elements_list_ = param_.getValue("elements");
+    elements_ = elements_from_string_(elements_list_);
+  }
+
+
+  std::vector<const Element*> FeatureFindingMetabo::elements_from_string_(const std::string&elements_string) const
```
timosachsenberg: maybe use parsing of EmpiricalFormula directly

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -504,7 +565,7 @@ namespace OpenMS
     }
   }
 
-  double FeatureFindingMetabo::scoreMZ_(const MassTrace& tr1, const MassTrace& tr2, Size iso_pos, Size charge) const
+  double FeatureFindingMetabo::scoreMZ_(const MassTrace& tr1, const MassTrace& tr2, Size iso_pos, Size charge, std::vector<const Element*> elements) const
```
timosachsenberg: better only calculate mass range once -> no passing of elements required

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -316,6 +335,12 @@ namespace OpenMS
     defaults_.setValue("remove_single_traces", "false", "Remove unassembled traces (single traces).");
     defaults_.setValidStrings("remove_single_traces", ListUtils::create<String>("false,true"));
 
+    //TODO does this make other parameters obsolete, or do we still want to support all different options?
+    defaults_.setValue("mz_scoring_by_elements", "false", "Use the m/z range of the assumed elements to detect isotope peaks. A expected m/z range is computed from the isotopes of the assumed elements. If enabled, this ignores 'mz_scoring_13C'");
```
timosachsenberg: better to have the user select from a list of methods instead of exclusive flags

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
marcus-ludwig: I moved the parts which are different for the scorings into separate methods.

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -346,6 +351,20 @@ namespace OpenMS
...
+    use_mz_scoring_by_element_range = param_.getValue("mz_scoring_by_elements").toBool();
+    std::string elements_list_ = param_.getValue("elements");
+    elements_ = elementsFromString_(elements_list_);
+  }
+
+
+  std::vector<const Element*> FeatureFindingMetabo::elementsFromString_(const std::string& elements_string) const
+  {
+    std::vector<const Element*> elements;
+    for (auto element_with_amount : EmpiricalFormula(elements_string)) {
```
hroest:
```suggestion
    for (const auto& element_with_amount : EmpiricalFormula(elements_string))
    {
```

oliveralka:
rebuild jenkins

jpfeuffer:
For easy access (the remaining errors):
CodeStyle:
```
/home/travis/build/OpenMS/OpenMS/src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp:546:  Missing space before {  [whitespace/braces] [5]
/home/travis/build/OpenMS/OpenMS/src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp:607:  If an else has a brace on one side, it should have it on both  [readability/braces] [5]
/home/travis/build/OpenMS/OpenMS/src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp:614:  If an else has a brace on one side, it should have it on both  [readability/braces] [5]
Done processing /home/travis/build/OpenMS/OpenMS/src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
Total errors found: 3
```

Compilation:
```
/.../OpenMS/src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp: In member function ‘OpenMS::Range OpenMS::FeatureFindingMetabo::getTheoreticIsotopicMassWindow_(std::vector<const OpenMS::Element*>, int) const’:
src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp:709:25: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
       for (int k = 1; k < iso.size(); ++k) {
```

oliveralka:
I think this should be it for this PR.

Maybe we should add that it is better to set 
` -algorithm:ffm:isotope_filtering_model <choice> = none`
when using the "element" method. 

We can discuss this and I will play around with the other suggestions (#4483) and open a new PR.

timosachsenberg:
Yes we should warn that the isotope model is not supported

oliveralka:
rebuild jenkins

jpfeuffer:
New commits automatically cancel and restart old PR jobs.

oliveralka:
You are right - sorry - thanks!

oliveralka:
In general as already mentioned in the class test, the simulated data should probably be replaced by a slice of real data. 

The feature reduction from 81 to 80 features when using the element scoring is based on a few traces which could potentially be two features, which is hard to say since the feature is cut off in retention time. The peak apex of one of the "isotope" peaks might be a little bit off, so I guess the implementation of the grouping based on the element scoring method and in peak apex in addition, as suggested by Marcus is a good idea.

timosachsenberg:
if it is simulated these might be indeed two features... let's merge

PR #4488, opened 2020-01-19
[BUILD] disable python memory leak tests

hroest:
due to many false positives and little utility

@jpfeuffer can you also disable them in the docker build?


.travis.yml
```diff
@@ -79,7 +79,7 @@ matrix:
       compiler: gcc
 
 ## pyopenms test
-    - env: ENABLE_STYLE_TESTING=OFF WITH_GUI=OFF ENABLE_TOPP_TESTING=OFF ENABLE_CLASS_TESTING=OFF ADDRESS_SANITIZER=Off BUILD_TYPE=Release PYOPENMS=ON PY_NO_OPTIMIZATION=ON OPENMP=Off
+    - env: ENABLE_STYLE_TESTING=OFF WITH_GUI=OFF ENABLE_TOPP_TESTING=OFF ENABLE_CLASS_TESTING=OFF ADDRESS_SANITIZER=Off BUILD_TYPE=Release PYOPENMS=ON PY_NO_OPTIMIZATION=ON OPENMP=Off -PY_MEMLEAK_DISABLE=On
```
jpfeuffer: Can you check if the hyphen is actually needed and does what it is supposed to do?

hroest: not sure, probably doesnt do much :-)

jpfeuffer:
I deactivated them in the docker build as well now.

Issue #4485, opened 2020-01-17
MaRaClusterAdapter provide option for writing the original output


timosachsenberg:
Provide optional outputfile parameter to keep the MaRaCluster output file: 
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/MaRaClusterAdapter.cpp#L366

Leon-Bichmann:
Maracluster has two modes of running batch and consensus, which files would you like to keep:

`echo "Clustering spectra"
maracluster batch -b file_list.txt -f ./clusters -a sample -t -10 -c -10,-15,-20

echo "Creating consensus spectra"
maracluster consensus -l ./clusters/sample.clusters_p10.tsv -f ./consensus -o ./consensus/sample_consensus.ms2
`

Leon-Bichmann:
In case both, I would suggest to register two string options for the /clusters and /consensus directories and then copy the temp directories there.

Would this work?
```
String temp_directory_body = makeAutoRemoveTempDirectory_();

File::copyDirRecursively(temp_directory.toQString(), maracluster_cluster_directory.toQString());
```

Leon-Bichmann:
#4490

PR #4490, opened 2020-01-21
MaRaCluster: keep original output in temp directory

Leon-Bichmann:
This PR provides the option to keep the MaRacluster original output stored in a temp directory in a designated output directory.

### DIFF
--- a/src/topp/MaRaClusterAdapter.cpp
+++ b/src/topp/MaRaClusterAdapter.cpp
@@ -287,7 +287,6 @@ protected:
     const String maracluster_executable(getStringOption_("maracluster_executable"));
     writeDebug_(String("Path to the maracluster executable: ") + maracluster_executable, 2);
 
-    String maracluster_output_directory = getStringOption_("output_directory");
     String maracluster_output_directory = getStringOption_("output_directory");   
     const String consensus_out(getStringOption_("consensus_out"));
     const String out(getStringOption_("out"));
--- a/src/topp/MaRaClusterAdapter.cpp
+++ b/src/topp/MaRaClusterAdapter.cpp
@@ -372,9 +372,6 @@ protected:
     //if specified keep original output in designated directory
     if (!maracluster_output_directory.empty())
     {
-      // convert path to absolute path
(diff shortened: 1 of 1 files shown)


src/topp/MaRaClusterAdapter.cpp
```diff
@@ -366,6 +369,24 @@ class MaRaClusterAdapter :
     readMClusterOutputAsMap_(consensus_output_file, specid_to_clusterid_map, filename_to_file_idx);
     file_idx = 0;
 
+    //if specified keep original output in designated directory
+    if (!maracluster_output_directory.empty())
+    {
+      // convert path to absolute path
+      QDir mara_out_dir(maracluster_output_directory.toQString());
```
timosachsenberg: is this conversion to absolutePath needed? 
File::copyDirRecursively in 
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/SYSTEM/File.cpp
seems to do something similar though I don't know in detail how that works. Can you check?

Leon-Bichmann: I'll check, it is the same code as in the SiriusAdapter so I guess it works.

oliveralka: I guess it should not be needed any more since it should be handled by the function directly. I think this was introduced in the SiriusAdapter and then refactored into File.cpp - I guess we forgot to remove it in the SiriusAdapter.

```
    QString canonical_source_dir = source_dir.canonicalPath();
    QString canonical_target_dir = target_dir.canonicalPath();
```

Leon-Bichmann: ok so when removing .absolutePath() compiling fails as it can't compile QDir to String

timosachsenberg: Why QDir? Should work without it.

Leon-Bichmann: ok, tests passed. Seems like that was used for another purpose in SiriusAdapter

timosachsenberg:
Thanks!

oliveralka:
#4485

PR #4491, opened 2020-01-21
[FEATURE] MascotGenericFile: keep TITLE info when reading and writing files

enetz:
The content of the TITLE field of MGF files is not strictly defined and the only place where some additional free text info can be stored.
Previously OpenMS discarded that info when reading and then writing out MGF.
It was also not possible to add any info to it in an OpenMS tool or in pyOpenMS.

The changes here will write out the TITLE info, if it is available, and add the nativeID to it to make the new titles unique. Previously only the nativeID was written out as the TITLE.

This assumes, that a meta value of a spectrum called "TITLE" was read in from an MGF, or created for writing out MGF by a tool on purpose and is unlikely to be created in another way.

We encountered this problem in a hackathon recently, where some important info was lost after reading, processing and writing MGF using pyOpenMS, although that info was available and easily accessible as a MetaValue in python.


src/openms/source/FORMAT/MascotGenericFile.cpp
```diff
@@ -337,34 +337,66 @@ namespace OpenMS
       os << "BEGIN IONS\n";
       if (!store_compact_)
       {
-        os << "TITLE=" << precisionWrapper(mz) << "_" << precisionWrapper(rt)
-           << "_" << spec.getNativeID() << "_" << filename << "\n";
+        // if a TITLE is available, it was (most likely) parsed from an MGF file
+        // in that case, a nativeID was constructed from the TITLE and the spectrum index
+        if (spec.metaValueExists("TITLE"))
+        {
+          if (String(spec.getMetaValue("TITLE")).hasSubstring("_index="))
```
cbielow: is `_index` canonical? Could it be something else?

src/openms/source/FORMAT/MascotGenericFile.cpp
```diff
@@ -337,34 +337,66 @@ namespace OpenMS
...
+        }
       }
       else
       {
-        os << "TITLE=" << fixed << setprecision(HIGH_PRECISION) << mz << "_"
-           << setprecision(LOW_PRECISION) << rt << "_"
-           << spec.getNativeID() << "_" << filename << "\n";
+        // if a TITLE is available, it was (most likely) parsed from an MGF file
+        // in that case, a nativeID was constructed from the TITLE and the spectrum index
+        if (spec.metaValueExists("TITLE"))
```
cbielow: This code duplication is quite ugly.
The `store_compact_` flag somewhat provokes this, but can you just wrap the behavior of this flag into an inline function?! would make the code much more maintainable

src/openms/source/FORMAT/MascotGenericFile.cpp
enetz: `index=spectrum_number` is the native ID, defined by the mzTab and mzIdentML spectrum Id specification for MGFs. I added this to make sure all titles are unique, even if the loaded MGF had non-unique titles. Also this way the nativeIDs in ID files can somewhat be mapped back to the titles without relying purely on the order of spectra. The separator `_`was chosen arbitrarily and could be something else.

Here I just test whether the TITLE was already generated with the `else` case with `TITLE + "_" + nativeID`, meaning the loaded MGF was probably written out by OpenMS previously and already has the index. This is to prevent adding the index multiple times when processing MGFs with multiple OpenMS tools.

The nativeID is generated while reading in an MGF, so potentially it can be changed before writing the data out again, but usually OpenMS tools are not supposed overwrite nativeIDs of spectra.

Another option would be to make the new TITLE while reading in the file with `TITLE = TITLE + "_" + nativeID` and then writing out only the TITLE meta value. That way OpenMS tools or a pyOpenMS script could define their own titles freely by overwriting this, if necessary.

src/openms/source/FORMAT/MascotGenericFile.cpp
jpfeuffer: @enetz Can you "quickly" do that if you agree?

enetz: @jpfeuffer the code duplication that @cbielow mentioned was already reduced by a previous commit.
The remaining duplicated code here is just this:
       ```
 if (spec.metaValueExists("TITLE"))
        {
          os << "TITLE=" << spec.getMetaValue("TITLE") << "\n";
        }
        else
```
and the `else` cases are different

PR #4492, opened 2020-01-21
some minor fixes

cbielow:
PP-HiRes now gives the user some feedback of which MS-levels were actually picked in `auto-mode`.

```
11:10:22 NOTICE: PeakPickerHiRes of node #4 started. Processing ...
Picked spectra by MS-level:
  MS-level 1: 8259 / 8259
  MS-level 2: 0 / 7482
```

### DIFF
--- a/src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
+++ b/src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
@@ -402,7 +402,14 @@ namespace OpenMS
     pickExperiment(input, output, boundaries_spec, boundaries_chrom, check_spectrum_type);
   }
 
-  void PeakPickerHiRes::pickExperiment(const PeakMap& input, PeakMap& output, 
+  struct SpectraPickInfo
+  {
+    int picked{0}; //< number of picked spectra
+    int total{0}; //< overall number of spectra
+  };
+
+  void PeakPickerHiRes::pickExperiment(const PeakMap& input,
+                                       PeakMap& output, 
                                        std::vector<std::vector<PeakBoundary> >& boundaries_spec, 
                                        std::vector<std::vector<PeakBoundary> >& boundaries_chrom,
                                        const bool check_spectrum_type) const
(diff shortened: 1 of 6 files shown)


src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
```diff
@@ -402,7 +402,14 @@ namespace OpenMS
     pickExperiment(input, output, boundaries_spec, boundaries_chrom, check_spectrum_type);
   }
 
-  void PeakPickerHiRes::pickExperiment(const PeakMap& input, PeakMap& output, 
+  struct SpectraPickInfo
+  {
+    int picked{0}; //< number of picked spectra
```
timosachsenberg: why int?

src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
```diff
@@ -445,18 +457,18 @@ namespace OpenMS
         else
         {
           std::vector<PeakBoundary> boundaries_s; // peak boundaries of a single spectrum
-
-                                                  // determine type of spectral data (profile or centroided)
-          SpectrumSettings::SpectrumType spectrum_type = input[scan_idx].getType();
-
+          SpectrumSettings::SpectrumType spectrum_type = input[scan_idx].getType(true);
```
timosachsenberg: Maybe add in comment what the true means

src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
```diff
@@ -402,7 +402,14 @@ namespace OpenMS
     pickExperiment(input, output, boundaries_spec, boundaries_chrom, check_spectrum_type);
   }
 
-  void PeakPickerHiRes::pickExperiment(const PeakMap& input, PeakMap& output, 
+  struct SpectraPickInfo
+  {
+    uint32_t picked{0}; //< number of picked spectra
```
hroest:
```suggestion
    uint32_t picked{0}; ///< number of picked spectra
```

src/openms/source/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.cpp
```diff
@@ -402,7 +402,14 @@ namespace OpenMS
...
+    uint32_t total{0}; //< overall number of spectra
```
hroest:
```suggestion
    uint32_t total{0}; ///< overall number of spectra
```

cbielow: good catch... I always get this wrong...

PR #4493, opened 2020-01-21
[Fix] check only db filename during merge

jpfeuffer:
some workflow engine might symlink or copy databases around.
Better to just check the basename during merging.

### DIFF
--- a/src/openms/include/OpenMS/METADATA/ProteinIdentification.h
+++ b/src/openms/include/OpenMS/METADATA/ProteinIdentification.h
@@ -254,6 +254,7 @@ public:
 
       std::pair<int,int> getChargeRange() const;
       int getChargeValue_(String& charge_str) const;
+      bool mergeable(const ProteinIdentification::SearchParameters& sp, const String& experiment_type) const;
 
     };
 
@@ -406,6 +407,7 @@ public:
     /// if this object has inference data
     bool hasInferenceData() const;
     bool hasInferenceEngineAsSearchEngine() const;
+    bool peptideIDsMergeable(const ProteinIdentification& idRun, const String& experiment_type) const;
     //@}
 
 protected:
--- a/src/openms/include/OpenMS/METADATA/ProteinIdentification.h
+++ b/src/openms/include/OpenMS/METADATA/ProteinIdentification.h
@@ -252,10 +252,17 @@ public:
 
       bool operator!=(const SearchParameters& rhs) const;
 
+      /// returns the charge range from the search engine settings as a pair of ints
       std::pair<int,int> getChargeRange() const;
-      int getChargeValue_(String& charge_str) const;
+
+      /// Tests if these search engine settings are mergeable with @param sp
+      /// depending on the given @param experiment_type.
+      /// Modifications are compared as sets. Databases based on filename.
+      /// "labeled_MS1" experiments additionally allow different modifications.
       bool mergeable(const ProteinIdentification::SearchParameters& sp, const String& experiment_type) const;
 
+      private:
+      int getChargeValue_(String& charge_str) const;
     };
 
     /** @name Constructors, destructors, assignment operator <br> */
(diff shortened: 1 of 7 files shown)


src/openms/source/ANALYSIS/ID/ConsensusMapMergerAlgorithm.cpp
```diff
@@ -32,11 +32,10 @@
 // $Authors: Julianus Pfeuffer $
 // --------------------------------------------------------------------------
 
+#include <include/OpenMS/ANALYSIS/ID/ConsensusMapMergerAlgorithm.h>
```
cbielow: `include/OpenMS` ?

jpfeuffer: hmm good question. just moved it without thinking.

src/tests/class_tests/openms/source/IDMergerAlgorithm_test.cpp
```diff
@@ -38,6 +38,8 @@
 #include <OpenMS/FORMAT/IdXMLFile.h>
 #include <OpenMS/SYSTEM/StopWatch.h>
 #include <OpenMS/test_config.h>
+#include <QFile>
```
cbielow: remnants from testing? Not used?

jpfeuffer: wow quick spotting. yes, found it as well.

src/openms/source/ANALYSIS/ID/ConsensusMapMergerAlgorithm.cpp
```diff
@@ -377,9 +376,13 @@ namespace OpenMS
         break;
       }
       const ProteinIdentification::SearchParameters& sp = idRun.getSearchParameters();
+      String spdb = sp.db;
```
cbielow: seeing this code duplicated 4x and the code below (which looks like a tolerant version of `operator==` for ProteinIdentification::SearchParameters): can we please move this whole comparison to a member of ProteinIdentification::SearchParameters?!

jpfeuffer:
By the way, I was kind of hoping that our File classes abstract the handling of native separators but it seems like reading a string like "C:\foo\bar.txt" on non-Windows machines is tricky.
Workarounds exist in other places as well.

src/openms/source/ANALYSIS/ID/ConsensusMapMergerAlgorithm.cpp
jpfeuffer: true. will do.

src/openms/source/METADATA/ProteinIdentification.cpp
```diff
@@ -448,6 +492,34 @@ namespace OpenMS
         se == "ProteinInference";
   }
 
+  bool ProteinIdentification::peptideIDsMergeable(const ProteinIdentification& idRun, const String& experiment_type) const
```
cbielow: should be `id_run`

src/openms/source/ANALYSIS/ID/ConsensusMapMergerAlgorithm.cpp
```diff
@@ -359,140 +357,19 @@ namespace OpenMS
...
-      {
-        //TODO actually introduce a flag for labelling modifications in the Mod datastructures?
-        //OR put a unique ID for the used mod as a UserParam to the mapList entries (consensusHeaders)
-        //TODO actually you would probably need an experimental design here, because
-        //settings have to agree exactly in a FractionGroup but can slightly differ across runs.
-        //Or just ignore labelling mods during the check
-       OPENMS_LOG_WARN << "Used modification settings from IDRun " + idRun.getIdentifier() + " does not match with the others."
-        " Although it seems to be an MS1-labeled experiment, check carefully that only non-labelling mods differ.";
-      }
+      ok = ok && ref.peptideIDsMergeable(idRun, experiment_type);
```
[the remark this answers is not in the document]
cbielow: ahh.. I see now... you want to see all the warnings.
Maybe just mention that in a comment.

src/openms/include/OpenMS/METADATA/ProteinIdentification.h
```diff
@@ -254,6 +254,7 @@ namespace OpenMS
 
       std::pair<int,int> getChargeRange() const;
       int getChargeValue_(String& charge_str) const;
```
cbielow: side question: why does this look like a private member?

PR #4494, opened 2020-01-22
Feature/pydoc

hroest:
adds the following documentation to all classes:

```
Documentation is available at " + classdocu_base + "class%(namespace)s_1_1%(cpp_name)s.html
```

so that the class documentation reads as follows:

```
cdef class FloatDataArray:
    """   
    Cython implementation of _FloatDataArray

    Documentation is available at http://www.openms.de/current_doxygen/html/classOpenMS_1_1DataArrays_1_1FloatDataArray.html

     -- Inherits from ['MetaInfoDescription']

    The representation of extra float data attached to a spectrum or chromatogram.
    Raw data access is proved by `get_peaks` and `set_peaks`, which yields numpy arrays
    """
```

NOTE: We will have to adjust the URL to point to the release documentation

@jpfeuffer do we have a stable redirect from the release docu that we can use? E.g. http://www.openms.de/release24_doxygen or something like that?

### DIFF
--- a/src/pyOpenMS/addons/AASequence.pyx
+++ b/src/pyOpenMS/addons/AASequence.pyx
@@ -5,3 +5,18 @@
         for k,v in mmap.iteritems():
             c_mmap[ _String(<char *>k) ] = v # add <size_t> ?
         self.inst.get().getAAFrequencies(c_mmap)
+
+    def __iter__(self):
+        cdef int n = self.inst.get().size()
+        cdef int i = 0
+
+        cdef Residue py_result 
+        while i < n:
+
+            py_result = Residue.__new__(Residue)
+            py_result.inst = shared_ptr[_Residue](new _Residue( deref(self.inst.get())[i] ))
+
+            yield py_result
+
+            i += 1
+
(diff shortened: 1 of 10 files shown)


hroest:
PS: depends on https://github.com/uweschmitt/autowrap/pull/87

jpfeuffer:
Hi! www.openms.de/documentation/*
should dynamically redirect to the html folder of the latest release docu.
E.g. www.openms.de/documentation/TOPP_documentation.html
will lead to the latest TOPP docu.
Does that help?

hroest:
>  Does that help?

yes, but it would be even better to have stable releases. This is because we will build pyOpenMS 2.3 and pyOpenMS 2.4 and it would be important to have the right pyOpenMS version point to the matching documentation

jpfeuffer:
OK I added a new redirect rule:
`www.openms.de/doxygen/(.*)`
to
`https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/$1``

see e.g. http://www.openms.de/doxygen/release/2.3.0/html/index.html

jpfeuffer:
The documentation folders will have the same name as the github branches.

timosachsenberg:
hi pyopenms passed - so can we merge that for the release (even if we don't use the new autowrap yet=?

PR #4497, opened 2020-01-22
[FIX] avoid stall on Windows when TOPPAS calls a TOPP tool with a ver…

cbielow:
…y long list of input files (exceeding the 8k length limit)

Fixed by moving the list of files into the .ini file.

### DIFF
--- a/src/openms_gui/source/VISUAL/TOPPASToolVertex.cpp
+++ b/src/openms_gui/source/VISUAL/TOPPASToolVertex.cpp
@@ -34,6 +34,7 @@
 
 #include <OpenMS/VISUAL/TOPPASToolVertex.h>
 
+#include <OpenMS/CONCEPT/LogStream.h>
 #include <OpenMS/FORMAT/FileHandler.h>
 #include <OpenMS/FORMAT/ParamXMLFile.h>
 #include <OpenMS/SYSTEM/File.h>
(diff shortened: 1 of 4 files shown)


src/openms_gui/source/VISUAL/TOPPASToolVertex.cpp
```diff
@@ -525,19 +525,22 @@ namespace OpenMS
 
         String param_name = in_params[param_index].param_name;
 
+        const QStringList& file_list = ite->second.filenames.get();
+
         bool store_to_ini = false;
         // check for GenericWrapper input/output files and put them in INI file:
-        if (param_name.hasPrefix("ETool:"))
+        // OR if there are a lot of input files (which might exceed the 8k length limit of cmd.exe on Windows)
+        if (param_name.hasPrefix("ETool:") || file_list.size() > 10)
```
timosachsenberg: can you make the threshold a constant?

timosachsenberg: as it is reused in other parts of the code

cbielow:
done

timosachsenberg:
src/openms_gui/source/VISUAL/TOPPASToolVertex.cpp:539:64: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

PR #4504, opened 2020-01-29
Update install-linux.doxygen

timosachsenberg:


hroest:
should not be necessary of `libhdf5-dev` is installed ?

timosachsenberg:
I'll try again on google cloud

PR #4506, opened 2020-01-31
MapAlignerTreeGuided: new tool

jthueringer:
As requested in pull request #4418 
* moved functions to a class inside the library
* tests provided


src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,175 @@
...
+    ~MapAlignmentAlgorithmTreeGuided() override;
+
+    /**
+     * @brief Extract RTs given for individual features of each map, calculate distances for each pair of maps and cluster hierarchical using average linkage.
+     *
+     * @param feature_maps Vector of input maps (FeatureMap) whose distance is to be calculated.
+     * @param tree Vector of BinaryTreeNodes that will be computed
+     * @param maps_ranges Vector to store all sorted RTs of extracted identifications for each map in @p feature_maps; needed to determine the 10/90 percentiles
+    */
+    static void buildTree_(std::vector<FeatureMap>& feature_maps, std::vector<BinaryTreeNode>& tree, std::vector<std::vector<double>>& maps_ranges);
```
timosachsenberg: the underscore usually indicates that this is a protected or private member. Here it is in public namespace

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * @brief Align feature maps tree guided using align() of @ref OpenMS::MapAlignmentAlgorithmIdentification and use TreeNode with larger 10/90 percentile range as reference.
+     *
+     * @param tree Vector of BinaryTreeNodes that contains order for alignment.
+     * @param feature_maps_transformed Vector with copied input maps for transformation process.
+     * @param maps_ranges Vector that contains all sorted RTs of extracted identifications for each map; needed to determine the 10/90 percentiles.
+     * @param map_transformed FeatureMap to store all features of combined maps with original and transformed RTs in order of alignment.
+     * @param trafo_order Vector to store indices of maps in order of alignment.
+    */
+    void treeGuidedAlignment_(const std::vector<BinaryTreeNode> &tree, std::vector<FeatureMap> feature_maps_transformed,
```
timosachsenberg: can this be static?

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @brief Extract original RT ("original_RT" MetaInfo) and transformed RT for each feature to compute RT transformations.
+     *
+     * @param feature_maps Vector of input maps for size information.
+     * @param map_transformed FeatureMap that contains all features of combined maps with original and transformed RTs in order of alignment.
+     * @param transformations Vector to store transformation descriptions for each map. (output)
+     * @param trafo_order Vector that contains the indices of aligned maps in order of alignment.
+    */
+    void computeTrafosByOriginalRT_(std::vector<FeatureMap> &feature_maps, FeatureMap &map_transformed,
```
timosachsenberg: can this be static?

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,175 @@
...
+    String model_type_;
+
+    /// Default params of transformation models linear, b_spline, lowess and interpolated
+    Param model_param_;
+
+    /// Instantiation of alignment algorithm
+    MapAlignmentAlgorithmIdentification align_algorithm_;
+
+    /**
+     * @brief Similarity functor that provides similarity calculations with the ()-operator for protected type SeqAndRTList that stores retention times given for individual peptide sequences of a feature map
```
timosachsenberg: please break sentence into several ones - hard to read this way.

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,175 @@
...
+      Using pearson correlation, calculate the retention time similarity of two maps from their intersection of the peptide identifications.
+      Small intersections are penalized by multiplication with the quotient of intersection to union.
+    */
+    class PeptideIdentificationsPearsonDistance_;
+
+    /**
+     * @brief For given peptide identifications extract sequences and store with associated feature RT.
+     *
+     * @param peptides Vector of peptide identifications to extract sequences.
+     * @param peptide_rts Map to store a list of feature RTs for each peptide sequence as key.
```
timosachsenberg: don't understand this comment

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,175 @@
...
+     * @param map_range Vector in which all feature RTs are stored for given peptide identifications.
+     */
```
timosachsenberg: feature_rt is not documented

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  MapAlignmentAlgorithmTreeGuided::MapAlignmentAlgorithmTreeGuided() :
+          DefaultParamHandler("MapAlignmentAlgorithmTreeGuided"),
+          ProgressLogger()
+  {
+    defaults_.insert("model:", TOPPMapAlignerBase::getModelDefaults("b_spline"));
+    defaults_.setValue("model_type", "b_spline", "Options to control the modeling of retention time transformations from data");
+    defaults_.setValidStrings("model_type", {"linear","b_spline","lowess","interpolated"});
+    defaults_.insert("align_algorithm:", MapAlignmentAlgorithmIdentification().getDefaults());
+    defaults_.setValue("align_algorithm:use_feature_rt", "true", "Because only aligning feature maps with this algorithm, use the retention time of the centroid of the feature (apex of the elution profile) that the peptide was matched to. If different identifications are matched to one feature, only the peptide closest to the centroid in RT is used.#br#Precludes &apos;use_unassigned_peptides&apos;.");
```
timosachsenberg: Please rework description of this parameter

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+    defaultsToParam_();
+  }
+
+  MapAlignmentAlgorithmTreeGuided::~MapAlignmentAlgorithmTreeGuided()
+  {
+  }
+
+  void MapAlignmentAlgorithmTreeGuided::updateMembers_()
+  {
+    align_algorithm_.setParameters(param_.copy("align_algorithm:", true));
```
timosachsenberg: hmm I think if the user changes the model_type, then the defaults of model: also need to be updated?

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+    model_type_ = param_.getValue("model_type");
+    model_param_ = model_param_.copy(model_type_+":", true);
+  }
+
+  // Similarity functor that provides similarity calculations with the ()-operator for protected type SeqAndRTList
+  // that stores retention times given for individual peptide sequences of a feature map
+  class MapAlignmentAlgorithmTreeGuided::PeptideIdentificationsPearsonDistance_
+  {
+  public:
+    float operator()(SeqAndRTList& map_first, SeqAndRTList& map_second) const
```
timosachsenberg: why float?

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+        }
+        ++union_size;
+      }
+      Size intercept_size = intercept_rts1.size();
+
+      // pearsonCorrelationCoefficient(rt_map_i, rt_map_j)
+      float pearson_val;
+      pearson_val = static_cast<float>(Math::pearsonCorrelationCoefficient(intercept_rts1.begin(), intercept_rts1.end(),
+                                                                     intercept_rts2.begin(), intercept_rts2.end()));
+      if (pearson_val > 1)
```
timosachsenberg: how can this happen?

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+      {
+        throw Exception::InvalidRange(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
+      }
+      // Small intersections are penalized by multiplication with the quotient of intersection to union.
+      return pearson_val * intercept_size / union_size;
```
timosachsenberg: what if the union is 0 ? Is this case handled?

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  // For given peptide identifications extract sequences and store with associated feature RT.
+  void MapAlignmentAlgorithmTreeGuided::addPeptideSequences_(const std::vector<PeptideIdentification> &peptides,
+          SeqAndRTList &peptide_rts, std::vector<double> &map_range, double feature_rt)
+  {
+    for (const auto & peptide : peptides)
+    {
+      if (!peptide.getHits().empty())
+      {
+        const String& sequence = peptide.getHits()[0].getSequence().toString();
```
timosachsenberg: @cbielow throwing different charge states together might not be optimal. Maybe append charge state to string?

timosachsenberg: I think some tools (Percolator, MapAlignerIdentification) do that

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+    }
+  }
+
+  // For each input map, extract peptide identifications (sequences) of existing features with associated feature RT.
+  void MapAlignmentAlgorithmTreeGuided::extractSeqAndRt_(const vector<FeatureMap>& feature_maps,
+          vector<SeqAndRTList>& maps_seq_and_rt, vector<vector<double>>& maps_ranges)
+  {
+    for (Size i = 0; i < feature_maps.size(); ++i)
+    {
+      for (auto feature_it = feature_maps[i].begin(); feature_maps[i].end() != feature_it; ++feature_it)
```
timosachsenberg: should be possible to write loop without iterators here for(const FeatureMap& fm : feature_maps)

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  // Extract RTs given for individual features of each map, calculate distances for each pair of maps and cluster hierarchical using average linkage.
+  void MapAlignmentAlgorithmTreeGuided::buildTree_(std::vector<FeatureMap> &feature_maps, std::vector<BinaryTreeNode> &tree,
+          std::vector<std::vector<double>> &maps_ranges)
+  {
+    vector<SeqAndRTList> maps_seq_and_rt(feature_maps.size());
+    extractSeqAndRt_(feature_maps, maps_seq_and_rt, maps_ranges);
+    PeptideIdentificationsPearsonDistance_ pep_dist;
+    AverageLinkage al;
+    DistanceMatrix<float> dist_matrix; // will be filled
```
timosachsenberg: ok I guess that's why float is used

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
cbielow: both has pro's and con's.
The RT should be identical for charge states (give or take some inaccuracies during FF).
So, if MapA only has q=2 but MapB has q=3, we actually get an anchor (which is good).
However, if both maps have both charge states, it gets a bit messy (I'd need to check how these duplicates would actually be handled...)

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
timosachsenberg: I guess this might be because of tests?

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
timosachsenberg: maybe we need to extract the defaults for all models here

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
jthueringer: No. During the refactoring, I just missed adapting it. I'll fix it.

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
jthueringer: Only if the `defaults_` parameters are passed with the method instead.

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
jthueringer: Can you explain what you mean? I think by overwriting 'model_param_' with 'model_param_.copy(model_type_+":", true);' the user input is used (because the trafoXML output contains the changed parameters). Am I missing something?

jthueringer: Sorry, my fault! You are right. I'll change it.

timosachsenberg:
Thanks a lot for the contribution. This is a long missing tool. Great job. To give other users a recommendation on which alignment tool to use in openms it would be great if you could comment on how robust the alignment is. e.g. the empty union case seems to be handled now but I guess there are other (edge?) cases. Also, I would be interested in how many different datasets has this be evaluated?

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+
+// instead of guard
+#pragma once
+
+// included dependencies
+#include <OpenMS/CONCEPT/ProgressLogger.h>
+#include <OpenMS/DATASTRUCTURES/DefaultParamHandler.h>
+#include <OpenMS/DATASTRUCTURES/BinaryTreeNode.h>
+#include <OpenMS/APPLICATIONS/MapAlignerBase.h>
+#include "MapAlignmentAlgorithmIdentification.h"
```
cbielow: this include looks weird. Please use full path

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer:  Julia Thueringer$
+// $Authors: Julia Thueringer $
+// --------------------------------------------------------------------------
+
+// instead of guard
```
cbielow: this is a guard... :)

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+#include <OpenMS/DATASTRUCTURES/BinaryTreeNode.h>
+#include <OpenMS/APPLICATIONS/MapAlignerBase.h>
+#include "MapAlignmentAlgorithmIdentification.h"
+
+namespace OpenMS
+{
+  /**
+    @brief A map alignment algorithm based on peptide identifications from MS2 spectra.
+
+    ID groups with the same sequence in different maps represent points of correspondence between the maps. They are used to evaluate the distances between the maps for hierarchical clustering and form the basis for the alignment.
```
cbielow: `correspondence in RT`

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+    Only the best PSM per spectrum is considered as the correct identification.
+
+    For each pair of maps, the distance is determined based on the intersection of the contained identifications and hierarchically clustered using average linkage.
+    Following the tree, the maps are aligned, resulting in a transformed feature map that contains both the original and the transformed retention times.
+    As long as there are at least two clusters, the alignment is done as follows:\n
```
cbielow: remove `\n`

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+    Of every pair of clusters, the one with the larger 10/90 percentile retention time range is selected as reference for the align() method of @ref OpenMS::MapAlignmentAlgorithmIdentification.
+    align() maps the median retention time of each ID group in the second cluster to the reference retention time of this group.
```
cbielow: replace `maps` with another verb. Could be mistaken as a noun and then the reader becomes confused

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+    Cubic spline smoothing is used to convert this mapping to a smooth function.
+    Retention times in the second cluster are transformed to the reference scale by applying this function.
+    Additionally, the original retention times are stored in the meta information of each ID group.
```
cbielow: `ID group` means PeptideIdentification in Openms terms? If yes, please use it

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+    The reference is combined with the transformed cluster.
+
+    The resulting map is used to extract transformation descriptions for each input map.
+    For each map cubic spline smoothing is used to convert the mapping to a smooth function.
+    Retention times of each map are transformed applying corresponding function.
```
cbielow: `by applying the smoothed function`

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -0,0 +1,177 @@
...
+
+namespace OpenMS
+{
+  /**
+    @brief A map alignment algorithm based on peptide identifications from MS2 spectra.
+
+    ID groups with the same sequence in different maps represent points of correspondence between the maps. They are used to evaluate the distances between the maps for hierarchical clustering and form the basis for the alignment.
+    Only the best PSM per spectrum is considered as the correct identification.
+
+    For each pair of maps, the distance is determined based on the intersection of the contained identifications and hierarchically clustered using average linkage.
```
cbielow: can you add the actual formula for computing the distance here?

src/openms/source/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.cpp
```diff
@@ -0,0 +1,294 @@
...
+// --------------------------------------------------------------------------
+
+#include <OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h>
+// calculate pearson distance
+#include <OpenMS/MATH/STATISTICS/StatisticFunctions.h>
+// create binary tree
+#include <OpenMS/DATASTRUCTURES/DistanceMatrix.h>
+#include <OpenMS/DATASTRUCTURES/BinaryTreeNode.h>
+#include <OpenMS/COMPARISON/CLUSTERING/ClusterHierarchical.h>
+#include <include/OpenMS/COMPARISON/CLUSTERING/AverageLinkage.h>
```
cbielow: remove the `include/`

jthueringer:
@timosachsenberg That no identification contains a sequence should be the only edge case. The analysis of two data sets has shown that to get the best possible result, the parameter `max_rt_shift` has to be a bit higher than when using the MapAlignerIdentification. In general, TreeGuided performed better than Identification and PoseClustering.

The attached pictures show some of the results: For [iPRG2015](https://www.omicsdi.org/dataset/massive/MSV000079843) (12 runs) and [PXD003133](https://www.ebi.ac.uk/pride/archive/projects/PXD003133) (4 runs) the results of the three map aligners were compared. The tests were done with a max_rt_shift of 1000 s, 3500 s, 6000 s and 9000 s.

1. wall time measurements
wall_all

2. consensus feature analysis (grouped with FeatureLinkerUnlabeledQT, RT tolerance was set to 300s)
combined_sizes

3. results of the R- tool [PTXQC](https://pubs.acs.org/doi/abs/10.1021/acs.jproteome.5b00780) for PXD003133 with 9000s max_rt_shift. 
Left: results of MBRAlign 
right: results of MBRIDTransfer - ID-transfer performance of Match-between-runs (the transferred column is very interesting, because it shows that with TreeGuided the amount of false-positive ID-transfers to a file is less (red bar) then with the results of Identification and PoseClustering.
PXD_MBR_9000

4. The recall was calculated based on the formula of [Eva Lange](https://bmcbioinformatics.biomedcentral.com/articles/10.1186/1471-2105-9-375). For the calculation the formula was slightly modified.
### Recall values for iPRG2015
recall_iPRG

### Recall values for PXD003133
recall_PXD

hroest:
@jthueringer why are you using max RT differences of thousands of seconds? That seems a very large amount to me, basically you are saying that a peptide can shift 9000 seconds, is that correct? Usually, after alignment the shifts should be less than 60 seconds, or am I missing something here?

jthueringer:
@hroest The decision to use the large values in max_rt_shift was based on the combination of the composition of the raw data as well as the default settings of the MapAligner:

As can be seen in the following figure, the RT scales of the data sets ranged from 120 to 300 minutes.
beideDatensaetzeRTSpannen

And the analysis of the RT shifts for the identifications in the raw data revealed the following:
rawData-rtShift

Explanation for chosen max_rt_shifts:
  * 1000s: Is the default value of PoseClustering (with about 16 minutes clearly above the average RT shift of one minute of the iPRG2015 dataset, but less then the maximum shift).
  * 3500s: Corresponds to the default value of Identification for iPRG2015 because it uses 50% of the range of the reference RT scale, which is about 60 minutes.
  * 6000s: The average RT difference of identified sequences of the PXD003133 dataset has been exceeded, but the maximum has not yet been reached (should be comparable to 1000s for iPRG2015).
  * 9000s: As with 3500s for iPRG2015, the determined maximum RT shift of PXD003133 is exceeded. In addition, this value corresponds to the default setting of Identification for PXD003133.

And because PXD was the more challenging dataset: The analysis of its RT shifts after the alignment revealed that the shifts were not less than 60 seconds, but significantly less than before:
PXD_rt-shifts-after-alignment

jpfeuffer:
The question probably would be, if those are then true shifts.
After annotation with IDs, maybe the following paper proposes a nice way to determine error rates:
https://www.ncbi.nlm.nih.gov/pubmed/31547658

Nonetheless, a very useful tool! Thank you.

hroest:
Are these RT-shifts after alignment? I would also look at the distributions, but there is simply no conceivable way where such large RT shifts are real. We usually expect shifts of +/- 5 minutes maximum across different runs on different instruments and I dont think it makes any sense to go beyond that, I think 600 seconds is the maximum I would feel comfortable to use and even 1000 s seems way to high for modern instrumentation. This may be a special case but I would not feel comfortable to use shifts > 5min

PR #4507, opened 2020-01-31
an initial version of a Code of Conduct

cbielow:
like every good open-source project should have :)

Adapted from guidelines written by the [The Carpentries Project](https://github.com/carpentries/handbook/blob/master/topic_folders/policies/code-of-conduct.md).

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,73 @@
+### OpenMS Code of Conduct
+
+#### Code of Conduct (Summary View)
+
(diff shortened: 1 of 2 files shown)


cbielow:
any comments?

hendrikweisser:
The document itself is a bit repetitive, but the guidelines make sense and don't seem too overbearing.

Just one question:
> We do not tolerate discrimination or harassment based on [...] technical choices

Does that apply to vi(m) users as well? ;-)

timosachsenberg:
I think it's a good start - let's merge

PR #4508, opened 2020-02-03
FIA-MS data processing class

sgalkina:
@dmccloskey 
Flow-injection analysis class allows for the basic FIA-MS data processing steps such as: acquiring the data for the certain time interval, summing along the time axis, smoothing the peaks, peak picking and accurate mass search. The batch runs are to be managed with the scheduler class that takes a simple csv file as an input.


src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+#include <OpenMS/ANALYSIS/OPENSWATH/SpectrumAddition.h>
+#include <OpenMS/FILTERING/SMOOTHING/SavitzkyGolayFilter.h>
+#include <OpenMS/FORMAT/MzMLFile.h>
+
+using namespace std;
+
+namespace OpenMS
+{
+  /**
+    description
```
hroest: please add a description for this class (as detailed as possible)

hroest:
thanks a lot for contributing, make sure you read https://github.com/OpenMS/OpenMS/blob/develop/CONTRIBUTING.md#opening-a-pull-request -- you are already doing very well but there are  few points that need attention, specifically the documentation and pyOpenMS part (Python bindings).

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+// 
+// --------------------------------------------------------------------------
+// $Maintainer: Chris Bielow $
+// $Authors: Marc Sturm, Hendrik Weisser, Chris Bielow $
```
hroest: I suggest to change this

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+    /// Process the given file
+    void run(float n_seconds, OpenMS::MzTab & output);
+
+    /// Cut the spectra for time
+    void cutForTime(const MSExperiment & experiment, vector<MSSpectrum> & output, float n_seconds=6000);
+
+    /// Merge spectra from different retention times into one
+    MSSpectrum mergeAlongTime(const std::vector<OpenMS::MSSpectrum> & input);
+
+    /// Pick peaks from merged spectrum and return as featureMap with the corresponding polarity
```
hroest: return as featureMap ??? does not seem to be accurate?

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+    const String getDBMapping();
+
+    /// Get the path to the db:struct for passing to AccurateMassSearch
+    const String getDBStruct();
+
+    /// Get the path to the positive adducts for passing to AccurateMassSearch
+    const String getPositiveAdducts();
+
+    /// Get the path to the negative adducts for passing to AccurateMassSearch
+    const String getNegativeAdducts();
```
hroest: have you thought about using the `DefaultParamHandler` instead of using these class members and all the `get` methods?

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+    String positive_adducts_;
+    String negative_adducts_;
+    bool store_progress_;
+    std::vector<float> mzs_; 
+    std::vector<float> bin_sizes_;
+    MSExperiment experiment_;
+  };
+
+/*
+
```
hroest: description?

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+#include <OpenMS/FORMAT/MzTabFile.h>
+#include <OpenMS/FORMAT/MzTab.h>
+#include <OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimatorMedianRapid.h>
+#include <OpenMS/FILTERING/BASELINE/MorphologicalFilter.h>
+#include <OpenMS/KERNEL/SpectrumHelper.h>
+#include <OpenMS/ANALYSIS/OPENSWATH/SpectrumAddition.h>
+#include <OpenMS/FILTERING/SMOOTHING/SavitzkyGolayFilter.h>
+#include <OpenMS/FORMAT/MzMLFile.h>
+
+using namespace std;
```
hroest: please remove from .hpp file

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+      float min_mz=50, 
+      float max_mz=1500, 
+      float bin_step=20
+    );
+
+    /// Default desctructor
+    ~FIAMSDataProcessor();
+
+    /// Copy constructor
+    FIAMSDataProcessor(const FIAMSDataProcessor& cp);
```
hroest:
```suggestion
    FIAMSDataProcessor(const FIAMSDataProcessor& cp) = default;
```

saves some space in the cpp file

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,422 @@
...
+  positive_adducts_ = rhs.positive_adducts_;
+  negative_adducts_ = rhs.negative_adducts_;
+  store_progress_ = rhs.store_progress_;
+  mzs_ = rhs.mzs_;
+  bin_sizes_ = rhs.bin_sizes_;
+  return *this;
+}
+
+void FIAMSDataProcessor::cutForTime(const MSExperiment & experiment, vector<MSSpectrum> & output, float n_seconds) {
+    for (auto s : experiment.getSpectra()) {
```
hroest:
```suggestion
    for (const auto& s : experiment.getSpectra())
    {
```

otherwise you are making a copy of each spectrum first

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,422 @@
...
+    }
+    output.sortByPosition();
+    return output;
+}
+
+MSSpectrum FIAMSDataProcessor::extractPeaks(const MSSpectrum & input) {
+  MSSpectrum spectrum(input);
+
+  SavitzkyGolayFilter filter;
+  filter.filter(spectrum);
```
hroest: SGolay filter assumes equally spaced points, is that the case here? Will this always work well? Also, would you need a way to control parameters?

Finally, this will be pretty slow since constructing an object derived from `DefaultParamHandler` for each spectrum is rather slow...

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,422 @@
...
+}
+
+FeatureMap FIAMSDataProcessor::convertToFeatureMap(const MSSpectrum & input) {
+  FeatureMap output;
+  for (auto it = input.begin(); it != input.end(); ++it) {
+      Feature f;
+      f.setIntensity(it->getIntensity());
+      f.setMZ(it->getMZ());
+      output.push_back(f);
+      output[0].setMetaValue("scan_polarity", polarity_);
```
hroest: why not `f.setMetaValue("scan_polarity", polarity_);` ?

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,422 @@
...
+}
+
+MSSpectrum FIAMSDataProcessor::trackNoise(const MSSpectrum & input) {
+  SignalToNoiseEstimatorMedianRapid sne(50);
+  MSSpectrum output;
+  if (input.size() == 0) {
+    return output;
+  }
+  std::vector<double> mzs, intensities;
+  for (auto it = input.begin(); it != input.end(); ++it)
```
hroest:
```suggestion
  for (const auto& it = input.begin(); it != input.end(); ++it)
```

.gitignore
```diff
@@ -20,10 +20,13 @@ OpenMS.includes
 .DS_Store
 src/.DS_Store
 src/openms/.DS_Store
+src/openms/OpenMS_autogen
```
dmccloskey: This line and the one below should probably be removed.

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,422 @@
...
+  ams.run(input, output);
+}
+
+MSSpectrum FIAMSDataProcessor::trackNoise(const MSSpectrum & input) {
+  SignalToNoiseEstimatorMedianRapid sne(50);
+  MSSpectrum output;
+  if (input.size() == 0) {
+    return output;
+  }
+  std::vector<double> mzs, intensities;
```
dmccloskey: I would recommend using `.reserve(...)` with std::vector for large vector sizes particularly if you know the size ahead of time.  MSSpectrum and other OpenMS types also use std::vector under the hood so there maybe opportunities in to use it there as well.

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,422 @@
...
+  }
+  FeatureMap picked_features = convertToFeatureMap(picked_spectrum);
+  MSSpectrum signal_to_noise = trackNoise(picked_spectrum);
+  storeSpectrum_(signal_to_noise, dir_output_ + "/" + filename_ + "_signal_to_noise_" + postfix + ".mzML");
+  runAccurateMassSearch(picked_features, output);
+  OpenMS::MzTabFile mztab_outfile;
+  mztab_outfile.store(dir_output_ + "/" + filename_ + "_" + postfix + ".mzTab", output);
+}
+
+void FIAMSDataProcessor::loadExperiment_() {
```
dmccloskey: `loadExperiment_` is a bit out of scope for the class.  I think it would be better if the user calls those snippets of code prior to calling the method.

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,209 @@
...
+    bool store_progress_;
+    std::vector<float> mzs_; 
+    std::vector<float> bin_sizes_;
+    MSExperiment experiment_;
+  };
+
+/*
+
+*/
+class OPENMS_DLLAPI FIAMSScheduler
```
dmccloskey: It might be good to move this to its own file.

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,422 @@
...
+    for (auto s : experiment.getSpectra()) {
+        if (s.getRT() < n_seconds) output.push_back(s);
+    }
+}
+
+MSSpectrum FIAMSDataProcessor::mergeAlongTime(
+  const std::vector<MSSpectrum> & input
+  ) {
+    MSSpectrum output;
+    for (size_t i = 1; i < mzs_.size() - 1; i++) {
```
oliveralka: Why do you start at i = 1?

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
timosachsenberg: and feel free to add yourself to the AUTHORS file

sgalkina:
@hroest @oliveralka @dmccloskey thank you for the great help, I've tried to address all the comments. Please let me know if you have any more questions.

src/pyOpenMS/pxds/FIAMSDataProcessor.pxd
```diff
@@ -0,0 +1,29 @@
+from Types cimport *
+from libcpp cimport bool
+from libcpp.vector cimport vector as libcpp_vector
+from SavitzkyGolayFilter cimport *
+from PeakPickerHiRes cimport *
+from MzTab cimport *
+from MzMLFile cimport *
```
hroest:
```suggestion
from MzMLFile cimport *
from FeatureMapping cimport *
```

src/pyOpenMS/pxds/FIAMSScheduler.pxd
```diff
@@ -0,0 +1,20 @@
+from Types cimport *
+from libcpp cimport bool
+from libcpp.map cimport map as libcpp_map
+from libcpp.vector cimport vector as libcpp_vector
+from CsvFile cimport *
```
hroest:
```suggestion
from String cimport *
```

hroest:
I added some changes to the pyOpenMS import

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,265 @@
...
+      sgfilter_(),
+      picker_()
+    {
+    defaults_.setValue("filename", "fiams", "The filename to use for naming the output files");
+    defaults_.setValue("dir_output", "", "The path to the directory where the output files will be placed");
+
+    defaults_.setValue("resolution", 120000.0, "The instrument settings: resolution");
+
+    defaults_.setValue("polarity", "positive", "The instrument settings: polarity");
+    defaults_.setValidStrings("polarity", ListUtils::create<String>("positive,negative"));
```
cbielow: for new code please use C++11 style:
`defaults_.setValidStrings("polarity", {"positive", "negative"});`

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,173 @@
...
+    */
+    MSSpectrum trackNoise(const MSSpectrum & input);
+
+    /**
+      @brief Perform accurate mass search
+
+      @param input  Input a feature map
+      @param output  [out] mzTab file with the accurate mass search results
+    */
+    void runAccurateMassSearch(FeatureMap & input, OpenMS::MzTab & output);
```
cbielow: coding style: please put the ampersand next to the type, e.g. `FeatureMap& input`

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,265 @@
...
+    
+    defaults_.setValue("sgf:frame_length", 11, "SavitzkyGolayFilter parameter. The number of subsequent data points used for smoothing");
+    defaults_.setValue("sgf:polynomial_order", 4, "SavitzkyGolayFilter parameter. Order or the polynomial that is fitted");
+    defaults_.setValue("sne:window", 10, "SignalToNoiseEstimatorMedianRapid parameter. Signal-to-noise estimation window (in mz)");
+    
+    defaultsToParam_();
+  }
+
+  /// Copy constructor
+  FIAMSDataProcessor::FIAMSDataProcessor(const FIAMSDataProcessor& source) :
```
cbielow: can be defaulted... nothing special happens here

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,265 @@
...
+        mzs_.push_back((i+1)*bin_step_);
+        bin_sizes_.push_back(mzs_[i] / (resolution_*4.0));
+    }
+    Param p;
+    p.setValue("frame_length", param_.getValue("sgf:frame_length"));
+    p.setValue("polynomial_order", param_.getValue("sgf:polynomial_order"));
+    sgfilter_.setParameters(p);
+  }
+
+  void FIAMSDataProcessor::cutForTime(const MSExperiment & experiment, const float & n_seconds, std::vector<MSSpectrum> & output) {
```
cbielow: why `const float & n_seconds`?
`const float n_seconds` should do fine?!

src/openms/source/ANALYSIS/ID/FIAMSScheduler.cpp
```diff
@@ -0,0 +1,122 @@
...
+      for (Size j = 0; j < headers.size(); ++j) {
+        mapping[headers[j]] = row[j];
+      }
+      samples_.push_back(mapping);
+    }
+  }
+
+  void FIAMSScheduler::run() {
+    #pragma omp parallel for
+    for (size_t i = 0; i < samples_.size(); ++i) {
```
cbielow: this will not compile on Visual Studio 2017 and earlier. Please use a signed type 
(@timosachsenberg  @jpfeuffer  our CI should catch this... is the jenkins machine actually running on VS2015 as the name implies?)

src/openms/source/ANALYSIS/ID/FIAMSScheduler.cpp
```diff
@@ -0,0 +1,122 @@
...
+      p.setValue("polarity", samples_[i].at("charge"));
+      p.setValue("db:mapping", ListUtils::create<String>(base_dir_ + samples_[i].at("db_mapping")));
+      p.setValue("db:struct", ListUtils::create<String>(base_dir_ + samples_[i].at("db_struct")));
+      p.setValue("positive_adducts", base_dir_ + samples_[i].at("positive_adducts"));
+      p.setValue("negative_adducts", base_dir_ + samples_[i].at("negative_adducts"));
+      fia_processor.setParameters(p);
+
+      String time = samples_[i].at("time");
+      std::vector<String> times;
+      boost::split(times, time, boost::is_any_of(";"));
```
cbielow: why boost and not simply `String().split()`?

cbielow:
there are a few new empty `.gitkeep` files added by accident? Please remove them

cbielow:
are the maintainers for the new classes choosen for a specific reason? If not, please use the authors as maintainers :)

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,265 @@
...
+    MSSpectrum signal_to_noise = trackNoise(picked_spectrum);
+    FeatureMap picked_features = convertToFeatureMap(picked_spectrum);
+    storeSpectrum_(signal_to_noise, dir_output_ + "/" + filename_ + "_signal_to_noise_" + postfix + ".mzML");
+    runAccurateMassSearch(picked_features, output);
+    OpenMS::MzTabFile mztab_outfile;
+    mztab_outfile.store(dir_output_ + "/" + filename_ + "_" + postfix + ".mzTab", output);
+    return is_cached;
+  }
+
+  void FIAMSDataProcessor::storeSpectrum_(const MSSpectrum & input, String filename) {
```
cbielow: should be `FIAMSDataProcessor::storeSpectrum_(const MSSpectrum& input, const String& filename)`

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -0,0 +1,265 @@
...
+      MzMLFile mzml;
+      MSExperiment exp;
+      exp.addSpectrum(input);
+      mzml.store(filename, exp);
+  }
+
+  /// Get mass-to-charge ratios to base the sliding window upon
+  const std::vector<float> FIAMSDataProcessor::getMZs(){
```
cbielow: should be `const std::vector<float>& FIAMSDataProcessor::getMZs()`

same for  `getBinSizes()`

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,173 @@
...
+    FIAMSDataProcessor(const FIAMSDataProcessor& cp);
+
+    /// Assignment
+    FIAMSDataProcessor& operator=(const FIAMSDataProcessor& fdp);
+
+    /**
+      @brief Run the full analysis for the experiment for the given time interval
+
+      The workflow steps are:
+      - the time axis of the experiment is cut to n_seconds
```
cbielow: this means from 0:n_seconds ? Maybe make that a bit clearer?

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,173 @@
...
+// 
+// --------------------------------------------------------------------------
+// $Maintainer: Chris Bielow $
+// $Authors: Svetlana Kutuzova, Douglas McCloskey $
+// --------------------------------------------------------------------------
+ 
+#pragma once
+#include <OpenMS/FILTERING/SMOOTHING/SavitzkyGolayFilter.h>
+#include <OpenMS/TRANSFORMATIONS/RAW2PEAK/PeakPickerHiRes.h>
+#include <OpenMS/FORMAT/MzTab.h>
```
cbielow: you don't need these (very heavy) includes (MzTab and MzMLFile).
Forward declaring them is enough.

src/openms/source/ANALYSIS/ID/FIAMSScheduler.cpp
```diff
@@ -0,0 +1,122 @@
...
+      for (size_t j = 0; j < times.size(); ++j) {
+        OPENMS_LOG_INFO << "Started " << samples_[i].at("filename") << " for " << times[j] << " seconds" << std::endl;
+        MzTab mztab_output;
+        fia_processor.run(exp, std::stof(times[j]), mztab_output, load_cached_);
+        OPENMS_LOG_INFO << "Finished " << samples_[i].at("filename") << " for " << times[j] << " seconds" << std::endl;
+      }
+    }
+  }
+
+  const std::vector<std::map<String, String>> FIAMSScheduler::getSamples() {
```
cbielow: return by const ref

src/openms/source/ANALYSIS/ID/FIAMSScheduler.cpp
```diff
@@ -0,0 +1,122 @@
...
+    return samples_;
+  }
+
+  const String FIAMSScheduler::getBaseDir() {
```
cbielow: return by const ref

src/openms/include/OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h
```diff
@@ -0,0 +1,173 @@
...
+    though it is not the exact implementation.
+  */
+  class OPENMS_DLLAPI FIAMSDataProcessor  :
+    public DefaultParamHandler
+  {
+public:
+    /// Constructor
+    FIAMSDataProcessor();
+
+    /// Default desctructor
```
cbielow: typo?

cbielow:
nice PR and thanks for contributing!

There are still a few performance opportunities missed here (beside the ones I annotated), but if its fast enough, just leave it as is).

In general, on first sight, the new functions in the Processor seem to duplicate functionality we already have, but when looking at the implementation, they just use what is available already (which is nice!).
So to avoid confusion, could you add to the documentation of each method what is actually used inside (e.g. an sg-filter, or SN-filter class XY etc)?!

sgalkina:
Dear all, @hroest @oliveralka @cbielow @dmccloskey 

Thank you for all the very helpful comments. I have some questions about the CI builds I cannot resolve myself.
There are 4 builds currently failing and it looks like 3 of those fails are not related to the new code introduced.

sgalkina:
Dear all, @hroest @oliveralka @cbielow @dmccloskey

Thank you for all the helpful comments. I have some questions about the CI builds I cannot resolve myself.
It looks like 3 of the failed builds are not related to the new code introduced.
1) https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_ci/compiler=msvc-14.0,os_label=win10_64/3846/ all the failed tests are from TOPP section and FIA-MS class is not one of them
2)  https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms_pr_docker_pyopenms/2280/#showFailuresLink - failed pyOpenMS classes do not list FIA-MS
3) https://travis-ci.org/OpenMS/OpenMS/jobs/649531138?utm_medium=notification&utm_source=github_status I have troubles debugging this one since the error message is `Could not find executable /home/travis/build/OpenMS/OpenMS/_build/src/tests/class_tests/bin/FIAMSDataProcessor_test` for DataProcessor class but not for Scheduler class. Also, other platforms didn't have that problem.

If you have an idea of what causing it please help. Do those failed builds have to actually be fixed before merging or the errors are not caused by the new code you think? 

Thank you

jpfeuffer:
Hi @sgalkina 
I would suggest to start by looking at our CDash which provides a quicker overview of what went wrong (better summarized than the reports on jenkins).

You find it here:
http://cdash.openms.de/index.php?project=OpenMS&date=2020-02-12

Look for your PR number 4508.
In detail, it complains about:
`C:\jenkins\ws\openms_pr_ci\031d6063\source\src\openms\source\ANALYSIS\ID\FIAMSScheduler.cpp(74): error C3016: 'i': index variable in OpenMP 'for' statement must have signed integral type`

`src/tests/class_tests/openms/source/FIAMSDataProcessor_test.cpp:170:10: error: unused variable ‘is_cached_before’ [-Werror=unused-variable]`

jpfeuffer:
Additionally one test fails in your class tests with gcc 7 on ubuntu 18.04:
`-  line 204:  TEST_EQUAL(abs((float)picked_result.MZBegin(mz)->getIntensity() - (float)spec_picked.MZBegin(mz)->getIntensity()) < 100,true): got '0', expected '1'`

src/openms/source/ANALYSIS/ID/FIAMSScheduler.cpp
```diff
@@ -72,7 +72,7 @@ namespace OpenMS {
 
   void FIAMSScheduler::run() {
     #pragma omp parallel for
-    for (size_t i = 0; i < samples_.size(); ++i) {
+    for (Size i = 0; i < samples_.size(); ++i) {
```
timosachsenberg: this needs to be int for OpenMP

timosachsenberg:
```
for (int i =0; i < (int)samples_size(); ++i)
```

src/pyOpenMS/pxds/FIAMSScheduler.pxd
```diff
@@ -0,0 +1,20 @@
...
+    cdef cppclass FIAMSScheduler "OpenMS::FIAMSScheduler":
+        #
+        # wrap-doc:
+        #     ADD PYTHON DOCUMENTATION HERE
+        #
+        FIAMSScheduler() nogil except +
+        FIAMSScheduler(FIAMSScheduler) nogil except +
+        FIAMSScheduler(String filename, String base_dir, bool load_cached_) nogil except +
+        void run() nogil except +
+        libcpp_vector[ libcpp_map[ String, String ] ] getSamples() nogil except +
```
hroest: this is not going to work in pyOpenMS, we cannot automatically wrap such complex STL constructs. I suggest to comment it out unless you really need it in Python
```suggestion
        # libcpp_vector[ libcpp_map[ String, String ] ] getSamples() nogil except +
```

src/pyOpenMS/pxds/FIAMSDataProcessor.pxd
```diff
@@ -0,0 +1,30 @@
...
+    cdef cppclass FIAMSDataProcessor(DefaultParamHandler) :
+        #
+        # wrap-doc:
+        #     ADD PYTHON DOCUMENTATION HERE
+        #
+        # wrap-inherits:
+        #  DefaultParamHandler
+        FIAMSDataProcessor() nogil except +
+        FIAMSDataProcessor(FIAMSDataProcessor) nogil except +
+        # NAMESPACE # bool run(MSExperiment & experiment, float & n_seconds, OpenMS::MzTab & output, bool load_cached_spectrum) nogil except +
```
hroest: this should work
```suggestion
        bool run(MSExperiment & experiment, float & n_seconds, MzTab & output, bool load_cached_spectrum) nogil except +
```

hroest:
some issues with pyOpenMS that are fixable if you comment out your return of STL construct

sgalkina:
> some issues with pyOpenMS that are fixable if you comment out your return of STL construct

thanks you @hroest. After applying your suggested changes I can see that the build is still failing on pyOpenMS side. Is there anything else you can spot that can fix it? http://cdash.openms.de/viewTest.php?onlyfailed&buildid=397836

hroest:
yes, it seems compiling to C++ is the problem

```

pyopenms/pyopenms_6.cpp: In function 'PyObject* __pyx_pf_8pyopenms_10pyopenms_6_14FIAMSScheduler_6_init_0(__pyx_obj_8pyopenms_10pyopenms_6_FIAMSScheduler*)':
pyopenms/pyopenms_6.cpp:119692:44: error: no matching function for call to 'OpenMS::FIAMSScheduler::FIAMSScheduler()'
     __pyx_t_1 = new OpenMS::FIAMSScheduler();
                                            ^
In file included from pyopenms/pyopenms_6.cpp:747:0:
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/include/OpenMS/ANALYSIS/ID/FIAMSScheduler.h:72:5: note: candidate: OpenMS::FIAMSScheduler::FIAMSScheduler(const OpenMS::FIAMSScheduler&)
     FIAMSScheduler(const FIAMSScheduler& cp) = default;
     ^
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/include/OpenMS/ANALYSIS/ID/FIAMSScheduler.h:72:5: note:   candidate expects 1 argument, 0 provided
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/include/OpenMS/ANALYSIS/ID/FIAMSScheduler.h:62:5: note: candidate: OpenMS::FIAMSScheduler::FIAMSScheduler(OpenMS::String, OpenMS::String, bool)
     FIAMSScheduler(
     ^
/home/jenkins/ws/openms_pr_docker_pyopenms/OpenMS/src/openms/include/OpenMS/ANALYSIS/ID/FIAMSScheduler.h:62:5: note:   candidate expects 3 arguments, 0 provided
At global scope:
```

it seems like you dont have a default constructor for the scheduler? either you have to add one in C++ or add `#wrap-ignore` to the pxd file

```
FIAMSScheduler() = default;
```

or 

```
FIAMSScheduler() nogil except + #wrap-ignore
```

sgalkina:
@hroest thanks again for your help, the builds are finally fixed :) please let me know if there is anything else blocking the merge

PR #4509, opened 2020-02-03
[GUI,FIX] removal of temporary theoretical spectrum layers

enetz:
This fixes a TOPPView crash caused by accumulating DataLayers from theoretical spectra in IDView.

The parameter `caption` in `addData()` is not used as the layer name. The `filename` parameter is used for that and was set as an empty string. The removal of these temporary layers relies on using the caption as the layer name (to identify them as temporary layers), so they were not removed and caused several problems.
After adding exactly 10 layers TOPPView just crashes for some reason.

Now the `layer_caption` of the theoretical spectra is also used as the filename and the layers are removed correctly. This makes the IDView for normal proteomics data usable again.
Tested on Linux and Windows.


timosachsenberg:
Looks good to me

PR #4511, opened 2020-02-05
[GUI,FIX] precalculated ppm errors were always read from the first PeptideHit

enetz:
The `precursor error (|ppm|)` column in TOPPView IdentificationView is read from a MetaValue if it is available. For some reason it was always read from the first PeptideHit of the spectrum. This fixes this and shows the correct number for each PeptideHit.


timosachsenberg:
good catch!

PR #4512, opened 2020-02-06
AssayGeneratorMetabo: Filter fragments based on min/max m/z

oliveralka:
Allows restriction of the fragment ion mass (min, max), when choosing the transitions instead of reducing it on spectrum level (mzML) e.g. via FileFilter. 

It seems it is better to use all available data for SIRIUS, Passatutto and restrict annotated fragments at a later stage (transition extraction). Otherwise, the fragmentation trees provided by SIRIUS may lose smaller fragmentation reactions e.g. loss of CH2. This may lead to a shift in the decoy fragment mass distribution down the road. 

Note: Decoy generation is not yet included in the AssayGeneratorMetabo, which may be introduced with the next release of SIRIUS 4.3.0 - currently it is handled via a python script and will be available soon for this targeted case. 

In addition:
Tweaked some of the initial parameters. 

Edit: FYI - The change of the tests was necessary due to the change of initial parameters


oliveralka:
@jpfeuffer I will close this PR, since the changes are already on the release branch (2.5.0) - all additional commits to the current release branch will be added to develop later on anyway, or (this, THIRDPARTY, and so on)?

jpfeuffer:
Yes, we will merge back to master and develop

PR #4514, opened 2020-02-06
update THIRDPARTY to include TRFP 1.2.0

timosachsenberg:


timosachsenberg:
Regarding the failing tests, there is currently a bug in ThermoRawFileParser that will be fixed (probably) today.

PR #4515, opened 2020-02-06
[Fix] proteomics lfq fdr

jpfeuffer:
There was a small error when writing the debug file because unreferenced proteins were not correctly filtered.
I added a potential PSM/peptide FDR filtering.


timosachsenberg:
looks good

jpfeuffer:
travis style failed randomly.

PR #4516, opened 2020-02-07
[Fix] idpep outlier handling

jpfeuffer:
Happened for OMSSA in our Tutorial.

### DIFF
--- a/src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
+++ b/src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
@@ -70,6 +70,8 @@ namespace OpenMS
       defaults_.setValue("max_nr_iterations", 1000, "Bounds the number of iterations for the EM algorithm when convergence is slow.", ListUtils::create<String>("advanced"));
       defaults_.setValidStrings("incorrectly_assigned", ListUtils::create<String>("Gumbel,Gauss"));
       defaults_.setValue("neg_log_delta",6, "The negative logarithm of the convergence threshold for the likelihood increase.");
+      defaults_.setValue("outlier_handling","ignore", "What to do with outliers outside of 3*IQR from Q1/Q3.");
+      defaults_.setValidStrings("outlier_handling", {"ignore","set_to_closest_valid","none"});
       defaultsToParam_();
       getNegativeGnuplotFormula_ = &PosteriorErrorProbabilityModel::getGumbelGnuplotFormula;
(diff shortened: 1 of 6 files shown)


src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
```diff
@@ -781,27 +805,74 @@ namespace OpenMS
...
+                  x_scores.end()
+          );
+          nr_outliers = before - x_scores.size();
+        }
+        else //set_to_closest_valid
+        {
+          auto closest_lower = std::lower_bound(x_scores.begin(), x_scores.end(), q1 - 3 * iqr);
+          auto closest_upper = --std::upper_bound(x_scores.begin(), x_scores.end(), q3 + 3 * iqr);
+
+          for (auto it = x_scores.begin(); it != closest_lower; ++it)
```
timosachsenberg: if x_scores is empty the loop will not terminate. Is this handled?

src/topp/IDPosteriorErrorProbability.cpp
```diff
@@ -221,7 +222,10 @@ class TOPPIDPosteriorErrorProbability :
       }
 
       // fit to score vector
-      bool return_value = PEP_model.fit(score.second[0]);
+      //TODO choose outlier handling based on search engine? If not set by user?
+      //XTandem is prone to accumulation at min values/censoring
```
timosachsenberg: could this be detected by looking at the smallest non-empty bin? if that one contains e.g., the largest number of counts it is set to 0?

timosachsenberg: or is this properly handled now?

jpfeuffer: No, this case is still open. Unfortunately we don't work with bins here. We could look for a big number of equal scores in the low range. AFAIR it's always the same minimum score but I would need more data of many settings to see when and why xtandem actually produces such weird distributions sometimes.

jpfeuffer: Ignoring would be a first step but IMHO censoring should be handled during fitting by e.g. taking the likelihood of being smaller than the censoring threshold (i.e. CDF) instead of a point estimate at that threshold. But I would like to see how often and how pronounced it happens and especially why (maybe it can be avoided in the search settings).

timosachsenberg: A quick fix with probably little impact on fits in general would be to ignore all values equal or above 99.9 % percentile or equal and below 0.01.

jpfeuffer: I gave it a little try but I have no data to test it.

PR #4523, opened 2020-02-17
update release 2.5 branch

cbielow:
it looks like all of current develop could be merged into release.
Any objections? (e.g. is FFMetabo experimental?)


oliveralka:
The FFM elemental scoring works as expected, but it can be refined further (e.g. rt scoring - check if small trace within larger trace (99% overlap); Check if the Apex is the same in between traces which should be grouped). 

Summary: It can be used and since it is not the standard yet - it should be fine to have it in the release. 
The other changes can be integrated later on.

PR #4526, opened 2020-02-18
Check only 1x if output file can be written

bernt-matthias:
Happens already here https://github.com/OpenMS/OpenMS/blob/ca57038b730ba5995662476e64b4019130cc024e/src/openms/source/APPLICATIONS/TOPPBase.cpp#L1296


bernt-matthias:
Actually I wanted to target develop...

timosachsenberg:
k will rebase

Issue #4404, opened 2019-12-08
SeedListGenerator output by prefix for consensuXML input


bernt-matthias:
For consensusXML input the number of outputs depends on the contents of the input. 

In order to automatize this step it would be great to just specify a prefix of the outputs. 

For instance, one could have an optional `-out FILE` argument that needs to be specified if the input is not consensusXML and `-out_prefix PREFIX` if the input is otherwise.  

Background: For the CTD -> galaxy tool the automatic conversion of the consensusXML input case seems impossible. I'm wondering if there is a workaround.

jpfeuffer:
True. This would be much better but this also needs to be reflected in the CTD for other wrappers. We do not yet support its "output prefix capabilities" but it is probably not too much work (it is basically just another tag).

bernt-matthias:
There are seemingly already some other places where directories are reflected as simple string. A quick grep in the ctd files gave me this list:

MzMLSplitter -out
PepNovoAdapter -dir
IDRipper -out 
MascotAdapter -mascot_directory temp_data_directory
InspectAdapter -inspect_directory temp_data_directory
PrecursorIonSelector -tmp_dir
IDFileConverter -in 
OpenSwathFileSplitter -outputDirectory

Maybe adding a simple string option would be a good first step.

bernt-matthias:
For the case of IDRipper there is `-out_path` and `-out`:

```
  -out <file>        The path to this file is used as the output directory. (valid formats: 'idXML')
  -out_path <file>   Directory for the output files after ripping according to 'file_origin'. If 'out_pa
                     th' is set, 'out' is ignored.
```

When specifying `-out_path DIR` the files are located in the current working dir. Only with `-out_path DIR/FILE` (where FILE does not even need to be existent) the generated files are located in `DIR`.

bernt-matthias:
Also `DTAExtractor` seems to be an example

cbielow:
> Background: For the CTD -> galaxy tool the automatic conversion of the consensusXML input case seems impossible. I'm wondering if there is a workaround.

So how would this PREFIX help Galaxy? Would you assume all files with this prefix in the target directory where generated by the tool?

bernt-matthias:
Exactly. The prefix would be an existing directory in Galaxy. Galaxy has means to take all files from an directory (optionally matching a regexp). 

[...]

PR #4527, opened 2020-02-18
initial implementation of output prefix parameters

bernt-matthias:
as discussed in https://github.com/OpenMS/OpenMS/issues/4404

implementation for SeedListGenerator currently only works for consensusXML input. For this tool one could think about to have the -out and the -out_prefix that could be used exclusively. for the -out_prefix parameter a question would be how the names should be generated for the other input types.

TODOs:
- [ ] CTD changes


jpfeuffer:
This looks like a great start. Indeed the CTD changes are missing, but at least the prefix type already exists there. I will try to add to this PR when I find time again.

timosachsenberg:
any update here?

src/openms/source/APPLICATIONS/TOPPBase.cpp
```diff
@@ -1360,6 +1370,10 @@ namespace OpenMS
     {
       outputFileWritable_(param_value, param_name);
     }
+    else if (p.type == ParameterInformation::OUTPUT_PREFIX)
+    {
+      outputFileWritable_(param_value + "_0", param_name);
```
timosachsenberg, line 1375: only tests one file

src/openms/include/OpenMS/APPLICATIONS/TOPPBase.h
```diff
@@ -524,12 +524,33 @@ namespace OpenMS
...
+      Tools should follow the convention to name output files PREFIX_[0..N-1].EXTENSION.
+      For example, a tool that splits mzML files into multiple mgf files should create files:
+      splitted_0.mgf, splitted_1.mgf, ... if splitted got passed as prefix.
+      
+      Note: setting format(s) via setValidFormat_ for an output prefix will can used to export
+            e.g. valid CTD files that contain information on the expected output file types.            
+
+      @param name Name of the option in the command line and the INI file
+      @param argument Argument description text for the help output
+      @param default_value Default argument
```
jpfeuffer, line 539:
```suggestion
      @param default_value Default value (remember, no extension is specified here)
```

src/openms/include/OpenMS/APPLICATIONS/TOPPBase.h
```diff
@@ -524,12 +524,33 @@ namespace OpenMS
...
     void registerOutputFile_(const String& name, const String& argument, const String& default_value, const String& description, bool required = true, bool advanced = false);
 
+    /**
+      @brief Registers an output file prefix used for tools with multiple file output.
+
+      Tools should follow the convention to name output files PREFIX_[0..N-1].EXTENSION.
+      For example, a tool that splits mzML files into multiple mgf files should create files:
+      splitted_0.mgf, splitted_1.mgf, ... if splitted got passed as prefix.
+      
+      Note: setting format(s) via setValidFormat_ for an output prefix will can used to export
```
jpfeuffer, line 534:
```suggestion
      Note: setting format(s) via setValidFormat_ for an output prefix can be used to export
```

src/openms/source/APPLICATIONS/TOPPBase.cpp
```diff
@@ -1360,6 +1371,10 @@ namespace OpenMS
...
```
timosachsenberg, line 1376:
```suggestion
      outputFileWritable_(param_value + "_0", param_name); // only test one file
```

src/openms/include/OpenMS/APPLICATIONS/TOPPBase.h
```diff
@@ -524,12 +524,33 @@ namespace OpenMS
...
+            e.g. valid CTD files that contain information on the expected output file types.
```
jpfeuffer, line 535:
```suggestion
            e.g. valid CTD files that contain information on the expected output file types. In theory, it is possible
            to output different types and list them here but this should be avoided for cleanlyness (prefer multiple
            separate outputs). This could be left empty in case of an unknown amount of different extensions that
            are produced but is highly recommended.
```

timosachsenberg (approved):

PR #4528, opened 2020-02-18
[Feature] FFID batch extraction of chromatograms

jpfeuffer:


src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.h
```diff
@@ -259,6 +264,48 @@ class OPENMS_DLLAPI FeatureFinderIdentificationAlgorithm :
 
   void calculateFDR_(FeatureMap& features);
 
+  template <typename It>
+  std::vector<std::pair<It,It>>
+  chunk_(It range_from, It range_to, const std::ptrdiff_t portion)
```
timosachsenberg: maybe also use batch_size here

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.h
```diff
@@ -259,6 +264,48 @@ class OPENMS_DLLAPI FeatureFinderIdentificationAlgorithm :
 
   void calculateFDR_(FeatureMap& features);
 
+  template <typename It>
```
timosachsenberg: add docu

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.h
```diff
@@ -259,6 +264,48 @@ class OPENMS_DLLAPI FeatureFinderIdentificationAlgorithm :
...
+  {
+    /* Aliases, to make the rest of the code more readable. */
+    using std::vector;
+    using std::pair;
+    using std::make_pair;
+    using std::distance;
+    using diff_t = std::ptrdiff_t;
+
+    /* Total item number and portion size. */
+    const diff_t total
```
timosachsenberg: in one line?

timosachsenberg: in general some minor style issues (spaces after comma etc.)

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -72,6 +72,9 @@ namespace OpenMS
     defaults_.setValue("debug", 0, "Debug level for feature detection.", ListUtils::create<String>("advanced"));
     defaults_.setMinInt("debug", 0);
 
+    defaults_.setValue("extract:batch_size", 1000, "Nr of peptides for which chromatograms are extracted at the same time."
```
timosachsenberg: "Nr of peptides used in each batch of chromatogram extraction. Smaller values decrease memory usage but increase runtime."

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -347,53 +350,66 @@ namespace OpenMS
...
-      OpenSwath::SpectrumAccessPtr spec_temp =
-        SimpleOpenMSSpectraFactory::getSpectrumAccessOpenMSPtr(shared);
-      extractor.extractChromatograms(spec_temp, chrom_temp, coords, mz_window_,
-          mz_window_ppm_, "tophat");
-      extractor.return_chromatogram(chrom_temp, coords, library_, (*shared)[0],
-          chrom_data_.getChromatograms(), false);
-    }
+      createAssayLibrary_(chunk.first, chunk.second, ref_rt_map);
+
+      /*if (debug_level_ >= 666)
```
timosachsenberg: could we keep this debug code by moving it up out of the loop?

timosachsenberg: is it possible to move that out of the loop and do an 
extra createAssayLibrary( with begin and end? or is ref_rt_map changed?

jpfeuffer: But that defeats the purpose of chunking, no?

jpfeuffer: yes, we add to ref_rt_map in every iteration

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -597,12 +613,12 @@ namespace OpenMS
...
   }
 
-  void FeatureFinderIdentificationAlgorithm::createAssayLibrary_(PeptideMap& peptide_map, PeptideRefRTMap& ref_rt_map)
+  void FeatureFinderIdentificationAlgorithm::createAssayLibrary_(const PeptideMap::iterator& begin, const PeptideMap::iterator& end, PeptideRefRTMap& ref_rt_map)
   {
     std::set<String> protein_accessions;
 
-    for (PeptideMap::iterator pm_it = peptide_map.begin();
-         pm_it != peptide_map.end(); ++pm_it)
+    for (auto pm_it = begin;
```
timosachsenberg: not sure if removing all autos here is not detrimental to understanding the code

timosachsenberg: and below

jpfeuffer: I did it because I wanted to make it const, until I saw that we clear the values in the end.
auto would allow for changing it to const iterators later.
And it gets rid of my clang-tidy warnings..

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
timosachsenberg: yes but this would only be intended for writing the information of the traml file

timosachsenberg: so you could keep that debug traml thingy outside the loop

PR #4529, opened 2020-02-18
PercolatorAdapter and CruxAdapter: use underscore in arguments

bernt-matthias:
instead of dashes. makes it consistent and allows easier integration in Galaxy.

hope that I got all of them.


timosachsenberg:
this is a breaking change (for workflows) so we would need to discuss how to handle this

bernt-matthias:
An alternative (or extension) would be to modify the functions that get the values such that they don't care about dash / underscore.

jpfeuffer:
To be honest, I like underscores better, but indeed this will break every workflow, no matter if bash, nextflow, KNIME or TOPPAS.

cbielow:
I like the improved consistency but why does Galaxy stumble across dashes?

bernt-matthias:
> To be honest, I like underscores better, but indeed this will break every workflow, no matter if bash, nextflow, KNIME or TOPPAS.

Any idea how we could do this backward compatible.

Maybe hack the `get...Option` to check for the parameter with underscores and dashes, i.e. 

check here 

https://github.com/OpenMS/OpenMS/blob/b705b33848ac875c99d53783510bc72c69b01db8/src/openms/source/APPLICATIONS/TOPPBase.cpp#L1176

and a few other places where parameters are looked up. 

Alternatively the replacement could be done when parsing the command line (maybe this is easier?).

> I like the improved consistency but why does Galaxy stumble across dashes?

Problem is that the automatic conversion just gets more complex, since I need to check at many places in the code if its underscores or dashes. In addition Galaxy tools rely in many places on python and the command line parameters correspond to python variables which can't contain dashes.

timosachsenberg:
I think this might be more complex on the OpenMS side as some parameters are passed through to the underlying options that directly work on parameter objects. This would need to be hacked into the command line parsing in OpenMS

timosachsenberg:
Ok let's merge this even if it means we need to update some workflows. We have some breaking changes anyway

PR #4530, opened 2020-02-18
Topic/3rdparty test fix

bernt-matthias:


bernt-matthias:
duplicate of https://github.com/OpenMS/OpenMS/pull/4531

PR #4533, opened 2020-02-19
RTEvaluation: fix output formats

bernt-matthias:
- csv output was not csv so I just made it tsv (which is better anyway)
- latex was missing line ends


bernt-matthias:
As test `RTEvaluation -in PeptideIndexer_1.idXML -out RTEvaluation.tsv` would work.

PR #4534, opened 2020-02-19
fix doxygen errors

cbielow:
this fixes quite a lot of broken links in our documenation, especially with TOPP/UTILS

I also tried to switch to MARKDOWN support, but accidental use of indentation will break a lot of our pages since the markdown interpreter runs first and converts these parts to verbatim blocks...
If at some point this is fixed, we could use

| potential predecessor tools   |    MapAlignerTreeGuided    | potential successor tools |
| ----------------------------- | -------------------------- | --------------------------  |
| TOPP_IDMapper                 | ^                          | TOPP_FeatureLinkerUnlabeledKD |

as table format

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/QUANTITATION/ProteinResolver.h
+++ b/src/openms/include/OpenMS/ANALYSIS/QUANTITATION/ProteinResolver.h
@@ -164,50 +164,50 @@ public:
     */
     void resolveID(std::vector<PeptideIdentification> & peptide_identifications);
 
-    /**
-      @brief NOT IMPLEMENTED YET
+    // /**
+      // @brief NOT IMPLEMENTED YET
 
-      @param protein_nodes
-      @param peptide_nodes
-      @param reindexed_proteins
-      @param reindexed_peptides
-      @param peptide_identifications
-      @param output
-    */
+      // @param protein_nodes
+      // @param peptide_nodes
+      // @param reindexed_proteins
+      // @param reindexed_peptides
+      // @param peptide_identifications
+      // @param output
+    // */
     // void writeProteinsAndPeptidesmzTab(std::vector<ProteinEntry>& protein_nodes, std::vector<PeptideEntry>& peptide_nodes, std::vector<Size>& reindexed_proteins, std::vector<Size>& reindexed_peptides, std::vector<PeptideIdentification>& peptide_identifications, String& output  );
-    /**
-      @brief Writing peptide table into text file
+    // /**
+      // @brief Writing peptide table into text file
 
-      @param peptides
-      @param reindexed_peptides
-      @param identifications
-      @param output_file
-    */
+      // @param peptides
+      // @param reindexed_peptides
+      // @param identifications
+      // @param output_file
+    // */
     // void writePeptideTable(std::vector<PeptideEntry> & peptides, std::vector<Size> & reindexed_peptides, std::vector<PeptideIdentification> & identifications, String & output_file); // not implemented
-    /**
-      @brief Writing peptide table into text file
+    // /**
+      // @brief Writing peptide table into text file
 
-      @param peptides
-      @param reindexed_peptides
-      @param consensus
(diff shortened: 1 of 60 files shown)


cbielow:
rebuild jenkins

jpfeuffer:
Seems unrelated.

jpfeuffer:
feel free to merge when travis passes.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/ProteinResolver.h
```diff
@@ -164,50 +164,50 @@ namespace OpenMS
...
-      @brief NOT IMPLEMENTED YET
-
-      @param protein_nodes
-      @param peptide_nodes
-      @param reindexed_proteins
-      @param reindexed_peptides
-      @param peptide_identifications
-      @param output
-    */
+    // /**
```
timosachsenberg: is this correct?

src/openms/source/APPLICATIONS/ToolHandler.cpp
```diff
@@ -107,8 +107,6 @@ namespace OpenMS
     tools_map["NoiseFilterGaussian"] = Internal::ToolDescription("NoiseFilterGaussian", "Signal processing and preprocessing");
     tools_map["NoiseFilterSGolay"] = Internal::ToolDescription("NoiseFilterSGolay", "Signal processing and preprocessing");
     tools_map["OMSSAAdapter"] = Internal::ToolDescription("OMSSAAdapter", "Identification");
-    tools_map["OpenPepXL"] = Internal::ToolDescription("OpenPepXL", "Identification");
```
timosachsenberg: hmm I thought we had in our changelog that these are now tools..

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/ProteinResolver.h
cbielow: yes. If you leave the doxygen comments like they were, all these parameters get attributed to the first active function below...

src/openms/source/APPLICATIONS/ToolHandler.cpp
cbielow: Ok, I did not know that. I used the minimal number of changes to make it consistent:
They had the 'bool official = false' set in TOPPBase and were still named '@page UTILS_..." in Doxygen, but were in the wrong category in ToolHandler. All of these need to be consistent for our docs to work properly.

timosachsenberg: @enetz can you fix this?

enetz: @timosachsenberg @cbielow It is very possible that I forgot some details when turning my tools from UTILs to TOOLs. Probably also applies to OpenPepXLLF and XFDR, not just OpenPepXL. I will look into it tomorrow.

timosachsenberg:
@chris can you revert the downgrade to util and change the missing stuff to make it a tool?

cbielow:
> @chris can you revert the downgrade to util and change the missing stuff to make it a tool?

what about UTILS_RNPxlSearch and UTILS_RNPxlXICFilter. They remain in UTIL?

timosachsenberg:
yes that's fine as there will be a successor tool at some point

cbielow:
Done.

Im attaching the remaining errors, if someone else wants to have a go... (this should really be a CI test...)
[doxygen-error.log](https://github.com/OpenMS/OpenMS/files/4230845/doxygen-error.log)

jpfeuffer:
We have the doxygen logs nightly. I can add a parser for it maybe. I am a bit hesitant to add it to general CI as it takes quite some additional times.

jpfeuffer:
You even see them in CDash as Warnings: http://cdash.openms.de/viewBuildError.php?type=1&buildid=401832

PR #4535, opened 2020-02-19
[FIX] quantification based on seeds was broken

timosachsenberg:


timosachsenberg:
failed windows build seems unrelelated

PR #4329, opened 2019-10-25
Feature/pyopenms acquisitioninfo access

mwalzer:
Enables AcquisitionInfo access to items and from items to metainfointerface within PyOpenMS.
Turns out this is not integrated by default, so here the pxd fix.

PR #4536, opened 2020-02-19
[FEATURE] correct return value for operator[]

hroest:
correctly use reference return value


timosachsenberg:
cool this would also fix https://github.com/OpenMS/OpenMS/pull/4329 if we add it to AQInfo

hroest:
no, not yet - but this is the first step towards a solution, but we need to also adopt autowrap to generate both `__setitem__` and `__getitem__` functions, which it currently does not do

hroest:
this would probably only work together with https://github.com/uweschmitt/autowrap/pull/91

jpfeuffer:
rebuild jenkins

jpfeuffer:
seems like the autowrap PR was merged. What about this one?

timosachsenberg:
ping

PR #4436, opened 2019-12-21
Float Data Array fixes

hroest:
a couple of features:

- Keep the name of float data arrays (instead of deleting them)
- remember precision of float data arrays when storing 
- represent float data arrays be default with double precision
- fix check for numerical value for data arrays (relates to content of array)
- store unit metadata for float data array descriptor

- add test for all of this

PR #4537, opened 2020-02-19
Store name of float data arrays and unit names

hroest:
a couple of features:

- Keep the name of float data arrays (instead of deleting them)
- fix check for numerical value for data arrays (relates to content of array)
- store unit metadata for float data array descriptor
- add test for all of this

this is a partial PR based on #4436


timosachsenberg:
looks good to me

PR #4540, opened 2020-02-24
use separate testfile for FidoAdapter tests

bernt-matthias:
Before test1 and test6 used the same file for the comparison
which is fine if compared with FuzzyDiff, but leads to problems
for my Galaxy tests.

My change anctually does not change the current situation
for OpenMS, since the file used for comparison for test6 is
a symlink.


jpfeuffer:
does that work on win?

timosachsenberg:
@jpfeuffer should we duplicate that file or rely on symlinks?

jpfeuffer:
Well, I think it failed on Windows. So, duplicating would be the only platform-independent option.

jpfeuffer:
Hm sorry, but this is not a copy of FidoAdapter_1_output.idXML. At least not from develop. Maybe from 2.4?
We updated the path to the XSL.

bernt-matthias:
Done. Rebased and squashed the commits. Hope that I got it correct this time.

jpfeuffer:
Looks good. failures unrelated, I am currently testing external code snippets in the CI.
Let me restart travis with cache tho.

PR #4541, opened 2020-02-24
SpectraSTAdapter: remove apparently redundant extension

bernt-matthias:
and accept only the one that as commonly used in OpenMS


timosachsenberg:
could have some consequences if other tools produce pep.xml . I think I recall that some use that extension

bernt-matthias:
A grep for pep.xml in the ctd files generated by the OpenMS (2.5) tools gives only SpectraSTAdapter. 
(for pepXML there are 6 tools using this).

Indeed CometAdapter used pep.xml internally but outputs idXML. 

Apart from this 'pep.xml' is only mentioned in test data, a unit test, and doc.tex as far as I can see. 

Currently the PR only accects this Adapter, so I guess there should be no (negative) side effects within OpenMS. (But one could eliminate some more places where pep.xml is used)

bernt-matthias:
If you agree I would suggest to rename all test files from pep.xml to pepXML.

timosachsenberg:
It seems rather save tome but I would like to have some more input here @grosenberger @hroest I think you used e.g. SpectraST quite extensively with OpenSWATH. Do you think this will disrupt hybrid workflows (e.g. TPP + OpenMS) a lot?

PR #4542, opened 2020-02-24
Let NovorAdapter also use mzML

bernt-matthias:
It apparently does in the test, but now also the help is consistent (seems that Novor is the only tool that uses mzml instead of mzML)


giangiacomoberetta1:
Deas the use of mzML or mzml affect the peptide search results?

bernt-matthias:
No. As far as I recall, the test already uses mzML

PR #4544, opened 2020-02-26
uncomment a test comparison for IDPosteriorErrorProbability

bernt-matthias:
At least on Linux systems the comparison seems to work now. Lets see what the CI tests say.


timosachsenberg:
rebuild jenkins

bernt-matthias:
> rebuild jenkins

Do I need to do anything?

timosachsenberg:
tests are failing so probably this test is not stable on every platform

bernt-matthias:
If I understand it correctly the test that is failing (https://travis-ci.org/github/OpenMS/OpenMS/jobs/655271917) is on linux. Is there a easy possibility to see why it is failing?

PR #4549, opened 2020-02-28
change default number of reported hits per spectrum in comet to 1

timosachsenberg:


hroest:
does this affect any downstream tools? e.g. percolator or anything that re-ranks hits (you cant rerank hits if there is only one hit)

timosachsenberg:
no, not really. it's more about making the defaults a bit more consistent between tools.

PR #4552, opened 2020-02-28
MetaboliteAdductDecharger: don't write to opt out if absent

bernt-matthias:


FabianAicheler:
The consensusmap output is mandatory/the main output, wheras the other two are optional. So I would leave as was, at most one could consider whether to clarify this fact as comments.

bernt-matthias:
Thanks for the comment. In this case the parameter should be changed to mandatory. I change this accordingly.

timosachsenberg:
hmm not sure if that is the best solution.
I think it should be all optional and an error if none of them is set (we don't model that in CTD).
@jpfeuffer what is your take on that?

jpfeuffer:
Would someone ever want two outputs? If not, just make one "out" with restrictions "fxML,cXML" and potentially add an "out_type" parameter.

bernt-matthias:
OK. Also added a test. Since I currently don't have the time to compile it locally I can not contribute the test result files. Alternatively I could remove the DIFF calls from the tests.

bernt-matthias:
Argument help could need an update. Need your input on this.

timosachsenberg:
@FabianAicheler can you assist?

jpfeuffer:
@FabianAicheler You can use https://gitpod.io/#https://github.com/OpenMS/OpenMS/pull/4552
to quickly edit this pull request and push your changes (in case you do not want to set up a new build environment)

FabianAicheler:
Currently looking into whether collaborators need both outputs.

> 
> 
> Argument help could need an update. Need your input on this.

What specifically do you mean?

bernt-matthias:
> What specifically do you mean?

Currently the command line help of the outputs just says "output" or "output file" which seems way to minimalistic :)

FabianAicheler:
So we learned that there are indeed use case scenarios where one is interested in both consensusXml together with featureXml outputs.

For use with Knime, it is preferable that outputs are not flagged as mandatory.

@bernt-matthias Would it be okay for use with Galaxy if we flag all outputs as optional, and only write to them if output filenames were specified for them?

FabianAicheler:
Also sorry about changing statements regarding whether/which output should be mandatory^^ Historically, I recalled there to be some restrictions for Knime, but nowadays node requirements are more flexible.

So the first commit (0fb3e57 ) looks good to me (if it passes tests), I'll think of more verbose descriptions of the outputs.

timosachsenberg:
anything else needed here?

bernt-matthias:
I guess most of the 2nd and 3rd commit should be dropped (I guess the test `!outfile_cm.empty()` should be kept). 

Will try to do this soon. 

@FabianAicheler might add a "more verbose descriptions of the outputs"

bernt-matthias:
OK. Dropped the last two commits (but kept the tests .. without the time to think about them).

FabianAicheler:
I think(?) that removal of the out_type in the test is required, and a small adaption for the second test in order to cover all three potential outputs. Will look into it later today.

FabianAicheler:
OK, so I did some changes to the tests (and added test files) in gitpod. But the gitpod interface is a bit finnicky, and I am not sure whether I really pushed my changes somewhere (or whether that was transformed into a pull request to bernt-matthias).

Corresponding tests worked there. Just wanted to let you all know, will try to clarify with Julianus or other gitpod experts later in Discord(?)^^

bernt-matthias:
@FabianAicheler just checked there is no pull request to my fork.

FabianAicheler:
@bernt-matthias Ok, you should hopefully have a pull request now^^

bernt-matthias:
merged

oliveralka:
looks good!

PR #4553, opened 2020-02-28
MascotAdapter: specify allowed output formats

bernt-matthias:


jpfeuffer:
mascot_out is a flag. It can't have restrictions. You would need to allow idXML and mgf in the out Parameter and later set whatever depends on the old mascot_out accordingly.
This is a nice change though, thanks.

bernt-matthias:
Sure. Then lets add also input format restrictions. 

Then the two flags are not really necessary, or? Seems that everything is determined by the by formats given by the -in and -out parameters.

src/topp/MascotAdapter.cpp
```diff
@@ -353,11 +351,13 @@ class TOPPMascotAdapter :
...
-    mascot_in = getFlag_("mascot_in");
-    mascot_out = getFlag_("mascot_out");
+    FileTypes::Type in_type = FileHandler::getType(inputfile_name);
+    FileTypes::Type out_type = FileHandler::getType(outputfile_name);
+    mascot_out = in_type == FileTypes::MASCOTXML;
+    mascot_in = out_type == FileTypes::MGF;
     if (mascot_out && mascot_in)
     {
-      writeLog_("Both Mascot flags set. Aborting! Only one of the two flags [-mascot_in|-mascot_out] can be set!");
+      writeLog_("When the input file is a mascotXML, only idXML can be written! Please change the output type.");
```
bernt-matthias: Would also add the other in/out combination to the error message.

jpfeuffer:
tests need to be adapted. windows is currently not functional, can be ignored for now.

PR #4560, opened 2020-03-05
[Fix] doc

jpfeuffer:
Should work now. But if someone introduces another layer of folders or long template classes it will crash.


jpfeuffer:
I also restructured the build jobs into folders, so we can shorten the build workspace names a little more.

jpfeuffer:
Won't be testable in CI since we don't build docu. I built using the manual build job, however.

PR #4562, opened 2020-03-05
MassCalculator: output is csv

bernt-matthias:
not sure about about the `-in` parameter. Could be csv,tsv,txt since 

> Whitespace, commas or semicolons can de used to delimit the different items.


timosachsenberg:
I think we can just use tsv for the input but keep the "\t,;"  code for backwards compatibility

bernt-matthias:
Sure. It would also be easy to change the output to tsv..?

timosachsenberg:
sure go for it

jpfeuffer:
> Could be csv,tsv,txt

Why did you go for tsv-only in the end? In any case, the filenames of the test input should be changed accordingly or "txt" has to be allowed additionally.

bernt-matthias:
OK. I can add csv to the allowed extensions for the input.

For the output I would suggest to drop the separator argument and replace it with an optional out_type argument. Then the separator would be determined from this argument or the file extension. Is there another program where I could get inspiration for this?

timosachsenberg:
I would not make it too complicated and just use tab as fixed separator and tsv consistently.

bernt-matthias:
As you wish, just tell me if there is anything else to do.

timosachsenberg:
As always, thanks a lot Bernt! Highly appreciated

PR #4565, opened 2020-03-06
[FIX] prevent filtering of very sharp peaks

timosachsenberg:


jpfeuffer:
Was that regarding the mailing list error?

timosachsenberg:
yes we looked into their data and they have really nice sharp peaks with < 3 s FWHM.
This should definitely not be filtered out by default

Issue #4563, opened 2020-03-05
Some Elements missing in in Elements.xml


jamesETsmith:
Hello, 

I'm using `pyOpenMS` to help build some data analysis tools for a MS experiment involving inorganic compounds. When trying to predict the isotopic abundance of Indium containing compound I noticed that indium and several other elements were missing from the `Elements.xml` file. Is this intentional? If not, would it be possible to add them? I'm happy to make a pull request if you want. 

I understand that I can modify the file locally, but it would be a better long term solution, in my opinion, if `pyOpenMS` contained elements like Indium natively.

Here is a minimal example to reproduce the error:

```python
>>> from pyopenms import EmpiricalFormula, CoarseIsotopePatternGenerator
>>> wm = EmpiricalFormula("In")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pyopenms/pyopenms_4.pyx", line 2778, in pyopenms.pyopenms_4.EmpiricalFormula.__init__
  File "pyopenms/pyopenms_4.pyx", line 2757, in pyopenms.pyopenms_4.EmpiricalFormula._init_1
RuntimeError: 'In' found. Please use only valid element identifiers or modify share/OpenMS/CHEMISTRY/Elements.xml! in: Unknown element 'In'
```

Thanks,
James

timosachsenberg:
Hi, thanks for reporting! We would be really happy if you could open a PR and add the missing elements.

hroest:
We currently only represent 63 out of 118 elements:

```
$ grep -I AtomicNumber share/OpenMS/CHEMISTRY/Elements.xml   | sort | uniq | wc
     63     311    3578
```

and we are missing 21, 36, 39, 41, 43, 45, 49, 54, 59 (Sc, Kr, Y, Nb, Tc, Rh, In, Xe, Pr) and then whole bunch after. Granted, not the most important ones but still we should add a few more in that range

jpfeuffer:
We should decide which other Elements we would like and add them, too.

poshul:
@cbielow I see you have a comment in the Elements.xml about Deuterium and Tritium causing mem-leaks. Can you expand on that. Is this something that means that we should limit the addition of new elements?

poshul:
Elements have been moved to:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/ElementDB.cpp

PR #4566, opened 2020-03-06
Adding isotopic abundances for several elements

jamesETsmith:
This PR fixes the #4563 and adds Indium (and its isotopes) along with the other elements mentioned in the issue. 

If there are other elements you'd like me to add feel free to let me know.

-James


jpfeuffer:
Thanks a lot! I found some one-off errors, can you double-check if I corrected them right?

jamesETsmith:
Shoot, thanks for catching those! They look good to me.

timosachsenberg:
Thanks!

PR #4571, opened 2020-03-09
Remove python 2 from manylinux script

### LINKED ISSUE #4567: pyopenms dist scripts do not work
https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/dist-scripts/create-manylinux.sh#L35

`/data/create-manylinux.sh: line 36: /opt/python/cp27*/bin/pip: No such file or directory`

Seems like they dropped support for py2?
@hroest should we just remove the py2 parts?

jpfeuffer:
fixes #4567


jpfeuffer:
close in favor of #4574

PR #4574, opened 2020-03-09
pyOpenMS Release fix

hroest:


hroest:
fixes #4567

jpfeuffer:
I can confirm that it works (first py version ran through).
What about removing the restriction on autowrap? Wasn't our extension fixed to work with newer ones or do I remember wrongly?

PR #4578, opened 2020-03-10
Support for SLIM-labeling in FeatureFinderIdentification

hendrikweisser:
Adds a parameter ("model:each_trace") to FFId which enables fitting of elution models (Gaussian or EGH) to each mass trace in a feature. This is useful for data analysis in SLIM-labeling experiments (https://pubmed.ncbi.nlm.nih.gov/28821603), where intensities of different isotopologue peaks are compared.


src/openms/source/TRANSFORMATIONS/FEATUREFINDER/ElutionModelFitter.cpp
```diff
@@ -113,16 +116,97 @@ double ElutionModelFitter::calculateFitQuality_(const TraceFitter* fitter,
...
+  // record model parameters:
+  double center = fitter->getCenter(), height = fitter->getHeight();
+  feature.setMetaValue("model_height", height);
+  feature.setMetaValue("model_FWHM", fitter->getFWHM());
+  feature.setMetaValue("model_center", center);
+  feature.setMetaValue("model_lower", fitter->getLowerRTBound());
+  feature.setMetaValue("model_upper", fitter->getUpperRTBound());
+  if (asymmetric)
+  {
+    EGHTraceFitter* egh = static_cast<EGHTraceFitter*>(fitter);
```
timosachsenberg: these casts smell bad but not sure if its worth to refactor the TraceFitter class

timosachsenberg:
looks good!

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/ElutionModelFitter.cpp
hendrikweisser: That's how it's been before, so maybe an issue for another time.

hendrikweisser:
Test failures on Windows seem unrelated (InspectAdapter failed to write index file), so good to merge?

PR #4582, opened 2020-03-13
[Feature] Refactor QualityControl tool

jpfeuffer:
To make it more useable from the library.


src/openms/include/OpenMS/METADATA/MetaInfoInterfaceUtils.h
```diff
@@ -100,6 +102,17 @@ namespace OpenMS
       return common_keys;
     }
 
+    /// function to copy all meta values from one object to another
+    //TODO get a MetaValue list to copy only those that have been set
+    static void copyMetaValues(const MetaInfoInterface& from, MetaInfoInterface& to)
```
timosachsenberg: this could also go into the MetaInfoInterface class as: addMetaValues(...) but fine with me as is

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
```diff
@@ -102,6 +102,8 @@ namespace OpenMS
      * @return Status for RAWMZML and POSTFDRFEAT
      */
     QCBase::Status requires() const override;
+
+    void addMetaDataMetricsToMzTab(MzTabMetaData& meta);
```
timosachsenberg: hmm don't know if I would add the dependency to MzTab to this class.
if you don't want to go with a new class that takes QC results and annotates in mzTab then 
I would suggest making this method static.

src/openms/include/OpenMS/QC/QCBase.h
```diff
@@ -52,6 +53,44 @@ namespace OpenMS
   class OPENMS_DLLAPI QCBase
   {
   public:
+
+    /// two way mapping from ms-run-path to protID|pepID-identifier
+    struct Mapping
```
timosachsenberg: sounds useful for other parts in OpenMS as well

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
jpfeuffer: But it is using the results from inside this class instance. I don't think it can be static, can it?

jpfeuffer: I guess I could forward declare the class to save the import

src/openms/include/OpenMS/QC/QCBase.h
jpfeuffer: true. suggestion? new class? a bit of a hassle for this little thing, but ok.

src/openms/include/OpenMS/METADATA/MetaInfoInterfaceUtils.h
jpfeuffer: yes, thought about it, too. I think I will do it.

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
jpfeuffer: What about putting it into MzTab? Then it's just the other way around I guess.

timosachsenberg: Not too important right now. Alternatively one could also think about a more low level return type instead of annotating directly. returning a String or StringList or key value pairs maybe. Don’t know if that is too cumbersome though

src/openms/include/OpenMS/QC/QCBase.h
timosachsenberg: Maybe in protein Id ?

src/openms/include/OpenMS/KERNEL/FeatureMap.h
```diff
@@ -264,6 +264,19 @@ namespace OpenMS
     */
     OPENMS_DLLAPI void clear(bool clear_meta_data = true);
 
+    /// applies a function on all PeptideHits or only assigned ones
+    OPENMS_DLLAPI void applyFunctionOnPeptideHits(std::function<void(PeptideHit&)>& f, bool include_unassigned = true);
```
cbielow: std::function is SLOW...
probably better to use a function-template and pass in a lambda...

src/openms/include/OpenMS/METADATA/MetaInfoInterfaceUtils.h
cbielow: the function name is also misleading. Should be 'addMetaValues()'...
This is already fixed in a branch of mine somewhere but you can go ahead .. I'll rebase and merge later

jpfeuffer:
ubuntu unrelated. merge?

PR #4584, opened 2020-03-13
Added global visibility hidden preset, to silence warnings on macos. …

jpfeuffer:
…Before just libraries were built with hidden.


timosachsenberg:
feel free to merge

jpfeuffer:
hmm wait with it, I think it was not this but rather https://github.com/OpenMS/contrib/pull/99
that fixed it.

PR #4586, opened 2020-03-15
[TEST] Github style action

jpfeuffer:


jpfeuffer:
Interesting idea, but this specific action seems to be very badly implemented. Will close. Our own cppcheck suffices.

PR #4587, opened 2020-03-15
[FEATURE] Nightly pyopenms wheel through GitHub Actions

jpfeuffer:
TODO:
- [x] test wheels
- [x] create pyopenms nightly pip package (@hroest should we use a common account?)
- [x] add twine upload to workflow (there is a GitHub action for it; depends on the new pip package) We have to discuss who opens the package. We will need to upload it with the corresponding account.
- [x] add windows build
- [ ] remove unnecessary logs by piping to a log file or dev/null
- [x] (optional) try modifying the linux build script to only build the library once to save execution time


oliveralka:
@jpfeuffer  Would it be easier to have an additional workflow for windows? Since the contrib/OpenMS has to be build with the Visual Studio 16 2019 generator. The paths have to be set differently etc. What do you think?

jpfeuffer:
Windows has bash support in GitHub actions. This should be fine. See my actions in the contrib. Generators can be controlled by variables.
We can share as many steps as possible but if we think we should use a separate one it's also no problem.

jpfeuffer:
You need to use "shell: bash" for example. And load the Visual Studio environment variables first. But then it should be roughly the same.

.github/workflows/pyomslnxwheel.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     - name: Load contrib build
       run: |
-          curl -o contribbld.tar.gz https://abibuilder.informatik.uni-tuebingen.de/archive/openms/contrib/windows/x64/msvc-15/contrib_build.tar.gz
+          curl -o contribbld.tar.gz https://abibuilder.informatik.uni-tuebingen.de/archive/openms/contrib/windows/x64/msvc-14.2/contrib_build.tar.gz
```
timosachsenberg: so contrib is build with 14.2 and this is build with 16?

jpfeuffer: no, I dont know where CMake takes their numbers from. They probably just internally increase numbers now. But default toolchain in VS2019 is msvc14.2

jpfeuffer: I recently switched the labeling of the contrib builds to the actual compiler version, so it is easier to see which are compatible.

oliveralka: contrib = VS 2019 x64 (?) 
github env = Windows Server 2019 (VS 2019)
qt = win64_msvc2017_64 

I guess that is not going to work anyhow.

jpfeuffer: Haha confusing. Server2019 is the name of the OS.
And VS2017 and 2019 are compatible. But yeah, we should change to Qt for vs2019 if already provided by the Qt Company (they are always a bit late)

oliveralka: as far as I have seen https://github.com/jurplel/install-qt-action
It is not yet available (newest: win64_msvc2017_64), but that should work / compatibility wise?

.github/workflows/pyomslnxwheel.yml
```diff
@@ -45,8 +45,11 @@ jobs:
 
     - name: Load contrib build
       run: |
+          mkdir contribbld
+          cd contribbld
           curl -o contribbld.tar.gz https://abibuilder.informatik.uni-tuebingen.de/archive/openms/contrib/windows/x64/msvc-14.2/contrib_build.tar.gz
           tar -xzf contribbld.tar.gz
+          rm contribbld.tar.gz
```
oliveralka: should it automatically extract the directory with the given name?
 tar -xzf contribbld.tar.gz -> contribbld

oliveralka:
Maybe we have to use the visual studio command prompt: 
https://github.com/marketplace/actions/enable-developer-command-prompt
instead of the git bash?

oliveralka:
What is your suggestion in terms of testing the wheels? 
In general testing them on the system they have been build usually works - since it will find all the libraries even if there are problems with the paths (e.g. on other systems)? 

Or do you want to see if it can be installed and run a small ".py" test. 

We could for example build and install: 
```
from pyopenms import *
spectrum = MSSpectrum()
mz = range(1500, 500, -100)
i = [0 for mass in mz]
spectrum.set_peaks([mz, i])
```
then export the mzML and against a previously imported one?

jpfeuffer:
For now I thought about manual testing. But true we should extend the workflow to at least spawn fresh instances without a built OpenMS and its dependencies, copy the artifacts from the last job and test in a conda environment. For Linux we could even switch from the manylinux container to Ubuntu-latest.

jpfeuffer:
Looks good. Remember for the rest (lnx, win), you can easily make a matrix build. See the contrib actions:
https://github.com/OpenMS/contrib/blob/master/.github/workflows/main.yml

jpfeuffer:
And for debugging, maybe it works if you remove the "needs:" directive (maybe/hopefully it uses the artifacts of the last build)

jpfeuffer:
Puh.. seems like their artifact support is pretty crap. (You can't download artifacts from older runs or other workflows). https://github.com/actions/download-artifact/issues/3

Maybe use the direct link for debugging now: https://github.com/OpenMS/OpenMS/suites/524734819/artifacts/2984697

.github/workflows/pyomslnxwheel.yml
```diff
@@ -382,27 +382,27 @@ jobs:
...
+          wget --no-parent -r -P $HOME/macwheels_abi "https://abibuilder.informatik.uni-tuebingen.de/archive/openms/pyopenms/macOS/"  
         #uses: actions/download-artifact@v1
         #with:
           #name: winwheels 
-        shell: bash 
+
+        if: runner.os == 'macOS'
         run: |  
           wget --no-parent -r -P $HOME/macwheels_abi "https://abibuilder.informatik.uni-tuebingen.de/archive/openms/pyopenms/macOS/"  
-      if: runner.os == 'macos-latest'
```
jpfeuffer: Why not use the runner.os variable to construct the correct link? Then you only need one step.

oliveralka: I have seen that - similar to your miniconda code, but I am unsure it that works as well if we want to use the "download/artifacts" later on.

jpfeuffer: I see. I would hope it works like `artifact: name: ${{runner.os}}-wheels` but you never know with their crap actions.

oliveralka:
@jpfeuffer 
I guess we could switch them off for the builds, or? 
WITH_GUI=On/Off
ENABLE_TUTORIALS=On/Off
ENABLE_DOCS=On/Off

If we have log files - should they be uploaded as well? 
Do you have an example of what you deem unnecessary which could be piped to dev/null (just as a reference)?

jpfeuffer:
I think the flags are not necessary (the OpenMS and pyopenms targets do not build Docu or GUI anyway) but won't hurt (especially since we might not need QtSvg during CMake config stage then).
The first thought about piping to log files came from the Miniconda installation and environment setup. We would only need to see errors there (but uploading logs with the artifacts would also be nice).
I did not think about the stripping on linux. Hmm let's see.

jpfeuffer:
Just copy the libs instead of the whole build folder:
```
for lib in *so
do
  cp $lib $lib.tmp
done
```
strip...
copy back

.github/workflows/pyomslnxwheel.yml
```diff
@@ -329,6 +337,14 @@ jobs:
               cd ../..
               rm -rf wheelhouse/fix/
               mv wheelhouse/* $GITHUB_WORKSPACE/data/wheelhouse/
+
```
oliveralka: probably in the wrong directory?

oliveralka:
@jpfeuffer: Ok, that did the trick - I think a little bit of cleanup is still necessary.
Linux build dropped from 2h 54 min to 1h 58min!

hroest:
sounds good, so the plan is a new Python package "pyopenms-nightly" and upload it nightly with twine? Yes, sounds like a good idea, I suggest to create a new account for this on PyPI and we have to think about versioning, maybe just call it after the day? Eg pyopenms-nightly.2020.03.25 as an example? The version number needs to increase for `pip` to install a newer version.

jpfeuffer:
Ah yes, then I agree that the date in yyyy.mm.dd format is a good idea.

jpfeuffer:
pypi down.. have to try tomorrow.

jpfeuffer:
Ok I added a step but we need to set the version number in setup.py correctly first. Maybe also do the first uploads to testpypi.
Maybe add a build number in case we have multiple per day (otherwise I think we can't upload for testing)

.github/workflows/pyopenms-wheels.yml
```diff
@@ -0,0 +1,415 @@
...
+        path: OpenMS
+
+    - name: Build on manylinux2014 
+      run: |
+            # Apply a patches / fixes
+            # 1. Updated readme (due to pypi.org not displaying markdown)
+            # 2. Updated version number
+            # 3. Patch init (since the linked libraries are all in .lib)
+            cd OpenMS
+            wget https://raw.githubusercontent.com/hroest/OpenMS/0f3a2e8833d18fc54b7de86e6d67ab7349e828a0/src/pyOpenMS/README.rst -O src/pyOpenMS/README.rst
```
jpfeuffer: Can we somehow get rid of those patches?
1) E.g. just include that separate README.rst in the pyopenms directory?
2) not necessary when building from a git repo. We need to use a nightly version anyway. I am preparing a commit.
3) Figure out why boost is suddenly needed and maybe find a way to autodetermine what needs to be linked?
4) Why do we keep /OpenMS/src/pyOpenMS/pxds/SwathMapMassCorrection.pxd if it causes problems?

jpfeuffer:
I read that the pypi upload action only works on Linux. So we need to make a separate step after the tests on Linux where we get all the wheels again and upload.
We also need to check the wheels. Seems like they are getting bigger with every pyversion. We might include things multiple times due to the incremental builds. The last one exceeds the 60 mb limit of pypi. We need to write them if we can't go under that.

jpfeuffer:
And: the hour suggestion of @oliveralka for builds on the same day (mainly for testing) does not work if we reconfigure. The first builds might be one hour earlier than others. The incomplete results are here: https://test.pypi.org/project/pyopenms-nightly/2.5.0.2020032820/#files

oliveralka:
Ok, sizewise it seems that the pyopenms libraries get dragged over to the new iteration: 
In the wheel of macOS p38 (last) you can see:


```
Constants.py
License.txt
Plotting.py
__init__.py
all_modules.py
docompile.py
libOpenMS.dylib
libOpenSwathAlgo.dylib
libSuperHirn.dylib
libz.1.dylib
pyopenms_1.cpython-36m-darwin.so
pyopenms_1.cpython-37m-darwin.so
pyopenms_1.cpython-38-darwin.so
pyopenms_2.cpython-36m-darwin.so
pyopenms_2.cpython-37m-darwin.so
pyopenms_2.cpython-38-darwin.so
pyopenms_3.cpython-36m-darwin.so
pyopenms_3.cpython-37m-darwin.so
pyopenms_3.cpython-38-darwin.so
pyopenms_4.cpython-36m-darwin.so
pyopenms_4.cpython-37m-darwin.so
pyopenms_4.cpython-38-darwin.so
pyopenms_5.cpython-36m-darwin.so
pyopenms_5.cpython-37m-darwin.so
pyopenms_5.cpython-38-darwin.so
pyopenms_6.cpython-36m-darwin.so
pyopenms_6.cpython-37m-darwin.so
pyopenms_6.cpython-38-darwin.so
pyopenms_7.cpython-36m-darwin.so
pyopenms_7.cpython-37m-darwin.so
pyopenms_7.cpython-38-darwin.so
pyopenms_8.cpython-36m-darwin.so
pyopenms_8.cpython-37m-darwin.so
pyopenms_8.cpython-38-darwin.so
python_extras.py
qt_version_info.py
share
sysinfo.py
version.py

```

oliveralka:
Size-wise seems to be fine now - linux: 40 MB, macOS: 25 MB, windows: 30 MB per .whl.

jpfeuffer:
I am rearranging the steps. So many incredible workarounds since GitHub's own actions can't even do basic stuff.

hroest:
Seems like it worked: https://pypi.org/project/pyopenms-nightly/#history

jpfeuffer:
Yep, needed two smaller fixes after the merge. But now it should work every night unless there is a hiccup with the GitHub hosted runners.

PR #4589, opened 2020-03-16
less error-prone PEP/PSM mapping in mzTab

cbielow:
#4588, just against release2.5

@jpfeuffer I know its a "bit" of manual work (do you accept bribes? Mate tea perhaps? )


timosachsenberg:
I think we want to go for a 2.6 release soonish and @jpfeuffer is untouchable so could you live with that?

PR #4590, opened 2020-03-17
expose scaling parameters in SimpleSVM and add decoy generation in SSE

timosachsenberg:
Not ready for merge yet. Still needs some cleanup.


jpfeuffer:
Why not use the DecoyGenerator functionality? Actually I think it is not trivial to construct good decoys. E.g. just reversing proteins will not properly account for trypsin with P rule cutting sites.

timosachsenberg:
I agree. Will move to class

timosachsenberg:
python bindings are still missing

src/openms/source/CHEMISTRY/DecoyGenerator.cpp
```diff
@@ -35,86 +35,112 @@
...
-AASequence DecoyGenerator::shufflePeptide(
+AASequence DecoyGenerator::shufflePeptides(
         const AASequence& protein,
         const String& protease,
         const int max_attempts,
         int seed)
 {
-    std::mt19937 rng(seed);
+  OPENMS_PRECONDITION(!protein.isModified(), "Decoy generation only supports unmodified proteins.");
+  std::mt19937 rng(seed);
```
timosachsenberg: hmm seems to lead to different results on windows. Any suggestions on how to make sure the tests have identical results on all platforms?

jpfeuffer: Seems like you have to write your own shuffle. the twister should be platform-independent.

timosachsenberg: yep seems so... would have been nice to let users choose if they need the X% faster implementation or something that is platform-independent 
https://stackoverflow.com/questions/51929085/cross-platform-random-reproducibility

timosachsenberg: @jpfeuffer @cbielow where should I put a "portable_shuffle" somewhere in Math or Concept?

timosachsenberg: or just use https://www.boost.org/doc/libs/1_71_0/libs/range/doc/html/range/reference/algorithms/mutating/random_shuffle.html ?

timosachsenberg:
Ok ready for review

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -245,39 +260,73 @@ void SimpleSearchEngineAlgorithm::postProcessHits_(const PeakMap& exp,
...
-#pragma omp parallel for default(none) shared(annotated_hits, exp, fixed_modifications, variable_modifications, peptide_ids, max_variable_mods_per_peptide)
+    bool annotation_precursor_error_ppm = std::find(annotate_psm_.begin(), annotate_psm_.end(), "precursor_error_ppm") !=  annotate_psm_.end();
+    bool annotation_fragment_error_ppm = std::find(annotate_psm_.begin(), annotate_psm_.end(), "median_fragment_error_ppm") != annotate_psm_.end();
+
+#pragma omp parallel for default(none) shared(annotated_hits, exp, fixed_modifications, variable_modifications, peptide_ids, max_variable_mods_per_peptide, annotation_precursor_error_ppm, annotation_fragment_error_ppm)
     for (SignedSize scan_index = 0; scan_index < (SignedSize)annotated_hits.size(); ++scan_index)
     {
       if (!annotated_hits[scan_index].empty())
       {
+        const MSSpectrum& spec = exp[scan_index];
```
jpfeuffer, line 271: maybe add a bounds/size check at least before the loop?

src/openms/source/CHEMISTRY/DecoyGenerator.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include <boost/date_time/posix_time/posix_time_types.hpp>
+
+#include <algorithm>
+
+using namespace OpenMS;
+
+DecoyGenerator::DecoyGenerator()
+{
+  // find a seed:
+  // C++11 note: C++ build-in alternative once C++11 can be presumed: 'std::chrono::high_resolution_clock'
```
jpfeuffer, line 49: aren't we requiring C++11?

src/openms/include/OpenMS/CHEMISTRY/DecoyGenerator.h
```diff
@@ -0,0 +1,103 @@
...
+            );
+    
+    private:
+      // sequence identity by matching AAs
+      static double SequenceIdentity_(const String& decoy, const String& target);
+
+      template <class RandomAccessIterator>
+        void shuffle_ (RandomAccessIterator first, RandomAccessIterator last)
+      {
+        for (auto i = (last-first)-1; i > 0; --i) // OMS_CODING_TEST_EXCLUDE
```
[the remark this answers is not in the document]
timosachsenberg: yes because last is either .end() or end() - 1 in this class.

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -124,6 +130,13 @@ namespace OpenMS
     defaults_.setValue("enzyme", "Trypsin", "The enzyme used for peptide digestion.");
     defaults_.setValidStrings("enzyme", all_enzymes);
 
+    defaults_.setValue("decoys", "false", "Should decoys be generated?");
+    defaults_.setValidStrings("decoys", {"true","false"} );
+
+    defaults_.setValue("annotate:PSM", StringList{}, "Annotations added to each PSM.");
+    defaults_.setValidStrings("annotate:PSM", StringList{"median_fragment_error_ppm","precursor_error_ppm"});
```
timosachsenberg, line 137: TODO: add string constants from Constants

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -245,39 +260,73 @@ void SimpleSearchEngineAlgorithm::postProcessHits_(const PeakMap& exp,
       annotated_hits.shrink_to_fit();
     }
 
-#pragma omp parallel for default(none) shared(annotated_hits, exp, fixed_modifications, variable_modifications, peptide_ids, max_variable_mods_per_peptide)
+    bool annotation_precursor_error_ppm = std::find(annotate_psm_.begin(), annotate_psm_.end(), "precursor_error_ppm") !=  annotate_psm_.end();
```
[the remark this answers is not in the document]
timosachsenberg: yep. TODO: add string constant from Constants

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
timosachsenberg: If the spectrum has been annotated it must exist.

src/openms/source/CHEMISTRY/DecoyGenerator.cpp
timosachsenberg: ok will check if supported

PR #4591, opened 2020-03-18
add additional debug options and prelimary support for percolator 3.0…

timosachsenberg:
…4 features


src/topp/PercolatorAdapter.cpp
```diff
@@ -452,16 +470,24 @@ class PercolatorAdapter :
...
+          calc_mass = hit.getMetaValue("CalcMass");
+        }
 
         if (hit.metaValueExists("IsotopeError"))  // MSGFPlus
         {
           float isoErr = hit.getMetaValue("IsotopeError").toString().toFloat();
           exp_mass = exp_mass - (isoErr * Constants::C13C12_MASSDIFF_U) / charge;
         }
-        else if (hit.metaValueExists("isotope_error")) // e.g. SimpleSearchEngine /RNPxlSearch
+        else if (hit.metaValueExists("isotope_error")) // e.g. SimpleSearchEngine /OpenNuXL
```
jpfeuffer: Can't we unify this MetaValue?

timosachsenberg: you mean put it here https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CONCEPT/Constants.cpp#L173

timosachsenberg: the thing is that MSGFPlus uses IsotopeError and we usually the lower case names.

jpfeuffer: And why?

timosachsenberg: maybe add isotope_error to the Constants file and make MSGFPlus adapter name it IsotopeError

timosachsenberg: https://github.com/OpenMS/OpenMS/blob/d17cc251fd0c4068eb253b03c9fb107897771fdc/src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp#L49

https://github.com/OpenMS/OpenMS/blob/c6348e527cd58d7bad637fe8849deae4ca7a2ed2/src/topp/MSGFPlusAdapter.cpp#L608

https://github.com/OpenMS/OpenMS/blob/c6348e527cd58d7bad637fe8849deae4ca7a2ed2/src/topp/PercolatorAdapter.cpp#L459

These seem to be the only occurances

jpfeuffer: Yes, I would say stick to one for writing: But for now, allow both for reading.

jpfeuffer:
Great opportunity for updating THIRDPARTY btw ;)

timosachsenberg:
@enetz I renamed OMS:precursor_mz_error_ppm to precursor_mz_error_ppm in a hopefully backward compatible way.

PR #4593, opened 2020-03-18
[FEATURE] Allow calculation of FDR from protein group scores.

jpfeuffer:


jpfeuffer:
dont merge. has error for old gcc/boost apparently

jpfeuffer:
Note the mac build is a hiccup from CTest (merged lines parsed as error instead of warning.)

PR #4594, opened 2020-03-18
[Feature] Allow inference on consensusXML

jpfeuffer:
- Adds methods for working with IDs on cXML (FDR, Filtering)
- Refactored IDSwitcher to Algorithm
- Implement requestGeneralScoreType (with which you can choose a wanted score type in your ID and it checks for the dozens of different meta values)
- Implemented resolution on the IDBoostGraph data structure (before it was building its own)
- TODO more tests. Only the new resolution has one.
- Fixes HUGE bug in idXML writer, where it didn't bounds check for an existing protein accession and just associated those peptides with a random (usually first) protein.
- Don't allow Percolator to run on idXMLs with multiple runs since it just does nonsense.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h
@@ -114,6 +114,7 @@ namespace OpenMS
     /// Loops over all runs in the ConsensusMaps' protein IDs. (experimental)
     void inferPosteriorProbabilities(
         ConsensusMap& cmap,
+        bool greedy_group_resolution,
         boost::optional<const ExperimentalDesign> exp_des = boost::optional<const ExperimentalDesign>());
 
   private:
--- a/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
@@ -500,6 +500,11 @@ namespace OpenMS {
     void getUpstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
         bool stop_at_first, std::vector<vertex_t>& result);
 
+    /// @todo untested
+    /// Removes all edges from a peptide (and its PSMs) to its parent protein groups (and its proteins)
+    /// except for the best protein group
+    /// Also removes the corresponding PeptideEvidences in the underlying ID data structure for now
+    /// @todo make the removals optional
     void resolveGraphPeptideCentric_(Graph& fg);
 
     template<class NodeType>
--- a/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
@@ -366,10 +366,11 @@ namespace OpenMS {
 
     /// @todo untested
     /// Removes all edges from a peptide (and its PSMs) to its parent protein groups (and its proteins)
-    /// except for the best protein group
+    /// except for the best protein group.
+    /// @pre Graph must contain PeptideCluster nodes (e.g. with clusterIndistProteinsAndPeptides).
     /// @param removeAssociationsInData Also removes the corresponding PeptideEvidences in the underlying
     ///     ID data structure. Only deactivate if you know what you are doing.
-    void resolveGraphPeptideCentric(bool removeAssociationsInData);
+    void resolveGraphPeptideCentric(bool removeAssociationsInData = true);
 
 
 
@@ -506,6 +507,9 @@ namespace OpenMS {
 
     void getUpstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
         bool stop_at_first, std::vector<vertex_t>& result);
+
+    void getDownstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
+                                      bool stop_at_first, std::vector<vertex_t>& result);
 
 
     /// see equivalent public method
--- a/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
@@ -335,7 +335,7 @@ namespace OpenMS {
 
 
     /// Do sth on connected components (your functor object has to inherit from std::function or be a lambda)
-    void applyFunctorOnCCs(const std::function<unsigned long(Graph&)>& functor);
+    void applyFunctorOnCCs(const std::function<unsigned long(Graph&, unsigned int)>& functor);
     /// Do sth on connected components single threaded (your functor object has to inherit from std::function or be a lambda)
     void applyFunctorOnCCsST(const std::function<void(Graph&)>& functor);
 
--- a/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
@@ -364,6 +364,13 @@ namespace OpenMS {
     /// Splits the initialized graph into connected components and clears it.
     void computeConnectedComponents();
 
+    /// @todo untested
+    /// Removes all edges from a peptide (and its PSMs) to its parent protein groups (and its proteins)
+    /// except for the best protein group
+    /// @param removeAssociationsInData Also removes the corresponding PeptideEvidences in the underlying
+    ///     ID data structure. Only deactivate if you know what you are doing.
+    void resolveGraphPeptideCentric(bool removeAssociationsInData);
+
 
 
     /// Zero means the graph was not split yet
@@ -500,12 +507,9 @@ namespace OpenMS {
     void getUpstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
         bool stop_at_first, std::vector<vertex_t>& result);
 
-    /// @todo untested
-    /// Removes all edges from a peptide (and its PSMs) to its parent protein groups (and its proteins)
-    /// except for the best protein group
-    /// Also removes the corresponding PeptideEvidences in the underlying ID data structure for now
-    /// @todo make the removals optional
-    void resolveGraphPeptideCentric_(Graph& fg);
+
+    /// see equivalent public method
+    void resolveGraphPeptideCentric_(Graph& fg, bool removeAssociationsInData);
 
     template<class NodeType>
     void getDownstreamNodes(vertex_t start, Graph graph, std::vector<NodeType>& result)
--- a/src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
@@ -138,7 +138,14 @@ namespace OpenMS
         if (!ids.empty())
         {
           new_type = findScoreType(ids[0], type);
-          break;
+          if (new_type == ids[0].getScoreType())
+          {
+            return;
+          }
+          else
+          {
+            break;
+          }
         }
       }
       if (new_type.empty())
@@ -158,6 +165,7 @@ namespace OpenMS
 
       function<void(PeptideIdentification&)> f = [&counter,this](PeptideIdentification& id){switchScores(id,counter);};
       cmap.applyFunctionOnPeptideIDs(f, unassigned_peptides_too);
+
     }
 
 
@@ -170,7 +178,7 @@ namespace OpenMS
       if (possible_types.find(curr_score_type) != possible_types.end())
       {
         OPENMS_LOG_INFO << "Requested score type already set as main score: " + curr_score_type + "\n";
-        return "";
+        return curr_score_type;
       }
       else
       {
--- a/src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
@@ -1,0 +1,229 @@
+// --------------------------------------------------------------------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
+// ETH Zurich, and Freie Universitaet Berlin 2002-2020.
+//
+// This software is released under a three-clause BSD license:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of any author or any participating institution
+//    may be used to endorse or promote products derived from this software
+//    without specific prior written permission.
+// For a full list of authors, refer to the file AUTHORS.
+// --------------------------------------------------------------------------
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL ANY OF THE AUTHORS OR THE CONTRIBUTING
+// INSTITUTIONS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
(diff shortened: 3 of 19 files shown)


src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -500,7 +508,12 @@ namespace OpenMS {
     void getUpstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
         bool stop_at_first, std::vector<vertex_t>& result);
 
-    void resolveGraphPeptideCentric_(Graph& fg);
+    void getDownstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
```
timosachsenberg: underscore missing?

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,237 @@
...
+namespace OpenMS
+{
+
+  class OPENMS_DLLAPI IDScoreSwitcherAlgorithm:
+    public DefaultParamHandler
+  {
+  public:
+    IDScoreSwitcherAlgorithm ();
+
+    enum class ScoreType
```
timosachsenberg: add comments?  maybe even refer to MS OBO ?

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,237 @@
...
+        hit_it->setScore(hit_it->getMetaValue(new_score_));
+      }
+      id.setScoreType(new_score_type_);
+      id.setHigherScoreBetter(higher_better_);
+    }
+
+    /// Looks at the first Hit of the given @p id and according to the given @p type ,
+    /// deduces a fitting score and score direction to be switched to.
+    /// Then tries to switch all hits.
+    void switchToGeneralScoreType(vector<PeptideIdentification>& id, ScoreType type, Size& counter)
```
timosachsenberg: this sounds a bit like what the MS obo models with parent CV nodes of certain score types. I guess MS obo is not complete here otherwise it could be a good way to document that

jpfeuffer: Yes, I looked into it and it is far from complete. Also, it would be quite a large change and is more something for the new ID structure.

src/openms/source/FORMAT/IdXMLFile.cpp
```diff
@@ -357,7 +357,21 @@ namespace OpenMS
             // empty accessions are not written out (legacy code)
             if (!protein_accession.empty())
             {
-              protein_accessions.push_back("PH_" + String(accession_to_id[protein_accession]));
+              const auto acc = accession_to_id.find(protein_accession);
+              if (acc != accession_to_id.end())
+              {
+                protein_accessions.emplace_back("PH_" + String(prot_count));
```
jpfeuffer: should be acc->second. fixed.

src/openms/include/OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h
```diff
@@ -114,6 +114,7 @@ namespace OpenMS
     /// Loops over all runs in the ConsensusMaps' protein IDs. (experimental)
     void inferPosteriorProbabilities(
         ConsensusMap& cmap,
+        bool greedy_group_resolution,
         boost::optional<const ExperimentalDesign> exp_des = boost::optional<const ExperimentalDesign>());
```
timosachsenberg: downside of the boost::optional is that it is not easily wrapped in pyOpenMS

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,247 @@
...
+      {
+        if (!hit_it->metaValueExists(new_score_))
+        {
+          String msg = "Meta value '" + new_score_ + "' not found for " +
+                       describeHit_(*hit_it);
+          throw Exception::MissingInformation(__FILE__, __LINE__,
+                                              OPENMS_PRETTY_FUNCTION, msg);
+        }
+
+        String old_score_meta = (old_score_.empty() ? id.getScoreType() :
```
timosachsenberg: const&

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,247 @@
...
+                                 old_score_);
+        DataValue dv = hit_it->getMetaValue(old_score_meta);
```
timosachsenberg: const&

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,247 @@
...
+      {
+        String msg = "First encountered ID does not have the requested score type.";
+        throw Exception::MissingInformation(__FILE__, __LINE__,
+                                            OPENMS_PRETTY_FUNCTION, msg);
+      }
+      new_score_ = t;
+      new_score_type_ = t;
+      if (type != ScoreType::RAW && higher_better_ != type_to_better_[type])
+      {
+        OPENMS_LOG_WARN << "Requested non-raw score type does not match the expected score direction. Correcting!\n";
```
timosachsenberg: does this actually happen? would it be better to fail hard then?

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,247 @@
...
+      new_score_ = new_type;
+      new_score_type_ = new_type;
+
+      if (type != ScoreType::RAW && higher_better_ != type_to_better_[type])
+      {
+        OPENMS_LOG_WARN << "Requested non-raw score type does not match the expected score direction. Correcting!\n";
+        higher_better_ = type_to_better_[type];
+      }
+
+      function<void(PeptideIdentification&)> f = [&counter,this](PeptideIdentification& id){switchScores(id,counter);};
```
timosachsenberg: I prefer auto f =

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,247 @@
...
+      cmap.applyFunctionOnPeptideIDs(f, unassigned_peptides_too);
+    }
+
+
+    /// finds a certain score type in an ID and its metavalues if present, otherwise returns empty string
+    template <typename IDType>
+    String findScoreType(IDType& id, IDScoreSwitcherAlgorithm::ScoreType type)
+    {
+      String curr_score_type = id.getScoreType();
```
timosachsenberg: const & ?

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,247 @@
...
+      set<String>& possible_types = type_to_str_[type];
```
timosachsenberg: const set&  ?

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,247 @@
...
+        {
+          if (hit.metaValueExists(poss_str) || hit.metaValueExists(poss_str + "_score")) return poss_str;
+        }
+        OPENMS_LOG_WARN << "Score of requested type not found in the UserParams of the checked ID object.\n";
+        return "";
+      }
+    }
+
+  private:
+    String describeHit_(const PeptideHit& hit);
```
timosachsenberg: add one line comment that this is for error messages etc.

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
jpfeuffer: I mean, if someone forgets to set the param for the algorithm, this happens. E.g. if you just wanna use this method without setting up the whole algorithm.

src/openms/source/ANALYSIS/ID/ConsensusMapMergerAlgorithm.cpp
```diff
@@ -298,12 +299,23 @@ namespace OpenMS
...
+      OPENMS_LOG_WARN << "Inference was already performed on the runs in this ConsensusXML."
+                         " Merging their proteins, will invalidate correctness of the inference."
+                         " You should redo it.\n";
+      // deliberately do not take over old inference settings.
+    }
+
+    //we do it based on the IDRuns since ID Runs maybe different from quantification in e.g. TMT
     vector<String> mergedOriginFiles{};
-    for (const auto& chead : cmap.getColumnHeaders())
+    //set<String> mergedOriginFiles{};
```
timosachsenberg: remove commented out?

src/openms/source/FILTERING/ID/IDFilter.cpp
```diff
@@ -237,6 +237,43 @@ namespace OpenMS
   }
 
 
+  void IDFilter::removeUnreferencedProteins(ConsensusMap& cmap, bool include_unassigned)
+  {
+    // collect accessions that are referenced by peptides for each ID run:
+    map<String, unordered_set<String> > run_to_accessions;
+
+    function<void(const PeptideIdentification&)> f =
```
timosachsenberg: I prefer having auto and a meaningful name for the functor

src/openms/source/ANALYSIS/ID/ConsensusMapMergerAlgorithm.cpp
jpfeuffer: I am actually not sure, what to take here.

src/openms/include/OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h
jpfeuffer: Will think about when doing py bindings.

timosachsenberg:
Are there some new features that should be exposed via pyOpenMS? Otherwise looks good to me

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -364,20 +364,58 @@ namespace OpenMS {
...
+    /// @param removeAssociationsInData Also removes the corresponding PeptideEvidences in the underlying
+    ///     ID data structure. Only deactivate if you know what you are doing.
+    void resolveGraphPeptideCentric(bool removeAssociationsInData = true);
+
 
 
     /// Zero means the graph was not split yet
     Size getNrConnectedComponents();
 
+    /// \brief Returns a specific connected component of the graph as a graph itself
```
cbielow: I don't mind the `\`' style, but its unusual in our codebase `@`
Maybe change?

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -364,20 +364,58 @@ namespace OpenMS {
...
 
+    /// \brief Searches for all upstream nodes from a (set of) start nodes that are lower
+    ///    or equal than a given level. The ordering is the same as in the IDPointer variant typedef.
+    /// \param q a queue of start nodes
+    /// \param graph the graph to look in (q has to be part of it)
+    /// \param lvl the level to start reporting from
+    /// \param stop_at_first do you want to stop at the first node <= lvl or also report its
+    ///    upstream "predecessors"
+    /// \param result vector of reported nodes
+    void getUpstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
```
cbielow: Graph by value?

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
```diff
@@ -364,20 +364,58 @@ namespace OpenMS {
...
+
+    /// \brief Searches for all downstream nodes from a (set of) start nodes that are higher
+    ///    or equal than a given level. The ordering is the same as in the IDPointer variant typedef.
+    /// \param q a queue of start nodes
+    /// \param graph the graph to look in (q has to be part of it)
+    /// \param lvl the level to start reporting from
+    /// \param stop_at_first do you want to stop at the first node >= lvl or also report its
+    ///    upstream "predecessors"
+    /// \param result vector of reported nodes
+    void getDownstreamNodesNonRecursive(std::queue<vertex_t>& q, Graph graph, int lvl,
```
cbielow: Graph by value?

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,255 @@
...
+        OPENMS_LOG_WARN << "Score of requested type not found in the UserParams of the checked ID object.\n";
+        return "";
+      }
+    }
+
+  private:
+    /// \brief describes a PeptideHit for debug/exception messages
+    /// \param hit the hit
+    /// \return a String description to include in a message
+    String describeHit_(const PeptideHit& hit);
```
cbielow: does it make sense to move this to PeptideHit as operator<<? (same with ProteinHit)

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -0,0 +1,255 @@
...
+#pragma once
+
+#include <OpenMS/DATASTRUCTURES/DefaultParamHandler.h>
+#include <OpenMS/CONCEPT/Exception.h>
+#include <OpenMS/KERNEL/ConsensusMap.h>
+
+#include <vector>
+#include <set>
+
+using namespace std;
```
cbielow: in a header ... tse... :)

src/openms/source/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.cpp
```diff
@@ -70,13 +69,18 @@ namespace OpenMS
...
-    unsigned long operator() (IDBoostGraph::Graph& fg) {
+    unsigned long operator() (IDBoostGraph::Graph& fg, unsigned int idx) {
       //TODO do quick bruteforce calculation if the cc is really small?
-      cnt_++;
+
       // this skips CCs with just peps or prots. We only add edges between different types.
       // and if there were no edges, it would not be a CC.
       if (boost::num_vertices(fg) >= 2)
       {
+        OPENMS_LOG_DEBUG << "Running cc " << String(idx) << "..." << std::endl;
```
cbielow: this gets called very often (potentially), right?
Don't use `std::endl` then... awfully slow...

src/openms/source/ANALYSIS/ID/IDBoostGraph.cpp
```diff
@@ -954,7 +978,50 @@ namespace OpenMS
...
+
+        #ifdef INFERENCE_DEBUG
+        OPENMS_LOG_INFO << "Processing cc " << i << " with " << boost::num_vertices(curr_cc) << " vertices." << std::endl;
+        OPENMS_LOG_INFO << "Printing cc " << i << std::endl;
+        printGraph(LOG_INFO, curr_cc);
+        OPENMS_LOG_INFO << "Printed cc " << i << std::endl;
+        #endif
+
+        resolveGraphPeptideCentric_(curr_cc, removeAssociationsInData);
+        pl.setProgress(i);
```
cbielow: this will make the progress jump forth and back...
Depending on how expensive `resolveGraphPeptideCentric_` is, you can use 
`#pragma omp parallel for schedule(dynamic, 1)` to only process 1 chunk per thread at a time and then only use
`#pragma omp master` to set the progress. This will be somewhat accurate then.
For full accuracy, you'd need an atomic counter and report progress in masterthread.

src/openms/source/ANALYSIS/ID/IDBoostGraph.cpp
```diff
@@ -974,17 +1041,55 @@ namespace OpenMS
...
+        //TODO how to/if resolve ties AND allow preferring targets? We need to merge the PR with TD info for groups first.
+        set<String> accsToRemove;
         for (const auto& prot : prots)
         {
-          if (prot == *best_prot)
+          if (prot != *best_prot)
           {
+            if (fg[prot].which() == 1) // if the node is a group, find their members first.
+            {
+              vector<vertex_t> single_prots;
```
cbielow: lots of allocate/deallocate due to local vectors.
Pull out of loop and just clear to better perf.

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
jpfeuffer: you are right, my IDE autocompletes to this style. I found a way to change it though! 👍

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
jpfeuffer: whoops. thanks a lot. this would have been expensive.

src/openms/include/OpenMS/ANALYSIS/ID/IDBoostGraph.h
jpfeuffer: thx

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
jpfeuffer: Copy-paste from the refactored tool. Certainly! 💯

src/openms/source/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.cpp
jpfeuffer: Thanks. I was actually thinking about ways to avoid logging completely when multithreading. What do you think?

cbielow:
Please do not include `ostream` in a header. It's huge and compile time will suffer.
Use `iosfwd` instead.

PR #4598, opened 2020-03-20
question: shallow clones

timosachsenberg:
@jpfeuffer if we would make this a default... do you know by chance if its easy to do a full checkout if needed


jpfeuffer:
should be as easy as `git fetch --unshallow`
As long as all the automated scripts we have still work with this: 👍

timosachsenberg:
you know best ;)

timosachsenberg:
let's close this. seems a bit too dangerous

PR #4600, opened 2020-03-23
[Feature] Include calculation of QC parameters in ProteomicsLFQ tool

jpfeuffer:
TODO:
 - [ ] Adapt tests
 - [ ] Make optional?
 - [ ] Refactor copied functions
 - [ ] Check report when applying PTXQC on the gerenated mzTab
 - [ ] Check/Make fraction support of PTXQC compatible

### DIFF
--- a/src/utils/ProteomicsLFQ.cpp
+++ b/src/utils/ProteomicsLFQ.cpp
@@ -1037,10 +1037,12 @@ protected:
       for (auto & f : tmp)
       {
         //keep FWHM meta value for QC
-        double fwhm = static_cast<double>(f.getMetaValue("FWHM"));
+        double fwhm = static_cast<double>(f.getMetaValue("model_FWHM"));
         f.clearMetaInfo();
         f.setSubordinates({});
         f.setConvexHulls({});
+        // QC tools will create a FWHM metavalue as a copy anyway,
+        // so save it as FWHM right away
         f.setMetaValue("FWHM", fwhm);
       }
       feature_maps.push_back(tmp);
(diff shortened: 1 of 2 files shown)


jpfeuffer:
Current state (thanks @cbielow for running the tool):

[report_v1.0.3_lfqBSA_Downloads.html.zip](https://github.com/OpenMS/OpenMS/files/4371361/report_v1.0.3_lfqBSA_Downloads.html.zip)

src/utils/ProteomicsLFQ.cpp
```diff
@@ -1014,11 +1103,144 @@ class ProteomicsLFQ :
...
+                             << ", RT '" << pep_id.getRT() << "', m/z '" << pep_id.getMZ() << "' and identifier '" << pep_id.getIdentifier()
+                             << "' does not appear in any of the given FeatureXMLs. Check your input!\n";
+            incomplete_features = true;
+          }
+        };
+    consensus_fraction.applyFunctionOnPeptideIDs(f, true);
+    if (incomplete_features) return ILLEGAL_PARAMETERS;
+
+    // add new PeptideIdentifications (for unidentified MS2 spectra)
+    consensus_fraction.getUnassignedPeptideIdentifications().insert(consensus_fraction.getUnassignedPeptideIdentifications().end(), all_new_upep_ids.begin(), all_new_upep_ids.end());
```
timosachsenberg, line 1237: consider moving calculations from R to cpp QC classes

timosachsenberg:
anyupdate here?

jpfeuffer:
We still do not have any consensus on how to avoid the Unassigned IDs for every MS2 scan.
With the current state, this PR is out of the question for me.

jpfeuffer:
Also, as long as ptxqc does not support fractions in mzTab, I also don't think it is worth it

jpfeuffer:
I think I am going to deprecate this in favor of the new mzQC capabilities.
I will write out small mzQC files in addition to the mzTab and pmultiQC will read from both. Plus additional mzQC files written in intermediate steps.
I think this is the sanest way to go for a pipeline. This is basically the standard for genomics as well. I don't think it is feasible to have everything in one file. Especially since you only want summaries for intermediate steps.

jpfeuffer:
@timosachsenberg @cbielow What do you think?

jpfeuffer:
See also https://github.com/OpenMS/OpenMS/projects/50

timosachsenberg:
I agree.

timosachsenberg:
This one has been stale for quite a while. Any plans to continue or finish?

PR #4605, opened 2020-03-24
make add_decoys an advanced option in MSGF+ adapter

timosachsenberg:
rarely used as we need to index we a t/d database


jpfeuffer:
`Internal error: Request for parameter with wrong type 'add_decoys'`

timosachsenberg:
yep was a web edit.... was expecting some errors ;) will fix in gitpod later

cbielow:
Off topic: Can we silence this bot and only allow him to talk IFF there is a problem?

PR #4606, opened 2020-03-24
Added findHighestInWindow in MSSpectrum. Changed PrecursorCorrection …

JeeH-K:
…accordingly.

A piece of PrecursorCorrection code is extracted to findHighestInWindow for further use.

### DIFF
--- a/src/openms/source/FILTERING/CALIBRATION/PrecursorCorrection.cpp
+++ b/src/openms/source/FILTERING/CALIBRATION/PrecursorCorrection.cpp
@@ -201,22 +201,33 @@ namespace OpenMS
           continue;
         }
 
-        // get tolerance window and left/right iterator
+//        // get tolerance window and left/right iterator
+//        std::pair<double,double> tolerance_window = Math::getTolWindow(mz, mz_tolerance, ppm);
+//        MSSpectrum::ConstIterator left = rt_it->MZBegin(tolerance_window.first);
+//        MSSpectrum::ConstIterator right = rt_it->MZEnd(tolerance_window.second);
+//
+//        // no MS1 precursor peak in +- tolerance window found
+//        if  (left == right)
+//        {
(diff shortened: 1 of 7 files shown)


src/openms/source/FILTERING/CALIBRATION/PrecursorCorrection.cpp
```diff
@@ -201,23 +201,34 @@ namespace OpenMS
...
+//        MSSpectrum::ConstIterator max_intensity_it = std::max_element(left, right, Peak1D::IntensityLess());
+//
+//        // find peak (index) with highest intensity to expected position
+//        Size highest_peak_idx = max_intensity_it - rt_it->begin();
+
+        // get tolerance window and index of highest peak
         std::pair<double,double> tolerance_window = Math::getTolWindow(mz, mz_tolerance, ppm);
-        MSSpectrum::ConstIterator left = rt_it->MZBegin(tolerance_window.first);
-        MSSpectrum::ConstIterator right = rt_it->MZEnd(tolerance_window.second);
+        auto highest_peak_idx = rt_it->findHighestInWindow(mz, mz-tolerance_window.first, mz+tolerance_window.second);
```
timosachsenberg: I would use Int here instead of auto. It's even shorter ;)

timosachsenberg: if tests pass you can also remove the code in comments

timosachsenberg:
looks good!
two things are missing:

- [ ] pyOpenMS bindings to the new function (see the MSSpectrum.pxd file)
- [ ] a test

src/openms/source/FILTERING/CALIBRATION/PrecursorCorrection.cpp
JeeH-K: sure, thanks for the suggestion

PR #4608, opened 2020-03-25
[Feature] use merge idx to keep track of primaryRuns in Consenus/IDMerger

jpfeuffer:


jpfeuffer:
Failures only because of newly introduced external code tests.

jpfeuffer:
TODO would be more tests. But the general merging is tested in Epifany on cxml. I would need to create another downstream test with mztabexporter for example.

jpfeuffer:
@timosachsenberg merge and check additionally with the workflow in the nightlies?

Issue #1260, opened 2015-02-16
IsobaricAnalyzer: Refactor handling of 0 intensity channels


timosachsenberg:
see also: https://github.com/OpenMS/OpenMS/issues/1258#issuecomment-74483691

hendrikweisser:
There's an unrelated warning produced by IsobaricAnalyzer that could be fixed together with this:
> ConsensusXMLFile::store():  found 1 invalid unique ids

hendrikweisser:
I've fixed the warning in #1269. The actual issue here is still open. Until it gets fixed, a work-around is to filter out zero-intensity features with FileFilter by setting a minimum intensity ("int" parameter).

jpfeuffer:
I agree that would be helpful.

jpfeuffer:
I saw that something even annotates "all_empty". It would be one line, to remove them.

jpfeuffer:
@timosachsenberg Would it make sense to additionally remove a channel completely if its intensity is zero to create "missing values"?

timosachsenberg:
at the level of each consensus feature? Yes, I think these could be removed.
Removing "all_empty" consensus features also sounds like a good idea. I would like to have some second opinion on that.

hendrikweisser:
In my opinion the "all_empty" consensus features should be removed.

Individual channels with zero intensity should stay, because "zero" is the actual readout for those channels.

cbielow:
if all_empty means neither ID nor any kind of quant in any channel, then yes: kick it.
I would vote against single channels which are 0. This information is actually useful.

hendrikweisser:
> I would vote against single channels which are 0.

"Vote against" having them or removing them? (I assume removing.)

cbielow:
> 
> 
> > I would vote against single channels which are 0.
> 
> "Vote against" having them or removing them? (I assume removing.)

fixed my wording.
`I would keep single channels which are 0`

timosachsenberg:
me too

PR #4610, opened 2020-03-27
support QC of labeled experiments (iTRAQ/TMT) 

cbielow:
by just supplying the consensusXML to TOPP `QualityControl`.

This PR also separates some QC classes for a cleaner interface (FWHM, PeptideMass).


jpfeuffer:
Would be a great time to look at https://github.com/OpenMS/OpenMS/issues/1260
and how this interferes with QC (e.g. number of quantified, ided spectra).

cbielow:
g++ 4.8.4 from Travis does not like
```
std::map<UInt64, BaseFeature> new_feats;
for (const FeatureHandle& fh : cf.getFeatures())
{
  UInt64 index = fh.getMapIndex();
  new_feats.emplace(index, fh);   // <- cannot create the pair. There is a ctor from FeatureHandle to BaseFeature available
...
```
and fails to compile, see http://cdash.openms.de/viewBuildError.php?buildid=422916. All other compilers happily do it (as they should).
We could fix this by writing less efficient code, but honestly, this compiler is 6 years old now.
Now, to bring up this conversation again: can we get rid of GCC 4.8.4 at this point?
There is also other parts of OpenMS where we needed to do handstands: https://github.com/OpenMS/OpenMS/pull/4273/files

cbielow:
rebuild jenkins

cbielow:
ready for merge. Fails on Win are Comet-related...

timosachsenberg:
@jpfeuffer do you agree that this is ready for merge or do we need some additional checks on how the input (merged vs. not merged) etc. is formated?

jpfeuffer:
I think it will fail if you merged two TMT consensusXMLs but I don't know how often this happens.

cbielow:
This is not how the workflow is currently handling this. only single itraq/tmt experiments are supported.
 ATM.
How hard would you guesstimate to make marging two consensusXMLs valid? There is probably a lot of traps... ??

timosachsenberg:
I think the proper way would be to merge consensusXML column-wise as we do for the fractionated design (also see the FileMerger option) and adapt other code accordingly if necessary.

timosachsenberg:
-append_method append_cols

PR #4613, opened 2020-03-29
change OpenMS version to 2.6-pre (for nightly installers)

cbielow:
change to v2.6 and appends a `-pre` to all nightly installers which are not in Release or Master branch.


jpfeuffer:
I fixed the test, but maybe it is best if you look for other harcoded versions (I think in pyopenms there might be some)

cbielow:
Ok, I was searching for hardcoded '2.5.0' and only found this one script... updated to 2.6 now

PR #4616, opened 2020-03-30
[FIX] IdXML Allow missed cleavages to be a signed Integer

### LINKED ISSUE #4614: PeptideIndexer crashes with MSGFPlusAdapter output
:\jenkins\ws\openms_release_packaging\9447518b\source\src\openms\source\FORMAT\HANDLERS\XMLHandler.cpp(107): While loading 'D:\Downloads\OF_20181027_LEC_02_HeLa_Cyto_n1.idXML': string representation overflows output binary result( in line 4 column 336)

Maybe related to https://github.com/OpenMS/OpenMS/issues/2666 ?

jpfeuffer:
e.g. for MSGF+ which uses -1 to denote unlimited

fixes #4614

### DIFF
--- a/src/openms/source/FORMAT/IdXMLFile.cpp
+++ b/src/openms/source/FORMAT/IdXMLFile.cpp
@@ -493,7 +493,7 @@ namespace OpenMS
 
       optionalAttributeAsString_(param_.taxonomy, attributes, "taxonomy");
       param_.charges = attributeAsString_(attributes, "charges");
-      optionalAttributeAsUInt_(param_.missed_cleavages, attributes, "missed_cleavages");
+      optionalAttributeAsInt_(param_.missed_cleavages, attributes, "missed_cleavages");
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Does anyone know about any implications in other tools that read this parameter?

jpfeuffer:
Well, officially it is an unsigned int in our schema as well. Another solution would be to catch this when reading MSGF or Myrimatch or whatever kind of input. But then, I don't know if 1000 was the best choice either.

jpfeuffer:
Hmm it was a UInt in the datastructure as well.. might lead to a lot of sign comparison warnings now. Need to check locally.

timosachsenberg:
I think -1 could mainly result in problems converting to other Id formats. Is -1 allowed in mzID? Then I would say we should adopt it

timosachsenberg:
just checked. in mzIdentML it's an xsd:int

jpfeuffer:
@timosachsenberg  Do we still want this or do we go with the workaround to just set it to a UInt whenever it is read?

PR #4618, opened 2020-03-31
File name fix

### LINKED ISSUE #4555: Filenames with multiple . characters in them are truncated in output from TOPPAS
Example: MS2_515.9086_(-13)_Iso2_10AINETD50_OT60_AGC5e4_MaxIT1000.mzML produces output files with the name MS2_515.$EXTENSION. Likewise the layer name is truncated in TOPPView the same way.

poshul:
Fixes  #4555


poshul:
rebuild jenkins

src/openms_gui/source/VISUAL/Spectrum1DWidget.cpp
```diff
@@ -325,13 +325,16 @@ namespace OpenMS
 
   void Spectrum1DWidget::saveAsImage()
   {
-    QString file_name = QFileDialog::getSaveFileName(this, "Save File", "", "Raster images *.bmp *.png *.jpg *.gif (*.bmp *.png *.jpg *.gif);;Vector images *.svg (*.svg)");
+    QString filter = "Raster images *.bmp *.png *.jpg *.gif (*.bmp *.png *.jpg *.gif);;Vector images *.svg (*.svg)";
+    QString* sel_filter = new QString;
```
timosachsenberg, line 329: I would use 
QString sel_filter;
and pass &sel_filter so you don't need the delete later

poshul: done.

hroest:
looks good, why do all the CometAdapter tests fail on Win?

timosachsenberg:
we updated thirdparty and got some failing tests. Chris fixed comet and I think this should now work on windows. He just reported that he get's some failing test on linux now. We look into it.

PR #4619, opened 2020-04-01
fix crash in PeptideIndexer caused by -1 in missed cleavage

timosachsenberg:


timosachsenberg:
1000 is safe as MSGF+ has an upper limit on peptide length

timosachsenberg:
fixes #4619

cbielow:
i did not follow the complete issue, but if PeptideIndexer crashes the underlying bug wont be fixed by this PR, right? should we not fix whatever was causing that?

timosachsenberg:
MSGF+ now writes out missed cleavages = -1 previous version wrote out 1000.
-1 is not supported for this attribute in idXML and Leads to downstream errors

PR #4620, opened 2020-04-02
added minimum and maximum peptide length to comet adapter

timosachsenberg:


src/topp/CometAdapter.cpp
```diff
@@ -159,6 +159,12 @@ class TOPPCometAdapter :
     registerIntOption_("allowed_missed_cleavages", "<num>", 0, "Number of possible cleavage sites missed by the enzyme. It has no effect if enzyme is unspecific cleavage.", false, false);
     setMinInt_("allowed_missed_cleavages", 0);
     setMaxInt_("allowed_missed_cleavages", 5);
+
+    registerIntOption_("min_peptide_length", "<num>", 5, "Minimum peptide length to consider.", false);
+    setMinInt_("min_peptide_length", 5);
+    registerIntOption_("max_peptide_length", "<num>", 63, "Maximum peptide length to consider.", false);
```
jpfeuffer, line 165: should we standardize across SEs or use the default specified by each SE?

timosachsenberg: hmm I don't have a strong opinion on that. for now I think the defaults are fine as they are usually the maximum (and minimum) a search engine can handle

jpfeuffer: I see that MSGF is 6 to 40. hmm

jpfeuffer: In one way we would get good comparability inside OpenMS (e.g. for ConsensusID).
In the other, it would be better to compare against other frameworks/ tools outside OpenMS.

jpfeuffer: let's leave it like this. if you use consensusID, you should make sure on the workflow level, that the settings are the same.

src/topp/CometAdapter.cpp
```diff
@@ -159,6 +159,12 @@ class TOPPCometAdapter :
...
+    setMinInt_("max_peptide_length", 63);
```
jpfeuffer, line 166: setMaxInt?? Is this really the limit though?

timosachsenberg: ups thanks

PR #4623, opened 2020-04-03
[DOC] extended documentation of experimental design

timosachsenberg:


timosachsenberg:
TODO: add code that sample column is not needed in one table design

PR #4626, opened 2020-04-05
[FIX] implement HashGrid.find to avoid "at" with exceptions

### LINKED ISSUE #4625: QTCluster happily working with exceptions
making it impossible to use exception breakpoints.
https://github.com/OpenMS/OpenMS/blob/d17cc251fd0c4068eb253b03c9fb107897771fdc/src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp#L457

jpfeuffer:
fixes #4625

### DIFF
--- a/src/openms/include/OpenMS/COMPARISON/CLUSTERING/HashGrid.h
+++ b/src/openms/include/OpenMS/COMPARISON/CLUSTERING/HashGrid.h
@@ -123,7 +123,7 @@ private:
       }
 
 public:
-      Iterator(Grid & grid) :
+      explicit Iterator(Grid & grid) :
         grid_(grid), grid_it_(grid.end())
       {}
 
@@ -140,7 +140,7 @@ public:
         return *this;
       }
 
-      Iterator operator++(int)
+      const Iterator operator++(int)
       {
         Iterator ret(*this);
         operator++();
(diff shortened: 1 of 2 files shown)


src/openms/include/OpenMS/COMPARISON/CLUSTERING/HashGrid.h
```diff
@@ -405,16 +401,28 @@ namespace OpenMS
 
     /**
      * @warning Currently needed non-const by HierarchicalClustering.
-     */
-    grid_iterator grid_begin() { return cells_.begin(); }
+    */
+    typename Grid::mapped_type & grid_at(const CellIndex & x) { return cells_.at(x); }
```
timosachsenberg, line 405: can this be const (as well as the other grid_finds)

jpfeuffer: this one yes.  other grid_find, no, otherwise you cannot overload since only the return type is different.

timosachsenberg:
looks good

src/openms/include/OpenMS/COMPARISON/CLUSTERING/HashGrid.h
jpfeuffer: ah no, actually none of them.

jpfeuffer: both have colliding overloads

PR #4628, opened 2020-04-06
MapAlignerIdentification: option to use "identity" transformation when data is missing

hendrikweisser:
As requested by @timosachsenberg. When the "force" flag is set, MapAlignerIdentification will now finish by applying the "identity" transformation (i.e. no change to RTs) when there isn't enough suitable RT data to calculate an alignment. Previously (and still without "-force") it would fail with an error.
I think not having the expected data is actually an error condition, hence the requirement to use "-force".

### DIFF
--- a/src/topp/MapAlignerIdentification.cpp
+++ b/src/topp/MapAlignerIdentification.cpp
@@ -177,7 +177,7 @@ private:
         model_type = "identity";
         transformations.resize(data.size());
       }
-      else throw(err);
+      else throw;
     }
 
     if (model_type != "none")
--- a/src/topp/MapAlignerIdentification.cpp
+++ b/src/topp/MapAlignerIdentification.cpp
@@ -175,6 +175,7 @@ private:
           << "\nSince 'force' is set, processing will continue using 'identity' transformations."
           << endl;
         model_type = "identity";
+        transformations.resize(data.size());
       }
       else throw(err);
     }
--- a/src/topp/MapAlignerIdentification.cpp
+++ b/src/topp/MapAlignerIdentification.cpp
@@ -116,7 +116,7 @@ public:
 
 private:
   template <typename MapType, typename FileType>
-  void loadInitialMaps_(vector<MapType>& maps, StringList& ins, 
+  void loadInitialMaps_(vector<MapType>& maps, StringList& ins,
                         FileType& input_file)
   {
     // custom progress logger for this task:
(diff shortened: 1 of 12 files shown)


src/topp/MapAlignerIdentification.cpp
```diff
@@ -158,11 +158,27 @@ class TOPPMapAlignerIdentification :
...
+
+    try
+    {
+      algorithm.align(data, transformations, reference_index);
+    }
+    catch (Exception::MissingInformation& err)
+    {
+      if (getFlag_("force"))
+      {
+        OPENMS_LOG_ERROR
```
timosachsenberg, line 173: in PeptideIndexer we can select how to handle the error situation with some nice string lists.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/ID/PeptideIndexing.cpp#L51
which might be more expressive

src/topp/MapAlignerIdentification.cpp
```diff
@@ -158,11 +158,27 @@ class TOPPMapAlignerIdentification :
...
+          << "Error: alignment failed. Details:\n" << err.getMessage()
+          << "\nSince 'force' is set, processing will continue using 'identity' transformations."
+          << endl;
+        model_type = "identity";
```
timosachsenberg, line 177: didn't know that this works out of the box.
The fit (which is later called) is then just a NOP?

timosachsenberg:
Thanks! This simplifies some of our workflows.
Adding a small test using e.g. two different fractions (test files BSA_F1.idXML and BSA_F3.idXML ) would be great.
https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/examples/FRACTIONS/BSA1_F1.idXML

src/topp/MapAlignerIdentification.cpp
hendrikweisser: Good to know, but I don't want to make this more complex than necessary at the moment.

src/topp/MapAlignerIdentification.cpp
hendrikweisser: Yep. :-)

hendrikweisser:
I tested this with empty idXML files and it seemed to work, but now I tried with the BSA fractions and got a segfault. So unfortunately this isn't ready to merge yet.

src/topp/MapAlignerIdentification.cpp
```diff
@@ -158,11 +158,27 @@ class TOPPMapAlignerIdentification :
...
+      }
+      else throw(err);
```
timosachsenberg, line 179: better rethrow with throw()
https://www.learncpp.com/cpp-tutorial/14-6-rethrowing-exceptions/

hendrikweisser: Thanks for the hint! (I tried `throw()` initially, but that didn't compile...)

hendrikweisser:
Issues should be fixed now. Jenkins gets test failures for "CometAdapter_1" and "..._3" on Windows, but I don't see how that could be related to the changes here.

PR #4630, opened 2020-04-06
[Feature][Fix] consensusXML and mzTab protein group support

jpfeuffer:
"Just" wanted to look at a ProteinQuantifier issue, when I saw that neither cXML nor mzTab output works correctly.

TODO:
- [x] implement correct quant info writing for other mzTab row types. Would be nice to have this for other row types as well but I don't know enough about the metabo, XL, oligo datastructures.
- [x] probably fix some more tests?


src/openms/source/FORMAT/ConsensusXMLFile.cpp
```diff
@@ -88,7 +86,11 @@ namespace OpenMS
     }
     else if (tag == "IdentificationRun")
     {
-      consensus_map_->getProteinIdentifications().push_back(prot_id_);
+      // post processing of ProteinGroups (hack)
```
timosachsenberg, line 89: maybe quickly say what the hack consists of

src/openms/source/FORMAT/ConsensusXMLFile.cpp
```diff
@@ -812,7 +814,14 @@ namespace OpenMS
         os << "\t\t\t</ProteinHit>\n";
       }
 
-      writeUserParam_("UserParam", os, current_prot_id, 3);
+      // add ProteinGroup info to metavalues (hack)
```
timosachsenberg, line 817: same here

src/openms/source/FORMAT/ConsensusXMLFile.cpp
```diff
@@ -812,7 +814,14 @@ namespace OpenMS
...
+      MetaInfoInterface meta = current_prot_id;
+      addProteinGroups_(meta, current_prot_id.getProteinGroups(),
+                        "protein_group", accession_to_id_, STORE);
+      addProteinGroups_(meta, current_prot_id.getIndistinguishableProteins(),
+                        "indistinguishable_proteins", accession_to_id_, STORE);
+      writeUserParam_("UserParam", os, meta, 3);
+      //writeUserParam_("UserParam", os, current_prot_id, 3);
```
timosachsenberg, line 824: remove?

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2825,19 +2867,19 @@ namespace OpenMS
...
-        if (column_entries[i].first == *it)
+      bool found = false;
+      for (Size i = 0; i != column_entries.size(); ++i)
       {
-          output.push_back(column_entries[i].second.toCellString());
-        found = true;
-        break;
+        if (column_entries[i].first == *it)
+        {
+            output.push_back(column_entries[i].second.toCellString());
```
timosachsenberg, line 2876: indentation

src/openms/source/FORMAT/ConsensusXMLFile.cpp
jpfeuffer: haha I dont know. it was copied from idXML

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2866,10 +2908,11 @@ namespace OpenMS
       Size n_best_search_engine_score = mz_tab.getMetaData().protein_search_engine_score.size();
 
       // add header
-      out.push_back(generateMzTabProteinHeader_(protein_section[0], n_best_search_engine_score, mz_tab.getProteinOptionalColumnNames()));
+      // use the last element as a reference row, since they contain group abundance data if set
```
timosachsenberg, line 2911: why use the last element? maybe add another comment as it kind of seems like a hack we needed to introduce

src/topp/ProteinQuantifier.cpp
```diff
@@ -830,9 +831,16 @@ class TOPPProteinQuantifier :
         // annotate quants to protein(groups) for easier export in mzTab
         auto const & protein_quants = quantifier.getProteinResults();
         PeptideAndProteinQuant::annotateQuantificationsToProteins(protein_quants, proteins_, quantifier.getStatistics().n_samples);
-        vector<ProteinIdentification> proteins = consensus.getProteinIdentifications();
-        proteins.insert(proteins.begin(), proteins_); // insert inference information as first protein identification
-        consensus.setProteinIdentifications(proteins);
+        if (!inference_in_cxml)
+        {
+          vector<ProteinIdentification> proteins = consensus.getProteinIdentifications();
```
timosachsenberg, line 836: get by mutable reference does not exist?

timosachsenberg:
looks good but I think some issues remain and some tests would be really good

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -542,6 +542,15 @@ namespace OpenMS
                   items.end());
     }
 
+    /// Move items that satisfy a condition to a container (e.g. vector)
+    template <class Container, class Predicate>
+    static void moveMatchingItems(Container& items, const Predicate& pred, Container& target)
```
timosachsenberg, line 547: maybe consider a move_if  implementation as described in https://stackoverflow.com/questions/32155810/move-all-elements-which-satisfy-some-condition-from-one-container-to-another-i

timosachsenberg: but not that important

src/openms/source/ANALYSIS/ID/IDBoostGraph.cpp
```diff
@@ -1526,16 +1552,16 @@ namespace OpenMS
...
-                {
-                  ProteinHit* ph = boost::get<ProteinHit*>(graph[*ui]);
-                  scores_and_tgt_fraction.emplace_back(
-                      ph->getScore(),
-                      static_cast<double>(ph->getMetaValue("target_decoy").toString()[0] == 't')); // target = 1; false = 0;
-                }
+              }
+              if (!part_of_group)
+              {
+                ProteinHit* ph = boost::get<ProteinHit*>(graph[*ui]);
```
timosachsenberg, line 1558: const?

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
jpfeuffer: I actually implemented the first answer of the link. Since we have to shrink the vector, I used the partition approach.

PR #4631, opened 2020-04-07
[FIX] wrap changed static method in TransformationModelLowess

### LINKED ISSUE #4585: pyopenms: method should be static
This is a declaration for a member method, but underlying C++ method is static:

https://github.com/OpenMS/OpenMS/blob/e26c2ced530423f0c3fcfa077e38d41da1cb12dd/src/pyOpenMS/pxds/TransformationModelLowess.pxd#L16

jpfeuffer:
fixes #4585
I hope this syntax is correct?!


timosachsenberg:
I think you need to do it like in https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/pxds/AASequence.pxd

PR #4634, opened 2020-04-09
[FIX] Do not fill spectrum data and check all spectra for centroided in...

jpfeuffer:


src/topp/CometAdapter.cpp
```diff
@@ -633,11 +634,9 @@ class TOPPCometAdapter :
     }
 
     // determine type of spectral data (profile or centroided)
-    SpectrumSettings::SpectrumType spectrum_type = exp[0].getType();
-
-    if (spectrum_type == SpectrumSettings::PROFILE)
+    for (const auto& s : exp)
     {
-      if (!getFlag_("force"))
+      if (s.getType() == SpectrumSettings::PROFILE && !getFlag_("force"))
```
timosachsenberg, line 639: @jpfeuffer changed by me

jpfeuffer: weird. gitpod did not store my last save correctly I guess.

jpfeuffer: thx

PR #4636, opened 2020-04-09
[FIX] limit double precision to 17 digits during write

### LINKED ISSUE #4627: [Discussion] Unreasonable output precision of our new boost::karma generators?
https://github.com/jpfeuffer/OpenMS/blob/master/src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h#L63-L63

Imho this should be writtenDigits() - log10(n)

We are having problems with other languages parsing our output since the precision is too high.

jpfeuffer:
fixes #4627


src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
```diff
@@ -60,7 +60,9 @@ namespace OpenMS
     class BK_PrecPolicy : public boost::spirit::karma::real_policies<T>
     {
     public:
-      static unsigned int precision(T) { return writtenDigits<T>(T()); }
+        static unsigned int precision(T n) {
```
timosachsenberg, line 63: - bracket in new line
- add comment why this is sensible to do

oliveralka:
@jpfeuffer I probably test if that fixes the issue before merging or?

jpfeuffer:
Sure. If you have workspace ready where the test data is, it should be quick.

I wonder how many tests fail or if FuzzyDiff always catches the mini differences.

timosachsenberg:
in1:  /home/travis/build/OpenMS/OpenMS/_build/src/tests/class_tests/openms/AccurateMassSearchEngine_test_319.tmp   (line: 13, position/column: 14/36)
			<intensity>!
			           36310.1992</intensity>

in2:  /home/travis/build/OpenMS/OpenMS/src/tests/class_tests/openms/data/AccurateMassSearchEngine_output1.featureXML   (line: 13, position/column: 14/36)
			<intensity>!
			           36310.199219</intensity>

jpfeuffer:
Yes, so the problem is, we used our writtenDigits() function before which returns std::numeric_limits<T>::digits10, which is (in my opinion) not what you want if you want to read
back in serialized doubles and guarantee that they will be the same.
I am happy about input here. https://www.boost.org/doc/libs/1_72_0/libs/multiprecision/doc/html/boost_multiprecision/tut/limits/constants.html gives a good overview of those two values.

jpfeuffer:
> 36310.1992

this is enough to represent the underlying 36310.19921875 uniquely when reading it back in.

src/openms/include/OpenMS/DATASTRUCTURES/StringUtils.h
```diff
@@ -60,7 +60,16 @@ namespace OpenMS
...
-      static unsigned int precision(T) { return writtenDigits<T>(T()); }
+        static unsigned int precision(T n)
+        {
+          // max_digits10 specifies the max digits needed to uniquely read floating points
+          // back in. The precision member only controls the fraction part.
+          // -> if fixed point, subtract the number of digits of the integer part
+          // -> if scientific, subtract one
+          return BK_PrecPolicy<T>::floatfield(n) ?
+            std::numeric_limits<T>::max_digits10 - (floor(log10(n)) + 1) : 
+            std::numeric_limits<T>::max_digits10 - 1;
```
timosachsenberg, line 71: why subtract one?
From the boost docu:
The extra two or three least-significant digits are 'noisy' and may be junk, but if you want to 'round-trip' - printing a value out as a decimal digit string and reading it back in - (most commonly during serialization and de-serialization) you must use os.precision(std::numeric_limits<T>::max_digits10)

jpfeuffer: for scientifc notation only.

oliveralka:
Let's see what the others think. 

Tested locally in addition:
```
	 39 - DataValue_test (Failed)
	 62 - String_test (Failed)
	173 - MascotInfile_test (Failed)
	182 - MzMLFile_test (Failed)
	187 - MzTab_test (Failed)
	196 - ParamXMLFile_test (Failed)
	205 - SVOutStream_test (Failed)
	263 - DataFilters_test (Failed)
	388 - AccurateMassSearchEngine_test (Failed)
	442 - ItraqConstants_test (Failed)
	2221 - TOPP_OpenPepXL_1_out_2 (Failed)
	2226 - TOPP_OpenPepXLLF_1_out_2 (Failed)
	2286 - TOPP_FidoAdapter_1 (Failed)
	2287 - TOPP_FidoAdapter_1_out (Failed)
	2288 - TOPP_FidoAdapter_2 (Failed)
	2289 - TOPP_FidoAdapter_2_out (Failed)
	2290 - TOPP_FidoAdapter_3 (Failed)
	2291 - TOPP_FidoAdapter_3_out (Failed)
	2292 - TOPP_FidoAdapter_4 (Failed)
	2293 - TOPP_FidoAdapter_4_out (Failed)
	2294 - TOPP_FidoAdapter_5 (Failed)
	2295 - TOPP_FidoAdapter_5_out (Failed)
	2296 - TOPP_FidoAdapter_6 (Failed)
	2297 - TOPP_FidoAdapter_6_out (Failed)
```

jpfeuffer:
So, me and @timosachsenberg figured that we:
a) need an absolute value of the number first
b) that this formula will lead to longer strings for small values that do not yet undergo scientific writing mode (e.g. 0.001). Here it would ADD 4 to account for the first significant digit starting at position 4. Leading to an output of size 21, making it unreadable by the Java library.

So the solution to b) would be to switch to scientific mode starting from < 0.1

timosachsenberg:
if abs(N) < 0.1

timosachsenberg:
did I get it right that the scientific notation has a fixed number of digits? And that lower number of written digits now works with luciphor?

jpfeuffer:
We set the precision to writtenDigits<T>(), which is digits10<T> which is 15 for double.
Before, we allowed numbers from 0.0001 to 99999 to be written in fixed notation. Since boost karma
does not care about significant digits you still get an additional 15 digits after. 5+15 = 20 = too much.
If you switch to scientific starting at 10000 you get 1e05 plus the fractional part, so in total 16 digits.

PR #4637, opened 2020-04-13
Feat: Auto QC calculations

dmccloskey:
# New features
- automated estimation of default QC filter/flagging values
- ability to filter based off of %RSD values as calculated by e.g., pooled QCs or replicate Samples
- ability to filter based off of %BackgroundInterference values as calculated by e.g., Blank samples
- store method for the MRMFeatureQC class

# Code changes
- removed old code around potential QCML support
- comparison methods for MRFeatureQC


dmccloskey:
The CI failures appear to be due to the lint checks add a test that appears quite unrelated to the PR on windows.  Please let me know how best to proceed.

jpfeuffer:
Correct, the test is unrelated. But could you try to address the linting errors quickly?
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=432435

And you seem to have pulled over changes to the referenced version of the contribs from another branch or so. Can you try to revert that part?
Potentially with
`git submodule --init contrib`

dmccloskey:
Please let me know if something is still up with the submodules.  I am not sure what happened exactly to cause the changes.  I reverted them using `git submodule init contrib`.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -94,13 +95,81 @@ namespace OpenMS
...
+    /**
+      @brief Flags or filters features and subordinates in a FeatureMap based on a user
+        defined set of filter values derived from calling `EstimateBackgroundInterferences`.  The user supplied
+        filter_criteria represents the bounds on acceptable %BackgroundInterference values.
+        NOTE that filtering is only done on the `Intensity` member.
+
+
+      @param features FeatureMap to flag or filter
+      @param filter_criteria MRMFeatureQC class defining QC parameters defining the acceptable limits of %BackgroundInterference
+        where %BackgroundInterference = (value Sample)/(value Blank)*100%
```
cbielow, line 120: could you check if this actually generates the desired documentation? `%` and doxygen gives me the creeps

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -94,13 +95,81 @@ namespace OpenMS
...
+    /**
+      @brief Estimate the lower and upper bound values for the MRMFeatureQC class based on a
+        user supplied template.  The template can either be initialized from the first sample 
+        (meaning all initial template values are over written) or not (meaning the initial template
+        values will be updated if the ranges are found to be too narrow)
+
+      @param[in] samples Samples (typically Standards) from which to estimate the lower and upper bound values for the MRMFeatureQC members
+      @param[in, out] filter_template A MRMFeatureQC class that will be used as a template to fill in the estimated lower and upper values.
+        A "template" is needed so that the MRMFeatureQC::meta_value_qc parameters that the FeatureMap::MetaValues that user would like estimated are known.
+      @param [in] transitions transitions from a TargetedExperiment
```
cbielow, line 135: must be `@param[in]`. The extra space will lead to `[in]` being interpreted as the variable name.
See https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/SYSTEM/File.h#L241for how*NOT* to do it (if you could fix that while being at it... :)).
From Doxygen: `Possible values are "[in]", "[in,out]", and "[out]"`

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -94,13 +95,81 @@ namespace OpenMS
...
+    void FilterFeatureMapBackgroundInterference(FeatureMap& features, const MRMFeatureQC& filter_criteria, const MRMFeatureQC& filter_values);
+
+    /**
+      @brief Estimate the lower and upper bound values for the MRMFeatureQC class based on a
+        user supplied template.  The template can either be initialized from the first sample 
+        (meaning all initial template values are over written) or not (meaning the initial template
+        values will be updated if the ranges are found to be too narrow)
+
+      @param[in] samples Samples (typically Standards) from which to estimate the lower and upper bound values for the MRMFeatureQC members
+      @param[in, out] filter_template A MRMFeatureQC class that will be used as a template to fill in the estimated lower and upper values.
```
cbielow, line 133: remove space before `out`

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -153,6 +222,57 @@ namespace OpenMS
...
+      @brief Updates the metaValue ranges based on the value given
+
+      @param[in] component component of the numerator
+      @param[in] meta_value_key Name of the metaValue
+      @param[in, out] meta_value_l Lower bound (inclusive) for the metaValue range
+      @param[in, out] meta_value_u Upper bound (inclusive) for the metaValue range
+      @param[out] key_exists true if the given key is found, false otherwise
+    */
+    void updateMetaValue(
+      const Feature & component,
```
cbielow, line 235: coding convention: `Feature& component....`

src/openms/source/FORMAT/MRMFeatureQCFile.cpp
```diff
@@ -208,10 +208,98 @@ namespace OpenMS
...
-  //   // TODO: pending fix to CsvFile::fstore()
-  // }
+  void MRMFeatureQCFile::store(const String & filename, const MRMFeatureQC & mrmfqc, const bool is_component_group)
+  {
+    if (is_component_group) {
+      // Store the ComponentGroupQCs
+      clear(); // clear the buffer_
+
+      // Make the ComponentGroupQCs headers
+      String headers = "component_group_name,retention_time_l,retention_time_u,intensity_l,intensity_u,overall_quality_l,overall_quality_u";
```
cbielow, line 218: why create a joined string, only to split it again?
how about
```
StringList headers = {"one", "two" ...};
headers.push_back("meta"+x);
```

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
```diff
@@ -360,12 +645,247 @@ namespace OpenMS
...
+          // iterate through multi-feature/multi-sub-feature QCs/filters
+          // iterate through component_groups
+          for (size_t cg_qc_it = 0; cg_qc_it < filter_template.component_group_qcs.size(); ++cg_qc_it)
+          {
+            if (filter_template.component_group_qcs.at(cg_qc_it).component_group_name == component_group_name)
+            {
+              const double rt = samples.at(sample_it).at(feature_it).getRT();
+              if (sample_it == 0 && init_template_values) {
+                initRange(rt,
+                  filter_template.component_group_qcs.at(cg_qc_it).retention_time_l,
```
cbielow, line 673: all occurences of `filter_template.component_group_qcs.at(cg_qc_it)` could be replace by `somevar` if the loop would be range-based. Much easier to understand and probably faster (if thats a concern).
`for (auto& somevar: filter_template.component_group_qcs) ...`

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
dmccloskey: I changed it just in case.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
dmccloskey: You don't like the default MSVC linting I see?

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFilter.cpp
dmccloskey: The iterator was used because at certain points we were working with two aligned lists.  We have tested the speed using the range based for loop and it did not make a difference at least for the particular logic we were using and the data sets we were running.

timosachsenberg:
@cbielow is there some overlap to your QC approach or something we could consider in the future to harmonize QC approaches?

dmccloskey:
I wanted to check in to see if this branch is ready to be merged.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
```diff
@@ -36,6 +36,7 @@
 
 #include <OpenMS/ANALYSIS/OPENSWATH/MRMFeatureQC.h>
 #include <OpenMS/ANALYSIS/TARGETED/TargetedExperiment.h>
+#include <OpenMS/ANALYSIS/QUANTITATION/AbsoluteQuantitationMethod.h>
```
cbielow, line 39: you should be able to forward-declare `AbsoluteQuantitationMethod` and avoid the include

cbielow:
> 
> 
> @cbielow is there some overlap to your QC approach or something we could consider in the future to harmonize QC approaches?

if this is for MRM data, there is probably not a lot of overlap. Nevertheless, computation of QC might be extracted into a separate class and put into /src/QC/, just for clearer interfaces??

cbielow:
looks good.
Maybe have a look at the codefactor warnings. Some of them seem easy to fix.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
dmccloskey: I am not sure I see how forward-declaration would be desirable here nor the benefit of avoiding the include.  Can you elaborate?

cbielow: its all about compile time.
Unrelated code will not get pulled in compile units which include MRMFeatureFilter.h and speed up compilation.

dmccloskey:
Please advice on the appleclang failure.

jpfeuffer:
Unrelated. Wanna merge?

dmccloskey:
Yes

Issue #4632, opened 2020-04-07
TOPPBase does not check if files in InputFileList exist


timosachsenberg:
see
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/APPLICATIONS/TOPPBase.cpp#L1373

it passes an INPUTFILE_LIST but the code only handles INPUT_FILE

timosachsenberg:
@cbielow you probably know that code better than us. Can you have a quick look at those lines? If you agree that this is the case I can provide a simple fix.

cbielow:
yep. looks buggy.

PR #4643, opened 2020-04-14
[FIX] test presence of files in input file lists

timosachsenberg:


timosachsenberg:
Exit Value | 1
-- | --
1

jpfeuffer:
does TOPPBase use our File.find functions? Cause only those respect searching in share and OpenMS home.

timosachsenberg:
File:exists

timosachsenberg:
https://github.com/OpenMS/OpenMS/blob/develop/src/utils/AccurateMassSearch.cpp#L107

timosachsenberg:
fixes https://github.com/OpenMS/OpenMS/issues/4632

PR #4644, opened 2020-04-15
[FIX] add fall-back if alignment not possible

timosachsenberg:
add recent fix by Hendrik


timosachsenberg:
I think if we analyze a lot of files and one is bad, it would make sense to have the identity transformation for that single file. But this is currently not supported I think - fall-back is always to identity for all transformations, right @hendrikweisser ?

timosachsenberg:
I added the force parameter for now

PR #4646, opened 2020-04-15
[NOP] added additional checks to prevent wrong order of input files

timosachsenberg:


cbielow:
can the code be pulled into the library? The main method seems pretty large...

timosachsenberg:
For just checking Input basename - yes. The code that involves reading the full Id files to access the annotated ms run I don’t think so.

timosachsenberg:
Where would you put it? TOPPBase would be an option.

cbielow:
Toppbase is already a beast. 
In File.cpp or a new class?

cbielow:
very nice!
+1 for merge if tests pass

PR #4649, opened 2020-04-16
[FIX] disable frequent warning message in SN estimator (as done in Op…

timosachsenberg:
…enSWATHWorkflow by default).


timosachsenberg:
tests passed and it is a trivial PR that just reduced log output to be consistent with OpenSWATH workflow so I merge

PR #4651, opened 2020-04-17
[FIX] pyopenms-nightly

### LINKED ISSUE #4650: pyopenms-nightly builds broken
Conda installation on win and mac not working anymore.

Also we have to fix the overnight day-change bug for pip.

And unify the contribs installation: on mac it still uses a cached version but probably should download as well.

oliveralka:
Fix #4650
1) Download contribs 
2) Try to set timezone to fix overnight day-change bug


jpfeuffer:
Looks good, but did you figure out, why miniconda is not working anymore. Otherwise it will be hard to debug your other changes.

oliveralka:
Haha yeah, I will try it directly within setup.py.

In case of miniconda it seems that the corresponding download link has changed to 
`https://repo.anaconda.com/miniconda`

jpfeuffer:
Looks v promising. thx

jpfeuffer:
No module pytz.

And we need to reset the install names of the contribs libs. I forgot that. On Jenkins we do that automatically. I can add that.

oliveralka:
Added to the .yml. 

Not sure if needed here as well:
```
tools/travis/lnx-cibuild.before.sh
src/pyOpenMS/dist-scripts/create-manylinux.sh
```

oliveralka:
> And we need to reset the install names of the contribs libs. I forgot that. On Jenkins we do that automatically. I can add that.

The would be great - thanks!

oliveralka:
@jpfeuffer Could you take another look, should be ok like that. Then I will remove the debug branch - and it should be ready for merge.

jpfeuffer:
I mean looks good but now we have to change all the build scripts and build machines. I did not think about that :( I am wondering if a CMake based configure solution would be better. I think the string TIMESTAMP function has timezone support.

oliveralka:
Ok, that was not clear to me as well - I will have a look at a CMAKE based solution.

oliveralka:
It might be not possible to set a different timezone it is either current or UTC...let's see! 
https://cmake.org/cmake/help/v3.0/command/string.html

jpfeuffer:
Hmm I see. Another option would be to use the Git last change date that we parse.
GIT_LC_DATE I think in CMake. This might also be more robust, since all the builds will have the same stamp then. But it wont account for rebuilds.

Issue #4359, opened 2019-11-15
FeatureFinderCentroided is too slow


cbielow:
using a single thread takes over 4h on a rather simple dataset.
There is definitely room for performance improvements there.

```
05:54:39 NOTICE: Output file '...\iPRG2015\JD_06232014_sample4_C.idXML' written.
Found 240664 seeds for charge 3.
Found 42727 feature candidates for charge 2.
Found 223719 seeds for charge 3.
Found 15634 feature candidates for charge 3.
Found 48480 seeds for charge 4.
Found 1302 feature candidates for charge 4.
Removed 43561 overlapping features.
65424 features left.

Abort reasons during feature construction:
- Could not extend seed: 583362
- Could not find good enough isotope pattern containing the seed: 131674
- Feature quality too low after fit: 26119
- Invalid feature after fit - too few traces or peaks left: 731121
- Invalid fit: Center outside of feature bounds: 5568
- Invalid fit: Fitted model is bigger than 'max_rt_span': 30451
- Invalid fit: Less than 'min_rt_span' left after fit: 34
FeatureFinderCentroided took 04:36:41 h (wall), 04:36:25 h (CPU), 16.08 s (system), 04:36:09 h (user).
```

timosachsenberg:
Ok stupid question but was that run on the centroided data?

cbielow:
yes (you almost scared me for a sec there, since I uses PPHR in auto-mode and thought it maybe did not detect the data correctly -- but it did).
Inspecting the input, the problem might be that there are a lot of 0 intensity peaks shouldering real peaks. When removing 0-intensity peaks before FF, the runtime drops to 39:16 m (wall). This is still quite long, but at 4x better than before... its a start...

hroest:
@cbielow maybe we can remove zeros by default when (i) peak-picking and (ii) before starting the FFC

timosachsenberg:
Note: This can be done by calling ThresholdMower without parameter

mafreitas:
FWIW.  As @timosachsenberg mentions using ThresholdMower before FFC greatly speeds up the processing.  It would be nice if there were an option in FFC to set a threshold (relative or absolute).

For example, a 3 Gb mzML with ~6000 MS1 and total ~200,000 scans took only 8 minutes to run on an m5.large ec2 instance using 2 CPUs and 8GB ram

[...]

PR #4652, opened 2020-04-17
[FEATURE] FeatureFinderCentroided Performance

IceFreez3r:
As mentioned here: #4359, the FeatureFinderCentroided is too slow. 
@cbielow gave me the original file from the issue. 
Times before:
`FeatureFinderCentroided took 02:06:33 h (wall), 02:06:26 h (CPU), 6.18 s (system), 02:06:20 h (user).`
Then I've implemented the idea of using ThresholdMover from the issue and:
Times afterwards:
`FeatureFinderCentroided took 31:11 m (wall), 31:07 m (CPU), 4.42 s (system), 31:03 m (user).`


cbielow:
while looking at something completely unrelated I remembered that there is an even faster solution:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/FORMAT/OPTIONS/PeakFileOptions.h#L106
Set an Intensity range while loading the mzML. Apply somewhere here:
https://github.com/OpenMS/OpenMS/pull/4652/files#diff-a5a737054268e8beb7abdc43a9a61c43R178

timosachsenberg:
Is there a difference? Maybe the threshold filter also takes care that associated float/string/int dataarrays are filtered

jpfeuffer:
One does not even load the 0 intensity data, so I agree there is an advantage.

timosachsenberg:
Agree but we should check if the data arrays are also correctly filtered e.g. ion mobility

timosachsenberg:
Though for FFC it currently does not use them

cbielow:
metadata is handled correctly AFAIKT.
This would also be a different issue -- this is just using the functionality offered by MzMLFile....

timosachsenberg:
ok so the maximum of the range is: PositionType::maxPositive()
and the minimum std::numeric_limits<PositionType>::epsilon()
I guess

src/topp/FeatureFinderCentroided.cpp
```diff
@@ -178,6 +180,9 @@ class TOPPFeatureFinderCentroided :
     PeakFileOptions options;
     options.setMSLevels(vector<Int>(1, 1));
 
+    using RP_TYPE = DRange<1>::PositionType;
+    options.setIntensityRange({std::numeric_limits<RP_TYPE>::epsilon(), RP_TYPE::maxPositive()});
```
cbielow, line 184: should be `std::numeric_limits<RP_TYPE>::min` (which is 1.175494351e-38F) not `epsilon` (which is 1.192092896e-07F)

cbielow: note: min is positive. lowest would be -max.

timosachsenberg: yep right will chance

src/topp/FeatureFinderCentroided.cpp
```diff
@@ -42,6 +42,9 @@
 #include <OpenMS/FORMAT/MzQuantMLFile.h>
 #include <OpenMS/METADATA/MSQuantifications.h>
 #include <OpenMS/SYSTEM/File.h>
+#include <OpenMS/FILTERING/TRANSFORMERS/ThresholdMower.h>
```
cbielow, line 45: this could be removed

timosachsenberg: k will do in another PR - thanks

IceFreez3r: Changed that on my fork. I don't know if I need to open another PR or if you can just reopen this one

timosachsenberg: Thanks but I already pushed it to develop. So I guess its fine

Issue #4348, opened 2019-11-11
Comet does not identify any spectra if mzML is not indexed


timosachsenberg:

jpfeuffer:
you were suggesting this to check for an index:
OnDiscMSExperiment().openFile("filename", skipMetaData=true)
correct?

I am not sure if we should be adding an index by default though, this would basically copy the whole mzML file without the users knowledge right?

timosachsenberg:
Yes, I think this should return false if there is no index.
From a usability point of view, creating the index would be better (maybe while outputting a warning)

hroest:
Maybe only copy it if no index is found.

jpfeuffer:
Okay. But it should be temporary right? Otherwise we need another output param that might be used in some cases only.

And then we warn the user that they should better index to potentially save time for following steps and in case of reexecution?

timosachsenberg:
Yes I think this would make the workflow more robust

hroest:
> And then we warn the user that they should better index to potentially save time for following steps and in case of reexecution?

where are non-indexed files even coming from? pwiz writes an index by default and OpenMS too since a few releases.

jpfeuffer:
Ah it's from our old BSA example in the share. I agree that they should be rare these days.

jpfeuffer:
This should also have been fixed in the newest version. See Comet changelog.
We should update.

timosachsenberg:
problem persists in newest comet 2019.01 rev. 5

PR #4653, opened 2020-04-17
fixes for CometAdapter

cbielow:
[FIX] fixes indexedMZML problem (#4348, always write an Index)
[FEATURE] CometAdapter does now read the comet version. No need for the user to specifiy it via commandline.


jpfeuffer:
Nice. But what kind of mzml is this writing. Afaik We only read the metadata to double check centroidedness.

jpfeuffer:
And do you really want to copy gigabytes of spectra hidden from the user?

cbielow:
> And do you really want to copy gigabytes of spectra hidden from the user?

With a warning (added), yes. Its 'only' MS2 spectra.

jpfeuffer:
Ok looks good. Warning should be enough I guess.

jpfeuffer:
Quick question before merge: what if you want to search MS3?

cbielow:
good point!

jpfeuffer:
This broke the Address Sanitizer. Please have a look.
Is this really how you initialize a smart pointer? Wouldn't you rather use make_unique/shared?

jpfeuffer:
Ah it needs to be unique_ptr<char[]>

cbielow:
ahh darn.. wrong delete operator...
beginner trap :)
and make_unique is C++14 ....
However, I hear we are switching to 17 soon :)

PR #4663, opened 2020-04-23
Percolator: change comment when using psm or peptide fdr

Leon-Bichmann:
Here comes the PR, as discussed previously.

We can also change it in another way if you have a better idea.

### DIFF
--- a/src/topp/PercolatorAdapter.cpp
+++ b/src/topp/PercolatorAdapter.cpp
@@ -1215,13 +1215,13 @@ protected:
 
       if (!peptide_level_fdrs)
       {
-      OPENMS_LOG_INFO << "PSM-level FDR: The all PSMs per Peptide should be returned by percolator. Reannotating the all PSMs in input data with percolator output." << endl;
+      OPENMS_LOG_INFO << "PSM-level FDR: All PSMs are returned by percolator. Reannotating all PSMs in input data with percolator output." << endl;
       }
       else
       {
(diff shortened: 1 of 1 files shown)


src/topp/PercolatorAdapter.cpp
```diff
@@ -1212,7 +1212,15 @@ class PercolatorAdapter :
           }
         }
       }
+
+      if (!peptide_level_fdrs)
+      {
+      OPENMS_LOG_INFO << "Suitable PSM Hits for " << cnt << " of " << all_peptide_ids.size() <<  " PSMs found." << endl;
+      }
+      else if (peptide_level_fdrs)
```
timosachsenberg, line 1220: remove redundant if here

timosachsenberg:
hmm I don't like the log message. 
I would not understand what it means.

timosachsenberg:
can you improve the log message? Maybe tell the user what input percolator will get?

src/topp/PercolatorAdapter.cpp
```diff
@@ -1212,7 +1212,17 @@ class PercolatorAdapter :
           }
         }
       }
-      OPENMS_LOG_INFO << "Suitable PeptideHits for " << cnt << " of " << all_peptide_ids.size() <<  " PSMs found." << endl;
+
+      if (!peptide_level_fdrs)
+      {
+      OPENMS_LOG_INFO << "PSM-level FDR: The all PSMs per Peptide should be returned by percolator. Reannotating the all PSMs in input data with percolator output." << endl;
```
timosachsenberg, line 1218: the all? the all?

PR #4666, opened 2020-04-24
[FIX] pyopenms-nightly - variable for push 

oliveralka:
PyPI push not working due to unresolved variable ("OPENMS_GIT_LC_DATE").

Uploading pyopenms_nightly-2.6.0.devOPENMS_GIT_LC_DATE-cp35-cp35m-manylinux2014_x86_64.whl

### DIFF
--- a/src/pyOpenMS/CMakeLists.txt
+++ b/src/pyOpenMS/CMakeLists.txt
@@ -402,7 +402,7 @@ if(NOT PY_NUM_MODULES)
 endif()
 
 # set data variable nightly builds use repository last change date
-set(OPENMS_GIT_LC_DATE OPENMS_GIT_LC_DATE)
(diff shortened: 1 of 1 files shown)


cbielow:
can someone explain why this self assignment does anything useful?

oliveralka:
No - I thought it has to be there to be set in env.py, but probably not.

jpfeuffer:
It was not enough. The GIT_LC_DATE seems to include time and timezone, which is probably not allowed in a wheel name. Also I would remove the underscores.

So that it just reads pyopenms_nightly-2.6.0.dev20200424

```
ERROR: pyopenms_nightly-2.6.0.dev2020_04_24.13_34_37._0200-cp38-cp38-manylinux2014_x86_64.whl is not a supported wheel on this platform.
```

jpfeuffer:
@oliveralka

PR #4673, opened 2020-04-26
pyopenms-nightly - change date format

oliveralka:
@jpfeuffer is there an easy way to test the CMake syntax locally?


jpfeuffer:
In theory you just have to configure the project with -DPYOPENMS=ON after configuration you will see in the setup file if it is correct.

jpfeuffer:
But looks good

oliveralka:
Representation of the initial format looked different:

yyyy-mm-dd 08:04:17 +0200

Corrected - now it should work as expected in the nightlies.

jpfeuffer:
Great. I assume you could test it locally now?

oliveralka:
Yes, tested it locally - thanks for the hint ;)

PR #4677, opened 2020-04-27
Sum formula output in RNAMassCalculator

hendrikweisser:
OpenMS has useful functionality in the library for calculating molecular sum formulas of RNAs. This change exposes this functionality to the user via additional output in RNAMassCalculator.


jpfeuffer:
Only style I think. You can ignore macOS for now.

hendrikweisser:
Fixed now.

Issue #4655, opened 2020-04-18
NovorAdapter: maximum number of MGF spectra exceeded


giangiacomoberetta1:
Hi !

While running a search using NovorAdapter, I got the following error message.

The input raw file was converted in centroid mode, so it seems that the problem I related to the  number of spectra that can be written as MGF has exceeded the maximum set value (10.000).

Is it possible to confirm this? In affirmative case, is it possible to increase this upper limit?

Thanks in advance and take care,

GB

ERROR NovorAdapter         0:9        Failing process stdout: [Executable is: D:/KNIME_project/OpenMS-2.4.0/NovorAdapter/novor/lib/novor.jar, Progress of 'loading spectra list':, ,   2.06 %               ,   4.94 %               ,   7.59 %               ,   10.29 %               ,   14.44 %               ,   23.71 %               ,   33.06 %               ,   39.75 %               ,   45.83 %               ,   52.13 %               ,   57.08 %               ,   61.56 %               ,   64.92 %               ,   68.18 %               ,   71.33 %               ,   77.15 %               ,   80.60 %               ,   84.10 %               ,   87.33 %               ,   90.46 %               ,   93.74 %               ,   96.05 %               ,   98.78 %               , -- done [took 10.39 s (CPU), 23.28 s (Wall)] -- , Progress of 'loading chromatogram list':, , -- done [took 0.00 s (CPU), 0.01 s (Wall)] -- , Progress of 'storing mascot generic file':, ,   8.59 %               ,   21.59 %               ,   31.35 %               ,   37.80 %               ,   43.84 %               ,   49.33 %               ,   54.40 %               Error: Unexpected internal error (Spectrum to be written as MGF has 10142 peaks; the upper limit is 10,000. Only centroided data is allowed - this is most likely profile data.)]
ERROR NovorAdapter         0:9        Failing process stderr: []
ERROR NovorAdapter         0:9        Return code: 8
ERROR NovorAdapter         0:9        Execute failed: Failed to execute node NovorAdapter

jpfeuffer:
Hi!

Hmm never realized this limitation. Seems to be related to our MGF writer, so I would assume that we might be able to increase it. Unfortunately it is a hard-coded limit: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/MascotGenericFile.cpp#L319

So it needs to be changed in the code.
You could try to remove 0 or low intensity peaks with ThresholdMower.

[...]

PR #4682, opened 2020-04-28
[FEATURE ]First try on memory efficient streaming/checking for mzmls in SE adapter

jpfeuffer:


src/openms/source/FORMAT/MascotGenericFile.cpp
```diff
@@ -318,10 +318,8 @@ namespace OpenMS
     }
     if (spec.size() >= 10000)
     {
-      String msg = "Spectrum to be written as MGF has " + String(spec.size()) +
-        " peaks; the upper limit is 10,000. Only centroided data is allowed - this is most likely profile data.";
-      throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION,
-                                       msg);
+      OPENMS_LOG_WARN << "Spectrum to be written as MGF has " + String(spec.size()) +
```
cbielow, line 321: MGF file format actually mandates less than 10k peaks,
see http://www.matrixscience.com/help/data_file_help.html -->
```
There is an upper limit of 10,000 peaks per individual MS/MS spectrum.
If you see an error message reporting that this limit has been exceeded,
 it almost certainly means that your data are profile data, and not peak lists.
 It is very unlikely that a single MS/MS spectrum could ever contain more than 
1000 genuine peaks, never mind 10,000.```

File: src/openms/source/FORMAT/MzMLFile.cpp [hunk 2]
```diff
@@ -252,4 +253,32 @@ namespace OpenMS
     consumer->setExperimentalSettings(experimental_settings);
   }
 
+  std::map<UInt,std::pair<Size,Size>> MzMLFile::getCentroidInfo(const String& filename)
+  {
+    bool oldoption = options_.getFillData();
+    options_.setFillData(false);
+    MSDataTransformingConsumer* c = new MSDataTransformingConsumer();
```
cbielow, line 260: Why create on heap?

src/topp/CometAdapter.cpp
```diff
@@ -627,42 +628,38 @@ class TOPPCometAdapter :
...
 
-    PeakMap exp;
-    MzMLFile mzml_file;
-    mzml_file.getOptions().setFillData(false); // only load metadata for spectra
-    mzml_file.getOptions().setMSLevels({2, 3}); // only load MS2 and MS3
-    mzml_file.setLogType(log_type_);
-    mzml_file.load(inputfile_name, exp);
+    int ms_level = getIntOption_("ms_level");
+    const auto& centroid_info = MzMLFile().getCentroidInfo(inputfile_name);
+    std::cout << centroid_info.size() << std::endl;
```
cbielow, line 633: remove cout message
?!

src/openms/source/FORMAT/MascotGenericFile.cpp
jpfeuffer: see https://github.com/OpenMS/OpenMS/issues/4655

jpfeuffer: I can revert it, but IMHO this is just plain arbitrary.

cbielow: but its their format...

and 10k peaks in an MS2 spectrum... this has to be wrong...

jpfeuffer:
http://cdash.openms.de/testDetails.php?test=91827267&build=443403

False positive? Can't see anything wrong. Codacy also has cppcheck and does not complain.

PR #4688, opened 2020-04-30
[Fix] mzTab single proteins

jpfeuffer:
Somehow they were previously just ignored.


jpfeuffer:
Since I changed some indents my editor auto-formatted (e.g. split) some lines.

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1877,20 +1881,20 @@ namespace OpenMS
         if (!ms_run_in_data.empty())
         {
           // prepend file:// if not there yet
-          for (const String& s : ms_run_in_data)
+          for (const String &s : ms_run_in_data)
```
timosachsenberg, line 1884: do we prefer String &s over String& s ?

jpfeuffer: ah good catch. have to change my editor settings

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -2179,15 +2128,68 @@ namespace OpenMS
...
+        // map general protein groups to the contained proteins
         // We only report quantitative data for indistinguishable groups (which may be composed of single proteins).
         // We skip the more extensive reporting of general groups with complex shared peptide relations.
-        std::vector<ProteinIdentification::ProteinGroup> protein_groups2;
+        std::vector<ProteinIdentification::ProteinGroup> empty_protein_groups;
+        auto protein_groups = std::cref(empty_protein_groups);
         if (quant_study_variables == 0)
         {
-          protein_groups2 = it->getProteinGroups();
+          protein_groups = std::cref(it->getProteinGroups());
```
timosachsenberg, line 2168: is there a difference to const std::vector<ProteinIdentification::ProteinGroup>& protein_groups2 = quant_study_variables_ == 0 ? pid.getProteinGroups() : std::vector<ProteinIdentification::ProteinGroup>();
?

timosachsenberg:
integrated into streaming PR

PR #4694, opened 2020-05-03
Feature/mz tab stream

timosachsenberg:
Export from consensusMap and IDs possible now.
There is currently some code duplication between two stream writer.  Moving stuff to a common base would solve that problem but I thought it might be better to wait until a stream interface for the new Identification data structure is implemented.  This PR fixes the memory consumption problem that was the primary aim.


timosachsenberg:
@jpfeuffer maybe you can add the test files to this PR and close https://github.com/OpenMS/OpenMS/pull/4688

Currently marked as draft because some tests fail but stream export from consensusMap is nearly complete

timosachsenberg:
rebuild jenkins

timosachsenberg:
@jpfeuffer tests passed. one upload timeout

src/openms/include/OpenMS/FORMAT/MzTabFile.h
```diff
@@ -68,6 +68,25 @@ namespace OpenMS
     // store MzTab file
     void store(const String& filename, const MzTab& mz_tab) const;
 
+    // stream IDs to file
+    void store(
```
timosachsenberg, line 72: would be great if our ID and feature datastructures would also provide a stream reader interface ...

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1412,78 +1448,85 @@ namespace OpenMS
...
-        else if (peptide_evidences[i].getAABefore() == PeptideEvidence::N_TERMINAL_AA)
-        {
-          row.pre = MzTabString("-");
-        }
-        else
-        {
-          row.pre = MzTabString(String(peptide_evidences[i].getAABefore()));
-        }
+      // report without pep evidence information
+      row.pre = MzTabString("null");
```
jpfeuffer, line 1457: doesn't an empty MzTabString default to "null" when written?
Saves a string comparison when initialized.

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1634,769 +1677,858 @@ namespace OpenMS
...
-      f.getKeys(keys); //TODO: why not just return it?
-      for (String & s : keys)
-      {
-        if (s.has(' '))
-        {
-          s.substitute(' ', '_');
-        }
-      }
+      auto row = peptideSectionRowFromFeature_(f, feature_user_value_keys, peptide_hit_user_value_keys, fixed_mods);
+      mztab.getPeptideSectionRows().emplace_back(row);
```
jpfeuffer, line 1688: move row? At least for the vectors and maps inside it should work. Move for MzTabString etc might also be nice.

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1634,769 +1677,858 @@ namespace OpenMS
...
-      row.search_engine_score_ms_run[1][1] = MzTabDouble();
-
-      // create opt_ column for peptide sequence containing modification
-      MzTabOptionalColumnEntry opt_global_modified_sequence;
-      opt_global_modified_sequence.first = String("opt_global_modified_sequence");
-      row.opt_.push_back(opt_global_modified_sequence);
-
-      // create and fill opt_ columns for feature (peptide) user values
-      addMetaInfoToOptionalColumns(feature_user_value_keys, row.opt_, String("global"), f);
+      window.emplace_back(MzTabDouble(f.getConvexHull().getBoundingBox().minX()));
```
jpfeuffer, line 1712: do you need MzTabDouble when emplacing?

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1634,769 +1677,858 @@ namespace OpenMS
...
-        }
-      }
+    const set<String>& accessions = best_ph.extractProteinAccessionsSet();
+    const vector<PeptideEvidence>& peptide_evidences = best_ph.getPeptideEvidences();
 
-      // create and fill opt_ columns for psm (PeptideHit) user values
-      addMetaInfoToOptionalColumns(peptide_hit_user_value_keys, row.opt_, String("global"), best_ph);
+    row.unique = accessions.size() == 1 ? MzTabBoolean(true) : MzTabBoolean(false);
+    // select accession of first peptide_evidence as representative ("leading") accession
+    row.accession = peptide_evidences.empty() ? MzTabString("null") : MzTabString(peptide_evidences[0].getProteinAccession());
```
jpfeuffer, line 1762: maybe MzTabString()?

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1634,769 +1677,858 @@ namespace OpenMS
...
+    set<String> pid_key_set(pid_keys.begin(), pid_keys.end());
+    addMetaInfoToOptionalColumns(pid_key_set, row.opt_, String("global"), pid);
 
-      // map indistinguishable groups to the contained proteins
-      const std::vector<ProteinIdentification::ProteinGroup>& indist_groups = prot_ids.front().getIndistinguishableProteins();
-      Size ind_idx{0};
-      for (const ProteinIdentification::ProteinGroup & p : indist_groups)
+    // link to spectrum in MS run
+    String spectrum_nativeID = pid.getMetaValue("spectrum_reference").toString();
+    size_t run_index = idrun_2_run_index[pid.getIdentifier()];
```
jpfeuffer, line 2064: I know it was probably already in the old code but maybe (especially since it is in a separate function now) we should use .at() in this part to not get any bad surprises.
Probably not very speed-critical here.

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1634,769 +1677,858 @@ namespace OpenMS
...
+    vector<MzTabString> entries;
+    for (Size j = 0; j != group.accessions.size() ; ++j)
+    {
+      // set accession and description to first element of group
+      if (j == 0)
+      {
+        protein_row.accession = MzTabString(group.accessions[j]);
+        // protein_row.description  // TODO: how to set description? information not contained in group
+      }
+      entries.emplace_back(MzTabString(group.accessions[j]));
```
jpfeuffer, line 2370: constructor needed?

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -1634,769 +1677,858 @@ namespace OpenMS
...
+    // set accession and description to first element of group
+    protein_row.accession = MzTabString(leader_protein.getAccession());
 
-          // set accession and description to first element of group
-          protein_row.accession = MzTabString(leader_protein.getAccession());
+    // TODO: check with standard if it is important to also place leader at first position
+    //       (because order in set and vector may differ)
+    for (Size j = 0; j != group.accessions.size() ; ++j)
+    {
+      entries.emplace_back(MzTabString(group.accessions[j]));
```
jpfeuffer, line 2419: superfluous constructor?

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2888,6 +2887,216 @@ namespace OpenMS
...
+          Size n_best_search_engine_score = row.best_search_engine_score.size();
+          tab_file << "\n" << generateMzTabPeptideHeader_(search_ms_runs, n_best_search_engine_score, n_search_engine_score, assays, study_variables, s.getPeptideOptionalColumnNames()) + "\n";
+          first = false;
+/*
+ *        TODO: consider removing this below as well. Does not add much value and looks convoluted
+          if (complete)
+          { // all ms_runs mandatory
+            search_ms_runs = ms_runs;
+          }
+          else // only report all scores if user provided at least one
```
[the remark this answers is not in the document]
timosachsenberg: yes, I think this is only relevant if exporting from a user constructed mzTab data structure.

src/openms/source/FORMAT/MzTab.cpp
timosachsenberg: I think so because the constructor is explicit: explicit MzTabDouble(const double v);

src/openms/source/FORMAT/MzTab.cpp
timosachsenberg: same

src/openms/source/FORMAT/MzTab.cpp
timosachsenberg: jep explicit

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2914,12 +3123,13 @@ namespace OpenMS
       // add header
       // use the first row as a reference row to get the optional cols from meta values
       // TODO this is very fragile!
+      out.push_back("");
```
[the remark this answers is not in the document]
timosachsenberg: newline

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -587,28 +588,51 @@ namespace OpenMS
     mz_tab_version.fromCellString(String("1.0.0"));
   }
 
-	// static method remapping the target/decoy column from an opt_ to a standardized column
-	static void remapTargetDecoy_(std::vector<MzTabOptionalColumnEntry>& opt_entries)
+  // static method remapping the target/decoy column from an opt_ to a standardized column
+  static void remapTargetDecoyPSMAndPeptideSection_(std::vector<MzTabOptionalColumnEntry>& opt_entries)
```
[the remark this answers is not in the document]
timosachsenberg: names are generated form the meta data stored in the hits so I think this is possible.

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2914,12 +3123,13 @@ namespace OpenMS
       // add header
       // use the first row as a reference row to get the optional cols from meta values
       // TODO this is very fragile!
```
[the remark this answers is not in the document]
timosachsenberg: no unfortunately not... with the MzTab object users could still mess around

PR #4695, opened 2020-05-03
[Feature] refactor PeakPickers

### LINKED ISSUE #4686: [REVIEW] Consolidate PeakPickers
@OpenMS/core We have PPHiRes (with low_memory mode), PPHiResLoMem, PPHiResLoMemRA

which basically all do the same thing.

Can we remove the latter two?

jpfeuffer:
fixes #4686


timosachsenberg:
can we recycle the failing tests?

jpfeuffer:
There are already two tests for the lowmem option. And it should just yield the same results as the default option.

jpfeuffer:
tests should be unrelated but I can rerun windows later.

PR #4700, opened 2020-05-05
fix: struct vs class

dmccloskey:
Clash on Windows (and only Win) with forward declared struct for a later declared class when building an external application that calls OpenMS.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMFeatureFilter.h
@@ -49,7 +49,7 @@
 
 namespace OpenMS
 {
-  struct AbsoluteQuantitationMethod;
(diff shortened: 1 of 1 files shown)


timosachsenberg:
seems some submodules were not updated on your side. can you check?

dmccloskey:
Do you have any recommendations for reverting the submodule changes?  My attempts as resetting the submodules with the usual `git submodule update --init` appear not yield any tracked changes.

timosachsenberg:
when I am lost I usually go into the submodule folder. Check where it points to with git branch -va .
Then you can do a manual change to the correct commit with: git reset --hard COMMITHASH

timosachsenberg:
I will check

timosachsenberg:
I did a "git checkout master" in both submodules and then a git add and commit with the updated submodule references

dmccloskey:
That looks to have made quite a difference.  There are still 3 changes in THIRDPARTY that do not match.

PR #4704, opened 2020-05-06
MapAlignerTreeGuided: Better memory usage and new option

jthueringer:
Because the alignment of large data sets can lead to an out of memory problem, I have changed the following:
1. internally the growing FeatureMap is no longer copied. Instead, all maps that are no longer needed are deleted from the temporary vector.
2. the user of the TOPP tool is given the option to choose whether the input data is copied or loaded twice during the alignment.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
+++ b/src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
@@ -92,12 +92,12 @@ public:
      * @brief Align feature maps tree guided using align() of @ref OpenMS::MapAlignmentAlgorithmIdentification and use TreeNode with larger 10/90 percentile range as reference.
      *
      * @param tree Vector of BinaryTreeNodes that contains order for alignment.
-     * @param feature_maps_transformed Vector with copied input maps for transformation process.
+     * @param feature_maps_transformed Vector with input maps for transformation process.
      * @param maps_ranges Vector that contains all sorted RTs of extracted identifications for each map; needed to determine the 10/90 percentiles.
      * @param map_transformed FeatureMap to store all features of combined maps with original and transformed RTs in order of alignment.
      * @param trafo_order Vector to store indices of maps in order of alignment.
     */
-    void treeGuidedAlignment(const std::vector<BinaryTreeNode>& tree, std::vector<FeatureMap> feature_maps_transformed,
+    void treeGuidedAlignment(const std::vector<BinaryTreeNode>& tree, std::vector<FeatureMap>& feature_maps_transformed,
                              std::vector<std::vector<double>>& maps_ranges, FeatureMap& map_transformed,
                              std::vector<Size>& trafo_order);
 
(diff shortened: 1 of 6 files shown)


src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmTreeGuided.h
```diff
@@ -92,12 +92,12 @@ namespace OpenMS
      * @brief Align feature maps tree guided using align() of @ref OpenMS::MapAlignmentAlgorithmIdentification and use TreeNode with larger 10/90 percentile range as reference.
      *
      * @param tree Vector of BinaryTreeNodes that contains order for alignment.
-     * @param feature_maps_transformed Vector with copied input maps for transformation process.
+     * @param feature_maps_transformed Vector with input maps for transformation process.
```
cbielow, line 95: can you document, why they are not 'const&'

cbielow:
some tests are failing:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=449610

cbielow:
I have no idea why they fail though... lets try to rebuild the tests..

cbielow:
rebuild jenkins

jthueringer:
I have no idea, too. It's strange that the tests work with the msvc, but not with the other two compiler.

src/tests/topp/CMakeLists.txt
```diff
@@ -1165,6 +1165,14 @@ add_test("TOPP_MapAlignerTreeGuided_2_out2" ${DIFF} -in1 MapAlignerTreeGuided_2_
 set_tests_properties("TOPP_MapAlignerTreeGuided_2_out2" PROPERTIES DEPENDS "TOPP_MapAlignerTreeGuided_2")
 add_test("TOPP_MapAlignerTreeGuided_2_out3" ${DIFF} -in1 MapAlignerTreeGuided_2_output3.tmp -in2 ${DATA_DIR_TOPP}/MapAlignerTreeGuided_2_output3.trafoXML )
 set_tests_properties("TOPP_MapAlignerTreeGuided_2_out3" PROPERTIES DEPENDS "TOPP_MapAlignerTreeGuided_2")
+# loading files twice option
+add_test("TOPP_MapAlignerTreeGuided_3" ${TOPP_BIN_PATH}/MapAlignerTreeGuided -test -ini ${DATA_DIR_TOPP}/MapAlignerTreeGuided_parameters2.ini -in ${DATA_DIR_TOPP}/MapAlignerTreeGuided_1_input1.featureXML ${DATA_DIR_TOPP}/MapAlignerTreeGuided_1_input2.featureXML ${DATA_DIR_TOPP}/MapAlignerTreeGuided_1_input3.featureXML -out MapAlignerTreeGuided_3_output1.tmp MapAlignerTreeGuided_3_output2.tmp MapAlignerTreeGuided_3_output3.tmp)
+add_test("TOPP_MapAlignerTreeGuided_3_out1" ${DIFF} -in1 MapAlignerTreeGuided_3_output1.tmp -in2 ${DATA_DIR_TOPP}/MapAlignerTreeGuided_1_output1.featureXML )
+set_tests_properties("TOPP_MapAlignerTreeGuided_1_out1" PROPERTIES DEPENDS "TOPP_MapAlignerTreeGuided_3")
```
jpfeuffer, line 1171: copy paste

jpfeuffer: this is why they fail

jthueringer: You're absolutely right. I'll fix it. Thanks.

But I don't see why this led to non-existent *.tmp files.. Besides, the appleclang was running successfully this time.

jpfeuffer: Ah. So, we run the tests in parallel (4 at a time). If the tool test and its corresponding out tests are scheduled at the same time, the tmp files were not even produced yet.

jthueringer: Okay, I got it. Thanks again :)

jpfeuffer:
Otherwise, nice addition. We will merge when tests pass.

jpfeuffer:
since it worked on Win before, I think we can merge. Win is too full right now anyway.
@cbielow merge if you agree.

cbielow:
@jthueringer thanks!

PR #4705, opened 2020-05-07
[FEATURE] clean up file suffix renaming

cbielow:
some easier and more safe way of renaming of files.


cbielow:
tests pass (except CodeFactor, but the method is not really complex...)

timosachsenberg:
why got this moved from File to FileHandler?

timosachsenberg:
I would have voted for keeping it in File

cbielow:
The new functions are closely linked to FileType, thus FileHandler seemed more logical.
We could move it, or duplicate the function signature in File.h, simply calling FileHandler internally.

timosachsenberg:
it's fine with me but as a (new) user I would probably rather expected that method in a class called file

PR #4706, opened 2020-05-07
SwathWizard progress

cbielow:
other round of SwathWizard...

- can run OpenSwathWorkflow now
- supports drag'n'drop of input files
- some other minor fixes here and there


src/openms/source/APPLICATIONS/TOPPBase.cpp
```diff
@@ -1596,21 +1596,10 @@ namespace OpenMS
       writeLog_("Exit code: " + String(qp.exitCode()));
       if (any_failure)
       {
-        qp.close();
         return EXTERNAL_PROGRAM_ERROR;
       }
     }
 
-    if (debug_level_ >= 10)
```
timosachsenberg, line 1604: would it make sense to put it into the any failure block?

src/openms/source/DATASTRUCTURES/Param.cpp
```diff
@@ -1222,7 +1222,7 @@ OPENMS_THREAD_CRITICAL(oms_log)
             }
             this->root_.insert(local_entry, prefix); //->setValue(it.getName(), local_entry.value, local_entry.description, local_entry.tags);
           }
-          else
+          else if(verbose)
```
timosachsenberg, line 1225: so this is now silenced? might be a bit dangerous

src/openms_gui/include/OpenMS/VISUAL/DIALOGS/SwathTabWidget.h
```diff
@@ -46,23 +47,53 @@ namespace Ui
...
+      Param swath_param_; ///< the global Swath parameters which will be passed to OpenSwathWorkflow.exe, once updated with parameters the Wizard holds separately
+      Param swath_param_wizard_; ///< small selection of important parameters which the user can directly change in the Wizard
     };
 
   }
-}
+} // ns OpenMS
+
+// this is required to allow Ui_SwathTabWidget (auto UIC'd from .ui) to have a InputFile member
+using InputFile = OpenMS::InputFile;
```
timosachsenberg, line 97: would it make sense to put them in an extra GUI or VISUAL namespace?

timosachsenberg: Ui ?

src/openms_gui/source/VISUAL/DIALOGS/SwathTabWidget.cpp
```diff
@@ -62,19 +84,190 @@ namespace OpenMS
...
+        connect(ui->input_swath_windows, &InputFile::updatedCWD, this, &SwathTabWidget::broadcastNewCWD_);
     }
 
     SwathTabWidget::~SwathTabWidget()
     {
         delete ui;
     }
 
-    void SwathTabWidget::on_pushButton_clicked()
+    String infileToOSW(const String& infile)
```
timosachsenberg, line 120: needed as public member?

timosachsenberg:
looks good but it seems some tests are failing.
Screenshots?
Is this already in a state that e.g. @Leon-Bichmann could test it?

src/openms/source/APPLICATIONS/TOPPBase.cpp
cbielow: it was already .. see some lines above.

src/openms/source/DATASTRUCTURES/Param.cpp
cbielow: if you do 'verbose == true' then its active again :)
But you mean for other use cases like `-write_ini + -ini` ?

timosachsenberg: yep. Is it active by default if one uses an old ini file in a new version?

src/openms_gui/include/OpenMS/VISUAL/DIALOGS/SwathTabWidget.h
cbielow: the problem is the auto-generated header from the .ui file: it references an `InputFile* obj;`. 
We'd need to convince the UIC to generate `OpenMS::Inputfile* obj` or similar, but I did not bother to look for how to do that (yet)

src/openms_gui/source/VISUAL/DIALOGS/SwathTabWidget.cpp
cbielow: no. Its only used in SwathWidget.cpp. No need to export this.

cbielow:
feel free to test; not very exciting ATM though :)
image

cbielow:
tests pass. Any further comments?

PR #4709, opened 2020-05-11
Optimization in TheoreticalSpectrumGenerator and MSSpectrum + Fixing multithreaded SimpleSearchEngine

akriese:
While profiling the SimpleSearchEngine, we have found several points to optimize:
1. Too much time consumed by String construction and concatenation (metadata).
2. Sorting of the generated peak spectrum not working with partially sorted chunks of peaks
3. Lots of copying happening in MSSpectrum::select()
4. If losses shall be added, EmpiricalFormula::toString() takes up a huge amount of runtime.

This is how we optimized the above points:
1. String construction was pulled outside of loops, if possible. emplace_back() was used for temporary Strings.
2. As many peaks are added in sorted order to the spectrum, we can remember those chunks of peaks to sort them faster. MSSpectrum::sortByPositionPresorted() implements this with an idea similar to mergesort. For now, the new function is only used in TheoreticalSpectrumGenerator::getSpectrum(), as the chunk information has be added manually to a vector. Changes to the code were made to make sure that some chunks are actually pushed into the spectrum in sorted order (e.g. in addAbundantImmoniumIons_()). 
Also, a test for the new sort function was added to the MSSpectrum_test.
3. While working on the new sort function it became apparent, that select() takes up a big amount of time to move around the meta value arrays. By using moves and reserves we lowered that amount.
4. EmpiricalFormula::toString() spends a lot of time accessing maps. We cut that time 1. by caching already computed formula-strings in addLosses_faster_() in a map and 2. by changing the EmpiricalFormula::operator<() to make a search in sets and maps of EmpiricalFormula faster.

Results:
Depending on the parameters for the SimpleSearchEngine, the speed-up lies between 7% and 20%.

More information on the progress of this pull request can be seen [here](https://github.com/cbielow/OpenMS/pull/89)


timosachsenberg:
@enetz what do you think about these changes? I think you also made some attempts to optimize TSG

src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
```diff
@@ -643,16 +643,9 @@ namespace OpenMS
...
 
-    // both maps have same size
-    auto it = formula_.begin();
-    auto rhs_it = rhs.formula_.begin();
-    for (; it != formula_.end(); ++it, ++rhs_it)
-    {
-      if (*(it->first) != *(rhs_it->first)) return *(it->first) < *(rhs_it->first); // element
-      if (it->second != rhs_it->second) return it->second < rhs_it->second; // count
-    }
+    if (charge_ != rhs.charge_) return charge_ < rhs.charge_;
```
timosachsenberg, line 646: changes semantics / ordering
did you try the simpler:
return std::tie(formula_size(), charge_, formula_) < std::tie(rhs.formula_size(), rhs.charge_, rhs.formula_)
should be optimized to the same code

src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp
```diff
@@ -133,39 +133,39 @@ namespace OpenMS
...
     {
       if (spectrum.getIntegerDataArrays().size() > 0)
       {
-        charges = spectrum.getIntegerDataArrays()[0];
+        charges = std::move(spectrum.getIntegerDataArrays()[0]);
       }
       if (spectrum.getStringDataArrays().size() > 0)
       {
-        ion_names = spectrum.getStringDataArrays()[0];
+        ion_names = std::move(spectrum.getStringDataArrays()[0]);
```
timosachsenberg, line 155: why is there so much moving? would it be possible to work with pointer or references PeakSpectrum::IntegerDataArray* charges ?

src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp
```diff
@@ -351,58 +337,51 @@ namespace OpenMS
...
-          losses_names.emplace_back(formula.toString());
-        }
-      }
-    }
+    const String charge_str((Size)abs(charge), '+');
+    const String residue_str(Residue::residueTypeToIonLetter(res_type));
+    const String ion_ordinal_str(String(ion_ordinal) + "-");
+
+    // formula.toString() is extremely expensive, so we use a static map to remember what String belongs to which formula
+    static std::map<EmpiricalFormula, String> formula_str_cache;
```
timosachsenberg, line 367: the cache should be precomputed and passed to the function. please no static function variables if it can be avoided

src/openms/source/CHEMISTRY/EmpiricalFormula.cpp
akriese: std::tie() takes only lvalue references as far as I understand. Your approach works, when leaving the formula.size() out of the tie. Is 
`if (formula_.size() != rhs.formula_size()) return formula_.size() < rhs.formula_.size();
return std::tie(charge_, formula_) < std::tie(rhs.charge_, rhs.formula_);`
more favourable than the current solution?

timosachsenberg: yes I was thinking if it makes sense to have formula_.size() begin tested in formula_ < rhs.formula_
but I guess that is out of scope here. so feel free to leave it as is

src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp
akriese: I managed to use pointers. There is less moving now.

src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp
akriese: I've made the map a member of the TSG, so it is not static anymore. Precomputing seems rather complicated, as the set fx_losses is built up continously before every execution of addLossesFaster_().

timosachsenberg: seems still static here.

akriese: Just pushed the commit :)

timosachsenberg: I think this might not work in a multi-threaded context

akriese: True. I am gonna have a look at how to precompute the map, so it can be thread-safe.

timosachsenberg: sorry for giving so much headache. Most of my complaints are because of the non-ideal original design

timosachsenberg: and not because of your (by the way highly appreciated) changes

akriese: Still different results with different threads, but I tested a develop-build and actually there, the results vary too. So maybe, we should take a look into the algorithm in general.

akriese: Confirmed. The SimpleSearchEngine has different outputs for different threadnums in the current develop. The differences are very slight. Also, e.g. with four threads, the results vary between two runs.

akriese:
I've taken another look into the threading problem. There was an omp atomic missing, but fixing this didn't solve the whole problem.

akriese:
I've found out, that the sorting of the best n hits for each scan are sorted using the AnnotatedHit_::hasBetterScore(). Problem here is, that the sorting only considers the score of the compared hits. If two sequences have the same score for one scan (which happens fairly often), there is no rule on how to order them. That made the SimpleSearchEngine have different results when using multiple threads.
I changed the hasBetterScore()-function to compare more members of the AnnotatedHit_-struct, so that two equally scoring hits are sorted in a specified way.

Also, I had to build in a sorting of the peptide_ids-vector in postProcessHits_(), to make sure, the idXML-File always writes the hits in the same order (as using threads "shuffles" peptide_ids).

The results with these new changes differ slightly compared to what the SimpleSearchEngine has put out (using one thread) before the changes were made. But at least the results are always the same now indepent on how many threads are being used.
I'd appreciate, if someone could give me a feedback on this. Only if approved, I would change the tests (if needed), which may take some more time.

src/openms/include/OpenMS/DATASTRUCTURES/String.h
```diff
@@ -524,6 +527,25 @@ namespace OpenMS
...
 
+    bool operator==(const StringView other) const
+    {
+      if (size_ != other.size_) return false;
+
+      //same size
+      // same sequence, if both Views point to the same start
+      if (begin_ == other.begin_) return true;
+
+      const char * b = begin_;
```
cbielow, line 538: you can do this with less code: `strcmp(...)`
http://www.cplusplus.com/reference/cstring/strcmp/

src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp
```diff
@@ -340,6 +340,14 @@ void SimpleSearchEngineAlgorithm::postProcessHits_(const PeakMap& exp,
       }
     }
 
+#ifdef _OPENMP
+    // we need to sort the peptide_ids by scan_index in order to have the same output in the idXML-file
+    std::sort(peptide_ids.begin(), peptide_ids.end(), [](const PeptideIdentification& a, const PeptideIdentification& b)
```
cbielow, line 345: ouch. this looks expensive?

you can omit the sorting, if #threads == 1.
The ifdef usually does not help a lot, since most of the time, OpenMS is compiled with openmp enabled.

timosachsenberg: sorting should still be fine. I would say 95% of all time are spend in spectrum generation and comparison

akriese: So, should it be `if (omp_get_num_threads() > 1) sort();` ? Can I access the number of threads outside of a parallel region?

cbielow: no, should be `omp_get_max_threads()`
see  https://blogs.oracle.com/d/openmp-getting-the-maximum-number-of-threads

akriese: Okay, but should I leave the #ifdef _OPENMP? omp_get_num_threads() is not defined for certain systems, right?

src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp
```diff
@@ -133,39 +133,48 @@ namespace OpenMS
...
     }
 
-    PeakSpectrum::StringDataArray ion_names;
-    PeakSpectrum::IntegerDataArray charges;
+    MSSpectrum::Chunks chunks(spectrum);
+    PeakSpectrum::StringDataArray* ion_names;
+    PeakSpectrum::IntegerDataArray* charges;
 
-    if (add_metainfo_)
+    bool charges_dynamic = false, ion_names_dynamic = false;
```
cbielow, line 147: no need to use pointers.
Just initialize 
```
PeakSpectrum::StringDataArray* ion_names = 0;
PeakSpectrum::IntegerDataArray* charges = 0;
```

and call delete on it unconditionally.
(that is safe)

src/openms/include/OpenMS/CHEMISTRY/TheoreticalSpectrumGenerator.h
```diff
@@ -143,5 +147,8 @@ namespace OpenMS
     double pre_int_;
     double pre_int_H2O_;
     double pre_int_NH3_;
+
+    // formula.toString() is extremely expensive, so we use a member map to remember what String belongs to which formula
+    //mutable std::map<EmpiricalFormula, String> formula_str_cache_;
```
timosachsenberg, line 152: what about this comment line?

PR #4710, opened 2020-05-11
[FIX] writing of protein terminal mods correctly in Comet and MSGF adapter

### LINKED ISSUE #4681: Protein terminal mods not correctly passed to Search engines!
https://github.com/OpenMS/OpenMS/blob/ce59617cb0ca05a69c59736db58d8015b60c1276/src/topp/CometAdapter.cpp#L489

Will not work with Protein N-term.
Just by looking at it.

jpfeuffer:
fixes #4681


timosachsenberg:
cool!

PR #4712, opened 2020-05-11
report peak mem usage on Linux

cbielow:
- support peak memory usage calculation on Linux 
- report it for all platforms (which support it) per default

### DIFF
--- a/src/openms/configh.cmake
+++ b/src/openms/configh.cmake
@@ -155,6 +155,7 @@ CHECK_INCLUDE_FILE_CXX("time.h" OPENMS_HAS_TIME_H)
 CHECK_INCLUDE_FILE_CXX("sys/types.h" OPENMS_HAS_SYS_TYPES_H)
 CHECK_INCLUDE_FILE_CXX("sys/times.h" OPENMS_HAS_SYS_TIMES_H)
 CHECK_INCLUDE_FILE_CXX("sys/time.h"  OPENMS_HAS_SYS_TIME_H)
+CHECK_INCLUDE_FILE_CXX("sys/resource.h"  OPENMS_HAS_SYS_RESOURCE_H)
 CHECK_INCLUDE_FILE_CXX("stdint.h"  OPENMS_HAS_STDINT_H)
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


jpfeuffer:
https://stackoverflow.com/questions/669438/how-to-get-memory-usage-at-runtime-using-c

to potentially add for Mac. Should be simple.

cbielow:
> 
> 
> https://stackoverflow.com/questions/669438/how-to-get-memory-usage-at-runtime-using-c
> 
> to potentially add for Mac. Should be simple. Same function just different header and different scale.

Done. But the code on SO multiplies by 1024 for linux, yet the docs https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getrusage.2.html say that the number is kilobytes (same as linux). So I did not divide, hoping that the docs are correct. 

Someone with MacOS please confirm.

jpfeuffer:
```
While loading '../share/OpenMS/examples/BSA/BSA1.mzML': Obsolete CV term 'MS:1000763 - PeakPicker' used in tag 'software'.
Progress of 'loading spectra list':
-- done [took 1.80 s (CPU), 1.87 s (Wall)] -- 


Memory usage (loading MS data): 17 MB (working set delta), 21320 MB (peak working set delta)

-- General information --

File name: ../share/OpenMS/examples/BSA/BSA1.mzML
File type: mzML

MS levels: 1, 2
Total number of peaks: 479455
Number of spectra: 1684

Number of spectra per MS level:
  level 1: 564
  level 2: 1120

Ranges:
  retention time: 1501.41 .. 2499.52
  mass-to-charge: 85.81 .. 799.95
  intensity:      0.67 .. 11977811.00

Peak type metadata (estimated)
  level 1: Centroid (Centroid)
  level 2: Centroid (Centroid)


FileInfo tooks 3.05 s (wall), 2.93 s (CPU), 0.10 s (system), 2.83 s (user); Peak Memory Usage: 81932 MB.
```

for BSA_1.mzML so yes, probably divide by 1024

jpfeuffer:
as described in  https://stackoverflow.com/questions/59913657/strange-values-of-get-rusage-maxrss-on-macos-and-linux my man page also says:
`ru_maxrss    the maximum resident set size utilized (in bytes).`

cbielow:
ok, this should do it...

timosachsenberg:
Nice

PR #4713, opened 2020-05-11
[FEATURE] add new MSGF options

### LINKED ISSUE #4711: Allow setting new parameters of new MSGF+ version(s)
"-tasks" and "-missedCleavages": Which are both vital for huge mzMLs to reduce memory footprint.

jpfeuffer:
fixes #4711


jpfeuffer:
linux error unrelated (and not significant for this java program).

cbielow:
this breaks with any older MSGF+, e.g.  (v2018.01.30) (30 January 2018).
That is probably not intentional?

PR #4720, opened 2020-05-12
[Fix] repeating secondary search engine settings in mzTab MTD section.

jpfeuffer:
Plus little fix in IDScoreSwitcherAlgo. Added test.


jpfeuffer:
Wait with merge. Will add a test with multiple runs.

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
```diff
@@ -124,8 +124,17 @@ namespace OpenMS
         // also already have the correct score set
         return;
       }
+
+      if (t.hasSuffix("_score"))
```
timosachsenberg, line 128: this is kind of ugly in OpenMS. Would probably be better to deal with that on the parse / writer / data structure side. we just need to make sure to get rid of that stuff at some point in the future

src/tests/topp/MzTabExporter_4_output.mzTab
```diff
@@ -6,16 +6,20 @@ MTD	description	OpenMS export from consensusXML
...
+MTD	software[3]-setting[2]	db_version:
+MTD	software[3]-setting[3]	enzyme:Trypsin
+MTD	software[3]-setting[4]	precursor_mass_tolerance:5.0
+MTD	software[3]-setting[5]	precursor_mass_tolerance_ppm:ppm
+MTD	software[3]-setting[6]	fragment_mass_tolerance:0.1
+MTD	software[3]-setting[7]	fragment_mass_tolerance_ppm:Da
+MTD	software[3]-setting[8]	charges:
+MTD	software[3]-setting[9]	missed_cleavages:1
+MTD	software[3]-setting[10]	fixed_modificiations:Carbamidomethyl (C)
+MTD	software[3]-setting[11]	variable_modificiations:Acetyl (N-term) Gln->pyro-Glu (N-term Q) Glu->pyro-Glu (N-term E)
```
timosachsenberg, line 22: maybe a different separator?

src/tests/topp/ProteinQuantifier_9_output_mztab.mzTab
```diff
@@ -6,16 +6,19 @@ MTD	description	OpenMS export from consensusXML
...
 MTD	software[2]	[, , Mascot, ]
-MTD	software[2]-setting[1]	null
-MTD	software[2]-setting[2]	null
-MTD	software[2]-setting[3]	null
-MTD	software[2]-setting[4]	fragment_mass_tolerance:0.0
-MTD	software[2]-setting[5]	fragment_mass_tolerance_unit:Da
-MTD	software[2]-setting[6]	precursor_mass_tolerance:0.0
-MTD	software[2]-setting[7]	precursor_mass_tolerance_unit:Da
-MTD	software[2]-setting[8]	enzyme:unknown_enzyme
+MTD	software[2]-setting[1]	db:
```
timosachsenberg, line 11: should we print out empty settings at all? if empty use null?

src/tests/topp/MzTabExporter_1_output.mzTab
```diff
@@ -18898,53 +18901,53 @@ PEP	AITHLNNNFMFGQK	sp|P14866|HNRPL_HUMAN	1	null	null	null	0.0	0.0	0-UNIMOD:737,1
...
-PSM	AQLLVNPK	39	sp|Q13416|ORC2_HUMAN	1	uniprot_homo-sapiens_and_E-coli_June-2017_canonical-without-isoforms_reviewed_plus-contaminants_incl_bovinetrypsin_plus_shuffled	null	[, , MS-GF+, Release (v2017.08.23)]	2.40963855421687e-04	0-UNIMOD:737,8-UNIMOD:737	2014.885800000000018	2	670.934020996093977	670.936937418220964	ms_run[1]:controllerType=0 controllerNumber=1 scan=11506	R	K	40	47	HCD	0	74.0	77.0	6.760485e-11	1.610822e-03	6.760485e-11	670.936950683593977	47	1	unique	40	0
+PSM	AQLLVNPK	39	sp|Q13416|ORC2_HUMAN	1	uniprot_homo-sapiens_and_E-coli_June-2017_canonical-without-isoforms_reviewed_plus-contaminants_incl_bovinetrypsin_plus_shuffled	null	[, , MS-GF+, Release (v2017.08.23)]	2.40963855421687e-04	0-UNIMOD:737,8-UNIMOD:737	2014.885800000000018	2	670.934020996093977	670.936937418221078	ms_run[1]:controllerType=0 controllerNumber=1 scan=11506	R	K	40	47	HCD	0	74.0	77.0	6.760485e-11	1.610822e-03	6.760485e-11	670.936950683593977	47	1	unique	40	0
```
timosachsenberg, line 18942: can we limit the idigits in RT and m/z output  to e.g. 9 fractional digits?

src/tests/topp/QualityControl_1_out.mzTab
```diff
@@ -4,18 +4,23 @@ MTD	mzTab-type	Quantification
 MTD	title	ConsensusMap export from OpenMS
 MTD	description	OpenMS export from consensusXML
 MTD	protein_search_engine_score[1]	[, , one-peptide-rule XTandem, ]
+MTD	protein_search_engine_score[2]	[, , one-peptide-rule XTandem, ]
```
timosachsenberg: why are there 3?

src/tests/topp/MzTabExporter_5_output.mzTab
```diff
@@ -0,0 +1,145 @@
+MTD	mzTab-version	1.0.0
+MTD	mzTab-mode	Summary
+MTD	mzTab-type	Quantification
+MTD	title	ConsensusMap export from OpenMS
+MTD	description	OpenMS export from consensusXML
+MTD	protein_search_engine_score[1]	[, , Epifany Posterior Probability, 2.6.0-pre-feature-cXMLProtGroups-2020-04-06]
+MTD	peptide_search_engine_score[1]	[, , Percolator Posterior Error Probability, 3.02]
+MTD	peptide_search_engine_score[2]	[, , XTandem, ]
+MTD	psm_search_engine_score[1]	[, , Percolator Posterior Error Probability, 3.02]
```
timosachsenberg, line 9: different number of peptide and psm search engine scores...

src/tests/topp/ProteinQuantifier_9_output_mztab.mzTab
jpfeuffer: But if it is just says "null" no one knows what parameter was "null". That was my point.

src/tests/topp/MzTabExporter_4_output.mzTab
jpfeuffer: Instead of ":"?

src/tests/topp/MzTabExporter_1_output.mzTab
jpfeuffer: I don't think so. I think the float type gets lost due to conversion to MzTabDouble.

src/tests/topp/MzTabExporter_4_output.mzTab
timosachsenberg: no the space character

src/tests/topp/ProteinQuantifier_9_output_mztab.mzTab
timosachsenberg: db:null ?

src/tests/topp/MzTabExporter_1_output.mzTab
jpfeuffer: We can only go through all occurrences where we write an RT and modify the output generator precision beforehand and change it back.

src/tests/topp/MzTabExporter_5_output.mzTab
jpfeuffer: Yes, which is now correct. In theory we should be able to conserve the underlying search engine score. It is not really implemented yet since you need a big map of possible score types for each search engine and do a pre lookup through all the peptide hits meta values.

src/tests/topp/MzTabExporter_4_output.mzTab
jpfeuffer: Ah ok. What is best for mztab? Just comma?

src/tests/topp/ProteinQuantifier_9_output_mztab.mzTab
jpfeuffer: Ok that is fine for me.

src/tests/topp/QualityControl_1_out.mzTab
jpfeuffer: So yes, this is kind of a Todo and made me think if it wouldn't be nicer to have run level info. But yes, since for now I make peptide search scores unique, we should probably do the same for proteins.

src/openms/include/OpenMS/ANALYSIS/ID/IDScoreSwitcherAlgorithm.h
jpfeuffer: Yes definitely. It also makes it super hard to recover old scores since they are not registered at the run level anymore. But this would be solved in the new ID structure.

src/tests/topp/ProteinQuantifier_9_output_mztab.mzTab
timosachsenberg: it's more a question than a suggestion

src/tests/topp/MzTabExporter_1_output.mzTab
timosachsenberg: ok let's keep for now

src/tests/topp/QualityControl_1_out.mzTab
timosachsenberg: yes I think so but this is currently not supported by mzTab

PR #4721, opened 2020-05-13
Fibonacci heap

Felix-Droop:
# FeatureLinkerUnlabeledQT Runtime Optimization

## Benchmarks

The optimization lead to a huge improvement in runtime and a neglectible increase of peak memory usage.

```
small = 2 featureXMLs (total 291 MB), large = 5 featureXMLs (total 814 MB)
ref = before (no optimizations, branch "develop"), opt = now (fibonacci heap)
CPU time only for 'linking features' step, RAM usage for whole program:

opt small:    took  1.19 s (CPU), Peak Memory Usage  420 MB
ref small:    took 02:54 m (CPU), Peak Memory Usage  414 MB
opt large:    took  6.24 s (CPU), Peak Memory Usage 1157 MB
ref large:    took 41:41 m (CPU), Peak Memory Usage 1123 MB
```

## Tests

We ran the reference implementation (before any change) and our improved version with the `-test` flag on the sample data set which we also used for the benchmarks. We used `FuzzyDiff` with default parameters (meaning no whitelist and no numerical error tolerance) to compare the output files. No differences were found.

We changed the interface of the `QTCluster` class and therefore had to adapt the corresponding test. Finally, we had to slightly modifiy the input data of the FeatureFinderUnlabeled test 6, because this test had two possible correct solutions.

## Fibonacci Heap

The main optimization idea was to eliminate the runtime dominating linear search for the cluster with the best quality. In order to avoid this we stored the clusters in a fibonacci heap. This data structure enables us to extract the wanted cluster efficiently. It is cheap to perform a priority update of already inserted clusters as well.

We are using the `boost::heap::fibonacci_heap` from the boost version 1.49. Hopefully this doesn't cause any continuous integration problems.

## Hot-Cold Splitting

We implemented hot-cold splitting in the `QTCluster` class. This makes moving instances of it cheap. The cold data has been outsourced to the class `Data_` and is pointed to by a member.

## Refactoring

To analyze the runtime and find the bottleneck we refactored some of the long functions. We extracted certain code blocks into new methods. It is now easier to understand the structure of the code. Unfortunately, git is not able to match the old code blocks to the new methods. Therefore this pull request a little looks messy and like more change than we actually did.


src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -239,52 +241,57 @@ namespace OpenMS
...
+    // number of clusters == number of data points:
+    Size size = cluster_heads.size();
+
+    // fill the element mapping
+    for (Heap::iterator it = cluster_heads.begin();
+         it != cluster_heads.end(); ++it)
     {
-      NeighborList neigh = it->getAllNeighbors();
-      for (NeighborList::iterator n_it = neigh.begin(); n_it != neigh.end(); ++n_it)
+      NeighborMap const& neigh = it->getAllNeighborsDirect();
```
timosachsenberg, line 266: we use a different style throughout OpenMS:
const NeighborMap&
So changed this and other places would make it more consistent.

src/openms/source/DATASTRUCTURES/QTCluster.cpp
```diff
@@ -182,47 +190,43 @@ namespace OpenMS
...
-    }
+    // add center point to the copy
+    neighbors_copy[data_->center_point_->getMapIndex()].second = data_->center_point_;
 
-    // since we are finalized, we do not need to care about the annotation
-    for (NeighborMap::const_iterator it = neighbors_.begin(); it != neighbors_.end(); ++it)
-    {
-      elements[it->first] = it->second.second;
-    }
+    return neighbors_copy;
```
timosachsenberg, line 204: I guess named return value optimization (NRVO) is in effect here... should we add a comment?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
Felix-Droop: Done.

src/openms/source/DATASTRUCTURES/QTCluster.cpp
Felix-Droop: Done.

src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
```diff
@@ -110,124 +109,134 @@ namespace OpenMS
...
-    Int x_coord_;
-
-    /// y coordinate in the grid cell
-    Int y_coord_;
+     * @brief Class to store the bulk internal data (neighbors, annotations, etc.)
+     * 
+     * Has no functionality without a QTCluster pointing to it.
+     * Create object of this class before calling constructor of QTCluster
+     */
+    class BulkData
```
timosachsenberg, line 140: do these internal classes and structs need to be public?

src/tests/topp/FeatureLinkerUnlabeledQT_6_output.consensusXML
```diff
@@ -56,11 +56,11 @@
 				<UserParam type="int" name="map_index" value="2"/>
 			</PeptideIdentification>
 		</consensusElement>
-		<consensusElement id="e_3332699010107892018" quality="0.73605">
-			<centroid rt="410.0" mz="500.010999999999967" it="600.0"/>
+		<consensusElement id="e_3332699010107892018" quality="0.736175">
+			<centroid rt="410.033333333333303" mz="500.010999999999967" it="600.0"/>
```
timosachsenberg, line 60: this seems to be a different element. is this correct?

timosachsenberg: ah I see , input changed

src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
Felix-Droop: Currently it has to be, because we want to store the `BulkData` objects in a vector and therefore construct them independently of the `QTCluster`objects. We hope to improve cache behavior by this. Canging it will probably lead to a small performance penalty. Only some seconds though. Would you prefer it if we changed it?

timosachsenberg: hmm, not sure. depends a bit on if this would give users an "easier" (e.g., fewer parameters passed) interface and if it hides some of the added complexity. I let you or @cbielow decide.

timosachsenberg:
By the way, this sounds like an awesome improvement!

Felix-Droop:
rebuild jenkins

src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
Felix-Droop: We decided to stick with the performance. 

One could implement a wrapper that hides the ugly interface and presents a better one. But this would probably result in yet another pretty big change in this code and would not be worth the effort in my opinion.

jpfeuffer:
Hi everyone,

it seems like this PR broke the nightly builds in Debug mode.
Debug mode has often some additional OPENMS_PRECONDITIONs that may fail.
I understand that for profiling most of the development was in Release mode.
A hotfix would be highly appreciated though. :+1:

PR #4723, opened 2020-05-14
temporary directory class

### LINKED ISSUE #4707: RAII temporary directory class
implement a non-copyable class which creates a new temporary directory (as subdirectory of File::getTempDirectory()) -- see TOPPBase::makeTempDirectory_() for an example.

On destruction of the object, the temp-dir should get deleted. see `TOPPBase::removeTempDirectory_` for an example.

The goal is basically to move the functionality of TOPPBase::makeTempDirectory_() and TOPPBase::removeTempDirectory_ into a separate class and make TOPPBase smaller.

The constructor of the new class should have a bool option to not delete the temp-dir on destruction (for debug purposes).

Waschi97:
- Moved the functionality of `TOPPBase::makeTempDirectory()` and `TOPPBase::removeTempDirectory` into the new class `TempDir` which is a nested class in `File`.

- Added a test for this new class.

- Replaced each usage of `TOPPBase::makeTempDirectory()` and `TOPPBase::removeTempDirectory` with a `File::TempDir` object

Closes #4707


jpfeuffer:
Looks good but I think you need to add a TempDir(debug_lvl_ >= 2) in the search adapters. Since this was the previous default. This might not even show up in the tests.

src/openms/include/OpenMS/SYSTEM/File.h
```diff
@@ -55,6 +55,34 @@ namespace OpenMS
...
+
+      /// Destroy temporary folder (can be prohibited in Constructor)
+      ~TempDir();
+
+      /// Return path to temporary folder
+      const String& getPath() const;
+
+    private:
+      /// Make class noncopyable
+      TempDir(const TempDir&);
```
cbielow, line 79: while this is technically correct, it might still be confusing, because the move-assignment-op and move-ctor and not mentioned here (So one might wonder if moving is allowed and what strange things would happen if it were).
Now, since you declared a custom copy-ctor, no move* will be generated by the compiler implicitly, which is good. But to be more explicit, it would be nice to state this, and do it the C++11 way:
```
public:
   /// delete all means to copy or move a TempDir
   TempDir(const TempDir&) = delete;
   TempDir& operator=(const TempDir&) = delete;
   TempDir(TempDir&& ) = delete;
   TempDir& operator=(TempDir&&) = delete;
```

src/openms/source/SYSTEM/File.cpp
```diff
@@ -65,6 +65,31 @@ using namespace std;
 namespace OpenMS
 {
 
+  File::TempDir::TempDir(bool keep_dir)
```
cbielow, line 68: now the nitpicking starts, sorry in advance :)

You could make use of the Ctor's initializer list and remove some redundant lines, which would make this a bit more readable (IMHO!)

```
File::TempDir::TempDir(bool keep_dir)
  : keep_dir_(keep_dir)
{
    temp_dir_ = File::getTempDirectory() + "/" + File::getUniqueName() + "/";

    OPENMS_LOG_DEBUG << "Creating temporary directory '" << temp_dir_ << "'" << std::endl;
    QDir().mkpath(temp_dir_.toQString());
  };
```

the `QDir::toNativeSeparators` is technically correct, but usually does not provide any benefit (Windows can deal fine with / in paths). I vaguely remember there was a disadvantage doing \, but I cannot recall right now...
But thats a matter of style.

src/topp/LuciphorAdapter.cpp
```diff
@@ -490,7 +490,8 @@ class LuciphorAdapter :
     }
 
     //tmp_dir
-    String temp_dir = makeAutoRemoveTempDirectory_();
+    File::TempDir dir(debug_level_ >= 2);
```
cbielow, line 493: nitpicking again: to avoid a level of indirection for the reader, you could just use
`File::TempDir temp_dir (debug_level_ >= 2);`
i.e. without `String temp_dir = dir.getPath();`
and replace occurences of `temp_dir` with `temp_dir.getPath()`

cbielow:
everything I commented on in this second round is optional, and just meant to show another approach.
Feel free to ignore it, or, pull in the changes.

cbielow:
changes to MSFraggerAdapter yield a compile error:
http://cdash.openms.de/viewBuildError.php?buildid=458457
(order of initialization vs. declaration order of members)

Otherwise, I'd say ready for merge.
Thanks!

PR #4725, opened 2020-05-15
Refactoring and Integration of SIRIUS 4.6.0

oliveralka:
SIRIUS 4.4.17 switched to a sub-tool based environment which can be called more or less individually. 

Here we refactored the current version to retain the current functionality. 

TODO: 
Integration of decoy generation via "passatutto", which is called internally by SIRIUS. I am currently not sure what the best way would be to allow to use decoy generation in a flexible fashion e.g. for the AssayGeneratorMetabo and an additional DecoyGeneration tool for metabolomics, which can used additional/different methods. Since for this specific method SIRIUS has to be called. - maybe an additional input and output format is needed for the SiriusAdapter.


oliveralka:
@jpfeuffer Anything else I need to do for the tests to use the correct SIRIUS version (added a branch to OpenMS/Thirdparty and added the respective one to the current PR).

jpfeuffer:
No, I think at least the openms/PR/bldtst jobs should checkout a fresh version of THIRDPARTY

jpfeuffer:
Did you have a look at a direct invocation of Sirius in my old PR?

oliveralka:
As I recall there still have been some problems with that PR, or? 
In that regard, I talked to Marcus - so we can actually get a .jar which we can call directly in the future, but  I think we should still finish this version - to include the decoy generation.

oliveralka:
test - still needs some clean-up and refactoring.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -109,68 +163,156 @@ namespace OpenMS
...
+        enclose->defaults_.setValidStrings(openms_param_name, choices);
+      }
+
+      void withMinInt(int value)
+      {
+        enclose->defaults_.setMinInt(openms_param_name, value);
+      }
+    };
+
+    class ParameterSection
```
jpfeuffer, line 208: This almost looks like a full general abstraction of command line tools and should probably go into a separate class.

src/openms/include/OpenMS/ANALYSIS/TARGETED/MetaboTargetedTargetDecoy.h
```diff
@@ -0,0 +1,94 @@
...
+    {
+      String identifier;
+      String target_compound_ref;
+      String decoy_compound_ref;
+      std::vector<double> target_product_masses;
+      std::vector<double> decoy_product_masses;
+    };
+
+    /**
+    @brief Constructs a mass mapping of targets and decoys using a the unique mids identifier.
```
jpfeuffer, line 66: spelling

src/openms/include/OpenMS/FORMAT/DATAACCESS/SiriusFragmentAnnotation.h
```diff
@@ -42,9 +42,35 @@ namespace OpenMS
...
+
+          @param sirius_workspace_subdirs: Vector of paths to SIRIUS subdirectories.
+          @param use_exact_mass: Option to use exact mass instead of peak mz in MSSpectrum.
+          */
+          static std::vector<SiriusTargetDecoySpectra> extractAndResolveSiriusAnnotations(const std::vector<String>& sirius_workspace_subdirs, const double& score_threshold, bool use_exact_mass);
+
           /**
           @brief extractSiriusFragmentAnnotationMapping  
-          Extract native id (./spectrum.ms) and fragment annotation (./spectra/1_sumformula.ms) from SIRIUS output (per compound).
+          Extract concatenated native ids and concatenated mids (unique identifier) from (./spectrum.ms) and fragment annotation (./spectra/1_sumformula.tsv) from SIRIUS output (per compound).
```
jpfeuffer, line 73: Is this "mids" or "m_ids"?

jpfeuffer: See also other occurrences

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -46,12 +46,38 @@ using namespace std;
...
+    return adduct_charge;
+  }
+
   bool MetaboTargetedAssay::intensityLess_(Peak1D a, Peak1D b)
   {
     return (a.getIntensity() < b.getIntensity());
   }
 
-  // method to extract a potential transistions based on the ms/ms based of the highest intensity precursor or a consensus spectrum
+  // method to extract a potential transitions based on the ms/ms based of the highest intensity precursor or a consensus spectrum
```
jpfeuffer, line 80: still wrong. plural vs singular

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
```diff
@@ -135,26 +265,28 @@ namespace OpenMS
       }
 
       // use first file in folder (rank 1)
-      dir.setFilter(QDir::Files | QDir::Hidden | QDir::NoSymLinks);
-      QFileInfoList list = dir.entryInfoList();
-      QFileInfo firstfile = list[0]; 
+      String filename = rank_filename.at(1); // rank 1
```
jpfeuffer, line 268: what if folder empty?

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
```diff
@@ -184,19 +316,91 @@ namespace OpenMS
...
+    std::map< Size, double > rank_score = SiriusFragmentAnnotation::extractCompoundRankingAndScore_(path_to_sirius_workspace);
+
+    const std::string sirius_spectra_dir = path_to_sirius_workspace + "/decoys/";
+    QDir dir(QString::fromStdString(sirius_spectra_dir));
+    if (dir.exists())
+    {
+      msspectrum_to_fill.setMetaValue("peak_mz", DataValue("mz"));
+
+      // use first file in folder (rank 1)
+      String filename = rank_filename.at(1); // rank 1
```
jpfeuffer, line 355: same here, what if map is empty?

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
```diff
@@ -72,28 +72,40 @@ String SiriusMzTabWriter::extractFeatureId(const String& path)
   return feature_id;
 }
 
+std::map< String, Size > SiriusMzTabWriter::extract_columnname_to_columnindex(CsvFile& csvfile)
```
jpfeuffer, line 75: This file share A LOT with CSIFingerIDMzTabWriter.
In general, why can't it use the general MzTabWriter? No small molecule support? Or is there a SMMZTabWriter?

jpfeuffer (commented):
In general, you should also check if you can get error logs from sirius with the new version.
If not, please adopt the direct calling mechanism from:
https://github.com/OpenMS/OpenMS/pull/4231

src/openms/include/OpenMS/FORMAT/DATAACCESS/SiriusFragmentAnnotation.h
oliveralka: Should be mids, based on 
`std::vector<String> mids; ///< native ids and identifier for multiple possible identification via AMS ("|" separator)
`
In the SiriusMSConverter.h 

Where does the m_ids come from?

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
oliveralka: Log_Debug output see line 332

src/openms/include/OpenMS/FORMAT/DATAACCESS/SiriusFragmentAnnotation.h
jpfeuffer: I mean, it confuses me since I read it as "mids" = "middle points".
If it is a contraction of m... and IDs, maybe it is better called m_ids or mIDs

oliveralka: ah, yeah that makes sense - will call it differently then!

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
oliveralka: Yes, there are some overlapping points. Most of the overlap is refactored (I extracted an additional function - next push), but SiriusMzTabWriter will read `"/formula_candidates.tsv"`, which is structurally different from the CSIFingerID 
`"/structure_candidates.tsv"`. That why I have two individual readers/writers. 

The overlapping methods are implemented in the SiriusMzTabWriter and then used in the CSI one
```
   /**
    @brief Extract scan_index from filepath
    */
    static int extractScanIndex(const String& path);

    /**
    @brief Extract scan_number from filepath
    */
    static int extractScanNumber(const String& path);

    /**
    @brief Extract feature_id from filepath
    */
    static String extractFeatureId(const String& path);

    /**
    @brief Extract columnname and index based in SIRIUS entries
    */
    static std::map< String, Size > extract_columnname_to_columnindex(CsvFile& csvfile);

    /**
     @brief Extract mz, rt of the precursor and the nativeID of the corresponding MS2 spectra in the spectrum.ms file
    */
    static SiriusSpectrumMSInfo extractSpectrumMSInfo(const String& single_sirius_path);
```

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
oliveralka: Maybe. SIRIUS since 4.4 is now actually not one algorithm/command-line tool anymore, but a growing aggregation of different algorithms, which can be called either in dependence on each other or independently. 

This abstraction allows the integration of additional algorithms/tools. Currently, SIRIUS, FingerID & passatutto are integrated. Later we could integrate ZODIAC as well. Of course at some point the SiriusAdapter might just blow up.

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
oliveralka: Yes, you are correct. I moved the functions now, that the directory is checked beforehand. Since if a spectra/decoy directory is available, then the specific files are there as well. 

```
      std::map< Size, String > rank_filename = SiriusFragmentAnnotation::extractCompoundRankingAndFilename_(path_to_sirius_workspace);
      std::map< Size, double > rank_score = SiriusFragmentAnnotation::extractCompoundRankingAndScore_(path_to_sirius_workspace);
```

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
jpfeuffer: I just mean, that by providing a list of subcommands and a mapping from TOPPBase parameters to tool parameters and with parameter modifiers, these classes could be generalized to be used for any command-line tool that we are wrapping. But I don't have a strong opinion on that. Can stay here for now.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -109,68 +163,156 @@ namespace OpenMS
...
-
-      // getter (used to call functions from SiriusMSConverter, SiriusMzTabWriter, CsiFingerIDMzTabWriter)
-      String getFeatureOnly();  
-      String getNoMasstraceInfoIsotopePattern();
-      int getIsotopePatternIterations();
-      int getCandidates();
-      int getTopNHits();
+      const vector<String> callSiriusQProcess(const String& tmp_ms_file,
+                                              const String& tmp_out_dir,
+                                              String& executable,
```
cbielow, line 183: can be `const`?

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -109,68 +163,156 @@ namespace OpenMS
...
+                                   const FeatureMapping::FeatureToMs2Indices& feature_mapping,
+                                   const MSExperiment& spectra);
 
-      
       /**
       @brief Call SIRIUS with QProcess
 
-      @return Vector with paths to a compound 
-
       @param tmp_ms_file: path to temporary .ms file
```
cbielow, line 174: the `:` is probably confusing doxygen?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -44,99 +43,470 @@
...
+      }
+      return false;
+    }
+
+    String SiriusAdapterAlgorithm::determineSiriusExecutable(String &executable)
+    { 
+      // if executable was not provided
+      if (executable.empty())
+      {
+        const String &qsiriuspathenv = QProcessEnvironment::systemEnvironment().value("SIRIUS_PATH");
```
cbielow, line 494: pheww...
https://en.cppreference.com/w/cpp/utility/program/getenv
?

cbielow:
there seems to be a second ecosystem for parameter passing and parsing...
I've only skimmed it, but is that really required?

oliveralka:
@cbielow Yes, I think so. Since it makes the integration of further SIRIUS tools/algorithms way easier. The idea comes from @lukaszimmermann maybe he can comment on that.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
oliveralka: ....ups

oliveralka:
Next week we should get a multiOS build of SIRIUS 4.5 - I will integrate it into THIRDPARTY and test everything. Then this PR should be ready.

oliveralka:
@jpfeuffer Error logs are reported now if SIRIUS fails e.g.
```
ut sirius --ppm-max-ms2=10 --db=all --tree-timeout=0 --candidates=5 --compound-timeout=100 --ppm-max=10 --profile=qtof passatutto
Executing: /home/alka/software/sirius-linux64-headless-4.4.29/bin/sirius
FATAL: External invocation of Sirius failed. Standard output and error were:
178 redundant branch up with value 0.571429 current lower 1 solution 1 dj -0.346554
84 redundant branch down with value 0.454545 current upper 0 solution 0 dj 0.770051
2204 redundant branch down with value 0.5 current upper 0 solution 0 dj 0.224938
93 redundant branch up with value 0.5 current lower 1 solution 1 dj -0.0845271
7 redundant branch down with value 0.333333 current upper 0 solution 0 dj 0.140972
115 redundant branch down with value 0.357143 current upper 0 solution 0 dj 0.769314
87 redundant branch up with value 0.333333 current lower 1 solution 1 dj -1.19273
11 redundant branch down with value 0.5 current upper 0 solution 0 dj 0.150306
.
.
.
INFO  02:01:41 - Solution is infeasible
SEVERE02:01:41 - RecalibrationJob (50699): Failed!
java.lang.NullPointerException
77
```
Will close #4231

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -178,9 +199,316 @@ class TOPPAssayGeneratorMetabo :
...
+                            return assay.molecular_formula != pr->first.first ||
+                                   assay.compound_adduct != pr->first.second;
+                          }), mta.end());
+    }
+  }
+
+  // Use FeatureGroupingAlgorithmQT
+  // based on mz and rt (minimum feature) - MetaValue index to vector<MetaboTargetedAssay>
+
+  struct Intermediate
```
oliveralka, line 257: Class
more descriptive
add further comments (also to functions)

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -545,42 +878,29 @@ class TOPPAssayGeneratorMetabo :
...
     } // end iteration over all files
 
-    // use first rank based on precursor intensity
-    std::map< std::pair <String,String>, MetaboTargetedAssay > map_mta;
-    for (const auto& it : v_mta)
-    {
-      pair<String,String> pair_mta = make_pair(it.compound_name, it.compound_adduct);
+    // group ambiguous identification based on precursor_mz and feature retention time
+    // Use featureMap and use FeatureGroupingAlgorithmQT
+    std::map< std::pair <double,double>, vector<MetaboTargetedAssay> > ambiguity_groups = buildAmbiguityGroup(v_mta, ar_mz_tol, ar_rt_tol, ar_mz_tol_unit_res, in.size());
```
oliveralka, line 887: Maybe additional class instead of map<pair, vector>

`map<additional class>`

```
additional class
double
double
vector<MTA>
```

src/utils/SiriusAdapter.cpp
```diff
@@ -131,49 +133,48 @@ class TOPPSiriusAdapter :
...
     setValidFormats_("in_featureinfo", ListUtils::create<String>("featureXML"));
 
-    registerOutputFile_("out_sirius", "<file>", "", "MzTab Output file for SiriusAdapter results", false);
+    registerOutputFile_("out_sirius", "<file>", "", "MzTab output file for SIRIUS results", false);
     setValidFormats_("out_sirius", ListUtils::create<String>("mzTab"));
 
     registerOutputFile_("out_fingerid","<file>", "", "MzTab output file for CSI:FingerID, if this parameter is given, SIRIUS will search for a molecular structure using CSI:FingerID after determining the sum formula", false);
     setValidFormats_("out_fingerid", ListUtils::create<String>("mzTab"));
 
+    registerOutputFile_("out_decoys","<file>", "", "Experimental output in project space.", false, true);
```
oliveralka, line 142: Maybe remove - does not yet work in SIRIUSAdapter - another PR 
What would be needed here and why would someone need the decoys in this case?

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -85,7 +86,11 @@ namespace OpenMS
     Peak1D peak;
 
     // monoisotopic_trace
-    peak_index = getHighestIntensityPeakInMZRange(precursor_mz, precursor_spectrum, 10, true);
+    const int tolerance = 10;
```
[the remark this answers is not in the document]
oliveralka: I think it is fine to set the error to 10 ppm (max), since is about finding the peak with the highest intensity based on a test mass-to-charge in the spectrum .e.g the precursor.

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -331,188 +365,299 @@ namespace OpenMS
...
+  // method to pair compound information (SiriusMSFile) with the annotated target spectrum from Sirius based on the m_id (unique identifier)
+  std::vector< MetaboTargetedAssay::CompoundTargetDecoyPair > MetaboTargetedAssay::pairCompoundWithAnnotatedSpectra(const std::vector<SiriusMSFile::CompoundInfo>& v_cmpinfo,
+                                                                                                                  const std::vector<SiriusFragmentAnnotation::SiriusTargetDecoySpectra>& annotated_spectra)
+  {
+    vector< MetaboTargetedAssay::CompoundTargetDecoyPair > v_cmp_spec;
+    for (const auto& cmp : v_cmpinfo)
+    {
+      for (const auto& spectra : annotated_spectra)
+      {
+        if (cmp.m_ids_id == spectra.target.getName())
```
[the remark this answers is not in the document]
oliveralka: SiriusFragmentAnnotation.cpp
```
    msspectrum_to_fill.setNativeID(native_id);
    msspectrum_to_fill.setName(m_id);
```
The concatenated native_ids and m_ids are stored in the nativeID and name it is probably better for consistency to use metavalues instead?

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedTargetDecoy.cpp
```diff
@@ -0,0 +1,273 @@
...
+        potential_decoy_compound.id = it.decoy_compound_ref;
+        potential_decoy_compound.setMetaValue("decoy", DataValue(1));
+
+        if (TransitionsMap.find(it_target->id) != TransitionsMap.end())
+        {
+          potential_decoy_transitions = TransitionsMap[it_target->id];
+          for (size_t i = 0; i < potential_decoy_transitions.size(); ++i)
+          {
+            potential_decoy_transitions[i]
+                .setNativeID(std::regex_replace(potential_decoy_transitions[i].getNativeID(),
```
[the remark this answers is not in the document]
oliveralka: See above - is there a MetaValue interface for Transition?

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
```diff
@@ -33,97 +33,232 @@
...
+    OpenMS::String native_id = SiriusFragmentAnnotation::extractNativeIDFromSiriusMS_(path_to_sirius_workspace);
+    OpenMS::String m_id = SiriusFragmentAnnotation::extractMIDFromSiriusMS_(path_to_sirius_workspace);
+    SiriusFragmentAnnotation::extractAnnotationFromDecoyFile_(path_to_sirius_workspace, msspectrum_to_fill);
+    msspectrum_to_fill.setNativeID(native_id);
+    msspectrum_to_fill.setName(m_id);
   }
   
   // extract native id from SIRIUS spectrum.ms output file (workspace - compound specific)
   // first native id in the spectrum.ms
   OpenMS::String SiriusFragmentAnnotation::extractNativeIDFromSiriusMS_(const String& path_to_sirius_workspace)
```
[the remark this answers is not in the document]
oliveralka: Renamed and added comments.

oliveralka:
Note: New Windows SIRIUS 4.6.0 MultiOS build does currently not seem to work with the flat third-party structure.

oliveralka:
rebuild jenkins

src/openms/source/ANALYSIS/OPENSWATH/MRMAssay.cpp
```diff
@@ -1097,4 +1100,85 @@ void MRMAssay::detectingTransitionsCompound(OpenMS::TargetedExperiment& exp, int
...
+    vector<TargetedExperiment::Compound> compounds = exp.getCompounds();
+    vector<std::string> descriptions_targets;
+    vector<std::string> descriptions_decoys;
+    vector<std::string> difference_target_decoys;
+    vector<std::pair<std::string, std::string>> reference_decoys;
+    vector<std::string> single_decoy_id;
+
+    for (const auto &it : compounds)
+    {
+      // extract potential target TransitionIds based on the decoy annotation
```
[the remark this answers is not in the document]
oliveralka: "_decoy" for compound description and later transitions ID is set automatically in the MetaboTargetedAssay. 

https://github.com/OpenMS/OpenMS/pull/4725/files

oliveralka:
Windows not working due to missing FIX for SIRIUS start script, but all in all IMHO ready to review

oliveralka:
rebuild jenkins

oliveralka:
Note:
Differences unix vs windows in two tests:
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/bldtst/compiler=msvc-14.2,os_label=win10_64/594/
SiriusAdapter_1/_3

oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

oliveralka:
@jpfeuffer Do you have an idea where the discrepancy between TRAVIS clang and the pyopenms container could come from? 
Travis clang: https://travis-ci.com/github/OpenMS/OpenMS/jobs/485831422
Does not seem to compile correctly.

jpfeuffer:
One is python3, the other python2

oliveralka:
rebuild jenkins

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -591,8 +601,34 @@ class TOPPAssayGeneratorMetabo :
     // use MRMAssay methods for filtering
     MRMAssay assay;
 
-    // filter: min/max transitions
-    assay.detectingTransitionsCompound(t_exp, min_transitions, max_transitions);
+    // sort by highest intensity - filter:  min/max transitions (targets), filter: max transitions (decoys)
```
jpfeuffer, line 604: Have you thought about filtering before you resolve? In case you resolve for something that you would filter anyway (then you could have used the second best).

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -44,99 +43,471 @@
...
+
+    void SiriusAdapterAlgorithm::Project::parameters()
+    {
+      // This will be called internal using the "out_project_space" parameter (SiriusAdapter)
+      // -o, -p, --output, --project-space=<projectSpaceLocation>
+      // Specify project-space to read from and also write to if
+      // nothing else is specified. For compression use the File
+      // ending .zip or .sirius
+
+      // This should not be changes since the mztab writers/readers are depend on the default naming schema on it.
```
jpfeuffer, line 137:
```suggestion
      // This should not be changed since the mztab writers/readers are depending on the default naming schema of it.
```

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -44,99 +43,471 @@
...
+      // --naming-convention=<projectSpaceFilenameFormatter>
+      //    Specify a naming scheme for the  compound
+      // directories ins the project-space. Default %index_%filename_%compoundname
```
jpfeuffer, line 140:
```suggestion
      //    Specify a naming scheme for the compound
      //    directories in the project-space. Default %index_%filename_%compoundname
```

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -44,99 +43,471 @@
...
+
+      // This will be called internal using the preprocessed .ms file
+      // -i, --input=<input>
+      // Input for the analysis. Ths can be either preprocessed mass
+      // spectra in .ms or .mgf file format, LC/MS runs in .mzML/.
+      // mzXml format or already existing SIRIUS project-space(s)
+      // (uncompressed/compressed).
+
+      // --maxmz=<maxMz>
+      // Just consider compounds with a precursor mz lower or equal
```
jpfeuffer, line 150: Remove the comments where they are the same in the description.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -44,99 +43,471 @@
...
+                                    "gnps",
+                                    "zincbio",
+                                    "train",
+                                    "undp",
+                                    "pantcyc",
+                                    "ymdb",
+                                    "keggmine",
+                                    "ecocycmine",
+                                    "ymdbmine",
+                                    "custom",
```
jpfeuffer, line 437: Any help on how to use the custom database?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -384,14 +739,14 @@ namespace OpenMS
         qp.close();
 
         throw Exception::InvalidValue(__FILE__,
-                                      __LINE__, 
-                                      OPENMS_PRETTY_FUNCTION, 
+                                      __LINE__,
+                                      OPENMS_PRETTY_FUNCTION,
                                       "FATAL: External invocation of Sirius failed!",
-                                       "");
+                                      "");
```
jpfeuffer, line 745: Is there a better Exception. I think this will print weirdly.

jpfeuffer (changes requested):

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
oliveralka: These are basically the descriptions from the SIRIUS tool, should I still remove them?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
oliveralka: I can add that to the description. As far as I know, they have to be renamed to custom_n and then copied to a specific directory.

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
```diff
@@ -33,97 +33,233 @@
...
+      if (annotated_spectrum.empty())
+      {
+        continue;
+      }
+      else
+      {
+        // only use spectra over a certain score threshold (0-1)
+        if (double(annotated_spectrum.getMetaValue("score")) >= score_threshold)
+        {
+          // resolve multiple use of the same concatenated nativeids based on the sirius score (used for multiple features/identifications)
```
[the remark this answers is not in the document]
oliveralka: As far as have seen decoys in SIRIUS are currently only generated for the first rank.

src/tests/topp/AssayGeneratorMetabo_decoy_generation_output_both_multids.tsv
```diff
@@ -0,0 +1,9 @@
...
+373.040751999999998	288.946851999999979	1	1	1.0	608.394000000000005					Second,Calonectrin,Millefin,3,17beta-dihydroxy-5,9-dioxo-4,5-9,10-diseco-androsta-1(10),2-dien-4-oicacid,(1E,4Z)-14,15-dihydroxy-8alpha-(2-methylpropanoyloxy)germacra-1(10),4,11(13)-trieno-12,6alpha-lactone,?	C14H17IN2O2	NA	[M+H]+					-1	C8H5IN2O2	-1.0	-1.0	0_Second,Calonectrin,Millefin,3,17beta-dihydroxy-5,9-dioxo-4,5-9,10-diseco-androsta-1(10),2-dien-4-oicacid,(1E,4Z)-14,15-dihydroxy-8alpha-(2-methylpropanoyloxy)germacra-1(10),4,11(13)-trieno-12,6alpha-lactone,?_[M+H]+_608_0	0_0_Second,Calonectrin,Millefin,3,17beta-dihydroxy-5,9-dioxo-4,5-9,10-diseco-androsta-1(10),2-dien-4-oicacid,(1E,4Z)-14,15-dihydroxy-8alpha-(2-methylpropanoyloxy)germacra-1(10),4,11(13)-trieno-12,6alpha-lactone,?_[M+H]+_608_0	0	1	0	1
```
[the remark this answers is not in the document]
oliveralka: Yes, they are harmless and stem from the AMS, when no name existed for a library entry.

src/tests/topp/THIRDPARTY/third_party_tests.cmake
```diff
@@ -79,7 +79,7 @@ OPENMS_FINDBINARY(FIDOCHOOSEPARAMS_BINARY "FidoChooseParameters" "FidoChoosePara
 
 #------------------------------------------------------------------------------
 # Sirius
-OPENMS_FINDBINARY(SIRIUS_BINARY "sirius;sirius-console-64.exe" "Sirius")
+OPENMS_FINDBINARY(SIRIUS_BINARY "sirius;sirius.sh;sirius.bat;sirius-console-64.exe" "Sirius")
```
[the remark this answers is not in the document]
oliveralka: sirius.sh;sirius.bat: used for THIRDPARTY
sirius: unix and macOS stand alone.
sirius.bat: stand alone

I will remove the "sirius-console-64.exe"

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -635,11 +640,12 @@ namespace OpenMS
         }
     }
 
-    // if not mappend information available (e.g. empty featurexml or only a few features)
+    // ms2 spectra without an associated feature based on the provided featureXML
```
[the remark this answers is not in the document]
oliveralka: @jpfeuffer I am not sure what you mean by your comment.

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
```diff
@@ -541,22 +546,22 @@ namespace OpenMS
...
             {
-              for (Size i = 0; i < masstrace_centroid_mz.size(); ++i)
-              {
-                pair<double, double> masstrace_mz_int(masstrace_centroid_mz[i],masstrace_intensity[i]);
-                f_isotopes.push_back(masstrace_mz_int);
-              }
+              pair<double, double> masstrace_mz_int(masstrace_centroid_mz[i], masstrace_intensity[i]);
+              f_isotopes.push_back(masstrace_mz_int);
             }
+          }
```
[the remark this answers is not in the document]
oliveralka: Done

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedTargetDecoy.cpp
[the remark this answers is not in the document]
oliveralka: @jpfeuffer 

SiriusFragmentAnnotation.cpp
```
    msspectrum_to_fill.setNativeID(concat_native_ids);
```

The problem here is that Transition does not have a MetaValue Interface. To keep it consistent the concatenated native ids are stored in the NativeID data structure.

src/openms/include/OpenMS/ANALYSIS/TARGETED/MetaboTargetedAssay.h
```diff
@@ -39,65 +39,83 @@
...
     {
-      std::pair <SiriusMSFile::CompoundInfo, MSSpectrum> compoundspectrumpair;
+    public:
+      int target_index = -1;
+      int decoy_index = -1;
+      double target_mz = 0.0;
+      double target_rt = 0.0;
+      double decoy_mz = 0.0;
+      double decoy_rt = 0.0;
+      int target_file_number;
```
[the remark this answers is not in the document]
oliveralka: Done

src/openms/source/ANALYSIS/OPENSWATH/MRMAssay.cpp
```diff
@@ -1097,4 +1097,85 @@ void MRMAssay::detectingTransitionsCompound(OpenMS::TargetedExperiment& exp, int
...
+      if (it.id.find("decoy") != std::string::npos)
+      {
+        String current_decoy = it.id;
+        String potential_target = std::regex_replace(current_decoy, std::regex("_decoy"), "");
```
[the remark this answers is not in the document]
oliveralka: Yes, you are right, since the description is set as "_decoy" in the MetaboTargetedAssay based on the decoy annotation - one can assume that it is "_decoy" and the regex is not needed.

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -32,26 +32,105 @@
...
 
+#include <regex>
+
 using namespace OpenMS;
 using namespace std;
 
 namespace OpenMS
 {
 
+  // get charge from adduct in standard format [M+H]+ or [M+H]1+
```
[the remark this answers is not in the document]
oliveralka: It is in the header :)

src/openms/source/ANALYSIS/ID/SiriusMSConverter.cpp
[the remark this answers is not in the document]
jpfeuffer: Me neither haha. Forget it.

src/openms/source/ANALYSIS/OPENSWATH/MRMAssay.cpp
```diff
@@ -1055,8 +1056,7 @@ void MRMAssay::detectingTransitionsCompound(OpenMS::TargetedExperiment& exp, int
         for (TransitionVectorType::iterator tr_it = m->second.begin(); tr_it != m->second.end(); ++tr_it)
         {
           ReactionMonitoringTransition tr = *tr_it;
-
-          if ((std::find(LibraryIntensity.begin(), LibraryIntensity.end(), boost::lexical_cast<double>(tr.getLibraryIntensity())) != LibraryIntensity.end()) && tr.getDecoyTransitionType() != ReactionMonitoringTransition::DECOY && j < (Size)max_transitions)
+          if ((std::find(LibraryIntensity.begin(), LibraryIntensity.end(), boost::lexical_cast<double>(tr.getLibraryIntensity())) != LibraryIntensity.end()) && j < (Size)max_transitions)
```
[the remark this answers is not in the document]
jpfeuffer: I could not comment on old lines.

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -331,191 +418,544 @@ namespace OpenMS
...
+      // use the one with the highest occurrence
+      MetaboTargetedAssay::filterBasedOnMolFormAdductOccurrence_(targets);
+      MetaboTargetedAssay::filterBasedOnMolFormAdductOccurrence_(decoys);
+
+      // sort by precursor intensity
+      if (!targets.empty())
+      {
+        sort(targets.begin(),
+             targets.end(),
+             [](const MetaboTargetedAssay &a, const MetaboTargetedAssay &b) -> bool
```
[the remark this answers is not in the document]
jpfeuffer: I.e. define the function and reuse it?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
jpfeuffer: Yeah I know. I just thought, since you copy them word for word into the parameter description, I don't see a need.
The parameter name is also always just the OpenMS name with -- right?

oliveralka: Yes, you are right. I guess then they are not needed there, or at least the description.

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -331,191 +418,544 @@ namespace OpenMS
...
-      double exact_mass_precursor = 0.0;
-      for (auto spec_it = transition_spectrum.begin();
-           spec_it != transition_spectrum.end();
-           ++spec_it)
+  // method to pair compound information (SiriusMSFile) with the annotated target spectrum from Sirius based on the m_id (unique identifier)
+  std::vector< MetaboTargetedAssay::CompoundTargetDecoyPair > MetaboTargetedAssay::pairCompoundWithAnnotatedSpectra(const std::vector<SiriusMSFile::CompoundInfo>& v_cmpinfo,
+                                                                                                                  const std::vector<SiriusFragmentAnnotation::SiriusTargetDecoySpectra>& annotated_spectra)
+  {
+    vector< MetaboTargetedAssay::CompoundTargetDecoyPair > v_cmp_spec;
+    for (const auto& cmp : v_cmpinfo)
```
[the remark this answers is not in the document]
oliveralka: Actually a lot - all spectra SIRIUS was able to annotate. 
In terms of runtime, it should be still far less than SIRIUS needs.
If you have an idea on how to optimize it let me know (maybe also after initial merge).

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp
```diff
@@ -331,191 +418,544 @@ namespace OpenMS
...
+      if (!current_entry.potential_rmts.empty()) // should never be empty
+      {
+        // remove "decoy" tag from compound id for correct mapping
+        if (current_entry.potential_rmts[0].getDecoyTransitionType() ==
+            ReactionMonitoringTransition::DecoyTransitionType::DECOY)
         {
-          explanation_array = transition_spectrum.getStringDataArrays()[0];
-          if (explanation_array.getName() != "explanation")
+          String compoundId = current_entry.potential_cmp.id;
+          compoundId = std::regex_replace(compoundId, std::regex("decoy_"), "");
```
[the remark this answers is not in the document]
oliveralka: Yes, you are right. Will also use string.erase() instead of regex.

src/openms/source/ANALYSIS/TARGETED/MetaboTargetedTargetDecoy.cpp
```diff
@@ -0,0 +1,273 @@
...
+                            decoy_product_masses.begin(),
+                            decoy_product_masses.end(),
+                            std::back_inserter(intersection));
+
+      if (!intersection.empty())
+      {
+        std::transform(intersection.begin(),
+                       intersection.end(),
+                       std::back_inserter(replace_intersection),
+                       [mass_to_add](double d) -> double { return d + mass_to_add; });
```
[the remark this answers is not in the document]
oliveralka: I tried my best :D 

But I think it is ok, even if it happens it should only happen on a limited set of decoys.
Even some kind of overlap is not too bad for scoring, since some of the metabolites might also overlap in some of the transitions.

src/utils/AssayGeneratorMetabo.cpp
oliveralka: Yes, the problem here is that the two data structures used are really different and I think it does not make sense to parse to MRMAssay and then back for the resolution and then to MRMAssay again.

oliveralka:
Before merge: 
We have to merge THIRDPARTY (https://github.com/OpenMS/THIRDPARTY/tree/add/sirius460)
and add it to this PR.

oliveralka:
@jpfeuffer could you please have another look at the "new" void resolve function in the MetaboTargetedAssay?
https://github.com/OpenMS/OpenMS/blob/ebcad6cdb92b2c57fa5c9b4d400812b757d76bb2/src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp#L910

I am not sure if I did that correctly with using the key and operator[] again even with the unordered_map, but I could not think
of any other way to update the entries in the map, since it was split into targets and decoys beforehand.
https://github.com/OpenMS/OpenMS/blob/ebcad6cdb92b2c57fa5c9b4d400812b757d76bb2/src/openms/source/ANALYSIS/TARGETED/MetaboTargetedAssay.cpp#L955

I think the other issues should be fixed now. - Thanks! :)

jpfeuffer:
Do you mean `map_it->second = targetdecoy` or is this not what you want to do?
You can probably also store the iterators in the "empty_keys" (instead of the keys) and remove by iterator but not sure.

jpfeuffer (approved):

jpfeuffer:
Jenkins does not pick up the sirius and AGM tests anymore: http://cdash.openms.de/viewTest.php?onlyfailed&buildid=121031
please check

oliveralka:
Will investigate!

oliveralka:
@jpfeuffer Can it be that the "ENABLE_SIRIUS_TEST" variable was switched off at some point?

jpfeuffer:
Hm but if you compare to recent older runs, the tests were not missing. So I am not sure.

oliveralka:
Yes, I compared them as well:

New:
http://cdash.openms.de/viewTest.php?buildid=119501&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=AssayGeneratorMetabo
Old:
http://cdash.openms.de/viewTest.php?onlyfailed&buildid=121031&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=AssayGeneratorMetabo

I currently not sure why/how that happened.

jpfeuffer:
I think so, too.

jpfeuffer:
Your executables are under bin! Only the root folder of each thirdparty is put into the path.

PR #4726, opened 2020-05-15
Feature/plfq spec count

timosachsenberg:
Added spectral counting as quantification method to ProteomicsLFQ


timosachsenberg:
rebuild jenkins

jpfeuffer:
only pyopenms errors. I will review now.

jpfeuffer:
Does this support calculating ratios/FCs inside ProteinQuantifier, together with all the settings (top3, same peptides, etc.)?

jpfeuffer:
Would be nice to have information on both unique and total peptides. Is this possible already?

timosachsenberg:
yes, this should work exactly the same as only the counts are used to fill abundances.
In addition, I currently annotate (but not yet output in mzTab) the number of unique peptides also if feature intensity-based quant is used.
Maybe you have a hint how to calculate unique/total peptides properly. I think I am annotating and counting the group-unique ones.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.h
```diff
@@ -64,18 +64,23 @@ namespace OpenMS
       /// mapping: fraction -> charge -> sample -> abundance
       std::map<Int, std::map<Int, SampleAbundances>> abundances;
 
+      /// mapping: fraction -> charge -> sample -> abundance
+      std::map<Int, std::map<Int, SampleAbundances>> psm_counts;
```
timosachsenberg, line 68: These are the PSM counts on the peptide level

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.h
```diff
@@ -87,15 +92,19 @@ namespace OpenMS
       /// mapping: peptide (unmodified) -> sample -> abundance
       std::map<String, SampleAbundances> abundances;
 
+      std::map<String, SampleAbundances> psm_counts;
```
timosachsenberg, line 95: these are the ones summarized over all peptidoforms of the same peptide

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -370,21 +388,31 @@ namespace OpenMS
...
     {
       String accession = getAccession_(pep_q.second.accessions,
                                        accession_to_leader);
       OPENMS_LOG_DEBUG << "Peptide id mapped to leader: " << accession << endl;
-      if (!accession.empty()) // proteotypic peptide
+
+      // not enough evidence or mapping to multiple groups
+      if (accession.empty()) continue;
+
+      // proteotypic peptide
```
timosachsenberg, line 401: here it says proteotypic peptide but I guess it is proteogrouptypic ;)

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -427,16 +455,30 @@ namespace OpenMS
...
+      map<UInt64, DoubleList> abundances; // all peptide abundances by sample
+      map<UInt64, DoubleList> psm_counts; // all PSM counts by sample
+      for (const auto & pep : peptides) // for all selected peptides
+      { 
+        for (auto & sa : prot_q.second.abundances[pep]) // copy over abundances
         {
           abundances[sa.first].push_back(sa.second);
         }
+
+        for (auto & sa : prot_q.second.psm_counts[pep]) // copy over psm counts
```
timosachsenberg, line 470: here only the PSMs of the "selected peptides" are counted.

timosachsenberg: e.g. TOP3

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -427,16 +455,30 @@ namespace OpenMS
...
+        {
+          psm_counts[sa.first].push_back(sa.second);
+        }
+      }
+
+      // summarize all peptides of this protein to get map: sample->psm_count
+      // note: we take all for spectral counting (not only the top n ones) and always sum up
```
timosachsenberg, line 477: need to check again if this is the case

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -710,26 +773,41 @@ namespace OpenMS
...
+        const SampleAbundances& total_abundances = q.second.total_abundances;
+        const SampleAbundances& total_psm_counts = q.second.total_psm_counts;
+
+       // TODO: OPENMS_ASSERT(id_group->float_data_arrays.empty(), "Protein group float data array not empty!.");
+        id_group->getFloatDataArrays().resize(2);
         ProteinIdentification::ProteinGroup::FloatDataArray & abundances = id_group->getFloatDataArrays()[0];
         abundances.setName("abundances");        
         abundances.resize(n_samples);
 
+        auto & psm_counts = id_group->getFloatDataArrays()[1];
```
timosachsenberg, line 785: this float data array is added similar to the abundances array

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
timosachsenberg: ok the comment seems to be wrong. psm_count is only filled for the "selected peptides".
Hmm no exactly sure what to do here. If I would do pure spectral counting I would set top n to 0 (all).
Here quantification at feature-intensity level could be TOP3 so the question would be if still all should be used.

jpfeuffer: I agree that you would most likely use "all" for spectral counting but I would leave options open if possible.

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
jpfeuffer: I think proteogrouptypic is the better measure. For singletons proteotypic = proteogrouptypic and for groups proteotypic would be 0 anyway, so the way you are counting here just adds information but does not lose any.

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -214,10 +214,14 @@ namespace OpenMS
         {
           pep_q.second.total_abundances[sa.first] = sa.second;
         }
+        // spectral counting according to the best fraction and charge state only:
```
timosachsenberg, line 217: ok probably need to move this one too

src/tests/topp/ProteomicsLFQ_2_out.csv
```diff
@@ -46,6 +41,7 @@ P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,2,NA,0,L,2,2,3,1,4.102819e06,"BSA2_F1
 P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,3,NA,0,L,2,2,3,1,1.358117e07,"BSA2_F1.mzML"
 P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,2,NA,0,L,3,3,5,1,3.918333e05,"BSA3_F1.mzML"
 P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,3,NA,0,L,3,3,5,1,8.359427e05,"BSA3_F1.mzML"
+P00761|TRYP_PIG,LSSPATLNSR,2,NA,0,L,3,3,5,1,2.406531e05,"BSA3_F1.mzML"
```
timosachsenberg, line 44: this one is new

src/tests/topp/ProteomicsLFQ_3_out.csv
```diff
@@ -59,6 +53,9 @@ P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,2,NA,0,L,2,2,3,1,4.102819e06,"BSA2_F1
 P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,3,NA,0,L,2,2,3,1,1.358117e07,"BSA2_F1.mzML"
 P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,2,NA,0,L,3,3,5,1,3.918333e05,"BSA3_F1.mzML"
 P02769|ALBU_BOVIN,LC(Carbamidomethyl)VLHEK,3,NA,0,L,3,3,5,1,8.359427e05,"BSA3_F1.mzML"
+P00761|TRYP_PIG,LSSPATLNSR,2,NA,0,L,1,1,1,1,5.088512e06,"BSA1_F1.mzML"
```
timosachsenberg, line 56: these 3 are new

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -70,26 +70,28 @@ namespace OpenMS
...
     for (auto & pep : peptides)
     {
-      if (!pep.getHits().empty())
+      if (pep.getHits().empty()) continue;
+      pep.sort(); // TODO: move this out of count peptides
+      const PeptideHit& hit = pep.getHits()[0]; // get best hit
+      PeptideData& data = pep_quant_[hit.getSequence()];
+      data.psm_count++;
+
+      // TODO: why is this needed
```
timosachsenberg, line 86: why is empty element added to fraction?

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -375,16 +388,25 @@ namespace OpenMS
...
+      if (accession.empty()) continue;
+
+      // proteotypic peptide
+      const String peptide = pep_q.first.toUnmodifiedString();
+
+      prot_quant_[accession].psm_count += pep_q.second.psm_count;
+
+      // transfer abundances and counts from peptides->protein
+      // summarize abundances and counts between different peptidoforms       
+      for (auto const & sta : pep_q.second.total_abundances)
```
timosachsenberg, line 402: TODO: should we exclude modified peptides

src/openms/source/FORMAT/MSstatsFile.cpp
```diff
@@ -727,49 +727,30 @@ bool OpenMS::MSstatsFile::isQuantifyable_(
...
+  if (accs.empty()) return false;
 
-  if (accs.empty())
-  {
-    quantifyable = false;
-  }
-  else if (accs.size() > 1)
-      // every prot accession in the set needs to belong to the same indist. group to make this peptide
-      // eligible for quantification
+  if (accs.size() == 1) return true;
```
jpfeuffer, line 732: I don't think this is true. If this accession is not part of the given, considered proteins it should not be quantified.

jpfeuffer: Unless you add a check/precondition that all accessions in evidences occur in the map (which is built from the first run only).

jpfeuffer: Another thing that we probably should define at the beginning of the function, is if every protein in this first run needs to be part of a group (i.e. all singletons are there) or not. I just realized that this also heavily affects the logic.

timosachsenberg: I think PLFQ makes sure this is the case. I think we should still add a consistency check if data is consistent with the inference run.

timosachsenberg: I will add an issue as this has some higher priority

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
timosachsenberg: added an issue

PR #4733, opened 2020-05-19
Update IsoSpec sources to v 2.1.0

michalsta:
Hi,

This is a pull request which updates the sources of IsoSpec in OMS to v2.1.0. Some notes on this:

There are some if constexpr(...) warnings. I do check for availability of if constexpr through __cpp_if_constexpr in platform.h and provide the fallback to (possibly less efficient) plain old if in case it is missing. The trouble is that g++ and clang with -std=c++11 define that macro, provide the if constexpr, and then warn if someone uses it. tl;dr it's a harmless warning

I've changed the default behaviour on thresholding by total probability to perform p_trimming by default. The previous behaviour seems a bit non-obvious, which is even reflected in tests - for example, the number of configurations is tested, when, without p-trimming it is undefined. The more efficient but less intuitive behaviour is still available in the low-level IsoSpecWrapper classes.

The IsoSpec v2.1.0 is still in beta, but the plan is to release it in a few days, and so far it looks to be a "v2.1.0b1 released as final with no modifications". Just wanted to drop that pull req now and possibly get some comments so it's ready to be integrated into OMS once we're fairly happy with testing.


jpfeuffer:
You seem to know your way around ;) Looks great. Let me know if we can help.
travis test failure should be unrelated.
For the g++7 test, there was just an error in the upload to the CDash overview.

There are only some little errors on Windows:
http://cdash.openms.de/viewBuildError.php?buildid=464805

I need to take a closer look myself. Maybe some of their C++ non-standard "features".

michalsta:
Yeah, I'd really appreciate it if you could look into those compile errors. I did manage to run this through MSVC2019, but I can't seem to get MSVC2015 to work (even the 19 took two days...), so, I'd have to try and fix those blindly...

jpfeuffer:
I guess that's it. Looks good, only spurious upload errors. Ready when you are.

michalsta:
Right. I think we're ready for a merge now.

michalsta:
ooo, the CI, it's all green! :D Yes, definitely ready :)

src/tests/class_tests/openms/source/TheoreticalSpectrumGenerator_test.cpp
```diff
@@ -591,23 +591,15 @@ START_SECTION(([EXTRA] test isotope clusters for all peak types))
     263.15390,
 
     // 405: POS: 78.54256367545 INT: 0.921514272689819
-    79.04108117545,
     79.04424117545,
```
timosachsenberg, line 594: why is there this difference between old / new?

src/tests/class_tests/openms/source/IsoSpec_test.cpp
```diff
@@ -521,7 +514,7 @@ START_SECTION(( void IsoSpecTotalProbWrapper::run() ))
 
   // human insulin
   IsotopeDistribution iso_result3 = IsoSpecTotalProbWrapper(EmpiricalFormula("C520H817N139O147S8"), total_prob, do_trim).run();
-  TEST_EQUAL(iso_result3.size(), 19616);
+  TEST_EQUAL(iso_result3.size(), 19615);
```
timosachsenberg, line 517: now it's correct?

src/tests/class_tests/openms/source/FineIsotopeDistribution_test.cpp
```diff
@@ -77,7 +77,7 @@ START_SECTION(( IsotopeDistribution run(const EmpiricalFormula&) const ))
 
   // simple way of getting an IsotopeDistribution using total probability
   test_id = ef.getIsotopeDistribution(FineIsotopePatternGenerator(0.01, true, false));
-  TEST_EQUAL(test_id.size(), 6)
+  TEST_EQUAL(test_id.size(), 3)
```
timosachsenberg, line 80: why now 3?

src/tests/class_tests/openms/source/IsoSpec_test.cpp
```diff
@@ -282,24 +282,17 @@ END_SECTION
...
-  bool do_trim = true;
 
-  IsoSpecTotalProbGeneratorWrapper ITPW(EmpiricalFormula("C6H12O6"), total_prob, do_trim);
+  IsoSpecTotalProbGeneratorWrapper ITPW(EmpiricalFormula("C6H12O6"), total_prob);
   TEST_EQUAL(compare_generator_to_reference(ITPW, fructose_expected_oms, -1), true);
 
-  IsoSpecTotalProbGeneratorWrapper ITPW2(EmpiricalFormula("C6H12O6"), total_prob, do_trim);
-  TEST_EQUAL(generator_length(ITPW2), 17);
+  IsoSpecTotalProbGeneratorWrapper ITPW2(EmpiricalFormula("C6H12O6"), total_prob);
+  TEST_EQUAL(generator_length(ITPW2), 2548); // Total prob is a hint, if we keep extracting configurations
```
timosachsenberg, line 290: because there is not trimming?

src/tests/class_tests/openms/source/TheoreticalSpectrumGenerator_test.cpp
michalsta: This one is because we do p_trim by default now. When computing to a total probability threshold, IsoSpec computes some extra configurations above what's necessary to cover the requested probability, and without p_trim those configurations are reported (the rationale being, well, since we already computed them, we might as well report them). The problem is that exactly which excess configurations are included is undefined, the only guarantee is that sum of reported probabilities is going to be above the selected p. Which means that the exact number of confs of the reported envelope is undefined, and might change as we improve the algorithm (and this number shouldn't be tested for).

I thought that this is a bit counter-intuitive, and now we do p-trim by default. With p-trim we do have the guarantee that the reported set is optimal, that is, it is the smallest set of configurations that sums up to more than the requested probability.

michalsta: And yes, with p-trim the size should not change (much... see below) as we modify the algorithm.

src/tests/class_tests/openms/source/IsoSpec_test.cpp
michalsta: Actually, in a sense both values are sorta correct. We compute the probabilities of configurations a bit differently now, compared to what used to be. Mathematically it's equivalent, but we do different rounding errors during calculations now. In this case it resulted in one less configuration being needed to cross the requested probability threshold. In either case the rounding errors are much less than the tolerances on element abundances given by IUPAC, so it shouldn't be important.

src/tests/class_tests/openms/source/FineIsotopeDistribution_test.cpp
michalsta: Again, because we're doing p_trimming by default

src/tests/class_tests/openms/source/IsoSpec_test.cpp
michalsta: Yes. The way it used to be is that TotalProbGenerator used to be kind of a hybrid between a generator and something that stored configurations and performed p-trimming. This led to it being somewhat memory inefficient, and unclear in design. Now the two parts are separated and the generator just goes over the entire configuration space in an order that's close enough to sorted to be able to perform the p-trim efficiently (while not exactly sorted to avoid the nlogn complexity), and another piece of code accumulates the configurations, terminates the search in the appropriate moment and performs the p-trim.

timosachsenberg:
Thanks a lot! Great PR

Issue #4684, opened 2020-04-28
Unify precision across all writers


jpfeuffer:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/DATAACCESS/MSDataWritingConsumer.cpp#L56

Here mzML gets written with the old? way of using std streams. Not boost karma.

cbielow:
we have a possible solution for this. Please wait :)

PR #4736, opened 2020-05-20
Improved writing to files by implementing the class FastOStream

akriese:
As writing to a file takes up some time in almost all TOPP-Tools, we thought, that this could be improved by avoiding the construction and destruction of sentries, as normally there is only one thread writing output to a file. For that, we built a new class FastOStream, that references a given std::ostream and writes to it using the 'os.rdbuf()->sputn(s, n)' method. Overloads for the operator<<() include different string-like types, arithmetic types (using boost::karma for conversions), DataValue and vector.

We implemented the class in multiple files (all judged to be relevant by @cbielow) and replaced the old ostreams.

Tests / testfiles had to be changed as boost::karma's conversions are different compared to the std::ios_base's. (as discussed in #4684 and #4627)

Note that this class has undefined behaviour if accessed by multiple threads at once.

The speedup of storing files lies at up to 15% on linux and up to 30% in Windows.
E.g. (IDFileConverter, Windows): 
Before changes:
`Progress of 'Storing...':
-- done [took 11.16 s (CPU), 11.16 s (Wall)]`
After changes:
`Progress of 'Storing...':
-- done [took 7.95 s (CPU), 7.96 s (Wall)]`


src/openms/source/ANALYSIS/OPENSWATH/OpenSwathTSVWriter.cpp
```diff
@@ -87,203 +89,203 @@ namespace OpenMS
...
           {
-            group_label = id;
+            full_peptide_name += pep.sequence[loc];
           }
-
-          // If a protein is present, take the first one
-          String protein_name = "";
-          if (!pep.protein_refs.empty() )
+          // C-terminal and N-terminal modifications may be at positions -1 or pep.sequence
+          for (auto modification : pep.modifications)
```
jpfeuffer, line 156: const-ref?

jpfeuffer:
Unfortunately, this seems to break the pyOpenMS tests.
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/pyOMSDckr/2777/testReport/junit/projectroot/pyOpenMS/pyopenms_unittest_test000_py/

Let us know if you need help building pyOpenMS. But maybe the error is obvious to you.

akriese:
The pyopenms-error most certainly is connected to the FOS using boost::karma's String-construction. The generated numbers are longer than what ostream used to create. So, the output String is longer than expected... We're gonna fix the test

cbielow:
any progress here?

timosachsenberg:
no progress and diverged significantly. feel free to reopen an updated PR

PR #4737, opened 2020-05-20
[FIX] PLFQ check score types again after reloading IDs.

jpfeuffer:
And add trypticity to search engine parameters.


src/openms/include/OpenMS/METADATA/ProteinIdentification.h
```diff
@@ -251,6 +251,16 @@ namespace OpenMS
       AVERAGE,
       SIZE_OF_PEAKMASSTYPE
     };
+
+    /// Trypticity
```
timosachsenberg, line 255: Would it make sense to have this neun only once?

timosachsenberg: neun=enum

timosachsenberg:
hmm or would it better to use something generic instead of NumberTrypticTermini
NumberEnzymaticTermini?

src/openms/include/OpenMS/METADATA/ProteinIdentification.h
jpfeuffer: IMHO this should go into the enzyme structures, but this is too much effort right now to support it there.

src/openms/include/OpenMS/CHEMISTRY/DigestionEnzyme.h
```diff
@@ -53,6 +53,19 @@ namespace OpenMS
   {
   public:
 
+    /// Terminus Specificity
+    ///@todo Currently this is just here for encapsulation
```
timosachsenberg, line 57: which I think is totally fine ;)

timosachsenberg: but can't we just use the one from EnzymaticDigestion?

src/openms/source/ANALYSIS/ID/PeptideIndexing.cpp
```diff
@@ -51,20 +51,22 @@ using namespace std;
...
 
-    defaults_.setValue("enzyme:specificity", EnzymaticDigestion::NamesOfSpecificity[0], "Specificity of the enzyme."
-                                                                                               "\n  '" + EnzymaticDigestion::NamesOfSpecificity[0] + "': both internal cleavage sites must match."
-                                                                                                                                                     "\n  '" + EnzymaticDigestion::NamesOfSpecificity[1] + "': one of two internal cleavage sites must match."
-                                                                                                                                                                                                           "\n  '" + EnzymaticDigestion::NamesOfSpecificity[2] + "': allow all peptide hits no matter their context. Therefore, the enzyme chosen does not play a role here");
+    defaults_.setValue("enzyme:specificity", "", "Specificity of the enzyme. Default: deduce from input."
+      "\n  '" + EnzymaticDigestion::NamesOfSpecificity[2] + "': both internal cleavage sites must match."
```
timosachsenberg, line 62: hmm on the left we have [0] und here we have[2] could we use the enum as index directly?

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
```diff
@@ -80,11 +80,11 @@ namespace OpenMS
 
   EnzymaticDigestion::Specificity EnzymaticDigestion::getSpecificityByName(const String& name)
   {
-    for (Size i = 0; i < SIZE_OF_SPECIFICITY; ++i)
+    for (Size i = 0; i < SPEC_UNKNOWN; ++i)
```
timosachsenberg, line 83: SIZE_OF_SPECIFICITY is a safer pattern (until we switch to enum classes)

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -2997,11 +2999,24 @@ namespace OpenMS
...
+        {
+          minCharge = upit->second.toString().toInt();
+        }
+        else if (upit->first == "MaxCharge")
+        {
+          maxCharge = upit->second.toString().toInt();
+        }
+        else if (upit->first == "NumTolerableTermini")
+        {
+          sp.nr_termini = static_cast<EnzymaticDigestion::Specificity>(upit->second.toString().toInt());
```
timosachsenberg, line 3012: nice fixes

src/openms/source/METADATA/ProteinIdentification.cpp
```diff
@@ -141,7 +142,8 @@ namespace OpenMS
            fragment_mass_tolerance_ppm == rhs.fragment_mass_tolerance_ppm &&
```
timosachsenberg, line 142: I prefer std::tie constructs if possible

src/openms/source/METADATA/ProteinIdentification.cpp
```diff
@@ -164,7 +166,8 @@ namespace OpenMS
         this->fragment_mass_tolerance_ppm != sp.fragment_mass_tolerance_ppm ||
```
timosachsenberg, line 166: same here

src/openms/include/OpenMS/CHEMISTRY/DigestionEnzyme.h
jpfeuffer: yes, forgot to remove

src/openms/source/CHEMISTRY/EnzymaticDigestion.cpp
jpfeuffer: Add it or use SIZE as unknown?

timosachsenberg: I would add SIZE as last and have UNKNOWN extra. At least that's how others do it

src/tests/topp/QualityControl_1_out.mzTab
```diff
@@ -15,7 +15,7 @@ MTD	software[2]-setting[4]	fragment_mass_tolerance_unit:Da
 MTD	software[2]-setting[5]	precursor_mass_tolerance:10.0
 MTD	software[2]-setting[6]	precursor_mass_tolerance_unit:ppm
 MTD	software[2]-setting[7]	enzyme:Trypsin
-MTD	software[2]-setting[8]	nr_enzymatic_termini:10
+MTD	software[2]-setting[8]	nr_enzymatic_termini:3
```
timosachsenberg, line 18: 3?

jpfeuffer: yes, I renamed everything again now anyway.

PR #4742, opened 2020-05-22
[FIX] term. mods with specific origin in Comet

### LINKED ISSUE #4741: Error in IDMerger
Hi all,

I have been attempting to test Epifany and downloaded the example workflows for Knime.

I used the example mzMLs (iPRG2016 data B sample) as suggested from the https://www.openms.de/comp/epifany/ page. After running the CometAdaptor successfully the IDMerger showed the error below:

```
ERROR IDMerger             0:6        Failing process stdout: [Warning: Parameters file version (2.4.0-HEAD-2019-01-29) does not match the version of this tool (2.5.0)., Your current parameters are still valid, but there might be new valid values or even new parameters. Upgrading the INI might be useful., Error: Unexpected internal error (the value 'Cannot convert string to peptide modification. No terminal modification matches to term specificity and origin.' was used but is not valid; .(Gln->pyro-Glu)ETMGPIPPVHPC(Carbamidomethyl)TIM(Oxidation)GIYK)]
ERROR IDMerger             0:6        Failing process stderr: []
ERROR IDMerger             0:6        Return code: 8
ERROR IDMerger             0:6        Execute failed: Failed to execute node IDMerger
```

The error is caused by the variable modification `Gln -> pyroGlu (Q)` that I added to the parameters, removing that specific modification solves the problem, but it would be good to have the ability to select it in the future. 

Thank you,

Giuseppe

jpfeuffer:
fixes #4741

### DIFF
--- a/src/topp/CometAdapter.cpp
+++ b/src/topp/CometAdapter.cpp
@@ -365,7 +365,10 @@ protected:
       //TODO support agglomeration of Modifications to same AA. Watch out for nc_term value then.
       if (mod.getTermSpecificity() == ResidueModification::C_TERM)
       {
-        residues = "c";
+        if (mod.getOrigin() == "X")
+        {
+          residues = "c";
+        } // else stays mod.getOrigin()
         term_distance = 0;
         // Since users need to specify mods that apply to multiple residues/terms separately
         // 3 and -1 should be equal for now.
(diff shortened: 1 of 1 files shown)


timosachsenberg:
looks good

PR #4744, opened 2020-05-22
Add a missing const operator[] for IsotopeDistribution

nateglasser:
I wasn't able to pass a const IsotopeDistribution& to functions and read their peaks because the only operator[] was non-const. This adds the missing const operator[] to IsotopeDistribution.


src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsotopeDistribution.h
```diff
@@ -221,6 +221,8 @@ namespace OpenMS
     //@{
     /// operator which access a cell of the distribution and wraps it in SpectrumFragment struct
     Peak1D& operator[](const Size& index){ return distribution_[index];}
+    /// const operator which access a cell of the distribution and wraps it in SpectrumFragment struct
```
cbielow, line 224: `const operator to access a cell....` for extra bonus points (also for the non const version)

cbielow:
Failed test is unrelated

PR #4747, opened 2020-05-23
SIMDe library 

feldroop:
SIMD support for future performance optimizations:

- Added the SIMDe library to the source tree. 
  Details: https://github.com/nemequ/simde

- Added an SSE implementaion of the `StringManager::AppendASCII` function from the `XMLHandler`.
  This implementation showed a significant speedup in our microbenchmarks, but no effect inside an actual OpenMS application. Therefore it is only an example of what future optimizations might look like.


cbielow:
ready for merge. 
Any comments?

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -43,6 +43,7 @@
 #include <OpenMS/FORMAT/VALIDATORS/MzMLValidator.h>
 #include <OpenMS/INTERFACES/IMSDataConsumer.h>
 #include <OpenMS/SYSTEM/File.h>
+#include <simde/x86/sse2.h>
```
cbielow, line 46: this include is not required ?!

cmake/compiler_flags.cmake
```diff
@@ -52,7 +52,8 @@ if (CMAKE_COMPILER_IS_GNUCXX)
     -Wno-long-long 
     -Wno-unknown-pragmas
     -Wno-unused-function
-    -Wno-variadic-macros)
+    -Wno-variadic-macros
+    -msse2)
```
cbielow, line 56: this should not be required, since `For the x86-32 compiler, you must use -march=cpu-type, -msse or -msse2 switches to enable SSE extensions and make this option effective. For the x86-64 compiler, these extensions are enabled by default. `
(see https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/x86-Options.html).

At the moment, we only use SSE2 instructions, right?
But what if someone wants to add SSE3 code using  SIMDe? Do we need to add `-msse2 -msse3 -mssse3 -msse4 -msse4.1 -msse4.2 ` etc here manually?
Maybe add a comment what is required.

mr-c:
ping @Felix-Droop

Felix-Droop:
This can be closed, if noone is currently interested in using SIMDe in OpenMS. (which I assume is the case)

mr-c:
@Felix-Droop I'm not an openms user, but I do contribute to SIMDe and I know some of the OpenMS people; may I take this over?

timosachsenberg:
Hi. Thanks for your contribution. We are of course interested in SIMD but I wasn't aware that any input is required at this point. It is also marked as a draft.

Felix-Droop:
> @Felix-Droop I'm not an openms user, but I do contribute to SIMDe and I know some of the OpenMS people; may I take this over?

Sure!

> Hi. Thanks for your contribution. We are of course interested in SIMD but I wasn't aware that any input is required at this point. It is also marked as a draft.

Just to clarify, my comment wasn't meant in an angry way :D
When I did this in an internship with Chris, we chose to leave this PR stale until the next effort for SIMD optimization in OpenMS happens, because the one optimization I actually implemented did not provide a benefit.

timosachsenberg:
I will close this PR for now as it is not actively worked on. We consider reopening in the future.

PR #4750, opened 2020-05-25
Caching of validated CV terms in MzMLHandler + Alternative to getAllChildTerms()

akriese:
While measuring the performance of the FileConverter, we noticed, that the programm spends a lot of time in the recursive function ControlledVocabulary::getAllChildTerms() which builds a set of Strings.
As this set is not actually needed completely, an alternative function was then created to iterate over all children recursively without building the set. That function (iterateAllChildren()) takes a lambda-function which shall return a bool to decide if to quit the recursion.
The new function was applied where possible.

Furthermore, the search for an existing CVTerm in SematicValidator::locateTerm_() is very expensive and also done repeatedly for the same terms and paths. So, a map<pair<Path, CV-name>, bool> was created for MzMLHandler::validateCV_(), which is looked up to see if a term-path-combination was located already. Only if not included in the map, that combination is determined with locateTerm_().

These changes lead to a constant 50% speedup of MzMLFile::store().


akriese:
rebuild jenkins

akriese:
rebuild jenkins

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -3377,14 +3377,22 @@ namespace OpenMS
 
     bool MzMLHandler::validateCV_(const ControlledVocabulary::CVTerm& c, const String& path, const Internal::MzMLValidator& validator) const
     {
-      SemanticValidator::CVTerm sc;
+      // static map<<Path, c.id>, isValid> for storing, if a cvterm is valid in the given path
+      const auto it = cached_terms_.find(std::make_pair(path, c.id));
```
cbielow, line 3381: can you write a bit of local documentation to make clearer why the cache is useful?

PR #4754, opened 2020-05-26
Cherry-picked commits for feature

poshul:
Added utility function MSChromatogram to allow merging the ChromatogramPeaks from two MSChromatograms. Function is useful for summing intensity of multiple isotope traces in an isotopic envelope.

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
+++ b/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
@@ -424,7 +424,7 @@ public:
 
       @param other a reference to the MSChromatogram to take ChromatogramPeaks from
     */
-    void mergePeaks(MSChromatogram& other);
+    void mergePeaks(MSChromatogram& other, bool add_meta=true);
 
 protected:
 
--- a/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
+++ b/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
@@ -422,9 +422,11 @@ public:
       @note Make sure BOTH chromatograms are sorted with respect to RT. Otherwise the result is
       undefined.
 
+      MZ of the destination MSChromatogram remains unchanged. If add_meta is true a metavalue "merged_with" is added with the MZ of the source MSChromatogram
+
       @param other a reference to the MSChromatogram to take ChromatogramPeaks from
     */
-    void mergePeaks(MSChromatogram& other, bool add_meta=true);
+    void mergePeaks(MSChromatogram& other, bool add_meta=false);
 
 protected:
 
--- a/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
+++ b/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
@@ -416,6 +416,16 @@ public:
 
     ///@}
 
+    /**
+      @brief Adds all the ChromatogramPeaks from another MSChromatogram and updates the metadata to indicate a merge
+
+      @note Make sure BOTH chromatograms are sorted with respect to RT. Otherwise the result is
+      undefined.
+
+      @param other a reference to the MSChromatogram to take ChromatogramPeaks from
+    */
+    void mergePeaks(MSChromatogram& other);
+
 protected:
 
     /// Name
--- a/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
+++ b/src/openms/include/OpenMS/KERNEL/MSChromatogram.h
@@ -422,6 +422,8 @@ public:
       @note Make sure BOTH chromatograms are sorted with respect to RT. Otherwise the result is
       undefined.
 
+      @note Peak level metadata stored in float_array string_array and int_array of the destination MSChromatogram is not guaranteed to be correct after merging
+
       MZ of the destination MSChromatogram remains unchanged. If add_meta is true a metavalue "merged_with" is added with the MZ of the source MSChromatogram
 
       @param other a reference to the MSChromatogram to take ChromatogramPeaks from
--- a/src/openms/source/CONCEPT/Constants.cpp
+++ b/src/openms/source/CONCEPT/Constants.cpp
@@ -154,6 +154,9 @@ namespace OpenMS
 
     namespace UserParam
     {
+      // User parameter name for the M/Z of other chromatograms which have been merged into this one
+      const std::string MERGED_WITH = "merged_with";
+
       // User parameter name for precursor mz error in ppm
       const std::string PRECURSOR_ERROR_PPM_USERPARAM = "precursor_mz_error_ppm";
 
--- a/src/openms/source/CONCEPT/Constants.cpp
+++ b/src/openms/source/CONCEPT/Constants.cpp
@@ -155,7 +155,7 @@ namespace OpenMS
     namespace UserParam
     {
       // User parameter name for the M/Z of other chromatograms which have been merged into this one
-      const std::string MERGED_WITH = "merged_chromatogram_mzs";
+      const std::string MERGED_CHROMATOGRAM_MZS = "merged_chromatogram_mzs";
 
       // User parameter name for precursor mz error in ppm
       const std::string PRECURSOR_ERROR_PPM_USERPARAM = "precursor_mz_error_ppm";
--- a/src/openms/source/CONCEPT/Constants.cpp
+++ b/src/openms/source/CONCEPT/Constants.cpp
@@ -155,7 +155,7 @@ namespace OpenMS
     namespace UserParam
     {
       // User parameter name for the M/Z of other chromatograms which have been merged into this one
-      const std::string MERGED_WITH = "merged_with";
+      const std::string MERGED_WITH = "merged_chromatogram_mzs";
 
       // User parameter name for precursor mz error in ppm
       const std::string PRECURSOR_ERROR_PPM_USERPARAM = "precursor_mz_error_ppm";
--- a/src/openms/source/KERNEL/MSChromatogram.cpp
+++ b/src/openms/source/KERNEL/MSChromatogram.cpp
@@ -416,23 +416,36 @@ void MSChromatogram::clear(bool clear_meta_data)
   }
 }
 
+
+OpenMS::MSChromatogram::Iterator set_sum_similar_union(OpenMS::MSChromatogram::Iterator first1,
+                    OpenMS::MSChromatogram::Iterator last1,
+                    OpenMS::MSChromatogram::Iterator first2,
+                    OpenMS::MSChromatogram::Iterator last2,
+                    OpenMS::MSChromatogram::Iterator d_first)
+{
+    for (; first1 != last1; ++d_first) {
+        if (first2 == last2)
+            return std::copy(first1, last1, d_first);
+        if ( round( ( first1->getRT() ) * 1000) == round( (first2->getRT()  * 1000 ) ) )
+        {
+          *d_first = *first1++;
+          d_first->setIntensity(d_first->getIntensity() + first2->getIntensity());
+          first2++;
+        }
+        else if (first2->getRT() < first1->getRT()) {
+            *d_first = *first2++;
+        } else {
+            *d_first = *first1++;
+        }
+    }
+    return std::copy(first2, last2, d_first);
+}
+
+
 void MSChromatogram::mergePeaks(MSChromatogram& other)
 {
   vector<ChromatogramPeak> temp;
-  std::merge(this->begin(), this->end(), other.begin(), other.end(), std::back_inserter(temp), ChromatogramPeak::RTLess());
-  this->clear(false);
-  for(iterator it = temp.begin(); it != temp.end(); it++)
-  {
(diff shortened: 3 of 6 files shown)


poshul:
Retry DeepCode

jpfeuffer:
Dont worry about deepcode if it does not work. Maybe because of the cherry-pick..

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -415,3 +415,24 @@ void MSChromatogram::clear(bool clear_meta_data)
     integer_data_arrays_.clear();
   }
 }
+
+void MSChromatogram::mergePeaks(MSChromatogram& other)
+{
+  vector<ChromatogramPeak> temp;
+  std::merge(this->begin(), this->end(), other.begin(), other.end(), std::back_inserter(temp), ChromatogramPeak::RTLess());
```
timosachsenberg, line 422: would be interesting to benchmark if appending the other vector and a sort is faster. @cbielow what do you think?

timosachsenberg: I think I would take the merge code and instead of merging the element add it on top if they are equal in RT. that way you don't need the extra copying.

timosachsenberg: or even closer: https://en.cppreference.com/w/cpp/algorithm/set_union
this merges two ranges but keeps only one element if two are equal. I did not check but this sounds like it could be adapted to create the summed up version?

poshul: Yes, I think I can do something clever with a modified set_union interpretation. I'll let you know.

poshul: Got it working.

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -448,4 +454,4 @@ void MSChromatogram::mergePeaks(MSChromatogram& other)
   temp.reserve((this->end() - this->begin()) + (other.end() - other.begin()) );
   ContainerType::assign(temp.begin(), set_sum_similar_union(this->begin(), this->end(), other.begin(), other.end(), temp.begin()));
   this->setMetaValue("merged_with",String(other.getMZ()));
```
timosachsenberg, line 456: I would make this optional. There could be many merges and also setting metaValues is super expensive

poshul: Added.

src/openms/include/OpenMS/KERNEL/MSChromatogram.h
```diff
@@ -424,7 +424,7 @@ namespace OpenMS
 
       @param other a reference to the MSChromatogram to take ChromatogramPeaks from
     */
-    void mergePeaks(MSChromatogram& other);
+    void mergePeaks(MSChromatogram& other, bool add_meta=true);
```
timosachsenberg, line 427: I would make this false as default

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -448,10 +448,23 @@ OpenMS::MSChromatogram::Iterator set_sum_similar_union(OpenMS::MSChromatogram::I
...
-void MSChromatogram::mergePeaks(MSChromatogram& other)
+void MSChromatogram::mergePeaks(MSChromatogram& other, bool add_meta)
 {
   vector<ChromatogramPeak> temp;
   temp.reserve((this->end() - this->begin()) + (other.end() - other.begin()) );
   ContainerType::assign(temp.begin(), set_sum_similar_union(this->begin(), this->end(), other.begin(), other.end(), temp.begin()));
-  this->setMetaValue("merged_with",String(other.getMZ()));   
+  if (add_meta)
+  {
+    if (this->getMetaValue("merged_with") == DataValue::EMPTY)
```
timosachsenberg, line 458: please add a telling name to Constants::

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -448,10 +448,23 @@ OpenMS::MSChromatogram::Iterator set_sum_similar_union(OpenMS::MSChromatogram::I
...
+    {
+      this->setMetaValue("merged_with",DoubleList(other.getMZ()));
```
timosachsenberg, line 460: is there a list constructor from single double?

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -448,10 +448,23 @@ OpenMS::MSChromatogram::Iterator set_sum_similar_union(OpenMS::MSChromatogram::I
...
+    }
+    else
+    {
+      DoubleList ls = this->getMetaValue("merged_with").toDoubleList();
```
timosachsenberg, line 464: did you check if you can just append to the existing one? this is a lot of copying

src/openms/include/OpenMS/KERNEL/MSChromatogram.h
timosachsenberg: document if m/z stays the same

timosachsenberg: also document what happens to
  MSChromatogram::FloatDataArray float_array;
  MSChromatogram::StringDataArray string_array;
  MSChromatogram::IntegerDataArray int_array;

timosachsenberg: just figured that add_meta could be misunderstood that meta data is added from the other chromatogram

poshul: Documented.

src/openms/source/KERNEL/MSChromatogram.cpp
poshul: Done.

src/openms/source/KERNEL/MSChromatogram.cpp
poshul: Not the correct one. I've fixed this.

src/openms/source/CONCEPT/Constants.cpp
```diff
@@ -154,6 +154,9 @@ namespace OpenMS
 
     namespace UserParam
     {
+      // User parameter name for the M/Z of other chromatograms which have been merged into this one
+      const std::string MERGED_WITH = "merged_with";
```
timosachsenberg, line 158: this string constant is probably too general for the narrow task it is used for and there might be some potential name clashes (i.e. it's not that unrealistic that a user already stored a merged_with meta value with different meaning).
So maybe something like "merged_chromatogram_mzs" ?

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -415,3 +415,52 @@ void MSChromatogram::clear(bool clear_meta_data)
     integer_data_arrays_.clear();
   }
 }
+
+
+OpenMS::MSChromatogram::Iterator set_sum_similar_union(OpenMS::MSChromatogram::Iterator first1,
```
cbielow, line 420: should be `setSumSimilarUnion`

src/tests/class_tests/openms/source/MSChromatogram_test.cpp
```diff
@@ -1070,6 +1078,27 @@ START_SECTION(([MSChromatogram::MZLess] bool operator()(const MSChromatogram &a,
...
+  a.push_back(p3);
+  b.push_back(p5);
+  c.push_back(p1);
+  c.push_back(p2);
+  c.push_back(p4);
+  a.sortByPosition();
+  b.sortByPosition();
+  c.sortByPosition();
+  a.mergePeaks(b, true);
+  DoubleList dl = DoubleList();
```
cbielow, line 1095: for brevity: `DoubleList dl = { b.getMZ() };`

src/tests/class_tests/openms/source/MSChromatogram_test.cpp
```diff
@@ -1070,6 +1078,27 @@ START_SECTION(([MSChromatogram::MZLess] bool operator()(const MSChromatogram &a,
 }
 END_SECTION
 
+START_SECTION(void mergePeaks(MSChromatogram & other) )
+{
+  MSChromatogram a, b, c;
```
cbielow, line 1083: does `MSChromatogram a{p1, p2, p3};` work?

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -415,3 +415,52 @@ void MSChromatogram::clear(bool clear_meta_data)
...
+
+
+void MSChromatogram::mergePeaks(MSChromatogram& other, bool add_meta)
+{
+  vector<ChromatogramPeak> temp;
+  temp.reserve((this->end() - this->begin()) + (other.end() - other.begin()) );
+  ContainerType::assign(temp.begin(), set_sum_similar_union(this->begin(), this->end(), other.begin(), other.end(), temp.begin()));
+  if (add_meta)
+  {
+    DoubleList ls = DoubleList();
```
cbielow, line 458: `= DoubleList()` is not required

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -415,3 +415,52 @@ void MSChromatogram::clear(bool clear_meta_data)
...
+OpenMS::MSChromatogram::Iterator set_sum_similar_union(OpenMS::MSChromatogram::Iterator first1,
+                    OpenMS::MSChromatogram::Iterator last1,
+                    OpenMS::MSChromatogram::Iterator first2,
+                    OpenMS::MSChromatogram::Iterator last2,
+                    OpenMS::MSChromatogram::Iterator d_first)
+{
+    for (; first1 != last1; ++d_first) {
+        if (first2 == last2)
+            return std::copy(first1, last1, d_first);
+        if ( round( ( first1->getRT() ) * 1000) == round( (first2->getRT()  * 1000 ) ) )
```
cbielow, line 429: can you document a bit what is going on here?

src/openms/source/CONCEPT/Constants.cpp
```diff
@@ -154,6 +154,9 @@ namespace OpenMS
 
     namespace UserParam
     {
+      // User parameter name for the M/Z of other chromatograms which have been merged into this one
+      const std::string MERGED_WITH = "merged_chromatogram_mzs";
```
cbielow, line 158: is there a more descriptive name?
`MERGED_WITH` is a bit unspecific.
Maybe `MERGED_WITH_MZ_USERPARAM`?

src/openms/source/KERNEL/MSChromatogram.cpp
timosachsenberg: should this helper be private?

cbielow: its a free function in a cpp... private by design :)

src/openms/source/CONCEPT/Constants.cpp
timosachsenberg: or even MERGED_CHROMATOGRAM_MZS

src/tests/class_tests/openms/source/MSChromatogram_test.cpp
poshul: nope.

src/openms/source/CONCEPT/Constants.cpp
poshul: done.

src/openms/source/KERNEL/MSChromatogram.cpp
poshul: I've added some documentation for the function. Let me know if there is more that I should add.

src/openms/source/KERNEL/MSChromatogram.cpp
poshul: Fixed

src/tests/class_tests/openms/source/MSChromatogram_test.cpp
poshul: fixed

jpfeuffer:
This probably broke the nightlies on Win in Debug mode http://cdash.openms.de/viewTest.php?onlyfailed&buildid=473271

src/openms/source/KERNEL/MSChromatogram.cpp
```diff
@@ -415,3 +415,53 @@ void MSChromatogram::clear(bool clear_meta_data)
...
+        }
+        else // Otherwise choose whichever peak is smaller and add it to the output list
+        {
+          if (first2->getRT() < first1->getRT()) 
+          {
+            *d_first = *first2++;
+          }
+          else 
+          {
+            *d_first = *first1++;
```
timosachsenberg, line 444: do we need a ++first2; here?

timosachsenberg:
I think the implementation might have some other problems (e.g., if one of the vectors contain the same RT).
Will give it a try.

PR #4755, opened 2020-05-26
[FIX] Comet pepXML modification parsing

### LINKED ISSUE #4751: IDFileConverter to IDMerger error from Comet search
HI, 

When running a Comet search with version `2019.01 rev. 5` there is an issue with the conversion of the results from Comet's pep.xml to idXML, which makes the merging of the idXML files error out.

The id conversion only issues warnings:
```
Progress of 'Loading...':
Non-fatal error while loading '/home/ubuntu/data/protein_inference/comet_data//1a7d100b-b01e-441e-8acf-2a78fd92c151.pep.xml': No modification description given. Trying to define by modification mass.
Non-fatal error while loading '/home/ubuntu/data/protein_inference/comet_data//1a7d100b-b01e-441e-8acf-2a78fd92c151.pep.xml': Modification '147.035384999999991' is not uniquely defined by the given data. Using 'Oxidation (M)' to represent any of 'Oxidation (M), oxidation to L-methionine sulfoxide (M), oxidation to L-methionine (R)-sulfoxide (M), oxidation to L-methionine (S)-sulfoxide (M)'.
Non-fatal error while loading '/home/ubuntu/data/protein_inference/comet_data//1a7d100b-b01e-441e-8acf-2a78fd92c151.pep.xml': Modification '170.105527999999993' is not uniquely defined by the given data. Using 'Acetyl (K)' to represent any of 'Acetyl (K), acetate labeling reagent light form (N-term and K) (K), N-acetylated L-lysine (K)'.
-- done [took 01:09 m (CPU), 01:09 m (Wall)] --
Progress of 'Storing...':
IDFileConverter took 01:30 m (wall), 01:30 m (CPU), 2.11 s (system), 01:28 m (user).
<Non-fatal error while loading '/home/ubuntu/data/protein_inference/comet_data//1a7d100b-b01e-441e-8acf-2a78fd92c151.pep.xml': No modification description given. Trying to define by modification mass.> occurred 5 times
```

The IDMerger instead fails with this error:
```
Error: Unexpected internal error (the value 'Cannot convert string to peptide modification. No terminal modification matches to term specificity and origin.' was used but is not valid; .(Gln->pyro-Glu)KDAQQRALQK)
```

This looked a bit odd for a pyro-Glu modification at a n-terminal K. In Comet's pep.xml file  I think this is actually a peptide matching a decoy protein:

file name: `23487e81-c94b-4d4a-b044-a4723e9746af.pep.xml`

```
    217    <search_hit hit_rank="4" peptide="KDAQQRALQK" peptide_prev_aa="R" peptide_next_aa="K" protein="DECOY_sp|Q6PL18|ATAD2_HUMAN" num_tot_proteins="1" num_matched_ions="2" tot_num_ions="36" calc_neutral_    217 pep_mass="1167.636020" massdiff="-0.000140" num_tol_term="2" num_missed_cleavages="2" num_matched_peptides="986">
    218     <modification_info modified_peptide="KDAQQRALQ[111]K">
    219      <mod_aminoacid_mass position="9" mass="111.032029" variable="-17.026549" source="param"/>
    220     </modification_info>
    221     <search_score name="xcorr" value="0.495"/>
    222     <search_score name="deltacn" value="0.001"/>
    223     <search_score name="deltacnstar" value="0.000"/>
    224     <search_score name="spscore" value="17.2"/>
    225     <search_score name="sprank" value="3"/>
    226     <search_score name="expect" value="1.82E+01"/>
    227    </search_hit>
```
The peptide has a "modification" on the 9th amini acid Q.

The same peptide in the idXML looks like this:
file name: `23487e81-c94b-4d4a-b044-a4723e9746af.pep.xml.idXML`
```
 345263                         <PeptideHit score="18.199999999999999" sequence=".(Gln-&gt;pyro-Glu)KDAQQRALQK" charge="3" aa_before="R" aa_after="K" protein_refs="PH_24" >
 345264                                 <UserParam type="string" name="MS:1002258" value="2"/>
 345265                                 <UserParam type="string" name="MS:1002259" value="36"/>
 345266                                 <UserParam type="string" name="num_matched_peptides" value="986"/>
 345267                                 <UserParam type="float" name="MS:1002252" value="0.495"/>
 345268                                 <UserParam type="float" name="MS:1002253" value="0.001"/>
 345269                                 <UserParam type="float" name="MS:1002254" value="0.0"/>
 345270                                 <UserParam type="float" name="MS:1002255" value="17.199999999999999"/>
 345271                                 <UserParam type="float" name="MS:1002256" value="3.0"/>
 345272                                 <UserParam type="float" name="MS:1002257" value="18.199999999999999"/>
 345273                         </PeptideHit>
``` 
where the protein it matches to is a decoy and is:
```
     61                         <ProteinHit id="PH_24" accession="DECOY_sp|Q6PL18|ATAD2_HUMAN" score="0.0" sequence="" >
     62                         </ProteinHit>
```

This modification seems to have been moved to the N-terminus of the peptide sequence in the idXML and it's invalid for IDMerger.

Needless to say, removing the `Gln->pyro-Glu (Q)` from the list of modification solves the issue.

If needed:
The pep.xml file is [here](https://drive.google.com/file/d/1HQ7NwEkfdkD5s-o4m80aJ0t0YSuL_hCV/view?usp=sharing)

the idXML file is [here](https://drive.google.com/file/d/1EQAFdQMHfkEUvhD6LiHDoL_oNJSXbMO-/view?usp=sharing)

jpfeuffer:
AKA Avoid putting terminal modifications to internal AAs when reading pepXML. Instead put them as unknown delta_mass.
Also parses decoys from comet search. (Note: this is not well-tested with downstream tools).
Should fix #4751 though

Limitations: 
1) pepXML only encodes the location of the peptide in the first protein
2) depending on the settings, comet might not report all parent proteins
3) only works for decoys from comet, not pepxml in general

### DIFF
--- a/src/openms/source/CHEMISTRY/AASequence.cpp
+++ b/src/openms/source/CHEMISTRY/AASequence.cpp
@@ -1019,6 +1019,7 @@ namespace OpenMS
       }
       else
       {
+        //TODO why are we allowing Protein Term here?
         aas.n_term_mod_ = terminalResidueHelper(mod_db, 'n', true, str, mod, String(*next_aa));
         return mod_end;
       }
@@ -1034,6 +1035,7 @@ namespace OpenMS
     else if (specificity == ResidueModification::C_TERM)
     {
       ModificationsDB* mod_db = ModificationsDB::getInstance();
+      //TODO why are we allowing Protein Term here?
       aas.c_term_mod_ = terminalResidueHelper(mod_db, 'c', true, str, mod, res);
       return mod_end;
     }
@@ -1253,7 +1255,9 @@ namespace OpenMS
 
     // ----------------------------------- 
     // Dealing with an unknown modification
-    // ----------------------------------- 
+    // -----------------------------------
+
+    const ResidueModification* new_mod = ResidueModification::createUnknownFromMass(mass, delta_mass, specificity, residue);
 
     // Notes on mass calculation: AASequence::getMonoWeight uses DiffMonoMass
     // for its calculation of C/N-terminal modification mass and it uses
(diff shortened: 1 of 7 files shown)


src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1034,6 +1035,7 @@ namespace OpenMS
     else if (specificity == ResidueModification::C_TERM)
     {
       ModificationsDB* mod_db = ModificationsDB::getInstance();
+      //TODO why are we allowing Protein Term here?
```
timosachsenberg, line 1038: I think because it can occur at the Protein Term but is not specific to that.

jpfeuffer: Hmm ok we dont have knowledge about the protein here. Actually it is only used with "Any X-term", never with Protein N-term.
We should allow an additional optional parameter in the fromString method in case we know this peptide is at a protein terminus. Otherwise we may get Protein term mods in the middle of a protein, when no other mod matches.

jpfeuffer:
fixes #4751 partly

src/openms/source/CHEMISTRY/AASequence.cpp
timosachsenberg: yeah I think this would be important

jpfeuffer:
What?? Why are there differences on linux?

PR #4756, opened 2020-05-27
Hotfix PRECONDITION after optimization for QTCluster

Felix-Droop:
### Fixed the FeatureLinkerunlabeledQT after the nightly tests failed.

The reason for the failed test (`OPENMS_PRECONDITION`) was the follwing:

The `tmp_element_mapping` of the `QTCluster` class used to be a pointer to a map. When we implemented the hot-cold-splitting we decided that two indirections were unnecessary. We also wanted to increase the code quality by removing some `new`s and `delete`s. We then changed the `tmp_element_mapping` to be just a map inside the `BulkData` class.

In the preconditions there were a lot of checks whether `tmp_element_mapping` is or isn't a nullptr. This was the case whenever the cluster was in a finalized state (nullptr) or not (pointer to map). We changed these checks to ask whether it is empty (finalized) or not (not finalized). This does not work, because the `tmp_element_mapping` can be empty during a finalized state for certain inputs. In these cases it used to be a pointer to an empty map, but now we don't have the pointer anymore and therefore can't perform this check.

We believe these checks were made because one wanted to make absolutely sure that no nullptr can be dereferenced. The new code does not use a pointer anymore and is less dangerous. Furthermore these checks are redundant, because there is a bool `finalized_` which monitors the exact same functionality with preconditions.

I removed two preconditions at locations where `finalized_` is also checked and I believe the check is not necessary anymore.

I built the DEBUG version on Linux and the tests do not fail anymore. I also fixed a memory leak in the QTClusterFinder_test.cpp.

### Reference

QTCluster.cpp at the last commit before our PR (https://github.com/OpenMS/OpenMS/pull/4721):

https://github.com/OpenMS/OpenMS/blob/94790282ebf0b65932da4a34a03f16eace253f77/src/openms/source/DATASTRUCTURES/QTCluster.cpp

### DIFF
--- a/src/tests/class_tests/openms/source/QTClusterFinder_test.cpp
+++ b/src/tests/class_tests/openms/source/QTClusterFinder_test.cpp
@@ -74,7 +74,7 @@ START_SECTION((static BaseGroupFinder* create()))
 	BaseGroupFinder* base_ptr = nullptr;
 	base_ptr = QTClusterFinder::create();
   TEST_NOT_EQUAL(base_ptr, base_nullPointer);
-  if (base_ptr != nullptr) delete base_ptr;
+  delete base_ptr;
 }
 END_SECTION
 
--- a/src/tests/class_tests/openms/source/QTClusterFinder_test.cpp
+++ b/src/tests/class_tests/openms/source/QTClusterFinder_test.cpp
@@ -74,6 +74,7 @@ START_SECTION((static BaseGroupFinder* create()))
 	BaseGroupFinder* base_ptr = nullptr;
 	base_ptr = QTClusterFinder::create();
   TEST_NOT_EQUAL(base_ptr, base_nullPointer);
+  if (base_ptr != nullptr) delete base_ptr;
 }
 END_SECTION
 
--- a/src/openms/source/DATASTRUCTURES/QTCluster.cpp
+++ b/src/openms/source/DATASTRUCTURES/QTCluster.cpp
@@ -146,8 +146,6 @@ namespace OpenMS
     OPENMS_PRECONDITION(distance <= data_->max_distance_,
         "Distance cannot be larger than max_distance")
     // collect_annotations_ implies tmp_neighbors_ != NULL, 
-    OPENMS_PRECONDITION(!collect_annotations_ || !tmp_neighbors_.empty(), 
-        "Initialize the cluster first before adding elements")
 
     Size map_index = element->getMapIndex();
 
@@ -462,8 +460,6 @@ namespace OpenMS
 
   void QTCluster::finalizeCluster()
   {
-    OPENMS_PRECONDITION(!data_->tmp_neighbors_.empty(),
-        "Try to finalize QTCluster that was not initialized")
     OPENMS_PRECONDITION(!finalized_,
         "Try to finalize QTCluster that was not initialized")
 
--- a/src/openms/source/DATASTRUCTURES/QTCluster.cpp
+++ b/src/openms/source/DATASTRUCTURES/QTCluster.cpp
@@ -145,7 +145,6 @@ namespace OpenMS
     // ensure we only add compatible peptide annotations
     OPENMS_PRECONDITION(distance <= data_->max_distance_,
         "Distance cannot be larger than max_distance")
-    // collect_annotations_ implies tmp_neighbors_ != NULL, 
 
     Size map_index = element->getMapIndex();
 


src/tests/class_tests/openms/source/QTClusterFinder_test.cpp
```diff
@@ -74,6 +74,7 @@ START_SECTION((static BaseGroupFinder* create()))
 	BaseGroupFinder* base_ptr = nullptr;
 	base_ptr = QTClusterFinder::create();
   TEST_NOT_EQUAL(base_ptr, base_nullPointer);
+  if (base_ptr != nullptr) delete base_ptr;
```
cbielow, line 77: no need for `if ...`, 
its enough to do `delete base_ptr;` since delete will check for nullptr internally.

timosachsenberg:
I get a floating point exception on one of my branches (release mode). Will try to reproduce it.

timosachsenberg:
ok I am able to reproduce it on my branch for some specific data.
Might be related to comparing identifications at Cluster.
Will dig into it later

timosachsenberg:
Progress of 'reading input':
-- done [took 59.94 s (CPU), 59.94 s (Wall)] --
Progress of 'Linking features':
  56.86 %
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff602e06a in OpenMS::QTCluster::update(std::vector<OpenMS::QTCluster::Element, std::allocator<OpenMS::QTCluster::Element> > const&) () from /home/sachsenb/OMS/OpenMS-build/lib/libOpenMS.so
Missing separate debuginfos, use: debuginfo-install glib2-2.56.1-2.el7.x86_64 glibc-2.17-196.el7.x86_64 keyutils-libs-1.5.8-3.el7.x86_64 krb5-libs-1.15.1-8.el7.x86_64 libcom_err-1.42.9-9.el7.x86_64 libicu-50.1.2-15.el7.x86_64 libselinux-2.5-6.el7.x86_64 openssl-libs-1.0.2k-16.el7_6.1.x86_64 pcre-8.32-15.el7_2.1.x86_64

timosachsenberg:
For reference: command line argument was
/home/sachsenb/OMS/OpenMS-build/bin/FeatureLinkerUnlabeledQT -in debug_fraction_1_1.featureXML debug_fraction_1_2.featureXML debug_fraction_1_3.featureXML debug_fraction_1_4.featureXML debug_fraction_1_5.featureXML debug_fraction_1_6.featureXML debug_fraction_1_7.featureXML debug_fraction_1_8.featureXML debug_fraction_1_9.featureXML debug_fraction_1_10.featureXML debug_fraction_1_11.featureXML debug_fraction_1_12.featureXML -out test.consensusXML -algorithm:use_identifications -algorithm:distance_MZ:max_difference 10.0 -algorithm:distance_MZ:unit ppm

timosachsenberg:
Test data https://drive.google.com/file/d/16xpIKc2y5AgRv6PlMQeMSXfcKGtV3smt/view?usp=sharing

timosachsenberg:
If I merge against develop that now contains these changes I get:
Progress of 'Linking features':
  56.86 %               Error: Unexpected internal error (finalized_ Cannot perform operation on cluster that is not finalized)
in debug mode

Felix-Droop:
Thank you for the help and sorry for the problems. I will look into it now.

timosachsenberg:
No problem. This would be really great. I just checked again and this seems to only happens on this branch. https://github.com/OpenMS/OpenMS/pull/4726
I did not mess with the Linker so I don't know why this breaks. But as it fails in QtCluster update I suspect that it might be related to your original PR

timosachsenberg:
gdb now reports a segfault in for (const auto& removed_element : removed)
```
211       {
212         OPENMS_PRECONDITION(finalized_,
213             "Cannot perform operation on cluster that is not finalized")
214
215         // check if the cluster center was removed:
216         for (const auto& removed_element : removed)
217         {
218           // If center point was removed, then we are done and no more work is
219           // required
220           if (removed_element.feature == data_->center_point_)
```

Felix-Droop:
A few more questions to make sure I understand all this correctly:

1. The exact same command with the same data did not crash at OpenMS:develop?
2. Did this error occur after the hotfix yesterday or did it possibly exist already before?

We have a little history of segfaults with this optimization. The function `QTClusterFinder::removeTopFromElementMapping` should fix them (see also the comments in this function). It is plausible that your test data triggers a case that we did not have in mind and that also slipped through all the tests (after we found the error, we should maybe add another test?). But  if it works on OpenMS:develop (see question 1.), I am confused because it should also fail, shoulnd't it?

Felix-Droop:
If there is a segfault in update(), most probably the cluster, on which update() is called, was destructed (i.e. popped from the heap) and its id is still in the element mapping. The error in the debug mode is only explainable as UB: if the object is destructed, then we probably have a hidden segfault when we ask for finalized_, but the memory is somehow still owned by us and is all 0s, therefore finalized_ is false. I do not have too much knowledge about this, but is this a possible explanation? Does the debug mode use an address sanitizer, which could catch something like this?

timosachsenberg:
1. No but I remember to get a failed precondition when compiling in debug mode. 
2. I check also before the hotfix but *think* I got the same results (sorry I was switching branches a lot as I was working on other stuff and just did the debugging on the side)

This sounds plausible and might also explain why I had a bit of a hard time reproducing it. Can you download the zip file I linked here and try to run it using the command line I posted here?
Maybe with develop and my branch? I hope I am not asking for too much.

By default we don't do address sanitizer

timosachsenberg:
ok now I get the sefault on newest develop.
So maybe not 100% predictable / initialized or freed memory issue

Felix-Droop:
I will download the data and try to find the problem. Might take me some time though...

PR #4757, opened 2020-05-27
Rename vscode folder and init during gitpod startup

### LINKED ISSUE #4735: .vscode in source tree
Regarding this commit: https://github.com/OpenMS/OpenMS/commit/a08d97d09725d4e21023eb0ead3a71e3a0aa3db4

It might be necessary to put the `.vscode` folder into the source tree to make gitpod work. I don't know anything about it.

I am working with VSCode locally and it's a little bit annoying. First of all, the config files are not working for my setup/workspace. Then, if I change something, it gets tracked by git. Since I don't want to commit these changes, changing branches for example is harder than it should be.

Maybe there is an easy way for me to work around this that I am missing. Otherwise I hope we can find a solution that works for everyone! The above commit also deleted `.vscode` from the `.gitignore` file. I guess that's not an option.

jpfeuffer:
fixes #4735


timosachsenberg:
looks good

PR #4759, opened 2020-05-28
add OPENMS_GIT_LC_DATE defaults

cbielow:
when using GIT_TRACKING=Off

Otherwise configuring with PyOpenMS will fail here
https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/CMakeLists.txt#L406
with 
```
CMake Error at src/pyOpenMS/CMakeLists.txt:406 (string):
  string sub-command REPLACE requires at least four arguments.


CMake Error at src/pyOpenMS/CMakeLists.txt:409 (list):
  list GET given empty list```


timosachsenberg:
thanks!

PR #4761, opened 2020-05-28
Updated GNPSExport module

abisarvepalli:
- better memory management with caching mzml files in the heap

- updated binning and determining similarity across msbins when generating merged_spectra


src/topp/GNPSExport.cpp
```diff
@@ -80,296 +150,387 @@ class TOPPGNPSExport : public TOPPBase
...
-      mzml_file.load(mzml_file_path, map);
 
-      ms_maps.push_back(map);
-    }
+    //-------------------------------------------------------------
+    // preprocessing: allocate memory
+    //-------------------------------------------------------------
+    // max_msmap_cache = std::min(max_msmap_cache, static_cast<int>(mzml_file_paths.size()));
+    int max_msmap_cache = static_cast<int>(mzml_file_paths.size());
+    MzMLFile* mzml_files = new MzMLFile[max_msmap_cache];
```
cbielow, line 331: please don't do this!
There is no need to have more than one `MzMLFile`. One is enough to load any number of mzMLs into the internal MSExperiment datastructure.

src/topp/GNPSExport.cpp
```diff
@@ -80,296 +150,387 @@ class TOPPGNPSExport : public TOPPBase
...
+    MSExperiment* specs_list = new MSExperiment[max_msmap_cache];
```
cbielow, line 332: please avoid manual heap allocations and use `vector<MSExperiment>`

src/topp/GNPSExport.cpp
```diff
@@ -80,296 +150,387 @@ class TOPPGNPSExport : public TOPPBase
...
+
+    map<int,int> msmaps_cached; // <K, V> = <map_index, mzml_files index>
```
cbielow, line 334: you can also get rid of this mapping, by simply making sure the `vector<MSExperiment>` is valid at the index you want to access. Its a simple 1:1 mapping as far as I understand it, since the order of input mzMLs is expected to match the map_index.

src/topp/GNPSExport.cpp
```diff
@@ -80,296 +150,387 @@ class TOPPGNPSExport : public TOPPBase
...
-          return a.first.first < b.first.first;
-        });
+          int map_index = pepts_iter->first;          
+
+          // load missing MzMLFile files + MSExperiment specs
+          if (msmaps_cached.find(map_index) == msmaps_cached.end())
+          { 
+            // printf("\t\tregistering new msmap\n");
+            MzMLFile mzmlfile;
+            mzml_files[num_msmaps_cached] = mzmlfile;
```
cbielow, line 446: no need to do this. When creating the array (or vector) it is default initialized

cbielow:
forgot to say: Welcome!
... and thanks for your contribution to OpenMS :)

src/topp/GNPSExport.cpp
abisarvepalli: was concerned whether or not we needed multiple mzml_files, thanks for the clarification and will make this optimization asap!

src/topp/GNPSExport.cpp
abisarvepalli: i felt that it would be easier allocate memory for all the MSExperiments in the heap, and the reason that I maintained it as an array rather than wrapping it in a vector was to allow for quicker queries.

Instead, I can opt to wrap it as a vector<MSExperiment>(max_msmap_cache) as I may allow a flag to adjust how many maps to cache as it does consume memory when running the module on larger datasets that require 1000s of mzml files.

Would it be passable to commit this change to opting to wrap the array as a vector?

src/topp/GNPSExport.cpp
abisarvepalli: Good point! In the case that I want to check if the MSExperiment has been loaded, is it simply calling the (MSExperiment).empy()  to determine if the spectrum has been loaded?

src/topp/GNPSExport.cpp
abisarvepalli: Also because the consensusXML may not necessarily reference all mzml files passed in

abisarvepalli:
Thanks! Love to be part of an active community of researchers + developers 😄

src/topp/GNPSExport.cpp
cbielow: Vector also allocates its data on the heap. So speedwise, there is no difference whatsoever.
If you preallocate the vector using the same size as the array, it behaves exactly the same as the array.

cbielow: > 
> 
> Also because the consensusXML may not necessarily reference all mzml files passed in

This argues pro vector actually, if you use push_back instead of preallocating and `operator[]`.

But an empty MSExperiment is nothing to worry about memory-wise, so i'd go for the latter.

src/topp/GNPSExport.cpp
cbielow: yes. The only case where this goes wrong if your mzML file is indeed empty as well. If that's the case however, its also very fast to load each time you access it.

If you want the really safe way, maintain a vector<bool> of the same size as the number of experiments, and indicate there if it has been loaded already.

cbielow:
before merging, please squash/rebase your commits to avoid pulling in the large testfiles which were present in the intermediate commits. This would make the Git repo considerably larger, for no good reason.
Thanks!

abisarvepalli:
From what I've interpreted:

- It seems that the apple, ubuntu, and windows tests had not failed due to faulty GNPSExport tests
- The Tracis-CI build failed due to IDBoostGraph.cpp_cppcheck_test failing 'potential for loop copies'

abisarvepalli:
From what I've interpreted:

- It seems that the apple, ubuntu, and windows tests had not failed due to GNPSExport tests
- The Tracis-CI build failed due to IDBoostGraph.cpp_cppcheck_test failing 'potential for loop copies'

src/topp/GNPSExport.cpp
```diff
@@ -42,24 +43,91 @@
...
+      Eligible MS/MS scans have a pairwise cosine similarity with the MS/MS scan of highest precursor intensity above the Cosine Similarity Treshold.
+	    The fragment ions of merged MS/MS scans are binned in m/z (or Da) range defined by the Binning width parameter.    
+      .
+	  -# Most intense: most_intense - For each consensusElement, the GNPSExport will output the most intense MS/MS scan (with the highest precursor ion intensity) as consensus MS/MS spectrum.
+      .    
+
+Note that mass accuracy and the retention time window for the pairing between MS/MS scans and a LC-MS feature
+or consensusElement is defined at the IDMapper tool step.
+
+A representative OpenMS-GNPS workflow would sequencially use these OpenMS TOPP tools:
```
cbielow, line 96: typo `sequentially`

some other typos above (`ran` -> `run`) ...

src/topp/GNPSExport.cpp
```diff
@@ -42,24 +43,91 @@
...
-  8. Run the GNPSExport to export an .MGF
-  9. Upload the .MGF file on http://gnps.ucsd.edu and follow the instructions here:
+  2. Run the FeatureFinderMetabo tool on the mzML files.
+  3. Run the IDMapper tool on the featureXML and mzML files.
+  4. Run the MapAlignerPoseClustering tool on the featureXML files.
+  5. Run the MetaboliteAdductDecharger on the featureXML files.
+  6. Run the FeatureLinkerUnlabeledKD tool or FeatureLinkerUnlabeledQT, on the featureXML files and output a consensusXML file.
+  8. Run the FileFilter on the consensusXML file to keep only consensusElements with at least MS/MS scan (peptide identification).
+  9. Run the GNPSExport on the "filtered consensusXML file" to export an .MGF file.
+  10. Run the TextExport on the "filtered consensusXML file" to export an .TXT file.
```
cbielow, line 105: do you mean `TextExporter`?

cbielow: you can add hyperlinks to the above TOPP tools by using e.g. `@ref TOPP_FileFilter` instead of just `FileFilter`

src/topp/GNPSExport.cpp
```diff
@@ -42,24 +43,91 @@
...
+  11. Upload your files to GNPS and run the Feature-Based Molecular Networking workflow. Instructions are here:
 https://ccms-ucsd.github.io/GNPSDocumentation/featurebasedmolecularnetworking/
 
+The GitHub for that ProteoSAFe workflow and an OpenMS python wrappers is available here:
+https://github.com/Bioinformatic-squad-DorresteinLab/openms-gnps-workflow
+
+An online version of the OpenMS-GNPS pipeline for FBMN running on CCMS server (http://proteomics.ucsd.edu/) is available on GNPS:
+https://ccms-ucsd.github.io/GNPSDocumentation/featurebasedmolecularnetworking-with-OpenMS
+
+GNPS (Global Natural Products Social Molecular Networking, https://gnps.ucsd.edu/ProteoSAFe/static/gnps-splash2.jsp)
```
cbielow, line 115: again, URL not working.
Please check all URLs...

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+  static constexpr double DEF_MERGE_BIN_SIZE = static_cast<double>(BinnedSpectrum::DEFAULT_BIN_WIDTH_HIRES);
+
+  static constexpr double DEF_PREC_MASS_TOL = 0.5;
+  static constexpr bool DEF_PREC_MASS_TOL_ISPPM = false;
+
+  static constexpr double DEF_PEPT_CUTOFF = 5;
+  static constexpr double DEF_MSMAP_CACHE = 50;
+
+
+  void writeMS2BlockToFile(
```
cbielow, line 176: OpenMS coding convention: https://github.com/OpenMS/OpenMS/wiki/Coding-conventions#underscores
use trailing underscores for private things (variables and functions),
e.g. `writeMS2BlockToFile` --> `writeMS2BlockToFile_`

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+    const int& scan_index,
+    const String& feature_id,
+    const int& feature_charge,
+    const String& feature_mz,
+    const String& spec_index,
+    const String& feature_rt
+  ) 
+  {
+    output_file << "BEGIN IONS" << "\n";
+    output_file << "OUTPUT=" << output_type << "\n";
```
cbielow, line 189: hint/optional:
for better readability you can do
```
output_file << "BEGIN IONS" << "\n"
            << "OUTPUT=" << output_type << "\n";
```
i.e. omit repeated usage of the stream and indent accordingly

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+      if (ms2.second > 0)
+      {
+        output_file << ms2.first << "\t" << ms2.second << "\n";
+      }
+    }
+
+    output_file << "END IONS" << "\n\n";
+  }
+
+  void generateMSMSSpectrumBins(
```
cbielow, line 211: documentation is still missing

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+    // generate new spectrum
+    vector<double> mz_merged;
+    vector<double> intensity_merged;
+
+    // double last_mz = numeric_limits<double>::min();
+    double last_mz = sorted_mz_int_pairs[0].first;    
+    double sum_mz = 0;
+    int sum_intensity = 0;
+    double count = 0;
+    for (const auto mz_int : sorted_mz_int_pairs)
```
cbielow, line 226: this is probably where the auto-loop copy warning comes from.
use `const auto&` to avoid it

src/topp/GNPSExport.cpp
```diff
@@ -80,296 +150,387 @@ class TOPPGNPSExport : public TOPPBase
...
+  void getElementPeptideIdentificationsByElementIntensity(
+    const ConsensusFeature& feature,
+    vector<pair<int,double>>& sorted_element_maps,
+    vector<pair<int,int>>& pepts
+  )
+  {
+    for (pair<int,double> element_pair : sorted_element_maps)
+    {
+      int element_map = element_pair.first;
+      vector<PeptideIdentification> feature_pepts = feature.getPeptideIdentifications();
```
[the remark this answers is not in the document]
cbielow: todo

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+    const vector<pair<int,double>>& sorted_element_maps,
+    vector<pair<int,int>>& pept_indices
+  )
+  {
+    for (const pair<int,double> element_pair : sorted_element_maps)
+    {
+      int element_map = element_pair.first;
+
+      vector<PeptideIdentification> feature_pepts = feature.getPeptideIdentifications();
+      for (const PeptideIdentification pept_id : feature_pepts)
```
cbielow, line 288: this is another loop copy error in the checker (I guess)
use `const PeptideIdentification& pept_id`

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
-    setValidStrings_("output_type", ListUtils::create<String>("full_spectra,merged_spectra"));
+    registerStringOption_("output_type", "<choice>", "most_intense", "specificity of mgf output information", false);
+    setValidStrings_("output_type", ListUtils::create<String>("merged_spectra,most_intense"));
+
+    addEmptyLine_();
 
-    registerDoubleOption_("precursor_mz_tolerance", "<num>", DEF_PRECURSOR_MZ_TOLERANCE, "Tolerance mz window for precursor selection", false);
-    registerDoubleOption_("precursor_rt_tolerance", "<num>", DEF_PRECURSOR_RT_TOLERANCE, "Tolerance rt window for precursor selection", false);
+    // registerIntOption_("msmap_cache", "<num>", DEF_MSMAP_CACHE, "Number of msmaps that can be cached during export for optimized performance", false, true);
+    registerIntOption_("peptide_cutoff", "<num>", DEF_PEPT_CUTOFF, "Number of most intense peptides to consider per consensus element; '-1' to consider all identifications", false, true);
```
cbielow, line 337: you can use `setMinInt_()` and `setMaxInt_()` for limit the allowed range of values.
E.g. -2 does not make sense here, so go for [-1, inf) (i.e. just specify the minimum)

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+    registerDoubleOption_("ms2_bin_size", "<num>", DEF_MERGE_BIN_SIZE, "Bin size (Da) for fragment ions when merging ms2 scans", false, false);
```
cbielow, line 338: set lower limit of 0? (`setMinFloat_()`)

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
-
-      // store "mz rt" information from each scan
-      stringstream scans_output;
-      scans_output << setprecision(2) << setfill('0') << fixed;
-
-      // determining charge and most intense feature for header
-      BaseFeature::ChargeType charge = feature.getCharge();
-      for (ConsensusFeature::HandleSetType::const_iterator feature_iter = feature.begin();
-      feature_iter != feature.end(); ++feature_iter)
+      const ConsensusFeature feature = consensus_map[cons_i];
```
cbielow, line 393: efficiency: use `&` to avoid copy

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+
+        // printf("map_index %d\n", map_index);
+
+        // MSExperiment specs = specs_list->at(map_index);
+        MSExperiment& specs = getSpectraAtIndex(mzml_file_paths, specs_list, map_index);
+        // if (specs.empty())
+        // {
+        //   mzml_file.load(mzml_file_paths[map_index], specs);
+        // }
+        auto spec = specs[spec_index];
```
cbielow, line 430: `auto&`
copying a spectrum is not free

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
+      {
+        output_file << ms2.first << "\t" << ms2.second << "\n";
+      }
+    }
+
+    output_file << "END IONS" << "\n\n";
+  }
+
+  void generateMSMSSpectrumBins(
+    const vector<pair<double,int>>& sorted_mz_int_pairs,
```
cbielow, line 212: why do you go through the trouble of creating a vector of pairs from an MSSpectrum (which is esstentially the same thing)? Just sort the spectrum using https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1MSSpectrum.html#a86f7b200260209865e730a2e4fcaa628 and iterate over the spectrum:
```
for (const auto& peak : spec)
{
  peak.getMz() ...
  peak.getIntensity()...
}
```

saves code and runtime

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
-            double cosine_sim = bsca(binned_highest_int, binned_spectrum);
-            // OPENMS_LOG_DEBUG << cosine_sim << " >= " << cos_sim << endl;
+          BinnedSpectralContrastAngle bsca;
+          double cos_sim = bsca(binned_highest_int, binned_spectrum);
 
-            // compare calculated cosine sim to binned highest int
-            if (cosine_sim >= cos_sim)
+          if (cos_sim >= cos_sim_threshold)
+          {
+            for (auto spec_iter = test_spec.begin(); spec_iter < test_spec.end(); spec_iter++)
```
cbielow, line 478: again: no need for this conversion. Just pass a spectrum to `generateMSMSSpectrumBins`

src/topp/GNPSExport.cpp
```diff
@@ -79,297 +147,368 @@ using namespace std;
...
-    // writing output
-    //-------------------------------------------------------------
-    ofstream output_file(out);
-    progress_logger.startProgress(0, 1, "writing mgf file");
-    output_file << output_stream.str();
-    progress_logger.endProgress();
+    // delete / dealloc mem resources
     output_file.close();
 
+    mzml_file_paths.clear();
```
cbielow, line 508: no need to clear at the end of a function.
The destructor will take care of it automatically.

src/topp/GNPSExport.cpp
abisarvepalli: I am attempting to merge multiple MSSpectrum scans. I browsed the namespace and was unable to find a way to append/extend a MSSpectrum with more scans. What would you suggest to be most optimal?

oliveralka:
@abisarvepalli Do you have additional questions?

oliveralka:
rebuild jenkins

lfnothias:
@abisarvepalli would you be able to resolve these couple things so that can be part of the next OpenMS release ?

jpfeuffer:
rebuild jenkins

src/topp/GNPSExport.cpp
```diff
@@ -410,13 +357,12 @@ class TOPPGNPSExport : public TOPPBase
     addEmptyLine_();
 
     // registerIntOption_("msmap_cache", "<num>", DEF_MSMAP_CACHE, "Number of msmaps that can be cached during export for optimized performance", false, true);
+    // setMinInt_("msmap_cache", 0);
     registerIntOption_("peptide_cutoff", "<num>", DEF_PEPT_CUTOFF, "Number of most intense peptides to consider per consensus element; '-1' to consider all identifications", false, true);
     setMinInt_("peptide_cutoff", 1);
```
oliveralka, line 362: should that be -1?

FabianAicheler:
The test files GNPSExport_mzml1.mzML and GNPSExport_mzml2.mzML are ~31MB each, could/should those be smaller?

lfnothias:
@abisarvepalli can you please use these smaller mzML files instead ? You can simply rename them as the other previous files. The link expires in 7 days. https://we.tl/t-zYIWPzoWD3.

abisarvepalli:
> @abisarvepalli can you please use these smaller mzML files instead ? You can simply rename them as the other previous files. The link expires in 7 days. https://we.tl/t-zYIWPzoWD3.

yep went ahead and downloaded the updated file, will reflect it in the next commit

src/topp/GNPSExport.cpp
```diff
@@ -42,334 +59,523 @@
...
-          feature_stream << "CONSENSUSID=e_" << feature.getUniqueId() << endl;
-
-          feature_stream << "MSLEVEL=2" << endl;
-          feature_stream << "CHARGE=" << std::to_string(charge == 0 ? 1 : charge) << "+" << endl;
-          feature_stream << "PEPMASS=" << peptides[0].first.second.getMZ() << endl;
-          feature_stream << "FILE_INDEX=" << peptides[0].second.second << endl;
-          feature_stream << "RTINSECOND=" << peptides[0].first.second.getRT() << endl;
-
-          for (auto ms2_iter = ms2_block.rbegin(); ms2_iter != ms2_block.rend(); ++ms2_iter)
+        if (output_file.is_open())
```
FabianAicheler, line 543: I feel that code redundancies of the output_types branches could be at least somewhat reduced to improve maintenance/readability.

E.g. by moving the case logic to only the specific, really different parts, or maybe adapting the 'merged' part to support filtered pepts that contain only the most intense one.

src/topp/GNPSExport.cpp
```diff
@@ -42,334 +59,523 @@
...
+  9. Run the @ref TOPP_GNPSExport on the "filtered consensusXML file" to export an .MGF file.
+  10. Run the @ref TOPP_TextExporter on the "filtered consensusXML file" to export an .TXT file.
+  11. Upload your files to GNPS and run the Feature-Based Molecular Networking workflow. Instructions are here:
 https://ccms-ucsd.github.io/GNPSDocumentation/featurebasedmolecularnetworking/
 
+The GitHub for that ProteoSAFe workflow and an OpenMS python wrappers is available here:
+https://github.com/Bioinformatic-squad-DorresteinLab/openms-gnps-workflow
+
+An online version of the OpenMS-GNPS pipeline for FBMN running on CCMS server (http://proteomics.ucsd.edu/) is available on GNPS:
+https://ccms-ucsd.github.io/GNPSDocumentation/featurebasedmolecularnetworking-with-OpenMS
```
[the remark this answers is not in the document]
lfnothias: @abisarvepalli this is the correct link: https://ccms-ucsd.github.io/GNPSDocumentation/featurebasedmolecularnetworking-with-openms/

src/topp/GNPSExport.cpp
```diff
@@ -42,334 +59,523 @@
...
 
-This tool was developed for the OpenMS-GNPS workflow. It can be accessed on GNPS (https://gnps.ucsd.edu/ProteoSAFe/static/gnps-experimental.jsp). The steps used by that workflow are as following:
+This tool was developed for the Feature Based Molecular Networking (FBMN) workflow on GNPS (https://gnps.ucsd.edu/ProteoSAFe/static/gnps-splash2.jsp)
+
+Please cite our preprint: Nothias, L.F. et al, Feature-based Molecular Networking in the GNPS Analysis Environment
+bioRxiv 812404 (2019) (https://www.biorxiv.org/content/10.1101/812404v1)
```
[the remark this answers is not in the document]
lfnothias: This is the citation:
Nothias, LF., Petras, D., Schmid, R. et al. Feature-based molecular networking in the GNPS analysis environment. Nat Methods 17, 905–908 (2020). https://doi.org/10.1038/s41592-020-0933-6

src/topp/GNPSExport.cpp
```diff
@@ -441,88 +463,61 @@ class TOPPGNPSExport : public TOPPBase
...
+        feature.getMZ(),
+        best_speci,
+        best_spec.getRT()
+      );
+
+      // store outputted spectra in MSExperiment
+      MSExperiment exp;
+
+      // add most intense spectrum to MSExperiment
+      exp.addSpectrum(best_spec);
```
FabianAicheler, line 513: I'm unsure whether this line (without logic check that output_type!= 'merged_spectra') causes the most intense spectrum to be added twice (due to cosine similarity of 1 to itself in the 'merged_spectra' branch). And if so, would that make a difference after flattening and binning for reported intensities?

jpfeuffer:
rebuild jenkins

jpfeuffer:
I merged this in the other linked PR. Would be nice if the last question of @FabianAicheler will be addressed still.

PR #4772, opened 2020-06-05
[FEATURE] export localization scores in mzTab

timosachsenberg:


timosachsenberg:
tests passed

jpfeuffer:
I was wondering if a UserParam in the SearchParam object would be a better place. Since the other modifications etc are already stored there.

PR #4775, opened 2020-06-05
subsitute vector<double> for map<PeakType, double> in SignalToNoiseEstimator

xchen98:
# Optimize of SignalToNoiseEstimator in OpenSwathWorkflow

It coud be found that the type pf data, map<PeakType, doule>, was used in SignalToNoiseEstimator of Tool OpenSwathWorkflow, in which the values corresponded to the iteration of map are storaged. The stored values are synchronously updated contiguously with the iteration, using the function named getSignalToNoise. However, vector could be more efficient when processing large amounts of data. Thus, vectors are applied here. From the diagram shown below, it could be easily found that the processing speed differs a lot while using vector and map.(left: map | right: vector)
Signal_Vergleiche
# Benchmarks

runtime of OpenSwathWorkflow with map: 17:31 m
runtime of OpenSwathWorkflow with vector: 15:38m


src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.h
```diff
@@ -95,56 +89,26 @@ namespace OpenMS
...
-
-      return stn_estimates_[*data_point];
+      computeSTN_(c);
     }
 
-    virtual double getSignalToNoise(const PeakType & data_point)
+    ///Return to signal/noise estimate for date point @p index
+    ///@note you will get a warning to stderr if more than 20% of the
+    ///      noise estimates used sparse windows
+    virtual double getSignalToNoise(const Size index)
```
cbielow, line 108: could be `virtual double getSignalToNoise(const Size index) const` now?

src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimatorMedian.h
```diff
@@ -222,28 +225,21 @@ namespace OpenMS
...
-          ++run;
-        }
+        auto maxIt = std::max_element(c.begin(), c.end() ,[](const PeakType& a, const PeakType& b){ return a.getIntensity() > b.getIntensity();});
+        typename PeakType::IntensityType maxInt = maxIt->getIntensity();
 
         double bin_size = maxInt / 100;
 
         // fill histogram
-        run = scan_first_;
+        PeakIterator run = scan_first_;
```
cbielow, line 234: this could be a range-based for:
```for (const auto& peak : c)
{ ...
}
```
which would make it cleaner

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -325,11 +325,9 @@ namespace OpenMS
       p.setValue("noise_for_empty_window", 2.0);
       p.setValue("min_required_elements", 10);
       sne.setParameters(p);
-      sne.init(annotated_spectra[i].begin(), annotated_spectra[i].end());
+      sne.init(annotated_spectra[i]);
       double avgSNR { 0 };
-      for (MSSpectrum::const_iterator it = annotated_spectra[i].begin();
-           it != annotated_spectra[i].end();
-           ++it)
+      for (Size it = 0; it < annotated_spectra[i].size(); ++it)
```
cbielow, line 330: call the variable `i`, not `it` (since this suggests an iterator, which is not true anymore)

src/topp/FileFilter.cpp
```diff
@@ -851,14 +851,14 @@ class TOPPFileFilter :
         SignalToNoiseEstimatorMedian<MapType::SpectrumType> snm;
         Param const& dc_param = getParam_().copy("algorithm:SignalToNoise:", true);
         snm.setParameters(dc_param);
-        for (MapType::Iterator it = exp.begin(); it != exp.end(); ++it)
+        for (Size it = 0; it != exp.size(); ++it)
```
cbielow, line 854: again, range-based for is cleaner (and `it` is not a good name)
```for (auto & spec : exp)
{ ....
}
```

cbielow:
looks good to me. only minor changes

src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.h
xchen98: why could be this function `const`?

cbielow: because it does read-only access to the underlying member, i.e. does not change the state of the object

cbielow:
@xchen98 could you have a look at the comments. Afterwards this is ready for merge

xchen98:
here are the new modifications

cbielow:
rebuild jenkins

cbielow:
pyOpenMS fails.
is this expected?

jpfeuffer:
Yes, this PR changes the types in the interface.
```
error: no matching function for call to 'OpenMS::SignalToNoiseEstimatorMedian<OpenMS::MSSpectrum>::getSignalToNoise(boost::shared_ptr<OpenMS::Peak1D>::element_type&)'
```

src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.h
```diff
@@ -95,56 +89,26 @@ namespace OpenMS
...
+    virtual double getSignalToNoise(const Size index) const
     {
-      if (!is_result_valid_)
-      {
-        // recompute ...
-        init(first_, last_);
-      }
-
-      return stn_estimates_[data_point];
+      OPENMS_POSTCONDITION(index < stn_estimates_.size(),"SignalToNoiseEstimator estimates beyond container size was requested.");
```
hroest, line 110:
```suggestion
      OPENMS_PRECONDITION(index < stn_estimates_.size(), "SignalToNoiseEstimator estimates beyond container size was requested.");
```

hroest:
:+1: this looks good, needs fixing in pyOpenMS. Also it may be interesting to make the computation faster for the edge case when the S/N window is larger than the length of the input -- in that case the computation should be significantly faster. Have you thought about that?

hroest:
for pyOpenMS I suggest to modify this:

https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/pxds/SignalToNoiseEstimatorMedian.pxd#L19

to `void getSignalToNoise(const Size index) nogil except +`

timosachsenberg:
Thanks! this one got replaced so I closed it

PR #4778, opened 2020-06-08
Good Fix for FeatureLinkerUnlabeledQT Segfault Issue

Felix-Droop:
Regarding this conversation https://github.com/OpenMS/OpenMS/pull/4756.

Removing this temporary fix https://github.com/OpenMS/OpenMS/pull/4760.

I found the real cause of this issue and fixed it. It is confusing (see the big comment I added in the Code).

This fixed the segfault in @timosachsenberg's dataset.

One could consider renaming the Function `addClusterElements()`. I did not do it, because the function itself only adds elements to the cluster. The removing of an element (rare case that caused this issue) happens somewhere inside the `QTCluster` class after calling `addClusterElements()`. At least that's my theory, as I did not work on all the code involved there.


cbielow:
rebuild jenkins

PR #4779, opened 2020-06-08
Try fixing pyopenms on win with GH actions

jpfeuffer:
Since they upgraded to VS2019 update 16.6,
our builds fail. This is a PR with attempts to fix this.
In the beginning, I revert some changes that I tried with direct commits.

### DIFF
--- a/.github/workflows/msvc.py.patch
+++ b/.github/workflows/msvc.py.patch
@@ -1,8 +1,10 @@
 --- msvc.py	2020-08-24 17:38:24.000000000 +0200
-+++ msvc.py	2020-08-25 11:28:25.000000000 +0200
-@@ -276,7 +276,7 @@
++++ msvc.py	2020-08-25 13:04:26.000000000 +0200
+@@ -275,8 +275,9 @@
+             stderr=subprocess.STDOUT,
          ).decode('utf-16le', errors='replace')
      except subprocess.CalledProcessError as exc:
++        print("Error executing {}. Error was {}".format(exc.cmd, exc.output))
          raise distutils.errors.DistutilsPlatformError(
 -            "Error executing {}".format(exc.cmd)
 +            "Error executing {}. Error was {}".format(exc.cmd, exc.output)
--- a/.github/workflows/msvc.py.patch
+++ b/.github/workflows/msvc.py.patch
@@ -1,0 +1,11 @@
+--- msvc.py	2020-08-24 17:38:24.000000000 +0200
++++ msvc.py	2020-08-25 11:28:25.000000000 +0200
+@@ -276,7 +276,7 @@
+         ).decode('utf-16le', errors='replace')
+     except subprocess.CalledProcessError as exc:
+         raise distutils.errors.DistutilsPlatformError(
+-            "Error executing {}".format(exc.cmd)
++            "Error executing {}. Error was {}".format(exc.cmd, exc.output)
+         ) from exc
+ 
+     env = {
--- a/.github/workflows/pyopenms-wheels.yml
+++ b/.github/workflows/pyopenms-wheels.yml
@@ -40,10 +40,10 @@ jobs:
 #      run: |
 #        "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" modify --channelId VisualStudio.16.Release --productId Microsoft.VisualStudio.Product.Enterprise -q --includeRecommended --add "Microsoft.VisualStudio.Component.Windows10SDK.18362" --add "Microsoft.VisualStudio.Component.VC.CLI.Support" --add "Microsoft.VisualStudio.Component.VC.Modules.x86.x64" --add "Microsoft.VisualStudio.Component.VC.Tools.x86.x64" --add "Microsoft.VisualStudio.Component.VC.v141.x86.x64" --add "Microsoft.VisualStudio.Component.VC.140" --add "Microsoft.VisualStudio.Component.VC.Redist.14.Latest" --add "Microsoft.VisualStudio.Component.VC.CoreIde" --add "Microsoft.VisualStudio.Component.VC.CMake.Project"
 
-    - name: Install Buildtools 
-      shell: cmd
-      run: |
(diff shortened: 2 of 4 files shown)


jpfeuffer:
Actually I have no idea about what happens. I just keep this PR if someone finds/knows something.
Until then, no pyopenms nightlies.

oliveralka:
I found a comment, that it was necessary to update `setuptools` before pip. (https://stackoverflow.com/questions/44951456/pip-error-microsoft-visual-c-14-0-is-required)

We are currently not doing that see line 81 (https://github.com/OpenMS/OpenMS/blob/c1e815d875c704cd6a7933def9a10e02e1435967/.github/workflows/pyopenms-wheels.yml#L81)

jpfeuffer:
Yeah, I am pretty sure that 2019 should work (since it worked before, e.g. ~1 week ago). Github "just" updated the version from 16.5 to 16.6 (which included a slightly newer compiler version).
To add to the confusion, if I deactivate conda env beforehand, 3.6 fails already. Now after reverting, 3.6 passes but 3.7 fails..

I don't think we can do much here but hope that people will have a similar problem. I don't even know where to start raising issues: github? setuptools? pip? Microsoft?

oliveralka:
I have also seen these issues maybe they are related:

```
Building Custom Rule D:/a/OpenMS/OpenMS/OpenMS/src/openms/CMakeLists.txt

LINK : warning LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs; use /NODEFAULTLIB:library [D:\a\OpenMS\OpenMS\bld\src\openms\OpenMS.vcxproj]
EXEC : warning : cannot resolve item 'OpenSwathAlgo.dll' [D:\a\OpenMS\OpenMS\bld\src\pyOpenMS\prepare_pyopenms_libs.vcxproj]
EXEC : -- warning : gp_resolved_file_type non-absolute file 'OpenSwathAlgo.dll' returning type 'other' -- possibly incorrect [D:\a\OpenMS\OpenMS\bld\src\pyOpenMS\prepare_pyopenms_libs.vcxproj]
EXEC : warning : cannot resolve item 'OpenMS.dll' [D:\a\OpenMS\OpenMS\bld\src\pyOpenMS\prepare_pyopenms_libs.vcxproj]
  possible problems:
      need more directories?
      need to use InstallRequiredSystemLibraries?
      run in install tree instead of build tree?
```
I guess let's wait for a day or two and see if someone has the similar problems.

jpfeuffer:
Yes I was confused at first, too by these warnings, but I think I checked and they were there before, when it worked as well.

oliveralka:
@jpfeuffer It seems that the compiler is not found - other had a similar issue, but I am not sure if we can fix it the same way:

https://github.com/python-pillow/Pillow/pull/4201

jpfeuffer:
The thing is, we already used windows-2019 successfully before.
Seems more like they messed up the components during the update of VS on that environment:
https://github.com/actions/virtual-environments/issues/1028

oliveralka:
Read a comment, that it worked for someone after installing the "Desktop development with C++" for VS2019 (https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-enterprise?view=vs-2019).

It seems that python 3.6 builds correctly, but afterwards, 3.7 throws the error.

jpfeuffer:
Indeed, very weird. That's why I am starting to doubt that it is a missing component but maybe a setuptools incompatibility?

oliveralka:
Yes, I also stumbled over the issue you already commented on:
https://github.com/pypa/setuptools/issues/2028

I guess we can wait for a bit longer and see what comes out of it.

jpfeuffer:
Unfortunately, I cant even reproduce on a simple example. So it is related to our complex setup.
https://github.com/jpfeuffer/hello-github-actions/runs/780323701?check_suite_focus=true

jpfeuffer:
I am starting to think it has something to do with conda. Conda also installs a "vc" package together with its python.
We probably should test it with conda on a local Windows machine at some point to rule out errors there.

oliveralka:
Yes, I saw that as well, what we could try is:
Activate conda-env
Install needed packages
Deactivate conda-env
Add only the python executable path to CMake (-DPYTHON_EXECUTABLE:FILEPATH=C:/Python33/python.exe)

Then we would not stay in the conda environment. What do you think?

mattip:
Maybe the conda install of vs tools is changing values in the registry in a way that now requires conda to be activated in order to find the tools?

jpfeuffer:
So this is the problem: https://github.com/pypa/setuptools/blob/8d508965e49028b68e09b930a10340ba09687d35/setuptools/msvc.py#L272

It works outside of CMake. And it worked some time ago. I don't know what happens here.

mattip:
Maybe one of your environment variables (like PATH) is getting close to the limit, and running vcvarsall inside CMake pushes it over the limit? The problem is that stderr is being swallowed. Could you modify your version of setuptools to call `check_output` with `stderr=subprocess.STDOUT` (you may need `shell=True` as well) to catch the actual error?

jpfeuffer:
So, if I decode manually it says:
```
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.7.2
** Copyright (c) 2020 Microsoft Corporation
**********************************************************************
The input line is too long.
The syntax of the command is incorrect.
```

jpfeuffer:
Which confirms your idea @mattip .
However, I do not exactly understand why calling it from inside CMake makes a difference.
According to https://developercommunity.visualstudio.com/content/problem/257260/vcvarsallbat-reports-the-input-line-is-too-long-if.html you are also not supposed to call it repeatedly, but I cannot avoid it since setuptools does it internally.

mattip:
Can you trim some of the PATH components before the step?

jpfeuffer:
Thanks, it worked. And I know that "clearing" parts of the path is the better idea but not nice either. Fiddling with the PATH with a rather hardcoded hack will break the workflow whenever GitHub updates their runners.

mattip:
Maybe open an issue on the image repo https://github.com/actions/virtual-environments/issues

jpfeuffer:
superseded by simple commit 571d820633365b43aec243bb9829df8a61358fce

jpfeuffer:
Thanks a lot for reporting back! Maybe we can use your env var trick, too. It looks a lot cleaner than trying to reduce the PATH.

PR #4780, opened 2020-06-08
TOPPAS fixes

cbielow:
1)  fix 'FileNotWritable' exception in TOPPAS when temporary filenames only differ in case (upper vs lower) and thus renaming deletes the source (18e8d07)

2) [FIX] prevent Exception from skipping over important cleanup code (might lead to inoperable pipelines which cannot be executed anymore) (57581e2)


timosachsenberg:
feel free to merge

PR #4781, opened 2020-06-09
[Feature] Concatenate option for IDFileConverter

Waschi97:
- added a flag which allows the user to convert the top hits to one fasta entry

- changed the way the fasta output is written

- added TOPP tests for idXML to fasta conversion

This feature is needed for the database suitability metric.


Waschi97:
Retry DeepCode

src/topp/IDFileConverter.cpp
```diff
@@ -240,7 +240,7 @@ class TOPPIDFileConverter :
     registerFlag_("no_spectra_references_override", "[+mz_file only] Setting this flag will avoid overriding 'spectrum_reference' in PeptideIdentifications if mz_file is given and a 'spectrum_reference' is already present.", true);
     registerDoubleOption_("add_ionmatch_annotation", "<tolerance>", 0,"[+mz_file only] Will annotate the contained identifications with their matches in the given mz_file. Will take quite some while. Match tolerance is .4", false, true);
 
-    registerFlag_("concatenate_peptides", "[FASTA output only] Setting this flag will concatenate the top peptide hits to one entry, rather than write a new entry fro each hit.", true);
+    registerFlag_("concatenate_peptides", "[FASTA output only] Setting this flag will concatenate the top peptide hits to one entry, rather than write a new entry for each hit.", true);
```
cbielow, line 243: Maybe condense the information a bit and be more precise:
 - `Setting this flag will ...` is just additional payload.... remove.
 - replace `entry` with more specific `protein sequence` and `protein`

??

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +617,49 @@ class TOPPIDFileConverter :
     else if (out_type == FileTypes::FASTA)
     {
       Size count = 0;
-      ofstream fasta(out.c_str(), ios::out);
+
+      bool concat = getFlag_("concatenate_peptides");
+      //To avoid creating (K/R)P sites and losing information when the FASTA entry is used by
```
cbielow, line 622: [KP]|P sites are created in any case. The question is just how this affects the preceeding peptides if a peptide starts with P.

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +617,49 @@ class TOPPIDFileConverter :
...
+      //a search engine; Peptides beginning with 'P' are saved seperatly and later moved to the beginning of the concatenated sequence.
```
cbielow, line 623: separately

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +617,49 @@ class TOPPIDFileConverter :
...
+      String all_p(""); //peptides beginning with 'P'
```
cbielow, line 624: `String all_p;`

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +617,49 @@ class TOPPIDFileConverter :
...
+      f.writeStart(out);
+      FASTAFile::FASTAEntry entry;
       for (Size i = 0; i < peptide_identifications.size(); ++i)
       {
-        for (Size l = 0; l < peptide_identifications[i].getHits().size(); ++l)
+        if (concat)
+        {
+          if (peptide_identifications[i].getHits()[0].getSequence().toString()[0] == 'P')
+          {
+            all_p += peptide_identifications[i].getHits()[0].getSequence().toString();
```
cbielow, line 636: - do you really want the modified peptide sequence, not the unmodified one?
- calling `toString()` is expensive. Compute it once and use it for comparing to 'P'.
- you are missing an `else`...

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +617,49 @@ class TOPPIDFileConverter :
...
+      bool concat = getFlag_("concatenate_peptides");
+      //To avoid creating (K/R)P sites and losing information when the FASTA entry is used by
+      //a search engine; Peptides beginning with 'P' are saved seperatly and later moved to the beginning of the concatenated sequence.
+      String all_p(""); //peptides beginning with 'P'
+      String all_but_p(""); //all the others
+
+      FASTAFile f;
+      f.writeStart(out);
+      FASTAFile::FASTAEntry entry;
       for (Size i = 0; i < peptide_identifications.size(); ++i)
```
cbielow, line 630: use a range-based for loop for brevity

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +617,49 @@ class TOPPIDFileConverter :
...
-          String seq = hit.getSequence().toUnmodifiedString();
-          std::set<String> prot = hit.extractProteinAccessionsSet();
-          fasta << ">" << seq
-                << " " << ++count
-                << " " << hit.getSequence().toString()
-                << " " << ListUtils::concatenate(StringList(prot.begin(), prot.end()), ";")
-                << "\n";
-          // FASTA files should have at most 60 characters of sequence info per line
-          for (Size j = 0; j < seq.size(); j += 60)
+          for (Size l = 0; l < peptide_identifications[i].getHits().size(); ++l)
```
cbielow, line 642: writing out all hits when in non-concat mode is a bit inconsistent.
Do we actually want this?
If yes, should there be an option of how many hits to write (default=1) for both modes?

Waschi97: I'm not sure if we want this. This is the way it is done currently, but I'll happily add an input option for this.

jpfeuffer: I think it would be good to have this option. Yes, not necessarily for this project, but in general.

src/topp/IDFileConverter.cpp
```diff
@@ -619,9 +619,13 @@ class TOPPIDFileConverter :
     {
       Size count = 0;
       Int max_hits = getIntOption_("number_of_hits");
-      if (max_hits < 1)
+      if (max_hits == 0)
       {
-        OPENMS_LOG_WARN << "'number of hits' set to less then 1. No peptide hits will be exported!" << endl;
+        OPENMS_LOG_WARN << "'number of hits' set to 0. No peptide hits will be exported! Please enter the number of peptide hits you wish to export. To export all hits enter a negativ number." << endl;
+      }
+      if (max_hits < 0)
```
cbielow, line 626: Just make it <= 0.
Avoids extra code and exporting 0 entries is nothing anyone would ever want

src/tests/topp/CMakeLists.txt
```diff
@@ -962,6 +962,23 @@ set_tests_properties("TOPP_IDFileConverter_25_out1" PROPERTIES DEPENDS "TOPP_IDF
...
 add_test("TOPP_IDFileConverter_26_out1" ${DIFF} -in1 IDFileConverter_26_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_26_output.idXML -whitelist "IdentificationRun date" )
 set_tests_properties("TOPP_IDFileConverter_26_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_26")
+# Test idXML to FASTA
+# no concatenation
+add_test("TOPP_IDFileConverter_27" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/IDFileConverter_27_input.idXML -out IDFileConverter_27_output.tmp -out_type FASTA)
+add_test("TOPP_IDFileConverter_27_out1" ${DIFF} -in1 IDFileConverter_27_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_27_output.fasta)
+set_tests_properties("TOPP_IDFileConverter_27_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_27")
+add_test("TOPP_IDFileConverter_28" ${TOPP_BIN_PATH}/IDFileConverter -test -in ${DATA_DIR_TOPP}/IDFileConverter_27_input.idXML -out IDFileConverter_28_output.tmp -out_type FASTA -number_of_hits -1)
+add_test("TOPP_IDFileConverter_28_out1" ${DIFF} -in1 IDFileConverter_28_output.tmp -in2 ${DATA_DIR_TOPP}/IDFileConverter_28_output.fasta)
+set_tests_properties("TOPP_IDFileConverter_28_out1" PROPERTIES DEPENDS "TOPP_IDFileConverter_28")
```
cbielow, line 972: Very thorough. Nice

src/topp/IDFileConverter.cpp
```diff
@@ -239,6 +239,9 @@ class TOPPIDFileConverter :
     registerFlag_("no_spectra_data_override", "[+mz_file only] Setting this flag will avoid overriding 'spectra_data' in ProteinIdentifications if mz_file is given and 'spectrum_reference's are added/updated. Use only if you are sure it is absolutely the same mz_file as used for identification.", true);
     registerFlag_("no_spectra_references_override", "[+mz_file only] Setting this flag will avoid overriding 'spectrum_reference' in PeptideIdentifications if mz_file is given and a 'spectrum_reference' is already present.", true);
     registerDoubleOption_("add_ionmatch_annotation", "<tolerance>", 0,"[+mz_file only] Will annotate the contained identifications with their matches in the given mz_file. Will take quite some while. Match tolerance is .4", false, true);
+
+    registerFlag_("concatenate_peptides", "[FASTA output only] Will concatenate the top peptide hits to one peptide sequence, rather than write a new peptide for each hit.", true);
+    registerIntOption_("number_of_hits", "<integer>", 1, "[FASTA output only] Controls how many peptide hits will be exported. A negative value exports all hits.", false, true);
```
cbielow, line 244: A value of zero or less

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +618,67 @@ class TOPPIDFileConverter :
...
-                << " " << ListUtils::concatenate(StringList(prot.begin(), prot.end()), ";")
-                << "\n";
-          // FASTA files should have at most 60 characters of sequence info per line
-          for (Size j = 0; j < seq.size(); j += 60)
+          if (curr_hit > max_hits) break;
+          ++curr_hit;
+
+          if (concat)
+          {
+            String seq = hit.getSequence().toUnmodifiedString();
```
cbielow, line 651: Seq is used in all cases. Move before if clause?

src/topp/IDFileConverter.cpp
```diff
@@ -615,28 +618,67 @@ class TOPPIDFileConverter :
...
+            ++count;
           }
         }
       }
-      fasta.close();
+      if (concat)
+      {
+        entry.sequence = all_p + all_but_p;
+        entry.identifier = protein_identifications[0].getSearchEngine();
+        entry.description = "concatenated peptides";
```
cbielow, line 678: Maybe take this description as ´const char* CONCAT_PEPTIDE = "concatenated peptide";´ which you place into the library and use from all places you need it (see Constants.h?)

Waschi97: Your thought is that this can later be used again for searching for those peptides in the db suitability tool, right?
This wouldn't work though, because the description is lost during the peptide identification search, isn't it?
Or am I missing something here?

cbielow: very good thinking! :)
prepend it to the identifier then?

cbielow: (omit spaces and other funny stuff, use underscores if required)

PR #4788, opened 2020-06-12
[FEATURE] SwathWizard: show summary statistics about a TargetedExperiment

cbielow:
followup on https://github.com/OpenMS/OpenMS/pull/4773

shows stats about a pqp library file:
image


cbielow:
failing tests are unrelated (not rebuilding to safe the plant) :)

Any comments?

jpfeuffer:
Looks good from my side. I assume the deepcode warning is some Qt (auto generation?) related false positive?

cbielow:
> 
> 
> Looks good from my side. I assume the deepcode warning is some Qt (auto generation?) related false positive?

DeepCode does not understand that heap-allocated widgets are deleted by the parent widget in Qt and are thus not a mem-leak.

PR #4795, opened 2020-06-24
Toppview refactor

cbielow:
some refactoring to make TOPPViewBase a bit more managable...

reduces line count of TOPPViewBase from 4400 to 3700 lines...
Some more refactoring will be required to make the class readable again...


src/openms/include/OpenMS/KERNEL/MSExperiment.h
```diff
@@ -545,6 +545,13 @@ namespace OpenMS
     */
     void clear(bool clear_meta_data);
 
+    /// returns true if at least one of the spectra has level==1
+    bool containsMS1Scans() const;
```
timosachsenberg, line 549: I would be fine with having this in MSExperiment. But maybe more generic by passing the level?

src/openms/include/OpenMS/KERNEL/MSExperiment.h
```diff
@@ -545,6 +545,13 @@ namespace OpenMS
...
+
+    /// returns true if any MS1 spectra contain at least one peak with intensity of 0.0
+    bool hasMS1Zeros() const;
```
timosachsenberg, line 552: same here

src/openms/source/KERNEL/MSSpectrum.cpp
```diff
@@ -699,4 +699,17 @@ namespace OpenMS
   bool MSSpectrum::RTLess::operator()(const MSSpectrum &a, const MSSpectrum &b) const {
     return a.getRT() < b.getRT();
   }
+
+  // static
+  bool MSSpectrum::containsIMData() const
```
timosachsenberg, line 704: maybe better to return bool and type?

src/openms_gui/include/OpenMS/VISUAL/FilterList.h
```diff
@@ -0,0 +1,94 @@
...
+  class TOPPViewBase;
+
+  namespace Internal
+  {
+    /**
+      @brief A widget shows a list of input files (i.e. existing files on a mounted drive),
+             which allows adding/removing files and supports drag'n'drop from the window manager.
+
+    */
+    class FilterList : public QWidget
```
timosachsenberg, line 61: much cleaner. nice

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -262,13 +263,11 @@ namespace OpenMS
     menuBar()->addMenu(help);
     help->addAction(QWhatsThis::createAction(help));
     help->addSeparator();
-    QAction* action = help->addAction("OpenMS website", this, SLOT(showURL()));
-    action->setData("http://www.OpenMS.de");
-    action = help->addAction("Tutorials and documentation", this, SLOT(showURL()), Qt::Key_F1);
-    action->setData(String("html/index.html").toQString());
+    help->addAction("OpenMS website", [&]() { GUIHelpers::openURL("http://www.OpenMS.de"); });
```
timosachsenberg, line 266: so much cleaner 👍

timosachsenberg:
nice

src/openms/source/KERNEL/MSSpectrum.cpp
cbielow: what type?

timosachsenberg: I thought type of ion mobility that is stored... maybe some CVTerm ... but not urgent

cbielow: I'm not sure this is well defined. Will omit that for now

timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=2256

PR #4796, opened 2020-06-24
FIX: MSstatsTMTline_ bool operator

oliveralka:
This fix resolves the issue with skipped channels in the MSstatsConverter.

Thanks at @timosachsenberg and @jpfeuffer.


timosachsenberg:
Looks good! Thanks!

PR #4797, opened 2020-06-25
[FIX] bug introduced when trying to reuse vector

jpfeuffer:
Was not cleared correctly and sometimes led to type mismatch during get.


cbielow:
looks good (i guess).

PR #4798, opened 2020-06-25
[Feature] id filter on consensusXML

jpfeuffer:


timosachsenberg:
nice!

src/topp/IDFilter.cpp
```diff
@@ -243,7 +245,38 @@ class TOPPIDFilter :
...
+      IdXMLFile().load(inputfile_name, proteins, peptides);
+    }
+    else if (FileHandler::getTypeByFileName(inputfile_name) == FileTypes::CONSENSUSXML)
+    {
+      ConsensusXMLFile().load(inputfile_name, cmap);
+      for (auto& f : cmap)
+      {
+        UInt64 id = f.getUniqueId();
+        id_to_featureref[id] = &f;
+        for (auto&& p : f.getPeptideIdentifications())
```
cbielow, line 264: I find this confusing, since there is no case where we use reference proxies in our code (that I'm aware of).
Use `auto& p: ..`?

src/topp/IDFilter.cpp
```diff
@@ -705,7 +738,31 @@ class TOPPIDFilter :
...
              << IDFilter::countHits(peptides) << " spectrum matches." << endl;
 
-    IdXMLFile().store(outputfile_name, proteins, peptides);
+    if (FileHandler::getTypeByFileName(inputfile_name) == FileTypes::IDXML)
+    {
+      IdXMLFile().store(outputfile_name, proteins, peptides);
+    }
+    else if (FileHandler::getTypeByFileName(inputfile_name) == FileTypes::CONSENSUSXML)
+    {
+      for (auto&& p : peptides)
```
cbielow, line 747: same with auto&& and emplace here

PR #4801, opened 2020-06-30
fix unequal column bug

timosachsenberg:


src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2958,14 +2970,17 @@ namespace OpenMS
...
         if (first)
         { // add header
-          tab_file << "\n" << generateMzTabPSMHeader_(n_search_engine_scores, s.getPSMOptionalColumnNames()) + "\n";
+          tab_file << "\n" << generateMzTabPSMHeader_(n_search_engine_scores, s.getPSMOptionalColumnNames(), n_header_columns) + "\n";
           first = false;
         }
-        tab_file << generateMzTabSectionRow_(row, s.getPSMOptionalColumnNames(), meta_data) + "\n";
+        size_t n_section_columns = 0;
+        tab_file << generateMzTabSectionRow_(row, s.getPSMOptionalColumnNames(), meta_data, n_section_columns) + "\n";
+        if (n_header_columns != n_section_columns)  throw Exception::Postcondition(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "PSM header and content differs in columns. Please report this bug to the OpenMS developers.");
```
timosachsenberg, line 2983: now makes sure that the number of header columns never differ from the number of section columns

timosachsenberg: as a side effect: also makes sure that all section rows have the same number of columns

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -3024,31 +3040,44 @@ namespace OpenMS
...
+      size_t n_header_columns = 0;
       while (s.nextPEPRow(row))
       {
         if (first)
         {
           assays = row.peptide_abundance_assay.size();
           study_variables = row.peptide_abundance_study_variable.size();
-          Size n_search_engine_score = row.search_engine_score_ms_run.size();
+          Size n_search_engine_score = row.search_engine_score_ms_run.size(); // scores to runs          
+          Size search_ms_runs = n_search_engine_score != 0 ? row.search_engine_score_ms_run.at(1).size() : 0; // take number of searched MS runs from first score. TODO: handle this more generic
```
timosachsenberg, line 3066: this is new

PR #4802, opened 2020-07-02
[FEATURE] added filtering for protein group scores

jpfeuffer:


src/topp/IDFilter.cpp
```diff
@@ -157,7 +157,7 @@ class TOPPIDFilter :
     registerTOPPSubsection_("score", "Filtering by peptide/protein score.");
     registerDoubleOption_("score:pep", "<score>", 0, "The score which should be reached by a peptide hit to be kept.", false);
     registerDoubleOption_("score:prot", "<score>", 0, "The score which should be reached by a protein hit to be kept. Use in combination with 'delete_unreferenced_peptide_hits' to remove affected peptides.", false);
-
+    registerDoubleOption_("score:protgroup", "<score>", 0, "The score which should be reached by a protein group to be kept. Use in combination with 'delete_unreferenced_peptide_hits' to remove affected peptides.", false);
```
timosachsenberg, line 160: don't know maybe this needs a bit more explanation what the difference between score:prot and score:protgroup is.
Would it even make sense to just use score:prot and have the new semantics?

jpfeuffer: No, I think it is important if people actually care about specific isoforms to filter for single protein scores.
Hmm.. I think the difference is clear, the question is whether people know the implications.

timosachsenberg: I wouldn't say for the casual user. 
E.g., is it clear that score:protgroup includes single proteins? Most would probably guess no

PR #4803, opened 2020-07-03
[UPDATE] Unimod xml and obo files

enetz:
Updated Unimod.xml and Unimod.obo to newest version available on 02.07.2020

- Added the CUSTOM0-9, RNPXlink1-5 and other modifications that we had added to our previous unimod.xml ourselves
- Fixed tests related to changed format of cross-linking modifications
- AASequence tests: one tested mod changed so much, that it doesn't have the tested behavior anymore. I found another mod with a different mass that has the same behavior as the previous mod to make sure the behavior is still tested
- AASequence tests: added test for new type of modification with square brackets in their name
- MRMAssay tests: I was able to fix two tests, but the rest requires changes to TraML input or output test files, that are not obvious to me. The changes to this test class probably all stem from additional origin residues to previously used mods and the new possible Peptidoforms that are created from that.

A few more tests are still failing, but I am not sure how they are related to the Unimod update.

These are the tests still failing for me so far:
```
The following tests FAILED:
	578 - MRMAssay_test (Failed)
	2295 - TOPP_MSGFPlusAdapter_1 (Failed)
	2296 - TOPP_MSGFPlusAdapter_1_out1 (Failed)
	2297 - TOPP_MSGFPlusAdapter_1_out2 (Failed)
	2301 - TOPP_CometAdapter_1_out1 (Failed)
	2304 - TOPP_CometAdapter_2_out1 (Failed)
	2352 - TOPP_NovorAdapter_1_out (Failed)
```


jpfeuffer:
Seems like only the MRMAssay tests fail.
And the problem is that they added another potential neutral loss for deamidation.
Unfortunately, from quickly skimming over the code, our parser is again quite dumb and just takes the last occurring neutral loss in Unimod as the one and only (since it overwrites with each /NeutralLoss element)
This makes the R Deamidation different from the others and they are not grouped anymore for the transitions.
This is my explanation. Potential fixes are: 
- Add multiple mods, one per neutral loss. How to differentiate them?
- Use the first neutral loss instead of the last? Are they sorted by importance?

timosachsenberg:
Couldn't we read them into https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/CHEMISTRY/Residue.h#L260

jpfeuffer:
That works too of course. I don't know how often this is used in our algorithms at all.
Honestly, I also don't know why they would list nothing ("0") as a potential neutral loss. Shouldn't that be the default?
Or does a single annotated neutral loss mean that it should be considered (almost) always?

src/tests/class_tests/openms/source/AASequence_test.cpp
```diff
@@ -1422,16 +1422,16 @@ START_SECTION([EXTRA] testing terminal modifications)
   TEST_EQUAL(AASequence::fromString("(UniMod:385)CFPIANGER.").toBracketString(true, true), "n[-17]CFPIANGER") // relative
 
   // UniMod 1009 can only occur on N-terminal peptide on an C
```
timosachsenberg, line 1424: still refering to UniMod1009

timosachsenberg:
@enetz can you take a look how much work it would be to read all losses into a vector of neutral losses for a ResidueModification?

enetz:
This PR should work in general, but right now I can't fix the last failing MRMAssay test.
Some modifications got new origins and several new Peptidoforms were added (same peptide, same mass, same modification, but new additional positions for that modification).
The test uses two different file versions for different versions of installed Boost:

```
#if OPENMS_BOOST_VERSION_MINOR < 56
  String out1 = "MRMAssay_uisTransitions_output_1.TraML";
#else
  String out1 = "MRMAssay_uisTransitions_output_1_boost58.TraML";
#endif
```

I was able to update the `...boost58.TraML` files, but the Travis test has an older boost version and uses the other file.
I think the files
```
MRMAssay_uisTransitions_output_1.TraML
MRMAssay_uisTransitions_output_2.TraML
MRMAssay_uisTransitions_output_4.TraML
MRMAssay_uisTransitions_output_5.TraML
```
need to be updated for the older boost versions. I had to update these four for the newer boost version.

@timosachsenberg  @jpfeuffer   do you have an older boost version installed, or can we copy the tmp files from the Travis test directly? I don't want to mess with my boost installation just for this test.

jpfeuffer:
Yes, you can login to the travis machines when you restart them as a debug build.
https://docs.travis-ci.com/user/running-build-in-debug-mode/

Or check our abidocker and spin up an old linux.

enetz:
One additional feature that is added with this PR is the parsing of multiple neutral losses for modified residues.
Previously only one such neutral loss was parsed for a modification, even though we already use vectors of neutral losses for Residues in general.

The `OMSSAAdapter` is the only tool, that addresses neutral losses of modifications directly instead of addressing them from the modified residues and it expects one neutral loss. In this case, the first neutral loss is used.
This might be a design flaw in `OMSSAAdapter`, since I don't believe it should be necessary to look for neutral losses of modifications specifically. Rather the `Residue` class has vectors of loss names and loss formulas that are updated when a modification is set.

enetz:
One more issue: forgot to update python bindings
Will start fixing now

jpfeuffer:
Weird error only on macOS :( ?? https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/bldtst/compiler=appleclang-9.0.0,os_label=highsierra/11/testReport/junit/projectroot.src.tests/topp/TOPP_IDFileConverter_22_out1/

enetz:
Now only neutral losses with a non-empty empirical formula are stored. This removes empty, zero-mass neutral losses.

jpfeuffer:
Perfect! Thank you.

PR #4806, opened 2020-07-07
[FIX] colliding ProteinIDRun  IDs

jpfeuffer:
Fixes 
```
unable to read file (<filename) ProteinIdentifications are not unique, which leads to loss of unique peptide assignment. Duplicated Protein-Id is: unique_run_identifier.
```

when adding the "candidates_out" option and chunking.


timosachsenberg:
look good

PR #4814, opened 2020-07-15
FDR filtering in DBSuitability Tool

Waschi97:
- added FDR input option

- tool now only counts hits with q-value less than or equal to given FDR

- changed test output files accordingly

### DIFF
--- a/src/tests/topp/DatabaseSuitability_out_1.tsv
+++ b/src/tests/topp/DatabaseSuitability_out_1.tsv
@@ -1,7 +1,7 @@
 key	value
-#top_db_hits	438
+#top_db_hits	430
 #top_novo_hits	1
-db_suitability	0.997722095671982
+db_suitability	0.997679814385151
 #total_novo_seqs	1118
 #unique_novo_seqs	1111
 #ms2_spectra	1120
--- a/src/topp/DatabaseSuitability.cpp
+++ b/src/topp/DatabaseSuitability.cpp
@@ -94,7 +94,8 @@ To generate the de novo "database":
       - @ref TOPP_IDFilter can filter out unwanted ones.
       - @ref TOPP_IDFileConverter generates the de novo fasta file.
 
-For re-ranking all cases where a peptide hit only found in the de novo "database" scores above a peptide hit found in the actual database are checked. In all these cases the cross-correlation scores of those peptide hits are compared. If they are similar enough, the database hit will be re-ranked to be on top of the de novo hit. You can control how much of cases with similar scores will be re-ranked by using the @p novor_fract option.
+For re-ranking all cases where a peptide hit only found in the de novo "database" scores above a peptide hit found in the actual database are checked. In all these cases the cross-correlation scores of those peptide hits are compared. If they are similar enough, the database hit will be re-ranked to be on top of the de novo hit. You can control how much of cases with similar scores will be re-ranked by using the @p novor_fract option.@n
+For this to work it is important that FDR filtering is done in this tool and not beforehand by @ref TOPP_FalseDiscoveryRate. You can provide the wanted FDR using the corresponding flag.
 
 @note For identification search the only supported search engine for the time being is Comet because the Comet cross-correlation score is needed for re-ranking.@n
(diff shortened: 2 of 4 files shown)


src/topp/DatabaseSuitability.cpp
```diff
@@ -428,6 +439,23 @@ class DatabaseSuitability :
     }
     return true;
   }
+
+  bool scoreHigherThanFDR_(const PeptideHit& hit, double FDR, bool q_value_score)
```
jpfeuffer, line 443: I would probably call it scoreBetterThanFDR_ since the score is actually lower.

src/topp/DatabaseSuitability.cpp
```diff
@@ -220,6 +225,9 @@ class DatabaseSuitability :
       // skip if the top hit is a decoy hit
       if (top_hit.getMetaValue("target_decoy") == "decoy") continue;
 
+      // skip if top hit is out ouf FDR
```
jpfeuffer, line 228: The new statements and maybe some old stuff assumes that the hits are sorted. I wonder if we should add a Precondition or make sure that this is actually the case.

src/tests/topp/DatabaseSuitability_out_1.tsv
```diff
@@ -1,7 +1,7 @@
 key	value
-#top_db_hits	438
+#top_db_hits	430
```
jpfeuffer, line 2: Only 8 hits are filtered at a 1% FDR?

src/topp/DatabaseSuitability.cpp
Waschi97: Yeah.. I think I changed the name but forgot to change the return values. Since I use this to check if a hit should be skipped, I thought `scoreHigherThanFDR_` would be the better name. But you are right: true and false should be the other way around.

src/topp/DatabaseSuitability.cpp
Waschi97: Yes, this would probably be a good idea.
Is there a convenient way of doing this? Of cause I can just check while calculating the decoy cut-off..

src/tests/topp/DatabaseSuitability_out_1.tsv
Waschi97: Actually because the return calues `scoreHigherThanFDR_` are the wrong way around, there will only be 8 hits left. Which makes me want to add a seperate test for the FDR parameter..

jpfeuffer: Hm yes or just change the FDR parameter for the tests.

src/topp/DatabaseSuitability.cpp
jpfeuffer: I think a hits.sort() before you do anything on the hits should be fine. Not too much overhead.

src/topp/DatabaseSuitability.cpp
```diff
@@ -209,35 +214,64 @@ class DatabaseSuitability :
...
+          });
+      }
+      else
+      {
+        if (!hits[0].metaValueExists("q-value"))
+        {
+          throw(Exception::Precondition(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No q-value found at peptide identification nor at peptide hits. Make sure 'False Discovery Rate' is run beforehand."));
+        }
+
+        sort(hits.begin(), hits.end(),
```
jpfeuffer, line 240: Ah yes you want to sort by q-value.. I think the ranking should be the same with the normal score. Then you can use pep_id.sort(). Which is more concise and a bit faster (since it does not access meta values). But yes, maybe this is a bit safer. Up to you since it is probably not a bottleneck.

Waschi97: While the ranking should be the same for the score from which the q-value was calculated, I'm not sure that this is necessarily the case for any other score. That's why I decided to make sure the sorting is done by q-value.

jpfeuffer:
Ok. Will merge if tests pass.

PR #4815, opened 2020-07-16
change spectrum index to integer type in novoradapter

timosachsenberg:


timosachsenberg:
forgot test file... will add

jpfeuffer:
This might fail now.

PR #4820, opened 2020-07-20
[FIX] PR enforcing equal column sizes showed bug probably introduced …

jpfeuffer:
…in stream export.


jpfeuffer:
Unfortunately, there are still many errors with the exported mzTabs.
Even if we only export one search engine score, the order is not correct:
If in the header it is
1) MSGF
2) Percolator

In the PEP section we report search_engine_score[1] with the Percolator score.

For TMT specifically:
PEP:
- search_engine is always "null"
- search_engine_ms_run is always "null"

jpfeuffer:
Let's merge. I will open an issue with the remaining problems.

Issue #4817, opened 2020-07-17
failure to execute database search 


pavel-shliaha:
I have installed openMS on LINUX KNIME and downloaded the latest proteomics_LFQ workflow. 

https://github.com/OpenMS/Tutorials/tree/master/Workflows

The modification I made is insert DecoyDatabase node. Both search engines in them fail. The console log is:

WARN  Input File           0:146      File does not exist.
WARN  FontStore                       Using the system default font for annotations: Font {140719039767808}
ERROR OMSSAAdapter         0:180:98   Failing process stdout: [Warning: Parameters file version (2.3.0) does not match the version of this tool (2.5.0)., Your current parameters are still valid, but there might be new valid values or even new parameters. Upgrading the INI might be useful., OMSSAAdapter took 0.07 s (wall), 0.00 s (CPU), 0.00 s (system), 0.00 s (user).]
ERROR OMSSAAdapter         0:180:98   Failing process stderr: [Input database '/tmp/knime_Proteomics_LFQ21238/fs-Decoy_0-181-21239/000/000/DecoyDatabase_0/Human_rev_can_uniprot_24_09_15.fasta.psq' not found (the file '/tmp/knime_Proteomics_LFQ21238/fs-Decoy_0-181-21239/000/000/DecoyDatabase_0/Human_rev_can_uniprot_24_09_15.fasta.psq' could not be found). Make sure it exists (and check 'OpenMS.ini:id_db_dir' if you used relative paths. Aborting!, Unable to find database '/tmp/knime_Proteomics_LFQ21238/fs-Decoy_0-181-21239/000/000/DecoyDatabase_0/Human_rev_can_uniprot_24_09_15.fasta.psq' (searched all folders). Did you mistype its name?]
ERROR OMSSAAdapter         0:180:98   Return code: 6
ERROR OMSSAAdapter         0:180:98   Execute failed: Failed to execute node OMSSAAdapter
[...]

PR #4822, opened 2020-07-21
Doc extra

cbielow:
this PR adds some extra information for the user in multiple places:

1) amend our docs, for the zlib error of X!Tandem in #4817
2) write the search engine used in idXML files when calling FileInfo
3) some more stats when calling InternalCalibration (which calibration points were excluded and why)
4) allow MSGF+Adapter to pass a `conf` file with additional parameters

### DIFF
--- a/src/topp/MSGFPlusAdapter.cpp
+++ b/src/topp/MSGFPlusAdapter.cpp
@@ -77,17 +77,28 @@
 </CENTER>
 
     MS-GF+ must be installed before this wrapper can be used. Please make sure that Java and MS-GF+ are working.@n
-    The following MS-GF+ version is required: MS-GF+ 2019/07/03. At the time of writing, it could be downloaded from https://github.com/MSGFPlus/msgfplus/releases. Older versions will not work properly.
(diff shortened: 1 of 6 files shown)


cbielow:
tests pass.
Comments?

src/topp/MSGFPlusAdapter.cpp
```diff
@@ -208,6 +219,8 @@ class MSGFPlusAdapter :
 
     registerFlag_("legacy_conversion", "Use the indirect conversion of MS-GF+ results to idXML via export to TSV. Try this only if the default conversion takes too long or uses too much memory.", true);
 
+    registerInputFile_("conf", "<file>", "", "Optional MSGF+ configuration file (passed as -conf <file> to MSGF+). See documentation for examples.", false, false);
```
timosachsenberg, line 222: do we need to document here how deviating options set on command line and in config are handled?

cbielow: I briefly looked at the msgf docs but no info.
We'd need to test that (hoping they do not change the behavior)

timosachsenberg: one way would be to allow either conf or the other params

PR #4829, opened 2020-07-28
[FIX] ConsenusID reading from wrong meta data

jpfeuffer:
In the case of a search engine overwritten from Percolator, it read and wrote the wrong meta values (at the ProteinID, not the SearchParams)


jpfeuffer:
This should be small enough to pass without CI, which is currently down for updates.

jpfeuffer:
Dont merge yet.

jpfeuffer:
Now it could be merged.

PR #4831, opened 2020-07-29
Minor INI/TOPPAS improvements

cbielow:
- make the initial Parameter editor big enough
 - label store and load buttons for INI files more precisely


timosachsenberg:
+1

PR #4832, opened 2020-07-30
Labeling support

cbielow:
* unify the interface for Modifications a bit
* Add new UTIL 'StaticModification'

sidenote: the ResidueModification and Residue classes need a general rewrite. There are no invariants one can rely on, and everyone uses them as they please. This probably harbors some serious bugs...

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -39,6 +39,7 @@
 #include <OpenMS/CHEMISTRY/ResidueModification.h>
 
 #include <set>
+#include <memory>  // unique_ptr
 #include <unordered_map>
 
 namespace OpenMS
@@ -146,9 +147,12 @@ public:
 
     /**
        @brief Add a new modification to ModificationsDB.
-       Will skip adding if modification already exists (based on its fullID)
+       If the modification already exists (based on its fullID) it is not added.
+       The function returns a pointer to the modification in the ModificationDB (which can be differ from input if mod was already present).
+
+       @param new_mod Owning pointer, which transfers ownership to ModificationsDB (mod might get deleted if already present!)
     */
-    void addModification(ResidueModification * new_mod);
+    const ResidueModification* addModification(std::unique_ptr<ResidueModification> new_mod);
 
     /**
        @brief Returns the index of the modification in the mods_ vector; a unique name must be given
--- a/src/openms/include/OpenMS/CHEMISTRY/Residue.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/Residue.h
@@ -36,7 +36,6 @@
 #pragma once
 
 #include <OpenMS/CHEMISTRY/EmpiricalFormula.h>
-#include <OpenMS/CHEMISTRY/ResidueModification.h>
 #include <OpenMS/CONCEPT/Types.h>
 #include <OpenMS/DATASTRUCTURES/String.h>
 
@@ -46,6 +45,8 @@
 
 namespace OpenMS
 {
+  class ResidueModification;
+
   /**
       @ingroup Chemistry
 
@@ -396,6 +397,10 @@ public:
     /// helper for mapping residue types to letters for Text annotations and labels
     static char residueTypeToIonLetter(const ResidueType& res_type);
 
+    /// Write as Origin+Modification, e.g. M(Oxidation), or X[945.34] or N[+14.54] for user-defined mods.
+    /// This requires the Residue to have a valid OneLetterCode and an optional (but valid) ResidueModification (see ResidueModification::toString())
+    String toSequence() const;
+
     /// ostream iterator to write the residue to a stream
     friend OPENMS_DLLAPI std::ostream& operator<<(std::ostream& os, const Residue& residue);
 
@@ -469,6 +474,7 @@ protected:
     double internal_to_z_monoweight_ = getInternalToZIon().getMonoWeight();
   };
 
+  // write 'name threelettercode onelettercode formula'
   OPENMS_DLLAPI std::ostream& operator<<(std::ostream& os, const Residue& residue);
 
 }
--- a/src/openms/include/OpenMS/CHEMISTRY/Residue.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/Residue.h
@@ -304,6 +304,9 @@ public:
     /// sets the modification by name; the mod should be present in ModificationsDB
     void setModification(const String& name);
 
+    /// sets the modification by existing ResMod (make sure it exists in ModificationDB)
+    void setModification(const ResidueModification* mod);
+    
     /// returns the name (ID) of the modification, or an empty string if none is set
     const String& getModificationName() const;
 
@@ -464,10 +467,6 @@ protected:
     double internal_to_x_monoweight_ = getInternalToXIon().getMonoWeight();
     double internal_to_y_monoweight_ = getInternalToYIon().getMonoWeight();
     double internal_to_z_monoweight_ = getInternalToZIon().getMonoWeight();
-
-    /// sets the modification (helper function)
-    void setModification_(const ResidueModification& mod);
-
   };
 
   OPENMS_DLLAPI std::ostream& operator<<(std::ostream& os, const Residue& residue);
--- a/doc/doxygen/public/UTILS.doxygen
+++ b/doc/doxygen/public/UTILS.doxygen
@@ -83,6 +83,7 @@
   - @subpage UTILS_SequenceCoverageCalculator - Prints information about idXML files.
   - @subpage UTILS_SpecLibCreator - Creates an MSP-formatted spectral library.
   - @subpage UTILS_SpectraSTSearchAdapter - An interface to the 'SEARCH' mode of the SpectraST program (external, beta).
+  - @subpage UTILS_StaticModification - Allows to attach a set of fixed modifications to an idXML file (MS/MS search results), e.g. to add 15N (N15) labeling post-hoc.
 
   <b>Cross-linking</b>
   - @subpage UTILS_RNPxlSearch - Annotates RNA-to-peptide cross-links in MS/MS spectra.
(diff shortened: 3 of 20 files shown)


src/topp/CometAdapter.cpp
```diff
@@ -141,43 +141,26 @@ class TOPPCometAdapter :
...
-    setMaxInt_("min_peptide_length", 63);
-    registerIntOption_("max_peptide_length", "<num>", 63, "Maximum peptide length to consider.", false);
-    setMinInt_("max_peptide_length", 5);
-    setMaxInt_("max_peptide_length", 63);
-
     //Fragment Ions
-    registerDoubleOption_("fragment_bin_tolerance", "<tolerance>", 0.02, "Bin size (in Da) for matching fragment ions. Ion trap: 1.0005, high res: 0.02. CAUTION: Low tolerances have heavy impact on RAM usage. Consider using use_sparse_matrix and/or spectrum_batch_size.", false, true);
-    setMinFloat_("fragment_bin_tolerance", 0.01);
-    registerDoubleOption_("fragment_bin_offset", "<fraction>", 0.0, "Offset of fragment bins scaled by tolerance. Ion trap: 0.4, high res: 0.0.", false, true);
+    registerDoubleOption_("fragment_mass_tolerance", "<tolerance>", 0.3,
```
timosachsenberg, line 154: is it a good idea to change to lowers defaults?

src/topp/CometAdapter.cpp
```diff
@@ -141,43 +141,26 @@ class TOPPCometAdapter :
...
+                          "Fragment mass error (one-sided)."
+                          "CAUTION: Low tolerances have heavy impact on RAM usage (since Comet uses a lot of bins in this case). Consider using use_sparse_matrix and/or spectrum_batch_size.", false);
+    setMinFloat_("fragment_mass_tolerance", 0.01);
+    registerStringOption_("fragment_error_units", "<unit>", "Da", "Fragment monoisotopic mass error units", false);
+    setValidStrings_("fragment_error_units", { "Da" }); // only Da allowed
+    registerDoubleOption_("fragment_bin_offset", "<fraction>", 0.0, "Offset of fragment bins. Recommended by Comet: low-res: 0.4, high-res: 0.0", false);
```
timosachsenberg, line 160: this is the offset for high-res but tolerance up there is for low-res ?

src/topp/CometAdapter.cpp
```diff
@@ -429,15 +429,21 @@ class TOPPCometAdapter :
     // high res ms/ms:    0.02 tolerance, 0.0 offset (mono masses), theoretical_fragment_ions = 0
 
     String instrument = getStringOption_("instrument");
-    double bin_tol = getDoubleOption_("fragment_bin_tolerance");
+    double bin_tol = getDoubleOption_("fragment_mass_tolerance") * 2; // convert 1-sided tolerance to bin size
```
timosachsenberg, line 432: is this an ad-hoc solution?

src/topp/CometAdapter.cpp
cbielow: this is what other adapters have as well (xtandem, for example).
But given the fact that this is a bin-width, one might be off the center, so yeah... lets go with 0.5, which is what is was before.

src/topp/CometAdapter.cpp
cbielow: good catch

src/topp/CometAdapter.cpp
cbielow: no, just to make the parameters between engines more comparable.

cbielow:
ready for merge.
lets wait for CI to come back up...

cbielow:
rebuild jenkins

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -146,9 +147,12 @@ namespace OpenMS
...
     /**
        @brief Add a new modification to ModificationsDB.
-       Will skip adding if modification already exists (based on its fullID)
+       If the modification already exists (based on its fullID) it is not added.
+       The function returns a pointer to the modification in the ModificationDB (which can be differ from input if mod was already present).
+
+       @param new_mod Owning pointer, which transfers ownership to ModificationsDB (mod might get deleted if already present!)
     */
-    void addModification(ResidueModification * new_mod);
+    const ResidueModification* addModification(std::unique_ptr<ResidueModification> new_mod);
```
timosachsenberg, line 155: does std::unique_ptr work with python?

cbielow: there was no binding for `addModification` before, so I assume it did not even work with pointers.

timosachsenberg:
@hroest can we wrap functions that take std::unique_ptr<.> parameters?

hroest:
@timosachsenberg I dont think so, we currently can do raw ptrs and we can do shared_ptrs - we would have to write a new class here: https://github.com/uweschmitt/autowrap/blob/master/autowrap/ConversionProvider.py#L1546

hroest:
I dont think that modelling unique_ptr with pyOpenMS will be easy due to how unique_ptr conceptually work. Basically the idea of this approach would be to pass ownership and "give it up". Lets see how that would work in Python


```
r = ResidueModification() # Creates a new C++ "ResidueModification" and a Python object to hold it
db.addModification(r) 
```

Note that here `r` has a member `shared_ptr<ResidueModification>`. So, what state should the object `r` be in after this operation? Even if we can make it work, what does a Python user do with this object afterwards? can they still use `r` or will it be destroyed and how is that indicated/modelled? A lot of the guarantees that work in the C++ world are not carried to Python and need to be handled explicitly. How would this work:

```
r = ResidueModification()
r2 = ResidueModification(r)
db.addModification(r) 
```

where now both `r` and `r2` point to the same underlying C++ object (using their shared_ptr)? We cannot (easily) convert shared_ptr to unique ptr though! I guess we could try and use `std::shared_ptr<T>::use_count` and if equal to one then convert it to a unique_ptr but its a lot of work [1] 

1. https://en.cppreference.com/w/cpp/memory/shared_ptr/use_count

hroest:
I guess we could create a copy and that should work?

```
unique_ptr<ResidueModification>tmp (new ResidueModification(self.inst.get()) );
db.addModification(std::move(tmp)); 
```

@cbielow any concerns?

cbielow:
rebuild jenkins

cbielow:
> 
> 
> I guess we could create a copy and that should work?
> 
> ```
> unique_ptr<ResidueModification>tmp (new ResidueModification(self.inst.get()) );
> db.addModification(std::move(tmp)); 
> ```
> 
> @cbielow any concerns?

Where would this be implemented? As an add-on?
And we'd need Autowrap support for unique_ptr first, right?

cbielow:
ready to merge for now.
The python bindings need a bit more preparation on other sides, so lets deal with that in a separate PR

share/OpenMS/CHEMISTRY/XTandem_default_input.xml
```diff
@@ -76,7 +76,7 @@
   <note type="input" label="protein, cleavage semi">no</note>
   <note type="input" label="protein, quick acetyl">yes</note>
   <note type="input" label="protein, quick pyrolidone">yes</note>
-  <note type="input" label="protein, modified residue mass file"></note>
+  <note type="input" label="protein, modified residue mass file"></note>               !!!!! fill in a filename here for 15N labeling etc.. make sure the PATH is valid, or Tandem will silently ignore it!
```
timosachsenberg, line 79: what is this?

src/openms/include/OpenMS/CHEMISTRY/Residue.h
```diff
@@ -393,6 +397,10 @@ namespace OpenMS
     /// helper for mapping residue types to letters for Text annotations and labels
     static char residueTypeToIonLetter(const ResidueType& res_type);
 
+    /// Write as Origin+Modification, e.g. M(Oxidation), or X[945.34] or N[+14.54] for user-defined mods.
+    /// This requires the Residue to have a valid OneLetterCode and an optional (but valid) ResidueModification (see ResidueModification::toString())
+    String toSequence() const;
```
timosachsenberg, line 402: toString ?

src/topp/XTandemAdapter.cpp
```diff
@@ -101,6 +101,14 @@ using namespace std;
     If you want to use the XML configuration file and @em ignore most of the parameters set via this adapter, use the @p ignore_adapter_param flag.
     Then, the config given via @p default_config_file is used exclusively and only the values for the paramters @p in, @p out, @p database and @p xtandem_executable are taken from this adapter.
 
+    Hint: this adapter supports 15N labeling by specifying
+@code
+<note type="input" label="protein, modified residue mass file">/path/to/bioml.xml</note>
```
timosachsenberg, line 106: is the > in the middle correct?

hroest:
> Where would this be implemented? As an add-on?

yes, you would use `no-wrap` in the pxd file and then write an addon doing this. Similar to https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/addons/AASequence.pyx

share/OpenMS/CHEMISTRY/XTandem_default_input.xml
cbielow: a comment :)
since its chars between tags, the parser will ignore it

src/topp/XTandemAdapter.cpp
cbielow: yes

src/openms/include/OpenMS/CHEMISTRY/Residue.h
cbielow: I thought about that too, but I went for `toSequence` to make clear that its not just any string, but a proper sequence, which can be parsed again

timosachsenberg: hmm I don't like the name sequence for that... we have toString() and toUnmodifiedString() for AASequence
but never used sequence

timosachsenberg: alternatively to toString()
is there something we already use?
Also, note that we have on the Modification site something related (but different) string
M(Oxidation) vs Oxidation (M)
Would be good to have more telling names here in general

cbielow:
ready for merge

Issue #4835, opened 2020-08-02
mzTabExporter cannot deal with custom modifications


cbielow:
When generating a cXML with custom mods, currently (at least two) issue prevent the exporter to write an mzTab:

1)
```
const class OpenMS::ResidueModification *__cdecl OpenMS::ModificationsDB::searchModificationsFast(const class OpenMS::String &,bool &,const class OpenMS::String &,enum OpenMS::ResidueModification::TermSpecificity) constModification not found: C+58.018500
Error: Unexpected internal error (the value 'C+58.018500' was used but is not valid; Retrieving the modification failed. It is not available for the residue '' and term specificity 'none'. )
```

occurs because it tries to find a modification `C+58.018500` as obtained from `      <FixedModification name="C+58.018500" />`
which does not exist...
Error is raised here: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/MzTab.cpp#L1497

2)
After (hackily) deleting the line from the input, one gets yet another error:

```
Error: Unexpected internal error (Modification or Substitution identifier MUST NOT be null or empty in MzTabModification)

```

since the modification obviously also appears in PepIDs as custom mod, e.g. `<PeptideHit score="0.039528211667198" sequence="GC[161.027684999999991]TPEPTID" ...`.

I cannot really track what the actual problem is though...

[error.zip](https://github.com/OpenMS/OpenMS/files/5012508/error.zip)

Exemplary consensusXML attached (note: heavily reduced in size to reproduce the bug -- maybe some other downstream things do not work with this consensusXML...); the error should be easy to reproduce though using any other cxml by adding a custom mod anywhere.

jpfeuffer:
Regarding 1) Not sure if this is an mztab problem. I think we nowhere handle unknown modifications in the SearchParams object. I assume most if not all pieces of code will call getModification like in your first code snippet and fail. I think getModification needs to be adapted to parse these kinds of strings and add the missing mod to the DB if not present yet.
Regarding 2) the main problem is that mztab usually expects a unimod identifier for a mod which we were enforcing. It needs to be discussed how to handle that.

jpfeuffer:
@cbielow By the way, where did you get these unknown Modifications from? Comet with own params file?

cbielow:
> @cbielow By the way, where did you get these unknown Modifications from? Comet with own params file?
> Because if those mods are in unimod, you should properly look them up during pepXML parsing if possible.

[...]

PR #4841, opened 2020-08-03
add unknown modifications as chemmod

timosachsenberg:
Improved support for unknown modification export: 
https://github.com/OpenMS/OpenMS/issues/4835


hroest:
any chance you can add a test for this?

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -3027,10 +3053,20 @@ Not sure how to handle these:
         if (!is_fixed)
         {
           String unimod = res_mod.getUniModAccession();
+
```
cbielow, line 3056: this whole block looks awfully like copy and pasted a couple of times. Can you make that a function?

cbielow: :)

timosachsenberg: when I am through with the other issues :P

Issue #4834, opened 2020-07-31
OMSSAAdapter writes incomplete idXML


cbielow:
all `<PeptideIdentification>` nodes are missing the `spectrum_reference="controllerType=0 controllerNumber=1 scan=11608"` attribute

all `</ProteinIdentification>` nodes are missing the`<UserParam type="stringList" name="spectra_data" value="[....]"/>`
		

Thus, downstream tools such as `QualityControl` will fail

jpfeuffer:
I assume it is fixed with the PR. Please reopen if not.

cbielow:
confirmed fixed.

PR #4852, opened 2020-08-04
write out native ids and spectrum references

timosachsenberg:
add references to original data and native ids

### DIFF
--- a/src/topp/OMSSAAdapter.cpp
+++ b/src/topp/OMSSAAdapter.cpp
@@ -1032,7 +1032,7 @@ protected:
     {
       Size scan_id = lookup.findByRT(p.getRT());
       PeakSpectrum& temp = exp.getSpectrum(scan_id);
-      p.setMetaValue("spectrum_reference", temp.getIdentifier());
+      p.setMetaValue("spectrum_reference", temp.getNativeID());
     }
 
     //-------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


timosachsenberg:
might fix https://github.com/OpenMS/OpenMS/issues/4834

src/topp/OMSSAAdapter.cpp
```diff
@@ -1018,6 +1019,33 @@ class TOPPOMSSAAdapter :
...
+    // reannotate file oriin and native ids
+    MSExperiment exp;
+    MzMLFile().load(inputfile_name, exp);
+    protein_identification.setPrimaryMSRunPath({inputfile_name}, exp);
+
+    // add RT and precursor m/z to the peptide IDs (look them up in the spectra):
+    SpectrumLookup lookup;
+    lookup.readSpectra(exp);
+
+    for (auto& p : peptide_ids)
```
cbielow, line 1031: this for loop can be refactored into a member of SpectrumLookup?!
I think there are at least 2 copies of this code in place, which we could get rid of.
I can do that

timosachsenberg: this would be nice! thanks

timosachsenberg:
please don't merge... found an issue

jpfeuffer:
Looks good. Merge if you are ready.

Issue #4825, opened 2020-07-23
Warn in PSMFeatureExtractor already if an IdXML contains multiple runs


jpfeuffer:
Which are not supported yet.

jpfeuffer:
We do that in PercolatorAdapter already, but since PSMFeatureExtractor comes first it should fail/warn there already.

PR #4853, opened 2020-08-04
Warn about multiple runs when using PSMFeatureExtractor

Leon-Bichmann:
# Description

The PR is adding a warning about the use of an idXML resulting from multiple merged runs. One should use the Flag merge_proteins_add_psms.

#4825 

Not tested yet, will check once more before merging


jpfeuffer:
Feel free to merge, when tested.

Leon-Bichmann:
ok tested it on a local built on OSX and it works.

Issue #4828, opened 2020-07-27
TOPPView: error in projections 


JeeH-K:
In TOPPView 2.5 version, while showing mz and rt positions in the corner where mouse hovers (on the top and right panel of the main view), it shows as 'm/z' where it is rt value, and vice versa.
<img width="986" alt="Screen Shot 2020-07-27 at 12 47 58 PM" src="https://user-images.githubusercontent.com/7207883/88533965-68cf8380-d007-11ea-93c1-f91b2055d39d.png">

jpfeuffer:
I see. Should be a quick fix from one of our GUI guys?

timosachsenberg:
the text for the legend probably needs to be changed here: https://github.com/OpenMS/OpenMS/blob/d17cc251fd0c4068eb253b03c9fb107897771fdc/src/openms_gui/source/VISUAL/Spectrum2DWidget.cpp#L86
and here:
https://github.com/OpenMS/OpenMS/blob/d17cc251fd0c4068eb253b03c9fb107897771fdc/src/openms_gui/source/VISUAL/Spectrum2DWidget.cpp#L369

enetz:
I found the place where the labels are decided here:
https://github.com/OpenMS/OpenMS/blob/d17cc251fd0c4068eb253b03c9fb107897771fdc/src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp#L1241-L1250

The problem is that this currently ignores the mapping of the axes of the main window. So if the axes on the MS1 map are switched, it fits. I can also flip them, so that the work with the default settings, but not with flipped axes.
I couldn't find a way yet to let the `Spectrum1DCanvas`es of the projections know the axis settings of the MS1 map.
`isMzToXAxis()` seems to do something else.

hroest:
>  I can also flip them, so that the work with the default settings, but not with flipped axes.

does this mean the problem still exists?

enetz:
@hroest no, that comment is outdated, I found a way to adapt the labels to the orientation of the MS1 map.
It is described in the merged PR.

PR #4855, opened 2020-08-04
[FIX, GUI] fixes OpenMS#4828, wrong labels in the projections

enetz:
Switched the "m/z" and "RT" labels in `Spectrum1DCanvas.cpp`
This fixes the wrong labels in the projections as reported in #4828.

Previously `projection_horz_->canvas()->setSwappedAxis(true)` and `projection_vert_->canvas()->setSwappedAxis(true)` were always set. This caused the labels to be constant and independent from the orientation of the axes in the main MS1 map.
Now I set these both to `false` for one of the cases and the labels adapt to the MS1 map orientation correctly.


timosachsenberg:
sounds legit

timosachsenberg:
@JeeH-K may also give a look once it is merged

JeeH-K:
it worked like a charm! Thanks

PR #4862, opened 2020-08-05
add missing cachedMzML file type to parameter

bernt-matthias:
One question am I right that cachedMzML corresponds to to CML files with root tag the `indexedMzML`.

# Description

Also please:

# Checklist:

### DIFF
--- a/src/topp/OpenSwathAnalyzer.cpp
+++ b/src/topp/OpenSwathAnalyzer.cpp
@@ -150,7 +150,7 @@ protected:
                            "Swath files that were used to extract the transitions. "
                            "If present, SWATH specific scoring will be used.",
                            false);
-    setValidFormats_("swath_files", ListUtils::create<String>("mzML"));
+    setValidFormats_("swath_files", ListUtils::create<String>("mzML,cachedMzML"));
 
     registerDoubleOption_("min_upper_edge_dist", "<double>", 0.0,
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Unfortunately I don't think so.
Indexed mzml is often not declared in the file extension. I think it is officially an extension to the mzML standard.
CachedMzml from my understanding was a predecessor that has an (index?) cache in another file that it expects in the same folder: 'name.mzML.cache'

But let's also wait what @hroest says about this.

timosachsenberg:
I think chachedMzML is a custom binary representation of the mzML so it should be much faster.

hroest:
@bernt-matthias unfortunately this PR does not work as intended. At the moment there is no separate extension for the cached and non-cached version of an mzML file:

* Case 1: you have a single mzML file with all data: file.mzML
* Case 2: you have two files that are linked where the data is in the cached file: file.mzML + file.cachedMzML

in the second case, there is a "special" tag in the XML of file.mzML that makes it recognize that there is no data in the file and it will look for a file called "file" + ".cachedMzML" which should contain the data. So simply by looking at the extension of a file you cannot tell whether its Case 1 or 2. Also, you should never pass the "file.cachedMzML" into OpenMS since it will not be able to read it. We cannot merge this as it does not work as intended.

timosachsenberg:
close for now... feel free to reopen

PR #4863, opened 2020-08-05
start to unify traML vs TraML

bernt-matthias:
becomes anoying for autoconverters

seemingly traML is less common so I went for the majority

Seems that there are still some places left...

# Description

Also please:

# Checklist:

### DIFF
--- a/src/topp/MRMMapper.cpp
+++ b/src/topp/MRMMapper.cpp
@@ -124,7 +124,7 @@ protected:
     setValidFormats_("in", ListUtils::create<String>("mzML"));
 
     registerInputFile_("tr", "<file>", "", "transition file");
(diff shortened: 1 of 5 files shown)


hroest:
at least in OpenMS, currently `traML` is more often used:

```
$ grep -R setValidF src/topp/ src/utils/ | grep -c TraML
5
$ grep -R setValidF src/topp/ src/utils/ | grep -c traML
15
```

timosachsenberg:
me too. better to go for traML

bernt-matthias:
Maybe it would be a good idea then to check for file extensions in a case insensitive manner? Then one would have backward compatibility and no one needs to rename files and also automatic tool description converters would be happy.

timosachsenberg:
thanks!

PR #4864, opened 2020-08-06
[FIX] suppress misleading warning messages about unknown parameters

### LINKED ISSUE #4546: OpenPepXL: unknown or deprecated parameters
I see messages like `Unknown (or deprecated) Parameter 'in' given in outdated parameter file! Ignoring parameter.` when running 

```
OpenPepXL -test -decoy_string "decoy" -in OpenPepXL_input.mzML -consensus OpenPepXL_input.consensusXML -database OpenPepXL_input.fasta -out_xquestxml OpenPepXL_output.xquest.xml.tmp -out_xquest_specxml OpenPepXL_output.spec.xml.tmp -out_mzIdentML OpenPepXL_output.mzid.tmp -out_idXML OpenPepXL_output.idXML.tmp -debug 3
```

which is odd since I do not use a parameter file. The messages are also there if I use a parameter file generated with `OpenPepXL -write_ctd ./`. Testing currently in 2.5.0.

I'm asking in particular because the output also indicates that the `test` parameter is ignored even if I set it in the parameter file .. which would explain the differences in the output data that I observe in my Galaxy tool tests.

enetz:
The OpenPepXL, OpenPepXLLF and PeptideIndexer tools pass all input parameters on to their algorithm classes.
This includes the TOPPBase parameters like `threads` and `test` as well as inout and output file parameters. The algorithm classes don't know about these parameters and a warning about unknown or deprecated parameters is thrown when `debug != 0`.

This PR is a simple workaround that removes those unknown parameters from the `Param` object that is passed to the algorithm.
This method is already used in e.g. `FeatureFinderMultiplex`.

The better option (at least for new tools) would be to add a node to the parameters for each tool that includes all parameters needed for the respective algorithm class, but that would add prefixes to most of their parameters and break all existing `.ini` files and workflows.

Fixes #4546

### DIFF
--- a/src/topp/OpenPepXL.cpp
+++ b/src/topp/OpenPepXL.cpp
@@ -231,9 +231,9 @@ protected:
     PeakMap spectra;
 
     OpenPepXLAlgorithm search_algorithm;
-    Param this_param = getParam_().copy("", true);
+    Param this_param = getParam_();
     Param algo_param = search_algorithm.getParameters();
-    algo_param.update(this_param, false, OpenMS_Log_debug); // suppress param. update message
+    algo_param.update(this_param, false, false, false, false, OpenMS_Log_debug); // suppress param. update message
     search_algorithm.setParameters(algo_param);
     search_algorithm.setLogType(this->log_type_);
 
--- a/src/topp/OpenPepXLLF.cpp
+++ b/src/topp/OpenPepXLLF.cpp
@@ -220,9 +220,9 @@ protected:
     PeakMap spectra;
 
     OpenPepXLLFAlgorithm search_algorithm;
-    Param this_param = getParam_().copy("", true);
+    Param this_param = getParam_();
     Param algo_param = search_algorithm.getParameters();
-    algo_param.update(this_param, false, OpenMS_Log_debug); // suppress param. update message
+    algo_param.update(this_param, false, false, false, false, OpenMS_Log_debug); // suppress param. update message
     search_algorithm.setParameters(algo_param);
     search_algorithm.setLogType(this->log_type_);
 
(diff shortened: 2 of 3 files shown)


src/topp/OpenPepXL.cpp
```diff
@@ -232,9 +232,26 @@ class TOPPOpenPepXL :
...
+    this_param.remove("out_idXML");
+    this_param.remove("out_mzIdentML");
+    this_param.remove("out_xquestxml");
+    this_param.remove("out_xquest_specxml");
+    this_param.remove("ini");
+    this_param.remove("log");
+    this_param.remove("debug");
+    this_param.remove("threads");
+    this_param.remove("no_progress");
+    this_param.remove("force");
```
cbielow, line 251: please don't do this.

this list is not tested against TOPPBase and its very verbose.

Better:
add a method to TOPPBase which returns a Param which contains the common parameters.
You need to convert between `Param` and `ParameterInformation`, but there is code for that: `TOPPBase::paramToParameterInformation_(const Param& param) `.

cbielow: ideally, you move the `TOPPBase::paramToParameterInformation_(const Param& param) ` into the `ParameterInformation` class, where it belongs while you are at it :-)

cbielow: other solution: find overlapping set by comparing default params of TOPP tool vs default params of algorithm.

enetz:
@cbielow ok, I tried another alternative. Before, PeptideIndexer and OpenPepXL already used the `Param::update(...)` method, but it just turned warnings into a softer type of warnings that were misleading and printed only in debug mode. I found that that only happens because the function is verbose by default. I now call the function with `verbose = false`. The messages are gone, but the behavior stays the same as before. Also the function seems to do what we want already: update algorithm parameters without adding the ones the algorithm doesn't know about.

cbielow:
clean solution. nice.

PR #4865, opened 2020-08-06
IDDatastructure example of adding PSMs

timosachsenberg:
The use case I put here would probably reflect the natural order a user might want to build (and update) the ID data structure.
Is this possible? How?

### DIFF
--- a/src/tests/class_tests/openms/source/IdentificationData_test.cpp
+++ b/src/tests/class_tests/openms/source/IdentificationData_test.cpp
@@ -528,13 +528,19 @@ START_SECTION((static bool isBetterScore(double first, double second, bool highe
 }
 END_SECTION
 
-
 START_SECTION(([EXTRA] UseCaseBuildBottomUpProteomicsID()))
+{
   IdentificationData id;
   String file = "file://ROOT/FOLDER/SPECTRA.mzML";
   auto file_ref = id.registerInputFile(id);
  
+  // register a score type 
+  IdentificationData::ScoreType score("MySearchEngineScore", true);
+  auto score_ref = id.registerScoreType(score);
+
+  // register software (connected to score)
   IdentificationData::DataProcessingSoftware sw("MySearchEngineTool", "1.0");
+  sw.assigned_scores.push_back(score_ref);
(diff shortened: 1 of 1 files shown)


timosachsenberg:
Cool thanks. Can you add an example (I know this might be more involved given the design) how you
would e.g. assign all parent_matches of PSMs identified in (a) specific file and search. I would love to see how/if we can simplify how to deal with back-referencing

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -528,6 +529,84 @@ START_SECTION((static bool isBetterScore(double first, double second, bool highe
...
+  // some calculations, inference etc. could take place ...
+  IdentificationData::ParentMolecule protein("protein_1"); // accession is required
+  protein.sequence = "PRTTESTPEPTIDRPRT";
+  protein.description = "Human Random Protein 1";
+  auto protein_ref = id.registerParentMolecule(protein);
+  TEST_EQUAL(protein_ref->coverage, 0.0); // coverage not yet calculated
+
+  // add reference to parent (protein) and update peptide
+  IdentificationData::IdentifiedPeptide augmented_pep = *peptide_ref;
+  // @TODO: wrap this in a convenience function (like "match.addScore" above)
```
timosachsenberg, line 597: yes I think this would be beneficial

poshul:
Is this ready to get merged in yet?

hendrikweisser:
@poshul: I don't think this can be merged into "develop" because it depends on unmerged changes in my "idfile-integration" branch.

timosachsenberg:
In principle, we could promote this to a true PR, check if CI passes, and merge as the PR states "first example" and open one for each use-case.
Or we add to this PR. I don't mind.

poshul:
@hendrikweisser I was more wondering if it is complete enough that I can merge it into the branch I've been developing, or are there going to be significant updates.

jpfeuffer:
Do we still need this?

poshul:
I think we should merge it in if we can get all the builds to work. It's good reference for users.

PR #4866, opened 2020-08-06
DatabaseSuitability functions exported into the library

Waschi97:
- Added a new class `DBSuitability` which can be used to calculate database suitability.

- This class now also calculates q-values by itself.

- Added a second compute function to `Ms2IdentificationRate`, which works with a vector of PeptideIdentifications.
-> Also needed to add some new member functions to avoid code duplication.

- Changed the DatabaseSuitability TOPPTool to use these classes

- also in `Ms2IdentificationRate`: Changed the name of the parameter `no_fdr` to `no_index` since the metric doesn't need a q-value but target/decoy annotation. Therefore I also changed the name of this parameter in `QualityControl`.

- Test for new `DBSuitability` class

- Test the new compute function in `Ms2IdentificationRate`, change old tests to work with some of the changes

- Change TOPP docu of `DatabaseSuitability`


src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
```diff
@@ -69,6 +71,17 @@ namespace OpenMS
     /// container that stores results
     std::vector<IdentificationRateData> rate_result_;
 
+    /// returns number of all ms2 spectra in an MSExperiment
+    Size getMS2Count_(const MSExperiment& exp);
+
+    /// increases a 'counter' if PeptideIdentification is considered 'target' (target/decoy annotation needed)
```
cbielow, line 77: when referring to parameters, just `@p ...` to get proper highlighting, e.g. `@p ....`
but here, you can go classical
```
@param id ...
@param counter ...
``` 
to make it a bit more structured

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
```diff
@@ -69,6 +71,17 @@ namespace OpenMS
...
+    /// if force_index is true, 'counter' is increased regardless
+    /// is static so that it can be used with MapUtilities::applyFunctionOnPeptideIDs() without creating a new object for each ID
+    static void countPepID_(const PeptideIdentification& id, Size& counter, bool force_index);
```
cbielow, line 80: the interface is a bit... complicated?!
how about `static bool isTargetPeptide(PepID, count_unmatched)`
i.e. use the return value as counter and go
`counter += isTargetPeptide(...)` in calling code?

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
cbielow, line 77: document that you are only looking at the first hit....

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
```diff
@@ -69,6 +71,17 @@ namespace OpenMS
...
+
+    /// writes id-rate, num_ms2 and num_ids in IdentificationRateData object and appends it to result vector
+    void writeResults_(Size pep_ids, Size ms2_spectra);
```
cbielow, line 83: all these are counts right?
maybe `_count` as suffix?

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
```diff
@@ -77,20 +90,34 @@ namespace OpenMS
...
      *
      * stores results as a struct in a vector
-     * Only pep-ids with FDR metavalue annotation as 'target' are counted, unless force_fdr flag is set (assumes all pep-ids are target peptides)
+     * Only pep-ids with target/decoy annotation as 'target' are counted, unless force_index flag is set (assumes all pep-ids are target peptides)
      *
      * @param feature_map Input FeatureMap with target/decoy annotation
      * @param exp MSExperiment for counting number of MS2 spectra
-     * @param force_fdr Count all(!) PepIDs towards number of identified MS2 spectra (ignore target/decoy information if any)
-     * @exception Exception::MissingInformation is thrown if the FeatureXML is empty
+     * @param force_index Count all(!) PepIDs towards number of identified MS2 spectra (ignore target/decoy information if any)
```
cbielow, line 100: rename to `assume_all_target`?

src/openms/include/OpenMS/QC/Ms2IdentificationRate.h
```diff
@@ -77,20 +90,34 @@ namespace OpenMS
...
+
+    /**
+     * @brief computes Ms2 Identification Rate with PeptideIdentifications
+     *
+     * stores results as a struct in a vector
+     * Only pep-ids with target/decoy annotation as 'target' are counted, unless force_index flag is set (assumes all pep-ids are target peptides)
+     *
+     * @param pep_ids Input PeptideIdentifications with target/decoy annotation
+     * @param exp MSExperiment for counting number of MS2 spectra
+     * @param force_index Count all(!) PepIDs towards number of identified MS2 spectra (ignore target/decoy information if any)
```
cbielow, line 115: rename?

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,166 @@
...
+		* Counts top deNovo and top database hits. The ratio of db hits vs
+		* all hits yields the suitability.
+		* To re-rank cases, where a de novo peptide scores just higher than
+		* the database peptide, a decoy cut-off is calculated. This functionality
+		* can be turned off. This will result in an underestimated suitability,
+		* but it can solve problems like different search engines or to few decoy hits.
+		*
+		* Result is appended to the result member. This allows for multiple usage.
+		*
+		* @param pep_ids			vector containing pepIDs coming from a deNovo+database
```
cbielow, line 97: replace tabs with spaces

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,166 @@
...
+		*											identification search (currently only Comet-support)
+		* @throws							MissingInformation if decoy cut-off could not be calculated
+		* @throws							MissingInformation if no target/decoy annotation is found
+		* @throws							MissingInformation if no xcorr is found
+		* @throws							Precondition if FDR wasn't calculated
+		*/
+		void computeSuitability(std::vector<PeptideIdentification>& pep_ids);
+
+		/// return results
+		std::vector<SuitabilityData> getResults() const;
```
cbielow, line 107: return by `const&`

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+      }
+    }
+
+    if (double(diffs.size()) / pep_ids.size() < 0.2)
+    {
+      throw(Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Under 20 % of peptide identifications have two decoy hits. This is not enough for re-ranking. Use the 'force_no_re_rank' flag to still compute a suitability score."));
+    }
+
+    // sort the diffs decreasing and get the (1-novo_fract)*N one
+    auto sort_end = diffs.begin() + (1 - novo_fract) * diffs.size();
```
cbielow, line 279: using an iterator here is dangerous, since you do not know if its valid after the sorting (not sure if this is garanteed by the std.
Better just compute the index and do `begin() + index`.

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    partial_sort(diffs.begin(), sort_end + 1, diffs.end(), greater<double>());
```
cbielow, line 281: even better=faster should be https://en.cppreference.com/w/cpp/algorithm/nth_element.

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+      diff = abs(decoy_1 - decoy_2) / pep_id.getHits()[0].getSequence().getMonoWeight(); // normalized by mw
+    }
+
+    // if there aren't two decoy hits DBL_MAX is returned
+    return diff;
+  }
+
+  double Suitability::getDecoyCutOff_(const vector<PeptideIdentification>& pep_ids, double novo_fract)
+  {
+    // get all decoy diffs of peptide ids with at least two decoy hits
```
cbielow, line 262: double check that percentile is in the range [0,1]

src/openms/source/QC/Suitability.cpp
cbielow: and then check that the index is less than `diffs.size()` (segfaults otherwise)

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+    {
+      throw(Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Under 20 % of peptide identifications have two decoy hits. This is not enough for re-ranking. Use the 'force_no_re_rank' flag to still compute a suitability score."));
+    }
+
+    // sort the diffs decreasing and get the (1-novo_fract)*N one
+    auto sort_end = diffs.begin() + (1 - novo_fract) * diffs.size();
+
+    partial_sort(diffs.begin(), sort_end + 1, diffs.end(), greater<double>());
+
+    return *sort_end;
```
cbielow, line 283: use `diffs[index]` to avoid using an (invalid) iterator

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  bool Suitability::isNovoHit_(const PeptideHit& hit)
+  {
+    const set<String> accessions = hit.extractProteinAccessionsSet();
```
cbielow, line 288: taking by const& does not hurt (and might give speed if a non-temporary is returned one day)...

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+using namespace std;
+
+namespace OpenMS
+{
+  Suitability::Suitability()
+    : DefaultParamHandler("Suitability"), results_{}
+  {
+    defaults_.setValue("no_re_rank", "false", "Enable/Disable re-ranking");
+    defaults_.setValidStrings("no_re_rank", ListUtils::create<String>("true,false"));
```
cbielow, line 51: `{ "true", "false" }`

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+    defaults_.setValue("novo_fract", 1., "Fraction of how many cases, where a de novo peptide scores just higher than the database peptide, will be re-rank");
+    defaults_.setMinFloat("novo_fract", 0.);
+    defaults_.setMaxFloat("novo_fract", 1.);
+    defaults_.setValue("FDR", 0.01, "Filtering peptide hits based on this q-value");
```
cbielow, line 55: `[optional] FDR is only applied if input has not undergone FDR`

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+    defaultsToParam_();
+  }
+  
+  void Suitability::computeSuitability(vector<PeptideIdentification>& pep_ids)
+  {
+    bool no_re_rank = param_.getValue("no_re_rank").toBool();
+    double novo_fract = param_.getValue("novo_fract");
+    double FDR = param_.getValue("FDR");
+
+    SuitabilityData d;
```
cbielow, line 67: push back a result here, and work on a reference to the last item in the vector.
The default values should reflect that nothing could be computed

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    bool fdr_done = false;
+    bool q_value_score = false;
+    if (pep_ids[0].getScoreType() == "q-value")
```
cbielow, line 71: .. because then you can simply leave the function here 
`if (pep_ids.empty) return;` to avoid a segfault when checking the first element

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,316 @@
...
+      d.cut_off = getDecoyCutOff_(pep_ids, novo_fract);
+      if (d.cut_off == DBL_MAX)
+      {
+        throw Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Could not compute decoy cut off. Re-ranking impossible. If you want to ignore this, disable re-ranking.");
+      }
+    }
+
+    for (PeptideIdentification& pep_id : pep_ids)
+    {
+      vector<PeptideHit>& hits = pep_id.getHits();
```
cbielow, line 112: `pep_ids.sort()` and be done

src/topp/DatabaseSuitability.cpp
```diff
@@ -135,11 +136,11 @@ class DatabaseSuitability :
   // it gets automatically called on tool execution
   void registerOptionsAndFlags_() override
   {
-    registerInputFile_("in_id", "<file>", "", "Input idXML file from peptide search with combined database with added de novo peptide (after FDR)");
+    registerInputFile_("in_id", "<file>", "", "Input idXML file from peptide search with combined database with added de novo peptide. PeptideIndexer is needed, FDR is not.");
```
cbielow, line 139: FDR is optional

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    * To re-rank cases, where a de novo peptide scores just higher than
+    * the database peptide, a decoy cut-off is calculated. This functionality
+    * can be turned off. This will result in an underestimated suitability,
+    * but it can solve problems like different search engines or to few decoy hits.
+    * 
+    * Parameters can be set using the functionality of DefaultParamHandler.
+    * Parameters are:
+    *                no_re_rank - re-ranking can be turned off with this
+    *                novo_fract - percent of deNovo peptides to capture in re-ranking
+    *                FDR        - q-value that should be filtered for
```
cbielow, line 98: Maybe add `Preliminary tests have shown that DB suitability is rather stable across common FDR thresholds from 0-5%).` ?

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    *
+    * Result is appended to the result member. This allows for multiple usage.
+    *
+    * Attention: This function will calculate q-values for all peptide hits. Your input
```
cbielow, line 102: dangerous.
Commonly, one would pass by copy and document that this function needs to change the input.
Advanced users can then `std::move()` the argument into the function to avoid the copy (but loose their handle on the data).

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+    * @brief Calculates a xcorr cut-off based on decoy hits
+    *
+    * Decoy differences of all N pepIDs are calculated. The (1-novo_fract)*N highest
+    * one is returned.
+    * It is asssumed that this difference accounts for novo_fract of the re-ranking cases.
+    *
+    * @param pep_ids      vector containing the pepIDs
+    * @param novo_fract   fraction of how many cases, where a de novo peptide scores just higher than the database peptide, will be re-rank
```
cbielow, line 141: I still find this misleading, because this function only cares about decoy scores and has nothing to do with deNovo.
Rephrase to what the parameter actually does (and maybe rename it)?

Or maybe write both explanations, saying that the current one is the one from the paper?

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    * @returns            xcorr cut-off
+    * @throws             IllegalArgument if novo_fract isn't in range [0,1]
+    * @throws             IllegalArgument if novo_fract is to low for a decoy cut-off to be calculated
```
cbielow, line 144: `too low`

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    * @throws             MissingInformation if no more than 20 % of the pepIDs have two decoys in there top ten
```
cbielow, line 145: `their top10`

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    */
+    double getDecoyCutOff_(const std::vector<PeptideIdentification>& pep_ids, double novo_fract);
+
+    /**
+    * @brief Tests if a PeptideHit is considered a deNovo hit
+    *
+    * To test this the function looks into the protein accessions.
+    * If only the deNovo protein is found, 'true' is returned.
+    * If one database protein is found, 'false' is returned.
```
cbielow, line 154: `at least one database protein`

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    *
+    * @param hit      PepHit in question
+    * @returns        true/false
+    */
+    bool isNovoHit_(const PeptideHit& hit);
+
+    /**
+    * @brief Tests if a PeptideHit has a higher q-value the given FDR
```
cbielow, line 162: `has a higher q-value than a given FDR, i.e. does not pass FDR.`

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    *
+    * Q-value is searched at score and at meta-value level.
+    *
+    * @param hit            PepHit in question
+    * @param FDR            FDR to check against
+    * @returns              true/false
+    */
+    bool scoreHigherThanFDR_(const PeptideHit& hit, double FDR);
```
cbielow, line 170: `bool failsFDR(...)` ?

cbielow: or `passesFDR(...)` (usually one tries to use non-negative names)

src/openms/source/QC/Ms2IdentificationRate.cpp
```diff
@@ -149,4 +116,69 @@ namespace OpenMS
...
+    {
+      return false;
+    }
+    if (all_targets)
+    {
+      return true;
+    }
+    if (!(id.getHits()[0].metaValueExists("target_decoy")))
+    {
+      throw Exception::Precondition(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No target/decoy annotaion found. If you want to continue regardless use -MS2_id_rate:force_no_pep_index");
```
cbielow, line 155: `annotation`

cbielow: can you rename that `MS2_id_rate:force_no_pep_index` to `MS2_id_rate:assume_all_target` or so?

src/openms/source/QC/Ms2IdentificationRate.cpp
```diff
@@ -149,4 +116,69 @@ namespace OpenMS
...
+    }
+    if (id.getHits()[0].getMetaValue("target_decoy") == "target")
```
cbielow, line 157: `return (id.getHits()[0].getMetaValue("target_decoy") == "target")`?

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,294 @@
...
+using namespace std;
+
+namespace OpenMS
+{
+  Suitability::Suitability()
+    : DefaultParamHandler("Suitability"), results_{}
+  {
+    defaults_.setValue("no_re_rank", "false", "Enable/Disable re-ranking");
+    defaults_.setValidStrings("no_re_rank", { "true", "false" });
+    defaults_.setValue("novo_fract", 1., "Fraction of how many cases, where a de novo peptide scores just higher than the database peptide, will be re-rank");
```
cbielow, line 52: `deNovo` ?

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,294 @@
...
+    {
+      OPENMS_LOG_WARN << "No peptide identifications given to Suitability! No calculations performed." << endl;
+      return;
+    }
+
+    if (pep_ids[0].getScoreType() == "q-value")
+    {
+      throw Exception::Precondition(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "q-value found at PeptideIdentifications. That is not allowed! Please make sure FDR did not run previously.");
+    }
+    else
```
cbielow, line 81: no need for `else` after exception above.

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,294 @@
...
+    {
+      throw(Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Under 20 % of peptide identifications have two decoy hits. This is not enough for re-ranking. Use the 'force_no_re_rank' flag to still compute a suitability score."));
+    }
+
+    // sort the diffs decreasing and get the (1-novo_fract)*N one
+    UInt index = round((1 - novo_fract) * diffs.size());
+    
+    if (index >= diffs.size())
+    {
+      throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "'novo_fract' is set to low. Please set the parameter to a higher value.");
```
cbielow, line 268: `too low`

src/openms/source/QC/Suitability.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    nth_element(diffs.begin(), diffs.begin() + index + 1, diffs.end(), greater<double>());
```
cbielow, line 271: `index + 1`??
Then you must also lookup `diffs[index+1]`.
Please double check!

src/openms/include/OpenMS/QC/Suitability.h
jpfeuffer: ... threshold

src/topp/DatabaseSuitability.cpp
```diff
@@ -196,153 +189,38 @@ class DatabaseSuitability :
...
-
     for (const auto& pep_id : novo_peps)
     {
       if (pep_id.getHits().empty()) continue;
-      ++count_novo_seq;
       unique_novo.insert(pep_id.getHits()[0].getSequence());
     }
 
-    double id_rate = double(count_novo_seq) / count_ms2_lvl; // spectral quality (id rate of novo seqs)
+    Ms2IdentificationRate::IdentificationRateData quality = q.getResults()[0];
```
cbielow, line 210: name it `denovo_only_quality`?

src/topp/DatabaseSuitability.cpp
```diff
@@ -196,153 +189,38 @@ class DatabaseSuitability :
...
-      }
-    }
-
-    double suitability = double(count_db) / (count_db + count_novo); //db suitability
-
-    // spectra quality
-
-    Size count_novo_seq = 0;
+    Ms2IdentificationRate q;
+    q.compute(novo_peps, exp, true);
```
cbielow, line 201: add a sanity check that all PepIDs in `novo_peps` are actually deNovo protein hits?
Otherwise the user can accidentally pass the wrong idXML and hell will break loose?

src/openms/include/OpenMS/QC/Suitability.h
Waschi97: do this only in this function or change all occurrences?

jpfeuffer:
Looks good. Make it a non-draft as soon as it is ready from your side.

timosachsenberg:
can you check if these DeppCode warnings make sense https://github.com/OpenMS/OpenMS/pull/4529#issuecomment-672909519

Waschi97:
> can you check if these DeppCode warnings make sense [#4529 (comment)](https://github.com/OpenMS/OpenMS/pull/4529#issuecomment-672909519)

Those warnings make sense because usally one has to check a file for writability before writing into it. This is however done already when getting the string option.
But a check wouldn't hurt and would remove those warnings for good. I'll add it.

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+  *
+  * This class holds the functionality of calculating the database suitability.
+  * This can only be done if a combined deNovo+database identification search was performed.
+  * Currently only Comet search is supported.
+  *
+  * Allows for multiple usage, because results are stored internally and can be returned
+  * using getResults().
+  *
+  */
+  class OPENMS_DLLAPI Suitability:
```
hroest, line 60:
```suggestion
  class OPENMS_DLLAPI DBSuitability:
```
maybe call the class and file something more descriptive like DBSuitability?

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+    *
+    * @param pep_ids      vector containing pepIDs coming from a deNovo+database 
+    *                     identification search without FDR (currently only Comet-support)
+    * @throws             MissingInformation if no target/decoy annotation is found
+    * @throws             MissingInformation if no xcorr is found
+    * @throws             Precondition if a q-value is found in the input
+    */
+    void compute(std::vector<PeptideIdentification> pep_ids);
+
+    /// return results
```
hroest, line 113: can you document the output a bit more clear, how am I supposed to use this? What do I do with these results, what does the `suitability` parameter tell me? How do I now know whether the DB is suitable for the file or not? do I look at all entries in the vector?

src/openms/include/OpenMS/QC/Suitability.h
```diff
@@ -0,0 +1,173 @@
...
+  public:
+    /// struct to store results
+    struct SuitabilityData
+    {
+      Size num_top_novo = 0;
+      Size num_top_db = 0;
+      Size num_re_ranked = 0;
+      Size num_interest = 0;
+      double cut_off = DBL_MAX;
+      double suitability = 0;
```
hroest, line 72: please comment all members

```suggestion
      double suitability = 0; /// Value between 0 and 1 where 1 signifies ... 
```

src/openms/include/OpenMS/QC/Suitability.h
cbielow: good point

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -0,0 +1,298 @@
...
+      }
+      else
+      {
+        ++data.num_top_novo;
+      }
+    }
+
+    if (data.num_top_db == 0 && data.num_top_novo == 0)
+    {
+      OPENMS_LOG_WARN << "Identifications could not be assigned to either the database or the deNovo protein. Probably your FDR is to strict." << endl;
```
cbielow, line 186: `too strict`

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -47,7 +47,7 @@ namespace OpenMS
   class PeptideHit;
 
   /**
-  * @brief This class serves as the library representation @ref TOPP_DatabaseSuitability
+  * @brief This class serves as the library representation @ref TOPP_DatabaseDBSuitability
```
cbielow, line 50: revert to old name

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -0,0 +1,202 @@
...
+  *
+  * This class holds the functionality of calculating the database suitability.
+  * This can only be done if a combined deNovo+database identification search was performed.
+  * Currently only Comet search is supported.
+  *
+  * Allows for multiple usage, because results are stored internally and can be returned
+  * using getResults().
+  *
+  */
+  class OPENMS_DLLAPI DBSuitability:
```
timosachsenberg, line 60: please extend documentation a bit. e.g., without only refering back to the tool.
Also it's not clear how the multiple usage works.

timosachsenberg:
looks pretty complete to me.

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -47,15 +47,21 @@ namespace OpenMS
   class PeptideHit;
 
   /**
-  * @brief This class serves as the library representation @ref TOPP_DatabaseSuitability
+  * @brief This class holds the functionality of calculating the database suitability.
   *
-  * This class holds the functionality of calculating the database suitability.
-  * This can only be done if a combined deNovo+database identification search was performed.
+  * To calculate the suitability of a database for a specific mzML for identification search it
```
jpfeuffer, line 52:
```suggestion
  * To calculate the suitability of a database for a specific mzML for identification search, it
```

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -47,15 +47,21 @@ namespace OpenMS
...
+  * is vital to perform a combined deNovo+database identification search. Meaning that the database
+  * should be appended with an additional entry derived from concatenated deNovo sequences from said mzML.
   * Currently only Comet search is supported.
   *
-  * Allows for multiple usage, because results are stored internally and can be returned
-  * using getResults().
+  * This class will calculate q-values by itself and will throw an error if any q-value calculation is done beforehand.
```
jpfeuffer, line 57:
```suggestion
  * This class will calculate q-values by itself and will throw an error if any q-value calculation was done beforehand.
```

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -47,15 +47,21 @@ namespace OpenMS
...
   *
+  * The algorithm parameters can be set using setParams().
+  *
+  * Allows for multiple usage of the compute function. The result of each call is stored internally in a vector.
+  * Therefore old results will not be overridden by a new call. This vector than can be returned using getResults().
```
jpfeuffer, line 62:
```suggestion
  * Therefore old results will not be overridden by a new call. This vector then can be returned using getResults().
```

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -47,15 +47,21 @@ namespace OpenMS
...
+  *
+  * This class serves as the library representation @ref TOPP_DatabaseDBSuitability
```
jpfeuffer, line 64:
```suggestion
  * This class serves as the library representation of @ref TOPP_DatabaseDBSuitability
```

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -113,11 +119,11 @@ namespace OpenMS
     * 
     * Parameters can be set using the functionality of DefaultParamHandler.
     * Parameters are:
-    *                no_re_rank - re-ranking can be turned off with this
-    *                novo_fract - percent of deNovo peptides to capture in re-ranking
-    *                FDR        - q-value that should be filtered for
-    *                             Preliminary tests have shown that database suitability
-    *                             is rather stable across common FDR thresholds from 0 - 5 %
+    *                no_re_rank    - re-ranking can be turned off with this
+    *                cut_off_fract -  percentil that determines which cut-off will be returned
```
jpfeuffer, line 123:
```suggestion
    *                cut_off_fract -  percentile that determines which cut-off will be returned
```

jpfeuffer, line 123:
```suggestion
    *                cut_off_fract - percentile that determines which cut-off will be returned
```

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -133,41 +168,41 @@ namespace OpenMS
...
     * It is asssumed that this difference accounts for novo_fract of the re-ranking cases.
     *
-    * @param pep_ids      vector containing the pepIDs
-    * @param novo_fract   fraction of how many cases, where a de novo peptide scores just higher than the database peptide, will be re-rank
-    * @returns            xcorr cut-off
-    * @throws             IllegalArgument if novo_fract isn't in range [0,1]
-    * @throws             IllegalArgument if novo_fract is to low for a decoy cut-off to be calculated
-    * @throws             MissingInformation if no more than 20 % of the pepIDs have two decoys in there top ten
+    * @param pep_ids        vector containing the pepIDs
+    * @param cut_off_fract  percentil that determines which cut-off will be returned
```
jpfeuffer, line 176:
```suggestion
    * @param cut_off_fract  percentile that determines which cut-off will be returned
```

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -133,41 +168,41 @@ namespace OpenMS
...
+    * @returns              xcorr cut-off
+    * @throws               IllegalArgument if novo_fract isn't in range [0,1]
+    * @throws               IllegalArgument if novo_fract is too low for a decoy cut-off to be calculated
+    * @throws               MissingInformation if no more than 20 % of the pepIDs have two decoys in their top ten
```
jpfeuffer, line 180:
```suggestion
    * @throws               MissingInformation if no more than 20% of the peptide IDs have two decoys in their top ten peptide Hits
```

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -133,41 +168,41 @@ namespace OpenMS
...
+    * If at least one database protein is found, 'false' is returned.
     *
     * @param hit      PepHit in question
     * @returns        true/false
     */
     bool isNovoHit_(const PeptideHit& hit);
 
     /**
-    * @brief Tests if a PeptideHit has a higher q-value the given FDR
+    * @brief Tests if a PeptideHit has a lower q-value the given FDR threshold, i.e. passes FDR
```
jpfeuffer, line 197:
```suggestion
    * @brief Tests if a PeptideHit has a lower q-value than the given FDR threshold, i.e. passes FDR
```

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -38,30 +38,30 @@
...
-    : DefaultParamHandler("Suitability"), results_{}
+  DBSuitability::DBSuitability()
+    : DefaultParamHandler("DBSuitability"), results_{}
   {
     defaults_.setValue("no_re_rank", "false", "Enable/Disable re-ranking");
     defaults_.setValidStrings("no_re_rank", { "true", "false" });
-    defaults_.setValue("novo_fract", 1., "Fraction of how many cases, where a de novo peptide scores just higher than the database peptide, will be re-rank");
-    defaults_.setMinFloat("novo_fract", 0.);
-    defaults_.setMaxFloat("novo_fract", 1.);
+    defaults_.setValue("cut_off_fract", 1., "Percentil to determine which decoy cut-off to use.");
```
jpfeuffer, line 52:
```suggestion
    defaults_.setValue("cut_off_fract", 1., "Percentile to determine which decoy cut-off to use.");
```

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -184,15 +181,22 @@ namespace OpenMS
       }
     }
 
+    if (data.num_top_db == 0 && data.num_top_novo == 0)
+    {
+      OPENMS_LOG_WARN << "Identifications could not be assigned to either the database or the deNovo protein. Probably your FDR is too strict." << endl;
```
jpfeuffer, line 186:
```suggestion
      OPENMS_LOG_WARN << "Identifications could not be assigned to either the database or the deNovo protein. Probably your FDR threshold is too strict." << endl;
```

src/topp/DatabaseSuitability.cpp
```diff
@@ -142,9 +142,9 @@ class DatabaseSuitability :
     setValidFormats_("in_novo", { "idXML" });
     registerOutputFile_("out", "<file>", "", "Optional tsv output containing database suitability information as well as spectral quality.", false);
     setValidFormats_("out", { "tsv" });
-    registerDoubleOption_("novor_fract", "<double>", 1, "Set the fraction of how many cases, where a de novo peptide scores just higher than the database peptide, you wish to re-rank.", false, true);
-    setMinFloat_("novor_fract", 0);
-    setMaxFloat_("novor_fract", 1);
+    registerDoubleOption_("cut_off_fract", "<double>", 1, "Percentil to determine which decoy cut-off to use. '1' use the highest cut-off to '0' use the lowest one.", false, true);
```
jpfeuffer, line 145:
```suggestion
    registerDoubleOption_("cut_off_fract", "<double>", 1, "Percentile to determine which decoy cut-off to use. '1' use the highest cut-off to '0' use the lowest one.", false, true);
```

src/topp/QualityControl.cpp
```diff
@@ -141,7 +141,7 @@ class TOPPQualityControl : public TOPPBase
     registerInputFileList_("in_trafo", "<file>", {}, "trafoXMLs from MapAligners", false);
     setValidFormats_("in_trafo", {"trafoXML"});
     registerTOPPSubsection_("MS2_id_rate", "MS2 ID Rate settings");
-    registerFlag_("MS2_id_rate:force_no_pep_index", "Forces the metric to run if target/decoy annotation is missing (accepts all pep_ids as target hits).", false);
+    registerFlag_("MS2_id_rate:assume_all_target", "Forces the metric to run if target/decoy annotation is missing (accepts all pep_ids as target hits).", false);
```
jpfeuffer, line 144:
```suggestion
    registerFlag_("MS2_id_rate:assume_all_target", "Forces the metric to run even if target/decoy annotation is missing (accepts all pep_ids as target hits).", false);
```

jpfeuffer:
Only phrasing. Otherwise, looks good from my side.

jpfeuffer:
Merge @cbielow ?

src/topp/DatabaseSuitability.cpp
```diff
@@ -95,7 +94,7 @@ To generate the de novo "database":
       - @ref TOPP_IDFileConverter generates the de novo fasta file.
 
 For re-ranking all cases where a peptide hit only found in the de novo "database" scores above a peptide hit found in the actual database are checked. In all these cases the cross-correlation scores of those peptide hits are compared. If they are similar enough, the database hit will be re-ranked to be on top of the de novo hit. You can control how much of cases with similar scores will be re-ranked by using the @p novor_fract option.@n
```
cbielow, line 96: should be `@p cut_off_fract`... but see below

src/topp/DatabaseSuitability.cpp
```diff
@@ -95,7 +94,7 @@ To generate the de novo "database":
...
-For this to work it is important that FDR filtering is done in this tool and not beforehand by @ref TOPP_FalseDiscoveryRate. You can provide the wanted FDR using the corresponding flag.
+For this to work it is important that no FDR was performed. This tool does this itself and will crash if a q-value is found. You can still control the FDR that you want to establish using the corresponding flag.
```
cbielow, line 97: Mention that PepIndexer is required (its only mentioned in the parameter help so far).

src/topp/DatabaseSuitability.cpp
```diff
@@ -135,17 +134,17 @@ class DatabaseSuitability :
...
     setValidFormats_("in_spec", { "mzML" });
-    registerInputFile_("in_novo", "<file>", "", "Input idXML file containing de novo peptides");
+    registerInputFile_("in_novo", "<file>", "", "Input idXML file containing de novo peptides (unfiltered)");
     setValidFormats_("in_novo", { "idXML" });
     registerOutputFile_("out", "<file>", "", "Optional tsv output containing database suitability information as well as spectral quality.", false);
     setValidFormats_("out", { "tsv" });
-    registerDoubleOption_("novor_fract", "<double>", 1, "Set the fraction of how many cases, where a de novo peptide scores just higher than the database peptide, you wish to re-rank.", false, true);
-    setMinFloat_("novor_fract", 0);
-    setMaxFloat_("novor_fract", 1);
+    registerDoubleOption_("cut_off_fract", "<double>", 1, "Percentile to determine which decoy cut-off to use. '1' use the highest cut-off to '0' use the lowest one.", false, true);
```
cbielow, line 145: this is the `reranking_cutoff_percentile` (or along those lines)?

`Swap a top-scoring deNovo hit with a lower scoring DB hit if their xcorr score difference is less than threshold 't'.
't' is determined as the percentile (this parameter) in the list of all sorted score differences between two decoy hits of a PSM.` ??
.... phew...

cbielow, line 145: is `1` a good default?

src/topp/DatabaseSuitability.cpp
```diff
@@ -196,292 +189,69 @@ class DatabaseSuitability :
...
+    q.compute(novo_peps, exp, true);
 
-    // spectra quality
+    Ms2IdentificationRate::IdentificationRateData spectral_quality = q.getResults()[0];
 
-    Size count_novo_seq = 0;
-    set<AASequence> unique_novo;
+    DBSuitability s;
+    Param p;
+    p.setValue("no_re_rank", no_re_rank ? "true" : "false");
```
cbielow, line 211: having duplicated parameters (even though its just 3) is bad for maintenance.
You can define the parameters in the class and simply make new TOPPSubsection 'algorithm' as done in other tools which just takes the Param from the class.
To pass the parameters, simply extract them back (done in many other tools) -- see MSSimulator etc..

src/topp/DatabaseSuitability.cpp
```diff
@@ -162,26 +161,20 @@ class DatabaseSuitability :
     String in_spec = getStringOption_("in_spec");
     String in_novo = getStringOption_("in_novo");
     String out = getStringOption_("out");
-    double novo_fract = getDoubleOption_("novor_fract");
+    double cut_off_fract = getDoubleOption_("cut_off_fract");
     double FDR = getDoubleOption_("FDR");
     bool no_re_rank = getFlag_("force_no_re_rank");
```
cbielow, line 166: stay with `no_rerank` in the class (and also the TOPP tool by using a subsection).
Replace all occurrences of `force_no_re_rank` in the PR.

src/topp/DatabaseSuitability.cpp
jpfeuffer: Yes, thanks, this is really important.

src/topp/DatabaseSuitability.cpp
jpfeuffer: `Swap a top-scoring deNovo hit with a lower scoring DB hit if their xcorr score difference is in the given percentile of all score differences between the first two decoy hits of a PSM.`

jpfeuffer: Percentile implies sorted.
And it should be clear that it is about this param. Especially if we rename to ..._percentile.

Waschi97: should be, right?
'1' results in the lowest difference being used as a cut-off. This is the most strict and it's guaranteed to work.

jpfeuffer: Yes it is guaranteed to work, but does it produce nice results on average?
What do you use for benchmarking?

Waschi97: For benchmarking 0.99 was used because that's the setting from the original paper and we wanted comparable results. What would produce the nicest results on average wasn't tested. That's probably not a bad idea though...

jpfeuffer:
@Waschi97 So, no changes needed to make sure that 0.99 always works?

Waschi97:
Changes were needed! I just wanted to also add them to the test files first. :)

jpfeuffer:
ping @cbielow more comments?

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -0,0 +1,208 @@
...
+  * Currently only Comet search is supported.
+  *
+  * This class will calculate q-values by itself and will throw an error if any q-value calculation was done beforehand.
+  *
+  * The algorithm parameters can be set using setParams().
+  *
+  * Allows for multiple usage of the compute function. The result of each call is stored internally in a vector.
+  * Therefore old results will not be overridden by a new call. This vector then can be returned using getResults().
+  *
+  * This class serves as the library representation of @ref TOPP_DatabaseDBSuitability
```
cbielow, line 64: should be `TOPP_DatabaseSuitability`

src/topp/DatabaseSuitability.cpp
```diff
@@ -94,11 +93,11 @@ To generate the de novo "database":
...
-For this to work it is important that FDR filtering is done in this tool and not beforehand by @ref TOPP_FalseDiscoveryRate. You can provide the wanted FDR using the corresponding flag.
+For re-ranking all cases where a peptide hit only found in the de novo "database" scores above a peptide hit found in the actual database are checked. In all these cases the cross-correlation scores of those peptide hits are compared. If they are similar enough, the database hit will be re-ranked to be on top of the de novo hit. You can control how much of cases with similar scores will be re-ranked by using the @p reranking_cutoff_percentile.@n
+For this to work it is important @ref TOPP_PeptideIndexer ran before. However it is also crutial that no FDR was performed. This tool does this itself and will crash if a q-value is found. You can still control the FDR that you want to establish using the corresponding flag.
```
cbielow, line 97: `crucial`

cbielow: will it `crash` or rather `complain`?

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -0,0 +1,208 @@
...
+    *           reranking_cutoff_percentile - percentile that determines which cut-off will be returned
+    *           FDR                         - q-value that should be filtered for
+    *                                         Preliminary tests have shown that database suitability
+    *                                         is rather stable across common FDR thresholds from 0 - 5 %
+    *
+    * Since q-values need to be calculated the identifications are taken by copy.
+    *
+    * Result is appended to the result member. This allows for multiple usage.
+    *
+    * @param pep_ids      vector containing pepIDs with target/decoy annotation coming from a deNovo+database
```
cbielow, line 132: add a comment why this is not a const& (`vector is modified internally, and is thus copied`) or similar

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+    // sort the diffs decreasing and get the percentile one
+    UInt index = round(reranking_cutoff_percentile * diffs.size());
+    
+    if (index >= diffs.size())
+    {
+      return *min_element(diffs.begin(), diffs.end());
+    }
+
+    nth_element(diffs.begin(), diffs.begin() + index, diffs.end(), greater<double>());
```
cbielow, line 275: so this is still reversed?
Swap in the next PR?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -0,0 +1,298 @@
...
+using namespace std;
+
+namespace OpenMS
+{
+  DBSuitability::DBSuitability()
+    : DefaultParamHandler("DBSuitability"), results_{}
+  {
+    defaults_.setValue("no_rerank", "false", "Use this flag if you want to disable re-ranking. Cases, where a de novo peptide scores just higher than the database peptide, are overlooked and counted as a de novo hit. This might underestimate the database quality.");
+    defaults_.setValidStrings("no_rerank", { "true", "false" });
+    defaults_.setValue("reranking_cutoff_percentile", 0.01, "Swap a top-scoring deNovo hit with a lower scoring DB hit if their xcorr score difference is in the given percentile of all score differences between the first two decoy hits of a PSM.");
```
cbielow, line 52: mention what large/small values mean

cbielow (commented):
just some minor things.
good job and good to go otherwise!

src/openms/source/QC/DBSuitability.cpp
Waschi97: Yes, it is sorted descending. But since I take the actual percentile now and not the (1 - percentile) one like before, this is essentially swapped.

jpfeuffer:
tests fail now

Waschi97:
interesting, they don't seem to fail for me...

jpfeuffer:
http://cdash.openms.de/testDetails.php?test=18161249&build=42128

src/openms/source/QC/DBSuitability.cpp
Waschi97: My bad, the `greater<double>()` isn't needed anymore.

src/tests/topp/DatabaseSuitability_out_1.tsv
```diff
@@ -1,7 +1,7 @@
 key	value
-#top_db_hits	158
```
jpfeuffer, line 2: This change was expected?

Waschi97: Yeah, it was. Since the default for the `rerank_cutoff_percentile` changed, I had to change either the test or the test files.
I decided to change the test files, so that the default is tested.

jpfeuffer: Ah okay

PR #4871, opened 2020-08-10
Peptide indexer more user friendly

cbielow:
* `auto` mode for 3 settings in PeptideIndexer (all information is taken from input idXML)
  * enzyme specificity
  * enzyme name
  * FASTA filename

* allow removing weird peptides produced by XTandems's special cleavage rules ('X' is a stop codon...)

This PR changes some of the parameters of PI, i.e. old INI files need updating.


cbielow:
appleclang should be a false positive.

jpfeuffer:
Hmm yes, but we had this twice now. You recently changed this test and moved its data. Should be investigated in a separate PR.

cbielow:
> 
> 
> Hmm yes, but we had this twice now. You recently changed this test and moved its data. Should be investigated in a separate PR.

a first try here #4838

jpfeuffer:
I know. I think that caused it.

Issue #4654, opened 2020-04-17
Comet does not read our BSA1.mzML


cbielow:
even though it contains an Index...
CometAdapter reports:
```
FATAL ERROR: External invocation of 'C:/Program Files/OpenMS-2.4.0/share/OpenMS/THIRDPARTY/Comet/comet.exe' failed. Standard output and error were:
Standard output:  Comet version "2019.01 rev. 4"

 Search start:  04/17/2020, 07:33:40 PM
 - Input file: ..\openmsqt5\share\OpenMS\examples\BSA\BSA1.mzML
   - Load spectra:
Standard error: ..\openmsqt5\share\OpenMS\examples\BSA\BSA1.mzML(1) : parseOffset() 2: Syntax error parsing XML.

Exit code: -7
```

using `FileInfo.exe -v -in share\OpenMS\examples\BSA\BSA1_new.mzML` indeed shows some problems with our own mzML!!!
```
Error: CV term must have a unit: MS:1000045 - collision energy
Error: CV term must have a unit: MS:1000504 - base peak m/z
Error: CV term must have a unit: MS:1000505 - base peak intensity
Error: CV term must have a unit: MS:1000528 - lowest observed m/z
Error: CV term must have a unit: MS:1000527 - highest observed m/z
```

this might be the reason... but could be something entirely different...

cbielow:
also happens with Comet v "2019.01 rev. 5"

jpfeuffer:
Why is one file called `share\OpenMS\examples\BSA\BSA1.mzML`
and the other `share\OpenMS\examples\BSA\BSA1_new.mzML`?

PR #4874, opened 2020-08-10
[FIX] Protein terminal mods in AASequence

### LINKED ISSUE #4830: Variable Modification "Met-loss (Protein N-term M)" yields error in CometAdapter
This error is thrown:
```
Error: Unexpected internal error (the value 'Met-loss (Protein N-term M)' was used but is not valid; Retrieving the modification failed. It is not available for the residue '' and term specificity 'N-term'. )
Error occurred in line 265 of file C:\Development\OpenMS\src\openms\source\CHEMISTRY\ModificationsDB.cpp (in function: const class OpenMS::ResidueModification *__cdecl OpenMS::ModificationsDB::getModification(const class OpenMS::String &,const class OpenMS::String &,enum OpenMS::ResidueModification::TermSpecificity) const) !
```

This modification is supported by Comet as the 'clip_nterm_methionine' parameter in its param file and can also be passed to CometAdapter as a flag.
Maybe CometAdapter should check for this modification in the 'variable_modifications' parameter and set 'clip_nterm_methionine' accordingly instead of writting the modification.

jpfeuffer:
# Description
Fixes #4830 
Fixes bug in AASequence when trying to set terminal modifications.
Now correctly parses residue-specific terminal mods.
And if not found for general terminal mods, it automatically looks for protein terminal mods.
The best way in the future would be to actually check if this peptide is protein terminal and only look up the corresponding modifications but this logic needs to be outside of AASequence and implemented everywhere where we have access to this information.

# Checklist:


jpfeuffer:
What the hell?

jpfeuffer:
I did not touch anything related to XTandem. And the converted mzMLs are fine as seen on non-Windows.

jpfeuffer:
@cbielow Can you reproduce this?

cbielow:
Yes, can reproduce Comet's error.
When doing `FileConverter.exe -in C:/dev/openms_test/share/OpenMS/examples/FRACTIONS/BSA1_F1.mzML -out C:/dev/openms_test/share/OpenMS/examples/FRACTIONS/BSA1_F1.mzML`, then Comet does not crash anymore and finished successfully.. although the results are slightly different (suggesting that on Linux it also does something weird, but does not crash).

XTandem I cannot reproduce (X! TANDEM Vengeance (2015.12.15.2)).


btw, since it came up in the tests (printing `(exit code: ù).`): could you change
`(exit code: " + qp_->exitCode() + "). ` to `(exit code: " + int(qp_->exitCode()) + "). ` in
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/SYSTEM/ExternalProcess.cpp#L117
while you're at it?
Thanks!

jpfeuffer:
Weird. That Fileconversion command was exactly what I did for the files. This is why they changed.

jpfeuffer:
I just converted again. Results in the same mzML file (except for another dataprocessing entry) and updated offsets.
Same results in all tests.
I think Windows Comet just can't read non-Windows generated things.

jpfeuffer:
rebuild jenkins

cbielow:
solves https://github.com/OpenMS/OpenMS/issues/4654 (Comet crashes due to conversion of line endings from LF to CRLF by Git, which invalidates the mzML indeX)

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1429,23 +1429,53 @@ namespace OpenMS
...
-    if (modification == "")
+    if (modification.empty())
     {
       n_term_mod_ = nullptr;
       return;
     }
 
-    n_term_mod_ = ModificationsDB::getInstance()->getModification(modification, "", ResidueModification::N_TERM);
+    String residue = "";
+    char last_char_no_parentheses = modification[modification.length()-2];
```
hroest, line 1439: what happens if I enter a string of size 1? 

```
setNTerminalModification("T")
```

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1429,23 +1429,53 @@ namespace OpenMS
...
-    if (modification == "")
+    if (modification.empty())
     {
       c_term_mod_ = nullptr;
       return;
     }
-    c_term_mod_ = ModificationsDB::getInstance()->getModification(modification, "", ResidueModification::C_TERM);
+
+    String residue = "";
+    char last_char_no_parentheses = modification[modification.length()-2];
```
hroest, line 1465: same here

src/openms/source/CHEMISTRY/AASequence.cpp
hroest: @jpfeuffer can you please add that to the AASequence_test.cpp

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1445,9 +1444,19 @@ namespace OpenMS
...
+    // For strings in our most common UniMod format
     if (!modification.hasSubstring("Protein N-term"))
     {
-      n_term_mod_ = ModificationsDB::getInstance()->getModification(modification, residue, ResidueModification::N_TERM);
+      try
+      {
+        n_term_mod_ = ModificationsDB::getInstance()
+            ->getModification(modification, residue, ResidueModification::N_TERM);
+      }
+      catch (...)
```
timosachsenberg, line 1455: please comment try catch logic

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1476,7 +1485,14 @@ namespace OpenMS
 
     if (!modification.hasSubstring("Protein C-term"))
     {
-      c_term_mod_ = ModificationsDB::getInstance()->getModification(modification, residue, ResidueModification::C_TERM);
+      try
```
timosachsenberg, line 1488: same ehere

timosachsenberg:
do the pxds need an update?

jpfeuffer:
Unless pxds need special annotation when you catch exceptions, no. No interface changed.. yet.
It should heavily though, to make this pepXML parsing and general mod. handling from a nightmare to usable.

Failures unrelated.. I really need to check why their frequency increased after the update.

timosachsenberg:
I thought there were some references introduced that might be supported

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1429,23 +1428,84 @@ namespace OpenMS
...
+    if (!modification.hasSubstring("Protein C-term"))
+    {
+      // since this method is called setNTerminalModification without further specification
+      // we have to look for both general terminus and Protein terminus.
+      // For backwards compatibility we look for the general terminus first
+      // We have to use try-catch since getModification unfortunately throws Exceptions.
+      try
+      {
+        c_term_mod_ = ModificationsDB::getInstance()->getModification(modification, residue, ResidueModification::C_TERM);
+      }
```
cbielow, line 1499: return from function here?

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1429,23 +1428,84 @@ namespace OpenMS
...
+      catch (...)
+      {
+        c_term_mod_ = ModificationsDB::getInstance()->getModification(modification, residue, ResidueModification::PROTEIN_C_TERM);
```
cbielow, line 1502: catch, but don't do anything.

src/openms/source/CHEMISTRY/AASequence.cpp
```diff
@@ -1429,23 +1428,84 @@ namespace OpenMS
...
+      }
+    }
+    else
+    {
+      c_term_mod_ = ModificationsDB::getInstance()->getModification(modification, residue, ResidueModification::PROTEIN_C_TERM);
```
cbielow, line 1507: and then always do this.

Would make the code easier and less repetitive?

src/openms/source/FORMAT/PepXMLFile.cpp
```diff
@@ -1319,31 +1347,71 @@ namespace OpenMS
...
+      }
+
+      // BIG NOTE: According to the pepXML schema specification, protein terminus is either "c" or "n" if set.
+      // BUT: Many tools will put "Y" or "N" there in conjunction with the terminus attribute.
+      // Unfortunately there is an overlap for the letter "n". We will try to handle both based on the case:
+      String entry_lower = protein_terminus_entry;
+      entry_lower = entry_lower.toLower();
+      if (entry_lower == "y")
+      {
+        aa_mod.protein_terminus = true;
```
cbielow, line 1367: sorry, but all this code does not belong here!

collect all the information you need and refactor it to `aa_mod`'s implemention.
Ideally, make it a constructor and forbid changing properties individually (to maintain invariants).

This is a typical case of 'just add the code inline' which makes it
 - not testable
 - creates huge unmaintainable classes
 - is no fun to refactor at all

src/openms/source/FORMAT/PepXMLFile.cpp
```diff
@@ -1365,6 +1433,7 @@ namespace OpenMS
       if (desc.empty())
       {
         vector<String> mods;
+        // if terminus was not specified
```
cbielow, line 1436: creating a string from `aa_mod` should be a member function of aa_mod!

src/openms/source/FORMAT/PepXMLFile.cpp
jpfeuffer: aa_mod should not be here at all.
It is:
a) abused also for terminal mods
b) should be our own Residuemodification object, such that we do not lookup ever changing "mod-descriptions" dozens of times

jpfeuffer:
I have to give up now. Not sure when I get to it again.

src/openms/source/FORMAT/PepXMLFile.cpp
```diff
@@ -1863,7 +1863,8 @@ namespace OpenMS
 
       for (const auto& mod : fixed_modifications_)
       {
-        if (mod.getTerminus() == "n")
+        if (mod.getRegisteredMod()->getTermSpecificityName() == "N-term" ||
```
cbielow, line 1866: `.hasSuffix("N-term")`  to make it shorter?

PR #4876, opened 2020-08-10
[FEATURE] Fileinfo: more info

cbielow:
* extend mslevel + fragmentation method information 

also fixes a test on Apple Clang (amends #4838)

### DIFF
--- a/src/tests/topp/FileInfo_5_output.txt
+++ b/src/tests/topp/FileInfo_5_output.txt
@@ -13,7 +13,7 @@ Total number of peaks: 3149
 Number of spectra: 10
 
 Ranges:
-  retention time: 20.87 .. 39.97
+  retention time: 20.87 .. 39.97 sec (0.3 min)
   mass-to-charge: 500.07 .. 1497.28
   intensity:      16.30 .. 38920.12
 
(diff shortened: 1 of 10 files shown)


src/tests/topp/FileInfo_5_output.txt
```diff
@@ -13,15 +13,15 @@ Total number of peaks: 3149
 Number of spectra: 10
 
 Ranges:
-  retention time: 20.87 .. 39.97
+  retention time: 20.87 .. 39.97 sec (0.3 min)
```
timosachsenberg, line 16: start also in min?

timosachsenberg:
feel free to merge

src/tests/topp/FileInfo_5_output.txt
timosachsenberg: maybe as 
```
(0.2 min .. 0.3 min)
```

timosachsenberg: ah ok it's already the duration of the whole run

cbielow: should I use `(span 30 min)` to make it more clear?

PR #4880, opened 2020-08-12
Toppview refactor

### LINKED ISSUE #4812: TOPPView: Layers display not updated
In TOPPView, when a file is opened as a "new layer", it should appear in the "Layers" window. Due to a recent regression, this does not happen (immediately) any more. The layer is drawn in the main window, but the "Layers" window is not updated - that is, until I open another file, or minimise and un-minimise the TOPPView window. (There may be other triggers as well.)

Without knowing these workarounds, it is very confusing that the "Layer" entry is missing. This also prevents users from interacting with the layer (disabling it, setting it to active, removing it...).

This occurs with the current development version on Ubuntu 18.04.4.

cbielow:
# Description

fixes #4812

### DIFF
--- a/src/openms/include/OpenMS/SYSTEM/FileWatcher.h
+++ b/src/openms/include/OpenMS/SYSTEM/FileWatcher.h
@@ -62,17 +62,17 @@ namespace OpenMS
 
       @ingroup System
   */
-  class FileWatcher :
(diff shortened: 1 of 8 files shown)


cbielow:
tests pass.
ready for merge

PR #4884, opened 2020-08-14
subsitute vector<double> for map<PeakType, double> in SignalToNoiseEstimator

cbielow:
# Description

augments #4775 by fixing python bindings


timosachsenberg:
I did not look into the code but I guess you checked if changing from map to vector is still thread-safe (or this is only called in single-threaded context)

cbielow:
rebuild jenkins

cbielow:
> 
> 
> I did not look into the code but I guess you checked if changing from map to vector is still thread-safe (or this is only called in single-threaded context)

`init()` is/was never thread-safe, but `getSignalToNoise()` is (and was).

PR #4885, opened 2020-08-14
Toppview refactor

cbielow:
# Description

further refactoring of TOPPView into separate classes. Deletion of a lot of code duplications.
Fixes some bugs with tab handling and makes the code more modular.
This is still far from `ok`, but a first step.

This PR adds:
 * support for a `flag set` via OpenMS/DATASTRUCTURES/FlagSet.h (allows to combine multiple enum values into a single state using 'operator+', 'operator-' etc).
 * switching to compile-time checked Signal/Slot Qt5 mechanism for the refactored classes (much safer)

This PR fixes:
#4812
#4926
#4928


src/openms/include/OpenMS/FILTERING/DATAREDUCTION/DataFilters.h
```diff
@@ -70,24 +70,30 @@ namespace OpenMS
       EXISTS          ///< Only for META_DATA filter type, tests if meta data exists
     };
 
-    ///Representation of a peak/feature filter combining FilterType, FilterOperation and a value
+    /// Representation of a peak/feature filter combining FilterType, FilterOperation and a value (either double or String)
     struct OPENMS_DLLAPI DataFilter
```
timosachsenberg, line 74: should this class be moved to the GUI ? It seems to be only used there

src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.h
```diff
@@ -171,5 +172,9 @@ namespace OpenMS
     std::vector<double> stn_estimates_;
   };
 
+  /// Picks @p n_scans from level 1 randomly and returns either average intensity at a certain @p percentile.
+  /// If no MS1 scans are present, 0.0 is returned
+  OPENMS_DLLAPI float estimateNoiseFromRandomMS1Scans(const MSExperiment& exp, const UInt n_scans = 10, const double percentile = 80);
```
timosachsenberg, line 177: pass leve to make it more genericl?

src/openms/include/OpenMS/KERNEL/MSExperiment.h
```diff
@@ -551,6 +551,8 @@ namespace OpenMS
     /// returns true if any MS spectra of the specified level contain at least one peak with intensity of 0.0
     bool hasZeroIntensities(size_t ms_level) const;
 
+    /// do any of the spectra have a peptideID?
+    bool hasPeptideIdentifications() const;
```
timosachsenberg, line 555: not part of this PR but imho IDs should not be stored in MSExperiment

src/openms/source/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.cpp
```diff
@@ -29,7 +29,47 @@
...
+
+using namespace std;
+
+namespace OpenMS
+{
+
+  float estimateNoiseFromRandomMS1Scans(const MSExperiment& exp, const UInt n_scans, const double percentile)
+  {
+    vector<Size> ms1_indices;
+    for (Size i = 0; i < exp.size(); ++i) if (exp[i].getMSLevel() == 1 && !exp[i].empty()) ms1_indices.push_back(i);
```
timosachsenberg, line 50: would break this into multiple lines

src/openms/source/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.cpp
```diff
@@ -29,7 +29,47 @@
...
+
+    if (ms1_indices.empty()) return 0.0f;
+
+    std::default_random_engine generator(time(nullptr));
```
timosachsenberg, line 54: might not be protable / will introduce non-determinism

src/openms_gui/include/OpenMS/VISUAL/LayerData.h
```diff
@@ -82,7 +82,7 @@ namespace OpenMS
 
   @ingroup SpectrumWidgets
   */
-  class LayerData
+  class OPENMS_GUI_DLLAPI LayerData
```
timosachsenberg, line 85: maybe a good place for polymorphism in the future? to have different layer types  for different data?

src/openms_gui/include/OpenMS/VISUAL/LayerListView.h
```diff
@@ -28,68 +28,55 @@
...
 
-  void TOPPASLogWindow::setMaxLength(int max_length)
-  {
-    max_length_ = max_length;
-  }
+    /// rebuild list of layers and remember current widget (for context menu etc)
+    void update(SpectrumWidget* active_widget);
+
+  signals:
+    void layerDataChanged();
```
timosachsenberg, line 65: maybe document when it is emitted?

src/openms_gui/include/OpenMS/VISUAL/SpectraSelectionTabs.h
```diff
@@ -28,79 +28,82 @@
...
+    /// Default constructor
+    SpectraSelectionTabs(QWidget* parent, TOPPViewBase* tv);
+
+    /// update items in the two tabs according to the currently selected layer
+    void update();
+
+    /// invoked when user changes the active tab to @p tab_index
+    void currentTabChanged(int tab_index);
+    
+    void showSpectrumAs1D(int index);
```
timosachsenberg, line 77: if you know by chance what the next three methods are doing you could add a short comment

src/openms_gui/source/VISUAL/SpectraViewWidget.cpp
```diff
@@ -199,115 +206,80 @@ namespace OpenMS
...
         else
-        {
-          // open several chromatograms at once
-          for (Int i = 0; i != res.size(); ++i)
-          {
-            chrom_indices.push_back(res[i].toInt());
-          }
-          emit showSpectrumAs1D(chrom_indices);
+        { // open several chromatograms at once
+          emit showSpectrumAs1D(listToVec(res));
```
timosachsenberg, line 233: chromatogram handling is still suboptimal and something we might need to look at in the future

src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
```diff
@@ -940,8 +958,14 @@ namespace OpenMS
     if (peak.isValid())
     {
       QPoint begin;
-
-      const ExperimentType::PeakType& sel = getLayer_(layer_index).getCurrentSpectrum()[peak.peak];
+      const auto& spec = getLayer_(layer_index).getCurrentSpectrum();
+      if (peak.peak >= spec.size())
```
timosachsenberg, line 962: I guess the reference can probably change by other signals. Can we somehow block those?

timosachsenberg:
Very nice cleanup. Thanks!

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/DataFilters.h
cbielow: maybe, yes.
But I can imagine having use cases in the main lib as well.
Would keep it there for now.

src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.h
cbielow: possible, but this is just a refactor.
For MS2 its also less useful since MS2 is much less affected by noise.
But could be done.

src/openms/include/OpenMS/KERNEL/MSExperiment.h
cbielow: I agree!
But its used in many places...

src/openms_gui/include/OpenMS/VISUAL/LayerData.h
cbielow: Yes! People have been adding data members in the past without thinking about this... so currently, its a mess.

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/DataFilters.h
timosachsenberg: We have quite some filtering algorithms in the main lib. Don't know if these classes can be married. Otherwise, having different concepts might confuse users

src/openms/include/OpenMS/KERNEL/MSExperiment.h
timosachsenberg: It is ugly enough so I opened an issue

src/openms/source/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.cpp
cbielow: done

src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.h
cbielow: done

src/openms_gui/include/OpenMS/VISUAL/LayerData.h
timosachsenberg: Ok I'll add an issue for that

src/openms/source/FILTERING/NOISEESTIMATION/SignalToNoiseEstimator.cpp
cbielow: should be very portable. Just not deterministic

src/openms_gui/include/OpenMS/VISUAL/LayerListView.h
cbielow: done

src/openms_gui/include/OpenMS/VISUAL/SpectraSelectionTabs.h
cbielow: done

src/openms_gui/source/VISUAL/SpectraViewWidget.cpp
timosachsenberg: that would be awesome!

src/openms_gui/source/VISUAL/Spectrum1DCanvas.cpp
cbielow: since this is executed in the event loop until finished (the EL is not async), this is not possible IMHO

timosachsenberg (approved):

PR #4889, opened 2020-08-16
[FEATURE] combine SE settings during ConsensusID (instead of taking 1st)

jpfeuffer:
# Description

Important to have the same settings outcome, when executing ConsID on files in a different order. Before it was just taking first, now it takes the loosest settings.

Also please:

# Checklist:

### DIFF
--- a/src/topp/ConsensusID.cpp
+++ b/src/topp/ConsensusID.cpp
@@ -353,9 +353,10 @@ protected:
     int min_chg = 10000;
     int max_chg = -10000;
     Size mc = 0;
-    String enz;
+    // we sort them to pick the same entries, no matter the order of the inputs
+    set<String, std::greater<String>> enzymes;
+    set<String, std::greater<String>> dbs;
 
-    //TODO check if settings are same/similar
     bool allsamese = true;
     // use the first settings as basis (i.e. copy over db and enzyme and tolerance)
     // we assume that they are the same or similar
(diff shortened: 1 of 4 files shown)


src/topp/ConsensusID.cpp
```diff
@@ -373,6 +380,31 @@ class TOPPConsensusID :
       new_sp.setMetaValue(SE+":precursor_mass_tolerance_unit",sp.precursor_mass_tolerance_ppm  ? "ppm" : "Da");
       new_sp.setMetaValue(SE+":digestion_enzyme",sp.digestion_enzyme.getName());
       new_sp.setMetaValue(SE+":enzyme_term_specificity",EnzymaticDigestion::NamesOfSpecificity[sp.enzyme_term_specificity]);
+      
+      const auto& chg_pair = sp.getChargeRange();
+      if (chg_pair.first != 0 && chg_pair.first < min_chg) min_chg = chg_pair.first;
+      if (chg_pair.second != 0 && chg_pair.second > max_chg) max_chg = chg_pair.second;
+      if (sp.missed_cleavages > mc ) mc = sp.missed_cleavages;
+      if (sp.fragment_mass_tolerance > frag_tol) frag_tol = sp.fragment_mass_tolerance;
```
timosachsenberg, line 388: do you check if unit is the same?

src/topp/ConsensusID.cpp
```diff
@@ -373,6 +380,31 @@ class TOPPConsensusID :
...
+      if (sp.digestion_enzyme.getName().hasSubstring(enz))
+      {
+        enz = sp.digestion_enzyme.getName();
+      }
+      if (spec.empty())
+      {
+        spec = EnzymaticDigestion::NamesOfSpecificity[sp.enzyme_term_specificity];
+      }
+      //TODO does not handle no-cterm and no-nterm
+      else if (spec == "full" && (sp.enzyme_term_specificity == EnzymaticDigestion::SPEC_SEMI || sp.enzyme_term_specificity == EnzymaticDigestion::SPEC_NONE))
```
cbielow, line 400: this is quite hard to read and error prone.
what about a set<enum> just to keep track of what is present.
and at the end, check for presence from most loose setting to strictest once?

src/topp/ConsensusID.cpp
jpfeuffer: no, because then it cannot be combined. there is actually more to check, also for database and enzyme etc.

jpfeuffer: Suggestions welcome. It just has to be independent of the order. I currently assume e.g. same or similar enzymes, which I probably have to change.
Kind of a release blocker, since this blocks the PLFQ pipeline.

timosachsenberg:
code looks good. feel free to merge once it passes

PR #4892, opened 2020-08-19
Addon changes for pyopenms

24sharkS:
# Description

We currently cannot use the addon functions of ConvexHull2D.pyx and MapAlignmentAlgorithmIdentification.pyx. In the pxd files, these functions are declared without "#wrap-ignore" declarative and hence are wrapped by autowrap. The overloaded definitions in addons for these functions seem to be ignored by cython.
Thus, changing the name of these functions is one viable solution for exposing these in pyopenms and eventually the R bindings.

Also please:

# Checklist:

### DIFF
--- a/src/pyOpenMS/addons/ConvexHull2D.pyx
+++ b/src/pyOpenMS/addons/ConvexHull2D.pyx
@@ -3,15 +3,26 @@ from DPosition cimport DPosition2 as _DPosition2
 from DBoundingBox cimport DBoundingBox2 as _DBoundingBox2
 
 
-    def encloses(self, float x, float y):
-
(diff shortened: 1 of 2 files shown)


hroest:
looks good @24sharkS would it make sense to add these definitions to the `pxd` files with `wrap-ignore` just to document them better?

24sharkS:
> looks good @24sharkS would it make sense to add these definitions to the `pxd` files with `wrap-ignore` just to document them better?

Ideally as the pxd files contain C type declarations but since these are addon python functions to OpenMS, so I think parser will complain if we add python declarations to pxds with `# wrap-ignore` tag. Probably add these as comments instead in pxds ?

PR #4898, opened 2020-08-25
[FEATURE] Improve PSI MS ontology support in mzTab

jpfeuffer:
# Description

Adds support for some new and old PSI-MS terms

# Checklist:


src/utils/ProteomicsLFQ.cpp
```diff
@@ -1094,7 +1098,8 @@ class ProteomicsLFQ :
         f.setConvexHulls({});
       }
 
-      IDConflictResolverAlgorithm::resolve(tmp, false); // keep only best peptide per feature
+      IDConflictResolverAlgorithm::resolve(tmp,
+          getStringOption_("keep_feature_top_psm_only") == "true"); // keep only best peptide per feature per file
```
jpfeuffer, line 1102: should be false

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -3737,25 +3797,14 @@ Not sure how to handle these:
     const auto& refseq = curr_pep_ids[0].getHits()[0].getSequence();
     for (const auto& pep : curr_pep_ids)
     {
-      if (pep.getHits().size() > 1)
+      if (pep.getHits()[0].getSequence() != refseq)
```
jpfeuffer, line 3800: we could revert that. this relaxation was not needed in the end. but exception is also not really necessary since we always take the best hit only anyway. if we communicate that..

timosachsenberg:
Curious if these changes are sufficient. If so I think it is good to go

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -2686,11 +2703,54 @@ Not sure how to handle these:
...
+      MzTabParameter pep_score_type;
       const tuple<String, String, String>& name_version_score = se.first;
-      psm_score_type.fromCellString("[,," + get<0>(name_version_score) + " " + get<2>(name_version_score) + "," + get<1>(name_version_score) + "]");
+
+      //TODO we also have to consider the different q-value calculation methods of Percolator....
+      if (get<2>(name_version_score) == "peptide-level q-value")
+      {
+        if (get<0>(name_version_score) == "Percolator")
+        {
+          pep_score_type.fromCellString("[MS,MS:XXXXXXX,Percolator peptide-level q-value," + get<1>(name_version_score) + "]");
```
timosachsenberg, line 2716: what to do about these XXX?

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -2686,11 +2703,54 @@ Not sure how to handle these:
...
     for (auto const & se : search_engine_to_runs) // loop over (unique) search engine names
     {
+      //Huge TODO: we need to somehow correctly support peptide-level scores
       MzTabParameter psm_score_type;
+      MzTabParameter pep_score_type;
       const tuple<String, String, String>& name_version_score = se.first;
-      psm_score_type.fromCellString("[,," + get<0>(name_version_score) + " " + get<2>(name_version_score) + "," + get<1>(name_version_score) + "]");
+
+      //TODO we also have to consider the different q-value calculation methods of Percolator....
+      if (get<2>(name_version_score) == "peptide-level q-value")
```
timosachsenberg, line 2712: the problem I see is that the mzTab Peptide section is not really at the peptide level but basically just the consensusID level

jpfeuffer: But what meaning does this value have in mztab?
The scores for the PEP section or scores for peptide sequence level?
I assume the first.

timosachsenberg: yep first. though it is not 100% specified

jpfeuffer:
Will merge for testing. Feel free to still comment, I can address changes in a new PR.

PR #4902, opened 2020-08-29
fix: GLPK thread safety

dmccloskey:
# Description
GLPK is a non-thread safe linear program solver.  In particular, the routine `glp_free` is not thread safe and can result in following exception being thrown:

```
glp_free: memory allocation error
Error detected in file ..\src\glpenv05.c at line 145
```

# Fixes
- Removed the call to `glp_free` via `glp_delete_prob(lp_problem_);` upon calling `~LPWrapper` when using COIN OR
- Removed the instansiation of GLPK memory via the call `lp_problem_ = glp_create_prob();` upon calling `LPWrapper` when using COIN OR
- Added a few exception handling lines for good measure (which were not found to fix the problem) but should not hurt.


src/openms/source/DATASTRUCTURES/LPWrapper.cpp
```diff
@@ -742,21 +750,26 @@ namespace OpenMS
...
-      // if (verbose_level > 0) OPENMS_LOG_INFO << " Branch and cut took " << CoinCpuTime()-time1 << " seconds, "
-      //                                        << model.getNodeCount()<<" nodes with objective "
-      //                                        << model.getObjValue()
-      //                                        << (!model.status() ? " Finished" : " Not finished")
-      //                                        << std::endl;
-      for (Int i = 0; i < model_->numberColumns(); ++i)
-      {
-        solution_.push_back(model.solver()->getColSolution()[i]);
+      
+      try {
```
cbielow, line 754: not a good idea.
If an exception occurs noone will notice because the method returns a seemingly valid value

src/openms/source/DATASTRUCTURES/LPWrapper.cpp
```diff
@@ -66,23 +66,28 @@ namespace OpenMS
 {
 
 
-  LPWrapper::LPWrapper()
+  LPWrapper::LPWrapper(SOLVER solver)
```
cbielow, line 69: not a good idea.
All current code just calls a default c'tor.
This means that now, `solver` is set to COINOR (default-flag).
If COINOR is not installed, then the CTor will effectively do nothing and the object will be in an invalid state.

Currently, the `setSolver()` has a similar problem.

We need to decide if choosing between solvers is even desirable.
If COINOR is available, it should be used, because its much faster than GLPK for all scenarios I know of.

Thus the class choses the solver automatically and we should remove the `setSolver()` member to avoid trouble.

src/openms/source/DATASTRUCTURES/LPWrapper.cpp
dmccloskey: Good point.  The `try/catch` blocks that were inserted are not needed.  I will remove them.

src/openms/source/DATASTRUCTURES/LPWrapper.cpp
dmccloskey: > We need to decide if choosing between solvers is even desirable.
> If COINOR is available, it should be used, because its much faster than GLPK for all scenarios I know of.
> 
> Thus the class choses the solver automatically and we should remove the `setSolver()` member to avoid trouble.

I think this is the most straight forward solution.  An earlier solution that I tried that does just that is moving the GLPK sections below the `#if` COINOR section into an `#else` section so that by default COINOR is chosen if available and GLPK is used as a fall back.  However, I noticed many sections of the code and also tests are set up to run GLPK even if COINOR is available so I went with the current implementation as there would be quite a bit of refactoring needed on parts of the code that I am completely unfamiliar with.

Please let me know how you think we should move forward on this one.  I am not familiar with the other callers to LPWrapper well enough to make the call on this one in regards to what solver is most appropriate.  From my experience in optimization, I can say in general that COINOR is faster and and more accurate than GLPK, and in particular for our use cases COINOR would be preferred.  However, if there are application specific nuances in e.g., proteomics, I would not be in a good position to make that decision.

cbielow: I'd say that is enough anecdotal evidence :)
Lets just go ahead then and use COIN-OR if available and fall back to GLPK if needed.
Can you just revive the default Ctor and remove the setSolver(), and then lets see what happens.

dmccloskey:
@cbielow Please let me know your thoughts on the recent changes.  It looks like swapping GLPK for COINOR made no difference on any of the tests at least.

src/tests/class_tests/openms/source/LPWrapper_test.cpp
```diff
@@ -331,55 +330,48 @@ END_SECTION
 
 START_SECTION((void writeProblem(const String &filename, const WriteFormat format) const ))
 {
-  if(lp.getSolver() == LPWrapper::SOLVER_GLPK)
```
cbielow, line 334: I would leave the `if (getSolver() ==`) else if (getsolver ==)`, because it saves you from knowing anything about the implementation details and avoids the `#ifdef` ugliness in the test.

src/tests/class_tests/openms/source/LPWrapper_test.cpp
```diff
@@ -462,16 +454,14 @@ START_SECTION((void getMatrixRow(Int idx,std::vector<Int>& indexes)))
...
-  lp4.setSolver(LPWrapper::SOLVER_GLPK);
-  TEST_EQUAL(lp4.getSolver(),LPWrapper::SOLVER_GLPK)
-}
-END_SECTION
-
 START_SECTION((SOLVER getSolver() const ))
 {
-  TEST_EQUAL(lp4.getSolver(),LPWrapper::SOLVER_GLPK)
+
+#if COINOR_SOLVER==1
```
cbielow, line 460: although here it is required, I guess.

cbielow (approved):
perfect!
thanks!

PR #4903, opened 2020-08-30
[FIX] mzTab opt columns bug, now also for PEP section

jpfeuffer:


jpfeuffer:
Merging since it's a rather obvious fix

PR #4905, opened 2020-08-31
Calculating a Corrected Database Suitability

Waschi97:
**[edit from Chris: this is an extension of GSoC and already goes beyond what was required to complete the project successfully!]**

The number of top deNovo hits is corrected with a correction factor. This results in the database suitability behaving linear rather than hyperbolic.

This factor is calculated as the ratio of top database hits needed for a suitability of 1.0 and the number of deNovo identifications found when only searching with the deNovo "database".
The later is calculated with an internal call of a search adapter (the one found in the search parameters, i.e. the one the user used).
To calculate the number of top db hits at suitability 1.0 an extrapolation is needed. Since it was observed that both top deNovo and database hits behave linear this extrapolation can be done.
A second internal identification search is done with a smaller sample of the database in question. The top deNovo and database hits for this sampled database are calculated and used for the extrapolation.

----------------------------------------------------------------------------------------------------------------------------------------------

The algorithm described above was implemented into `DBSuitability`.

`DatabaseSuitability` TOPP tool input changed to have all needed data for the correction.

Still to do:
- [x] adapt docu

- [x] change class test for `DBSuitability` and TOPP test for `DatabaseSuitability`


src/openms/source/QC/DBSuitability.cpp
```diff
@@ -101,102 +91,51 @@ namespace OpenMS
...
-
-    if (data.num_top_db == 0 && data.num_top_novo == 0)
-    {
-      OPENMS_LOG_WARN << "Identifications could not be assigned to either the database or the deNovo protein. Probably your FDR threshold is too strict." << endl;
-      data.suitability = DBL_MAX;
-      return;
-    }
+    // calculate suitability
+    SuitabilityData d;
+    results_.push_back(d);
```
cbielow, line 96: `results_.push_back(SuitabilityData());` avoids accidentally referring back to `d`, which would be a mistake

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -101,102 +91,51 @@ namespace OpenMS
...
+    SuitabilityData& data = results_.back();
+    calculateSuitability_(pep_ids, data);
 
-    data.suitability = double(data.num_top_db) / (data.num_top_db + data.num_top_novo);
+    // calculate correction of suitability with extrapolation
+    
+    // sampled run
+    // maybe multiple runs? could be controlled with a parameter
+    double ratio = 0.5;
```
cbielow, line 104: call this `subsampling_rate`? `ratio` is a bit misleading

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -101,102 +91,51 @@ namespace OpenMS
...
+    vector<FASTAFile::FASTAEntry> sampled_db = getSubsampledFasta_(original_fasta, ratio);
+    sampled_db.insert(sampled_db.end(), novo_fasta.begin(), novo_fasta.end());
+    calculateDecoys_(sampled_db);
```
cbielow, line 107: `appendDecoys`?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -101,102 +91,51 @@ namespace OpenMS
...
+    vector<PeptideIdentification> subsampled_ids = runIdentificationSearch_(exp, sampled_db, search_info.first, search_info.second);
+
+    SuitabilityData sampled_data;
+    calculateSuitability_(subsampled_ids, sampled_data);
+
+    // slopes of db and deNovo hits
+    double db_slope = (int(sampled_data.num_top_db) - int(data.num_top_db)) / (-ratio);
```
cbielow, line 114: `*data` is very generic.
Prepend with `suitability_` + `_sampled/_full` instead?

cbielow: technically, you need `full - sampled / (1 - ratio)`, right?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -101,102 +91,51 @@ namespace OpenMS
...
+    calculateDecoys_(novo_fasta);
+    Int deNovo_intercept = countIdentifications_(runIdentificationSearch_(exp, novo_fasta, search_info.first, search_info.second));
+    // db_intercept is estimated to be 0
+
+    // calculate correction factor
+    double target_ratio = (-deNovo_intercept) / deNovo_slope; // ratio that corresponds to suitability of 1
+    double db_hits_at_ratio = db_slope * target_ratio; // db hits for suitability 1
+    double factor = db_hits_at_ratio / deNovo_intercept;
+
+    data.corr_factor = factor;
```
cbielow, line 127: put these three lines which compute corrected factors into a member of `data`, e.g. `setCorrectionFactor(double factor)`

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+    String adapter;
+    vector<String> keys;
+    search_params.getKeys(keys);
+    for (const String& key : keys)
+    {
+      if (adapter.empty()) // trying to find which adapter was used
+      {
+        for (const String& a : working_adapters)
+        {
+          if (key.find(a) == 0)
```
cbielow, line 253: you are looking for any infix?
Prefix + `:` should be enough?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+          {
+            // used adapter found
+            adapter = a;
```
cbielow, line 256: forgot putting the key of this hit into `p`?

cbielow: maybe make two separate loops?
one for finding the adapter name, the other for extracting the keys?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
     if (hit.getScore() > FDR) return false;
     return true;
   }
-}
+
+  pair<String, Param> DBSuitability::extractSearchAdapterInfoFromMetaValues_(const ProteinIdentification::SearchParameters& search_params) const
+  {
+    Param p;
+    // list of all allowed adapters
+    vector<String> working_adapters{ "CometAdapter", "CruxAdapter", "MSGFPlusAdapter", "MSFraggerAdapter", "MyriMatchAdapter", "OMSSAAdapter", "XTandemAdapter" };
```
cbielow, line 242: would automatic extraction using all names from ToolHandler that end in "Adapter" also work?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+        if (key.find(adapter) != 0) continue; // does adapter appear in meta value key?
+        p.setValue(key, search_params.getMetaValue(key));
+      }
+    }
+
+    if (p.empty()) // non of the allowed adapter names where found in the meta values
+    {
+      String message;
+      message = "No parameters found for any of the allowed adapters in the given meta values. Allowed are:\n";
+      for (const String& a : working_adapters)
```
cbielow, line 272: `ListUtils::concatenate(...)`

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+    }
+
+    OPENMS_LOG_DEBUG << "Parameters for the following adapter were found: " << adapter << endl;
+
+    return make_pair(adapter, p);
+  }
+
+  void DBSuitability::writeIniFile_(const Param& parameters, const String& filename) const
+  {
+    if (!File::writable(filename)) // check if file exists and is writable
```
cbielow, line 287: Does ParamXMLFile not take care of this?

src/openms/source/QC/DBSuitability.cpp
jpfeuffer: Dont think so, e.g. FidoAdapter and a few others do not do DB search

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+    }
+
+    // temporary folder for search in- und output files
+    File::TempDir tmp_dir;
+    String mzml_path = tmp_dir.getPath() + "spectra.mzML";
+    String db_path = tmp_dir.getPath() + "database.FASTA";
+    String out_path = tmp_dir.getPath() + "out.idXML";
+
+    // override the in- and output files in the parameters
+    parameters.setValue(adapter_name + ":1:in", mzml_path);
```
cbielow, line 309: sanity check: make sure these parameters exist. Otherwise you create new ones which will be ignored (or crash the tool)

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+    PeptideIndexing indexer;
+    FASTAContainer<TFI_Vector> proteins(fasta_data);
+    OPENMS_LOG_DEBUG << "Running PeptideIndexer functionalities ..." << endl << endl;
+    OPENMS_LOG_INFO.remove(cout); // prevent indexer from writing statistic
+    PeptideIndexing::ExitCodes indexer_exit = indexer.run(proteins, prot_ids, pep_ids);
+    OPENMS_LOG_INFO.insert(cout); // revert logging change
+    if (indexer_exit != PeptideIndexing::ExitCodes::EXECUTION_OK)
+    {
+      OPENMS_LOG_ERROR << "An error occured while trying to index the search results." << endl;
+      throw Exception::UnableToCreateFile(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Return state was: " + static_cast<Int>(indexer_exit));
```
cbielow, line 356: probably not the right kind of exception?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+    }
+
+    // calculate q-values
+    Param p;
+    p.setValue("use_all_hits", "true");
+    p.setValue("add_decoy_peptides", "true");
+    p.setValue("add_decoy_proteins", "true");
```
cbielow, line 363: we should really have a function `Param::overwriteValue` which checks if the value exists and fails otherwise....
This would make this much more safe...
@jpfeuffer , opinions?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+
+    FalseDiscoveryRate fdr;
+    fdr.setParameters(p);
+    OPENMS_LOG_DEBUG << "Calculating q-values ..." << endl << endl;
+    fdr.apply(pep_ids);
+
+    return pep_ids;
+  }
+
+  Size DBSuitability::countIdentifications_(std::vector<PeptideIdentification> pep_ids) const
```
cbielow, line 373: const &

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+  {
+    Size count{};
+    for (const auto& pep_id : pep_ids)
+    {
+      vector<PeptideHit> hits = pep_id.getHits();
```
cbielow, line 378: const &

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+      if (hits.empty()) continue;
+      if (!hits[0].metaValueExists("target_decoy"))
+      {
+        throw Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No target/decoy annotation found. Make sure PeptideIndexer ran beforehand.");
+      }
+      if (hits[0].getMetaValue("target_decoy") == "decoy") continue;
+      if (!passesFDR_(hits[0],this->getParameters().getValue("FDR"))) continue;
```
cbielow, line 385: slow. store FDR in double value outside of loop

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+
+      ++count;
+    }
+    return count;
+  }
+
+  std::vector<FASTAFile::FASTAEntry> DBSuitability::getSubsampledFasta_(const std::vector<FASTAFile::FASTAEntry>& fasta_data, double ratio) const
```
cbielow, line 392: ratio --> subsampling_rate ?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+      throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Ratio has to be between 0 and 1. Aborting!");
+    }
+    Size num_AS{};
+    for (const auto& entry : fasta_data)
+    {
+      num_AS += entry.sequence.size();
+    }
+    double num_AS_written = num_AS * ratio;
+
+    random_shuffle(fasta_data.begin(), fasta_data.end());
```
cbielow, line 405: `fasta_data` is const... why does this work???

please use 
```
  std::random_device rd;
  std::mt19937 g(rd());
  std::shuffle(v.begin(), v.end(), g);
```
since `random_shuffle` is removed in C++17...

src/openms/source/QC/DBSuitability.cpp
jpfeuffer: yes definitely. I often had problems when misspelling.

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -290,9 +229,320 @@ namespace OpenMS
...
+      data.suitability = DBL_MAX;
+      return;
+    }
+
+    data.suitability = double(data.num_top_db) / (data.num_top_db + data.num_top_novo);
+  }
+
+  void DBSuitability::calculateDecoys_(std::vector<FASTAFile::FASTAEntry>& fasta) const
+  {
+    vector<FASTAFile::FASTAEntry> decoys;
```
cbielow, line 526: append directly to `fasta`?
you can also reserve beforehand

cbielow (commented):
nice!

src/openms/source/QC/DBSuitability.cpp
cbielow: this would also have the advantage that the description etc stays the same. Currently, setValue() removes it....

src/openms/source/QC/DBSuitability.cpp
jpfeuffer: agree, I was confused.

cbielow:
todo: adapt dot plot with new inputs

src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
```diff
@@ -348,6 +348,25 @@ class DecoyHelper
...
+  // returns a regex string to find decoy prefixes
+  static std::string getPrefixRegex()
+  {
+    return std::string("^(") + ListUtils::concatenate<std::string>(getAffixes(), "_*|") + "_*)";
+  }
+
+  // returns regex string to find decoy suffixes
+  static std::string getSuffixRegex()
+  {
+    return std::string("(") + ListUtils::concatenate<std::string>(getAffixes(), "_*|") + "_*)$";
```
cbielow, line 367: I know this was just refactored from existing code, by in my mind, this should be 
`return std::string("(*_") + ListUtils::concatenate<std::string>(getAffixes(), "|*_") +")$";`

Is there a test for this?

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -87,8 +94,15 @@ namespace OpenMS
       /// this is normalized by mw
       double cut_off = DBL_MAX;
 
+      /// #IDs with only deNovo search / #IDs with only database search
+      /// used for correcting the number of deNovo hits
+      /// worse databases will have less IDs then good databases
```
cbielow, line 99: `then` --> `than`

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -189,110 +529,82 @@ namespace OpenMS
...
-      if (acc.find(Constants::UserParam::CONCAT_PEPTIDE) == String::npos)
-      {
-        return false;
-      }
+      throw(Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No suitability data found. Can't apply correction factor."));
     }
-    return true;
+    corr_factor = factor;
+    num_top_novo_corr = num_top_novo * factor;
+    suitability_corr = num_top_db / (num_top_db + num_top_novo_corr);
```
cbielow, line 607: these three member should be private, just to make sure they are not modified 'by hand', but only via this function.
You'd also need three small `double getCorrectionFactor() const { return corr_factor; }` etc functions in the header then.

src/tests/class_tests/openms/source/DBSuitability_test.cpp
```diff
@@ -49,9 +49,43 @@
...
+#include <OpenMS/FORMAT/IdXMLFile.h>
+#include <OpenMS/KERNEL/MSExperiment.h>
+
+
 START_TEST(Suitability, "$Id$")
 
-/////////////////////////////////////////////////////////////
+MSExperiment exp;
+MzMLFile m;
+m.load("C:\\Development\\TOPPAS_Outputs\\Corrected_Suitability_test_files\\TOPPAS_out\\006-FileConverter-out\\BSA1.mzML", exp);
```
cbielow, line 62: you need platform independent paths here.
Try `m.load(File::find("./examples/BSA/BSA1.mzML"));`

src/tests/class_tests/openms/source/DBSuitability_test.cpp
```diff
@@ -49,9 +49,43 @@
...
+p.setValue("FDR", 0.8);
+s.setParameters(p);
+s.compute(pep_ids, exp, fasta_data, novo_fasta, prot_ids[0].getSearchParameters());
+
+DBSuitability::SuitabilityData data = s.getResults()[0];
+cout << "top db:" << data.num_top_db << endl;
+cout << data.num_top_novo << " " << data.num_top_novo_corr << endl;
+cout << data.suitability << " " << data.suitability_corr << endl;
+cout << data.corr_factor << endl;
+
```
cbielow, line 86: this testfile still needs some actual tests... :)

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -203,7 +269,128 @@ namespace OpenMS
...
+    * @brief Creates a subsampled fasta with the given subsampling rate
+    *
+    * The subsampling is based on the number of amino acides and not on the number of fasta entries.
+    *
+    * @param fasta_data         fasta of which the subsampling should be done
+    * @param subsampling_rate   subsampling rate to be used [0,1]
+    * @returns                  fasta entries with total number of AA = original number of AA * subsampling_rate
+    * @throws                   IllegalArgument if subsampling rate is not between 0 and 1
+    */
+    std::vector<FASTAFile::FASTAEntry> getSubsampledFasta_(std::vector<FASTAFile::FASTAEntry> fasta_data, double subsampling_rate) const;
```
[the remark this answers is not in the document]
jpfeuffer: Ah you are shuffling. You could instead just draw random numbers in the range of the indices and copy out the according fasta entries. Might be worthwhile for gigabytes of fast.
There is also an on-disc fasta file container I think. But this has no random access.

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -110,33 +142,64 @@ namespace OpenMS
...
+    * @param search_params      SearchParameters object containing information which adapter
+    *                           was used with which settings for the identification search
+    *                           that resulted in @p pep_ids
+    * @throws                   MissingInformation if no target/decoy annotation is found on @p pep_ids
+    * @throws                   MissingInformation if no xcorr is found,
+    *                           this happends when another adapter than CometAdapter was used
+    * @throws                   Precondition if a q-value is found in @p pep_ids
     */
-    void compute(std::vector<PeptideIdentification> pep_ids);
+    void compute(std::vector<PeptideIdentification> pep_ids, const MSExperiment& exp, std::vector<FASTAFile::FASTAEntry> original_fasta, std::vector<FASTAFile::FASTAEntry> novo_fasta, const ProteinIdentification::SearchParameters& search_params);
```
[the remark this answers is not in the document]
jpfeuffer: See comment at subsampleFasta method

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+
+    // get the score of the first two decoy hits
+    double decoy_1 = DBL_MAX;
+    double decoy_2 = DBL_MAX;
+    UInt curr_hit = 0;
+
+    for (const auto& hit : pep_id.getHits())
     {
-      data.cut_off = getDecoyCutOff_(pep_ids, cut_off_percentile);
+      if (curr_hit > 10) break;
```
jpfeuffer, line 161: Is this documented?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+    Size num_AA{};
+    for (const auto& entry : fasta_data)
+    {
+      num_AA += entry.sequence.size();
+    }
+    double num_AS_written = num_AA * subsampling_rate;
+
+    //random_device rd();
+    //mt19937 g(rd());
+    mt19937 g(0);
```
jpfeuffer, line 418: fixed seed?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+    shuffle(fasta_data.begin(), fasta_data.end(), g);
```
jpfeuffer, line 419: How about shuffling the indices instead? Then you can take the fasta as const ref.

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+    for (const auto& entry : fasta_data)
+    {
+      if (curr_AA >= num_AS_written) break;
+      sampled_fasta.push_back(entry);
+      curr_AA += entry.sequence.size();
+    }
+    return sampled_fasta;
+  }
+
+  void DBSuitability::calculateSuitability_(std::vector<PeptideIdentification> pep_ids, SuitabilityData& data) const
```
jpfeuffer, line 432: Here, since it is a private method if you assume the Hits in the IDs to be sorted (by e.g. sorting them before manually).
Or by not caring if the IDs come out sorted, you could also take the IDs as (const) ref.

jpfeuffer (changes requested):

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+    }
+
+    nth_element(diffs.begin(), diffs.begin() + index, diffs.end());
+
+    return diffs[index];
+  }
+
+  bool DBSuitability::isNovoHit_(const PeptideHit& hit) const
+  {
+    const boost::regex decoy_pattern(DecoyHelper::getPrefixRegex() + "|" + DecoyHelper::getSuffixRegex());
```
cbielow, line 228: instanciating the regex is very expensive IFAIR.
Maybe make it a member variable?
How are we doing speedwise anyway?

src/topp/DatabaseSuitability.cpp
```diff
@@ -139,12 +143,16 @@ class DatabaseSuitability :
...
-    registerInputFile_("in_id", "<file>", "", "Input idXML file from peptide search with combined database with added de novo peptide. PeptideIndexer is needed, FDR is forbidden.");
+    registerInputFile_("in_id", "<file>", "", "Input idXML file from a peptide identification search with a combined database. PeptideIndexer is needed, FDR is forbidden.");
     setValidFormats_("in_id", { "idXML" });
     registerInputFile_("in_spec", "<file>", "", "Input MzML file used for the peptide identification");
     setValidFormats_("in_spec", { "mzML" });
     registerInputFile_("in_novo", "<file>", "", "Input idXML file containing de novo peptides (unfiltered)");
     setValidFormats_("in_novo", { "idXML" });
+    registerInputFile_("database", "<file>", "", "Input FASTA file of the database in question (with decoys)");
+    setValidFormats_("database", { "FASTA" });
+    registerInputFile_("novo_database", "<file>", "", "Input deNovo sequences derived from MzML given in 'in_spec' concatenated to one FASTA entry (with decoy)");
```
cbielow, line 154: should be 'without' decoy, right?
Because internally, you append decoys after subsampling.

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -203,7 +269,128 @@ namespace OpenMS
...
+    void appendDecoys_(std::vector<FASTAFile::FASTAEntry>& fasta) const;
+
+    /**
+    * @brief Returns the cross correlation score normalized by MW (if existing), else if the 'force' flag is set the current main score is returned
+    *
+    * @param pep_hit    PeptideHit of which the score is needed
+    * @returns          cross correlation score normalized by MW or current score
+    * @throws           MissingInformation if no xcorr is found and 'force' flag isn't set
+    */
+    double getRightScore_(const PeptideHit& pep_hit) const;
```
cbielow, line 367: the `Right` is a bit confusing IMHO.
Maybe `extractScore`?

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -203,7 +269,128 @@ namespace OpenMS
...
+    *                   this happends when another adapter than CometAdapter was used
+    */
+    void calculateSuitability_(std::vector<PeptideIdentification> pep_ids, SuitabilityData& data) const;
+
+    /**
+    * @brief Calculates and appends decoys to a given vector of FASTAEntry
+    *
+    * Each sequence is digested with Trypsin. The resulting peptides are reversed and appended to one another.
+    * This results in the decoy sequences.
+    * The idetifier is given a 'DECOY_' prefix.
```
cbielow, line 354: typo

src/openms/include/OpenMS/QC/DBSuitability.h
cbielow, line 202: this effectively copies `pep_ids` from the calling scope.
you need to modify the content, right?
the cleanest is probably to use `std::vector<PeptideIdentification>&& pep_ids` and then call it using `compute(std::move(peptides), ....)`...
Or just take by reference, but make it clear in the docs that the content will be modified.

cbielow: also the other vectors could be taken by reference instead of copy?

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -203,7 +269,128 @@ namespace OpenMS
...
+    /**
+    * @brief Returns a SuitabilityData object containing the data if re-ranking didn't happen
+    *
+    * Cases that are re-ranked are already counted. To get the 'no re-ranking' data these cases need to be
+    * subtracted from the number of top database hits and added to the number of top deNovo hits.
+    *
+    * @param data    actual suitability data
+    * @returns       simulated suitability data where re-ranking didn't happen
+    */
+    SuitabilityData simulateNoReRanking_(const SuitabilityData& data) const;
```
cbielow, line 378: maybe make this a member of `SuitabilityData`? Just to be closer to the data model?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+        diffs.push_back(diff);
+      }
+    }
+
+    if (double(diffs.size()) / pep_ids.size() < 0.2)
+    {
+      throw(Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Under 20 % of peptide identifications have two decoy hits. This is not enough for re-ranking. Use the 'no_rerank' flag to still compute a suitability score."));
+    }
+
+    // sort the diffs decreasing and get the percentile one
```
cbielow, line 213: could be any percentile, right? not the first necessarily.

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+    vector<String> keys;
+    search_params.getKeys(keys);
+
+    // find adapter name
+    String adapter;
+    for (const String& key : keys)
+    {
+      for (const String& a : working_adapters)
+      {
+        if (key.compare(0, a.size(), a) == 0)
```
cbielow, line 261: a complicated way of saying `if (key == a)` ???

cbielow: ahh.. ok. you actually want the prefix, right?
Maybe annotate that...
and maybe also inlcude the ':' in the comparison?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+
+    // load result
+    vector<ProteinIdentification> prot_ids;
+    vector<PeptideIdentification> pep_ids;
+    IdXMLFile id_file;
+    id_file.load(out_path, prot_ids, pep_ids);
+
+    if (keep_files)
+    {
+      id_file.store(tmp_dir.getPath() + "id_search_out.idXML", prot_ids, pep_ids);
```
cbielow, line 359: you could just move the temporary file?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
 
+    if (keep_files)
+    {
+      id_file.store(tmp_dir.getPath() + "indexed_post_FDR.idXML", prot_ids, pep_ids);
+    }
+
+    return pep_ids;
+  }
+
+  std::vector<FASTAFile::FASTAEntry> DBSuitability::getSubsampledFasta_(std::vector<FASTAFile::FASTAEntry> fasta_data, double subsampling_rate) const
```
cbielow, line 403: accept vector as `const &` and ...

src/openms/source/QC/DBSuitability.cpp
cbielow, line 419: shuffle a new vector of indices
(the DB could be huge...)
```
std::vector<int> rnd_indices(fasta_data.size()) ;
std::iota (std::begin(rnd_indices), std::end(rnd_indices), 0); // Fill with 0, 1, ....
```

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,52 +65,387 @@ namespace OpenMS
...
+    double num_AS_written = num_AA * subsampling_rate;
+
+    //random_device rd();
+    //mt19937 g(rd());
+    mt19937 g(0);
+    shuffle(fasta_data.begin(), fasta_data.end(), g);
+
+    Size curr_AA{};
+    vector<FASTAFile::FASTAEntry> sampled_fasta;
+    for (const auto& entry : fasta_data)
```
cbielow, line 423: ... then iterate through
`for (const int i : rnd_indices) ....`
until you have enough data collected

src/openms/source/QC/DBSuitability.cpp
cbielow, line 432: pass vector by const &

jpfeuffer:
I guess your decoy regex change needs changes in the PEptideIndexer test.

Waschi97:
hmm tests are working on my side ...

jpfeuffer:
Hmm maybe try to merge first

Waschi97:
Well, still working after merging on my side. Let's wait for the builds....

jpfeuffer:
Tada

Waschi97:
nice!

jpfeuffer:
There are a few open comments from me still, can you accept or answer them quickly? Then it is good to go I think

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -597,6 +616,78 @@ namespace OpenMS
...
+        {
+          throw(Exception::MissingInformation(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No target/decoy information found! Make sure 'PeptideIndexer' is run beforehand."));
+        }
+        if (hit.getMetaValue("target_decoy") == "decoy") continue;
+
+        // insert protein accessions
+        const set<String> accessions = hit.extractProteinAccessionsSet();
+        for (const String& acc : accessions)
+        {
+          String acc_copy(acc);
```
cbielow, line 641: why copy the string?

Waschi97: to search for the decoy pattern in line 646 the accession needs to be lower case

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -55,58 +66,419 @@ namespace OpenMS
...
+
+    if (pep_ids.empty())
+    {
+      OPENMS_LOG_WARN << "No peptide identifications found in given idXML! No calculations performed." << endl;
+      return;
+    }
+
+    if (!no_re_rank)
+    {
+      data.cut_off = getDecoyCutOff_(pep_ids, cut_off_fract);
```
cbielow, line 476: setting `data.cutoff` (or any other member of `data`) is a bit dangerous, since you never know what `data` was previously populated with (maybe the user passed in an old object?).
Better create a fresh `SuitabilityData` locally, and return it.
Or was there a specific reason to do it this way?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -597,6 +616,78 @@ namespace OpenMS
...
   }
 
+  UInt DBSuitability::numberOfUniqueProteins_(const std::vector<PeptideIdentification>& peps, UInt number_of_hits) const
+  {
+    set<String> proteins;
+
+    for (const auto& pep : peps)
+    {
+      const vector<PeptideHit>& hits = pep.getHits();
+      for (UInt i = 0; i < number_of_hits; ++i)
```
cbielow, line 626: possible reason why the number of proteins does not change much during subsamping: you use ALL hits.
Better use the top hit only?

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -597,6 +616,78 @@ namespace OpenMS
...
+        // insert protein accessions
+        const set<String> accessions = hit.extractProteinAccessionsSet();
+        for (const String& acc : accessions)
+        {
+          String acc_copy(acc);
+          if (acc.find(Constants::UserParam::CONCAT_PEPTIDE) != String::npos) // skip novo accessions
+          {
+            continue;
+          }
+          if (boost::regex_search(acc_copy.toLower(), decoy_pattern_)) // skip decoy accessions
```
cbielow, line 646: ok, that;s why you copy...
but then maybe do `if (boost::regex_search(String(acc).toLower(), decoy_pattern_))`, to make the copy later

src/openms/source/QC/DBSuitability.cpp
```diff
@@ -597,6 +616,78 @@ namespace OpenMS
...
+    for (int i = 0; i < data.size(); ++i)
+    {
+      Size num_top_novo = data[i].num_top_novo;
+      novo_data.push_back(num_top_novo);
+      novo_hits_to_data[num_top_novo] = i;
+    }
+
+    UInt curr_index = 0;
+    double best_diff = DBL_MAX;
+    double avg_novo_hits = Math::mean(novo_data.begin(), novo_data.end());
```
cbielow, line 677: median is more robust?!
If the number of items is even, it's a bit weird, because you must decide which of both 'median' items to take (they are equally far away from the computed median). Thus you either always underestimate, or always overestimate...
The only solution is to sample an odd number of times. Then the median is well defined.

src/openms/source/QC/DBSuitability.cpp
Waschi97: Ahh.. Yes, that could be the reason. I'll try that.

src/openms/source/QC/DBSuitability.cpp
Waschi97: I use the mean to calculate the median by hand. Because I want the value which is closest to the mean, right? `Math::median` just gives me the value which is located in the middle of the sorted list, which isn't always the same. Am I wrong?

src/openms/source/QC/DBSuitability.cpp
Waschi97: This is done because an empty `SuitabilityData` object is pushed into the results vector at the very beginning and all changes are made on a reference to that object.
I guess I could implement a `clear()` function and clear the object at the beginning of `calculateSuitability_`.

src/openms/source/QC/DBSuitability.cpp
cbielow: somewhat...
the median can be miles away from the mean... so better not trust the mean in any way.

cbielow:
rebuild jenkins

src/openms/include/OpenMS/QC/DBSuitability.h
```diff
@@ -429,8 +428,27 @@ namespace OpenMS
     * @returns        index to object with median number of de novo hits
     */
     Size getIndexWithMedianNovoHits_(const std::vector<SuitabilityData>& data) const;
+
+    /**
+    * @brief Extracts the worst score that still makes a FDR (q-value) threshold
```
jpfeuffer, line 433:
```suggestion
    * @brief Extracts the worst score that still passes a FDR (q-value) threshold
```

cbielow:
some minor refactoring could be done (e.g. move the subsampling to a more public place), but I think its good to go.
Lets wait for the tests...
BTW: any reason the workers status here is failed, even though the CDash upload worked and showed no errors?

jpfeuffer:
Maybe only build was uploaded? Test XML definitely failed with 500 in the logs

timosachsenberg:
tests passed

timosachsenberg (approved):

PR #4908, opened 2020-09-03
re-implement StopWatch to allow for 'resume()' and querying time while still running

cbielow:
Cleans up the implementation and fixes a bug (querying a stopwatch without stop() would yield arbitrary numbers -- at least on Windows)

Also allows to `start()` ... `stop()` `<do not time this>` `resume()`  to allow for more fine-grained measurements (e.g. benchmarking).

It also prevents accidental wrong usage (e.g. stop()'ing twice) by throwing exceptions.

### DIFF
--- a/src/openms/include/OpenMS/SYSTEM/StopWatch.h
+++ b/src/openms/include/OpenMS/SYSTEM/StopWatch.h
@@ -50,9 +50,15 @@
 namespace OpenMS
 {
   /**
-      @brief StopWatch Class.
+      @brief This class is used to determine the current process' CPU (user and/or kernel) and wall time.
 
-      This class is used to determine the current process time.
+      CPU time is measured as sum of all threads of the current process.
+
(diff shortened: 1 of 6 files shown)


cbielow:
ready for merge

timosachsenberg:
could some of the new code in the header files be moved to the c++ files?

cbielow:
yes, but the `TimeDiff_` stuff is not that much.
I thought about using PIMPL. This would even avoid including any`*time` headers. But it forces me to use new/delete... so I thought not worth the trouble... Any strong opinions?

timosachsenberg:
PIMPL might be overkill. 
Just thought we were pretty good in moving implementation to cpp files so I think we should continue that.

cbielow:
I tried PIMPL and it just makes the code look ugly, due to forwarding functions and new/delete (could be solved by unique_ptr, at the cost of including `<memory>`) and more code in general.
So I simply moved stuff to cpp.

PR #4909, opened 2020-09-03
[FEATURE] FIAMS CV params in mzML

jpfeuffer:
# Description

Adds new mzML CV params. See gitter.

# Checklist:

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -1647,7 +1647,8 @@ namespace OpenMS
         else if (accession == "MS:1001581") //FAIMS compensation voltage
         {
           //No member => meta data
-          spec_.setMetaValue("FAIMS compensation voltage", termValue);
+          //TODO CV term is wrongly annotated without an xref data type -> cast to double
+          spec_.setMetaValue("FAIMS compensation voltage", value.toDouble());
         }
         //scan polarity
         else if (accession == "MS:1000129") //negative scan
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -1646,7 +1646,8 @@ namespace OpenMS
         }
         else if (accession == "MS:1001581") //FAIMS compensation voltage
         {
-          //No member => meta data
+          // According to the PSI-MS ontology this term should be stored below the "scan" and not "spectrum" parent.
+          // Some pwiz version put this term on the "spectrum" level so we also read it here.
           //TODO CV term is wrongly annotated without an xref data type -> cast to double
           spec_.setMetaValue("FAIMS compensation voltage", value.toDouble());
         }
@@ -2111,6 +2112,12 @@ namespace OpenMS
 
           spec_.setDriftTime(value.toDouble());
           spec_.setDriftTimeUnit(unit);
+        }
+        else if (accession == "MS:1001581") //FAIMS compensation voltage.
+        {
+          //No member => meta data
+          //TODO CV term is wrongly annotated without an xref data type -> cast to double
+          spec_.setMetaValue("FAIMS compensation voltage", value.toDouble());
         }
         else if (accession == "MS:1000011") //mass resolution
         {
--- a/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
@@ -1644,6 +1644,11 @@ namespace OpenMS
           //No member => meta data
           spec_.setMetaValue("peak list raw scans", termValue);
         }
+        else if (accession == "MS:1001581") //FAIMS compensation voltage
+        {
+          //No member => meta data
+          spec_.setMetaValue("FAIMS compensation voltage", termValue);
+        }
         //scan polarity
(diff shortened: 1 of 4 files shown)


timosachsenberg:
also make sure it is written out in the same way... otherwise, it gets lost

jpfeuffer:
Question @OpenMS/core : Would it make sense to parse all CV params by default?

jpfeuffer:
@timosachsenberg We do not have a writing mechanism for parsed CV terms. So this is currently not possible.

timosachsenberg:
I think this would make sense but sounds quite involved. Could potentially also consume more memory.

hroest:
what about writing it back again? with the right term?

hroest:
> @timosachsenberg We do not have a writing mechanism for parsed CV terms. So this is currently not possible.
> I.e. we would need an equally long if-chain to write the values out.

I think that is what we currently do, isnt it?

jpfeuffer:
Ah maybe you are right. Does it try to look up the name (not ID) of the CV term here?
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp#L3468

I thought we would need to store the ID as well.

timosachsenberg:
ah ok I meant you want to store it as cvTerm on the MSExperiment side, too.

jpfeuffer:
Do we need to update the PSI.obo for this?

jpfeuffer:
Actually after getting the file, it seems like this is the only term missing in our parser.

jpfeuffer:
Unfortunately, the CV term is registered completely wrong. It is registered as a *scan* not *spectrum* attribute and does not even have an xref for the data type specified.

jpfeuffer:
If it is instead really a scan attribute, then the whatever converter was used writes it wrong.

jpfeuffer:
@hroest Where does this attribute fit better in your opinion? In the spectrum or spectrum/scanList/scan element? Then we can raise issues upstream.

jpfeuffer:
By the way, there is already an entry (filter string) in the spectrum/scanList/scan, that contains the "**c**ompensation **v**oltage".

```
<cvParam cvRef="MS" accession="MS:1000512" name="filter string" value="FTMS + p NSI cv=-50.00 Full ms [197.0777-1500.0000]"/>
              <cvParam cvRef="MS" accession="MS:1000927" name="ion injection time" value="50.0" unitCvRef="UO" unitAccession="UO:0000028" unitName="millisecond"
```

jpfeuffer:
So it seems like it indeed should be placed and expected in the scan element but it was not yet fixed upstream in pwiz.
https://github.com/ProteoWizard/pwiz/issues/747

When I find time I will update this PR to do it the correct way. Seems like pwiz plans to fix it in a non-backward compatible way anyhow (which makes sense).

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -1644,6 +1644,14 @@ namespace OpenMS
           //No member => meta data
           spec_.setMetaValue("peak list raw scans", termValue);
         }
+        else if (accession == "MS:1001581") //FAIMS compensation voltage
```
timosachsenberg, line 1647: below "spectrum" parent

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -1741,7 +1749,7 @@ namespace OpenMS
             chromatogram_.getPrecursor().getPossibleChargeStates().push_back(value.toInt());
           }
         }
-        else if (accession == "MS:1002476" || accession == "MS:1002815") //ion mobility drift time
+        else if (accession == "MS:1002476" || accession == "MS:1002815" || accession == "MS:1001581") //ion mobility drift time or FAIM compensation voltage
```
timosachsenberg, line 1752: below "scan"

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -2081,7 +2093,7 @@ namespace OpenMS
           //No member => meta data
           spec_.setMetaValue("dwell time", termValue);
         }
-        else if (accession == "MS:1002476" || accession == "MS:1002815") //ion mobility drift time
+        else if (accession == "MS:1002476" || accession == "MS:1002815" || accession == "MS:1001581") //ion mobility drift time or FAIMS compensation voltage
```
timosachsenberg, line 2096: below precursor (probably not used)

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -5121,8 +5139,13 @@ namespace OpenMS
         {
           os << "\t\t\t\t\t\t<cvParam cvRef=\"MS\" accession=\"MS:1000016\" name=\"scan start time\" value=\"" << spec.getRT()
              << "\" unitAccession=\"UO:0000010\" unitName=\"second\" unitCvRef=\"UO\" />\n";
-          // if drift time was never set, don't report it
-          if (spec.getDriftTime() >= 0.0)
+
+          if (spec.getDriftTimeUnit() == MSSpectrum::DriftTimeUnit::FAIMS_COMPENSATION_VOLTAGE)
+          {
```
timosachsenberg, line 5144: writing out below "scan" only

jpfeuffer:
Great! Thanks @timosachsenberg

PR #2080, opened 2016-07-29
[FIX] adopt activation modes to new obo

hroest:
It seems that some time ago, in 4fea083fd6a3fc820216ef3298860aad5cb5784d the HCD term in the obo file was replaced

```
-[Term]
-id: MS:1000422
-name: high-energy collision-induced dissociation
-def: "A collision-induced dissociation process wherein the projectile ion has the translational energy higher than approximately 1000 eV." [PSI:MS]
-synonym: "HCD" EXACT []
-is_a: MS:1000044 ! dissociation method
```

is now

```
+
+[Term]
+id: MS:1000422
+name: beam-type collision-induced dissociation
+def: "A collision-induced dissociation process that occurs in a beam-type collision cell." [PSI:MS]
+synonym: "HCD" EXACT []
+is_a: MS:1000133 ! collision-induced dissociation
+[Term]
+id: MS:1002481
+name: higher energy beam-type collision-induced dissociation
+def: "A collision-induced dissociation process wherein the projectile ion has the translational energy higher than approximately 1000 eV." [PSI:MS]
+is_a: MS:1000422 ! beam-type collision-induced dissociation

```

while OpenMS still has the old name hardcoded. This leads to warnings when writing and reading the same file again

```
While loading Name of CV term not correct: 'MS:1000422 - high-energy collision-induced dissociation' should be 'beam-type collision-induced dissociation'
```

it seems that in a rather unorthodox change, the `MS:1000422` term got changed from HCD to a very general "beam-type" collision mode

@mwalzer can you comment on whether this is a good idea?

@cbielow can you comment on the impact this will have on `ItraqChannelExtractor.cpp` which seems to use the HCD activation mode CV term

PR #4918, opened 2020-09-07
Fix/activation modes

timosachsenberg:
# Description

replaces https://github.com/OpenMS/OpenMS/pull/2080
Summary:
It seems that some time ago, in 4fea083fd6a3fc820216ef3298860aad5cb5784d the HCD term in the obo file was replaced

```
-[Term]
-id: MS:1000422
-name: high-energy collision-induced dissociation
-def: "A collision-induced dissociation process wherein the projectile ion has the translational energy higher than approximately 1000 eV." [PSI:MS]
-synonym: "HCD" EXACT []
-is_a: MS:1000044 ! dissociation method
```

is now

```
+
+[Term]
+id: MS:1000422
+name: beam-type collision-induced dissociation
+def: "A collision-induced dissociation process that occurs in a beam-type collision cell." [PSI:MS]
+synonym: "HCD" EXACT []
+is_a: MS:1000133 ! collision-induced dissociation
+[Term]
+id: MS:1002481
+name: higher energy beam-type collision-induced dissociation
+def: "A collision-induced dissociation process wherein the projectile ion has the translational energy higher than approximately 1000 eV." [PSI:MS]
+is_a: MS:1000422 ! beam-type collision-induced dissociation

```

while OpenMS still has the old name hardcoded. This leads to warnings when writing and reading the same file again

```
While loading Name of CV term not correct: 'MS:1000422 - high-energy collision-induced dissociation' should be 'beam-type collision-induced dissociation'
```

it seems that in a rather unorthodox change, the `MS:1000422` term got changed from HCD to a very general "beam-type" collision mode

Apparently, the change was made to introduce a hierarchy of beam-type and a trap-type "HCD". This was due to the fact that there is -to my understanding- Thermo HCD (higher radiofrequency voltage CID to keep the fragmented ions in the c-trap) and real HCD (high energy to induce collision).
So, the term obviously could have been used for both methods. Unfortunately, the existing term was swapped for one method and another introduced, instead of renaming and repurposing as a parent and introducing two new child ones.

This PR introduces the new terms and allows both HCD and BEAM where we currently expect HCD

# Checklist:


timosachsenberg:
just updated https://github.com/OpenMS/OpenMS/pull/2080
still needs the fixes in tools that read that information in OpenMS

timosachsenberg:
rebuild jenkins

jpfeuffer:
py bindings need to be updated

timosachsenberg:
Now we annotate MS:1000422
[Term]
id: MS:1000422
name: beam-type collision-induced dissociation
def: "A collision-induced dissociation process that occurs in a beam-type collision cell." [PSI:MS]
synonym: "HCD" EXACT []
is_a: MS:1000133 ! collision-induced dissociation
as BEAM

but I think we internally often expect HCID
[Term]
id: MS:1002481
name: higher energy beam-type collision-induced dissociation
def: "A collision-induced dissociation process wherein the projectile ion has the translational energy higher than approximately 1000 eV." [PSI:MS]
is_a: MS:1000422 ! beam-type collision-induced dissociation

would it make sense to just use HCID and remove BEAM?

These are the new enums for the additional activation modes:
      TRAP,                     ///< trap-type collision-induced dissociation (MS:1002472)
      BEAM,                     ///< beam-type collision-induced dissociation (MS:1000422) "HCD"
      INSOURCE,                 ///< in-source collision-induced dissociation (MS:1001880)
      LIFT,                     ///< Bruker proprietary method (MS:1002000)

this PR now introduces

src/openms/source/METADATA/Precursor.cpp
```diff
@@ -39,8 +39,44 @@ using namespace std;
...
+    "ECD", 
+    "IMD", 
+    "SORI", 
+    "HCID", 
+    "LCID", 
+    "PHD", 
+    "ETD", 
+    "PQD",
+    "TRAP",
+    "BEAM",
```
timosachsenberg, line 76: I would prefer to use HCD as short String

src/openms/source/METADATA/Precursor.cpp
```diff
@@ -39,8 +39,44 @@ using namespace std;
...
+    "INSOURCE",
+    "LIST"
```
timosachsenberg, line 78:
```suggestion
    "LIFT"
```

timosachsenberg:
Ok I changed the OpenMS part to use ::HCD and "HCD" instead of ::BEAM and "BEAM". Much more user friendly imho.
Note: HCD is the official synonym according to the ontology.

timosachsenberg:
Tests passed - ready to merge

src/openms/include/OpenMS/KERNEL/RangeUtils.h
```diff
@@ -364,8 +362,14 @@ namespace OpenMS
         }
       }
 
-      if (reverse_) return true;
-      else return false;
+      if (reverse_)
```
cbielow, line 365: `return reverse_;` ? :)

src/openms/source/ANALYSIS/QUANTITATION/IsobaricChannelExtractor.cpp
```diff
@@ -162,9 +162,10 @@ namespace OpenMS
 
   void IsobaricChannelExtractor::setDefaultParams_()
   {
-    defaults_.setValue("select_activation", Precursor::NamesOfActivationMethod[Precursor::HCID], "Operate only on MSn scans where any of its precursors features a certain activation method (e.g., usually HCD for iTRAQ). Set to empty string if you want to disable filtering.");
+    defaults_.setValue("select_activation", "auto", "Operate only on MSn scans where any of its precursors features a certain activation method (e.g., usually HCD for iTRAQ). Set to empty string if you want to disable filtering.");
```
cbielow, line 165: can you document what `auto` does?

src/openms/source/FORMAT/HANDLERS/MzMLHandler.cpp
```diff
@@ -1894,7 +1894,15 @@ namespace OpenMS::Internal
...
-          else if (accession == "MS:1000422") //high-energy collision-induced dissociation
+          else if (accession == "MS:1000422") //beam-type collision-induced dissociation / HCD
+          {
+            spec_.getPrecursors().back().getActivationMethods().insert(Precursor::HCD);
+          }
+          else if (accession == "MS:1002472") //trap-type collision-induced dissociation
+          {
+            spec_.getPrecursors().back().getActivationMethods().insert(Precursor::TRAP);
+          }          
+          else if (accession == "MS:1002481") //high-energy collision-induced dissociation
```
cbielow, line 1905: not really the fault of this PR, but this code duplication for spectra and chromatograms seems quite unneccessary... (e.g. grab a pointer to the respective precursor once, and write the insertion-mapping only once. However, totally optional :)

cbielow (commented):
looks good. Thanks!
Minor (mostly optional) things annotated.

Issue #4921, opened 2020-09-08
Move to Cython3 or activate binding=True for pyOpenMS


jpfeuffer:
To have better introspection.

https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html

jpfeuffer:
Add here:
https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/create_cpp_extension.py#L176

Why is it still language_level = 2 again? @hroest

jpfeuffer:
I don't think building with language_level = 2 is a long-term solution. I assume we won't be able to build with newer cython versions and won't get any potential bugfixes etc.

jpfeuffer:
This was the reason #4229

jpfeuffer:
Also maybe `language_level=3str` helps with our string problems? idk

jpfeuffer:
Ok. We cannot use binding=True since autowrap cannot wrap static methods correctly.

jpfeuffer:
https://github.com/uweschmitt/autowrap/issues/35

jpfeuffer:
However, now I think it looked worse than it is because I am using PyCharm. PyCharm creates fake python sources for extension modules to create function signatures that can be used for autocompletion. It tries to create a signature based on the docstring (which is generated nicely by autowrap). While this works well for simple methods (e.g. no overloaded methods) with simple param objects (e.g. no lists), it fails for anything more complex.

I am kind of happy with the docstrings for now, I just though something was broken with them. With autowrap we probably never will need "embed_signatures" (I wonder what the difference would be).

I am still wondering what binding=True would lead to. It should be able to get real signatures with that if we make autowrap decorate static methods. Seems to build a bit longer though and I read somewhere that there might be slight performance decreases.

timosachsenberg:
I think we moved to cython 3 right?

cbielow:
> I think we moved to cython 3 right?

True: https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/requirements_bld.txt#L6

timosachsenberg:
Done

PR #4922, opened 2020-09-08
Some tries to build pyopenms with language_level=3

jpfeuffer:
see #4921


jpfeuffer:
TODO make the all_modules.py generation py-version independent. Seems to fail with py2 on travis etc.

oliveralka:
Is it enough to remove the dot from the pyopenms import in all_modules.py for language independency? 
https://github.com/OpenMS/OpenMS/blob/4fff096878bbbf52172d8aebfb5aa82dad7cfdae/src/pyOpenMS/create_cpp_extension.py#L212

jpfeuffer:
Maybe? 😄

poshul (commented):
This looks good to me. What python versions have you tested it with?

jpfeuffer:
So actually the problem was, that out "tests" did not test the actual import.
I now realized that it builds with the additional include_paths but fails during import because of the missing relative imports in autowrap I think.

hroest:
I remember we had a similar discussion with @poshul a few months ago. First we have to fix this on the autowrap side and honestly we kind of gave up because we run into issues with relative imports:

- see https://github.com/uweschmitt/autowrap/pull/86
- see https://github.com/uweschmitt/autowrap/pull/82

this also had to do with the option `options.language_level = sys.version_info.major`

jpfeuffer:
Hi @hroest regarding that:
See my PR that fixes your fix:  https://github.com/hroest/autowrap/pull/2
I am currently making it cross-platform quickly.

jpfeuffer:
The upstream PR of autowrap was merged.

jpfeuffer:
Hmm but the other pyopenms PR needs to be integrated here.

jpfeuffer:
TODO: use newest contrib: on macOS https://abibuilder.informatik.uni-tuebingen.de/archive/openms/contrib/macOS/10.15.5/x64/appleclang-11.0.0/

jpfeuffer:
Yaaay!

poshul (approved):
Looks good!

Issue #4923, opened 2020-09-09
SwathWizard Icon Same as TOPPView (macOS) - OpenMS 2.6.0 RC


oliveralka:
Trying the OpenMS 2.6.0 RC installer - it seems there are some hickups with the SwathWizard.

The macOS icon ".icns" from TOPPView is used as icon for SwathWizard.

Maybe you can have a look @cbielow.

jpfeuffer:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms_gui/source/VISUAL/APPLICATIONS/GUITOOLS/SwathWizard-resources/SwathWizard.icns

jpfeuffer:
Also the Path workaround thingy does not seem to work

cbielow:
ahh classic vendo-lock-in problem:
I would have fixed that in the beginning but I could not find a way to create an .icns file without MacOSX (I do not have access to macs), see https://stackoverflow.com/questions/6337787/how-can-i-set-the-icon-for-a-mac-application-in-xcode
`iconutil` is not available on any system I have...

So, can someone with a Mac do the conversion?

oliveralka:
Trying to run SwathWizard either form outside the Contents directory or inside I get the following error:

Try to run the .app:
![Screenshot 2020-09-09 at 11 09 29](https://user-images.githubusercontent.com/15729190/92596342-f9a3ab00-f2a5-11ea-9edc-e9fd963015a2.png)
Try to run from inside the bundle (unix executable)
![Screenshot 2020-09-09 at 11 13 47](https://user-images.githubusercontent.com/15729190/92596353-fc060500-f2a5-11ea-827b-4d377062d0e9.png)

cbielow:
the wizard expects the OpenSwathWorkflow to be next to the current executable:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms_gui/source/VISUAL/DIALOGS/SwathTabWidget.cpp#L80

jpfeuffer:
I assume executable path is the path of the app on macOS

jpfeuffer:
If I fix the path with the app, it still does not work. So executable is executable and you need to introduce an if/else for macOS.

jpfeuffer:
Workaround for paths: Adapt info.plist, add path, run lsregister:
[...]

PR #4924, opened 2020-09-09
SwathWizard icns

oliveralka:
# Description
SwathWizard icons (.icns) for macOS

#4923 (partially)


jpfeuffer:
To save on the number of installers to test, maybe you could include the Path fix here as well?

oliveralka:
@jpfeuffer What was the consensus of that? Basically an additional path if the first one is empty, which looks at the relative path on the .app location?

jpfeuffer:
Yes,
a) modify plist to have only /use/bin
b) also look in ../../execPath (double-check) for OpenswathWorkflow, at least on mac.

cbielow:
> b) also look in ../../execPath (double-check) for OpenswathWorkflow, at least on mac.

so, https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/SYSTEM/File.h#L273, i.e.
`File::findSiblingTOPPExecutable("OpenSwathWorkflow");` should do it.
@oliveralka Can you modfiy https://github.com/OpenMS/OpenMS/blob/develop/src/openms_gui/source/VISUAL/DIALOGS/SwathTabWidget.cpp#L80 accordingly and test it?

oliveralka:
I will do that and test it.

oliveralka:
The .app worked for me in the bin directory and after copying it to the parent directory.

cbielow:
Thanks @oliveralka @jpfeuffer for taking care of this!

PR #4933, opened 2020-09-10
[FIX] CMake warning about sqlite package name

jpfeuffer:
just letter cases


timosachsenberg:
http://cdash.openms.de/viewConfigure.php?buildid=51344

timosachsenberg:
restartung travis... seems like a timeout only

PR #4935, opened 2020-09-12
[Feat/aut509] - Support for Isotopic Labelling Experiments (MDV)`IsotopeLabelingMDVs`

ahmedskhalil:
# Description

Adding `IsotopeLabelingMDVs` class to support and analyze isotopic labelling experiments (i.e. MDVs : Mass Distribution Vectors).

# Member Functions

- `void isotopicCorrection(...)` and `void isotopicCorrections(...)` to perform isotopic corrections on normalized Features & FeatureMaps respectively.
- `void calculateIsotopicPurity(...)` and `void calculateIsotopicPurities(...)` to calculate isotopic purities on normalized Features & FeatureMaps respectively.
- `void calculateMDVAccuracy(...)` and `void calculateMDVAccuracies(...)` to calculate the MDV(s) accuracy on normalized Features & FeatureMaps respectively.
- `void calculateMDV(...)` and `void calculateMDVs(...)` calculate the normalized MDV(s) on normalized Features & FeatureMaps respectively.
- Added support for dismissed MetaValue `SumFormula` in `TransitionTSVFile::createPeptide_(...)` for `TargetedExperiment::Peptide`, see [#5083](https://github.com/OpenMS/OpenMS/issues/5083).

# Checklist:


src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,342 @@
...
+      calculateIsotopicPurity( *feature_it, feature_with_isotopic_purity, experiment_data, isotopic_purity_name);
+      featureMap_with_isotopic_purity.push_back(feature_with_isotopic_purity);
+    }
+  }
+
+  
+  void IsotopeLabelingMDVs::calculateMDVAccuracy(
+    Feature& normalized_feature,
+    Feature& feature_with_accuracy_info,
+    std::vector<double>& fragment_isotopomer_measured,
```
timosachsenberg, line 129: please check all methods if parameters can be passed as const references otherwise a user might expect that this is an output parameter

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,342 @@
...
+  }
+
+
+  void IsotopeLabelingMDVs::calculateMDVAccuracies(
+    FeatureMap& normalized_featureMap,
+    FeatureMap& featureMap_with_accuracy_info,
+    std::vector<double>& fragment_isotopomer_measured,
+    std::vector<double>& fragment_isotopomer_theoretical)
+  {
+    for (FeatureMap::Iterator feature_it = normalized_featureMap.begin(); feature_it != normalized_featureMap.end(); feature_it++){
```
timosachsenberg, line 169: I don't know if applicable here but if possible we prefer range loops
```
for (const Feature&  feature : normalized_featureMap)
```

timosachsenberg: please check also other places in code

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,342 @@
...
+    std::string& isotopic_purity_name)
+  {
+    featuremap_with_isotopic_purity = normalized_featuremap;
+    
+    double experiment_data_peak = 0.0;
+    if ( !experiment_data.empty() )
+    {
+      std::vector<double>::iterator max_it  = std::max_element(experiment_data.begin(), experiment_data.end());
+      uint64_t experiment_data_peak_idx     = std::distance(experiment_data.begin(), max_it);
+      experiment_data_peak                  = experiment_data[experiment_data_peak_idx];
```
timosachsenberg, line 101: please remove additional white spaces (onle one space before and after equal sign)

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,342 @@
...
+    std::vector<std::vector<T>>& correction_matrix_inversed)
+  {
+    uint16_t correction_matrix_n = correction_matrix.size() == correction_matrix[0].size() ? correction_matrix.size() : 0;
+
+    // 1- get the inverse
+    double **CM_temp, **CM_inv;
+    double temp;
+    int i, j, k;
+
+    CM_temp = (double **)malloc(correction_matrix_n * sizeof(double *));
```
timosachsenberg, line 283: please don't use C-style manual memory management.
We use eigen for linear algebra in OpenMS so I would suggest to use that (which also comes with matrix inversion etc.)

timosachsenberg (changes requested):
Thanks you very much for your contribution.
My comments are mainly concerning style and parameter passing though I think replacing your matrix with the Eigen Matrix in OpenMS might be beneficial.

timosachsenberg:
Once the interface is done we can assist with the python binding.

timosachsenberg:
Would it be possible to expand tests using the CoarseIsotopePatternGenerator in OpenMS? This would allow to generate e.g. isotope patterns of molecules like EmpiricalFormula("(13)C5(12)C100")

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: Thanks for your feedback, I have updated all parameter to be passed as const reference wherever necessary

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: done

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: thanks for the recommendation, I have removed the function to rely solely on eigen for inversion

ahmedskhalil:
Sure, I'll look into it and come back with questions if I have any

ahmedskhalil:
> Would it be possible to expand tests using the CoarseIsotopePatternGenerator in OpenMS? This would allow to generate e.g. isotope patterns of molecules like EmpiricalFormula("(13)C5(12)C100")

I have used the `EmpiricalFormula(..)` and the `CoarseIsotopePatternGenerator(..)` in particular to generate the desired Isotope Distribution in `IsotopeLabelingMDVs::calculateMDVAccuracy` and `IsotopeLabelingMDVs::calculateMDVAccuracies` section tests, would you recommend using it elsewhere in `IsotopeLabelingMDVs _test` ?

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+        for (auto it = measured_feature_subordinates.begin(); it != measured_feature_subordinates.end(); it++)
+        {
+          intensities_vec.push_back(it->getIntensity());
+        }
+        std::vector<OpenMS::Peak2D::IntensityType>::iterator max_it = std::max_element(intensities_vec.begin(), intensities_vec.end());
+        double measured_feature_max = intensities_vec[std::distance(intensities_vec.begin(), max_it)];
+      
+        for (size_t i = 0; i < normalized_feature.getSubordinates().size(); ++i)
+        {
+          if(measured_feature_max != 0.0) {
```
timosachsenberg, line 228: space after if

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+            normalized_feature.getSubordinates().at(i).setIntensity(normalized_feature.getSubordinates().at(i).getIntensity() /  measured_feature_max);
+          }
+        }
+      }
+      
+      else
+      {
```
timosachsenberg, line 235: add comment what  the else case covers

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+        double measured_feature_max = intensities_vec[std::distance(intensities_vec.begin(), max_it)];
+      
+        for (size_t i = 0; i < normalized_feature.getSubordinates().size(); ++i)
+        {
+          if(measured_feature_max != 0.0) {
+            normalized_feature.getSubordinates().at(i).setIntensity(normalized_feature.getSubordinates().at(i).getIntensity() /  measured_feature_max);
+          }
+        }
+      }
+
```
timosachsenberg, line 233: remove extra line

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+          
+        for (size_t i = 0; i < normalized_feature.getSubordinates().size(); ++i)
+        {
+          if (measured_feature_max != 0.0) {
+            normalized_feature.getSubordinates().at(i).setIntensity((OpenMS::Peak2D::IntensityType)measured_feature_subordinates.at(i).getMetaValue(feature_name) / measured_feature_max);
+          }
+        }
+      }
+    }
+
```
timosachsenberg, line 252: remove extra line

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+        {
+          feature_peak_apex_intensity_sum += it->getIntensity();
+        }
+        
+        for (auto it = measured_feature_subordinates.begin(); it != measured_feature_subordinates.end(); it++)
+        {
+          normalized_feature.setMetaValue((it - measured_feature_subordinates.begin()), (it->getIntensity() / feature_peak_apex_intensity_sum));
+        }
+      }
+
```
timosachsenberg, line 269: remove extra line

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+          feature_peak_apex_intensity_sum += it->getIntensity();
+        }
+        
+        for (auto it = measured_feature_subordinates.begin(); it != measured_feature_subordinates.end(); it++)
+        {
+          normalized_feature.setMetaValue((it - measured_feature_subordinates.begin()), (it->getIntensity() / feature_peak_apex_intensity_sum));
+        }
+      }
+      
+      else
```
timosachsenberg, line 270: what is the else case?

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
```diff
@@ -0,0 +1,204 @@
...
+#pragma once
+
+#include <OpenMS/config.h>
+
+//Kernal classes
+#include <OpenMS/DATASTRUCTURES/DefaultParamHandler.h>
+#include <OpenMS/KERNEL/FeatureMap.h>
+#include <OpenMS/KERNEL/Feature.h>
+
+//Standard library
```
timosachsenberg, line 44: check for unneded includes

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+  
+  void IsotopeLabelingMDVs::isotopicCorrection(
+    const Feature& normalized_feature,
+    Feature& corrected_feature,
+    const std::vector<std::vector<double>> correction_matrix,
+    const std::string correction_matrix_agent)
+  {
+    // MDV_corrected = correction_matrix_inversed * MDV_observed (normalized_features)
+    
+    std::vector<std::vector<double>> selected_correction_matrix;
```
timosachsenberg, line 60: add comment like: // get correction matrix for correction_matrix_agent

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+    if (!correction_matrix_agent.empty() && correction_matrix_search != correction_matrices_.end())
+    {
+      selected_correction_matrix = correction_matrix_search->second ;
+    }
+    else
+    {
+      selected_correction_matrix = correction_matrix;
+    }
+    
+    uint16_t correction_matrix_n = selected_correction_matrix.size();
```
timosachsenberg, line 72: size() is usually a size_t (64bit)

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+    std::vector<std::vector<double>> correction_matrix_inversed(correction_matrix_n, std::vector<double>(correction_matrix_n,0));
```
timosachsenberg, line 73: is this one needed or could this be done just via the eigen matrix? so no extra copy would be needed

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+    
+    // 1- correction matrix inversion
+    Eigen::MatrixXd correction_matrix_eigen(selected_correction_matrix.size(), selected_correction_matrix[0].size());
+    for (uint i = 0; i < selected_correction_matrix.size(); i++){
+      for (uint j = 0; j < selected_correction_matrix[0].size(); j++){
+        correction_matrix_eigen(i,j) = selected_correction_matrix[i][j];
```
timosachsenberg, line 79: e.g. no need to copy over stuff

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+      }
+    }
+    Eigen::MatrixXd correction_matrix_eigen_inversed = correction_matrix_eigen.inverse();
+    for (int i = 0; i < correction_matrix_eigen_inversed.rows(); i++){
+      for (int j = 0; j < correction_matrix_eigen_inversed.cols(); j++){
+        correction_matrix_inversed[i][j] = correction_matrix_eigen_inversed(i,j);
```
timosachsenberg, line 85: e.g., is this copying needed? calculations below could also be done directly on eigen matrix

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,314 @@
...
+        corrected_value += correction_matrix_inversed[i][j] * MDV_observed[j];
+      }
+      corrected_feature.getSubordinates().at(i).setIntensity(corrected_value);
+    }
+  }
+
+  void IsotopeLabelingMDVs::isotopicCorrections(
+    const FeatureMap& normalized_featureMap,
+    FeatureMap& corrected_featureMap,
+    const std::vector<std::vector<double>> correction_matrix,
```
timosachsenberg, line 110: consider passing by reference

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: why a space after if is required? in a source file you created recenlty [SimpleSearchEngineAlgorithm.cpp](https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp) this convention isn't applied.

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: no longer needed

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: unnecessary copying has been removed

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
```diff
@@ -0,0 +1,204 @@
...
+ #include <Eigen/Dense>
+
+namespace OpenMS
+{
+  /**
+    @brief IsotopeLabelingMDVs is a class to support and analyze isotopic labeling experiments
+            (i.e. MDVs : Mass Distribution Vectors, also known as Mass Isotopomer Distribution (MID))
+  */
+  class OPENMS_DLLAPI IsotopeLabelingMDVs :
+    public DefaultParamHandler
```
cbielow, line 61: why derive from DefaultParamHandler? Does not seem to be used?

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+  {
+    for (const Feature& feature : measured_featureMap)
+    {
+      Feature normalized_feature;
+      calculateMDV(feature, normalized_feature, mass_intensity_type, feature_name);
+      normalized_featureMap.push_back(normalized_feature);
+    }
+  }
+
+  const std::unordered_map<std::string, std::vector<std::vector<double>> > IsotopeLabelingMDVs::correction_matrices_ =
```
cbielow, line 314: I think you can get away with
```
const std::unordered_map<std::string, std::vector<std::vector<double>> > correction_matrices_ {
      { "TBDMS", {{0.8213, 0.1053, 0.0734, 0.0000},
                  {0.8420, 0.0963, 0.0617, 0.0000},
                  {0.8466, 0.0957, 0.0343, 0.0233},
                  {0.8484, 0.0954, 0.0337, 0.0225}}
      }
};
```

and if you make this thing static, then all above functions can be static as well (if DefaultParamHandler can be removed, which I think is possible).

To ease usability, you could replace the  "TBDMS" `string` key with an ENUM because:
 - the allowed values are immediately clear
 - accidentally passing an invalid value is not possible

i.e.
```
enum class DerivAgent { 
  TBDMS 
};
```

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  void IsotopeLabelingMDVs::calculateMDVs(
+    const FeatureMap& measured_featureMap, FeatureMap& normalized_featureMap,
+    const String& mass_intensity_type, const String& feature_name)
+  {
```
cbielow, line 305: would it make sense to `.clear()` the output, before pushing more stuff into it?

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+        OpenMS::Peak2D::IntensityType feature_peak_apex_intensity_sum = 0.0;
+        std::vector<OpenMS::Peak2D::IntensityType> intensities_vec;
+        for (auto it = measured_feature_subordinates.begin(); it != measured_feature_subordinates.end(); it++)
+        {
+          feature_peak_apex_intensity_sum += (Peak2D::IntensityType)it->getMetaValue(feature_name);
+        }
+
+        for (size_t i = 0; i < normalized_feature.getSubordinates().size(); ++i)
+        {
+          if (feature_peak_apex_intensity_sum != 0.0)
```
cbielow, line 293: 'if' can go outside the loop?

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+          }
+        }
+      }
+    }
+    else if (mass_intensity_type == "norm_sum")
+    {
+      if (feature_name == "intensity")
+      {
+        OpenMS::Peak2D::IntensityType feature_peak_apex_intensity_sum = 0.0;
+        std::vector<OpenMS::Peak2D::IntensityType> intensities_vec;
```
cbielow, line 270: variable is not used?!

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+        for (auto it = measured_feature_subordinates.begin(); it != measured_feature_subordinates.end(); it++)
+        {
+          normalized_feature.setMetaValue((it - measured_feature_subordinates.begin()), (it->getIntensity() / feature_peak_apex_intensity_sum));
+        }
+      }
+      // for every other case where feature_name isn't 'intensity', i.e. 'peak_apex_int'
+      else
+      {
+        OpenMS::Peak2D::IntensityType feature_peak_apex_intensity_sum = 0.0;
+        std::vector<OpenMS::Peak2D::IntensityType> intensities_vec;
```
cbielow, line 285: variable is not used?

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
```diff
@@ -0,0 +1,204 @@
...
+      const FeatureMap& normalized_featureMap, FeatureMap& featureMap_with_accuracy_info,
+      const std::vector<double>& fragment_isotopomer_measured, const std::vector<double>& fragment_isotopomer_theoretical);
+ 
+    /**
+      @brief This function calculates the mass distribution vector (MDV)
+      either normalized to the highest mass intensity (norm_max) or normalized
+      to the sum of all mass intensities (norm_sum)
+     
+      @param[in]   mass_intensity_type Mass intensity type (either norm_max or norm_sum).
+      @param[in]   feature_name Feature name to determine which features are needed to apply calculations on.
```
cbielow, line 172: not clear from the interface what is allowed here.

I can see why 'string' is convenient (access to metavalues), but ENUM with an internal array which maps to the string you need would be much appreciated.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
```diff
@@ -0,0 +1,204 @@
...
+    void calculateMDVAccuracies(
+      const FeatureMap& normalized_featureMap, FeatureMap& featureMap_with_accuracy_info,
+      const std::vector<double>& fragment_isotopomer_measured, const std::vector<double>& fragment_isotopomer_theoretical);
+ 
+    /**
+      @brief This function calculates the mass distribution vector (MDV)
+      either normalized to the highest mass intensity (norm_max) or normalized
+      to the sum of all mass intensities (norm_sum)
+     
+      @param[in]   mass_intensity_type Mass intensity type (either norm_max or norm_sum).
```
cbielow, line 171: this can easily be made an enum

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+    feature_with_accuracy_info = normalized_feature;
+    
+    std::vector<double> fragment_isotopomer_abs_diff;
+    for (size_t i = 0; i < fragment_isotopomer_theoretical.size(); ++i)
+    {
+      fragment_isotopomer_abs_diff.push_back(std::abs(fragment_isotopomer_theoretical[i] - fragment_isotopomer_measured[i]));
+    }
+    
+    double diff_mean = 0.0;
+    for (size_t i = 0; i < fragment_isotopomer_abs_diff.size(); ++i)
```
cbielow, line 178: hint:
`double diff_mean = std::accumulate(fragment_isotopomer_abs_diff.begin(), fragment_isotopomer_abs_diff.end(), 0);`
done

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+    
+    diff_mean = 0.0;
+    for (size_t i = 0; i < fragment_isotopomer_abs_diff_.size(); ++i)
+    {
+      diff_mean += fragment_isotopomer_abs_diff_.at(i);
+    }
+    
+    diff_mean /= fragment_isotopomer_abs_diff_.size();
+    
+    feature_with_accuracy_info = normalized_feature;
```
cbielow, line 199: this is done above already. Why do it twice?

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+    feature_with_accuracy_info.setMetaValue("average_accuracy", diff_mean);
```
cbielow, line 200: the function's docs differ a bit from what is actually annotated. Update docs?

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,327 @@
...
+  {
+    feature_with_accuracy_info = normalized_feature;
+    
+    std::vector<double> fragment_isotopomer_abs_diff;
+    for (size_t i = 0; i < fragment_isotopomer_theoretical.size(); ++i)
+    {
+      fragment_isotopomer_abs_diff.push_back(std::abs(fragment_isotopomer_theoretical[i] - fragment_isotopomer_measured[i]));
+    }
+    
+    double diff_mean = 0.0;
```
cbielow, line 177: some of these loops are not needed (since you can calculate the sum on the fly),

but you might also want to consider:
OpenMS::StatisticFunctions::MAD -- for median absolute deviate (instead of mean absolute deviation -- which should be susceptible to outliers)

cbielow: just a thought: if you really want the code to do exactly as it is, you might want to consider adding 'MeanAbsoluteDeviation' to `OpenMS::StatisticFunctions` and use it from there.

cbielow (changes requested):
nice PR! Thanks for contributing!

Mostly just some interface recommendations.

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
cbielow: `size()` is `size_t` by definition! :)

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: yes it can, thanks for the hint

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
timosachsenberg: we have Math::MAD() already

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: thanks, it was supposed to be removed

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
cbielow: > 
> 
> we have Math::MAD() already

Yep. That's what I meant with `OpenMS::StatisticFunctions::MAD`.
But we do not have MeanAbsoluteDeviation (which I'm not sure makes even sense)

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
ahmedskhalil: thanks for the suggestion, is it recommended to have `mass_intensity_type` and `feature_name` passed in as ENUM ? for the latter there could be infinite amount of possibilities for `feature_name` beside intensity, or would it be sufficient to start with an ENUM with just 2 elements `{INTENSITY, PEAK_APEX_INT}` ?

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: it has been removed, thanks for pointing that out

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: thanks for your suggestion, it has been updated to accommodate the ENUM instead of passing in a string for the derivatization agent.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
cbielow: Exactly. There would be a final Value called `SIZE_OF...` which allows to know the number of entries in the enum.
See https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/METADATA/Instrument.h#L70
for an example, e.g.
```
enum IonOpticsType
{
  UNKNOWN,                  ///< unknown
  MAGNETIC_DEFLECTION,      ///< magnetic deflection
...
  SIZE_OF_IONOPTICSTYPE
};

/// Names of inlet types
static const std::string NamesOfIonOpticsType[SIZE_OF_IONOPTICSTYPE];
```

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: `normalized_featureMap` is supposed to be just declared by the user but I will clear it to make sure `normalized_feature` are pushed back into an empty vector

cbielow: good. Never trust the input!
And if you can do the work (= `.clear()`) in one place, prefer that over doing it in many places (whenever the method is called).
And if the vector is already empty (because it was created), clearing it is a NOOP, so win-win.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
ahmedskhalil: - `CoarseIsotopePatternGenerator.h` and `IsotopeDistribution.h` are used to calculate fragment isotopomer theoretical values. 
- `OpenMS/DATASTRUCTURES/Matrix.h` is used for operations on correction matrices. 
- `Feature.h` and `FeatureMap.h` are used for FeatureMap related operations.
- `cstddef.h` is used for `size_t`.
- `cmath.h` is used for `std::abs`.
- `unordered_map` is used as storage container for correction matrices.
- `algorithm` is used for `std::max`.
- `Eigen/Dense` is used for Eigen Matrix : `MatrixXf`.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
ahmedskhalil: thanks for the suggestion, it has been updated to used the enums elements.

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: updated.

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
ahmedskhalil: > > we have Math::MAD() already
> 
> Yep. That's what I meant with `OpenMS::StatisticFunctions::MAD`.
> But we do not have MeanAbsoluteDeviation (which I'm not sure makes even sense)

I have updated StatisticFunctions.h with `MeanAbsoluteDeviation`.

ahmedskhalil:
> nice PR! Thanks for contributing!
> 
> Mostly just some interface recommendations.

Thanks, all the changes has been pushed.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
```diff
@@ -0,0 +1,238 @@
...
+
+    /// Destructor
+    ~IsotopeLabelingMDVs();
+    //@}
+    
+    enum class DerivatizationAgent
+    {
+      NOT_SELECTED,
+      TBDMS
+    };
```
timosachsenberg, line 80: add SIZE_OF_DERIVATIZATIONAGENT

ahmedskhalil: done

ahmedskhalil:
All tests passed.

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
cbielow: I think what Timo meant was: check if you need the includes in the header. Most of them can actually be moved into the cpp since they are only used there.
Others, you can forward declare, e.g. 
```
class FeatureMap;
class Feature;
```

If i'm not mistaken, you only need `Matrix.h`

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.h
cbielow: can you remove DefaultParamhandler, please.

src/openms/include/OpenMS/MATH/STATISTICS/StatisticFunctions.h
```diff
@@ -205,6 +205,36 @@ namespace OpenMS
...
+
+    */
+    template <typename IteratorType>
+    double MeanAbsoluteDeviation(IteratorType begin, IteratorType end, double mean_of_numbers)
+    {
+      std::vector<double> diffs;
+      diffs.reserve(std::distance(begin, end));
+      for (IteratorType it = begin; it != end; ++it)
+      {
+        diffs.push_back(fabs(*it - mean_of_numbers));
```
cbielow, line 234: this seems inefficient.
what about:
```
double mean {0};
for (IteratorType it = begin; it != end; ++it)
{
  mean += fabs(*it - mean_of_numbers);
}
mean /= std::distance(begin, end);
```

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,322 @@
...
+  void IsotopeLabelingMDVs::isotopicCorrection(
+    const Feature& normalized_feature,
+    Feature& corrected_feature,
+    const Matrix<double>& correction_matrix,
+    const DerivatizationAgent& correction_matrix_agent)
+  {
+    // MDV_corrected = correction_matrix_inversed * MDV_observed (normalized_features)
+    
+    /// Correction Matrices for various derivatization agents
+    const std::map<DerivatizationAgent, std::vector<std::vector<double>> > correction_matrices_ =
```
cbielow, line 64: naming convention: variable is local, so should not carry an `_` at the end.
You could make it a static member though, depending on where you need it...

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,322 @@
...
+      }
+    };
+    
+    Eigen::MatrixXd correction_matrix_eigen;
+    auto correction_matrix_search = correction_matrices_.find(correction_matrix_agent);
+    
+    // use the internally saved correction matrix if the derivatization agent name is supplied and found
+    if (correction_matrix_agent != DerivatizationAgent::NOT_SELECTED && correction_matrix_search != correction_matrices_.end())
+    {
+      correction_matrix_eigen.resize(correction_matrix_search->second.size(), correction_matrix_search->second[0].size());
```
cbielow, line 79: OpenMS already contains code which can do the conversion:
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/namespaceOpenMS.html#a51eb708c742e3091feadac71be8bb50f
unfortunately, someone decided to allocate the EigenMatrix on the heap .. for whatever reason.
So the interface is inefficient.
A better one would be to just return the matrix on the stack.
If you want, feel free to add another function to MatrixUtils.h, otherwise I can also do that, since it's not really a part of this PR.

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,322 @@
...
+          correction_matrix_eigen(i,j) = correction_matrix(i,j);
+        }
+      }
+    }
+    
+    // 1- inversion of correction matrix
+    Eigen::MatrixXd correction_matrix_eigen_inversed = correction_matrix_eigen.inverse();
+    
+    // 2- element-wise expansion with MDV_observed
+    std::vector<Feature> normalized_feature_subordinates = normalized_feature.getSubordinates();
```
cbielow, line 105: to write the same thing in less code:
```
    std::vector<double> MDV_observed;
    for (const auto& feature : normalized_feature.getSubordinates())
    {
      MDV_observed.push_back(feature.getMetaValue("peak_apex_int"));
    }
```

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,322 @@
...
+    
+    std::vector<double> fragment_isotopomer_abs_diff;
+    for (size_t i = 0; i < fragment_isotopomer_theoretical.size(); ++i)
+    {
+      fragment_isotopomer_abs_diff.push_back(std::fabs(fragment_isotopomer_theoretical[i] - fragment_isotopomer_measured[i]));
+    }
+    
+    if (!fragment_isotopomer_abs_diff.empty())
+    {
+      double diff_mean = std::accumulate(fragment_isotopomer_abs_diff.begin(), fragment_isotopomer_abs_diff.end(), 0.0) / fragment_isotopomer_abs_diff.size();
```
cbielow, line 201: we have OpenMS::Math::mean(), right? Would be more expressive here

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,322 @@
...
+    
+      diff_mean = OpenMS::Math::MeanAbsoluteDeviation(fragment_isotopomer_abs_diff.begin(), fragment_isotopomer_abs_diff.end(), diff_mean);
+    
+      feature_with_accuracy_info.setMetaValue("average_accuracy", diff_mean);
+    }
+  }
+
+  void IsotopeLabelingMDVs::calculateMDVAccuracies(
+    const FeatureMap& normalized_featureMap,
+    FeatureMap& featureMap_with_accuracy_info,
```
cbielow, line 211: clear it before adding stuff to it? You never know what is being passed into the function...

src/openms/source/ANALYSIS/QUANTITATION/IsotopeLabelingMDVs.cpp
```diff
@@ -0,0 +1,322 @@
...
+        }
+      }
+    }
+  }
+
+  void IsotopeLabelingMDVs::calculateMDVs(
+    const FeatureMap& measured_featureMap, FeatureMap& normalized_featureMap,
+    const MassIntensityType& mass_intensity_type, const FeatureName& feature_name)
+  {
+    if (!normalized_featureMap.empty())
```
cbielow, line 310: no need for the surrounding `if ...`?! Just always clear..
It's less code and faster to parse for a human

src/tests/class_tests/openms/source/StatisticFunctions_test.cpp
```diff
@@ -118,6 +118,21 @@ START_SECTION([EXTRA](template <typename IteratorType> double MAD(IteratorType b
 }
 END_SECTION
 
+START_SECTION([EXTRA](template <typename IteratorType> double MeanAD(IteratorType begin, IteratorType end, double median_of_numbers)))
+{
+  int x1[] = {-1, 0, 1, 2, 3}; // mean = 1
```
cbielow, line 123: can you test corner cases, e.g.
empty input or just a single element-vector?

src/tests/class_tests/openms/source/StatisticFunctions_test.cpp
```diff
@@ -118,6 +118,21 @@ START_SECTION([EXTRA](template <typename IteratorType> double MAD(IteratorType b
...
+  TEST_EQUAL(Math::MeanAbsoluteDeviation(x1, x1 + 5, 1), 1.2);
+  
+  int x2[] = {-1, 0, 1, 2, 3, 4}; // mean = 1.5
+  TEST_REAL_SIMILAR(Math::MeanAbsoluteDeviation(x2, x2 + 6, 1.5), 1.5);
+  
+  DoubleList z_odd = ListUtils::create<double>("-1.0,-0.5,0.0,0.5,1.0,1.5,2.0"); // mean = 0.5
```
cbielow, line 129: hint: can be written as 
```
DoubleList z_odd {-1.0, -0.5, 0.0, 0.5, 1.0, 1.5, 2.0}; // mean = 0.5
```

src/tests/class_tests/openms/source/IsotopeLabelingMDVs_test.cpp
```diff
@@ -0,0 +1,684 @@
...
+  }
+  lactate_1_normalized.setSubordinates(L1_subordinates_normmax);
+
+  for (uint8_t i = 0; i < 3; ++i)
+  {
+    lactate_1_featureMap.push_back(lactate_1_normalized);
+  }
+
+  isotopelabelingmdvs.calculateMDVAccuracies(lactate_1_featureMap, lactate_1_with_accuracy_info_featureMap, accoa_C23H37N7O17P3S_MRM_measured_13, "C23H37N7O17P3S");
+  for (uint8_t i = 0; i < lactate_1_with_accuracy_info_featureMap.size(); ++i)
```
cbielow, line 669: `uint8_t` is pretty unusual and probably slower than just simply using `size_t`. Also a bit dangerous if the container is larger than 255 entries.

cbielow (commented):
I'm just nitpicking here and there.
Feel free to ignore it, but it would improve the code quality :)

cbielow:
and a pretty substantial code addition. Thanks for contributing @ahmedskhalil

ahmedskhalil:
> and a pretty substantial code addition. Thanks for contributing @ahmedskhalil

thanks for all the feedback, it should be ready to go now :)

cbielow:
I made two minor adaptations.
Feel free to merge when tests pass.

ahmedskhalil:
> I made two minor adaptations.
> Feel free to merge when tests pass.

tests have passed.

PR #4937, opened 2020-09-14
Extend OpenMSInfo (include build type, openmp info and missing infos on linux and macos)

### LINKED ISSUE #4664: OpenMSInfo: display build configuration (realse/debug)

jpfeuffer:
good to have to check if conda and pyopenms builds are working correctly.
Fixes #4664


src/openms/configh.cmake
```diff
@@ -183,6 +183,34 @@ configure_file(${PROJECT_SOURCE_DIR}/include/OpenMS/openms_package_version.h.in
...
+		COMMAND ${CMAKE_COMMAND} "-E" "copy_if_different" "${CONFIGURED_BUILD_CONFIG_H}" "${CONFIGURED_BUILD_CONFIG_CURRENT_H}"
+		VERBATIM
+		PRE_BUILD
+		DEPENDS  "${CONFIGURED_BUILD_CONFIG_H}"
+		OUTPUT   "${CONFIGURED_BUILD_CONFIG_CURRENT_H}"
+		COMMENT  "creating build_config.h file ({event: PRE_BUILD}, {filename: build_config.h})"
+)
+
 #------------------------------------------------------------------------------
 # export a list of all configured heders
```
timosachsenberg, line 214: typo headers

timosachsenberg (approved):
really useful!

timosachsenberg (approved):

timosachsenberg:
does this also include if release / debug was built?

jpfeuffer:
Yes, will print "Release, Debug, RelWithDebInfo, RelMinSize" or whatever the CMake config was at build time.

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
```diff
@@ -0,0 +1,230 @@
...
+    std::string OpenMS_ArchNames[] = {"unknown", "32 bit", "64 bit"};
+
+    #if WIN32
+    OpenMS_Architecture getArchOnWin();
+    String getWinOSVersion();
+    #endif
+
+    class OpenMSOSInfo
+    {
+      OpenMS_OS os;
```
cbielow, line 62: should be `os_`?

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
```diff
@@ -0,0 +1,230 @@
...
+      /*
+        @brief Get the build type used during building the OpenMS library
+      */
+      static String getBuildType()
+      {
+        return OPENMS_BUILD_TYPE;
+      }
+
+      /*
+        @brief Get the build type used during building the OpenMS library
```
cbielow, line 217: get number of system cores (incl. hyperthreading)

cbielow (commented):
nice!

only some tab/space mixups I think

jpfeuffer:
Ok I will address the comments and check indents but could also be some #ifdef indents.
How do we handle them usually? I think my IDE tries to do no indent for #ifdefs.

jpfeuffer:
Ah found some indent errors, yes

jpfeuffer:
rebuild jenkins

jpfeuffer:
rebuild jenkins

jpfeuffer:
I don't know. Macro clash?

jpfeuffer:
@cbielow Do you know of any clashes? Should I make an enum class to be safe?

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
```diff
@@ -0,0 +1,225 @@
...
+        }
+        #endif
+        return info;
+      }
+
+      //********************
+      //  Windows specific API calls
+      //********************
+      #ifdef WIN32
+      #include <windows.h>
```
cbielow, line 139: including this within our namespace is probably not a good idea

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
```diff
@@ -0,0 +1,225 @@
...
+    std::string OpenMS_OSNames[] = {"unknown", "MacOS", "Windows", "Linux"};
+    enum OpenMS_Architecture {ARCH_UNKNOWN, ARCH_32BIT, ARCH_64BIT};
+    std::string OpenMS_ArchNames[] = {"unknown", "32 bit", "64 bit"};
+
+    #if WIN32
+    OpenMS_Architecture getArchOnWin();
+    String getWinOSVersion();
+    #endif
+
+    class OpenMSOSInfo
```
timosachsenberg, line 60: does this need to be exported?

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
```diff
@@ -0,0 +1,225 @@
...
+          return ARCH_64BIT;
+        }
+        else
+        {
+          return ARCH_32BIT;
+        }
+        #endif
+      }
+
+      String getWinOSVersion()
```
cbielow, line 179: maybe comment why `QSysInfo::productVersion()` is not good enough? (lacking minor/path information?)

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
```diff
@@ -0,0 +1,225 @@
...
+        #elif (defined(__MACH__) && defined(__APPLE__)) // MacOS
+        info.os_ = OS_MACOS;
+        #else //Linux
+        info.os_ = OS_LINUX;
+        #endif
+        // check if we can use QSysInfo
+        #if (defined(Q_OS_MACOS) || defined(Q_OS_UNIX))
+        info.os_version_ = QSysInfo::productVersion();
+        // identify architecture
+        if (QSysInfo::WordSize == 32)
```
cbielow, line 123: so for non-win, this is always identical to pointer-size, but for Win, this could be larger (due to wow64)?
However, we do not do 32bit builds anymore... so this will never be used, right?

cbielow: I'd say, just report the  OpenMS binary architecture by looking at sizeof(size_t), and ignore the system arch in all cases (because that should not be our concern).

cbielow:
Windows.h is a beast, so hard to tell.

But if you move it out of our namespace, that might already do the trick (or 
get rid of it, see earlier comments) J

timosachsenberg:
side note: AFAIK  enum classes can't currently be wrapped using autowrap.

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
jpfeuffer: Ah, actually I don't know why? :D Maybe it was just not present when this function was written?

jpfeuffer: TBH personally I am not sure if I ever would need more information than https://doc.qt.io/qt-5/qsysinfo.html#productVersion

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
jpfeuffer: ok done. I agree

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
jpfeuffer: Let's go with just the Qt Info.

src/openms/include/OpenMS/SYSTEM/BuildInfo.h
jpfeuffer: removed

Issue #4942, opened 2020-09-16
TOPPView: deleting the wrong layer sometimes


cbielow:
adding a new layer inserts the layer according to its type in the layer stack.
However, updating the new index pointing to that layer is not always correct (sometimes it is assumed to be the last layer, which might not be true, e.g. when adding a peak layer to a feature layer).

PR #4943, opened 2020-09-16
Toppview refactor

cbielow:
# Description

[FEATURE] caches the spectra tree view (does not repaint it), because this could make selecting spectra in 1D view extremely slow (usually when there are a lot(!) of spectra/chromatograms, as is the case for OpenSwath XIC data).

[FEATURE] displaying chromatogram data with a lot of chroms is now feasible by making a quadratic precursor mapping function linear

[FEATURE] support loading sqMass sqlLite files from TOPPView and visualize Chromatograms

[FIXES] #4942 (Layer insertion/deletion failures), by introducing a LayerStack class which encapsulates the problem and gets rid of a lot of boilerplate code

[MISC] some other fixes, e.g. getting rid of std::map in FileTypes by using static arrays and having a compile-time type description available for all file types we can read/write. Useful for file dialogs and filters.

# Checklist:
- [x] New and existing unit tests pass locally with my changes (GUI so not applicable, unfortunately)
- [x] Updated or added python bindings for changed or new classes (GUI, so not applicable)


timosachsenberg:
just testing if the rebase action now works

cbielow:
rebuild jenkins

cbielow:
ready for review

src/openms_gui/source/VISUAL/SpectraSelectionTabs.cpp
```diff
@@ -103,9 +103,9 @@ namespace OpenMS
     });
 
     SpectrumCanvas* cc = tv_->getActiveCanvas();
-    int layer_row = (cc == nullptr ? -1 : (int)cc->activeLayerIndex());
+    Size layer_row = (cc == nullptr ? -1 : cc->getCurrentLayerIndex());
 
-    if (cc == nullptr || layer_row == -1)
+    if (cc == nullptr || layer_row == (Size)-1)
```
timosachsenberg, line 108: is this test redundant?
e.g., test only for layer_row == (Size)-1 should suffice

src/openms_gui/include/OpenMS/VISUAL/SpectrumCanvas.h
```diff
@@ -62,6 +62,110 @@ namespace OpenMS
 {
   class SpectrumWidget;
 
+  class LayerStack
```
timosachsenberg, line 65: move implementation to cpp file.
maybe add a line of comment that layers are automatically sorted by layer type?

src/openms_gui/include/OpenMS/VISUAL/LayerData.h
```diff
@@ -394,6 +405,40 @@ namespace OpenMS
     */
     void updateRanges();
 
+    /// Returns the minimum intensity of the internal data, depending on type
+    inline float getMinIntensity() const
```
timosachsenberg, line 409: move to cpp and remove inline?

timosachsenberg:
nice fixes and refactoring. Otherwise looks good (already had a look at it yesterday)

src/openms_gui/source/VISUAL/SpectraSelectionTabs.cpp
cbielow: good point. looking a bit closer, layer_row is not required at all..

cbielow: looking even closer, it is...

PR #4953, opened 2020-09-21
[fix] fixed space issue in test_AcquisitionInfo.py and added it to CmakeLists

poshul:
Fixed bug where test_AcquisitionInfo wasnt being run. Fixed syntax error in file.


jpfeuffer (approved):
@timosachsenberg into release?

timosachsenberg:
Hmm not critical I think

PR #4955, opened 2020-09-22
[FIX] Small Epifany fixes

jpfeuffer:
# Description

Make assertion to an exception so we realize when we have to skip the connected component.
Allow score switching internally.


timosachsenberg:
looks good but whitespaces are a bit of.
also a comment around the ifdef would be great

PR #4957, opened 2020-09-23
[FIX] Resolve compatibility issues between IDRipper and IDMerger

lkuchenb:
# Description

Currently `IDRipper` cannot rip what `IDMerger` has merged, since the latest changes in `IDMerger` hadn't been ported to `IDRipper`. This patch fixes `IDRipper` such that it can autodetect and handle the `IDMerger` outputs that were produced in any previous release.

# Main changes

 * Input format autodetection and handling as described above
 * Removes `-out_path` option, there is now only `-out` which is used to specify the output directory where the splitted files shall be written to
 * Adds a `-split_id_runs` flag which changes the behavior such that `IdentificationRun`s are also splitted into individual files
 * Adds a `-numeric_filenames` flag which enables output file names equal to the input file name with a suffix `_{idrun_index}_{origin_index}` (or just `_{origin_index}` if `-split_id_runs` is disabled). This is useful if otherwise there would be output filename conflicts.

# Checklist:
- [X] Updated or added python bindings for changed or new classes. (Tick if no upates were necessary.)


lkuchenb:
@jpfeuffer `IDRipper` tests 2+3 assume that `IDRipper` splits IDs from different `IdentificationRun`s into the same file if `file_origin` is specified equally. Is this actually a use case or was this just tested like this because of the previous behavior of `IDMerger`?

Following our discussion on Discord, this patch would make `IDRipper` split by `IdentificationRun` *and* origin.

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+  mode = new_mode;
+  return true;
+}
+
+IDRipper::OriginAnnotationFormat IDRipper::detectOriginAnnotationFormat_(map<String, UInt> & file_origin_map, const std::vector<PeptideIdentification> & peptide_idents)
+  {
+    // In case we observe 'file_origin' meta values, we assign an index to every unique meta value
+    file_origin_map.clear();
+
+    vector<string> modes = { "file_origin", "map_index", "id_merge_index" };
```
cbielow, line 225: add `const`

src/openms/include/OpenMS/ANALYSIS/ID/IDRipper.h
```diff
@@ -70,10 +70,14 @@ namespace OpenMS
       @param proteins Peptide identification annotated with file origin
       @param peptides Protein identification
     */
-    void rip(std::map<String, std::pair<std::vector<ProteinIdentification>, std::vector<PeptideIdentification> > > & ripped, std::vector<ProteinIdentification> & proteins, std::vector<PeptideIdentification> & peptides);
+    void rip(std::map<std::pair<UInt, UInt>, std::pair<std::vector<ProteinIdentification>, std::vector<PeptideIdentification> > > & ripped, std::vector<ProteinIdentification> & proteins, std::vector<PeptideIdentification> & peptides);
```
cbielow, line 73: this type looks very ugly.
Can you refactor that into a struct? or a class with members?

src/openms/include/OpenMS/ANALYSIS/ID/IDRipper.h
```diff
@@ -70,10 +70,14 @@ namespace OpenMS
...
 
 private:
 
+    enum OriginAnnotationFormat { FILE_ORIGIN = 0, MAP_INDEX = 1, ID_MERGE_INDEX = 2, UNKNOWN = 3 };
```
cbielow, line 77: we usually go with `SIZE_OF_ORIGIN_ANNOTATION_FORMAT`, as last member so you can do
`std::array<std::string, SIZE_OF_ORIGIN_ANNOTATION_FORMAT> names_of_OriginAnnotationFormat = { ...};`

this makes iteration over the string representations easy... see below

src/openms/include/OpenMS/ANALYSIS/ID/IDRipper.h
```diff
@@ -82,6 +86,8 @@ namespace OpenMS
     /// Assignment
     IDRipper & operator=(const IDRipper & rhs);
 
+    /// helper function, detects file origin annotation standard from collections of protein and peptide hits
+    OriginAnnotationFormat detectOriginAnnotationFormat_(std::map<String, UInt> & file_origin_map, const std::vector<PeptideIdentification> & peptide_idents);
```
cbielow, line 90: coding style.
`Type& name`

we also go with west-const usually

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -117,8 +144,8 @@ namespace OpenMS
       getProteinIdentification_(prot_ident, *pep_it, proteins);
       // TODO catch case that ProteinIdentification prot_ident is not found in the for-loop
 
-
-      map<String, pair<vector<ProteinIdentification>, vector<PeptideIdentification> > >::iterator it = ripped.find(file_);
+      pair<UInt,UInt> idx_pair(ident_run_idx, file_origin_idx);
+      map<pair<UInt, UInt>, pair<vector<ProteinIdentification>, vector<PeptideIdentification> > >::iterator it = ripped.find(idx_pair);
```
cbielow, line 148: very hard to understand this type...

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
         peptide_idents.push_back(*pep_it);
 
         //create and insert new map entry
-        ripped.insert(make_pair(file_, make_pair(protein_idents, peptide_idents)));
+        ripped.insert(make_pair(idx_pair, make_pair(protein_idents, peptide_idents)));
```
cbielow, line 190: inefficient. but maybe ok, if this is not a performance bottleneck.
better alternative:
`.emplace()` with `piecewise_construct` and std::move

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+      }
+    }
+    // Reduce the spectra data string list if that's what we ripped by
+    if (origin_annotation_fmt == MAP_INDEX || origin_annotation_fmt == ID_MERGE_INDEX)
+    {
+      map<pair<UInt, UInt>, pair<vector<ProteinIdentification>, vector<PeptideIdentification> > >::iterator it;
```
cbielow, line 196: range_based for instead?

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+      for (it = ripped.begin(); it != ripped.end(); ++it)
+      {
+        UInt const spectra_data_idx = it->first.second;
+        ProteinIdentification & prot_id = it->second.first.at(0);
+        StringList old_list;
+        prot_id.getPrimaryMSRunPath(old_list);
+        StringList new_list;
+        new_list.push_back(old_list.at(spectra_data_idx));
+        prot_id.setPrimaryMSRunPath(new_list);
```
cbielow, line 205: `prot_id.setPrimaryMSRunPath( { old_list.at(spectra_data_idx) } );`

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+      }
+    }
+  }
+
+bool setOriginAnnotationMode_(short & mode, short const new_mode)
```
cbielow, line 210: some docs would be nice :)

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+IDRipper::OriginAnnotationFormat IDRipper::detectOriginAnnotationFormat_(map<String, UInt> & file_origin_map, const std::vector<PeptideIdentification> & peptide_idents)
+  {
+    // In case we observe 'file_origin' meta values, we assign an index to every unique meta value
+    file_origin_map.clear();
+
+    vector<string> modes = { "file_origin", "map_index", "id_merge_index" };
+    short mode = -1;
+    for (vector<PeptideIdentification>::const_iterator it = peptide_idents.begin(); it != peptide_idents.end(); ++it)
+    {
+      bool mode_identified = false;
```
cbielow, line 229: not sure this bool is required unless you need to make sure that every instance in `peptide_idents` has a mode.
Please document.

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+      for (size_t i=0; i<modes.size(); ++i)
+      {
+        if (it->metaValueExists(modes[i]))
+        {
+          // Different mode identified for same or different peptide
+          if (mode_identified || !setOriginAnnotationMode_(mode, i))
```
cbielow, line 235: can be shortened to `if (!setOriginAnnotationMode_(mode, i))` to make it less complex?

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+  return true;
+}
+
+IDRipper::OriginAnnotationFormat IDRipper::detectOriginAnnotationFormat_(map<String, UInt> & file_origin_map, const std::vector<PeptideIdentification> & peptide_idents)
+  {
+    // In case we observe 'file_origin' meta values, we assign an index to every unique meta value
+    file_origin_map.clear();
+
+    vector<string> modes = { "file_origin", "map_index", "id_merge_index" };
+    short mode = -1;
```
cbielow, line 226: why not use `OriginAnnotationFormat mode = UNKNOWN`?
Would safe some `if ...`s below

lkuchenb:
@cbielow Thanks for the comprehensive review and in particular the OpenMS code style hints. I'll incorporate that.

src/openms/include/OpenMS/ANALYSIS/ID/IDRipper.h
timosachsenberg: could make pyOpenMS changes (see pxd file) necessary

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -160,8 +187,84 @@ namespace OpenMS
...
+    for (vector<PeptideIdentification>::const_iterator it = peptide_idents.begin(); it != peptide_idents.end(); ++it)
+    {
+      bool mode_identified = false;
+      for (size_t i=0; i<modes.size(); ++i)
```
timosachsenberg, line 230: we add spaces around operators (like here = and < )

src/openms/source/ANALYSIS/ID/IDRipper.cpp
lkuchenb: This happens once per output file, which are usually in the order of < 5.

lkuchenb:
I incorporated the code style remarks by @cbielow and @timosachsenberg.

Furthermore after working with a few more files I refactored the patch quite a bit (updated in [first comment](https://github.com/OpenMS/OpenMS/pull/4957#issue-491908255)).

jpfeuffer:
```
pyopenms/pyopenms_1.cpp:167147:101: error: no matching function for call to 'OpenMS::IDRipper::rip(std::map<OpenMS::String, std::pair<std::vector<OpenMS::ProteinIdentification>, std::vector<OpenMS::PeptideIdentification> > >&, std::vector<OpenMS::ProteinIdentification>&, std::vector<OpenMS::PeptideIdentification>&)'
   __pyx_v_self->inst.get()->rip(__pyx_v_c_ripped, (*__pyx_v_c_protein_vec), (*__pyx_v_c_peptide_vec));
                                                                                                     ^
In file included from pyopenms/pyopenms_1.cpp:767:0:
/home/jenkins/ws/openms/PR/pyOMSDckr/OpenMS/src/openms/include/OpenMS/ANALYSIS/ID/IDRipper.h:127:10: note: candidate: void OpenMS::IDRipper::rip(OpenMS::IDRipper::RipFileMap&, std::vector<OpenMS::ProteinIdentification>&, std::vector<OpenMS::PeptideIdentification>&, bool, bool)
     void rip(RipFileMap& ripped, std::vector<ProteinIdentification>& proteins, std::vector<PeptideIdentification>& peptides, bool full_split, bool split_ident_runs);
          ^
/home/jenkins/ws/openms/PR/pyOMSDckr/OpenMS/src/openms/include/OpenMS/ANALYSIS/ID/IDRipper.h:127:10: note:   candidate expects 5 arguments, 3 provided
```
in the Jenkins console log.
Refers to the python bindings in: https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/pxds/IDRipper.pxd#L29

lkuchenb:
Since `std::map` seems to be a tricky candidate for autowrap, there is now a separate interface for pyOpenMS

        void rip(
                libcpp_vector[RipFileIdentifier] & rfis,
                libcpp_vector[RipFileContent] & rfcs,
                libcpp_vector[ProteinIdentification] & proteins,
                libcpp_vector[PeptideIdentification] & peptides,
                bool full_split,
                bool split_ident_runs
                )

that returns two vectors (keys and values) instead, which can be `zip`ed in Python.

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -70,34 +73,154 @@ namespace OpenMS
     return *this;
   }
 
-  void IDRipper::rip(map<String, pair<vector<ProteinIdentification>, vector<PeptideIdentification> > >& ripped, vector<ProteinIdentification>& proteins, vector<PeptideIdentification>& peptides)
+  IDRipper::IdentificationRuns::IdentificationRuns(const vector<ProteinIdentification>& prot_ids)
+  {
+    for (const auto& prot_id : prot_ids)
```
timosachsenberg, line 78: please add a higher-level comment on what is happening in the loop

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -70,34 +73,154 @@ namespace OpenMS
...
+        throw Exception::InvalidValue(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "IdentificationRun IDs are not unique!", id_run_id);
+      }
+      UInt idx = this->index_map.size();
+      this->index_map[id_run_id] = idx;
+      const DataValue& mv_spectra_data = prot_id.getMetaValue("spectra_data");
+      spectra_data.push_back(mv_spectra_data.isEmpty() ? StringList() : mv_spectra_data.toStringList());
+      }
+  }
+
+  bool IDRipper::RipFileIdentifierIdxComparator::operator()(const RipFileIdentifier& left, const RipFileIdentifier& right) const
```
timosachsenberg, line 92: please document how the comparison operator compares in the header file

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -70,34 +73,154 @@ namespace OpenMS
...
+      catch (const std::out_of_range& e)
+      {
+          throw Exception::ParseError(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "input file",
+                  "Failed to identify corresponding spectra_data element for PeptideIdentification element.");
+      }
+  }
+
+  UInt IDRipper::RipFileIdentifier::getIdentRunIdx()
+  {
+      return this->ident_run_idx;
```
timosachsenberg, line 132: more a matter of style but I usually don't write this-> in such a context.
only if it could be really ambigious

src/openms/source/ANALYSIS/ID/IDRipper.cpp
```diff
@@ -70,34 +73,154 @@ namespace OpenMS
...
+  }
+
+  const std::vector<PeptideIdentification> & IDRipper::RipFileContent::getPeptideIdentifications()
+  {
+      return this->pep_idents;
+  }
+
+  bool IDRipper::registerBasename_(map<String, pair<UInt, UInt> >& basename_to_numeric, const IDRipper::RipFileIdentifier& rfi)
+  {
+      auto it = basename_to_numeric.find(rfi.out_basename);
```
timosachsenberg, line 162: maybe add some higher level comment what is done here

src/topp/IDRipper.cpp
```diff
@@ -114,9 +114,9 @@ class TOPPIDRipper :
   {
     registerInputFile_("in", "<file>", "", "Input file, in which the protein/peptide identifications must be tagged with 'file_origin'");
     setValidFormats_("in", ListUtils::create<String>("idXML"));
-    registerOutputFile_("out", "<file>", "", "The path to this file is used as the output directory.", false, false);
-    setValidFormats_("out", ListUtils::create<String>("idXML"));
-    registerStringOption_("out_path", "<file>", "", "Directory for the output files after ripping according to 'file_origin'. If 'out_path' is set, 'out' is ignored.", false, false);
+    registerOutputFile_("out", "<file>", "", "Path to the output directory to write the ripped files to.", false, false);
```
timosachsenberg, line 117: @jpfeuffer will removing the valid format get us into trouble? @jpfeuffer

jpfeuffer: Yes. Why was it removed? Because you can do folders? We probably need output prefix support soon.

jpfeuffer: Cause if that is the case, I don't think it can be avoided.

timosachsenberg: ok so do you think this can go before prefix support?

lkuchenb: With "output prefix" you mean all output files start with the same prefix? That would still not work with the default mode of ripper, since that extracts the original filenames from the merged idXML file.

timosachsenberg: if we omit the file type... what will happen on KNIME side if there is no mime type associated anymore?

jpfeuffer: It will generate an output file name without extension and pass it to the tool.
The next tool however will a) usually not accept the empty mimetype and b) not find any file, since the generated files are called different than the output name Knime generated. So you need another FileInput Node that loads the generated files based on the output prefix. Or we just correctly declare it a prefix in CTD.
A prefix could be a subfolder I think (i.e. they might not need to share anything hung in the filename).

timosachsenberg:
Ok I think the impact is rather small at this point and would say we merge after @lkuchenb added some of the remaining comment lines.

timosachsenberg:
rebuild jenkins

timosachsenberg:
rebuild jenkins

timosachsenberg:
rebuild jenkins

timosachsenberg:
tests passed. I think this is ready to merge

timosachsenberg:
thanks. maybe fixed by https://github.com/OpenMS/OpenMS/pull/5738

PR #4952, opened 2020-09-21
AGM/SiriusAdapter: FIX - unidentified masses & scan index not readable

oliveralka:
# Description
**Issue when running AssayGeneratorMetabo with HMDB, LipidMaps DB (multiple possible identifications).**
Multiple hits for the same feature stored as a compound in the .ms file. 
`>compound _4024150408384032977-148005Phenylpyruvicacid,m-Coumaricacid,4-Hydroxycinnamicacid,2-Hydroxycinnamicacid,Enol-phenylpyruvate,cis-p-Coumaricacid,Coumaricacid,MethylPhenylglyoxalate,Cis-2-coumarate,Trans-3-coumarate,3-(3-hydroxyphenyl)prop-2-enoicacid,3-phenyloxirane-2-carboxylicacid,3-      (4-hydroxyphenyl)prop-2-enoicacid,4-methoxy-1-benzofuran-6-ol,3-(2-hydroxyphenyl)prop-2-enoicacid,3-(2,3-dihydroxyphenyl)prop-2-enal,3-(2,5-dihydroxyphenyl)prop-2-enal,3-(2-hydroxyphenyl)oxirane-2-carbaldehyde,(2E)-3-(3,4-dihydroxyphenyl)prop-2-enal,3-(4-hydroxyphenyl)oxirane-2-carbaldehyde,9-hydroxy-7E-Nonene-3,5-diynoicacid,9-hydroxy-7Z-Nonene-3,5-diynoicacid779`

SIRIUS workspace truncates the name in the path, which leads to the loss of the index (779).
`1_20200921_144720_Olivers-MBP.fritz.box_58922_2_4024150408384032977-148005Phenylpyruvica`

Changed how / where the index is written in the path/compound. 
`1_20200921_144720_Olivers-MBP.fritz.box_58922_2_4024150408384032977-148005--779--Phenylpyruvica`

**Issue when using "keep unidentified masses" in AMS.** 
Usually, if no ID is performed Peptideidentification is empty. 
In the case of storing unidentified masses everything is set to either [null] or null, which was not handled in the SiriusMSConverter. 

```
			<PeptideIdentification identification_run_ref="PI_0" score_type="" higher_score_better="true" significance_threshold="0" >
				<PeptideHit score="0" sequence="" charge="0">
					<UserParam type="stringList" name="identifier" value="[null]"/>
					<UserParam type="stringList" name="description" value="[null]"/>
					<UserParam type="string" name="modifications" value="null"/>
[...]

PR #4976, opened 2020-09-29
FIX: CSI:FingerID parsing

oliveralka:
# Description

Unfortunately, the path parsing was only added to the SiriusMzTabWriter in the previous PR #4952. It was not noticed,
since the CSI:FingerId test is switched off by default, due to its long runtime and dependence on the cloud in Jena.

That this does not happen again I did a mini-refactoring to have the regex commands in one place. 

Also please:

# Checklist:
- [X] Updated or added python bindings for changed or new classes. (Tick if no upates were necessary.)


src/openms/include/OpenMS/FORMAT/DATAACCESS/SiriusMzTabWriter.h
```diff
@@ -102,7 +102,17 @@ namespace OpenMS
...
+
+    /**
+    @brief Extract scan_number from filepath
+    */
+    static int extract_scan_number(const String& path);
+
+    /**
+    @brief Extract feature_id from filepath
+    */
+    static String extract_feautre_id(const String& path);
```
timosachsenberg, line 114: extractFeatureId (without underscore and without the typo) also for the other methods

timosachsenberg (commented):
see comments

src/openms/include/OpenMS/FORMAT/DATAACCESS/SiriusMzTabWriter.h
oliveralka: hm, code convention hiccup...

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -73,6 +73,14 @@ namespace OpenMS
 
       };
 
+      /**
+      @brief Comparison function using the extracted scan_index from the sirius workspace file path
+
```
cbielow, line 78: so `i` and `j` are scan_indices or file paths?
maybe make that more clear and use variable names which reflect that
`static bool extractAndCompareScanIndexLess(const String& filepath_1, const String& filepath_2);`

src/utils/AssayGeneratorMetabo.cpp
```diff
@@ -418,7 +413,7 @@ class TOPPAssayGeneratorMetabo :
                                                              sirius_algo);
   
         // sort vector path list
-        std::sort(subdirs.begin(), subdirs.end(), extractAndCompareScanIndexLess_);
+        std::sort(subdirs.begin(), subdirs.end(), SiriusAdapterAlgorithm::extractAndCompareScanIndexLess);
```
cbielow, line 416: how many subdirs are we talking about? because for every comparison, the string will be parsed twice...

oliveralka: Depending on the files to be analysed 100-100.000. 
Hm, what would be a better solution here?

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
oliveralka: Ok, thanks.

`i` and `j` are the scan indices, which are extracted from the file paths in the sirius workspace.

`ws_known/sirius_out/2026_20200930_173624_abinode00.informatik.uni-tuebing_161994086002579952-880005--9645--LysoPC1819Z` --> 9645

`ws_known/sirius_out/2026_20200930_175021_abinode00.informatik.uni-tuebing_5289776317112608729-1041006--11475--Pentacosano` --> 11475

src/utils/AssayGeneratorMetabo.cpp
oliveralka: @cbielow  Would it be faster to have a struct e.g. 
```
      struct OPENMS_DLLAPI SiriusWorkspaceIndex
      {
      public:
        String file_path;
        int scan_index;
      };
```
perform the extraction of the scan index once and save them in a `vector<SiriusWorkspaceIndex> siriusworkspaceindex`- and compare the scan indices within the second vector,
```
std::sort(siriusworkspaceindex.begin(), siriusworkspaceindex.end(), 
               [](const auto& i, const auto& j) { return i.scan_index < j.scan_index; } );
```
And separate them later on - or is that too much overhead?

cbielow: yes. to make it a bit more efficient, its enough to store the array_index + scan index in the struct, then sort by scan index and rebuild the subdir vector from the array_indices. But your method also works. Stringviews would also be an option, but probably a bit slower than indices.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -183,7 +184,31 @@ namespace OpenMS
     {
       return tmp_ms_file_;
     }
-    
+
+    std::vector<String> SiriusAdapterAlgorithm::sortSiriusWorkspacePathsByScanIndex(const std::vector<String>& siriusworkspacepaths)
```
oliveralka, line 188: @cbielow I think that should go in the right direction - do you have further suggestions?

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -49,6 +49,13 @@ namespace OpenMS
       /// default constructor
       SiriusAdapterAlgorithm();
 
+      struct OPENMS_DLLAPI SiriusWorkspaceIndex
```
cbielow, line 52: technically, you can hide this implementation detail by just defining it in the cpp.
No need to do it here.

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -49,6 +49,13 @@ namespace OpenMS
...
+      {
+      public:
+        int array_index;
```
cbielow, line 55: and if you give it a ctor: `SiriusWorkspaceIndex(int, int)` ...

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -183,7 +184,31 @@ namespace OpenMS
...
+    {
+      std::vector<String> sorted_siriusworkspacepaths;
+      std::vector<SiriusAdapterAlgorithm::SiriusWorkspaceIndex> indices;
+
+      for (size_t i = 0; i < siriusworkspacepaths.size(); i++)
+      {
+        SiriusAdapterAlgorithm::SiriusWorkspaceIndex wsi;
+        wsi.array_index = i;
+        wsi.scan_index = SiriusMzTabWriter::extractScanIndex(siriusworkspacepaths[i]);
+        indices.emplace_back(wsi);
```
cbielow, line 198: you can do `indices.emplace_back(i, extractScanIndex(...));` and save some lines of code

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
```diff
@@ -49,6 +49,13 @@ namespace OpenMS
...
+        int scan_index;
+      };
+
       /// Struct for temporary folder structure
       struct OPENMS_DLLAPI SiriusTemporaryFileSystemObjects
```
timosachsenberg, line 60: we would use class instead of struct

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
timosachsenberg, line 52: make it class or remove redundant public:

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -183,7 +184,31 @@ namespace OpenMS
...
+      }
+
+      std::sort(indices.begin(), indices.end(), [](const SiriusAdapterAlgorithm::SiriusWorkspaceIndex& i, const SiriusAdapterAlgorithm::SiriusWorkspaceIndex& j) { return i.scan_index < j.scan_index; } );
+
+      std::vector<String> sorted_subdirs;
```
cbielow, line 203: not used, but it has the better name?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
timosachsenberg, line 198: if you add a constructor as chris suggested you could use in-place construction with: indices.emplace_back({i,SiriusMzTabWriter::extractScanIndex(siriusworkspacepaths[i])})

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
```diff
@@ -43,10 +43,33 @@
...
+  boost::regex regexp_feature("_(?<SCAN>\\d+)-");
+  bool found = boost::regex_search(path, match, regexp_feature);
+  if (found && match["SCAN"].matched) 
+  {
+    feature_id = "id_" + match["SCAN"].str();
+  }
+  if (feature_id.empty() || feature_id == "id_0")
+  {
+    feature_id = "null";
+  }
```
timosachsenberg, line 71: are all cases covered?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -183,7 +184,31 @@ namespace OpenMS
...
+
+      for (size_t i = 0; i < siriusworkspacepaths.size(); i++)
+      {
+        SiriusAdapterAlgorithm::SiriusWorkspaceIndex wsi;
+        wsi.array_index = i;
+        wsi.scan_index = SiriusMzTabWriter::extractScanIndex(siriusworkspacepaths[i]);
+        indices.emplace_back(wsi);
+      }
+
+      std::sort(indices.begin(), indices.end(), [](const SiriusAdapterAlgorithm::SiriusWorkspaceIndex& i, const SiriusAdapterAlgorithm::SiriusWorkspaceIndex& j) { return i.scan_index < j.scan_index; } );
```
timosachsenberg, line 201: break into multiple lines?

timosachsenberg (changes requested):

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -183,7 +184,31 @@ namespace OpenMS
...
+
+      std::vector<String> sorted_subdirs;
+      for (const auto& index : indices)
+      {
+        sorted_siriusworkspacepaths.emplace_back(siriusworkspacepaths[index.array_index]);
```
cbielow, line 206: emplace_back is misleading, because this is a copy, not a move.
It cannot be a move since `siriusworkspacepaths` is const&. If you pass it in a rvalue-ref (`&&`) you can move from it. Will be faster, but its probably not that important.

cbielow (commented):
only minor feedback from my side.
Its good to go :)

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
oliveralka: I think so - in any other case there should be a feature id

oliveralka:
Thanks you all for the feedback! 
I will incorporate the changes and check the pyopenms test!

src/openms/include/OpenMS/ANALYSIS/ID/SiriusAdapterAlgorithm.h
oliveralka: That is just preference and has no deeper meaning, or?

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
oliveralka: Just to be sure, since the initial `subdirs` vector is not needed anymore after calling the sort function to generate the `sorted_subdirs`, the initial `subdirs` vector can be used as a rvalue to perform move instead of copy.

Declaration: 
`std::vector<String> SiriusAdapterAlgorithm::sortSiriusWorkspacePathsByScanIndex(std::vector<String>&& subdirs);`

Call in e.g. SiriusAdapter.cpp
`vector<String> sorted_subdirs = SiriusAdapterAlgorithm::sortSiriusWorkspacePathsByScanIndex(std::move(subdirs));`

To call the function with move schematics, `std:move()` has to be used to allow the lvalue "subdirs" to be used as a rvalue for the move. 

Or is there more to it?

oliveralka: @cbielow Unfortunately it seems that the rvalue reference can not be wrapped in python via autowrap.

oliveralka:
Retry DeepCode

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
oliveralka: https://github.com/uweschmitt/autowrap/issues/95

oliveralka:
Do not merge yet - move schematics should not be used in this fashion.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -207,12 +207,12 @@ namespace OpenMS
                 indices.end(),
                 [](const SiriusWorkspaceIndex& i, const SiriusWorkspaceIndex& j) { return i.scan_index < j.scan_index; } );
 
-      for (const auto& index : indices)
+      for (auto index : indices)
```
timosachsenberg, line 210: can be const auto& I think

oliveralka: Yes, since the `subdirs` String is moved and not the `index`.

timosachsenberg (approved):

PR #4981, opened 2020-09-30
Osw datamodel

cbielow:
# Description

 * adds a basic OSW data model for reading .osw sql files from OpenSwathWorkflow and/or pyProphet
 * streamlines some OSW code and adds more checks


src/openms/include/OpenMS/CONCEPT/Exception.h
```diff
@@ -584,6 +584,20 @@ namespace OpenMS
...
+
+      E.g. when retrieving data from a table using the wrong column name or index.
+
+      @ingroup Exceptions
+    */
+    class OPENMS_DLLAPI SqlOperationFailed :
+      public BaseException
+    {
+public:
+      SqlOperationFailed(const char* file, int line, const char* function, const std::string& desciption)  noexcept;
```
timosachsenberg, line 598: description

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
```diff
@@ -0,0 +1,344 @@
...
+          return type_;
+        }
+        /// is this a decoy transition (from a decoy protein/peptide)
+        bool isDecoy() const
+        {
+          return is_decoy_;
+        }
+
+      private:
+        String annotation_; ///< e.g. y5/-0.002
```
timosachsenberg, line 86: alignment order of relevance?

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
```diff
@@ -0,0 +1,344 @@
...
+        UInt32 id_;         ///< ID as used in OSWPeakGroup::transition_ids
+        float product_mz_;  ///< observed product m/z value
+        char type_;         ///< b, y,
```
timosachsenberg, line 89: is char enough? would string hurt?

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
```diff
@@ -0,0 +1,344 @@
...
+        }
+        /// candidate explanations
+        const std::vector<OSWPeakGroup>& getFeatures() const
+        {
+          return features_;
+        }
+
+      private:
+        String seq_;
+        short charge_{0};
```
timosachsenberg, line 213: change order for better alignment?

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
```diff
@@ -0,0 +1,344 @@
...
+          return id_;
+        }
+
+        const std::vector<OSWPeptidePrecursor>& getPeptidePrecursors() const
+        {
+          return peptides_;
+        }
+
+      private:
+        String accession_;
```
timosachsenberg, line 255: alignment?

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
```diff
@@ -0,0 +1,344 @@
...
+        OSWProtein(OSWProtein&& rhs) = default;
+        /// move assignment operator
+        OSWProtein& operator=(OSWProtein&& rhs) = default;
+
+        const String& getAccession() const
+        {
+          return accession_;
+        }
+
+        Size getID() const
```
timosachsenberg, line 244: type of ID correct... saw a lot of UInt32

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
cbielow: ordered such that padding is minimized. Because there are a lot of these...

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
cbielow: Char[1] is used in the SQL schema. So its enough. String hurts memory usage, and so I went with char

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
```diff
@@ -0,0 +1,344 @@
...
+        {
+          transitions_.emplace(tr.getID(), tr);
+        }
+
+        /// Adds a protein, which has all its subcomponents already populated
+        /// All transition references internally are checked to make sure
+        /// they are valid.
+        /// You can add stub proteins, by omitting their peptide references.
+        /// @throws Exception::Precondition() if transition IDs within protein are unknown
+        void addProtein(OSWProtein&& prot);
```
timosachsenberg, line 283: do we need a const & version? What happens again if I just pass an lvalue?

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
cbielow: here order does not influence alignment.
the `short` is always padded

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
```diff
@@ -0,0 +1,344 @@
...
+          return proteins_;
+        }
+
+        /// Replace existing protein at position @index
+        /// Note: this is NOT the protein ID, but the index into the internal protein vector. See getProteins()
+        /// 
+        /// @p index must be a valid index into the getProteins() vector
+        /// @p protein The protein to replace the existing one
+        /// @throws Exception::Precondition() if transition IDs within protein are unknown
+        void setProtein(const Size index, OSWProtein&& protein)
```
timosachsenberg, line 298: same here

src/openms/include/OpenMS/FORMAT/OSWFile.h
```diff
@@ -61,8 +64,36 @@ namespace OpenMS
...
+    /// Use in conjunction with on-demand readProtein() to query proteins as needed
+    void readMinimal(OSWData& swath_result);
+
+    /**
+      @brief populates a protein at index @p index with Peptides, unless the protein already has peptides
+
+      Internally uses the proteins ID to search for cross referencing peptides and transitions in the OSW file.
+      @throws Exception::InvalidValue if the ID is not present in the OSW file
+    */
+    void readProtein(OSWData& swath_result, const Size index);
```
timosachsenberg, line 92: is this efficient to individually query them?

src/openms/include/OpenMS/FORMAT/OSWFile.h
```diff
@@ -88,14 +119,34 @@ namespace OpenMS
       @brief Reads an OSW SQLite file and returns the data on MS1-, MS2- or transition-level
              as ostream (e.g. stringstream or ofstream).
     */
-    static void readToPIN(const std::string& in_osw, const OSWFile::OSWLevel osw_level, std::ostream& pin_output,
+    static void readToPIN(const std::string& filename, const OSWFile::OSWLevel osw_level, std::ostream& pin_output,
```
timosachsenberg, line 122: is this the percolator pin file format?

src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
```diff
@@ -149,10 +149,21 @@ namespace OpenMS
...
+    /** Add a transition
+     *
+     * Transitions are internally mapped using their nativeID,
+     * i.e. TransitionType::getNativeID.
+     *
+     * When querying for a transition, make sure to use this key.
+     */
+    inline void addTransition(const TransitionType& transition, const String& key)
     {
+      auto result = transition_map_.emplace(key, int(transitions_.size()));
```
timosachsenberg, line 161: add comment what this line does

src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
```diff
@@ -181,12 +192,21 @@ namespace OpenMS
...
+    /** Add a chromatogram 
+     *
+     * Chromatograms are internally mapped using the provided key.
+     * The ChromatogramType::getNativeID is a good choice.
+     *
+     * When querying for a chromatogram, make sure to use this key.
+     */
+    inline void addChromatogram(const ChromatogramType& chromatogram, const String& key)
     {
+      auto result = chromatogram_map_.emplace(key, int(chromatograms_.size()));
```
timosachsenberg, line 204: same here

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
cbielow: total size is currently 40 and cannot be made smaller.
The order is somewhat hierachical and similar to the db schema.

Shall i change it anyway?

src/openms/include/OpenMS/KERNEL/MRMTransitionGroup.h
```diff
@@ -356,7 +377,7 @@ namespace OpenMS
       for (typename std::vector<ChromatogramType>::const_iterator pr_it = precursor_chromatograms_.begin(); pr_it != precursor_chromatograms_.end(); ++pr_it)
```
timosachsenberg, line 377: does auto work here?

timosachsenberg: as range for loop

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
cbielow: yep.
UInt32 is used for transitions only

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
cbielow: largest requirement is 8 and vector and string are both multiples of that.
So no matter where I put Size, I always loose 4 byte.

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
cbielow: > What happens again if I just pass an lvalue?

I does not compile and forces you to std::move.
And for now, that is exactly what I want :)

src/openms/include/OpenMS/FORMAT/OSWFile.h
cbielow: 32 sec for a full query and 44 if done individually.
The thing is, I don't want to let the user wait 44 secs in TOPPView when annotating, so I just read the protein list and fill them in on demand (very fast for single items)

src/openms/include/OpenMS/FORMAT/OSWFile.h
cbielow: I think so... not my code.

timosachsenberg (approved):

src/openms/include/OpenMS/FORMAT/OSWFile.h
timosachsenberg: ok that's faster than expected. maybe add this motivation as a comment?

src/openms/include/OpenMS/DATASTRUCTURES/OSWData.h
timosachsenberg: pyOpenMS entered the conversation

PR #4983, opened 2020-09-30
[CI] Add --output-on-failure to travis ctest invocations

lkuchenb:
Makes ctest print the tool output when a test fails.


timosachsenberg:
X!Tandem segfault seems unrelated

PR #4984, opened 2020-09-30
Make documentation links point to the exact version of the tool

jpfeuffer:
See https://github.com/bioconda/bioconda-recipes/pull/24604#issuecomment-701465355

### DIFF
--- a/src/openms/include/OpenMS/APPLICATIONS/TOPPBase.h
+++ b/src/openms/include/OpenMS/APPLICATIONS/TOPPBase.h
@@ -371,6 +371,7 @@ private:
     */
     String getSubsection_(const String& name) const;
 
+    /// Returns a link to the documentation of the tool (accessible on our servers and only after inclusion in the nightly branch or a release).
     String getDocumentationURL() const;
 
     /// Returns the default parameters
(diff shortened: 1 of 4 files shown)


timosachsenberg (approved):
makes sense.
Will we get some errors if we use a non-release, no git hash etc. version?
They might have other version strings and point to non-existing documentation.
I guess not as tests are passing

jpfeuffer:
Hm didn't think about that. Probably should point to nightly then.

bernt-matthias:
Note that there is still one missclassified docs / tools (topp vs utils):

- http://www.openms.de/doxygen/release/2.6.0/html/TOPP_FeatureFinderSuperHirn.html

and two absent docs pages

- http://www.openms.de/doxygen/release/2.6.0/html/UTILS_LowMemPeakPickerHiRes.html
- http://www.openms.de/doxygen/release/2.6.0/html/UTILS_LowMemPeakPickerHiResRandomAccess.html

jpfeuffer:
Where are the peak pickers referenced? They were removed in favor of a single tool so it should not show up anymore.

bernt-matthias:
My mistake. I still had 2.5 ctd files lying around. So ignore the peak pickers.

bernt-matthias:
So its only FeatureFinderSuperHirn

timosachsenberg:
add ,false here https://github.com/OpenMS/OpenMS/blob/d17cc251fd0c4068eb253b03c9fb107897771fdc/src/utils/FeatureFinderSuperHirn.cpp#L77

move from https://github.com/OpenMS/OpenMS/blob/5e1955dd1f07282105944d4f5e5370a2ca9b3286/src/openms/source/APPLICATIONS/ToolHandler.cpp#L68
to util part

Issue #4900, opened 2020-08-28
Add option to IDFilter to keep top n PeptideIdentifications based on best score


timosachsenberg:
-best:n_spectra

timosachsenberg:
and think about how to handle multiple runs

jpfeuffer:
Same could be done for IDMerger already.

PR #4994, opened 2020-10-01
[FEATURE] retain top n spectra (better spectrum = better PSM)

timosachsenberg:
# Description

Addresses https://github.com/OpenMS/OpenMS/issues/4900

# Checklist:

### DIFF
--- a/src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
+++ b/src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
@@ -1156,7 +1156,7 @@ public:
                                  all_peptides);
     }
 
-    /// Filter identifications by "N best" PepitdeIdentification objects (better PId means better PeptideHit than other).
+    /// Filter identifications by "N best" PeptideIdentification objects (better PeptideIdentification means better [best] PeptideHit than other).
     /// The vector is sorted and reduced to @p n elements. If the vector's size 's' is less than @p n, only 's' best spectra are kept.
     static void keepNBestSpectra(std::vector<PeptideIdentification>& peptides, Size n);
 
@@ -1370,4 +1370,3 @@ public:
   };
 
 } // namespace OpenMS
-
--- a/src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
+++ b/src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
@@ -1156,7 +1156,8 @@ public:
                                  all_peptides);
     }
 
-    /// Filter identifications by "N best" spectra. If a spectrum has a better PSM (PeptideHit) than another it is considered better.
+    /// Filter identifications by "N best" PepitdeIdentification objects (better PId means better PeptideHit than other).
+    /// The vector is sorted and reduced to @p n elements. If the vector's size 's' is less than @p n, only 's' best spectra are kept.
     static void keepNBestSpectra(std::vector<PeptideIdentification>& peptides, Size n);
 
     /// Filters a Consensus/FeatureMap by keeping the N best peptide hits for every spectrum
--- a/src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
+++ b/src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
@@ -1156,6 +1156,9 @@ public:
                                  all_peptides);
     }
 
+    /// Filter identifications by "N best" spectra. If a spectrum has a better PSM (PeptideHit) than another it is considered better.
+    static void keepNBestSpectra(std::vector<PeptideIdentification>& peptides, Size n);
+
     /// Filters a Consensus/FeatureMap by keeping the N best peptide hits for every spectrum
     template <class MapType>
     static void keepNBestPeptideHits(MapType& map, Size n)
--- a/src/openms/source/FILTERING/ID/IDFilter.cpp
+++ b/src/openms/source/FILTERING/ID/IDFilter.cpp
@@ -801,7 +801,7 @@ namespace OpenMS
       }                
     }
 
-    // there might be less spectra identified then n -> adapt
+    // there might be less spectra identified than n -> adapt
     n = std::min(n, peptides.size());
 
     auto has_better_peptidehit = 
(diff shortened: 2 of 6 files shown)


src/openms/source/FILTERING/ID/IDFilter.cpp
```diff
@@ -782,6 +782,34 @@ namespace OpenMS
...
+  void IDFilter::keepNBestSpectra(std::vector<PeptideIdentification>& spectra, Size n)
+  {
+    for (PeptideIdentification& s : spectra) { s.sort(); }
+
+    // there might be less spectra identified then n -> adapt
+    n = std::min(n, spectra.size());    
+    std::partial_sort(spectra.begin(), spectra.begin() + n, spectra.end(),
+      [] (const PeptideIdentification& l, const PeptideIdentification& r) 
+      {
+        if (r.getScoreType() != l.getScoreType())
```
cbielow, line 794: this compares nlogn times, but only n times would be required.
How about
`if (find_if(spectra.begin(), spectra.end(), [&](const PeptideIdentification& pep) { return (spectra.begin()->getScoreType() !=pep.getScoreType()) } != spectra.end()) throw ...` before the sorting
`for_each` would be even less code, but we're not at C++17 **yet**

src/openms/source/FILTERING/ID/IDFilter.cpp
```diff
@@ -782,6 +782,34 @@ namespace OpenMS
     }
   }
 
+  void IDFilter::keepNBestSpectra(std::vector<PeptideIdentification>& spectra, Size n)
```
cbielow, line 785: Why call this `spectra` and not `IDs`?
Technically you do not deal with spectra, but I guess ID might be mistaken for PSM, i.e. PeptideHit?
Any other suggestions?

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -1156,6 +1156,9 @@ namespace OpenMS
                                  all_peptides);
     }
 
+    /// Filter identifications by "N best" spectra. If a spectrum has a better PSM (PeptideHit) than another it is considered better.
```
cbielow, line 1159: `The vector is sorted and reduced to @p n elements. If the vector's size 's' is less than @p n, only 's' best spectra are kept.

src/openms/source/FILTERING/ID/IDFilter.cpp
timosachsenberg: It’s Approximately (last-first)log(middle-first) applications of cmp. So it is basically O(n) in practice. Don’t know if worth the effort

src/openms/source/FILTERING/ID/IDFilter.cpp
timosachsenberg: puh yeah I always think about spectra and PSMs in that context. What you want here is to retain the best n spectra and their PSMs. In the new data structure this would be InputItem but isn't much more telling either

src/openms/source/FILTERING/ID/IDFilter.cpp
cbielow: overhead aside, it just felt like a surprising place to find this checking code inside a sorting subfunction...
I have no strong feelings here, but it's a minor "WTF" moment.

timosachsenberg: ok I will do it in an external loop

src/openms/source/FILTERING/ID/IDFilter.cpp
timosachsenberg: Will rename

timosachsenberg:
I think it is good to go now

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -1156,6 +1156,10 @@ namespace OpenMS
                                  all_peptides);
     }
 
+    /// Filter identifications by "N best" PepitdeIdentification objects (better PId means better PeptideHit than other).
```
jpfeuffer, line 1159:
```suggestion
    /// Filter identifications by "N best" PeptideIdentification objects (better PeptideIdentification means better [best] PeptideHit than other).
```

src/tests/class_tests/openms/source/IDFilter_test.cpp
```diff
@@ -971,6 +971,33 @@ START_SECTION((template <class PeakT> static void keepNBestHits(MSExperiment<Pea
...
+{
+  vector<ProteinIdentification> proteins;
+  vector<PeptideIdentification> peptides;
+  IdXMLFile().load(OPENMS_GET_TEST_DATA_PATH("IDFilter_test5.idXML"),
+                   proteins, peptides);
+
+  cout << peptides[0].getHits()[0].getSequence().toString() << endl;
+  cout << peptides[1].getHits()[0].getSequence().toString() << endl;
+
+  IDFilter::keepNBestSpectra(peptides, 2); // keep best to spectra (those with best hits)
```
jpfeuffer, line 984:
```suggestion
  IDFilter::keepNBestSpectra(peptides, 2); // keep best two spectra (those with best hits)
```

src/openms/source/FILTERING/ID/IDFilter.cpp
```diff
@@ -782,6 +782,47 @@ namespace OpenMS
...
+      else
+      {
+        if (p.getScoreType() != score_type)
+        {
+          throw Exception::Precondition(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, String("PSM score types must be identical to allow proper filtering."));
+        }
+      }                
+    }
+
+    // there might be less spectra identified then n -> adapt
```
jpfeuffer, line 804:
```suggestion
    // there might be less spectra identified than n -> adapt
```

jpfeuffer (changes requested):

jpfeuffer:
AFAIK you can accept changes in a batch ;) for next time

jpfeuffer (approved):

timosachsenberg:
thanks for the fixes

PR #4997, opened 2020-10-05
remove XML file and store data locally in ResidueDB

JangJang3:
# Description

Removing the XML file from the ResidueDB files to disband the dependency of the XML file from OpenMS.

# Checklist:


timosachsenberg:
Thanks for trying to get rid of ResidueDB.
One problem with this approach is that there is still parsing going one. Only that this time it is from a map.
While this is much faster than parsing from a file, string operations are still very inefficient.
I think a better approach would be to fill the datastructures directly (e.g, create one residue after the other and add them to the internal maps).

timosachsenberg:
@cbielow what is your take on that? These changes already remove the need for parsing the Residue XML file. Should we accept as is and potentially improve on that later?

JangJang3:
@timosachsenberg thanks for the review! I will look into it and work on your idea.

timosachsenberg:
let's wait for some feedback by chris or others. maybe it is not necessary at this point

timosachsenberg:
Don't forget to register for https://hacktoberfest.digitalocean.com/
;)
After doing so you could take a look at ElementsDB ?

cbielow:
I agree with Timo that we could simply create a c'tor for Residue which fills all members and init them directly.
This should even be less code than now, and much faster as well.
If you really really want to keep it the way it is, then you can also use `vector<pair<string,string>>` instead of a map (I dont see why you'd need one).

timosachsenberg:
The constructor Chris mentions would be similar to this one
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/CHEMISTRY/Residue.h#L192
but with additional parameters, so one can fill all entries during construction. 
Then one can write:
Residue r = {"Alanine", "Ala", "A", "C12H12N12O12", ...} to construct the Residue object without any parsing

timosachsenberg:
TODO: remove residue xml file

JangJang3:
Thank you both for the feedback! Okay, then I will work on Timo's idea and change the files accordingly.

jpfeuffer:
I think it is nice.

What about an additional constructor without mono and avg mass such that it is automatically calculated from the Formula?
What about predefining common losses in the beginning?
Maybe the compiler does all that but it might be cleaner.

timosachsenberg:
I can take a quick look at that

timosachsenberg:
moved to https://github.com/OpenMS/OpenMS/pull/5092

PR #5004, opened 2020-10-08
TOPPView refactor

cbielow:
# Description

 * fix variable names in LayerData.h
 * rename enum due to macro clash (when adding OSWData.h header to SpectrumCanvas3D)
 * some refactoring in IDView (prep for OSW integration)

# Checklist:

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/SqliteConnector.h
+++ b/src/openms/include/OpenMS/FORMAT/SqliteConnector.h
@@ -92,6 +92,10 @@ public:
       return tableExists(db_, tablename);
     }
 
+    /// Counts the number of entries in SQL table @p table_name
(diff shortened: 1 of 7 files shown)


timosachsenberg (approved):

timosachsenberg:
one idea that came to me: could we just put all the exclusive layer elements in a boost variant?

cbielow:
rebuild jenkins

cbielow:
> 
> 
> one idea that came to me: could we just put all the exclusive layer elements in a boost variant?

This could work for the current use-case, but there might be new ones in the future where we'd need some composition, and then people will start storing pairs etc.. 
Lets keep the idea in mind when we decide what to do. Also, C++17 supports std::variant. No need for boost then.

cbielow:
ready for merge

PR #5010, opened 2020-10-12
use only pubchem DB for one of the SiriusAdapter tests

bernt-matthias:
The execution needs more than 2h on my machine (most of the time idling) .. and also times out on the Galaxy tests. All individual DBs seem to finish in seconds. 

Maybe: 

- could be done for the other SiriusAdapter tests as well
- its just a fix of a symptom

# Description

# Checklist:

### DIFF
--- a/src/tests/topp/THIRDPARTY/third_party_tests.cmake
+++ b/src/tests/topp/THIRDPARTY/third_party_tests.cmake
@@ -347,7 +347,7 @@ if (NOT (${SIRIUS_BINARY} STREQUAL "SIRIUS_BINARY-NOTFOUND"))
 
   # Note that with FingerID, output for compound 79 without feature only
   if (ENABLE_FINGERID_TEST)
-  add_test("TOPP_SiriusAdapter_4" ${TOPP_BIN_PATH}/SiriusAdapter -test -executable "${SIRIUS_BINARY}" -in ${DATA_DIR_TOPP}/THIRDPARTY/SiriusAdapter_2_input.mzML -in_featureinfo ${DATA_DIR_TOPP}/THIRDPARTY/SiriusAdapter_2_input.featureXML  -out_sirius SiriusAdapter_4_output.tmp -out_fingerid SiriusAdapter_4_foutput.tmp -sirius:auto_charge -sirius:profile qtof -sirius:database all)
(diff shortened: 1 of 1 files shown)


oliveralka:
The problem here is that SIRIUS calls the CSI:FingerID service, which is located at the University of Jena. Sometimes the service is unavailable or on heavy load. Usually, the test should take around 550s. Due to these issues, we disabled this test in our CI. This test should be disabled by default - CMake variable: "ENABLE_FINGERID_TEST"

What is the runtime with the PubChem fix on your system? 

I will have a look if and what changes if this is introduced to the other SiriusAdapter/AssayGeneratorMetabo tests.

bernt-matthias:
4.5 seconds

oliveralka:
`TOPP_SiriusAdapter_4` fails on my machine with the database pubchem. 

I think is is working that fast, since no valid match is found in the pubchem database and an empty file is returned. 

Did you have a look at the output? 

Since the CI is not testing - it does not fail here. I will look further into it.

oliveralka:
After consultation with the developers from Jena - it seems there are currently issues with their infrastructure and heavy load. 
Of course, it is not the best idea to not test something regularly, but if it is that dependent on a service we have no control over - I think it is best to use the test for development purpose only. 

The other SiriusAdapter tests are running locally, without any dependency on another service. This in unfortunately not possible for TOPP_SiriusAdapter_4, since it is dependent on the CSI:FingerID service. 

What would be your opinions here? @timosachsenberg @bernt-matthias

timosachsenberg:
In principle, it would be good to have it tested regularly. We could check if there is another category of "optional" tests in ctest that are allowed to fail. That way we could still get fail statistics in CDASH

timosachsenberg:
but I guess in this case it might be better to disable the test

bernt-matthias:
Maybe QProcess allows to specify a reasonable time out? Then the Adapter coud produce a exit code that tiggers just a warning in the test. 

... but this would still does not help me with the Galaxy test .. no idea how I would cope with it over there. 

one could use pre-computed data in the case of a time out and the `-test` flag is given. But seems quite a bit of extra work for "temporary" problems.

Best would be to have an offline version of this component.

oliveralka:
> Best would be to have an offline version of this component.
Unfortunately, an offline version is not possible due to the database size in the backend and licensing restrictions.  

I will try to think of something, but currently, I have no solution on hand.

PR #5015, opened 2020-10-15
FLASHDeconv 

JeeH-K:
# Description

Adding a new tool to OpenMS, FLASHDeconv.
More information: https://www.openms.de/comp/flashdeconv/ 

# Checklist:


cmake/Windows/Contrib/Inetc/inetc.vcxproj
```diff
@@ -42,42 +42,42 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release Unicode|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v110</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
```
timosachsenberg, line 45: @jpfeuffer probably not a good idea to update it here ?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,100 @@
...
+#include <OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvHelperStructs.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/QScore.h>
+
+namespace OpenMS
+{
+  struct PeakGroup;
+
+  struct OPENMS_DLLAPI DeconvolutedSpectrum
```
timosachsenberg, line 53: please add documentation to each struct/class and also to each member incl. parameters.
see e.g., https://github.com/cbielow/OpenMS/blob/efe0fce75beb1eaa040ab40c49aa737bcdae0ae7/src/openms/include/OpenMS/FORMAT/SqliteConnector.h how parameters are documented or how documentation is put into certain subpage etc.

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,100 @@
...
+  {
+
+  public:
+    typedef FLASHDeconvHelperStructs::Parameter Parameter;
+    typedef FLASHDeconvHelperStructs::LogMzPeak LogMzPeak;
+
+    DeconvolutedSpectrum();
+
+    //takes spectrum and scan number
+    explicit DeconvolutedSpectrum(MSSpectrum &s, int n);
```
timosachsenberg, line 63: why does the scan number needs to be passed? why is MSSpectrum not const?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,100 @@
...
+    static void writeThermoInclusionHeader(std::fstream &fs);
+
+    bool empty() const;
+
+    // for memory save... clear unnecessary information in mass tracing
+    void clearPeakGroupsChargeInfo();
+
+    bool registerPrecursor(DeconvolutedSpectrum &precursorSpectrum);
+
+    MSSpectrum *spec;
```
timosachsenberg, line 90: consider making those private (and the struct a class)
please add documentation

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
```diff
@@ -0,0 +1,104 @@
...
+    typedef FLASHDeconvHelperStructs::Parameter Parameter;
+    typedef FLASHDeconvHelperStructs::PrecalculatedAveragine PrecalculatedAveragine;
+    typedef FLASHDeconvHelperStructs::LogMzPeak LogMzPeak;
+
+    /// default constructor
+    /// 
+    FLASHDeconvAlgorithm(FLASHDeconvHelperStructs::PrecalculatedAveragine &avg, Parameter &param);
+
+    /// default destructor
+    ~FLASHDeconvAlgorithm();
```
timosachsenberg, line 78: ~FLASHDeconvAlgorithm() = default ?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
```diff
@@ -0,0 +1,104 @@
...
+
+    // deconvolution main function
+    void getPeakGroups(DeconvolutedSpectrum &spec, int &specIndex, int &massIndex);
+
+    // convert double to nominal mass
+    static int getNominalMass(double m);
+
+  protected:
+    // FLASHDeconv parameter set
+    Parameter &param;
```
timosachsenberg, line 94: inherit from DefaultParamHandler?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
```diff
@@ -0,0 +1,104 @@
...
+    // precalcualted averagine patterns
```
timosachsenberg, line 95: typo

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
```diff
@@ -0,0 +1,104 @@
...
+    FLASHDeconvHelperStructs::PrecalculatedAveragine &avg;
+
+    //The data structures for spectra overlapping.
+    std::vector<std::vector<Size>> prevMassBinMap;
+    std::vector<double> prevMinBinLogMassMap;
+  private:
```
timosachsenberg, line 101: remove?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
```diff
@@ -0,0 +1,104 @@
...
+
+    /// default constructor
+    /// 
+    FLASHDeconvAlgorithm(FLASHDeconvHelperStructs::PrecalculatedAveragine &avg, Parameter &param);
+
+    /// default destructor
+    ~FLASHDeconvAlgorithm();
+
+    /// copy constructor
+    // FLASHDeconvAlgorithm(const FLASHDeconvAlgorithm &);
```
timosachsenberg, line 81: what about the copy constructor? can be default?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
```diff
@@ -0,0 +1,104 @@
...
+    typedef FLASHDeconvHelperStructs::PrecalculatedAveragine PrecalculatedAveragine;
+    typedef FLASHDeconvHelperStructs::LogMzPeak LogMzPeak;
+
+    /// default constructor
+    /// 
+    FLASHDeconvAlgorithm(FLASHDeconvHelperStructs::PrecalculatedAveragine &avg, Parameter &param);
+
+    /// default destructor
+    ~FLASHDeconvAlgorithm();
+
```
timosachsenberg, line 79: add move constructor?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHIda.h
```diff
@@ -0,0 +1,94 @@
...
+                      int msLevel,
+                      char *name,
+                      double qScoreThreshold);
+
+    void getIsolationWindows(double *wstart, double *wend, double *qScores, int* charges, double* avgMasses);
+
+  protected:
+    std::map<int, std::vector<double>> selected; // int mass, rt, qscore
+
+    Parameter &param;
```
timosachsenberg, line 72: Parameter& same for pointers

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
timosachsenberg: would need to take a closer look how this plays with parameter passing from C'

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHProFilterAlgorithm.h
```diff
@@ -0,0 +1,78 @@
...
+  public:
+    FLASHProFilterAlgorithm(const String &fasta);
+
+    ~FLASHProFilterAlgorithm();
+
+    double *getScores(MSSpectrum &decovSpec, double intThreshold = .0);
+
+  protected:
+    std::vector<FASTAFile::FASTAEntry> fastaEntry;
+    //Eigen::MatrixXcf
```
timosachsenberg, line 73: remove commented lines?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/MassFeatureTrace.h
```diff
@@ -0,0 +1,94 @@
...
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Kyowon Jeong, Jihyung Kim $
+// $Authors: Kyowon Jeong, Jihyung Kim $
+// --------------------------------------------------------------------------
+
+#pragma once
+
+#ifndef OPENMS_HOST_MASSFEATURETRACE_H
```
timosachsenberg, line 37: required? is pragma once not sufficient?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/MassFeatureTrace.h
```diff
@@ -0,0 +1,94 @@
...
+    Parameter &param;
+    Param &mtd_param;
+    PrecalculatedAveragine &averagines;
+    std::unordered_map<double, std::unordered_map<double, PeakGroup>> peakGroupMap; // rt , mono mass, peakgroup
+  };
+}
+
+
+
+#endif //OPENMS_HOST_MASSFEATURETRACE_H
```
timosachsenberg, line 94: remove?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h
```diff
@@ -0,0 +1,102 @@
...
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Kyowon Jeong, Jihyung Kim $
+// $Authors: Kyowon Jeong, Jihyung Kim $
+// --------------------------------------------------------------------------
+
+
```
timosachsenberg, line 35: remove additional whitespaces here

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h
```diff
@@ -0,0 +1,102 @@
...
+namespace OpenMS
+{
+  // data structure for peak group. A mass contains multiple peaks of different charges and isotope indices
+  struct OPENMS_DLLAPI PeakGroup
+  {
+    typedef FLASHDeconvHelperStructs::Parameter Parameter;
+    typedef FLASHDeconvHelperStructs::LogMzPeak LogMzPeak;
+    typedef FLASHDeconvHelperStructs::PrecalculatedAveragine PrecalculatedAveragine;
+
+    // the spectrum from which a peak group is generated
```
timosachsenberg, line 56: check if some variables can be moved to protected (and strcut turn into class)

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h
```diff
@@ -0,0 +1,102 @@
...
+    MSSpectrum *spec;
+    // the peaks contained in this peak group
+    std::vector<LogMzPeak> peaks;
+    // information for identity
+    int scanNumber, specIndex;
```
timosachsenberg, line 61: if members of built-in types are initialized in the header file it should be done for all of them I guess

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/PeakGroupScoring.h
```diff
@@ -0,0 +1,133 @@
...
+#pragma once
+
+#include <Eigen/Dense>
+#include <OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvHelperStructs.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/QScore.h>
+
+#ifdef _OPENMP
+
```
timosachsenberg, line 45: remove extra line

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/PeakGroupScoring.h
```diff
@@ -0,0 +1,133 @@
...
+  #include <omp.h> // for test..
+#endif
+namespace OpenMS
```
timosachsenberg, line 48: add newline

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/SpectrumDeconvolution.h
```diff
@@ -0,0 +1,159 @@
...
+    boost::dynamic_bitset<> massBinsForThisSpectrum;
+    //massBins stores the selected bins for this spectrum + overlapped spectrum (previous a few spectra).
+    boost::dynamic_bitset<> massBins;
+    //mzBins stores the binned log mz peaks
+    boost::dynamic_bitset<> mzBins;
+    //peakGroups stores the decovnoluted mass peak groups
+    std::vector<PeakGroup> peakGroups;
+
+    //This stores the "universal pattern"
+    double *filter{};
```
timosachsenberg, line 98: would be great to get rid of the pointer stuff here. std::vector<double> - you can get the ptr via http://www.cplusplus.com/reference/vector/vector/data/

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/SpectrumDeconvolution.h
```diff
@@ -0,0 +1,159 @@
...
+    //This stores the patterns for harmonic reduction
+    double **harmonicFilter{};
+
+    //This stores the "universal pattern" in binned dimenstion
+    long *binOffsets{};
```
timosachsenberg, line 103: why long?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/SpectrumDeconvolution.h
```diff
@@ -0,0 +1,159 @@
...
+                      float *mzBinIntensities);
+
+    //this function takes the previous deconvolution results (from ovelapped spectra) for sensitive deconvolution of the current spectrum
+    void unionPrevMassBins(double &massBinMinValue,
+                           std::vector<std::vector<Size>> &prevMassBinVector,
+                           std::vector<double> &prevMassBinMinValue,
+                           UInt msLevel);
+
+    //Update mass bins from mz bins and universal pattern. It select candidate mass bins using the pattern, eliminate possible harmonic masses
+    std::vector<Byte *> updateMassBins(double &massBinMinValue,
```
timosachsenberg, line 129: can we get rid of the Byte* ?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/sources.cmake
```diff
@@ -0,0 +1,29 @@
+### the directory name
+set(directory include/OpenMS/ANALYSIS/TOPDOWN)
+
+### list all header files of the directory here
+set(sources_list_h
+		MassFeatureTrace.h
```
timosachsenberg, line 6: sort from A-Z

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/MassTraceDetection.h
```diff
@@ -114,6 +114,7 @@ namespace OpenMS
 
         // parameter stuff
         double mass_error_ppm_;
+        double mass_error_da_; // added by kyowon jeong
```
timosachsenberg, line 117: hehe don't need to sign extra lines. git blame is our friend if we want to know :P

src/openms/include/OpenMS/METADATA/Precursor.h
```diff
@@ -72,7 +72,7 @@ namespace OpenMS
       ECD,                      ///< Electron capture dissociation
       IMD,                      ///< Infrared multiphoton dissociation
       SORI,                     ///< Sustained off-resonance irradiation
-      HCID,                     ///< High-energy collision-induced dissociation
+      HCD,                     ///< High-energy collision-induced dissociation
```
timosachsenberg, line 75: I remember there was some discussion around it... we might need to revisit that one again to be sure this is correct

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+// --------------------------------------------------------------------------
+// $Maintainer: Kyowon Jeong, Jihyung Kim $
+// $Authors: Kyowon Jeong, Jihyung Kim $
+// --------------------------------------------------------------------------
+
+#include "include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h"
+
+namespace OpenMS
+{
+  DeconvolutedSpectrum::DeconvolutedSpectrum()
```
timosachsenberg, line 39: = default in header

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+  }
+
+  DeconvolutedSpectrum::DeconvolutedSpectrum(MSSpectrum &s, int n) :
+      spec(&s), scanNumber(n)
+  {
+  }
+
+  DeconvolutedSpectrum::~DeconvolutedSpectrum()
+  {
+    std::vector<PeakGroup>().swap(peakGroups);
```
timosachsenberg, line 50: not needed?

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+    auto outSpec = MSSpectrum(*spec);
+    outSpec.clear(false);
+    for (auto &pg : peakGroups)
+    {
+      if (pg.peaks.empty())
+      {
+        continue;
+      }
+      auto p = Peak1D(pg.monoisotopicMass, pg.intensity);
+      outSpec.push_back(p);
```
timosachsenberg, line 69: might be possible to outSpec.emplace_back(pg.monoisotopicMass, pg.intensity)

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+    }
+    if (precursorPeakGroup != nullptr && !spec->getPrecursors().empty())
+    {
+      std::vector<Precursor> precursors;
+      Precursor precursor(spec->getPrecursors()[0]);
+      precursor.setCharge(precursorPeakGroup->maxQScoreCharge);
+      precursor.setMZ(precursorPeakGroup->monoisotopicMass);
+      precursor.setIntensity(precursorPeakGroup->intensity);
+      precursors.push_back(precursor);
```
timosachsenberg, line 78: might be faster to provide a mutable getter to Precursors() and then use outSpec.getPrecursors().emplace_back({precursor}) if this part is speed relevant

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+      return;
+    }
+
+    for (auto &pg : peakGroups)
+    {
+      if (pg.peaks.empty())
+      {
+        continue;
+      }
+      double &m = pg.monoisotopicMass;
```
timosachsenberg, line 99: use const if possible: e.g., const double& m

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+        fs << "\t";
+        for (auto &p : pg.peaks)
+        {
+          fs << p.isotopeIndex << ";";
+        }
+        fs << "\t";
+
+        for (auto &p : pg.peaks)
+        {
+          auto tm = pg.monoisotopicMass + p.isotopeIndex * Constants::ISOTOPE_MASSDIFF_55K_U;
```
timosachsenberg, line 152: comment

timosachsenberg: also line below

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+      if (numMaxMS2 > 0 && peakGroups.size() > (Size) numMaxMS2)// max peak count for TopPic
+      {
+        scores.reserve(peakGroups.size());
+        for (auto &pg : peakGroups)
+        {
+          scores.push_back(pg.qScore);
+        }
+        std::sort(scores.begin(), scores.end());
+        qScoreThreshold = std::max(qScoreThreshold, scores[scores.size() - numMaxMS2]);
+        std::vector<double>().swap(scores);
```
timosachsenberg, line 306: I think the combination of reserve and the swap with empty idiom might be slower as it could lead to reallocations. maybe just scores.clear() ? it does not free the memory (fast)

share/OpenMS/CHEMISTRY/unimod_most_freq.xml
```diff
@@ -0,0 +1,1820 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
JeeH-K: this file wasn't needed for FLASHDeconv, removed

jpfeuffer (changes requested):
I added some preliminary comments. I will also wait for the first batch of corrections.

cbielow:
this looks really useful.
I will look over it, if the first round is resolved to avoid duplicates and confusion

src/utils/FLASHDeconv_prev.cpp
```diff
@@ -0,0 +1,3067 @@
+//
```
timosachsenberg, line 1: remove this file?

src/utils/DBsearchDeconvMass_masslist.cpp
```diff
@@ -0,0 +1,512 @@
+// --------------------------------------------------------------------------
```
timosachsenberg, line 1: change filename to match class name

timosachsenberg: or what is the difference to src/utils/DBsearchDeconvMass.cpp ?

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
KyowonJeong: What's the point of this one? Thanks!

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
KyowonJeong: It does not free... as far as I know  

https://stackoverflow.com/questions/35514909/how-to-clear-vector-in-c-from-memory

src/utils/DBsearchDeconvMass_masslist.cpp
JeeH-K: this file should not be included for this pull request. Thank you for pointing it out

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
timosachsenberg: your vector scores is declared outside of the loop. If you do the swap, the memory is freed and a fresh allocation is needed in the next iteration.
if you do a scores.clear() it will set the vector size to 0 but will keep the memory internally reserved (actually it will keep the old data in there without deleting it ) and no reallocation will occur in the next iteration (except if the internal capacity is exceeded). This can speed up things quite a bit compared to doing the swap.

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+           << ",?,?," << (precursorPeak.getMZ())
+           << "," << 0 << ",?,?,?,?"
+           << "," << precursorPeakGroup->qScore
+           << ",f\n";
+      }
+    }
+    else
+    {
+      //double scoreThreshold = 0;
+      std::vector<double> scores;
```
timosachsenberg, line 295: sorry I misread that somehow. Then it would be better to just put the scores vector inside the if block. THen the swap is not needed at all

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,654 @@
...
+          toExclude.insert(i.first);
+        }
+      }
+    }
+
+    nselected.swap(selected);
+    std::map<double, std::map<int, double>>().swap(nselected);
+
+    //double scoreThreshold = 0;
+    std::vector<double> scores;
```
timosachsenberg, line 410: can this be moved into if block? so no wap is needed?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/SpectrumDeconvolution.h
KyowonJeong: Is it OK to use vector<vector<Byte>> then? Or Byte is the problem?

src/openms/include/OpenMS/METADATA/Precursor.h
KyowonJeong: The mzmls that I used all had HCD instead of HCID... Should we keep both of them then?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/SpectrumDeconvolution.h
timosachsenberg: you could use Matrix<Byte> in OpenMS. This might even be faster because all bytes will occupy consecutive memory addresses.

timosachsenberg: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/DATASTRUCTURES/Matrix.h

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
```diff
@@ -0,0 +1,104 @@
...
+    /// copy constructor
+    // FLASHDeconvAlgorithm(const FLASHDeconvAlgorithm &);
+
+    /// assignment operator
+    FLASHDeconvAlgorithm &operator=(const FLASHDeconvAlgorithm &fd);
+
+    // deconvolution main function
```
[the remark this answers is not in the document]
KyowonJeong: I see. The parameters will be added soon. Thanks!

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h
KyowonJeong: PeakGroup contains so much information ... I would move some to protected but making it to class would not be possible..

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
timosachsenberg: I just meant to add a comment on the special constant here

KyowonJeong: Got it!

KyowonJeong: I actually added the comments in Constants.h

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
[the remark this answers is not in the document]
KyowonJeong: This will take some time. I will keep adding comments and parameters :)

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
KyowonJeong: I am on it! :)

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
KyowonJeong: Accessing scan number from spectrum is not easy.. as we discussed. I will change MSSpectrum into const.

But still I am not familiar with all the qualifier thing :)

src/openms/include/OpenMS/METADATA/Precursor.h
timosachsenberg: I think this was one part of the discussion: https://github.com/OpenMS/OpenMS/pull/2080

timosachsenberg: I am inclined to just keep HCD as you proposed but I think there might be some issue with just those changes. If you search for HCID you see that this shows up a few times. e.g., https://github.com/OpenMS/OpenMS/blob/1db1c512204b7345cdd1ba0668b4efd6d9624164/src/tests/topp/THIRDPARTY/spectra_spectrast.mzXML#L20 has HCID in the mzXML file. Is this properly parsed then and stored as HCD?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
KyowonJeong: They are necessary for spectra overlapping..

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
KyowonJeong: Should FLASHDeconvAlgorithm inherit from DefaultParamHandler? There are so many variables specific to FLASHDeconv so I made a separate structure for it. Is it not a good idea?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h
KyowonJeong: I moved them private

timosachsenberg:
Ok I moved the tool to util for now as it is marked as util in the TOPPBase constructor, pyOpenMS bindings are missing as well as class and tool tests. We can later promote it to a tool so there is no immediate hurry.

src/openms/source/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.cpp
```diff
@@ -346,8 +346,8 @@ namespace OpenMS
     long binEnd = (long) massBins.size();
     auto candidateMassBinsForThisSpectrum = boost::dynamic_bitset<>(massBins.size());
     // how many peaks of continuous charges per mass
-    int *continuousChargePeakPairCount = new int[massBins.size()];
-    std::fill_n(continuousChargePeakPairCount, massBins.size(), 0);
+    int *supportPeakPairCount = new int[massBins.size()];
```
timosachsenberg, line 349: Dynamic arrays
Using a pointer to a malloc-ed/new-ed array will be at best as fast as the std::vector version, and a lot less safe (see litb's post). So better use a std::vector.

KyowonJeong: I will change them all then! So just vector<...>(numberofelements, initval) is the way to go? Thanks!

src/topp/FLASHDeconv.cpp
```diff
@@ -0,0 +1,781 @@
...
+
+protected:
+  //typedef FLASHDeconvHelperStructs::Parameter Parameter;
+
+  // this function will be used to register the tool parameters
+  // it gets automatically called on tool execution
+  void registerOptionsAndFlags_() override
+  {
+    registerInputFile_("in", "<input_file/input_dir>", "", "Input file or directory");
+    registerOutputFile_("out", "<output_file_prefix/output_dir>", "",
```
timosachsenberg, line 84: please specify input and output file types. otherwise KNIME or TOPPAS don't know the MIME type of the output file.

KyowonJeong:
Timo - then is it not possible to specify input directory? 

Best 

Kyowon

timosachsenberg:
You can have only one or multiple input files. If possible use only one and let the workflow system handle the processing of directories.

cmake/Windows/Contrib/Inetc/inetc.vcxproj
timosachsenberg: @jpfeuffer revert?

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,456 @@
...
+        }
+      }
+      fs << "\t";
+      int isoEndIndex = 0;
+
+      for (auto &p : pg)
+      {
+        isoEndIndex = isoEndIndex < p.isotopeIndex ? p.isotopeIndex : isoEndIndex;
+      }
+      auto perIsotopeIntensity = new double[isoEndIndex + 1];
```
timosachsenberg, line 182: 1. use std::array or std::vector but please no C array. Why? because you forgot to delete it ;)

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,456 @@
...
+  {
+    for (auto &pg: *this)
+    {
+      pg.clearChargeInfo();
+    }
+  }
+
+  void DeconvolutedSpectrum::writeTopFD(std::fstream &fs, int id, const FLASHDeconvHelperStructs::PrecalculatedAveragine &avg)//, fstream &fsm, fstream &fsp)
+  {
+    auto msLevel = spec.getMSLevel();
```
timosachsenberg, line 272: Please don't use auto for simple types like int or double

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    /// original spectrum setter
+    MSSpectrum &getOriginalSpectrum();
+
+    /// peakGroup getter
+    PeakGroup getPrecursorPeakGroup();
+
+    /// precursor charge getter : set in registerPrecursor
+    int getPrecursorCharge();
+
+    /// get max mass - which is min mass between max mass specified by users or mass determined by precursor mass for MSn
```
timosachsenberg, line 126: better documentation needed.

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    double getCurrentMaxMass(double maxMass);
+
+    /// get max charge - which is min charge between max charge specified by users or charge determined by precursor charge for MSn
+    int getCurrentMaxCharge(int maxCharge);
```
timosachsenberg, line 130: better documentation needed.

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+namespace OpenMS
+{
+  class PeakGroup;
+
+  /**
+       @brief A class representing a deconvoluted spectrum. Also contains deconvoluted precursro information for MSn n>1.
+  */
+  class OPENMS_DLLAPI DeconvolutedSpectrum :
+      public std::vector<PeakGroup>
```
timosachsenberg, line 50: is it necessary to publicly derive from std::vector?
see 
https://stackoverflow.com/questions/4353203/thou-shalt-not-inherit-from-stdvector#:~:text=The%20main%20reason%20for%20not,from%20polymorphic%20use%20of%20descendants.
for details

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /// default constructor
+    DeconvolutedSpectrum() = default;
+
+    /**
+       @brief Constructor for DeconvolutedSpectrum
+       @param s spectrum
+       @param n scan number
+  */
+    explicit DeconvolutedSpectrum(const MSSpectrum &s, int n);
```
timosachsenberg, line 63: 1. please use spectrum instead of s and scan_number instead of n.
2. extent parameter documentation. It currently just repeats what is obvious from the types.

timosachsenberg, line 63: Use MSSpectrum& instead of MSSpectrum &T for consistent style in OpenMS

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    /// assignment operator
+    DeconvolutedSpectrum &operator=(const DeconvolutedSpectrum &fd) = default;
+
+    /**
+        @brief write the header in the output file (spectrum level)
+        @param fs file stream to the output file
+        @param n the index to the mass
+        @param detail if set true, detailed information of the mass (e.g., peak list for the mass) is written
+   */
+    static void writeDeconvolutedMassesHeader(std::fstream &fs, int &n, bool detail);
```
timosachsenberg, line 83: check for every parameter if it can be const.. e.g. I would expect
static void writeDeconvolutedMassesHeader(std::fstream &fs, const int n, const bool detail);

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+      @brief write the deconvoluted masses in the output file (spectrum level)
+      @param fs file stream to the output file
+      @param param FLASHDeconv paramter
+    */
+    void writeDeconvolutedMasses(std::fstream &fs,
+                                 const String &fileName,
```
timosachsenberg, line 91: use file_name instead of fileName for consistent style

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+                                 const FLASHDeconvHelperStructs::PrecalculatedAveragine &avg,
+                                 bool writeDetail);
```
timosachsenberg, line 93: check if parameter can be const

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+        @param detail if set true, detailed information of the mass (e.g., peak list for the mass) is written
+   */
+    static void writeDeconvolutedMassesHeader(std::fstream &fs, int &n, bool detail);
+
+    /**
+      @brief write the deconvoluted masses in the output file (spectrum level)
+      @param fs file stream to the output file
+      @param param FLASHDeconv paramter
+    */
+    void writeDeconvolutedMasses(std::fstream &fs,
```
timosachsenberg, line 90: check if method can be declared const

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+        @param fs file stream to the output file
+        @param n the index to the mass
+        @param detail if set true, detailed information of the mass (e.g., peak list for the mass) is written
+   */
+    static void writeDeconvolutedMassesHeader(std::fstream &fs, int &n, bool detail);
+
+    /**
+      @brief write the deconvoluted masses in the output file (spectrum level)
+      @param fs file stream to the output file
+      @param param FLASHDeconv paramter
```
timosachsenberg, line 88: typo: parameter

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+#include <OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvHelperStructs.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h>
+#include <iomanip>
+
+namespace OpenMS
+{
+  class PeakGroup;
+
+  /**
+       @brief A class representing a deconvoluted spectrum. Also contains deconvoluted precursro information for MSn n>1.
```
timosachsenberg, line 47: typo precursor

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+                                 const String &fileName,
+                                 const FLASHDeconvHelperStructs::PrecalculatedAveragine &avg,
+                                 bool writeDetail);
+
+    /**
+      @brief write the deconvoluted masses TopFD format
+      @param fs file stream to the output file
+      @param id the index to the spectrum. updated outside.
+ */
+    void writeTopFD(std::fstream &fs, int id, const FLASHDeconvHelperStructs::PrecalculatedAveragine &avg);
```
timosachsenberg, line 100: can id be const?
can the method be const?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /// cast DeconvolutedSpectrum into MSSpectrum object to write mzml format
+    MSSpectrum toSpectrum(int charge);
```
timosachsenberg, line 103: can charge be const?
can the method be const?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /// write the header for Thermo Inclusion List header format
+    static void writeThermoInclusionHeader(std::fstream &fs);
+
+    /// for memory save... clear unnecessary information in mass tracing
```
timosachsenberg, line 108: to save memory

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    void clearPeakGroupsChargeInfo();
+
+    /**
+     @brief register the precusor info in this from the precursor DeconvolutedSpectrum
+     @param precursorSpectrum the precursor DeconvolutedSpectrum
+     */
+    bool registerPrecursor(DeconvolutedSpectrum &precursorSpectrum);
```
timosachsenberg, line 115: - can precursorSpectrum be const?
- rename precursorSpectrum to precursor_spectrum

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /// original spectrum setter
```
timosachsenberg, line 117: update comment

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    MSSpectrum &getOriginalSpectrum();
```
timosachsenberg, line 118: add also const MSSpectrum& getOriginalSpectrum() const;
so we have both mutable and non-mutable access

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /// peakGroup getter
```
timosachsenberg, line 120: improve comment

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    double getCurrentMaxMass(double maxMass);
+
+    /// get max charge - which is min charge between max charge specified by users or charge determined by precursor charge for MSn
+    int getCurrentMaxCharge(int maxCharge);
+
+  private:
+    /// the original spectrum from which this is generated
+    MSSpectrum spec;
+    /// precursor peakGroup (or mass)
+    PeakGroup *precursorPeakGroup = nullptr;
```
timosachsenberg, line 136: why does this need to be a pointer?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /// get max mass - which is min mass between max mass specified by users or mass determined by precursor mass for MSn
+    double getCurrentMaxMass(double maxMass);
+
+    /// get max charge - which is min charge between max charge specified by users or charge determined by precursor charge for MSn
+    int getCurrentMaxCharge(int maxCharge);
+
+  private:
+    /// the original spectrum from which this is generated
+    MSSpectrum spec;
```
timosachsenberg, line 134: rename to spec_ as it is private

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
timosachsenberg, line 136: rename to precursor_peak_group_ as it is private

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    /// precursor peakGroup (or mass)
+    PeakGroup *precursorPeakGroup = nullptr;
+    /// precursor peak (not deconvoluted one)
+    Precursor precursorPeak;
```
timosachsenberg, line 138: rename to precursor_peak_

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    /// activation method for file output
+    std::string activationMethod;
```
timosachsenberg, line 140: - could be redundant. is this stored in the precursor?
- rename to activation_method_

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    /// scan number and precursor scan number
+    int scanNumber, precursorScanNumber;
```
timosachsenberg, line 142: - rename to scan_number_ and precursor_scan_number_
- is precursor_scan_number_ redundant?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+  };
+}
```
timosachsenberg, line 144: add newline at end of file

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+    MSSpectrum &getOriginalSpectrum();
+
+    /// peakGroup getter
+    PeakGroup getPrecursorPeakGroup();
+
+    /// precursor charge getter : set in registerPrecursor
+    int getPrecursorCharge();
+
+    /// get max mass - which is min mass between max mass specified by users or mass determined by precursor mass for MSn
+    double getCurrentMaxMass(double maxMass);
```
timosachsenberg, line 127: rename to max_mass

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
timosachsenberg, line 130: rename parameter to max_charge

timosachsenberg (changes requested):
I highlighted issues in this single header file that should be fixed in all files

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,456 @@
...
+      precursorScanNumber = precursorSpectrum.scanNumber;
+      auto startMz = p.getIsolationWindowLowerOffset() > 100.0 ?
+                     p.getIsolationWindowLowerOffset() :
+                     -p.getIsolationWindowLowerOffset() + p.getMZ();
+      auto endMz = p.getIsolationWindowUpperOffset() > 100.0 ?
+                   p.getIsolationWindowUpperOffset() :
+                   p.getIsolationWindowUpperOffset() + p.getMZ();
+
+      double maxSumIntensity = 0.0;
+      for (auto &pg: precursorSpectrum)
```
timosachsenberg, line 368: can this be const here? e.g. no sorting is required? because then the whole method could take a const DeconvolutedSpectrum&

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
```diff
@@ -0,0 +1,456 @@
...
+      {
+        std::sort(pg.begin(), pg.end());
+        if (pg[0].mz > endMz || pg[pg.size() - 1].mz < startMz)
+        {
+          continue;
+        }
+
+        double sumIntensity = .0;
+        double maxIntensity = .0;
+        LogMzPeak *tmp = nullptr;
```
timosachsenberg, line 378: can this be const LogMzPeak* ?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+       @brief Constructor for DeconvolutedSpectrum
+       @param s spectrum
+       @param n scan number
+  */
+    explicit DeconvolutedSpectrum(const MSSpectrum &s, int n);
+
+    /// default deconstructor
+    ~DeconvolutedSpectrum() = default;
```
timosachsenberg, line 66: do you need to free some memory? e.g. you have a pointer in this class. who has ownership?

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
```diff
@@ -0,0 +1,144 @@
...
+
+    /// copy constructor
+    DeconvolutedSpectrum(const DeconvolutedSpectrum &) = default;
```
timosachsenberg, line 69: you have a pointer in this class. who has ownership? important for copy and move !

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
KyowonJeong: Man shame on me.

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
KyowonJeong: That I did not know. I will apply this rule to all source codes..

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
KyowonJeong: You must be right. Actually I do not know very well how to derive from a class in C++. I will look into that.

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
KyowonJeong: It is being revised..

KyowonJeong: Oh it will take some time. Haha being revised.

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
KyowonJeong: This and comment update are being done

src/openms/source/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.cpp
```diff
@@ -0,0 +1,1831 @@
...
+    double isoNorm = avg.getNorm(mass);
+
+    int isoSize = (int) iso.size();
+
+    offset = 0;
+    double maxCosine = -1;
+    int isotopeLength = 0;
+    int maxIsotopeIndex = 0, minIsotopeIndex = -1;
+
+    for (int i = 0; i < avg.getMaxIsotopeIndex(); i++)
```
JeeH-K, line 986: why avg? why not depend on the size of the iso?

src/openms/source/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.cpp
```diff
@@ -0,0 +1,1831 @@
...
+                                                                        PrecalculatedAveragine &avg)
+  {
+    auto iso = avg.get(mass);
+    double isoNorm = avg.getNorm(mass);
+
+    int isoSize = (int) iso.size();
+
+    offset = 0;
+    double maxCosine = -1;
+    int isotopeLength = 0;
```
JeeH-K, line 983: this is not used

src/openms/source/ANALYSIS/TOPDOWN/FLASHDeconvHelperStructs.cpp
```diff
@@ -0,0 +1,230 @@
...
+        {
+          break;
+        }
+        norm -= iso[k].getIntensity() * iso[k].getIntensity();
+        leftIndex--;
+        iso[k].setIntensity(0);
+      }
+
+      Size rightIndex = iso.size() - 1 - mostAbundantIndex;
+      for (Size k = iso.size() - 1; k >= mostAbundantIndex; k--)
```
JeeH-K, line 92: Is this loop needed? iso is already trimmed on the right side at line 62

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h
KyowonJeong: For speed reason, I put that as a parameter. Otherwise for each line output activation method is obtained using for loop.

src/openms/source/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.cpp
KyowonJeong: It caused problems in assigning pointers..

src/openms/include/OpenMS/ANALYSIS/TOPDOWN/FLASHIda.h
```diff
@@ -0,0 +1,190 @@
...
+
+#include <OpenMS/ANALYSIS/TOPDOWN/PeakGroup.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvHelperStructs.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/FLASHDeconvAlgorithm.h>
+#include <OpenMS/ANALYSIS/TOPDOWN/DeconvolutedSpectrum.h>
+
+namespace OpenMS
+{
+  /**
+   * @brief FLASHIda class for real time deconvolution
```
hroest, line 45: can you improve the documentation here, it is not quite clear what this class is doing

hroest:
I looked at some of the header files, generally looks good but it needs more class-level documentation. What is the status of this, could this be merged soon?

PR #5018, opened 2020-10-16
[GITPOD] Better integration of CMake plugin with debugger + update of extensions

jpfeuffer:
# Description

Now one should be able to start debugging right away without additional setup. Also adds e.g. a doxygen and cmake extension and updates intellisense. We could think about clangd instead of Intellisense in the future.

### DIFF
--- a/.vscode-gitpod/launch.json
+++ b/.vscode-gitpod/launch.json
@@ -1,0 +1,6 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  "version": "0.2.0",
+  "configurations": []
+}
--- a/.vscode-gitpod/settings.json
+++ b/.vscode-gitpod/settings.json
@@ -1,4 +1,11 @@
 {
     "editor.tabSize": 2,
     "editor.insertSpaces": true,
+    "cmake.configureArgs": [
+      "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON",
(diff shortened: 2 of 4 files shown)


jpfeuffer:
Bildschirmfoto 2020-10-16 um 17 51 58

No excuses for not completing Hacktoberfest!!

.vscode-gitpod/launch.json
```diff
@@ -0,0 +1,6 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
timosachsenberg, line 2: is // a valid comment here?

.vscode-gitpod/settings.json
```diff
@@ -1,4 +1,11 @@
 {
     "editor.tabSize": 2,
     "editor.insertSpaces": true,
+    "cmake.configureArgs": [
```
timosachsenberg, line 4: what is this for?

timosachsenberg:
Is clangd better than intellisense?

jpfeuffer:
Some say so. At least for VS Code with gcc/clang.

.vscode-gitpod/settings.json
jpfeuffer: It is for the CMake plugin (our usual parameters). At first it was somehow cleaning the build folder and I had to add this to reconfigure. After deactivating the cleaning it might not be necessary since we initialize the cache at startup. But won't hurt unless we change it often.

timosachsenberg (approved):
Looks good.

Issue #5020, opened 2020-10-18
[Error] - Failed to compile with VS2019 MSVC 14.2


ahmedskhalil:
**Description**
OpenMS fails to compile with the aforementioned tools, the compiler complains about the missing header `algorithm` for the OpenSWATHAlgo, namely `Scoring.h`. The header was previously present as is required by `std::sort`.

The exact error is : `~\OpenMS\src\openswathalgo\source\OPENSWATHALGO\ALGO\Scoring.cpp(276,12): error C2039: 'sort': is not a member of 'std' [~\OpenMS\openms_build\src\openswathalgo\OpenSwathAlgo.vcxproj]`

jpfeuffer:
Interesting. It compiles successfully in the nightly builds: http://cdash.openms.de/buildSummary.php?buildid=64211

Which exact version of 14.2 do you use?

Nonetheless, if you add the missing header and it still compiles on our testing machines, it should be fine.

ahmedskhalil:
We're using MSVC `14.27.29110`. I had to automate adding the include directive to successfully build it in the CI.

jpfeuffer:
ok, then you have a newer update than on our build machines. I will upgrade ours and check again. But most likely you are right, feel free to make a pull request if you need it right now.

ahmedskhalil:
Thanks, I made a pull request #5022 to resolve this.

jpfeuffer:
Thanks a lot!!

PR #5022, opened 2020-10-20
[FIX] Failed to compile with VS2019 MSVC 14.2

ahmedskhalil:
# Description
OpenMS fails to compile with the aforementioned tools, the compiler complains about the missing header algorithm for the OpenSWATHAlgo, namely `Scoring.cpp`. The header was previously present as it's required by `std::sort`. See Issue #5020 .

The exact error is : `~\OpenMS\src\openswathalgo\source\OPENSWATHALGO\ALGO\Scoring.cpp(276,12): error C2039: 'sort': is not a member of 'std' [~\OpenMS\openms_build\src\openswathalgo\OpenSwathAlgo.vcxproj]`

# Changes
- Reinstated `#include <algorithm>` in `OpenMS\src\openswathalgo\source\OPENSWATHALGO\ALGO\Scoring.cpp`.

# Checklist:
- [x ] Make sure that you are listed in the AUTHORS file

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,6 +5,7 @@
 The following authors contributed to OpenMS at some point. For the 
 authors contributing to a specific piece of code, please refer to
 the authors tag in the respective file header.
+ - Ahmed Khalil
  - Alexandra Scherbart
  - Alexandra Zerck
  - Andreas Bertsch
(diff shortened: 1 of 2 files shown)


cbielow (approved):
thanks

PR #5024, opened 2020-10-20
[FIX] ConsID not taking over proteins in "linking" mode

jpfeuffer:
# Description

Otherwise you get invalid idXMLs during writing the results. Unfortunately, our tests did not include proteins so it was not covered.

# Checklist:

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -35,3 +35,5 @@ vscode:
     - mhutchie.git-graph@1.26.0:m8EvA1owa1dL2R6+Swieyg==
     - twxs.cmake@0.0.17:aQNzlYa74rJFJ5/BnUF1yA==
     - cschlosser.doxdocgen@0.7.2:g0DIKlW09t+j2Xi04fzrAA==
+    - hbenl.vscode-test-explorer@2.19.2:8rKHUShFJJd0UivektEkng==
+    - fredericbonnet.cmake-test-adapter@0.10.0:Fb9mp0fAtp8s6LKtpzJAyw==
--- a/src/topp/ConsensusID.cpp
+++ b/src/topp/ConsensusID.cpp
@@ -209,6 +209,22 @@ protected:
     StringList merged_spectra_data;
     String engine = prot_ids[0].getSearchEngine();
     String version = prot_ids[0].getSearchEngineVersion();
+
+    // merge proteins
+    unordered_set<String> seen_proteins;
+    vector<ProteinHit> merged_proteins;
+    for (Size i = 0; i < prot_ids.size(); ++i)
+    {
+      for (auto& hit : prot_ids[i].getHits())
+      {
+        const auto& iter_inserted = seen_proteins.emplace(hit.getAccession());
+        if (iter_inserted.second)
+        {
+          merged_proteins.emplace_back(std::move(hit));
+        }
+      }
+    }
+
     for (vector<ProteinIdentification>::iterator it_prot_ids = prot_ids.begin(); it_prot_ids != prot_ids.end(); ++it_prot_ids)
     {
       ProteinIdentification::SearchParameters search_params(it_prot_ids->getSearchParameters());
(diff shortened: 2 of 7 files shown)


.gitpod.yml
```diff
@@ -35,3 +35,5 @@ vscode:
     - mhutchie.git-graph@1.26.0:m8EvA1owa1dL2R6+Swieyg==
     - twxs.cmake@0.0.17:aQNzlYa74rJFJ5/BnUF1yA==
     - cschlosser.doxdocgen@0.7.2:g0DIKlW09t+j2Xi04fzrAA==
+    - hbenl.vscode-test-explorer@2.19.2:8rKHUShFJJd0UivektEkng==
```
timosachsenberg, line 38: just curious... what are these for?

src/topp/ConsensusID.cpp
```diff
@@ -209,6 +209,22 @@ class TOPPConsensusID :
...
+    unordered_set<String> seen_proteins;
+    vector<ProteinHit> merged_proteins;
+    for (Size i = 0; i < prot_ids.size(); ++i)
+    {
+      for (auto& hit : prot_ids[i].getHits())
+      {
+        const auto& iter_inserted = seen_proteins.emplace(hit.getAccession());
+        if (iter_inserted.second)
+        {
+          merged_proteins.emplace_back(std::move(hit));
```
timosachsenberg, line 223: I think emplace_back is used if the object is constructed in-place.
push_back(T&&) also exists and might be more appropriate to use here.

src/topp/ConsensusID.cpp
```diff
@@ -792,10 +810,22 @@ class TOPPConsensusID :
           }
         }
       }
-      else
+      else // link spectra by RT and mz proximity and do ConsensusID on their PSMs
       {
+        if (in.size() != 1)
+        {
+          OPENMS_LOG_WARN << "ConsensusID on idXML without the --per_spectrum flag, expects a single idXML file."
```
timosachsenberg, line 817: should this be a hard error?

src/topp/ConsensusID.cpp
```diff
@@ -792,10 +810,22 @@ class TOPPConsensusID :
...
+          "Please merge the files with IDMerger using its default settings." << std::endl;
+        }
+        // note: this requires a single merged idXML file.
         IdXMLFile().load(in[0], prot_ids, pep_ids, document_id);
 
+        if (prot_ids.size() == 1)
+        {
+          OPENMS_LOG_WARN << "ConsensusID on idXML without the --per_spectrum flag usually expects a merged idXML file"
+          "with multiple runs. Only one run found in the first file." << std::endl;
```
timosachsenberg, line 826: Maybe add a line what this implies (e.g., no processing is done, no new scores are calculated whatever. )

src/topp/ConsensusID.cpp
```diff
@@ -847,27 +877,30 @@ class TOPPConsensusID :
         ConsensusMap grouping;
         linker.group(maps, grouping);
 
+        Size old_size = prot_ids.size();
+        // create new identification run
+        setProteinIdentifications_(prot_ids);
+
         // compute consensus
         pep_ids.clear();
         for (ConsensusMap::Iterator it = grouping.begin(); it != grouping.end();
```
timosachsenberg, line 886: maybe change to range for loop for readability

timosachsenberg (changes requested):

.gitpod.yml
jpfeuffer: Visualizing CTests and potentially starting a debugging session for them. But I think we have too many tests 😂 it loads forever.

src/topp/ConsensusID.cpp
jpfeuffer: Good question. The featurelinker only maps between different maps right? (Not inside the same?) It is hard to predict but nothing should happen. But it will be converted to a FeatureMap and back etc so I bet at least minor things will change.

src/topp/ConsensusID.cpp
jpfeuffer: Yes maybe. I actually left it for "wrong" old workflows. But maybe they should better fail.

timosachsenberg (approved):

src/topp/ConsensusID.cpp
jpfeuffer: Now I am just failing it. Would probably nonsense to run it. I hope no one abused it like this.

jpfeuffer:
rebuild jenkins

PR #5027, opened 2020-10-22
[REFACTOR] introduce behavior base class

timosachsenberg:
# Description

Not much happening here yet. Removed a bit of code duplication and prepare for other refactoring.

# Checklist:

### DIFF
--- a/src/openms_gui/include/OpenMS/VISUAL/TOPPViewIdentificationViewBehavior.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/TOPPViewIdentificationViewBehavior.h
@@ -36,6 +36,7 @@
 
 #include <OpenMS/METADATA/SpectrumSettings.h>
 #include <OpenMS/VISUAL/LayerData.h>
+#include <OpenMS/VISUAL/TVBehaviorBase.h>
 #include <vector>
 
 namespace OpenMS
@@ -48,28 +49,9 @@ namespace OpenMS
   @brief Behavior of TOPPView in identification mode.
   */
   class TOPPViewIdentificationViewBehavior
-    : public QObject
+    : public TVBehaviorBase
   {
     Q_OBJECT
-    ///@name Type definitions
-    //@{
-    //Feature map type
-    typedef LayerData::FeatureMapType FeatureMapType;
-    //Feature map managed type
-    typedef LayerData::FeatureMapSharedPtrType FeatureMapSharedPtrType;
-
-    //Consensus feature map type
-    typedef LayerData::ConsensusMapType ConsensusMapType;
-    //Consensus  map managed type
-    typedef LayerData::ConsensusMapSharedPtrType ConsensusMapSharedPtrType;
-
-    //Peak map type
-    typedef LayerData::ExperimentType ExperimentType;
-    //Main managed data type (experiment)
-    typedef LayerData::ExperimentSharedPtrType ExperimentSharedPtrType;
-    ///Peak spectrum type
-    typedef ExperimentType::SpectrumType SpectrumType;
-    //@}
 
   public:
     /// Construct the behaviour with its parent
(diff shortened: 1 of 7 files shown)


src/openms_gui/include/OpenMS/VISUAL/TOPPViewIdentificationViewBehavior.h
```diff
@@ -36,6 +36,7 @@
 
 #include <OpenMS/METADATA/SpectrumSettings.h>
 #include <OpenMS/VISUAL/LayerData.h>
+#include <OpenMS/VISUAL/BehaviorBase.h>
```
cbielow, line 39: Filename and classname differ. please unify.

(and add the new files to the PR) :)

src/openms_gui/source/VISUAL/TOPPViewSpectraViewBehavior.cpp
```diff
@@ -48,6 +48,10 @@ using namespace std;
 
 namespace OpenMS
 {
+  TOPPViewSpectraViewBehavior::TOPPViewSpectraViewBehavior(TOPPViewBase* parent):
+    TOPPViewBehaviorBase(parent)
```
cbielow, line 52: reminder: `TOPPViewBehaviorBase() = delete;` to avoid accidental instanciations

src/openms_gui/include/OpenMS/VISUAL/TOPPViewIdentificationViewBehavior.h
timosachsenberg: Should we really put TOPPView in front of all those? TV would be much shorter. maybe even a namespace could work.
what is your preference?

cbielow: TV sounds good.
Its TOPPView specific, so `BehaviourBase` is a little too general.
But `TVBehaviourBase` should do.

timosachsenberg:
Ok, I think it is good to go... much more do to though :P

src/openms_gui/include/OpenMS/VISUAL/TVBehaviorBase.h
```diff
@@ -0,0 +1,90 @@
...
+// --------------------------------------------------------------------------
+// $Maintainer: Timo Sachsenberg$
+// $Authors: Timo Sachsenberg $
+// --------------------------------------------------------------------------
+
+#pragma once
+
+#include <OpenMS/METADATA/SpectrumSettings.h>
+#include <OpenMS/VISUAL/LayerData.h>
+#include <vector>
```
cbielow, line 39: not needed?

src/openms_gui/include/OpenMS/VISUAL/TVBehaviorBase.h
```diff
@@ -0,0 +1,90 @@
...
+  class TVBehaviorBase
+    : public QObject
+  {
+    Q_OBJECT
+
+protected:
+    ///@name Type definitions
+    //@{
+    /// Feature map type
+    typedef LayerData::FeatureMapType FeatureMapType;
```
cbielow, line 57: `using LayerData::FeatureMapType;` would be better, but might not be supported by all our compilers until switching to C++17

src/openms_gui/include/OpenMS/VISUAL/TVBehaviorBase.h
```diff
@@ -0,0 +1,90 @@
...
+    /// Peak spectrum type
+    typedef ExperimentType::SpectrumType SpectrumType;
+    //@}
+
+public:
+    TVBehaviorBase() = delete;
+
+    virtual ~TVBehaviorBase() = default;
+public slots:
+    /// Slot for behavior activation
```
cbielow, line 79: add `The default behaviour does nothing. Override in child class if desired.`

cbielow (commented):
nice!
only minor things

src/openms_gui/include/OpenMS/VISUAL/TVBehaviorBase.h
timosachsenberg: yeah let's do this globally after the switch

timosachsenberg:
rebuild jenkins

timosachsenberg:
don't merge yet... just found a bug

Issue #5014, opened 2020-10-14
Tools without category


bernt-matthias:
- DatabaseSuitability
- GNPSExport
- ProteomicsLFQ

oliveralka:
Do you mean in the documentation?

Since it seems they are registered in the ToolHandler with a category:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/APPLICATIONS/ToolHandler.cpp#L77
e.g. `tools_map["GNPSExport"] = Internal::ToolDescription("GNPSExport", "File Handling");`

oliveralka:
Hi @bernt-matthias,
could you shortly specify what you mean? Doxygen documentation or code? 

It seems that the tools have a registered category at least in the ToolHandler.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/APPLICATIONS/ToolHandler.cpp#L77
e.g. `tools_map["GNPSExport"] = Internal::ToolDescription("GNPSExport", "File Handling");`

timosachsenberg:
ok strange. @oliveralka can you check if they are wrongly registered as util or tool?

bernt-matthias:
At least in the generated ctd the category is missing. Also in the knime Screenshot that @timosachsenberg sent me on gitter the three tools were outside of the categories.

timosachsenberg:
It is registered as a util  https://github.com/OpenMS/OpenMS/blob/develop/src/topp/GNPSExport.cpp#L83
and here as a tool https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/APPLICATIONS/ToolHandler.cpp#L77
so I guess this is the issue

jpfeuffer:
This happens a lot. Can we maybe introduce a check? E.g. in the TOPPBase constructor? There were remnants of a check in the getDocumentationURL method but I removed it since it was non-functional and kind of out-of-place.

timosachsenberg:
How would this work in TOPPBase?
Is better to place the check into ToolHandler?

jpfeuffer:
No, you need the classification from TOPPBase. if official, then check ToolHandler::TOPP else check ToolHandler::Utils

jpfeuffer:
plus check that it is not double, etc

jpfeuffer:
Or yes, better, put the check for duplication in toolhandler and the check for matching in ToppBase

timosachsenberg:
[...]

PR #5032, opened 2020-10-25
[FIX] tool categories & additional checks.

oliveralka:
# Description

Fix for #5014
Fixes the tool categories for some tools. 
Adds functionality to check for duplicated entries (TOPP, Utils) 

Let me know if that is what you had in mind. 

# Checklist:

### DIFF
--- a/src/openms/source/APPLICATIONS/TOPPBase.cpp
+++ b/src/openms/source/APPLICATIONS/TOPPBase.cpp
@@ -135,19 +135,31 @@ namespace OpenMS
     // check if tool entries are in Utils and TOPP (duplication)
     if (ToolHandler::checkDuplicated(tool_name_))
     {
-      writeLog_(String("Warning: Message to maintainer - The '") + tool_name_ + "' has entries in the UTILS and TOPP category. Please add it to the correct category in the ToolHandler.");
+      throw Exception::InvalidValue(__FILE__,
+                                    __LINE__,
+                                    OPENMS_PRETTY_FUNCTION,
+                                    String("The '" + tool_name_ + "' has entries in the UTILS and TOPP category. Please add it to the correct category in the ToolHandler."),
+                                    tool_name_);
     }
 
     // check if tool is in official tools list
     if (official_ && tool_name_ != "GenericWrapper" && !ToolHandler::getTOPPToolList().count(tool_name_))
     {
-      writeLog_(String("Warning: Message to maintainer - If '") + tool_name_ + "' is an official TOPP tool, add it to the tools list in ToolHandler. If it is not, set the 'official' flag of the TOPPBase constructor to false.");
+      throw Exception::InvalidValue(__FILE__,
+                                    __LINE__,
+                                    OPENMS_PRETTY_FUNCTION,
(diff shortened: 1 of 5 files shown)


timosachsenberg (approved):

timosachsenberg:
nice. thanks!

jpfeuffer:
One thing missing would be a warning when running the tool (i.e. a check in TOPPBase).
Yes, it introduced a dependency on Toolhandler (if not already present) but might be worth it. Often developers do not build the documentation and then it just is found in the nightlies.

jpfeuffer:
https://github.com/OpenMS/OpenMS/blob/591cc3f1f130374b5aca12f38cd4c7b1fbf65508/src/openms/source/APPLICATIONS/TOPPBase.cpp#L39
We have a dependency already.

timosachsenberg:
ah ok you mean in addition write out a warning to OPENMS_LOG_WARN ?

jpfeuffer:
Maybe add a check for presence in either of the lists.

timosachsenberg:
but this check is already in there
https://github.com/OpenMS/OpenMS/pull/5032/files#diff-38dd76e7d29ade2c6a692b72f0ba8e04a92b6f63d118a6f930a9f1088727e347R142
https://github.com/OpenMS/OpenMS/pull/5032/files#diff-38dd76e7d29ade2c6a692b72f0ba8e04a92b6f63d118a6f930a9f1088727e347R148

cbielow:
I'd personally make this check a hard error.
Maybe this can even be done as `constexpr` compile warning once we are in C++17 land

oliveralka:
Ok, so we could throw an exception (e.g. InvalidValue), instead of the `writeLog_`?

cbielow:
yes

jpfeuffer:
? sorry I overlooked that completely.

oliveralka:
Since the "Tools/Utils" used in the TOPPBase_test are not registered in the ToolHandler - the tests fail. 

```
Message: the value 'TOPPBaseTest' was used but is not valid; If 'TOPPBaseTest' is a Util, add it to the util list in ToolHandler. If it is not, set the 'official' flag of the TOPPBase constructor to true.
515:  failed
```

I am actually not sure what to do here? Of course, we could add them to the ToolHandler, but they are not actually tools, or?

timosachsenberg:
hmm the easiest solution might be adding a protected member: 
bool disable_toolhandler_test_ = false; 
to TOPPBase and set it in these tests to true (which you can query and skip the if / throwing the exception).

timosachsenberg:
I would not add it to ToolHandler as we query those e.g., to get the list of executables etc.

src/openms/source/APPLICATIONS/TOPPBase.cpp
```diff
@@ -132,34 +133,37 @@ namespace OpenMS
       verboseVersion_ += String(", Revision: ") + VersionInfo::getRevision() + "";
     }
 
-    // check if tool entries are in Utils and TOPP (duplication)
-    if (ToolHandler::checkDuplicated(tool_name_))
+    if (toolhandler_test_)
```
timosachsenberg, line 136: nice but please add a comment why this can be disabled

timosachsenberg (approved):

timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=67203

PR #5041, opened 2020-10-31
[FIX] rare underflow error in OpenPepXLAlgorithm

enetz:
Fixes a rare underflow error in OpenPepXL.
Same as  #5040, but for the algorithm for isotopically labeled cross-linkers.

# Description

# Checklist:

### DIFF
--- a/src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
+++ b/src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
@@ -555,9 +555,11 @@ using namespace OpenMS;
           OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentSimple(matched_spec_xlinks_alpha, fragment_mass_tolerance_xlinks_, fragment_mass_tolerance_unit_ppm_, theoretical_spec_xlinks_alpha, xlink_peaks, exp_charges);
           OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentSimple(matched_spec_xlinks_beta, fragment_mass_tolerance_xlinks_, fragment_mass_tolerance_unit_ppm_, theoretical_spec_xlinks_beta, xlink_peaks, exp_charges);
         }
-        // maximal xlink ion charge = (Precursor charge - 1), minimal xlink ion charge: 2
+        // maximal xlink ion charge = (Precursor charge - 1), minimal xlink ion charge: 2, we need the difference between min and max here
         Size n_xlink_charges = (precursor_charge - 1) - 2;
-        if (n_xlink_charges < 1) n_xlink_charges = 1;
+        // check for reasonable numbers. In case precursor_charge is lower than 4, n_xlink_charges should always be 1
+        // it should never have a number higher than 7 or 8. If it's > 20, that is because of an underflow
+        if (n_xlink_charges < 1 | n_xlink_charges > 20) n_xlink_charges = 1;
 
         // compute match odds (unweighted), the 3 is the number of charge states in the theoretical spectra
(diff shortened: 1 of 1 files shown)


src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
```diff
@@ -555,9 +555,11 @@ using namespace OpenMS;
           OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentSimple(matched_spec_xlinks_alpha, fragment_mass_tolerance_xlinks_, fragment_mass_tolerance_unit_ppm_, theoretical_spec_xlinks_alpha, xlink_peaks, exp_charges);
           OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentSimple(matched_spec_xlinks_beta, fragment_mass_tolerance_xlinks_, fragment_mass_tolerance_unit_ppm_, theoretical_spec_xlinks_beta, xlink_peaks, exp_charges);
         }
-        // maximal xlink ion charge = (Precursor charge - 1), minimal xlink ion charge: 2
+        // maximal xlink ion charge = (Precursor charge - 1), minimal xlink ion charge: 2, we need the difference between min and max here
         Size n_xlink_charges = (precursor_charge - 1) - 2;
-        if (n_xlink_charges < 1) n_xlink_charges = 1;
+        // check for reasonable numbers. In case precursor_charge is lower than 4, n_xlink_charges should always be 1
+        // it should never have a number higher than 7 or 8. If it's > 20, that is because of an underflow
+        if (n_xlink_charges < 1 | n_xlink_charges > 20) n_xlink_charges = 1;
```
timosachsenberg, line 562: this test for > 20 is a bit hacky. Could you use an int? or just something like std::max(2, pc-3)

enetz: It was an int at some point before. I had to change it to Size because of another issue.
I could change the order here: use the formula only if the precursor is larger than 3, and directly set it to 1 otherwise.

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
```diff
@@ -555,11 +555,13 @@ using namespace OpenMS;
           OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentSimple(matched_spec_xlinks_alpha, fragment_mass_tolerance_xlinks_, fragment_mass_tolerance_unit_ppm_, theoretical_spec_xlinks_alpha, xlink_peaks, exp_charges);
           OPXLSpectrumProcessingAlgorithms::getSpectrumAlignmentSimple(matched_spec_xlinks_beta, fragment_mass_tolerance_xlinks_, fragment_mass_tolerance_unit_ppm_, theoretical_spec_xlinks_beta, xlink_peaks, exp_charges);
         }
+
         // maximal xlink ion charge = (Precursor charge - 1), minimal xlink ion charge: 2, we need the difference between min and max here
```
timosachsenberg, line 559: please adapt comment to match the precursor_charge - 3

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
```diff
@@ -779,10 +781,11 @@ using namespace OpenMS;
           double wTICold = XQuestScores::weightedTICScoreXQuest(alpha.size(), beta.size(), intsum_alpha, intsum_beta, total_current, type_is_cross_link);
 
           // maximal xlink ion charge = (Precursor charge - 1), minimal xlink ion charge: 2, we need the difference between min and max here
```
timosachsenberg, line 783: here as well and other PR

timosachsenberg (approved):

PR #5042, opened 2020-10-31
[FEATURE, PARAM] updated default parameters of OpenPepXL

enetz:
OpenPepXL default parameters set to a very sensitive search.
Also added default modifications.

# Description

# Checklist:


timosachsenberg:
looks good

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
```diff
@@ -74,28 +74,28 @@ using namespace OpenMS;
...
+    defaults_.setValue("precursor:min_charge", 2, "Minimum precursor charge to be considered.");
+    defaults_.setValue("precursor:max_charge", 8, "Maximum precursor charge to be considered.");
+    defaults_.setValue("precursor:corrections", ListUtils::create<int>("4, 3, 2, 1, 0"), "Monoisotopic peak correction. Matches candidates for possible monoisotopic precursor peaks for experimental mass m and given numbers n at masses (m - n * (C13-C12)). These should be ordered from more extreme to less extreme corrections. Numbers later in the list will be preferred in case of ambiguities.");
     defaults_.setSectionDescription("precursor", "Precursor filtering settings");
 
-    defaults_.setValue("fragment:mass_tolerance", 0.2, "Fragment mass tolerance");
-    defaults_.setValue("fragment:mass_tolerance_xlinks", 0.3, "Fragment mass tolerance for cross-link ions");
-    defaults_.setValue("fragment:mass_tolerance_unit", "Da", "Unit of fragment m");
+    defaults_.setValue("fragment:mass_tolerance", 20.0, "Fragment mass tolerance");
+    defaults_.setValue("fragment:mass_tolerance_xlinks", 20.0, "Fragment mass tolerance for cross-link ions");
```
cbielow, line 83: what is the unit of this?
does `fragment:mass_tolerance_unit` apply to both? (would be a bit unususal)

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
```diff
@@ -74,28 +74,28 @@ using namespace OpenMS;
...
-    defaults_.setValue("precursor:min_charge", 3, "Minimum precursor charge to be considered.");
-    defaults_.setValue("precursor:max_charge", 7, "Maximum precursor charge to be considered.");
-    defaults_.setValue("precursor:corrections", ListUtils::create<int>("2, 1, 0"), "Monoisotopic peak correction. Matches candidates for possible monoisotopic precursor peaks for experimental mass m and given numbers n at masses (m - n * (C13-C12)). These should be ordered from more extreme to less extreme corrections. Numbers later in the list will be preferred in case of ambiguities.");
+    defaults_.setValue("precursor:min_charge", 2, "Minimum precursor charge to be considered.");
+    defaults_.setValue("precursor:max_charge", 8, "Maximum precursor charge to be considered.");
+    defaults_.setValue("precursor:corrections", ListUtils::create<int>("4, 3, 2, 1, 0"), "Monoisotopic peak correction. Matches candidates for possible monoisotopic precursor peaks for experimental mass m and given numbers n at masses (m - n * (C13-C12)). These should be ordered from more extreme to less extreme corrections. Numbers later in the list will be preferred in case of ambiguities.");
```
cbielow, line 79: just a note: you can do `{4, 3, 2, 1, 0}`

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
enetz: @cbielow yes, the unit applies to both. Both tolerances are for fragments in the same spectrum, so should be very similar anyway. There are two tolerances mainly for older ion trap data, where people often use absolute numbers of 0.2 for most fragments and 0.3 is sometimes used for higher masses and cross-linked fragments are on the heavier side. When using relative tolerances, they should be the same.

src/openms/source/ANALYSIS/XLMS/OpenPepXLAlgorithm.cpp
timosachsenberg: yep, getting rid of ListUtils::create<int> now works with modern C++

timosachsenberg (approved):

PR #5049, opened 2020-11-03
Refactor TableViews in TOPPView

cbielow:
in prep for OSW data

 - uses named columns for more structured code
 - refactors common table operations into new TableView class
 - fixes some bugs, e.g.
    - referencing wrong columns due to index errors
    - the first click on a row would not yield any action (because there was no previous row)


src/openms_gui/source/VISUAL/SpectraIdentificationViewWidget.cpp
```diff
@@ -471,11 +407,12 @@ namespace OpenMS
 
     // generate flat list
     int selected_row(-1);
+    // index i is needed, so iterate the old way...
     for (Size i = 0; i < layer_->getPeakData()->size(); ++i)
     {
       QTableWidgetItem* item = nullptr;
 
-      const MSSpectrum & spectrum = (*layer_->getPeakData())[i];
+      const MSSpectrum& spectrum = (*layer_->getPeakData())[i];
```
timosachsenberg, line 376: is this safe? you head the lambda last time

src/openms_gui/source/VISUAL/SpectraIdentificationViewWidget.cpp
```diff
@@ -488,202 +386,60 @@ namespace OpenMS
...
-      if (id_count == 0)
-      {
-        c = Qt::white; // no identification
-      }
-      else
-      {
-        c = Qt::green; // identification
-      }
+      // set row background color
+      QColor bg_color = (id_count == 0 ? bg_color = Qt::white : bg_color = Qt::green);
```
jpfeuffer, line 390: Are the assignments in the cond. stmts needed?

src/openms_gui/source/VISUAL/SpectraIdentificationViewWidget.cpp
```diff
@@ -1059,65 +585,68 @@ namespace OpenMS
...
-    }
-
     // extract position of the correct Spectrum, PeptideIdentification and PeptideHit from the table
-    int r = item->row();
-    bool selected = item->checkState() == Qt::Checked;
-    int spectrum_index = table_widget_->item(r, 1)->data(Qt::DisplayRole).toInt();
-    int num_id = table_widget_->item(r, id_col)->text().toInt();
-    int num_ph = table_widget_->item(r, ph_col)->text().toInt();
+    int row = item->row();
+    String selected = item->checkState() == Qt::Checked ? "true" : "false";
```
jpfeuffer, line 599: Why String now? And shouldn't this be a property of the row and not the PeptideHit associated to that row? What if the rows represent sth else?

cbielow: it was string before. I just removed the intermediate step.
The row represents a single PH and it cannot represent sth else (except when we show unidentified MS2 spectra, but they do not have an editable checkbox, so this functions never gets called for those rows).

src/openms_gui/source/VISUAL/SpectraIdentificationViewWidget.cpp
cbielow: no. good catch. this only works because the assignment returns a reference. But its weird, so I'll remove it.

src/openms_gui/source/VISUAL/SpectraIdentificationViewWidget.cpp
cbielow: yes, its safe, because the MS data is not modified by the table

timosachsenberg (approved):
looks good

PR #5051, opened 2020-11-04
Tv dia table

cbielow:
# Description

Some fixes, mainly for the spectra tree view widget:
 - avoid useless conversion of numbers to qstring. Name roles properly.
 - fix bug, which caused the LayerMenu not to refresh when clicking a single chrom transition in SpectraView
 - get rid of unneeded and dangerous exposure to internal data
 - allow drag'n'drop of single chromatogram data from SpecView table to new Window
 - properly populate columns via enum indices (fixes wrong layout)
spec_view_chrom

### DIFF
--- a/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
+++ b/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
@@ -2494,19 +2494,15 @@ namespace OpenMS
       }
       else if (spec_view != nullptr)
       {
-        QTreeWidgetItem* item = spec_view->getTreeWidget()->currentItem();
-        if (item != nullptr)
+        ExperimentSharedPtrType new_exp_sptr(new ExperimentType());
+        if (spec_view->getSelectedScan(*new_exp_sptr))
         {
-          const LayerData& layer = getActiveCanvas()->getCurrentLayer();
-          Size index = (Size)(item->text(3).toInt());       // todo: wtf...
-          const ExperimentType::SpectrumType spectrum = (*layer.getPeakData())[index];
-          ExperimentSharedPtrType new_exp_sptr(new ExperimentType());
-          new_exp_sptr->addSpectrum(spectrum);
           ODExperimentSharedPtrType od_dummy(new OnDiscMSExperiment());
           FeatureMapSharedPtrType f_dummy(new FeatureMapType());
           ConsensusMapSharedPtrType c_dummy(new ConsensusMapType());
           vector<PeptideIdentification> p_dummy;
-          addData(f_dummy, c_dummy, p_dummy, new_exp_sptr, od_dummy, LayerData::DT_CHROMATOGRAM, false, false, true, layer.filename, layer.getName(), new_id);
+          const LayerData& layer = getActiveCanvas()->getCurrentLayer();
+          addData(f_dummy, c_dummy, p_dummy, new_exp_sptr, od_dummy, new_exp_sptr->getNrSpectra() > 0 ? LayerData::DT_PEAK : LayerData::DT_CHROMATOGRAM, false, false, true, layer.filename, layer.getName(), new_id);
         }
       }
       else if (source == nullptr)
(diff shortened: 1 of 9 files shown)


timosachsenberg:
Nice. Is it intended to be put in the "Scans" tab? Struggling with this abstraction and would probably have put it in a dedicated tab (similar how I do in the Top-Down-View) but that is open for discussion.

cbielow:
> 
> 
> Nice. Is it intended to be put in the "Scans" tab? 

No, definitely in a new tab! I'm just cleaning up the existing ones, because the new tab will probably make use of components existing ones. Even the currently merged functionality for Peak AND Chrom data in one tab is debatable... its troublesome at some points, but probably saves code in other places...

src/openms_gui/source/VISUAL/SpectraViewWidget.cpp
```diff
@@ -444,87 +479,75 @@ namespace OpenMS
       std::map<Precursor, std::vector<Size>, Precursor::MZLess>& map_precursor_to_chrom_idx = map_precursor_to_chrom_idx_cache_[(size_t)(exp.get())];
       if (!was_cached)
       { // create cache: collect all precursor that fall into the mz rt window
-        for (std::vector<MSChromatogram >::const_iterator iter = exp->getChromatograms().begin(); iter != exp->getChromatograms().end(); ++iter)
+        for (auto it = exp->getChromatograms().begin(); it != exp->getChromatograms().end(); ++it)
```
jpfeuffer, line 482: cbegin and cend? or did you change access patterns?

jpfeuffer (approved):

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -2494,19 +2494,15 @@ namespace OpenMS
       }
       else if (spec_view != nullptr)
       {
-        QTreeWidgetItem* item = spec_view->getTreeWidget()->currentItem();
-        if (item != nullptr)
+        ExperimentSharedPtrType new_exp_sptr(new ExperimentType());
+        if (spec_view->getSelectedScan(*new_exp_sptr))
```
timosachsenberg, line 2498: e.g., here copying the whole layer might be a more general solution

src/openms_gui/source/VISUAL/SpectraViewWidget.cpp
```diff
@@ -55,11 +54,68 @@ namespace OpenMS
     return out;
   }
 
+
+  QList<QVariant> vecToList(const std::vector<Size>& in)
```
timosachsenberg, line 58: this looks familiar... did you implement that somewhere else already?

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
cbielow: jep, I agree. But currently we have no means to subselect from a full dataset. Or did I miss something?

src/openms_gui/source/VISUAL/SpectraViewWidget.cpp
cbielow: the reverse operation. Just above :) ?

src/openms_gui/source/VISUAL/SpectraViewWidget.cpp
cbielow: good catch

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
timosachsenberg: Do we need to? If so, I think this should then work differently (e.g., letting the user select spectra via check boxes (like curated in id view and export them). We have the file filter tool.

timosachsenberg: I mean we can keep it but in my view a more general solution would be to copy the full layer and layer settings (aka current spectrum index)

PR #5054, opened 2020-11-05
Rename Table/Tree widgets

cbielow:
# Description

renamed View tabs
 - SpectraIdentificationViewWidget --> SpectraIDViewTab
 - SpectraViewWidget --> SpectraTreeTab


timosachsenberg:
feel free to merge if it passes

timosachsenberg (approved):

timosachsenberg:
Should we call this ...Tab ? I know it is used there but in principle it should be independent of the container it is placed in

cbielow:
> 
> 
> Should we call this ...Tab ? I know it is used there but in principle it should be independent of the container it is placed in

Assume it stands for tabular data, which is pretty close :)

timosachsenberg:
hmm Q**Tab**Widget vs Q**Table**Widget
but yeah I already approved ;)

PR #5063, opened 2020-11-12
keep decoys

timosachsenberg:
# Description

# Checklist:


timosachsenberg:
checked:
- feature finder skips decoys: check
- mztab export writes decoys: ?

timosachsenberg:
rebuild jenkins

timosachsenberg:
@jpfeuffer I would probably need some help for the last step of copying the decoys +/- non-quantified ids back into the consensusmap (and how this plays with the filtering of groups etc.).

jpfeuffer:
What do you mean with "copy back"?
Wouldn't it be better if you just load them from the beginning and just ignore them during alignment/linking?

timosachsenberg:
If that is possible it would make a lot of things simpler. Requires more memory though.

jpfeuffer (approved):
Looks good.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -472,9 +472,10 @@ namespace OpenMS
     }
 
     // add back peptide identification id if it got skipped (decoys)
-    for (auto p : peptides)
+    for (const auto& pep : peptides)
```
jpfeuffer, line 475: You can also flag the copy-for-loop check. Not sure if it makes a difference here now though.

PR #5071, opened 2020-11-19
Try to use the official Xerces findModule

jpfeuffer:
Then we can also use the imported target Xercesc::xercesc
And less maintenance.
Let's hope it works for 3.2 (only 3.1 is specified but the code seems general).


jpfeuffer:
Doesn't work. The official module is broken. No NAMES_PER_DIR

PR #5074, opened 2020-11-20
count and store how often a position in a protein was modified with a…

timosachsenberg:
… specific modification

# Description

# Checklist:


timosachsenberg:
would allow for some simple localization statistics

timosachsenberg:
For discussion:
how should we store modification / localization results at the protein level? 
In general think it might be a good idea to provide localization result support with core OpenMS data structures (in this PR it is just hacked in as a map).
E.g. we could provide a class ProteinLocalizationResult or ProteinModificationSummary that stores modification, position as key and spectral count, localization frequency/probability as value + meta values. This might cover most use cases I know of.

src/openms/include/OpenMS/METADATA/ProteinModificationSummary.h
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * @brief Maps a protein position to all observed modifications and associated statistics 
+   * 
+   * For example, to store that position 10 maps to Oxidation (M) which was observed in 123 PSMs.
+   * 
+   */
+  struct OPENMS_DLLAPI ProteinModificationSummary
+  {
+    /// basic modification statistic
+    struct OPENMS_DLLAPI Statistics : public MetaInfoInterface
```
cbielow, line 53: do you really really need the base class here? It makes for a hidden interface which noones knows what is actually stored in there...

timosachsenberg: I guess most may want to store false localization rates, probabilities etc.
what would be a better option? storing a vector of cv params as a member?

cbielow (commented):
can you add a unit test?

src/openms/include/OpenMS/METADATA/ProteinModificationSummary.h
cbielow: explicit members with 'NaN' defaults.
I think memory consumption should not be a problem here, given that the number of proteins is not too large to pose a problem.

timosachsenberg: One problem I see is that storing those extra "scores" in a file might get more difficult. Wouldn't we need a new format every time we add a new score here?

cbielow: how you store that in a file is a different matter :)

hendrikweisser:
Timo, have you thought about how this could fit into the new ID data structures? The new class corresponding to `ProteinHit` is `ParentSequence`, but it's kept deliberately agnostic to the type of "sequence" (protein or RNA). Right now I can think of two options for integration:
1. Analogous to your current implementation, add a member variable to `ParentSequence`, but use strings for the modifications (rather than `ResidueModification*`). Maybe not ideal because you lose static typing and need to translate to/from modification names.
2. Make a more general class `ParentModificationSummary` that takes `ResidueModification*` or `Ribonucleotide*` as a template parameter. Then add maps (maybe `protein_mod_summary` and `rna_mod_summary`) to `IdentificationData` that connect a `ParentSequenceRef` to a `ParentModificationSummary<...>`. Having the mod. summaries in separate places (variables) may actually be preferable because I assume only very few tools/algorithms would have to deal with them.

hendrikweisser:
Another thought: For option 1, a `variant` could be used instead of a string to store the mod. May still be a bit cumbersome.

timosachsenberg:
Without looking into detail option 2 sounds like something that could work 

@jpfeuffer what do you think? Our use case might be something like:
- n identifications on n ID files.
- one joint protein inference for all n identification runs 
- filtering of PSMs that did not pass a protein FDR
- calculation of localization summary on the inferred proteins at the run level AND maybe also at the level of several files (merged localization)

timosachsenberg:
Another thing that we might need/should add to the new ID data structure is a way to store custom modifications.
Otherwise, we can't transport the modifications (e.g., from an open search or a search that defines custom mods (e.g., glycans, whatever)) through to visualization.

timosachsenberg:
What would we need to store in the ParentModificationSummary?
e.g. to link to the Identification/localization run?

hendrikweisser:
> Another thing that we might need/should add to the new ID data structure is a way to store custom modifications.

Can that be delegated to `AASequence`/`NASequence` (or `ResidueModification*`/`Ribonucleotide*` for the "parents")? I think it should be handled at a fundamental level in the library, so that most code doesn't have to care whether a modification was pre-defined or custom.

> What would we need to store in the ParentModificationSummary?
> e.g. to link to the Identification/localization run?

I think only you can really decide that, based on what your needs are. If you keep track of the PMSs (`InputMatch`) as evidence, those would already link back to the identification run(s) that found them. However, if your mod. summary depends on a particular protein inference run (`ParentGroupSet`), you'd need to link to that. We would have to define how that works, because nothing else references `ParentGroupSet` currently, so there's no `...Ref` type for that yet.

jpfeuffer:
@timosachsenberg Isn't that roughly what we discussed today?

timosachsenberg:
Yes it is pretty much in line.

> Can that be delegated to AASequence/NASequence (or ResidueModification*/Ribonucleotide* for the "parents")? I think it should be handled at a fundamental level in the library, so that most code doesn't have to care whether a modification was pre-defined or custom.

I think it can be delegated but this means we need to encode custom mods in the AA/NA string itself.

> I think only you can really decide that, based on what your needs are. If you keep track of the PMSs (InputMatch) as evidence, those would already link back to the identification run(s) that found them. However, if your mod. summary depends on a particular protein inference run (ParentGroupSet), you'd need to link to that. We would have to define how that works, because nothing else references ParentGroupSet currently, so there's no ...Ref type for that yet.

Not an expert but there might be three main use cases:
1. output by localization tool (e.g., PTMProphet, Luciphor) that may report false localization rates at the PSM level
2. aggregation on the protein level (e.g., phosphosite occupancy, total number of spectral counts supporting that modification at that amino acid)
3. aggregation over multiple files (e.g., mean frequency this has been observer over biological replicates)

PR #5075, opened 2020-11-20
TOPPView DIA data browsing

cbielow:
# Description

- extend VerticalLineItem for 1D plot to show a band
- move data conversion for chomatogram to spectrum for 1D widget to a better place
- reduce redundant code (this is still not ideal)
 - show the correct menu entries when chromatogram data is present
 - enable double/single clicking of rows in DIA view
 - show the transitions with bands marking the feature-groups
 - faster sqMass reading (7%)
 - allow Annotation1DVerticalItem to be a band (and actually use its color member)
 - annotate OSW features using a band
 - fix bug (peptides get repeatedly added when browing data)
 - fix: map native IDs to chrom indices 

# Checklist:

### DIFF
--- a/src/openms_gui/include/OpenMS/VISUAL/MISC/GUIHelpers.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/MISC/GUIHelpers.h
@@ -39,32 +39,42 @@
 
 class QString; // declare this OUTSIDE of namespace OpenMS!
 class QStringList;
+class QPainter;
+class QPoint;
+
+#include <QColor>
+#include <QFont>
 
 namespace OpenMS
 {
   /**
-    @brief Class which holds static GUI-related helper functions.
-
-    Since all methods are static, the c'tor is private.
-    
-    @ingroup Visual
+    Namespace which holds static GUI-related helper functions.
   */
-  class OPENMS_GUI_DLLAPI GUIHelpers
+  namespace GUIHelpers
   {
-  public:
-
-    GUIHelpers() = delete;
     
     /// Open a folder in file explorer
     /// Will show a message box on failure
-    static void openFolder(const QString& folder);
+    void openFolder(const QString& folder);
 
     /// Open TOPPView (e.g. from within TOPPAS)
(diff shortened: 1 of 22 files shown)


src/openms_gui/include/OpenMS/VISUAL/MISC/GUIHelpers.h
```diff
@@ -39,32 +39,42 @@
...
+    /**
+       @brief draw a multi-line text at coordinates XY using a specific font and color
+       @param painter Where to draw
+       @param text Each item is a new line
+       @param where Coordinates where to start drawing (upper left corner of text)
+       @param col_fg Optional text color; if invalid (=default) will use the current painter's color
+       @param col_bg Optional background color of bounding rectangle; if invalid (=default) no background will be painted
+       @param Optional font; will use Courier by default
+    */
+    void drawText(QPainter& painter, const QStringList& text, const QPoint& where, const QColor col_fg = QColor("invalid"), const QColor col_bg = QColor("invalid"), QFont f = QFont("Courier"));
```
timosachsenberg, line 76: maybe not ideal place but better than before.
can this be static?

timosachsenberg (approved):

src/openms_gui/include/OpenMS/VISUAL/MISC/GUIHelpers.h
cbielow: I changed the class to a namespace. So its a global function (hence 'static') :)

timosachsenberg:
Just out of interest. In my view having static next to the method name is easier to spot than a namespace. Are there other advantages of the global function in namespace?

cbielow:
> 
> 
> Just out of interest. In my view having static next to the method name is easier to spot than a namespace. Are there other advantages of the global function in namespace?

Since the class GUIHelpers has no data members and no state, there did not seem to be any point in having it around. This also avoids acccidental instanciation of a class which is essentially an empty hull. But thats minor IMHO.
So its mostly a matter of taste, I think.

PR #5079, opened 2020-11-24
[FIX] Correctly sync greedy resolution in PLFQ with quant data

jpfeuffer:
# Description

Correctly sync greedy resolution in PLFQ with quant data.
The test data did not cover this. So I think it will not fail. We need to think about a test.

# Checklist:


timosachsenberg (approved):
let's merge as a hot fix

PR #5081, opened 2020-11-26
[FIX] Correctly sync greedy resolution in PLFQ with quant data

jpfeuffer:
# Description

Fixes a case where a peptide was completely filtered out by FDR. +refactor


timosachsenberg:
could need some comments what/why is this done

timosachsenberg (approved):

PR #5084, opened 2020-11-28
[FEATURE] Better debug messages and values.

jpfeuffer:
# Description

Provides a little bit easier debugging in PLFQ.


src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -1129,6 +1129,10 @@ namespace OpenMS
       }
       chromatogram.setMetaValue("product_mz", transition->getProductMZ());
       chromatogram.setMetaValue("precursor_mz", transition->getPrecursorMZ());
+      Precursor prec; prec.setPosition(transition->getPrecursorMZ());
```
jpfeuffer, line 1132: I wonder if that breaks the SWATH tests. But It looks like the metaValues are legacy anyways. Some debug messages indirectly depend on the Precursor being set.

jpfeuffer: Ok have to check if the failing tests are just cosmetic.

jpfeuffer:
How do we store UInt64 in Metavalues? Can we? @timosachsenberg 
See ProteomicsLFQ3 failing test.

jpfeuffer:
rebuild jenkins

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
jpfeuffer: @cbielow @hroest Any comments on this. Feels like it should be okay. Tests are running. Are the Precursor and Product fields the "new" alternative to the metavalues?

jpfeuffer:
@cbielow @oliveralka I am interested in seeing, where certain Features, Transitions, Chromatogram are getting filtered out due to bad fits, missing apices, bad alignment quality.
Do you have any inputs on where to put additional debug statements to find this out?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -665,7 +667,12 @@ namespace OpenMS
             // to not be combined
             seedcount++;
             //cout << peptide.sequence << " " << charge << endl;
-            String peptide_id = peptide.sequence + String(seedcount) + "/" + String(charge);
+            double mz = rt_pep.second->getMZ();
+            double rt = rt_pep.second->getRT();
+            String uid = rt_pep.second->getMetaValue("SeedFeatureID");
+
+            // UID should be enough, but let's add the seed count to be sure.
+            String peptide_id = peptide.sequence + "[" + uid + "][" + String(seedcount) + "]/" + String(charge);
```
cbielow, line 675: why not directly on `peptide.id = ...` ?

cbielow (approved):
looks good

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
timosachsenberg, line 1132: looks a bit strange to have both expressions in one line. below too

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -336,6 +336,8 @@ namespace OpenMS
         peptides.back().setRT(f_it->getRT());
         peptides.back().setMZ(f_it->getMZ());
         peptides.back().setMetaValue("FFId_category", "internal");
+        peptides.back().setMetaValue("FromUntargeted", "true");
```
timosachsenberg, line 339: do we need this? if there is a seed we should know that it is form untargeted?

timosachsenberg (approved):

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
jpfeuffer: no but then I have to redo all tests

timosachsenberg: Also, we never set FromUntargeted to "false" for the other features.
Maybe I am picky here but like this it doesn't really add value but just needs extra memory.

timosachsenberg: I think we need a script that updates all failed tests ;)

timosachsenberg: ok I will update them

PR #5085, opened 2020-11-30
TOPPView convenience (commandline, locking) and DIA plotting

cbielow:
# Description

- add nicer plotting to sqMass + OSW id view
image
    - this includes a basic overlap detection algorithm which avoids overlapping text blocks
- lock TOPPView GUI when large data is loaded, so the user gets feedback why nothing is responding
- automatically select the appropriate ID tab, depending on what data is available
- support ID annotation from commandline via the new `!` operator, e.g. `TOPPView BSA.mzML ! BSA.idXML` (or for any other annotation data we support, e.g. sqMass + osw)

### DIFF
--- a/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
+++ b/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
@@ -1895,7 +1895,7 @@ namespace OpenMS
   void TOPPViewBase::annotateWithAMS()
   { // this should only be callable if current layer's type is of type DT_PEAK
     LayerData& layer = getActiveCanvas()->getCurrentLayer();
-    LayerAnnotatorAMS annotator;
+    LayerAnnotatorAMS annotator(this);
     assert(log_ != nullptr);
     if (!annotator.annotate(layer, *log_, current_path_))
     {
@@ -1907,7 +1907,7 @@ namespace OpenMS
   void TOPPViewBase::annotateWithID()
   { // this should only be callable if current layer's type is one of DT_PEAK, DT_FEATURE, DT_CONSENSUS
     LayerData& layer = getActiveCanvas()->getCurrentLayer();
-    LayerAnnotatorPeptideID annotator;
+    LayerAnnotatorPeptideID annotator(this);
     assert(log_ != nullptr);
     if (!annotator.annotate(layer, *log_, current_path_))
     {
@@ -1919,7 +1919,7 @@ namespace OpenMS
   void TOPPViewBase::annotateWithOSW()
   { // this should only be callable if current layer's type is of type DT_CHROMATOGRAM
     LayerData& layer = getActiveCanvas()->getCurrentLayer();
-    LayerAnnotatorOSW annotator;
+    LayerAnnotatorOSW annotator(this);
     assert(log_ != nullptr);
     if (!annotator.annotate(layer, *log_, current_path_))
     {
--- a/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
+++ b/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
@@ -522,9 +522,6 @@ namespace OpenMS
 
   void TOPPViewBase::addDataFile(const String& filename, bool show_options, bool add_to_recent, String caption, UInt window_id, Size spectrum_id)
   {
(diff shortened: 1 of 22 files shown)


timosachsenberg:
Love the "TOPPView BSA.mzML ! BSA.idXML" option.

Would it be a good idea if BSA.idXML/OSW is opened from command line let TOPPView ask if the mzML file (if present in e.g. getPrimaryMSRun) should be also loaded and annotated?

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -1872,6 +1876,16 @@ namespace OpenMS
     return &(canvas->getCurrentLayer());
   }
 
+  LayerData* TOPPViewBase::getCurrentLayer()
```
timosachsenberg, line 1879: I think this TOPPViewBase should not have a function to get the "currentLayer".
The way it is done with getActiveCanvas()->getCurrentLayer() also not ideal.
I think cleaner would be to get the activeWorkspace and try to call the annotation there

src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
```diff
@@ -64,6 +64,7 @@ namespace OpenMS
     spectra_view_widget_(new SpectraTreeTab(this)),
     id_view_widget_(new SpectraIDViewTab(Param(), this)),
     dia_widget_(new DIATreeTab(this)),
+    tab_ptrs_{ spectra_view_widget_, id_view_widget_, dia_widget_ },   // make sure to add new tabs here!
```
timosachsenberg, line 67: Might also be better to have this generated based on the data stored in the workspace / the open window but well this would be quite some effort

src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
```diff
@@ -94,15 +95,24 @@ namespace OpenMS
...
-    setTabEnabled(DIAOSW_IDX, false);
+    // make sure initialization was correct
+    assert(tabBar()->count() == tab_ptrs_.size());
 
     // switch between different view tabs
     connect(this, &QTabWidget::currentChanged, this, &DataSelectionTabs::currentTabChanged);
     connect(this, &QTabWidget::tabBarDoubleClicked, this, &DataSelectionTabs::tabBarDoubleClicked);
   }
 
+  LayerData* getCurrentLayerData(TOPPViewBase* tv)
```
timosachsenberg, line 106: I think it is bad design to go back to the parent here and request the Layer

timosachsenberg: Not your fault though :)

cbielow:
> 
> 
> Love the "TOPPView BSA.mzML ! BSA.idXML" option.
> 
> Would it be a good idea if BSA.idXML/OSW is opened from command line let TOPPView ask if the mzML file (if present in e.g. getPrimaryMSRun) should be also loaded and annotated?

Ahh yes. Great idea! For this to work universally, we first need the common UID for sqMass and osw. This is next on my list. After that, we can tackle this.

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
cbielow: Agreed, this is not ideal. 
This was simply factoring out of some code, which should make it easier to replace in the future.
But one step at a time.

src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
cbielow: you mean across all layers?
This would break a central concept, but maybe its worth thinking about where this would lead.

src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
cbielow: Mhh.. with the way we currently do things, I don't see what we can do about it.
One could rewrite this using signal/slot, but that only makes it less obvious.

timosachsenberg: I think signal/slots would probably be better because it would move the design to a notify/observer model and not to the current "you need to update yourself go pull your data from the god object" model ;)

timosachsenberg (approved):
NIce changes!

PR #5087, opened 2020-12-01
PSMFeatureExtractor add tests

bernt-matthias:
The PR ended up in just adding tests .. which is good I think. All other changes discussed below have been reverted.

The tool has two input modes single file input and
multiple file input. For the later case `-multiple_search_engines`
needs to be added.

These two cases can not be distinguished automatically
for automatic tool building (Galaxy).

One possible solution is to use two inputs and require exactly one of
them. Seems to be the least painful change for Galaxy tool building.
Alternatively I could "manually" influence the tool building
process.

Also adds tests.

TODO

- [ ] remove `multiple_search_engines flag`?
- [ ] are there other tools with a similar problem?

# Checklist:


src/tests/topp/FeatureFinderIdentification_1.tmp
```diff
@@ -0,0 +1,2717 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
timosachsenberg, line 1: should this file be part of the commit? looks like one generated by a test

bernt-matthias:
Would be cool to have this in the next release, or?

bernt-matthias:
Or we leave it as is and also use the subworkflow workaround for Galaxy.

bernt-matthias:
For the most recent iteration of the Galaxy tool wrapper generation I add a conditional for each input that allows multiple input files that allows to switch between batch mode and joint processing mode.

bernt-matthias:
Anyway, we should keep the tests.

bernt-matthias:
@timosachsenberg do I need to cange a source file in order to run the tests?

timosachsenberg:
rebuild jenkins
Triggers the rebuilds and tests online. But there seem to be an issue with our jenkins

timosachsenberg (approved):

timosachsenberg:
There are some errors like:
```
Error: File not found (the file 'PSMFeatureExtractor_1_out.tmp' could not be found)
Cannot read input file given from parameter '-in1'!
```

bernt-matthias:
I could not reproduce the error locally (the tmp files were generated and processed by DIFF). I made some changes to the DIFF whitelists and files .. lets see if it works now.

enetz:
you forgot to set the test dependencies, like:
`set_tests_properties("UTILS_PSMFeatureExtractor_1_out" PROPERTIES DEPENDS "UTILS_PSMFeatureExtractor_1")`

PR #5089, opened 2020-12-02
SQL: common ID for sqMass and OSW

cbielow:
# Description

OpenSwathWorkflow now creates paired OSW and sqMass files (by common RUN.ID). This ensures that you always annotate the correct files.
This is enforced in TOPPView, for example.

Also fixes a bug in SqlConnection (which allows to open a non-existant database for reading without error).


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
```diff
@@ -125,13 +125,14 @@ namespace OpenMS
...
     OpenSwathOSWWriter(const String& output_filename,
+                       const UInt64 run_id,
                        const String& input_filename = "inputfile",
                        bool ms1_scores = false,
                        bool sonar = false,
                        bool uis_scores = false) :
       output_filename_(output_filename),
       input_filename_(input_filename),
-      run_id_(OpenMS::UniqueIdGenerator::getUniqueId()),
+      run_id_(static_cast<int64_t>(run_id & ~(1ULL << 63))),  // conversion from UInt64 to int64_t to support SQLite (and conversion to 63 bits)
```
timosachsenberg, line 135: I think this conversion is used at a few places in sqlite code. can we factor that out?

src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp
```diff
@@ -247,11 +247,11 @@ namespace OpenMS
     // the cost for initialization and copy should be minimal
     //  - a single C string is created
     //  - two ints
-    MzMLSqliteHandler::MzMLSqliteHandler(String filename) :
+    MzMLSqliteHandler::MzMLSqliteHandler(const String& filename, const UInt64 run_id) :
       filename_(filename),
       spec_id_(0),
       chrom_id_(0),
-      run_id_(0),
+      run_id_(static_cast<int64_t>(run_id & ~(1ULL << 63))), // conversion from UInt64 to int64_t to support SQLite (and conversion to 63 bits)
```
timosachsenberg, line 254: like here and maybe code that takes UInt meta values?

src/openms/source/FORMAT/SqMassFile.cpp
```diff
@@ -45,22 +45,22 @@ namespace OpenMS
 
   void SqMassFile::load(const String& filename, MapType& map)
   {
-    OpenMS::Internal::MzMLSqliteHandler sql_mass(filename);
+    OpenMS::Internal::MzMLSqliteHandler sql_mass(filename, 0);
```
timosachsenberg, line 48: 0 = uninitialized id?

timosachsenberg (approved):

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
cbielow: there were only 3 places left, but I refactored into a single function.
While doing that, I researched a bit on WHY this is actually necessary, and the short answer is: because we're doing it wrong.
We INSERT into a db by creating a large string which contains all the data. This is slow and forces us to remove the 'sign'-bit of a 64bit unsigned int because Sqlite cannot store a large 64bit uint into its signed64bit int (it does not have uint64). Thus, it will fall back to store it as double(!), which of course in accurate for large integers.
To do it right we'd need to `bind` variables to prepared statements. This would not only be faster (since we safe the roundtrip via strings), but also allows to store the 64bit integer - it will just show up as negative value in Sqlite, but can be restored without loss if information when reading it back... but to fix that would take days.

src/openms/source/FORMAT/SqMassFile.cpp
cbielow: yeah... somewhat.
The ID is only used for writing at the moment. Not for reading (because you'd need to peek the sqlite file to know that number).

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
timosachsenberg: thanks for looking into that...  maybe interesting for @hroest

PR #5092, opened 2020-12-05
remove XML file and store data locally in ResidueDB

timosachsenberg:
# Description

Removing the XML file from the ResidueDB files to disband the dependency of the XML file from OpenMS.

# Checklist:


timosachsenberg:
replaces https://github.com/OpenMS/OpenMS/pull/5092 (I just moved it to upstream as I then get tests build in gitpod)

src/openms/source/CHEMISTRY/ResidueDB.cpp
```diff
@@ -191,7 +194,7 @@ namespace OpenMS
         }
       }
     }
-    buildResidueName_(r);
+    buildResidueNames_();
```
timosachsenberg: I think it might be possible to use the buildResidueName_(r) which should be much faster.
Maybe just on the non-modified ones? Need to check.

src/openms/include/OpenMS/CHEMISTRY/ResidueDB.h
```diff
@@ -199,6 +194,9 @@ namespace OpenMS
     /// builds an index of residue names for fast access, synonyms are also considered
     void buildResidueNames_();
 
+    /// adds a single residue to the index
+    void buildResidueName_(Residue*);
```
cbielow, line 198: bad interface.
Who owns the Residue?

src/openms/source/CHEMISTRY/Residue.cpp
```diff
@@ -58,30 +58,39 @@ namespace OpenMS
   }
 
   Residue::Residue(const String& name,
-                   const String& three_letter_code,
-                   const String& one_letter_code,
-                   const EmpiricalFormula& formula) :
+            const String& three_letter_code,
+            const String& one_letter_code,
+            EmpiricalFormula formula,
```
cbielow, line 63: you use the formula read-only.
So pass by `const&`

src/openms/source/CHEMISTRY/Residue.cpp
```diff
@@ -58,30 +58,39 @@ namespace OpenMS
...
+            double pka,
+            double pkb,
+            double pkc,
+            double gb_sc,
+            double gb_bb_l,
+            double gb_bb_r,
+            set<String> synonyms):
```
cbielow, line 70: either pass by const& or move into member

src/openms/source/CHEMISTRY/ResidueDB.cpp
```diff
@@ -133,43 +132,42 @@ namespace OpenMS
...
     }     
   }
 
   void ResidueDB::addResidue_(Residue* r)
   {
     vector<String> names;
-    if (r->getName() != "")
+
+    // add name to lookup
+    if (r->getName() != "")
```
cbielow, line 149: a residue without name? should that be allowed?

src/openms/source/CHEMISTRY/ResidueDB.cpp
```diff
@@ -468,36 +388,41 @@ namespace OpenMS
     return rs;
   }
 
-  void ResidueDB::buildResidueNames_()
+  void ResidueDB::buildResidueName_(Residue* r)
```
cbielow, line 391: argument should be `const Residue* const r)`

src/openms/source/CHEMISTRY/ResidueDB.cpp
```diff
@@ -133,43 +132,42 @@ namespace OpenMS
...
+      buildResidues_();
       buildResidueNames_();
     }     
   }
 
   void ResidueDB::addResidue_(Residue* r)
   {
     vector<String> names;
-    if (r->getName() != "")
+
```
cbielow, line 147: there are two different methods `addResidue_` and `insertResidue`, where is the difference?
Can you get rid of one?

cbielow (changes requested):
nice.
But still a bit inconsistent :{

src/openms/include/OpenMS/CHEMISTRY/Residue.h
```diff
@@ -187,12 +187,19 @@ namespace OpenMS
 
     /// Move constructor
     Residue(Residue&&) = default;
-
-    /// Detailed constructor
+           
+    // Detailed constructor
```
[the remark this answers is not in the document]
timosachsenberg: Needed for pyOpenMS afaik

src/openms/include/OpenMS/CHEMISTRY/ResidueDB.h
timosachsenberg: Residues are always owned by ResidueDB

src/openms/source/CHEMISTRY/Residue.cpp
timosachsenberg: Not possible because it has a default

src/openms/source/CHEMISTRY/ResidueDB.cpp
timosachsenberg: I think these are modified residues. Not sure but I think they exist ;)

src/openms/source/CHEMISTRY/Residue.cpp
timosachsenberg: done thanks

timosachsenberg: thanks. done

src/openms/source/CHEMISTRY/ResidueDB.cpp
timosachsenberg: I think it currently needs Residue* to store it in
    std::set<Residue*> residues_;
and
    std::set<const Residue*> const_residues_;

probably because it provides a:
inline ResidueIterator beginResidue() { return residues_.begin(); }
inline ResidueIterator endResidue() { return residues_.end(); }

might be a good idea to see where these are used and get rid of them (including the non-const stored residues)

src/openms/include/OpenMS/CHEMISTRY/ResidueDB.h
cbielow: mhh.. should at least be documented :)

src/openms/source/CHEMISTRY/Residue.cpp
cbielow: the default does not prevent any of those two options.

timosachsenberg:
Good to merge as is? I will provide another PR with some more cleanup.

jpfeuffer:
Do you want to keep the xml for now?

timosachsenberg:
ahh sure. I knew I forgot something. Will remove it.

jpfeuffer (approved):

jpfeuffer:
Please double-check in the nightlies that there are no errors during packaging due to removal of the xml.

PR #5093, opened 2020-12-05
fix: bug in AbsoluteQuantitation

dmccloskey:
# Description

Small fix in `AbsoluteQuantitation`

## Fixes
- change the term that the `dilution_factor` is applied to (i.e., changed to the correct term :))

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,9 @@ src/.DS_Store
 src/openms/.DS_Store
 .vscode
 contrib-build
+contrib_build
 openms-build
+openms_build
 *.swp
 cmake-build-debug
 cmake-build-release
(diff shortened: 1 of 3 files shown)


timosachsenberg:
look good. can you add a small regression test?

dmccloskey:
Definitely.  Can you clarify what you mean by "regression test"?

timosachsenberg:
A test that would fail without your fix and that now produces the correct result.
Probably just a few lines in https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/AbsoluteQuantitation_test.cpp

dmccloskey:
Got it.  I will add that in shortly and ping you when it is ready.

dmccloskey:
Just added.  A bit lazy on my part not adding them in to start with :).

timosachsenberg (approved):

PR #5094, opened 2020-12-06
Fix: Bug in MRMFeatureQCFile

dmccloskey:
# Description
- Small bug in MRMFeatureQCFile where an empty vector in a member of MRMFeatureQC would throw an error

### DIFF
--- a/src/openms/source/FORMAT/MRMFeatureQCFile.cpp
+++ b/src/openms/source/FORMAT/MRMFeatureQCFile.cpp
@@ -211,6 +211,8 @@ namespace OpenMS
   void MRMFeatureQCFile::store(const String & filename, const MRMFeatureQC & mrmfqc, const bool is_component_group)
   {
     if (is_component_group) {
+      if (mrmfqc.component_group_qcs.size() <= 0) return;
+
       // Store the ComponentGroupQCs
       clear(); // clear the buffer_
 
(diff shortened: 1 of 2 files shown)


dmccloskey:
I just realized that this PR is pulling in two commits from a recent PR that was merged.  Let me know if this is a problem for the merging strategy or review and I can cherry pick in a new PR.

timosachsenberg:
no problem. I can try an automated rebase.
/rebase

src/openms/source/FORMAT/MRMFeatureQCFile.cpp
```diff
@@ -211,6 +211,8 @@ namespace OpenMS
   void MRMFeatureQCFile::store(const String & filename, const MRMFeatureQC & mrmfqc, const bool is_component_group)
   {
     if (is_component_group) {
+      if (mrmfqc.component_group_qcs.size() <= 0) return;
```
timosachsenberg, line 214: .empty() is probably better here and below

timosachsenberg (approved):
If you could add a regression test as well it would be great.

src/openms/source/FORMAT/MRMFeatureQCFile.cpp
timosachsenberg: I will add those changes

PR #5097, opened 2020-12-08
[FEATURE] allow subset of exp design to be given to PLFQ

jpfeuffer:
# Description

Allows to use a subset of mzMLs in the experimental design. Of course only if you filter samples or complete fractions. The checks wont allow partial filtering.

# Checklist:


src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -469,10 +470,12 @@ namespace OpenMS
         if (files_per_fraction == 0) // first fraction, initialize
         {
           files_per_fraction = f.second.size();
+          std::cout << f.second.size() << std::endl;
```
timosachsenberg, line 473: Debug code?

src/utils/ProteomicsLFQ.cpp
```diff
@@ -1638,6 +1605,43 @@ class ProteomicsLFQ :
...
+      in_basenames.insert(in_bn);
+    }
+
+    if (!std::includes(ed_basenames.begin(), ed_basenames.end(), in_basenames.begin(), in_basenames.end()))
+    {
+      throw Exception::InvalidParameter(__FILE__, __LINE__,
+        OPENMS_PRETTY_FUNCTION, "Spectra file basenames provided as input need to match a subset the experimental design file basenames.");
+    }
+
+    design.filterByBasenames(in_basenames);
```
timosachsenberg, line 1632: can we check if filtering needs to be performed and in that case output that only a subset is processed?

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -469,10 +470,12 @@ namespace OpenMS
...
         }
         else // fraction >= 2
         {
           // different number of associated MS files?
+          std::cout << f.second.size() << std::endl;
```
timosachsenberg, line 478: here as well

src/utils/ProteomicsLFQ.cpp
jpfeuffer: The LOG INFO in the method is not enough?

timosachsenberg: it is not that helpful if the user forgets to specify a file on the filename

jpfeuffer: done by returning nr of filtered files

timosachsenberg (approved):

jpfeuffer:
In the future we can replace more occurrences of __FILE__ with the new macro. Unfortunately this will lead to/need a bit of recompilation if you move the source folder (but I guess it's rare).

jpfeuffer:
Wait for merge. Made some improvements for logging in FFID.

CMakeLists.txt
```diff
@@ -269,6 +269,11 @@ set(BUILD_SHARED_LIBS true)
 set(OPENMS_HOST_DIRECTORY "${PROJECT_SOURCE_DIR}")
 set(OPENMS_HOST_BINARY_DIRECTORY "${PROJECT_BINARY_DIR}")
 
+# The additional / is important to remove the last character from the path.
+# Note that it does not matter if the OS uses / or \, because we are only
+# saving the path size.
+string(LENGTH "${PROJECT_SOURCE_DIR}/" CF_SOURCE_PATH_SIZE)
```
cbielow, line 275: dangerous.
In some (future) cases, a source file might also be configured and compiled from the build-tree.
If that build tree is in a shorter directory, then you might cut away parts of the filename or worse: point behind the last `\0` and funny things will happen.
So we'd need a check that the build-tree path length is longer, and if its not .. well... then advance only as much as its safe (minimum of both paths). This will look a bit weird in the final result, but better than undefined behaviour.
Maybe issue a warning in the CMake code if this happens..

jpfeuffer: I can also add another macro for the build tree, if you want.

jpfeuffer: I also found some solution with some constexpr magic but most of them assume "/" as the only separator.

cbielow: > 
> 
> I can also add another macro for the build tree, if you want.

Then the user would need to know when to use the source tree vs the build tree macro, right? Not ideal, and just as dangerous?

cbielow: ok, great fix. thanks!
i've added a line of docs

PR #5104, opened 2020-12-11
SwathWizard: support opening results in TOPPView

cbielow:
# Description

in pyProphet & TRIC tab now has another button: `open in TOPPView` which allows to open the result files and browse the data.


timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=82136

timosachsenberg (approved):

PR #5107, opened 2020-12-15
Allow SSEngine in IDPEP

### LINKED ISSUE #5086: SimpleSearchEngine not supported in IDPEP
It is just a matter of adding its internal (uppercase) name and the best transformation of its score(s) here:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp#L944

jpfeuffer:
fixes #5086

### DIFF
--- a/src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
+++ b/src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
@@ -958,9 +958,9 @@ namespace OpenMS
       const double fdr_for_targets_smaller)
     {
       std::set<Int> charges;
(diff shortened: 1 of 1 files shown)


src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
```diff
@@ -941,6 +941,9 @@ namespace OpenMS
       {
         return (-1) * log10(getScore_({"MS:1002257","expect"}, hit, current_score_type));
       }
+      else if (engine == "SIMPLESEARCHENGINE")
```
timosachsenberg, line 944: opening brackets?

timosachsenberg (approved):

PR #5112, opened 2020-12-17
[FEATURE] RT tolerance specific to seeds

timosachsenberg:
# Description

For seeds, the exact apex is known. We thus can lower the RT tolerance window for chromatogram extraction.
As a positive side effect, more seeds can be used for extraction (they don't overlap with known ID anymore).

# Checklist:


jpfeuffer (commented):
As discussed privately, this should make much more sense. The general rt_window for peptides can be up to 500s due to multiple elutions and RT shifts. We lost a lot of seeds due to overlap.
The chromatogram extractor will take care of correcting the beginning and the end if the window is not too short.

hendrikweisser:
I'm working on merging "develop" (with this change) into my "idfile-integration" branch. I wonder whether introducing a parameter specific to seeds is necessary, or if it wouldn't be better to adapt the calculations using the current parameters.

> For seeds, the exact apex is known.

If the seed RT position is expected to be at the feature apex, then the RT extraction window for seeds should just be the max. peak width (which is `2 * peak_width`, i.e. twice the "typical" peak width). The other factors that normally lead to a larger RT extraction window - uncertainty about the position of the ID in the feature (or even closely next to it) and uncertainty from the alignment of internal and external IDs - shouldn't play a role in that case. But if we can still derive the RT extraction window (for seeds) from `peak_width`, then another parameter isn't needed.

jpfeuffer:
I think you are right. I am actually not sure why we did not choose this way @timosachsenberg

timosachsenberg:
hmm I don't recall exactly why we introduced that but maybe it was intended as a hack to compensate for some problems with very narrow elution times (narrow peaks with tailing)? At least I think that gave more freedom to fine tune the extraction. But I don't know for sure and you say makes sense so maybe we just need to revisit that.

jpfeuffer:
Yes valid point but we should think about where to put the check for a "minimum" elution time. Should there be one in the tool, or do we expect users/devs to be smart enough to check for that before passing it. I think when using the tool, most people won't put anything super small. But when using the algorithm and you estimate average peak widths with some algorithm, you could in some corner cases get very small peak widths right?

PR #5113, opened 2020-12-18
query decoy result

timosachsenberg:
# Description

# Checklist:

### DIFF
--- a/src/openms/source/FORMAT/MascotRemoteQuery.cpp
+++ b/src/openms/source/FORMAT/MascotRemoteQuery.cpp
@@ -476,6 +476,7 @@ namespace OpenMS
 
   void MascotRemoteQuery::readResponse(QNetworkReply* reply)
   {
+    static QString dat_file_path_;
 #ifdef MASCOTREMOTEQUERY_DEBUG
     cerr << "\nvoid MascotRemoteQuery::readResponse(const QNetworkReply* reply): ";
     if (reply->error())
@@ -559,7 +560,8 @@ namespace OpenMS
       QRegExp rx("file=(.+/\\d+/\\w+\\.dat)");
       rx.setMinimal(true);
       rx.indexIn(response);
-      search_identifier_ = getSearchIdentifierFromFilePath(rx.cap(1));
+      dat_file_path_ = rx.cap(1);
+      search_identifier_ = getSearchIdentifierFromFilePath(dat_file_path_);
 
       if (param_.exists("skip_export") &&
           (param_.getValue("skip_export") == "true"))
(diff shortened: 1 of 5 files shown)


src/topp/MascotAdapterOnline.cpp
```diff
@@ -257,28 +265,59 @@ class TOPPMascotAdapterOnline :
...
+        {
+          mascot_tmp_file.remove(); // delete file
+        }
       }
-      else
+
+      if (internal_decoys)
       {
-        mascot_tmp_file.remove(); // delete file
+        String mascot_tmp_file_name(File::getTempDirectory() + "/" + File::getUniqueName() + "_Mascot_decoy_response");
```
cbielow, line 296: this section is too much copy-paste.
Use a function?

src/topp/MascotAdapterOnline.cpp
```diff
@@ -257,28 +265,59 @@ class TOPPMascotAdapterOnline :
...
+        mascot_tmp_file.write(mascot_query->getMascotXMLResponse());
+        mascot_tmp_file.close();
+      
+        // set up helper object for looking up spectrum meta data:
+        SpectrumMetaDataLookup lookup;
+        MascotXMLFile::initializeLookup(lookup, exp);
+
+        // read the response
+        vector<PeptideIdentification> pep_ids_decoy;
+        ProteinIdentification prot_id_decoy;
```
cbielow, line 308: ehh.... do you actually use the decoy result after loading it?

src/openms/source/FORMAT/MascotRemoteQuery.cpp
```diff
@@ -579,13 +579,27 @@ namespace OpenMS
 
       // see http://www.matrixscience.com/help/export_help.html for parameter documentation
       String required_params = "&do_export=1&export_format=XML&generate_file=1&group_family=1&peptide_master=1&protein_master=1&search_master=1&show_unassigned=1&show_mods=1&show_header=1&show_params=1&prot_score=1&pep_exp_z=1&pep_score=1&pep_seq=1&pep_homol=1&pep_ident=1&pep_expect=1&pep_var_mod=1&pep_scan_title=1&query_qualifiers=1&query_peaks=1&query_raw=1&query_title=1";
+      // TODO: check that export_params don't contain _show_decoy_report=1. This would add <decoy> at the top of the XML document and we can't easily distinguish the two files (target/decoy) from the search results later.
```
cbielow, line 582: worth mentioning that in the docs of `export_params` flag and maybe actually do the check?
Its only a single line of code.

src/topp/MascotAdapterOnline.cpp
timosachsenberg: No I am even not sure if it works if two different requests are processed async. I was hoping that someone who worked with MAO knows if that could work

timosachsenberg:
I think I found a way. I now export decoys after target results are fully downloaded (doing it asynchronously while the other query was still running (potentially going through several redirects) did not work.

src/openms/source/FORMAT/MascotRemoteQuery.cpp
timosachsenberg: I could add it thou it doesn't help a lot as a user should not mess with the export_params anyway. Basically if you change them in any way (not only related to the decoys) you will likely break the export, so you must know in detail what you are doing.

jpfeuffer (approved):

PR #5117, opened 2020-12-19
Automatically detected all avialable tools in ToolsDialog

dvdvgt:
# Description

This PR adresses issue  ##5117. All available/compatible tools are now being detected automatically instead of being hardcoded for each layer type.

# TODO

Some things still need to be fixed in my opinion but I want to discuss these changes first:

1. Generating an ini file for each tool for each file to determine with which layer types it is compatible with takes a considerable amount of time (~5 seconds). All these IO actions are quite a bottleneck. Maybe there exists a better/faster method?
2. Currently write errors in `getParamFromIni_` are caught but not properly reacted to. In case of an error the param object is still created (line 177, 182).

# Checklist:
- [ ] Make sure that you are listed in the AUTHORS file   **No permission**

### DIFF
--- a/src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
+++ b/src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
@@ -55,6 +55,10 @@
 #include <OpenMS/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.h>
 #include <OpenMS/TRANSFORMATIONS/FEATUREFINDER/FeatureFinder.h>
 
+#include <OpenMS/APPLICATIONS/ToolHandler.h>
+#include <OpenMS/DATASTRUCTURES/Map.h>
+#include <OpenMS/DATASTRUCTURES/Param.h>
+
 using namespace std;
 
 namespace OpenMS
@@ -83,37 +87,26 @@ namespace OpenMS
     main_grid->addWidget(label, 1, 0);
     QStringList list;
 
-    if (layer_type == LayerData::DT_PEAK)
+    // Determine all available tools compatible with the layer_type
+    tool_map_ = {
+            {"*.mzML", LayerData::DataType::DT_PEAK},
+            {"*.mzXML", LayerData::DataType::DT_PEAK},
+            {"*.featureXML", LayerData::DataType::DT_FEATURE},
+            {"*.consensusXML", LayerData::DataType::DT_CONSENSUS},
+            {"*.idXML", LayerData::DataType::DT_IDENT},
+    };
+    // Get a map of all tools
+    const auto& tools = OpenMS::ToolHandler::getTOPPToolList();
+    for (const auto& tool : tools)
     {
-      list << "FileFilter" << "FileInfo"
-           << "NoiseFilterGaussian" << "NoiseFilterSGolay"
-           << "BaselineFilter" << "PeakPickerHiRes"
-           << "PeakPickerWavelet" << "Resampler"
-           << "MapNormalizer" << "InternalCalibration"
-           << "TOFCalibration"
-           << "FeatureFinderCentroided" << "FeatureFinderIsotopeWavelet" << "FeatureFinderMultiplex"
-           << "MassTraceExtractor" << "FeatureFinderMetabo"
-           << "FeatureFinderMRM"
-           << "IsobaricAnalyzer" << "SpectraFilterWindowMower"
-           << "SpectraFilterThresholdMower" << "SpectraFilterSqrtMower"
-           << "SpectraFilterParentPeakMower" << "SpectraFilterMarkerMower"
-           << "SpectraFilterScaler" << "SpectraFilterBernNorm"
-           << "SpectraFilterNLargest" << "SpectraFilterNormalizer";
-    }
-    else if (layer_type == LayerData::DT_FEATURE)
-    {
-      list << "FileFilter" << "FileConverter"
-           << "FileInfo" << "Decharger"
-           << "FeatureLinkerLabeled";
-    }
-    else if (layer_type == LayerData::DT_CONSENSUS)
-    {
-      list << "FileFilter" << "FileConverter"
-           << "FileInfo";
-    }
-    else if (layer_type == LayerData::DT_CHROMATOGRAM)
-    {
-      //TODO CHROM
+        const String &toolName = tool.first;
+        Param p = getParamFromIni_(toolName);
+        std::vector<LayerData::DataType> toolTypes = getTypesFromParam_(p);
+        // Check if tool is compatible with the layer type
+        if (std::find(toolTypes.begin(), toolTypes.end(), layer_type) != toolTypes.end())
+        {
+            list << toolName.toQString();
+        }
     }
     //sort list alphabetically
     list.sort();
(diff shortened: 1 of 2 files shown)


src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
```diff
@@ -83,37 +87,26 @@ namespace OpenMS
...
-           << "FeatureLinkerLabeled";
-    }
-    else if (layer_type == LayerData::DT_CONSENSUS)
-    {
-      list << "FileFilter" << "FileConverter"
-           << "FileInfo";
-    }
-    else if (layer_type == LayerData::DT_CHROMATOGRAM)
+    // Determine all available tools compatible with the layer_type
+    tool_map_ = {
```
cbielow, line 91: please use FileTypeList (feel free to extend the class), instead of string for more robustness:
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1FileTypes_1_1FileTypeList.html

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
```diff
@@ -174,36 +167,65 @@ namespace OpenMS
...
+    // Containing all types a tool is compatible with
+    std::vector<LayerData::DataType> types;
+    for (const auto& entry : p)
+    {
+      if (entry.name == "in")
+      {
+        // Map all file extension to a LayerData::DataType
+        for (auto& fileExtension : entry.valid_strings)
+        {
+          const auto& iter = tool_map_.find(fileExtension);
```
dvdvgt, line 205: I changed to map the reflect your suggestion. I had to change the following lines:
```suggestion
          // valid string is of format "*.TYPE" -> convert to substr "TYPE" in order to be comparable.
          const String& fileTypeStr = fileExtension.substr(2, fileExtension.size());
          const auto& iter = tool_map_.find(FileTypes::nameToType(subStr));
```
Each entry in of `valid_strings` is of format "*.TYPE". I sliced of the first two characters in order to be able to convert it to a `FileTypes::Type` with `FileTypes::nametoType` so it can be mapped. Is this what you had in mind? I don't see a way to get completely rid of using strings to determine the type.

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
dvdvgt: I changed to map the reflect your suggestion. I had to change the following lines:
```suggestion
          // valid string is of format "*.TYPE" -> convert to substr "TYPE" in order to be comparable.
          const String& fileTypeStr = fileExtension.substr(2, fileExtension.size());
          const auto& iter = tool_map_.find(FileTypes::nameToType(subStr));
```
Each entry in of `valid_strings` is of format "*.TYPE". I sliced of the first two characters in order to be able to convert it to a `FileTypes::Type` with `FileTypes::nametoType` so it can be mapped. Is this what you had in mind? I don't see a way to get completely rid of using strings to determine the type.

timosachsenberg:
Just putting this here for discussion. If we add that many tools it might get cumbersome to select them from the drop down list. Eventually, it might be a good idea to organize them.
E.g., in TOPPAS (our workflow tool) we had a widget that displayed all tools in a treeview
image
One idea we could discuss is adding this tree view (but tools filtered as you did) to the left and get rid of the drop down list.
What do you guys think?

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
cbielow: jep. All good.

This is unfortunately yet another case where our internal TOPP datastructures are not fully modeled in ParamXML.
In all TOPP tools we just use ParameterInformation, which contains the extension without `*.`.
But that is not related to this PR ...

timosachsenberg:
And here is how it get's filled in TOPPAS https://github.com/OpenMS/OpenMS/blob/55598cf39c73a2092365fbdb51e63defa9b29c6f/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPASBase.cpp#L431

timosachsenberg:
Ideas for further enhancements:
- Some tools have **mandatory** input and output parameters. These should be definitely chosen as default in the input / output comboboxes.

dvdvgt:
> Ideas for further enhancements:
> 
>     * Some tools have **mandatory** input and output parameters. These should be definitely chosen as default in the input / output comboboxes. (I think these are currently marked in orange color in the ini editor below)
- [Currently the "in" and "out" parameters are chosen as default](https://github.com/OpenMS/OpenMS/blob/bbde6e5271ae549012dbc936f8ac9032c73f184c/src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp#L232). What changes do you have in mind? It appears that in some cases the `out` parameter is not required. Do you propose not to pre select that parameter if not necessary?
- In case multiple parameters (input or output) are mandatory what parameter would be chosen as default / prioritized?

I think the current implementation is sensible in that regard except for pre selecting the `out` parameter even if it is not required. I'd change that so that parameter is only chosen when required. What do you think?

dvdvgt:
So for the current commit there are 4 tests failing when I run them **locally**:
```
99% tests passed, 4 tests failed out of 2473

Total Test time (real) = 212.71 sec

The following tests FAILED:
         25 - StopWatch_test (Failed)
        490 - PrecursorIonSelectionPreprocessing_test (Failed)
        566 - DetectabilitySimulation_test (Failed)
        606 - MRMFeatureSelector_test (Failed)
Errors while running CTest
```
None of these test classes I changed and they did not fail with the previous commits even though no changes were done in between. Furthermore the same test did not fail on [travis](https://travis-ci.org/github/OpenMS/OpenMS/jobs/751212057) but instead others failed. I can't really make sense of this. I'd appreciate your help.

timosachsenberg:
http://cdash.openms.de/viewBuildError.php?buildid=89360

dvdvgt:
> Just putting this here for discussion. If we add that many tools it might get cumbersome to select them from the drop down list. Eventually, it might be a good idea to organize them.
> E.g., in TOPPAS (our workflow tool) we had a widget that displayed all tools in a treeview
> One idea we could discuss is adding this tree view (but tools filtered as you did) to the left and get rid of the drop down list.
> What do you guys think?

toppview

What do you guys think? I quite like it. Having the tools arranged by categories probably makes it easier finding a tool.

cbielow:
looks good!
One disadvantage here (which has been nagging me in TOPPAS as well), is that you cannot quickly find a tool when you already know its name, since its "somewhere" in the tree.
One solution is to add a `filter` textfield, where the user can enter partial names (infixes), and only tools which match the input are displayed. Something similar is already done in the Identification view.

So ideally, the treeview + filter would become a new custom widget, which can then be used in TOPPView and TOPPAS.

And one more idea: currently, the categories are ordered lexicographically. I usually find myself thinking about the categories in terms of where they occur in a processing pipeline. Maybe its worth introducing a custom order?!

timosachsenberg:
Should we go forward with merging and adding that to a new PR?

dvdvgt:
> Should we go forward with merging and adding that to a new PR?

Yes, I'd prefer opening a new PR for the GUI changes.

timosachsenberg:
One thing I forgot... can you also add the utils? 
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/APPLICATIONS/ToolHandler.h#L83

PR #5120, opened 2020-12-29
IDMapper map ids by native_id

### LINKED ISSUE #4965: IDMapper should support mapping by native_id
to avoid magic rt,m/z constants
used in, e.g., `LayerData::annotate`

RobertDeibel:
# Description
Implement IDMapper mapping by native_id. Rearranging code, and adding native_id retrieval and mapping to spectrum. 
Closes #4965

# Checklist:


RobertDeibel:
I'm unsure if the tests cover this. How would one go about writing test for this issue?

src/openms/source/ANALYSIS/ID/IDMapper.cpp
```diff
@@ -115,84 +118,107 @@ namespace OpenMS
...
 
     // store mapping of identification RT to index (ignore empty hits)
     multimap<double, Size> identifications_precursors;
     for (Size i = 0; i < peptide_ids.size(); ++i)
     {
       if (!peptide_ids[i].empty())
-      {
-        identifications_precursors.insert(make_pair(peptide_ids[i].getRT(), i));
+      { // mapping is done by either native id or by comparing peptide_id RT with experiment RT
+        if (peptide_ids[i].metaValueExists("spectrum_reference") == false)
```
jpfeuffer, line 132:
```suggestion
        if (!peptide_ids[i].metaValueExists("spectrum_reference")) 
```

src/openms/source/ANALYSIS/ID/IDMapper.cpp
```diff
@@ -115,84 +118,107 @@ namespace OpenMS
...
-    // store mapping of scan RT to index
-    multimap<double, Size> experiment_precursors;
-    for (Size i = 0; i < map.size(); i++)
-    {
-      experiment_precursors.insert(make_pair(map[i].getRT(), i));
-    }
+    lookup.readSpectra(map);
+
+    // remember which peptides were mapped (for stats later)
+    set<Size> peptides_mapped;
```
jpfeuffer, line 124: Maybe use unordered set if the order does not matter.

jpfeuffer (changes requested):
For tests, you can copy the input files from the old tests, add some "spectrum_reference" metavalues to the PeptideIdentifications and add an entry after the old one in the tests/topp/CmakeLists.txt

src/openms/source/ANALYSIS/ID/IDMapper.cpp
```diff
@@ -115,84 +118,107 @@ namespace OpenMS
...
 
-    // calculate the actual mapping
-    multimap<double, Size>::const_iterator experiment_iterator = experiment_precursors.begin();
-    multimap<double, Size>::const_iterator identifications_iterator = identifications_precursors.begin();
-    // to achieve O(n) complexity we now move along the spectra
-    // and for each spectrum we look at the peptide id's with the allowed RT range
-    // once we finish a spectrum, we simply move back in the peptide id window a little to get from the
-    // right end of the old interval to the left end of the new interval
-    while (experiment_iterator != experiment_precursors.end())
+    if (identifications_precursors.size() > 0)
```
timosachsenberg, line 153:
```suggestion
    if (!identifications_precursors.empty()) 
```

RobertDeibel:
I tried to add tests as you suggested @jpfeuffer, just copied test 1 and added some spectrum_reference meta values but I'm still not sure how the tests are actually performed. As in, I intentionally tried breaking the `IDMapper` class and while the `IDMapper_test` of tests/openms failed, all the TOPP tests still passed.

I'm obviously missing something but I'm not sure how I should go about on understanding the tests.

jpfeuffer:
Yes, so we have two types of tests:
- Class tests
- and TOPP/Tool tests

Class tests are standalone "additional" executables that include the class to be tested and our own testing utility classes to test outcomes of single functions of the class in question.

Tool tests are using the tool executable that the user would also receive. We use those executables to run the full algorithm on a small test dataset, to ensure that from version to version the results stay the same. The fact that the TOPP test did still pass is a bit weird, since it tests the complete algorithm. Maybe you did not break it enough. Often TOPP tests only use standard settings. Alternative "modes" are unfortunately not always tested. In your case, however, your new alternative mode needs to be added now.
TOPP tests are registered in the CMakeLists.txt that I mentioned and each test consists of
- an executable call on a test dataset (by using either fixed command line parameters or an ini file)
- a FuzzyDiff call that compares the temporary output file of the last call and a reference test output that you have to provide
- a line to add a dependency of the FuzzyDiff call on the actual executable call (so they get executed after each other)

RobertDeibel:
Added experiment test file with commit de76582

timosachsenberg:
OK I think we kind of added to the confusion. Sorry.
The IDMapper::annotate that you extended can not be tested by a topp tool test because we can't save the mzML with mapped ids.
Best way is probably to use the IDMapper_test.cpp. In that test I would load an IdXML without RT information (just copy one of the input idXML for IDMapper and delete the RT="..." attributes or set them to -1) and map it to it's mzML file.
If the spectra have the id attached it means that mapping without RT information just based on native id worked.

RobertDeibel:
Added the Test. Let me know if it's alright to place the new code it in the same test that tests the old behaviour. 

 Also: I didn't really think and rebased this already published branch. I'll probably close this PR and open a new one, that does not include all the junk. Let me know if that is unnecessary.

timosachsenberg:
don't nee to open a new one. I will do a "squash and merge" later so it will be only one commit

timosachsenberg (approved):
looks good

timosachsenberg:
please "squash and merge"  upon approval

Issue #4734, opened 2020-05-19
TextExporter output filetype(s)


hendrikweisser:
The output filetype of TextExporter was recently changed from CSV to TSV:
https://github.com/OpenMS/OpenMS/commit/fb675029ebb275a2fbf6a8c9b1a8b40bac7eddf1

While TSV output is indeed the default, the format can be changed (e.g. to CSV) by setting the "separator" parameter. Unfortunately simply allowing multiple output types (e.g. "tsv,csv,txt") wouldn't be enough, as e.g. TOPPAS requires an explicit "out_type" parameter in such cases. And then it should be checked whether the output type defined by "separator" actually matches the "out_type"...
Perhaps a better solution would be to replace "separator" by "out_type" and automatically set the separator based on that.

RobertDeibel:
I worked on this a little and have some remarks. See PR or copy pase:

> Some remarks:  
There are only two seperators here. Either `\t` of `,` which is more limiting than before.  
>
>I adjusted the `.ini` files of the tests but it seems that these are generated automatically. I generated a .ini (localy; see [test.ini.txt](https://github.com/OpenMS/OpenMS/files/5754920/test.ini.txt)) but I'm unsure if the others should just be replaced or if this is  done in some other place. 
>
>Does the `out_type` option have to be exposed somehow? The linked issue #4734 talks about tools requiring an out_type how would this be passed?
>
>Lastly maybe using both `seperator` and `out_type` would be an option? Making them mutually exclusive could be an option.

radusuciu:
Maybe a naive question, but why bother enforcing the extension at all? If the user wants to output a file with a `,` separator and have it be named `output.foo`, it seems like any issues stemming from that should be on them.

jpfeuffer:
I think it is mainly for workflow systems, so that downstream tools may know what file format to expect from upstream tools.
Such that workflow chaining errors can be caught during workflow-build time instead of at runtime after hours of execution.

PR #5121, opened 2020-12-30
TextExporter output filetype(s)

RobertDeibel:
# Description

- Replaced `seperator` option in `TextExporter` with `out_type`. 
- Seperator set implicitly by `out_type` option
- Adjusted .ini files of tests
- added more valid formats to `out` option  
Resolves issue #4734

# Checklist:


RobertDeibel:
Some remarks:  
There are only two seperators here. Either `\t` of `,` which is more limiting than before.  

I adjusted the `.ini` files of the tests but it seems that these are generated automatically. I generated a .ini (localy; see [test.ini.txt](https://github.com/OpenMS/OpenMS/files/5754920/test.ini.txt)) but I'm unsure if the others should just be replaced or if this is  done in some other place. 

Does the `out_type` option have to be exposed somehow? The linked issue #4734 talks about tools requiring an out_type how would this be passed?

Lastly maybe using both `seperator` and `out_type` would be an option? Making them mutually exclusive could be an option.

jpfeuffer:
Looks good. If the tests run, the inis were adapted correctly. I don't think they are autogenerated in this case.
If "txt" does not make a difference in the format, leave it out and remove it.

timosachsenberg:
If we specify -out "test.csv" but do not set out_type it will fall back to tsv, right?
Could this be surprising?
Maybe we make out_type optional and if it is not given determine it from the extension of the output file?

RobertDeibel:
Yes, you're right @timosachsenberg and I think this could be surprising. Especially for those not using out_type.
I'll apply the suggested code review and your suggestion.

RobertDeibel:
I'm unsure how to implement your suggestion @timosachsenberg. While it is possible to get the file extension I'm not sure how one would set the out_type parameter programmatically. 
As I see it, most functions of `TOPPBase` return `const` values when referring to parameters and those that do not are `private`.
Any suggestions?

jpfeuffer:
Not sure I understand the question but I would do the following to address @timosachsenberg s comment:
```
if out_type.empty()
 out_type = file.getExtension()
else
 if out_type != file.getExtension
   throw Error or Warn

getSeparatorFromOuttypeLogic(out_type)
```

I don't see a reason to retroactively set the tool's out_type parameter when all you need is the variable in the code. Or what did you have in mind?

RobertDeibel:
Yep you understood correctly. It just seemed me that it was intended by issue #4734, to have out_type always defined. 

But if your solution is the preferred way then everything is clear.

hendrikweisser:
Thanks for tackling this, @RobertDeibel.
There are a number of TOPP tools that allow different output types. Usually this is handled by trying to determine the type from "out" first, with "out_type" being optional. See FileConverter for example:
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/FileConverter.cpp#L342
https://github.com/OpenMS/OpenMS/blob/develop/src/topp/FileConverter.cpp#L412-L425

One could argue whether "out_type" shouldn't take precedence over "out" if it's specified (like @jpfeuffer suggested above), but I don't think it matters much.

timosachsenberg:
hi @RobertDeibel can you merge develop into your branch and add me as a collaborator? I might add something to the PR.

RobertDeibel:
> hi @RobertDeibel can you merge develop into your branch and add me as a collaborator? I might add something to the PR.

just merged `develop` and as a maintainer you should already have access to this branch.
Add a comment if that is not the case.

cbielow:
caused a CI regression.
Fixed in https://github.com/OpenMS/OpenMS/pull/5176

PR #5124, opened 2021-01-09
[FEATURE] Infer linking tolerance from RT distances of IDs in RT bins

jpfeuffer:
# Description

Calculates RT differences for IDs (per RT bin) to calculate a cutoff for adding unIDed features to clusters during QT linking.

# Checklist:


src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,6 +100,132 @@ namespace OpenMS
     // update parameters (dummy)
     setParameters_(1, 1);
 
+    if(use_IDs_)
+    {
+      std::unordered_map<String, std::vector<double>> ided_feat_rts;
```
timosachsenberg, line 105: add comment: map string "modified sequence/charge" to all RTs the feature has been observed in the different maps

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,6 +100,132 @@ namespace OpenMS
...
+      {
+        for (const auto& feat : map)
+        {
+          const auto& pepIDs = feat.getPeptideIdentifications();
+          if (!pepIDs.empty())
+          {
+            //TODO sort (lose const) or assume first is best?
+            if (!pepIDs[0].getHits().empty())
+            {
+              const auto it_inserted = ided_feat_rts.emplace(pepIDs[0].getHits()[0].getSequence().toString() + "/" + feat.getCharge(), std::vector<double>{feat.getRT()});
```
timosachsenberg, line 116: add const String key = pepIDs[0].getHits()[0].getSequence().toString() + "/" + feat.getCharge(); for better readability

timosachsenberg: is {feat.getRT()} sufficient?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,6 +100,132 @@ namespace OpenMS
...
+              if (!it_inserted.second)
```
timosachsenberg, line 117: add comment // key was already present? Append RT

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/QTClusterFinder.h
```diff
@@ -246,6 +249,23 @@ namespace OpenMS
      */ 
     void addClusterElements_(const Grid& grid, QTCluster& cluster);
 
+    bool distIsOutlier_(double dist, double rt);
```
timosachsenberg, line 252: add comment before merge

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/QTClusterFinder.h
```diff
@@ -246,6 +249,23 @@ namespace OpenMS
...
+
+    template <typename T1, typename T2> typename T1::value_type quantile_(const T1 &x, T2 q)
```
timosachsenberg, line 254: move to math functions?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,6 +100,132 @@ namespace OpenMS
...
+      }
+
+      //TODO this does not differentiate between the variety of differences between distinct map pairs. E.g.
+      // differences between map 1 and map 2 might be usually very small (e.g. they are replicates), while
+      // differences between map 1 and map 3 are large, since they are different conditions. But we might lose
+      // robust estimates and use more memory if we split them.
+      std::vector<std::pair<double,std::vector<double>>> medians_diffs;
+      medians_diffs.resize(ided_feat_rts.size());
+      Size c = 0;
+      for (auto& id_rts : ided_feat_rts)
```
timosachsenberg, line 134: please add a high level comment what is happening in this loop

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,6 +100,132 @@ namespace OpenMS
...
+        else
+        {
+          tmp_diffs.insert(tmp_diffs.end(), med_diffs.second.begin(), med_diffs.second.end());
+        }
+        cnt++;
+      }
+      // calculate allowed tolerance
+      std::sort(tmp_diffs.begin(), tmp_diffs.end());
+      q1 = quantile_(tmp_diffs, 0.25);
+      q3 = quantile_(tmp_diffs, 0.75);
```
timosachsenberg, line 207: we have MAD in statistics functions which might also be an option

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -513,6 +640,11 @@ void QTClusterFinder::createConsensusFeature_(ConsensusFeature& feature,
               {
                 continue; // conditions not satisfied
               }
+              if (use_IDs_ && neighbor_feature->getAnnotations().empty())
```
timosachsenberg, line 643: add comment what is checked here. e.g., // if IDs are used during linking check if unidentified features are too far off regarding RT

src/openms/source/DATASTRUCTURES/QTCluster.cpp
```diff
@@ -160,6 +160,7 @@ namespace OpenMS
       bool one_empty = (center_point.getAnnotations().empty() || element->getAnnotations().empty());
       if (!one_empty) // both are annotated
       {
+        //TODO shouldn't an overlap of at least one Sequence suffice?
```
timosachsenberg, line 163: yeah I think so

src/openms/source/DATASTRUCTURES/QTCluster.cpp
```diff
@@ -179,6 +180,7 @@ namespace OpenMS
     // Store best (closest) element:
     // Only add the element if either no element is present for the map or if
     // the element is closer than the current element for that map
+    //TODO shouldn't IDed features be preferred to unIDed ones?
```
timosachsenberg, line 183: yep e.g., if it would be possible to first cluster ided features (e.g., by setting distance to infinity) and then unided this could help potentially

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/QTClusterFinder.h
```diff
@@ -142,6 +142,9 @@ namespace OpenMS
     /// Set of features already used
     std::unordered_set<const OpenMS::GridFeature*> already_used_;
 
+    ///
+    std::map<double, double> bin_tolerances_;
```
timosachsenberg, line 146: is this mapping with double dangerous?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,6 +100,132 @@ namespace OpenMS
...
+      Size min_nr_diffs_per_bin = 50;
+
+      //TODO check if we can assume sorted
+      std::sort(medians_diffs.begin(), medians_diffs.end());
+
+      Size cnt = 0;
+      vector<double> tmp_diffs;
+      double start_rt = 0.;
+      double iqr_mult = 1.5;
+      double min_tolerance = 20;
```
timosachsenberg, line 170: I guess this is ok but could lead to situations where few features in a bin lead to e.g. bins 40, 13, 40 -> 40,20,40
but better (more robust) tolerances would be 40, 40, 40 in this case.
After non-linear alignment the tolerance should be a rather simple linear / non-linear function (maybe a linear fit on the RT tolerance ratios would work?).

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: You mean use the whole feature to calculate a note malized distance with FeatureDistance based on mz/RT/Int?
Yes I thought about that too. Do you think it is worth it?
By the way, if we include th experimental design we can have different weights for intensity (across replicates vs across conditions).

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: Yes, actually I thought about something completely different, since I think measures of spread are often used for symmetric distributions only. I currently use absolute differences.
But I am not sure. It kind of works already.

src/openms/source/DATASTRUCTURES/QTCluster.cpp
jpfeuffer: Yes I started on that as well. It is possible but a bit complicated since most of the code is not written in a reusable way to do incremental clustering.
But what I thought here: Since it is a normalized distance (0-1) we could add 1 to unidentified features so IDed ones are always preferred but not used exclusively.

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/QTClusterFinder.h
jpfeuffer: Since I only use lower_bound and don't really care in which of the corner case bins it falls, I don't think so.

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: Hmm good idea, so correct the tolerances again afterwards. Do we have simple code for such a fitting?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
timosachsenberg: here I just meant writing {rt} instead of vector<double>{rt}. but both ways is fine imho

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
timosachsenberg: I think for linear fit we have code. Including more robust versions like RANSAC

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/QTClusterFinder.h
timosachsenberg: ok if it can't happen that it falls into the one after last bin it should be fine

src/openms/source/DATASTRUCTURES/QTCluster.cpp
timosachsenberg: hmm sounds like it could work

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: ah no. Template types are not deduced for initializer_lists

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
timosachsenberg: not worth it as it seems

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -36,8 +36,9 @@
 #include <OpenMS/KERNEL/FeatureMap.h>
 #include <OpenMS/METADATA/PeptideIdentification.h>
 #include <OpenMS/KERNEL/FeatureHandle.h>
+#include <OpenMS/MATH/MISC/MathFunctions.h>
 
-// #define DEBUG_QTCLUSTERFINDER
+#define DEBUG_QTCLUSTERFINDER_IDS
```
jpfeuffer, line 41: TODO remove

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/QTClusterFinder.h
```diff
@@ -142,6 +155,9 @@ namespace OpenMS
     /// Set of features already used
     std::unordered_set<const OpenMS::GridFeature*> already_used_;
 
+    ///
```
jpfeuffer, line 158: TODO comment

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,10 +111,166 @@ namespace OpenMS
...
+      for (const auto& map : input_maps)
+      {
+        for (const auto& feat : map)
+        {
+          if (feat.getRT() < minRT) minRT = feat.getRT();
+          const auto& pepIDs = feat.getPeptideIdentifications();
+          if (!pepIDs.empty())
+          {
+            const auto& hits = pepIDs[0].getHits();
+            //TODO sort (lose const) or assume first is best? Especially since we sort later in run_internal_ anyway
```
jpfeuffer, line 129: This should be decided before merge

jpfeuffer:
Please someone review the changes to OpenSwath before I start fixing the dozens of tests.

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
```diff
@@ -127,55 +127,63 @@ namespace OpenMS
...
+    std::transform(intTheor.begin(), intTheor.end(), intTheor.begin(), OpenSwath::mySqrt());
 
-    manhattan = OpenSwath::manhattanDist(intExp.begin(), intExp.end(), theorint.begin());
+    double intExpTotal = std::accumulate(intExp.begin(), intExp.end(), 0.0);
+    double intTheorTotal = std::accumulate(intTheor.begin(), intTheor.end(), 0.0);
 
-    //std::cout << spectrum.size() << std::endl;
-    DIAHelpers::addPreisotopeWeights(firstIstotope, spectrum2, 2, -0.5);
-    std::vector<double>  theorint2;
-    DIAHelpers::extractSecond(spectrum, theorint2);
```
jpfeuffer, line 165: this uses the first spectrum, not the second.

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
```diff
@@ -127,55 +127,63 @@ namespace OpenMS
...
-    std::transform(theorint2.begin(), theorint2.end(), theorint2.begin(), OpenSwath::mySqrt());
+    OpenSwath::normalize(intExp, intExpTotal, intExp);
+    OpenSwath::normalize(intTheor, intTheorTotal, intTheor);
 
-    intExptotal = OpenSwath::norm(intExp.begin(), intExp.end());
```
jpfeuffer, line 168: This used euclidean norm on something that was previously normalized through sum?!

jpfeuffer:
This is ready for review/ A review is required. I am not going to fix the dozens of tests before people agree that I actually did the right thing.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -522,6 +530,10 @@ namespace OpenMS
 
     if (!svm_probs_internal_.empty()) calculateFDR_(features);
 
+    //TODO MRMFeatureFinderScoring already does an ElutionModel scoring. It uses EMG fitting.
```
timosachsenberg, line 533: @hendrikweisser

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -221,6 +228,7 @@ namespace OpenMS
     feat_finder_.setParameters(params);
     feat_finder_.setLogType(ProgressLogger::NONE);
     feat_finder_.setStrictFlag(false);
+    feat_finder_.setMS1Map(SimpleOpenMSSpectraFactory::getSpectrumAccessOpenMSPtr(boost::make_shared<MSExperiment>(ms_data_)));
```
timosachsenberg, line 231: what does this do?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -200,6 +200,13 @@ namespace OpenMS
     params.setValue("EMGScoring:max_iteration", param_.getValue("EMGScoring:max_iteration"));
     params.setValue("EMGScoring:init_mom", param_.getValue("EMGScoring:init_mom"));
     params.setValue("Scores:use_rt_score", "false"); // RT may not be reliable
+    params.setValue("Scores:use_ionseries_scores", "false"); // since FFID only uses MS1 spectra, this is useless
+    params.setValue("Scores:use_ms1_correlation", "false"); // this would be redundant to the "MS2" correlation and since
+    // precursor transition = first product transition, additionally biased
+    params.setValue("Scores:use_ms1_mi", "false"); // same as above. On MS1 level we basically only care about the "MS1 fullscan" scores
+    //TODO for MS1 level scoring there is an additional parameter add_up_spectra with which we can add up spectra around the apex,
+    // to complete ion ladders (and make this score more robust).
```
timosachsenberg, line 208: don't know from comment what is meant here. Complete ladders by merging MS1?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -128,18 +128,31 @@ namespace OpenMS
...
+                                            double charge = 1.);
+
     /// given an experimental spectrum add isotope pattern.
     OPENMS_DLLAPI void addIsotopes2Spec(const std::vector<std::pair<double, double> >& spec,
                           std::vector<std::pair<double, double> >& isotopeMasses, //[out]
-                          double charge = 1.);
+                          Size nr_isotopes, double charge = 1.);
+
+    /// given a peak of experimental mz and intensity, add isotope pattern to a "spectrum".
+    OPENMS_DLLAPI void addIsotopes2Spec(double mz, double ity,
```
timosachsenberg, line 153: is there a better name that is more expressive: 
addIsotopesOfASinglePeak2Spectrum (comment if it is according to averagine?)

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAScoring.h
```diff
@@ -136,6 +136,9 @@ namespace OpenMS
     /// Precursor isotope scores for precursors (peptides and metabolites)
     void dia_ms1_isotope_scores(double precursor_mz, SpectrumPtrType spectrum, size_t charge_state, 
                                 double& isotope_corr, double& isotope_overlap, const std::string& sum_formula = "") const;
+    void dia_ms1_isotope_scores(double precursor_mz, SpectrumPtrType spectrum,
```
timosachsenberg, line 139: this looks redundant
move charge_state e.g., before sum_formula and add a default value?
would make for a slimmer interface

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAScoring.h
```diff
@@ -205,6 +208,12 @@ namespace OpenMS
                                 int putative_fragment_charge,
                                 const std::string& sum_formula = "") const;
 
+    double scoreIsotopePattern_(const std::vector<double>& isotopes_int,
```
timosachsenberg, line 211: add documentation what these methods do- what is the score?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScores.h
```diff
@@ -67,6 +67,7 @@ namespace OpenMS
     bool use_ms1_fullscan = true;
     bool use_ms1_mi = true;
     bool use_uis_scores = true;
+    bool use_ionseries_scores = true;
```
timosachsenberg, line 70: does htis have an effect?

src/openms/include/OpenMS/MATH/MISC/MathFunctions.h
```diff
@@ -324,6 +324,24 @@ namespace OpenMS
       return lgamma(double(x+1));
     }
 
+    /**
+       @brief Returns the value of the @p q th quantile (0-1) in a sorted non-empty vector @x
+    */
+    template <typename T1, typename T2> typename T1::value_type quantile_(const T1 &x, T2 q)
```
timosachsenberg, line 330: why underscore in quantile_

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -55,9 +56,14 @@ namespace OpenMS
 
     defaults_.setValue("use_identifications", "false", "Never link features that are annotated with different peptides (only the best hit per peptide identification is taken into account).");
     defaults_.setValidStrings("use_identifications", ListUtils::create<String>("true,false"));
-    defaults_.setValue("nr_partitions", 100, "How many partitions in m/z space should be used for the algorithm (more partitions means faster runtime and more memory efficient execution )");
+    defaults_.setValue("nr_partitions", 100, "How many partitions in m/z space should be used for the algorithm (more partitions means faster runtime and more memory efficient execution).");
     defaults_.setMinInt("nr_partitions", 1);
-
+    defaults_.setValue("min_nr_diffs_per_bin", 50, "If IDs are used: How many differences from matching IDs should be used to calculate a linking tolerance for unIDed features in an RT region. RT regions will be extended until that number is reached.");
```
timosachsenberg, line 61: maybe document here what the fall-back is (e.g., if min not reached)

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,10 +111,166 @@ namespace OpenMS
     // update parameters (dummy)
     setParameters_(1, 1);
 
+    if(use_IDs_)
```
timosachsenberg, line 114: add space after if

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,10 +111,166 @@ namespace OpenMS
...
+          if (!pepIDs.empty())
+          {
+            const auto& hits = pepIDs[0].getHits();
+            //TODO sort (lose const) or assume first is best? Especially since we sort later in run_internal_ anyway
+            if (!hits.empty())
+            {
+              if ((hits[0].getScore() > min_score_ && pepIDs[0].isHigherScoreBetter()) ||
+                  (hits[0].getScore() < min_score_ && !pepIDs[0].isHigherScoreBetter()))
+              {
+                //TODO we could loosen the score filtering by requiring only ONE ID of a peptide to pass the threshold.
```
timosachsenberg, line 135: what does this mean in the case that e.g., 20 PSMs with same peptide are mapped to the feature?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,10 +111,166 @@ namespace OpenMS
...
+        Size i = 0;
+        for (const auto& rt : rts)
+        {
+          if (i++ == rts.size()/2) continue;
+          //TODO would relative diffs solve the RT dependency issue sufficiently? Probably not with non-linear shifts
+          //Note: One the one hand using abs. val. destroys the Gaussian distribution, on the other hand, if you only have
+          // two RTs for an ID, you will always have a negative difference from the median (and therefore the distribution
+          // would be biased anyway). We could use a stable median for evenly sized vectors but that would not reflect
+          // real distances then. Or always add positive AND negative differences.
+          medians_diffs[c].second.push_back(std::fabs(rt-medians_diffs[c].first));
```
timosachsenberg, line 186: space around minus

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,10 +111,166 @@ namespace OpenMS
...
+      std::sort(medians_diffs.begin(), medians_diffs.end());
+
+      Size cnt = 0;
+      vector<double> tmp_diffs, last_tmp_diffs;
+      // we calculate minRT (instead of starting first bin at 0) since RTs may start in the "negative" region after alignment.
+      double start_rt = minRT;
+      double min_tolerance = 20;
+      double tol, q2, q3 = 0.;
+      OPENMS_LOG_INFO << "Calculating RT linking tolerance bins...\n";
+      OPENMS_LOG_INFO << "RT_Bin_Start, Tolerance" << std::endl;
```
timosachsenberg, line 201: bin in lower case letters?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,10 +111,166 @@ namespace OpenMS
...
+
+      for (const auto& med_diffs : medians_diffs)
```
timosachsenberg, line 203: please add a brief comment what the loop is doing

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
jpfeuffer: Yeah but this is just a small optimization that we could do. So not too urgent. I think the scoring FFID is more elaborate also.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
jpfeuffer: It allows using MS1 scores.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
jpfeuffer: Ah.. oh.. should be isotopic envelopes of course.

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -615,6 +789,13 @@ void QTClusterFinder::createConsensusFeature_(ConsensusFeature& feature,
   {
     return feature_distance_(left->getFeature(), right->getFeature()).second;
   }
+
+  bool QTClusterFinder::distIsOutlier_(double dist, double rt)
+  {
+    if (bin_tolerances_.empty()) return false;
+    auto it = bin_tolerances_.lower_bound(rt);
```
timosachsenberg, line 796: - if rt is slightly larger than last this would return .end() is this still correct?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScores.h
jpfeuffer: yes, I deactivate it in FFID.

src/openms/include/OpenMS/MATH/MISC/MathFunctions.h
jpfeuffer: copy paste

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -615,6 +789,13 @@ void QTClusterFinder::createConsensusFeature_(ConsensusFeature& feature,
...
+    return dist >= (--it)->second;
```
timosachsenberg, line 797: - if bin_tolerances_ has exactly one element and rt is slightly smaller than this element. it will point on .begin(). --it is undefined. Can this happen (e.g. floating point) stuff?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: weird. I thought I did that.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
hendrikweisser: The elution model in FFId is used for quantification and filtering. In MRMFeatureFinderScoring, is the model also fit to multiple mass traces at once? And is there a way to pass on the model parameters (or are they lost after the score is calculated)?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: I intend to only use the best per feature. But if there are multiple features that have a best ID that maps to this peptide it could be enough if one exceeds the score to include all their differences. Just an idea.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -200,7 +200,8 @@ namespace OpenMS
     diascoring.dia_isotope_scores(transitions, spectrum, imrmfeature, scores.isotope_correlation, scores.isotope_overlap);
 
     // Peptide-specific scores
-    if (compound.isPeptide())
+    // useless for FFID
```
timosachsenberg, line 203: hmm not ideal to refer back to tool that calls OpenSwathScoring. Is there a better comment (you can stille add a note (e.g., as done in FFId) )

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -257,9 +258,20 @@ namespace OpenMS
 
       if (compound.isPeptide())
       {
-        diascoring.dia_ms1_isotope_scores(precursor_mz, ms1_spectrum,
-                                          precursor_charge, scores.ms1_isotope_correlation,
-                                          scores.ms1_isotope_overlap);
+        if (!compound.sequence.empty())
```
timosachsenberg, line 261: add comment why sequence can be empty

src/openms/source/DATASTRUCTURES/QTCluster.cpp
```diff
@@ -406,8 +409,10 @@ namespace OpenMS
       for (std::multimap<double, const GridFeature*>::const_iterator df_it =
              n_it->second.begin(); df_it != n_it->second.end(); ++df_it)
       {
+        std::set<AASequence> intersect;
         const std::set<AASequence>& current = df_it->second->getAnnotations();
-        if (current.empty() || (current == annotations_))
+        std::set_intersection(current.begin(), current.end(), annotations_.begin(), annotations_.end(), std::inserter(intersect, intersect.begin()));
+        if (!intersect.empty() || current.empty())
```
timosachsenberg, line 415: comment what you test here and what it means

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: yes, due to the "--" no?

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: > Returns an iterator pointing to the first element in the range [first, last) that is not less than (i.e. greater or equal to) value, or last if no such element is found.

I interpreted it differently. I can check in an online compiler.

jpfeuffer: https://godbolt.org/z/7jGdxz

jpfeuffer: https://godbolt.org/z/arfGKx

Feel free to play around. Looks fine to me.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
jpfeuffer: No, exactly, the FFID model does simultaneous optimization, while the other doesn't. And yes, the model parameters are currently gone. My favorite solution would be to use the FFID model during Swath scoring already and save the parameters. What do you need the params for?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -915,6 +930,8 @@ namespace OpenMS
       transition.setLibraryIntensity(iso_it->getIntensity());
       transition.setMetaValue("annotation", annotation);
       transition.setPeptideRef(peptide_id);
+
+      //TODO what about transition charge? A lot of DIA scores depend on it and default to charge 1 otherwise.
```
jpfeuffer, line 934: TODO remove. This comment should be solved by using the MS1 isotope scores only. I think "per-transition" isotopes do not make much sense here. They will overlap and the negative pre-isotopes during scoring may hit previous transitions in our case.

jpfeuffer: But I might need to check the other MS2 scores if they depend on a transition charge.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAScoring.h
oliveralka: I am not sure about the default charge state value, since for peptides it would most probably +2 (I guess). For metabolites, it would be either +1 (positive mode) or -1 (negative mode)

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -99,10 +111,166 @@ namespace OpenMS
...
+              {
+                //TODO we could loosen the score filtering by requiring only ONE ID of a peptide to pass the threshold.
+                // Would require a second pass though
+                const String key = pepIDs[0].getHits()[0].getSequence().toString() + "/" + feat.getCharge();
+                const auto it_inserted = ided_feat_rts.emplace(key, std::vector<double>{feat.getRT()});
+                if (!it_inserted.second) // already present
+                {
+                  it_inserted.first->second.push_back(feat.getRT());
+                }
+                /*const auto it_inserted_feat = ided_feats.emplace(key, std::vector<const typename MapType::FeatureType*>{&feat});
```
oliveralka, line 143: remove?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
oliveralka: maybe even addTheoIsotopes... -> to emphasise that it is the theoretical isotope pattern? I was actually confused in the beginning.

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
```diff
@@ -130,59 +130,67 @@ namespace OpenMS
...
-    intTheorTotal = OpenSwath::norm(theorint2.begin(), theorint2.end());
+    manhattan = OpenSwath::manhattanDist(intExp.begin(), intExp.end(), intTheor.begin());
 
-    OpenSwath::normalize(intExp, intExptotal, intExp);
-    OpenSwath::normalize(theorint2, intTheorTotal, theorint2);
+    // compare against the spectrum with negative weight preIsotope peaks
+    std::vector<double> intTheor2;
```
oliveralka, line 176: Maybe rename to intTheorNeg?

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
```diff
@@ -130,59 +130,67 @@ namespace OpenMS
...
-    dotprod = OpenSwath::dotProd(intExp.begin(), intExp.end(), theorint2.begin());
+    dotprod = OpenSwath::dotProd(intExp.begin(), intExp.end(), intTheor2.begin());
   }
 
   void DiaPrescore::updateMembers_()
   {
     dia_extract_window_ = (double) param_.getValue(
       "dia_extraction_window");
     nr_isotopes_ = (int) param_.getValue("nr_isotopes");
+    //TODO nr_charges_ is never used???
```
oliveralka, line 193: Hm, seem to be not actively used.
That is the only occurrence as far as I have seen:
DIAHelpers::addIsotopes2Spec(res, spectrum, nr_charges_);
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp#L140

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -201,17 +201,44 @@ namespace OpenMS
   }
 
   /// Precursor isotope scores
+  void DIAScoring::dia_ms1_isotope_scores(double precursor_mz, SpectrumPtrType spectrum,
+                                          double& isotope_corr, double& isotope_overlap, const EmpiricalFormula& sum_formula) const
+  {
+    //TODO precursor_mz can be received from the empirical formula (if required non-empty)
+
+    //collect the potential isotopes of this peak
+    int charge_state = sum_formula.getCharge();
```
oliveralka, line 210: I probably have overlooked it, but where was the charge set previously. Is there a standard charge based on the polarity of the experiment in the charge was not set properly?

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -201,17 +201,44 @@ namespace OpenMS
   }
 
   /// Precursor isotope scores
+  void DIAScoring::dia_ms1_isotope_scores(double precursor_mz, SpectrumPtrType spectrum,
```
oliveralka, line 204: What happens if you do not have identification for the compound and miss the EmpiricalFormula? Is the score just not reported or what happens? Since in theory, OpenSWATH should be able to quantify nonetheless?

oliveralka: That may be the problem why sum formula was reported as a string instead of EmpiricalFormula - but I am not sure.

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -236,27 +263,27 @@ namespace OpenMS
     double mz, intensity, left, right;
     std::vector<double> yseries, bseries;
     OpenMS::DIAHelpers::getBYSeries(sequence, bseries, yseries, generator, charge);
-    for (Size it = 0; it < bseries.size(); it++)
+    for (const auto& ion_mz : bseries)
```
oliveralka, line 266: Maybe rename to b_ion_mz and further down y_ion_mz?

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -150,8 +150,10 @@ namespace OpenMS
     scores_to_use.setValidStrings("use_ms1_fullscan", ListUtils::create<String>("true,false"));
     scores_to_use.setValue("use_ms1_mi", "false", "Use the MS1 MI score", ListUtils::create<String>("advanced"));
     scores_to_use.setValidStrings("use_ms1_mi", ListUtils::create<String>("true,false"));
-    scores_to_use.setValue("use_uis_scores", "false", "Use UIS scores for peptidoform identification ", ListUtils::create<String>("advanced"));
+    scores_to_use.setValue("use_uis_scores", "false", "Use UIS scores for peptidoform identification", ListUtils::create<String>("advanced"));
     scores_to_use.setValidStrings("use_uis_scores", ListUtils::create<String>("true,false"));
+    scores_to_use.setValue("use_ionseries_scores", "true", "Use MS2-level b/y ion-series scores for peptidoform identification", ListUtils::create<String>("advanced"));
```
oliveralka, line 155: Just for my understanding, this score was "just" decoupled from the DIAScores, or?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
```diff
@@ -139,7 +139,7 @@ namespace OpenMS
     // exclude some redundant/uninformative scores:
     // @TODO: intensity bias introduced by "peak_apices_sum"?
     // names of scores to use as SVM features
-    String score_metavalues = "peak_apices_sum,var_xcorr_coelution,var_xcorr_shape,var_library_sangle,var_intensity_score,sn_ratio,var_log_sn_score,var_elution_model_fit_score,xx_lda_prelim_score,var_isotope_correlation_score,var_isotope_overlap_score,var_massdev_score,main_var_xx_swath_prelim_score";
+    String score_metavalues = "peak_apices_sum,var_xcorr_coelution,var_xcorr_shape,var_library_sangle,var_intensity_score,sn_ratio,var_log_sn_score,var_elution_model_fit_score,xx_lda_prelim_score,var_ms1_isotope_correlation_score,var_ms1_isotope_overlap_score,var_massdev_score,main_var_xx_swath_prelim_score";
```
oliveralka, line 142: What about the RT score in FFID? 
iRT in the library can be set as normal RT in seconds (-Library:retentionTimeInterpretation 'seconds') and setting the rt normalization factor to the length of the gradient used in seconds (-Scoring:rt_normalization_factor 1200.0).

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
jpfeuffer: Yes which was wrong in my opinion and was changed here. It was used instead of nr isotopes. What do you think?

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
jpfeuffer: I think the constructor of EmpiricalFormula takes a charge. Or AASequence.getFormula sets it automatically, too.

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
jpfeuffer: yep. that is why I left the old overload. In that case, an averagine model is used. There is a comment somewhere that this is not accurate for metabolites but better than nothing.

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
jpfeuffer: Basically a sub-score. If DIAScores is off, I want this to be off, too. But I also want to be able to disable it, even if DIAScores are "on".

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
jpfeuffer: What exactly does it measure? There was a comment slightly above that says the RT scores are a bit unreliable.

src/openms/source/DATASTRUCTURES/QTCluster.cpp
```diff
@@ -433,22 +438,36 @@ namespace OpenMS
           df_it != n_it->second.end(); ++df_it)
```
timosachsenberg, line 438: if you know what is happening in the loops can you add a comment on top of them?

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
timosachsenberg: there seems to be too much code duplication with function below

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -362,14 +388,14 @@ namespace OpenMS
       }
       if (ratio > max_ratio) {max_ratio = ratio;}
 
-      double ddiff_ppm = std::fabs(mz - (mono_mz - 1.0 / (double) ch)) * 1000000 / mono_mz;
+      double ddiff_ppm = std::fabs(mz - (mono_mz - 1.0 / (double) ch)) * 1e6 / mono_mz;
```
timosachsenberg, line 391: is this correct? why 1.0? but doesn't seem to have a lot of impact

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -399,11 +424,29 @@ namespace OpenMS
     {
       // create the theoretical distribution from the peptide weight
       CoarseIsotopePatternGenerator solver(dia_nr_isotopes_ + 1);
-      isotope_dist = solver.estimateFromPeptideWeight(std::fabs(product_mz * putative_fragment_charge));
+      isotope_dist = solver.estimateFromPeptideWeight(std::fabs((product_mz - C13C12_MASSDIFF_U * putative_fragment_charge) * putative_fragment_charge));
```
timosachsenberg, line 427: Looks correct now but the difference in isotope intensities between new and old code will be super small

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAScoring.h
jpfeuffer: I can move the param, but I dont think it is redundant right? Since sum_formula can be empty, we need to require it. I have the new overload that is slimmer.

jpfeuffer: Hmm ok then charge_state should be an int. How did you do it currently @oliveralka :)

oliveralka: Usually in my case charge is already set by SIRIUS. It also can be assessed the annotated adduct, which currently only covers charge +1/-1.
https://github.com/OpenMS/OpenMS/pull/4725/files/14f627ec39174b4e499534a241465e1f9fc905dc#diff-d315275bb3622b58d06fd4ea2199d75255d7338e2f1aee06f421ea27bdad0d20R49

If neither sumformula nor adducts are set you would probably assume a positive/negative charge based on the polarity of the experiment.

jpfeuffer: No, I mean in the current state OpenSwath will evaluate this score with a charge of 2^32 - 1 since the integer -1 is converted to an unsigned int.

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
jpfeuffer: I guess decrementing begin is implementation defined. I will add an if-case.

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
```diff
@@ -130,59 +130,67 @@ namespace OpenMS
...
+    {
+      if (transition.fragment_charge != 0) chg = transition.fragment_charge;
+      DIAHelpers::addPreisotopeWeights(transition.getProductMZ(), spectrumWIso, nrNegPeaks, 0.0);
+      DIAHelpers::addPreisotopeWeights(transition.getProductMZ(),
+                                       spectrumWIsoNegPreIso,
+                                       nrNegPeaks,
+                                       -totalNegWeight / (double(lt.size() * nrNegPeaks)));
+    }
+    DIAHelpers::sortByFirst(spectrumWIso);
+    DIAHelpers::sortByFirst(spectrumWIsoNegPreIso);
```
timosachsenberg, line 157: this seems like the main difference to the previous implementation
can you comment that part?

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
timosachsenberg: yep looks like a bug to me

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
jpfeuffer: yes I thought so too, but still, it is more correct.

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
jpfeuffer: I'll change to C13C12 diff right?

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -201,17 +201,44 @@ namespace OpenMS
...
+    //collect the potential isotopes of this peak
+    int charge_state = sum_formula.getCharge();
+    double max_ratio;
+    int nr_occurrences;
+    std::vector<double> isotopes_int;
+    for (int iso = 0; iso <= dia_nr_isotopes_; ++iso)
+    {
+      double left  = precursor_mz + iso * C13C12_MASSDIFF_U / static_cast<double>(charge_state);
+      double right = left;
+      DIAHelpers::adjustExtractionWindow(right, left, dia_extract_window_, dia_extraction_ppm_);
```
cbielow, line 218: am I the only one who thinks this code is utterly non-intuitive?
I assumed it's broken, twice, when I saw it. But it seems to do the right thing, if you follow all the clues.
1) why pass two doubles by ref (which need to have the same value), instead of passing one and returning 2?
   (good place for using structured binding :))
2) switching the order of parameters (right, left) instead of (left, right), ... wow... just wow.

jpfeuffer: haha true, I was distracted by the actually wrong stuff around here.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -257,9 +258,20 @@ namespace OpenMS
 
       if (compound.isPeptide())
       {
-        diascoring.dia_ms1_isotope_scores(precursor_mz, ms1_spectrum,
-                                          precursor_charge, scores.ms1_isotope_correlation,
```
jpfeuffer, line 261: By the way, why was nothing complaining here? dia_ms1_isotope_scores takes a size_t as charge. If precursor_charge is negative (from the transition) you probably get quite some nonsense.

src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
timosachsenberg: I mean having it const would be great but if it is too much hassle we can go with the sorting for now.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -128,18 +128,33 @@ namespace OpenMS
...
                               double preIsotopePeaksWeight = -0.5, // weight of pre-isotope peaks
                               double mannmass = 1.000482, //
                               double charge = 1.);
 
-    /// given an experimental spectrum add isotope pattern.
+    /// add negative pre-isotope weights to spectrum
+    OPENMS_DLLAPI void addPreisotopeWeights(double mz,
+                                            std::vector<std::pair<double, double> >& isotopeSpec, // output
+                                            UInt nrpeaks = 2, //nr of pre-isotope peaks
+                                            double preIsotopePeaksWeight = -0.5, // weight of pre-isotope peaks
```
JeeH-K, line 143: param names don't seem like fitting the coding convention

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp
jpfeuffer: @oliveralka ping

oliveralka: The rt_score basically the difference to the retention time in the library. The smaller the difference the better. This comes in handy if the chromatography is similar / measured at the same column/instrument. It depends a little bit on how big your extraction window is. You have an extraction window of 600s seconds and find two similar peak groups one at 1300s (1) and the other one at 1600s rt (2) - your peak group in the library is at 1580s <- so in terms of the retention time score peak group 2 would be better off with a difference of 20s instead of 280s.

I am not sure about the unreliability statement. Chromatography always is dependent on the experimental setup!

oliveralka: I guess you have a really small extraction window in the FFMID, so the score is probably not needed here.

jpfeuffer: Hmm also, if transitions are "transferred" to other maps it depends on how good the alignment of the maps worked I guess.
The extraction windows are not so small since it sometimes tries to pick up same IDs that elude multiple times and because of the aforementioned transfer.

oliveralka: I guess this is why usually TRIC or DIAlignR are used after OpenSWATH, to realign "mismatched" peak groups based on the chromatograms.

hendrikweisser: The unreliability (as far as FFId is concerned) comes from the fact that the library RT is only an estimate (for "external" IDs). If we knew the exact RT where to expect a feature, we wouldn't need any sophisticated scoring to choose between feature candidates.

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
```diff
@@ -130,59 +131,85 @@ namespace OpenMS
...
 
-    intExptotal = OpenSwath::norm(intExp.begin(), intExp.end());
-    intTheorTotal = OpenSwath::norm(theorint2.begin(), theorint2.end());
+    manhattan = OpenSwath::manhattanDist(intExp.begin(), intExp.end(), intTheor.begin());
 
-    OpenSwath::normalize(intExp, intExptotal, intExp);
-    OpenSwath::normalize(theorint2, intTheorTotal, theorint2);
+    // compare against the spectrum with negative weight preIsotope peaks
+    std::vector<double> intTheorNeg;
+    // WARNING: This was spectrumWIso and therefore with 0 preIso weights in earlier versions! Was this a bug?
```
jpfeuffer, line 195: @grosenberger @hroest Do you know if it was intentional to not use negative weights for the preisotopes? E.g. after benchmarking you realized that it is penalizing too much and then just forgot to revert that part?

src/openms/source/ANALYSIS/OPENSWATH/DIAPrescoring.cpp
```diff
@@ -130,59 +131,85 @@ namespace OpenMS
...
                           double& dotprod,
-                          double& manhattan)
+                          double& manhattan) const
   {
     std::vector<std::pair<double, double> > res;
-    getMZIntensityFromTransition(lt, res);
-    std::vector<double> firstIstotope, theomasses;
-    DIAHelpers::extractFirst(res, firstIstotope);
-    std::vector<std::pair<double, double> > spectrum, spectrum2;
-    DIAHelpers::addIsotopes2Spec(res, spectrum, nr_charges_);
```
jpfeuffer, line 140: nr_charges_ here is at the parameter for nr_isotopes_, which is ok since the class here sets the same defaults. But this call does not set charge! Therefore this function was always called with charge=4 -> completely different isotope patterns since the mz is multiplied by 4 instead of usually 1.

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -267,60 +267,87 @@ namespace OpenMS
       double mass = product_mz;
       for (IsotopeDistribution::Iterator it = d.begin(); it != d.end(); ++it)
       {
-        isotopesSpec.push_back(std::make_pair(mass, it->getIntensity()));
+        isotopes_spec.push_back(std::make_pair(mass, it->getIntensity()));
         mass += mannmass;
```
jpfeuffer, line 271: This should be mannmass / charge. But luckily (for charge = 1) it cancels out other bugs where always charge 4 was used. So only intensities were wrong.

timosachsenberg:
If you want to split this PR into two parts (one for the linker and one for swath) you can create a copy of the branch with git checkout -b ...
and then do an interactive rebase git rebase -i
just add drop in front of the commits that you want to have removed from that branch

jpfeuffer:
Merge? I think we could potentially revert stuff later (although 95% were obvious bugfixes). The implications to OpenSwath were minor since those scores had very low pre-trained LDA weights (probably due to being wrongly calculated).

timosachsenberg:
k given that we want to test it now I would say a merge is fine

PR #5135, opened 2021-01-21
ParamEditor with more convenient StringList editing 

cbielow:
# Description

Valid strings can now be filtered for substrings, which should safe a lot of time scrolling through the long list...

image


src/openms/include/OpenMS/DATASTRUCTURES/ListUtils.h
```diff
@@ -134,6 +134,22 @@ namespace OpenMS
...
 
+
+    /**
+  @brief Converts a vector of T's to a vector of Strings.
+
+  @param s The vector of T's that should be converted.
+  @return A vector containing the elements of input vector converted into Strings.
+*/
+    template <typename T>
+    static std::vector<String> toStringList(const std::vector<T>& s)
```
timosachsenberg, line 145: could we also make the container a template so it can be used with lists and sets?

timosachsenberg: maybe too much trouble but boost has some traits for that

cbielow: My preferred solution would be a StringList class which has appropriate members. Simply because it's easier to find functionality than using free functions.
But that was too much trouble for now. But might make for a nice hacktoberfest project.

I'd like to avoid boost for such simple things because of unneeded dependency and the header bloat which boost usually entails. But that's just me.

cbielow:
For any reviewer: i'd encourage testing it live, because this delegate/commit shinnanigans which Qt is pulling there has many implicit and undocumented mechanisms and I think I got it right, but you never know. So try if editing different kinds of Parameters in INIFileEditor works. The new dialog as well of course :-)

poshul:
I've checked this with NucleicAcidSearchEngine and it works. Its a huge QoL improvement for me. One item to note, the filter box has pretty good regex support, but I cant figure out how to escape characters as literal.

poshul (commented):
Looks like a good addition to me.

cbielow:
> I've checked this with NucleicAcidSearchEngine and it works. Its a huge QoL improvement for me. One item to note, the filter box has pretty good regex support, but I cant figure out how to escape characters as literal.

I currently use `Wildcard` mode of Qt's regex, which is described here:
https://doc.qt.io/archives/qt-4.8/qregexp.html#wildcard-matching
So this is not a full regex support, because I figured it would be overwhelming for our rather simple lists...
Do you have an example where more powerful regex might be useful?

poshul:
> 
> 
> > I've checked this with NucleicAcidSearchEngine and it works. Its a huge QoL improvement for me. One item to note, the filter box has pretty good regex support, but I cant figure out how to escape characters as literal.
> 
> I currently use `Wildcard` mode of Qt's regex, which is described here:
> https://doc.qt.io/archives/qt-4.8/qregexp.html#wildcard-matching
> So this is not a full regex support, because I figured it would be overwhelming for our rather simple lists...
> Do you have an example where more powerful regex might be useful?

I think the wildcard support is sufficient for our uses. The one situation I noticed is that the literal ? is part of some RNA mods, so the lack of escaping means users need to track them down from the list rather than searching. Given the new list is such a huge improvement anyway I have no complaints leaving it as is.

cbielow:
i changed it to `wildcardUnix` which allows escaping of the supported wildcards by `\` (see updated screenshot above).

PR #5136, opened 2021-01-22
Option to Convert a Fasta to a mzML File in IDFileConverter using TSG

Waschi97:
added:
- additional valid in- and output file types
- new parameter subsection representing TSG params
- test for the new file conversion

Note:
TSG seems to be calculating spectra with a small error. Because of this and since MZ information is stored as base64 in mzML, testing the mzML output with FuzzyDiff wouldn't work. To resolve this issue the mzML is converted into MGF and is tested as such.


src/topp/IDFileConverter.cpp
```diff
@@ -566,6 +584,82 @@ class TOPPIDFileConverter :
...
       }
 
+      else if (in_type == FileTypes::FASTA)
+      {
+        MSExperiment exp;
+        TheoreticalSpectrumGenerator tsg;
+
+        // extract parameters and remove non tsg params
+        Param p = getParam_().copy("fasta_to_mzml:", true);
+        String enzym = p.getValue("enzym");
```
jpfeuffer, line 594: enzyme?

src/topp/IDFileConverter.cpp
```diff
@@ -566,6 +584,82 @@ class TOPPIDFileConverter :
...
+        Int min_charge = p.getValue("min_charge");
+        Int max_charge = p.getValue("max_charge");
+        Int pc_charge = p.getValue("precursor_charge");
+        p.remove("enzym");
+        p.remove("min_charge");
+        p.remove("max_charge");
+        p.remove("precursor_charge");
```
jpfeuffer, line 601: You could add max_missed_cleavages.
By the way: What is the difference to the MSSimulator when deactivating everything but MS2? Or is that not possible?

Waschi97: Missed cleavages is probably a good idea.

Not sure what you mean by "deactivating everything but MS2"... Deactivate where?

cbielow:
since the other PR is already merged, please correct the error message
https://github.com/OpenMS/OpenMS/commit/7547a7181a32ebd66979c4e0b4c42441f1b368ee#diff-863f27a9d733c2294f015ebe91d2ce4245cd1e0fdb5a87b9e5c126c92d5d0a1fR236
to `pc charge must be equal or larger than ...`

src/topp/IDFileConverter.cpp
cbielow: nope. And its not easy to control which peptides are ionized etc...

src/topp/IDFileConverter.cpp
```diff
@@ -566,6 +585,85 @@ class TOPPIDFileConverter :
...
+            tsg.getSpectrum(spec, peptide, min_charge, max_charge, pc_charge);
+            exp.addSpectrum(spec);
+          }
+        }
+        logger.endProgress();
+
+        // handle out type
+        const String out = getStringOption_("out");
+        FileTypes::Type out_type = FileTypes::nameToType(getStringOption_("out_type"));
+        if (out_type != FileTypes::MZML)
```
cbielow, line 650: this should be done before doing expensive computations, so the user gets the feedback immediately instead of waiting until the tool is almost done.

src/topp/IDFileConverter.cpp
```diff
@@ -566,6 +585,85 @@ class TOPPIDFileConverter :
...
+        p.remove("precursor_charge");
+
+        if (min_charge > max_charge)
+        {
+          writeLog_("Error: 'fasta_to_mzml:min_charge' must be smaller than or equal to 'fasta_to_mzml:max_charge'.");
+          printUsage_();
+          return ILLEGAL_PARAMETERS;
+        }
+
+        if (pc_charge != 0 && pc_charge <= max_charge)
```
cbielow, line 613: you can omit this check, right? because the TSG will throw an exception and let the user know

src/topp/IDFileConverter.cpp
```diff
@@ -566,6 +585,85 @@ class TOPPIDFileConverter :
...
+          vector<AASequence> digested_peptides;
+          AASequence seq = AASequence::fromString(entry.sequence);
+          digestor.digest(seq, digested_peptides);
+
+          // for each peptide calculate the theoretical spectrum
+          for (const auto& peptide : digested_peptides)
+          {
+            PeakSpectrum spec;
+            tsg.getSpectrum(spec, peptide, min_charge, max_charge, pc_charge);
+            exp.addSpectrum(spec);
```
cbielow, line 642: optimization: move it in

cbielow (commented):
looks good.
(test is missing, as indicated)

src/topp/IDFileConverter.cpp
timosachsenberg: Or keep it as an OpenMS precondition. It will only be enabled in debug mode

timosachsenberg (approved):
Add test and hope that abibuilder comes back soon

jpfeuffer:
@Waschi97 The nightlies fail with multiple errors in the IDFileconverter Topp tests which seem related to this. Did you run the tests in debug mode?
http://cdash.openms.de/testDetails.php?test=30925214&build=100943

PR #5137, opened 2021-01-24
Change TextExporter Output to Display Accession Details

### LINKED ISSUE #4668: TextExporter: export start and end position in protein

emilpaulitz:
# Description

This problem is described in issue #4668. If a peptide has multiple accessions, we currently only print the set of accessions, but not the start/end of the match, and not how many times a peptide matches in a protein. This change proposes to print multiple line for the same peptide, which are identical except for the accession column. That column contains the protein name and the start and end position of the match. 

Obviously, if there is no accession or only one, we only print one line.

The changes are reflected in outputs of TextExporter tests 3,4 and 5. For example:
Instead of:
PEPTIDE	1234.5	675.9	0.9	0	DEF	1	A	B	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	**PROT1;PROT2**	-1	-1

We write:
PEPTIDE	1234.5	675.9	0.9	0	DEF	1	A	B	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	**PROT1 start: 3 end: 5**	-1	-1
PEPTIDE	1234.5	675.9	0.9	0	DEF	1	A	B	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	**PROT2 start: 0 end: 2**	-1	-1

Closes #4668

# Checklist:


cbielow:
thanks for the PR!
I'm not certain this is a good way to solve the problem, because:

1) you are breaking backwards compatibility
2) this 'long table' strategy is IMHO causing more trouble than good, because:
  - it makes the file bigger/redundant
  - you cannot simply look at the number of rows and deduce how many PSMs were identified. Same goes for other quality metrics which one would like to compute quickly. And if you forget to remove duplicates, you are effectively reporting wrong things without even knowing it... dangerous..
3) The header name still says `accessions`, even though its now a single accession, with start/end coordinates on top (hard to parse).

So my suggestion is to keep the compact form as before. Add start and end as single columns, and concatenate the numbers using `;`, as is done in idXML for example.
This way you do not break compatibility, support easy computation of overview statistics and make it less error prone.

I can see that this format you suggest is similar to the way mzTab is doing it. but IMHO thats wrong, dangerous and cumbersome to parse, since usually you are not interested in `how many proteins does each peptide map to` (which is easy to answer in your format), but in basically everything else (which is easy to answer in the previous format -- and even the number of proteins can be easily determined by just counting `;` ...).
So to summarize: i dont see any benefit :(

cbielow:
up for discussion of course!

timosachsenberg:
I agree with chris. Actually we decided at some point that we break with the MzTab specification and do as Chris is proposing here. One of the reasons is also that users just do a grep "PEPTIDE" | wc -l to quickly derive the number of peptides etc.
So I would also suggest to write a single line and concatenate accessions, AA before/after etc.

emilpaulitz:
Alright, thank you for your feedback! I was thinking printing everything in one line makes it hard to read manually but I was not taking into account that one probably mostly uses scripts or excel to analyze larger files. I implemented this in the latest commit.

src/tests/topp/TextExporter_5_output_peptides.txt
```diff
@@ -1,6 +1,6 @@
...
-1234.5	675.9	0	0	TC(Carboxymethyl)VADESHAGC(Carboxymethyl)EK	2	K	S	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	PROT1;PROT2	359.970709999999997	4.0	0.51502823	-1
-1234.5	675.9	0	0	KC(Carboxymethyl)PSPC(Carboxymethyl)QLQAER	2	K	A	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	PROT1	892.753230000000031	4.0	4.7040273	-1
--1	-1	0	0	KC(Carboxymethyl)PSPC(Carboxymethyl)QLQAER	2	K	A	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	PROT2	892.753230000000031	6.0	4.7040273	-1
--1	-1	0.001507159	0	KC(Carboxymethyl)PSPC(Carboxymethyl)QLQAER	2	K	A	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	PROT2	892.753230000000031	6.0	4.7040273	-1
--1	-1	0	0	EYEATLEEC(Carboxymethyl)C(Carboxymethyl)AK	2	K	D	MOWSE	Mascot_2007-01-12T12:13:14_4835329514588776807	PROT1	900.572959999999966	6.0	3.9354399	-1
+#rt	mz	score	rank	sequence	charge	aa_before	aa_after	score_type	search_identifier	accessions	start	end	predicted_rt	rt_first_dim	predicted_rt_first_dim	predicted_pt
+1234.5	675.9	0	0	TC(Carboxymethyl)VADESHAGC(Carboxymethyl)EK	2	K	S	MOWSE	Mascot_2006-01-12T12:13:14_5233264595117471314	PROT1;PROT2	;	;	359.970709999999997	4.0	0.51502823	-1
```
timosachsenberg, line 2: There seem to be an issue that ; is printed even if the information is not present

emilpaulitz: I intended this, in case there exists start/end information for only some of the accessions. You would need to have some kind of indication which accession these belong to.

timosachsenberg: Sounds good! Did not think about that

emilpaulitz: Now it only prints redundant semincolons if there is information at all

timosachsenberg: @cbielow  looks good from my side. Should we do something about the number of digits for the score?

cbielow: we should solve this at a higher level. I remember @jpfeuffer fixing some precision issues, but from what I remember boost does not allow fine control over the precision (only the number of digits, which is not the same thing)

timosachsenberg (approved):

timosachsenberg:
Thanks!

PR #5140, opened 2021-01-26
[FEATURE] c17

jpfeuffer:
# Description

Fixes stuff for c17.
This PR needs contributions from everyone. This is to finally get it started.

TODOS:
- [x] update cpplint to e.g. support structured bindings. See current failure on CDash. Related: https://github.com/RobotLocomotion/drake/issues/12577
- [x] update openms-ci container for pyopenms to gcc7
- [x] debug pyopenms
- [x] do everything again for seqan, which fails heavily on Windows (although we should just remove it. We use maybe one class! Can't we just include that only?). I can't help here, since I do not have Windows and remote takes too much time.

# Checklist:

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ConfidenceScoring.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ConfidenceScoring.h
@@ -37,6 +37,7 @@
 #include <cmath> // for "exp"
 #include <ctime> // for "time" (random number seed)
 #include <limits> // for "infinity"
+#include <random>
 #include <boost/bimap.hpp>
 #include <boost/bimap/multiset_of.hpp>
 #include <boost/random/uniform_int.hpp>
@@ -62,9 +63,9 @@ namespace OpenMS
 
       /// Constructor
       explicit ConfidenceScoring(bool test_mode_ = false) :
-        generator_(), rand_gen_(generator_, boost::uniform_int<>())
+        generator_(), rand_gen_(std::random_device()())
       {
-        if (!test_mode_) rand_gen_.engine().seed(time(nullptr)); // seed with current time
+        if (!test_mode_) rand_gen_.seed(time(nullptr)); // seed with current time
       }
 
       virtual ~ConfidenceScoring() {}
@@ -118,7 +119,7 @@ namespace OpenMS
       boost::mt19937 generator_; ///< random number generation engine
 
       /// Random number generator (must be initialized in init. list of c'tor!)
-      boost::variate_generator<boost::mt19937&, boost::uniform_int<> > rand_gen_;
+      std::mt19937 rand_gen_;
 
       /// Randomize the list of decoy indexes
       void chooseDecoys_();
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ConfidenceScoring.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ConfidenceScoring.h
@@ -40,14 +40,12 @@
 #include <random>
 #include <boost/bimap.hpp>
 #include <boost/bimap/multiset_of.hpp>
-#include <boost/random/uniform_int.hpp>
-#include <boost/random/mersenne_twister.hpp>
-#include <boost/random/variate_generator.hpp>
 
 #include <OpenMS/FORMAT/FeatureXMLFile.h>
 #include <OpenMS/FORMAT/TraMLFile.h>
 #include <OpenMS/FORMAT/TransformationXMLFile.h>
 #include <OpenMS/KERNEL/FeatureMap.h>
+#include <OpenMS/MATH/MISC/MathFunctions.h>
 
 #include "OpenMS/OPENSWATHALGO/ALGO/Scoring.h"
 #include <OpenMS/ANALYSIS/MAPMATCHING/TransformationDescription.h>
@@ -62,10 +60,10 @@ namespace OpenMS
   public:
 
       /// Constructor
-      explicit ConfidenceScoring(bool test_mode_ = false) :
-        generator_(), rand_gen_(std::random_device()())
+      explicit ConfidenceScoring(bool test_mode_ = false)
       {
-        if (!test_mode_) rand_gen_.seed(time(nullptr)); // seed with current time
+        if (!test_mode_) shuffler_ = Math::RandomShuffler(0);
+        else shuffler_ = Math::RandomShuffler(time(nullptr));// seed with current time
       }
 
       virtual ~ConfidenceScoring() {}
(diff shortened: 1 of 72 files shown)


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ConfidenceScoring.h
```diff
@@ -118,7 +119,7 @@ namespace OpenMS
       boost::mt19937 generator_; ///< random number generation engine
 
       /// Random number generator (must be initialized in init. list of c'tor!)
-      boost::variate_generator<boost::mt19937&, boost::uniform_int<> > rand_gen_;
+      std::mt19937 rand_gen_;
```
timosachsenberg, line 122: I think we need to be very careful about the std::random stuff at is unfortunately not deterministically portable.
We had this problem before and the only solution was to not use std generators
https://www.reddit.com/r/cpp/comments/7i21sn/til_uniform_int_distribution_is_not_portable/

src/openms/include/OpenMS/MATH/MISC/RANSAC.h
```diff
@@ -237,6 +233,8 @@ namespace OpenMS
         return(bestdata);
       } // ransac()
 
+    private:
+      std::mt19937 rng{std::random_device()()};
```
timosachsenberg, line 237: I think this might not be portable

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/ConfidenceScoring.h
timosachsenberg: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/CHEMISTRY/DecoyGenerator.h#L89

timosachsenberg:
Good summary on C17: https://www.bfilipek.com/2017/01/cpp17features.html

jpfeuffer:
rebuild jenkins

jpfeuffer:
rebuild jenkins

jpfeuffer:
rebuild jenkins

oliveralka:
@jpfeuffer Der Win fail könnte ggf. an abibuilder8 liegen? Oder meinst du das liegt an C++17

timosachsenberg:
We include these seqan header on the top level.
```
#include <seqan/index.h>
#include <seqan/align.h>
#include <seqan/graph_align.h>

#include <seqan/basic.h>
#include <seqan/stream.h>
#include <seqan/seq_io/guess_stream_format.h>
#include <seqan/seq_io/read_fasta_fastq.h>
```

jpfeuffer:
Yes, it is because abibuilder8 should not have the "openms" label, until it is compatible with the current jobs.

timosachsenberg:
rebuild jenkins

jpfeuffer:
Ok Visual Studio out of heap space.....

timosachsenberg:
Does set PreferredToolArchitecture=x64
devenv help? It uses the 64Bit compiler

cbielow:
> 
> 
> Does set PreferredToolArchitecture=x64
> devenv help? It uses the 64Bit compiler

The cmake flag `-T host=x64 ` should already take care of this (if OpenMS was configured according to install instructions :))

jpfeuffer:
https://github.com/OpenMS/build-scripts/blob/d84eb8973f51e5f270c76ddca82891a320089582/cmakeScripts/run_tests.cmake#L282

timosachsenberg:
hmm this is C1002 some suggest following fix (https://social.msdn.microsoft.com/Forums/sqlserver/en-US/d2c4bb60-e558-4dc6-a0ba-47611d45bc86/c1002-compiler-is-out-of-heap-space-in-pass-2?forum=vcgeneral)
"Link Time Code Generation" should be set to "Profile Guided Optimization - Optimization (/LTCG:PGOptimize)" instead of being blank.

cbielow:
> 
> 
> hmm this is C1002 some suggest following fix (https://social.msdn.microsoft.com/Forums/sqlserver/en-US/d2c4bb60-e558-4dc6-a0ba-47611d45bc86/c1002-compiler-is-out-of-heap-space-in-pass-2?forum=vcgeneral)
> "Link Time Code Generation" should be set to "Profile Guided Optimization - Optimization (/LTCG:PGOptimize)" instead of being blank.

for PGO we'd need some test data, and it will need to run twice. Not sure this helps in any way.

I ran the branch on my machine (32 GB RAM) and it compiles just fine.
MSVC 14.27.29110
So maybe it's just the VM which needs a bit of extra RAM?

jpfeuffer:
Hm yes but I don't know if we should expect more than 8GB RAM from every developer. Not sure what GitHub Actions for the pyopenms builds will say to this as well (they have roughly 6GB)

cbielow:
looking at the peak mem usage for compiling `MSExperiment_test`, that's around 3 GB and that should work on any machine not from the stone age.
So maybe its enough to dial down on the number of parallel tasks (even though I would expect the OS to take of the swapping)...

cbielow:
another source suggest that this is a bug in visual studio 2019 version 16.1.6 and its fixed in visual studio 2019 version 16.4.
What is currently running on the build machines?
https://developercommunity.visualstudio.com/content/problem/653301/fatal-error-c1002-compiler-is-out-of-heap-space-in.html

cbielow:
rebuild jenkins

cbielow:
If I log into the Windows build machine (06), building it manually (from within VS, release build) for the class tests works just fine.
Go figure...

cbielow:
rebuild jenkins

jpfeuffer:
hehe now you killed it :(

timosachsenberg:
```
in1:  /Users/builder/jenkins/ws/openms/PR/bldtst/c797036f/build/src/tests/topp/InternalCalibration_1_models.csv.tmp   (line: 3, position/column: 5/6)
0.0, !
     -1.350175452690103, 2.193969606216441e-03, 0.0, local

in2:  /Users/builder/jenkins/ws/openms/PR/bldtst/c797036f/source/src/tests/topp/InternalCalibration_1_models.csv   (line: 3, position/column: 3/4)
0, !
   -1.05281918390288, 0.00167226980858603, 0, local
```

src/utils/DecoyDatabase.cpp
```diff
@@ -238,16 +238,15 @@ class TOPPDecoyDatabase :
           }
           vector<String> tokenized;
           boost::smatch m;
-          while (boost::regex_search(quick_seq, m, boost::regex("[^\\[]|(\\[[^\\[\\]]*\\])")))
+          while (boost::regex_search(quick_seq, m, boost::regex(R"([^\[]|(\[[^\[\]]*\]))")))
```
timosachsenberg, line 241: I love R" !

timosachsenberg (approved):
nice PR!

cbielow:
this is awesome!

PR #5141, opened 2021-01-27
fix warning 

cbielow:
`warning: iCCP: known incorrect sRGB profile` when opening TOPPView/TOPPAS because some PNGs use a shady embedded PNG color profile


timosachsenberg:
how did you fix that? just loading and saving again?

cbielow:
indeed.
Seems to be a Photoshop issue ...

jpfeuffer:
haha nice, was bugging me forever.

PR #5143, opened 2021-01-27
allow reading in a mod as protein-nterm in mzIdentML

cbielow:
# Description

# Checklist:

### DIFF
--- a/src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
@@ -990,18 +990,14 @@ namespace OpenMS
                         const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname, r, ResidueModification::C_TERM);
                         mod = m->getFullId();
                       }
-                      else if (spci->second.front().getAccession() == "MS:1002057")  // pro nterm
+                      else if (spci->second.front().getAccession() == "MS:1002057")  // protein nterm
                       {
-                        // TODO: add support for protein N-terminal modifications in unimod
-                        // ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::PROTEIN_N_TERM);
-                        const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::N_TERM);
+                        const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::PROTEIN_N_TERM);
                         mod = m->getFullId();
                       }
-                      else if (spci->second.front().getAccession() == "MS:1002058")  // pro cterm
+                      else if (spci->second.front().getAccession() == "MS:1002058")  // protein cterm
                       {
-                        // TODO: add support for protein C-terminal modifications in unimod
-                        // ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::PROTEIN_C_TERM);
-                        const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::C_TERM);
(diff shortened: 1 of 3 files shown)


jpfeuffer:
Well looks reasonable, but I think Eugen has to tell why this was not done in the first place. There must have been some reason.

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -990,18 +990,14 @@ namespace OpenMS
                         const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname, r, ResidueModification::C_TERM);
                         mod = m->getFullId();
                       }
-                      else if (spci->second.front().getAccession() == "MS:1002057")  // pro nterm
+                      else if (spci->second.front().getAccession() == "MS:1002057")  // protein nterm
```
timosachsenberg, line 993: maybe change nterm-> N-term

src/tests/class_tests/openms/source/MzIdentMLFile_test.cpp
```diff
@@ -72,8 +72,7 @@ START_SECTION(void load(const String& filename, std::vector<ProteinIdentificatio
 {
   std::vector<ProteinIdentification> protein_ids;
   std::vector<PeptideIdentification> peptide_ids;
-  std::vector<String> fm;
-  fm.push_back("Carbamidomethyl (C)");
+  std::vector<String> fm{"Carbamidomethyl (C)", "Xlink:DTSSP[88] (Protein N-term)"};
```
timosachsenberg, line 75: @enetz Eugen does this make sense

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -990,18 +990,14 @@ namespace OpenMS
...
                       {
-                        // TODO: add support for protein N-terminal modifications in unimod
-                        // ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::PROTEIN_N_TERM);
-                        const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::N_TERM);
+                        const ResidueModification* m = ModificationsDB::getInstance()->getModification(mname,  r, ResidueModification::PROTEIN_N_TERM);
                         mod = m->getFullId();
                       }
-                      else if (spci->second.front().getAccession() == "MS:1002058")  // pro cterm
+                      else if (spci->second.front().getAccession() == "MS:1002058")  // protein cterm
```
timosachsenberg, line 998:
```suggestion
                      else if (spci->second.front().getAccession() == "MS:1002058")  // protein C-term
```

timosachsenberg (commented):
looks good but as Julianus pointed out we would need some feedback from Eugen

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
timosachsenberg: these comments were just part of the session where I showed how to comment on PR for beginner. Feel free to ignore

src/tests/class_tests/openms/source/MzIdentMLFile_test.cpp
enetz: In the tests further below (line 96) only one of the fixed modifications is checked with `fm.back()`. That was fine before, since there was only one modification. Now it might be a better idea to check both of them.
Otherwise this PR looks fine to me.

timosachsenberg: Wasn't there some other problem that prevented us to include the protein terminal ones? I vaguely recall there were some discussions quite a while ago.

PR #5145, opened 2021-01-30
[FIX] some smaller fixes for the ID-based filtering during linking

jpfeuffer:
# Description

Some corner cases that I missed.


src/openms/source/ANALYSIS/MAPMATCHING/QTClusterFinder.cpp
```diff
@@ -249,9 +249,10 @@ namespace OpenMS
       std::merge(tmp_diffs.begin(), tmp_diffs.end(), last_tmp_diffs.begin(), last_tmp_diffs.end(), std::back_inserter(last_and_before_diffs));
       if (!last_and_before_diffs.empty())
       {
-        q2 = Math::quantile(tmp_diffs, 0.5);
-        q3 = Math::quantile(tmp_diffs, 0.75);
-        tol = q2 + 2. * 1.4826 * (q3-q2);
+        q2 = Math::quantile(last_and_before_diffs, 0.5);
+        q3 = Math::quantile(last_and_before_diffs, 0.75);
+        std::cout << "q2: " << q2 << " q3: " << q3 << std::endl;
```
timosachsenberg, line 254: debug code?

jpfeuffer: Yep

src/openms/include/OpenMS/MATH/MISC/MathFunctions.h
```diff
@@ -333,7 +333,7 @@ namespace OpenMS
       if (q > 1.0) q = 1.;
 
       const auto n  = x.size();
-      const auto id = (n - 1) * q;
+      const auto id = std::max(0., n * q - 1); // -1 for c++ index starting at 0
```
cbielow, line 336: this could be dangerous:
what if I pass an empty vector and `q=0`?
since n and q are unsigned then `n*q - 1` is std::numeric_limits<size_t>::max()-1 ... and we will get a segfault.
Can we:
1) throw an exception if `x` is empty (required for any value of q actually)
2) make sure `q` is a double by removing the template parameter.

PR #5146, opened 2021-01-31
[FEATURE] Show utils in ToolsDialog

dvdvgt:
# Description
This focuses on two things:
1. Making utils available for selection in ToolsDialog.
2. Increasing the performance of the compatibility checks (whether a tool or util can be applied to the current layer type) by using `QtConcurrency` i.e. threading `getParamFromIni_` so that IO action can be properly scheduled thus increasing the performance. This also makes sure the main thread isn't blocked and the GUI can still be used. All compatible tools/utils are then later added to the combo box when ready. Failures during the execution of `getParamFromIni_` are signaled to the main thread where they are added to TOPPView's LogWindow via a slot `addEntries_`.

# Checklist:

### DIFF
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -91,54 +91,53 @@ namespace OpenMS
     return temp_dir_;
   }
 
-  std::mutex mtx; // mutex for critical section
-
   String File::getExecutablePath()
   {
     // see http://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe/1024937#1024937 for more OS' (if needed)
-    static String spath = "";
     static bool path_checked = false;
-
-    const std::lock_guard<std::mutex> lock(mtx);
-
-    // short route. Only inquire the path once. The result will be the same every time.
-    if (path_checked) 
+    static String spath = []() -> String
     {
-      return spath;
-    }
+      String rpath = "";
+      // short route. Only inquire the path once. The result will be the same every time.
+      if ((path_checked))
+      {
+        return spath;
+      }
 
-    char path[1024];
+      char path[1024];
 
 #ifdef OPENMS_WINDOWSPLATFORM
-    int size = sizeof(path);
+        int size = sizeof(path);
     if (GetModuleFileName(NULL, path, size))
 #elif  defined(__APPLE__)
-    uint size = sizeof(path);
+        uint size = sizeof(path);
     if (_NSGetExecutablePath(path, &size) == 0)
 #else // LINUX
-    int size = sizeof(path);
-    int ch = readlink("/proc/self/exe", path, size);
-    if (ch != -1)
+        int size = sizeof(path);
+        int ch = readlink("/proc/self/exe", path, size);
+        if (ch != -1)
 #endif
-    {
-      spath = File::path(String(path));
-      if (File::exists(spath)) // check if directory exists
-      {
-        // ensure path ends with a "/", such that we can just write path + "ToolX", and to not worry about if its empty or a path.
-        spath.ensureLastChar('/');
-      }
-      else
-      {
-        std::cerr << "Path extracted from Executable Path does not exist! Returning empty string!\n";
-        spath = "";
-      }
-    }
-    else
-    {
-      std::cerr << "Cannot get Executable Path! Not using a path prefix!\n";
-    }
+        {
+          rpath = File::path(String(path));
+          if (File::exists(rpath))
+          {
+            // ensure path ends with a "/", such that we can just write path + "ToolX", and to not worry about if its empty or a path.
+            rpath.ensureLastChar('/');
+          }
+          else
+          {
+            std::cerr << "Path extracted from Executable Path does not exist! Returning empty string!\n";
+            rpath = "";
+          }
+        }
+        else
+        {
+          std::cerr << "Cannot get Executable Path! Not using a path prefix!\n";
+        }
 
-    path_checked = true; // enable short route for next run
+        path_checked = true;
+        return rpath;
+    }();
     return spath;
   }
 
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -99,12 +99,11 @@ namespace OpenMS
     static String spath = "";
     static bool path_checked = false;
 
-    mtx.lock();
+    const std::lock_guard<std::mutex> lock(mtx);
 
     // short route. Only inquire the path once. The result will be the same every time.
     if (path_checked) 
     {
-      mtx.unlock();
       return spath;
     }
 
@@ -140,7 +139,6 @@ namespace OpenMS
     }
 
     path_checked = true; // enable short route for next run
-    mtx.unlock();
     return spath;
   }
 
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -48,6 +48,8 @@
 #include <QtCore/QDir>
 #include <QtNetwork/QHostInfo>
 
+#include <mutex>
+
 #ifdef OPENMS_WINDOWSPLATFORM
 #include <Windows.h> // for GetCurrentProcessId() && GetModuleFileName()
 #endif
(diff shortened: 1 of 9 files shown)


src/openms/source/SYSTEM/File.cpp
```diff
@@ -89,14 +91,21 @@ namespace OpenMS
...
 
+  std::mutex mtx; // mutex for critical section
+
   String File::getExecutablePath()
   {
     // see http://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe/1024937#1024937 for more OS' (if needed)
     static String spath = "";
     static bool path_checked = false;
 
+    const std::lock_guard<std::mutex> lock(mtx);
```
cbielow, line 102: I was thinking about a more elegant solution, and I think there is one:
the `static` variable is already protected by a mutex implicitly added by the compiler.
We can make use of this by using an `immediately evaluated lambda` function, like here:
https://stackoverflow.com/questions/15581662/is-it-possible-to-initialize-static-variable-with-lambda

thus, no need for a mutex. and the code becomes easy:
```
String File::getExecutablePath()
{	  
  static String spath = ()[]{ ... code to create path here ... }();
  return spath;
}
```

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
```diff
@@ -86,36 +95,40 @@ namespace OpenMS
...
+    }
+    for (const auto& pair : utils)
+    {
+      keys.push_back(pair.first);
     }
+    // Check whether tools/utils are compatible with current layer in background and add all compatible tools/utils
+    // to the tools_combo_ when ready (addEntries_)
+    connect(&param_watcher_, &QFutureWatcher<Param>::finished, this, &ToolsDialog::addEntries_);
+    // Warnings shall be send to TOPPView's LogWindow i.e. the given LogWindow (log_)
+    connect(this, &ToolsDialog::warning, this, &ToolsDialog::logWarning_);
```
cbielow, line 124: using signal/slot for internal communication is a bit odd...

Either directly call `logWarning_` instead of emitting, or emit the signal and connect to the logwindow outside of the class.

dvdvgt: If I'd call `logWarning_` directly each thread would call that function meaning multiple threads access the LogWindow and changing widgets is only permitted for the main thread, no?
Using signals each thread can notify the main thread to log the error so that only the main thread accesses the log widget. That's the reason why I chose to do it that way.

cbielow: ah.. I did not catch that. all good. sorry for the confusion.

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
```diff
@@ -420,6 +434,24 @@ namespace OpenMS
     }
   }
 
+  void ToolsDialog::addEntries_()
+  {
+    QStringList names;
+    for (const auto& p : param_future_.results())
+    {
+      // Extract the tool's name
+      String name = p.begin().getName().substr(0, p.begin().getName().rfind(":"));
```
cbielow, line 443: does this still work if the INI file was not read correctly? Then the param object is empty, right?
Also, you only need that if the layer_type_ matches. So evaluation can be deferred to the actual push_back into `names`?

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
```diff
@@ -86,36 +95,40 @@ namespace OpenMS
...
+    for (const auto& pair : utils)
+    {
+      keys.push_back(pair.first);
     }
+    // Check whether tools/utils are compatible with current layer in background and add all compatible tools/utils
+    // to the tools_combo_ when ready (addEntries_)
+    connect(&param_watcher_, &QFutureWatcher<Param>::finished, this, &ToolsDialog::addEntries_);
+    // Warnings shall be send to TOPPView's LogWindow i.e. the given LogWindow (log_)
+    connect(this, &ToolsDialog::warning, this, &ToolsDialog::logWarning_);
+    param_future_ = QtConcurrent::mapped(keys, std::bind(&ToolsDialog::getParamFromIni_, this, std::placeholders::_1));
```
cbielow, line 125: not sure if the dependency on `QtConcurrent` is worth the trouble here.
@jpfeuffer does this have implications on packaging for e.g. our installers, conda, KNIME ... etc? It certainly has for the developer installation I imagine.

To avoid this, I think it's possible to do this using std::async, even though the callback is not as convenient e.g.
```
std::vector<std::future<Param>> result;
for (const QString& key : keys)
{
	result.push_back( std::move(std::async(std::launch::async, []() { return getParamFromIni_(key); });) );
}

for (auto& r : results)
{
	while (r.wait_for(std::chrono::milliseconds(10) != std::future_status::ready))
	{ // a bit of a crutch... (a callback would be better, but its a lot more code to get right)
		QCoreApplication::processEvents();
	}
}
// all params are available now ... evaluate them...
```

jpfeuffer: Hmm yes. We kind of automated the dependency deduction and copying of the DLLs but it is not very robust so it would need to be tested immediately after the merge. But if there is a solution that avoids using another dependency for one small part of the code, I would be in favor of that.

src/openms/source/SYSTEM/File.cpp
timosachsenberg: @cbielow  what happens if an exception is thrown? What is the state of the static?

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
timosachsenberg: Hmm does async have a fixed threadpool or just spawn >hundred. Is this an issue? @cbielow

src/openms/source/SYSTEM/File.cpp
cbielow: its still uninitialized and the next call will try again.

src/openms_gui/source/VISUAL/DIALOGS/ToolsDialog.cpp
cbielow: good point.
MSVC uses a threadpool, if i remember correctly.
And there is no build-in way to limit the number of threads.
Most linuxes will spawn several thousand. MSVC spawns twice as many as it detects CPU-cores (last time I tested).
So you might actually block the GUI thread...
The only save way to do the things here, would be to use std::hardware_concurrency and spawn only a limited number the async's, each getting assigned its share of TOPP Tools (more than one). The workload is fairly balanced I would imagine, since querying an INI should be dominated by the Tool starting up.

We might also get rid of this approach entirely, if we make sure that INI files for all TOPP tools are statically stored somewhere in OpenMS/share.. for example. This would need a test, which makes sure they are all up-to-date though... and that would run rather slow, but at least, we could read INI files directly. Just an idea...
How long does it take at the moment to populate the list?

timosachsenberg: depends on hdd type. Maybe 1-5s. The idea was to make this code multi-threaded and then later move it to the start of TOPPView so parsing is done during startup and should be finished when the user opens the tool dialog.

timosachsenberg: We used QConcurrency to have Qt handle all threads in a global thread-pool and we don't mix any multi-threading mechanisms.

timosachsenberg:
So what should we do here @cbielow and @jpfeuffer.
Fewer dependencies but more complex code and loss of thread pool management?
I don't have a strong opinion, but my guess is that it might take some extra work to find out if it will work with the std way?

cbielow:
I personally would not worry about thread-pool management.
We are talking about spawning 64 threads at most (if you have a really good machine) -- and that is done in 'no time' -- provided you chunk the list into as many sublists as there are CPU's on the system and then run `async` on them.
This can be done in like 20 lines of code.
So unless there are other applications which might come up for QtConcurrent in the future, I'd say: go native C++.
The solution is almost ready up there...

dvdvgt:
I will submit a new PR with the requested changes and leave this branch unchanged for now in case we might want to come back to use QtConcurrency over std::async after all.

PR #5154, opened 2021-02-04
Fix/mascot - spectral batch size option - not yet ready

Leon-Bichmann:
# Description

This PR should introduce a spectra batch size option fo the Mascot Adapter for it to process only a given number of spectra at a time, in case the files are too large to be processed in one go.

# Checklist:

### DIFF
--- a/src/topp/MascotAdapter.cpp
+++ b/src/topp/MascotAdapter.cpp
@@ -265,6 +265,7 @@ protected:
     registerStringOption_("boundary", "<string>", "", "MIME boundary for mascot output format", false);
     registerStringOption_("mass_type", "<type>", "Monoisotopic", "mass type", false);
     setValidStrings_("mass_type", ListUtils::create<String>("Monoisotopic,Average"));
+    registerIntOption_("mascot_spectral_batch_size", "<num>", 50000, "number of spectra processed in one batch by mascot (default 50000)", false);
     registerStringOption_("mascot_directory", "<dir>", "", "the directory in which mascot is located", false);
     registerStringOption_("temp_data_directory", "<dir>", "", "a directory in which some temporary files can be stored", false);
   }
@@ -419,6 +420,7 @@ protected:
       return ILLEGAL_PARAMETERS;
     }
 
+    n_chunks = getIntOption_("mascot_spectral_batch_size");
 
     if (mascot_in)
     {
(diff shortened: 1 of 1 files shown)


Leon-Bichmann:
Hey, so its not tested yet, I just opened it for now and will do further tests tomorrow.

Leon-Bichmann:
ok seems like there is sth. wrong I get the following error when trying to build it locally:

```ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/libOpenMS.dylib] Error 1
make[1]: *** [src/openms/CMakeFiles/OpenMS.dir/all] Error 2
make: *** [all] Error 2```

File: src/topp/MascotAdapter.cpp [hunk 1]
```diff
@@ -487,10 +487,35 @@ class TOPPMascotAdapter :
 
     if (!mascot_out)
     {
+    // determine number of chunks of size n
+    int size = (experiment.size() - 1) / n + 1;
+ 
+    // create array of vectors to store the chunks
+    std::vector<PeakMap> split_experiment[size];
```
Leon-Bichmann, line 496: is std::vector<PeakMap> a valid type specification?

cbielow: sure. If `vector` is `#include`d, either directly or indirectly :)

src/topp/MascotAdapter.cpp
```diff
@@ -487,10 +489,35 @@ class TOPPMascotAdapter :
...
+
+    for (int k = 0; k < size; ++k)
+    {
+        // get range for next set of n elements
+        auto start_itr = std::next(experiment.cbegin(), k*n_chunks);
+        auto end_itr = std::next(experiment.cbegin(), k*n_chunks + n_chunks);
+
+        // allocate memory for the chunk
+        split_experiment[k].resize(n_chunks);
```
cbielow, line 505: this is too much for the last bin, in most cases.

you can resize for all bins, based on `.resize(end_itr - start_itr);`

Leon-Bichmann: alright, thx i'll take a look at it next week again

Leon-Bichmann: ok so I thought the if clause would be fine to resize the last chunk. If I do it instead with the `end_itr - start_itr`. Can I use the same iterators twice?

timosachsenberg:
wrong file ;)

PR #5157, opened 2021-02-05
A searchengine base

cbielow:
# Description

Being annoyed by empty search results in MSGF+ without apparent reason, it turns out that the adapter and MSGF+ itself dont check for MS2 spectra being centroided. It will happily search profile spectra and give you: nothing (not even an error).

We had this functionality once, but it got buried in legacy code from TSVConversion inside the tool -- rather uncool.

Now, this is a first attempt at unifying some basic behaviour.
It comes at the cost of supporting fewer input formats for MSGF+, but I'm totally happy about that as long as it's safer to use (this bug just cost me a day!).
The scheme can (and should) be extended in the future to use common search parameters, where it makes sense...


src/openms/include/OpenMS/APPLICATIONS/SearchEngineBase.h
```diff
@@ -0,0 +1,104 @@
...
+
+      @param citations Add one or more citations if they are associated specifically to this TOPP tool; they will be printed during --help
+    */
+    SearchEngineBase(const String& name, const String& description, bool official = true, const std::vector<Citation>& citations = {}, bool toolhandler_test = true);
+
+    /// Destructor
+    virtual ~SearchEngineBase();
+
+    /**
+      @brief Reads the '-in' argument from internal parameters (i.e. the mzML file) and checks if MS2 spectra are present and are centroided. If no MS2 or profile MS2 data is found, an exception is thrown.
```
timosachsenberg, line 80: does two things at once

src/openms/include/OpenMS/APPLICATIONS/SearchEngineBase.h
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+      @brief Reads the '-database' argument from internal parameters (or from @p db) and tries to find the db in search directories (if it cannot be found immediately). If not found, an exception is thrown.
+      
+      @param db [Optional] Instead of reading the '-database', you can provide a custom name here (might be required for special db formats, see OMSSA)
+      @return filename for DB (might be a relative or absolute path)
+ 
+      @throws OpenMS::Exception::FileNotFound if database name could not be resolved
+    */
+    String getDBFilename(String db = "") const;
```
timosachsenberg, line 101: is = "" needed?

src/openms/include/OpenMS/FORMAT/MzMLFile.h
```diff
@@ -175,16 +175,25 @@ namespace OpenMS
     */
     bool isSemanticallyValid(const String& filename, StringList& errors, StringList& warnings);
 
+
+    struct SpecInfo
+    {
+      int count_centroided = 0;
```
timosachsenberg, line 181: reason why it is an  int?

src/openms/include/OpenMS/FORMAT/MzMLFile.h
```diff
@@ -175,16 +175,25 @@ namespace OpenMS
     */
     bool isSemanticallyValid(const String& filename, StringList& errors, StringList& warnings);
 
+
+    struct SpecInfo
```
timosachsenberg, line 179: better name? seems to be tied to getCentroidInfo

timosachsenberg: put Tandem somewhere? TandemSpecCounts ?

src/openms/source/APPLICATIONS/SearchEngineBase.cpp
```diff
@@ -0,0 +1,82 @@
...
+    TOPPBase(tool_name, tool_description, official, citations, toolhandler_test)
+  {
+  }
+
+  SearchEngineBase::~SearchEngineBase() {}
+
+  
+  String SearchEngineBase::getRawfileName(int ms_level) const
+  {
+    String inputfile_name = getStringOption_("in");
```
timosachsenberg, line 54: what if it is not an mzML?

src/topp/MSGFPlusAdapter.cpp
```diff
@@ -156,7 +156,7 @@ class MSGFPlusAdapter :
   void registerOptionsAndFlags_() override
   {
     registerInputFile_("in", "<file>", "", "Input file (MS-GF+ parameter '-s')");
-    setValidFormats_("in", ListUtils::create<String>("mzML,mzXML,mgf,ms2"));
+    setValidFormats_("in", {"mzML"});
```
timosachsenberg, line 159: this might make some people sad

timosachsenberg (commented):
Nice PR. Though we I think we need to make sure that stuff that worked before still works (e.g. ,zXML input?)

src/openms/include/OpenMS/APPLICATIONS/SearchEngineBase.h
cbielow: exactly the intend! :)

src/openms/include/OpenMS/APPLICATIONS/SearchEngineBase.h
cbielow: yes. For OMSSA, since we check the existance of the `.psq` file, even if only the fasta filename was provided in `-database`.

src/openms/include/OpenMS/FORMAT/MzMLFile.h
cbielow: usually the fastest type on a platform. nothing else.
Or would you rather prefer `UInt`?

src/openms/include/OpenMS/FORMAT/MzMLFile.h
cbielow: this is perfectly applicable to MS1 as well...

src/openms/include/OpenMS/FORMAT/MzMLFile.h
timosachsenberg: it counts so I would say size_t

src/openms/source/APPLICATIONS/SearchEngineBase.cpp
cbielow: then it does not work. 
:/

But the TOPP interface only speaks mzML that reason everywhere.
The thing is, the streaming consumers, are only available for mzML, and there is no fast way of checking for spectrum type for other formats. So, instead of taking the performance it, I'd gladly take the '1 format only` hit, which is the case for the majority of all TOPP tools anyway.
Its easy to fix via an upstream FileConverter.

src/topp/MSGFPlusAdapter.cpp
cbielow: yeah. Maybe.
But getting empty results (no IDs) without knowing why, might make them even more sad (and angry).

cbielow: Also, this makes the Adapter not easily swappable for other adapters, which do not support extra formats.

timosachsenberg: I see the reasons but still think that mzML should be encouraged (maybe even warned if something else is used). This will break existing workflows which will make users even more sad and angry ;)

src/openms/include/OpenMS/APPLICATIONS/SearchEngineBase.h
cbielow: ahh sorry, I misread. You mean, one has to pass the database filename explicitly for every adapter? Yeah... certainly possible.
I went with a default, so avoid repetitive code. But I have no strong feelings here.

cbielow:
rebuild jenkins

cbielow:
@jpfeuffer is there a way to restart only one jenkins build? Re-running already successful builds seems inefficient...

jpfeuffer:
no, unfortunately the Jenkins GitHub plugin does not support that. we would need to create a job for each platform, which is inefficient in more expensive manpower.

cbielow:
I see, thanks!

jpfeuffer:
There could be a workaround by putting a "matrix-subset" file into the repo. But then this would need to be edited for re-runs everytime. And edited back before merge I guess.
I am pretty sure you cannot extract keywords from the trigger phrase.

jpfeuffer:
Hmm maybe some Groovy script magic parsing the GitHub JSON payload?
https://gist.github.com/jimCresswell/7038683

No easy solution though.

cbielow:
good to merge.

PR #5161, opened 2021-02-08
[FEATURE] Allow skipping MS2 isotope scores in SWATH and disable it in FFID.

jpfeuffer:
# Description

See title. Let's see if we can get some speed back after activating the MS1 isotope scores.


timosachsenberg (approved):
wonder what the scores meant before

jpfeuffer:
I mean the MS2 scores basically lay an isotope profile over each product transition and score it.
But a) they were bugged (always using charge 4 and spacing 1 Da) and b) when your product transitions are basically the isotopes (as in FFID) you will have overlaps everywhere, and every product transition will look like it is part of a bigger profile (pre-isotope peaks are scored less/negatively).

PR #5167, opened 2021-02-12
Writting Spectrum Reference in NovorAdapter

Waschi97:
NovorAdapter used to just write the scan number as a MetaValue.
Using `SpectrumLookup()` now the spectrum reference is written.


timosachsenberg (approved):
nice! thanks

cbielow:
wait :)

cbielow:
the current solution loads the whole MSExperiment including data. should be restricted to level2 and without the actual peaks.

timosachsenberg:
Ah right that will make some difference.

Leon-Bichmann (approved):

PR #5169, opened 2021-02-16
Second Compute Function for FragmentMassError

Waschi97:
Added an additional compute function to `FragmentMassError` allowing for a `vector<PeptideIdentification>` input.
Also added a test.


src/openms/include/OpenMS/QC/FragmentMassError.h
```diff
@@ -90,6 +92,27 @@ namespace OpenMS
      */
     void compute(FeatureMap& fmap, const MSExperiment& exp, const QCBase::SpectraMap& map_to_spectrum, ToleranceUnit tolerance_unit = ToleranceUnit::AUTO, double tolerance = 20);
 
+    /**
+      * @brief computes FragmentMassError in ppm and Dalton (only of the first PeptideHit of each PepID)
+      *
+      * Stores average and variance of FragmentMassErrors in ppm as a struct in a vector.
+      * Each FragmentMassError (in ppm) is stored in the first PeptideHit of the corresponding PeptideIdentification as metavalue "fragment_mass_error_ppm".
```
timosachsenberg, line 99: The documentation could be improved as it leaves a lot of room for interpretation what is calculated and how it is stored.

E.g. is the only the average and variance stored? Average over what? All peptide hits of a spectrum? "as a struct in a vector" needs to be more precise. Is the FragmentMassError (in ppm) only calculated for the first PeptideHit or the average over all or both? etc.

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,6 +108,105 @@ namespace OpenMS
...
+    }
+
+    //---------------------------------------------------------------------
+    // FIND DATA FOR THEORETICAL SPECTRUM
+    //---------------------------------------------------------------------
+
+    // sequence
+    const AASequence& seq = pep_id.getHits()[0].getSequence();
+
+    // charge: re-calulated from masses since much more robust this way (PepID annotation of pep_id.getHits()[0].getCharge() could be wrong)
```
timosachsenberg, line 126: is this often the case? for which search engines?

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,6 +108,105 @@ namespace OpenMS
...
+
+    //---------------------------------------------------------------------
+    // CREATE THEORETICAL SPECTRUM
+    //---------------------------------------------------------------------
+    PeakSpectrum theo_spectrum = getTheoSpec_(act_method, seq, charge);
+
+    //-----------------------------------------------------------------------
+    // COMPARE THEORETICAL AND EXPERIMENTAL SPECTRUM
+    //-----------------------------------------------------------------------
+    if (exp_spectrum.empty() || theo_spectrum.empty())
```
timosachsenberg, line 163: does theo_spectrum.empty() need to be tested? are there corner cases you encountered where this happened?

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,6 +108,105 @@ namespace OpenMS
...
+    {
+      using MIV = MatchedIterator<MSSpectrum, PpmTrait, true>;
+      MIV mi(theo_spectrum, exp_spectrum_filtered, tolerance);
+      twoSpecErrors(mi, ppms, dalton, accumulator_ppm, counter_ppm);
+    }
+
+    //-----------------------------------------------------------------------
+    // WRITE PPM ERROR IN PEPTIDEHIT
+    //-----------------------------------------------------------------------
+    pep_id.getHits()[0].setMetaValue("fragment_mass_error_ppm", ppms);
```
timosachsenberg, line 193: Please define string for these user values in Constants.h around here https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/CONCEPT/Constants.h#L280

timosachsenberg (changes requested):
nice contribution!

timosachsenberg:
Another minor thing: FragmentMassError::FMEStatistics the FME seems to be redundant (also in computeFME). Doesn't really add value here.

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,6 +108,105 @@ namespace OpenMS
...
+  void calculateVariance_(FragmentMassError::FMEStatistics& result, const PeptideIdentification& pep_id)
+  {
+    if (pep_id.getHits().empty())
+    {
+      OPENMS_LOG_WARN << "There is a Peptideidentification(RT: " << pep_id.getRT() << ", MZ: " << pep_id.getMZ() << ") without PeptideHits. " << "\n";
+      return;
+    }
+    for (const auto& ppm : (pep_id.getHits()[0].getMetaValue("fragment_mass_error_ppm")).toDoubleList())
+    {
+      result.variance_ppm += pow((ppm - result.average_ppm), 2);
```
cbielow, line 206: replace with `x*x`

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -161,123 +260,106 @@ namespace OpenMS
...
+    result.average_ppm = accumulator_ppm / counter_ppm;
 
-      // sequence
-      const AASequence& seq = pep_id.getHits()[0].getSequence();
+    // computes variance
+    fmap.applyFunctionOnPeptideIDs(fVar);
 
-      // charge: re-calulated from masses since much more robust this way (PepID annotation of pep_id.getHits()[0].getCharge() could be wrong)
-      Int charge = static_cast<Int>(round(seq.getMonoWeight() / pep_id.getMZ()));
+    result.variance_ppm = result.variance_ppm / counter_ppm;
```
Waschi97, line 287: Move to `calculateVariance()`

src/openms/source/QC/FragmentMassError.cpp
Waschi97: I just copied existing code. So I really don't know why this is necessary..

src/openms/source/QC/FragmentMassError.cpp
Waschi97: same here

src/openms/include/OpenMS/CONCEPT/Constants.h
```diff
@@ -279,6 +279,10 @@ namespace OpenMS
       */
       extern OPENMS_DLLAPI const std::string   FRAGMENT_ERROR_MEDIAN_PPM_USERPARAM;
 
+      extern OPENMS_DLLAPI const std::string   FRAGMENT_ERROR_PPM_METAVALUE;
```
timosachsenberg, line 282: don't want to be pickt but we use _USERPARAM
Is the FRAGMENT_ERROR_PPM_METAVALUE for a single peak or e.g. the mean? if it is the mean I would
put _MEAN_ similar as we did for   FRAGMENT_ERROR_MEDIAN_PPM_USERPARAM;

src/openms/include/OpenMS/QC/FragmentMassError.h
```diff
@@ -128,7 +128,7 @@ namespace OpenMS
 
   private:
     /// container that stores results
-    std::vector<FMEStatistics> results_{};
+    std::vector<Statistics> results_{};
```
timosachsenberg, line 131: {} not needed

src/openms/include/OpenMS/CONCEPT/Constants.h
Waschi97: Alright, I'll add _USERPARAM.

It isn't the mean over a whole spectrum though. The Metavalue stores a list of all ppms (one for each peak in the spectrum).

jpfeuffer (approved):

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,10 +108,111 @@ namespace OpenMS
...
+      using MIV = MatchedIterator<MSSpectrum, PpmTrait, true>;
+      MIV mi(theo_spectrum, exp_spectrum_filtered, tolerance);
+      twoSpecErrors(mi, ppms, dalton, accumulator_ppm, counter_ppm);
+    }
+
+    //-----------------------------------------------------------------------
+    // WRITE PPM ERROR IN PEPTIDEHIT
+    //-----------------------------------------------------------------------
+    pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_PPM_METAVALUE_USERPARAM, ppms);
+    pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_DA_METAVALUE_USERPARAM, dalton);
```
timosachsenberg, line 194: do we also add variance as meta value? would it make sense

timosachsenberg: is it documented that these values are annotated / overwritten if already existing?

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -274,19 +286,80 @@ namespace OpenMS
...
+    // filter settings
+    WindowMower window_mower_filter;
+    Param filter_param = window_mower_filter.getParameters();
+    filter_param.setValue("windowsize", 100.0, "The size of the sliding window along the m/z axis.");
+    filter_param.setValue("peakcount", 6, "The number of peaks that should be kept.");
+    filter_param.setValue("movetype", "jump", "Whether sliding window (one peak steps) or jumping window (window size steps) should be used.");
+    window_mower_filter.setParameters(filter_param);
+
+    //-------------------------------------------------------------------
+    // find tolerance unit and value
```
timosachsenberg, line 321: please document that the tolerance parameter is overwritten if ToleranceUnit::AUTO is chosen. could be a bit surprisong

timosachsenberg (changes requested):

src/openms/source/QC/FragmentMassError.cpp
Waschi97: Currently variance is only calculated over all peptide identifications. But it would make sense to also have a variance for each hit, I'll add it.

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -192,9 +192,15 @@ namespace OpenMS
     // WRITE PPM ERROR IN PEPTIDEHIT
     //-----------------------------------------------------------------------
     pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_PPM_METAVALUE_USERPARAM, ppms);
-    pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_PPM_METAVALUE_USERPARAM + "_variance", Math::variance(ppms.begin(), ppms.end(), Math::mean(ppms.begin(), ppms.end())));
     pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_DA_METAVALUE_USERPARAM, dalton);
-    pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_DA_METAVALUE_USERPARAM + "_variance", Math::variance(dalton.begin(), dalton.end(), Math::mean(dalton.begin(), dalton.end())));
+    if (!ppms.empty())
```
timosachsenberg, line 196: can you check if variance works if only one value is present?

https://github.com/OpenMS/OpenMS/blob/55598cf39c73a2092365fbdb51e63defa9b29c6f/src/openms/include/OpenMS/MATH/STATISTICS/StatisticFunctions.h#L292

Waschi97: Well, it doesn't throw an error. It just returns `-nan(ind)`. But maybe I should still prevent that from happening...

axelwalter (approved):

src/openms/include/OpenMS/QC/FragmentMassError.h
```diff
@@ -64,24 +66,26 @@ namespace OpenMS
...
 
     /**
-     * @brief computes FragmentMassError in ppm and Dalton (only of the first PeptideHit of each PepID)
+     * @brief computes FragmentMassError (FME) in ppm and Dalton (only of the first PeptideHit of each PepID)
      *
-     * Stores average and variance of FragmentMassErrors in ppm as a struct in a vector.
-     * Each FragmentMassError (in ppm) is stored in the first PeptideHit of the corresponding PeptideIdentification as metavalue "fragment_mass_error_ppm".
-     * Each FragmentMassError (in Da) is stored in the first PeptideHit of the corresponding PeptideIdentification as metavalue "fragment_mass_error_da".
+     * Stores average FME over all spectra (one for each PeptideIdentification) and its variance in ppm as a struct in a vector.
+     * Each FME (in ppm) is stored at the first PeptideHit of the corresponding PeptideIdentification as metavalue Constants::UserParam::FRAGMENT_ERROR_PPM_METAVALUE_USERPARAM
```
timosachsenberg, line 79: add to doc in what cases variance is not stored

timosachsenberg (changes requested):
is the contrib up to date?
TODO:
I think pyOpenMS bindings are missing

jpfeuffer:
Yes contrib is uptodate. Coinor was recent.

jpfeuffer:
Regarding pyopenms. The whole QC folder does not have python bindings. Not sure if this should be done here.

timosachsenberg:
> Regarding pyopenms. The whole QC folder does not have python bindings. Not sure if this should be done here.

I think we should. Having one pxd template will make it significantly easier to add the others by mostly copying and pasting

src/openms/include/OpenMS/CONCEPT/Constants.h
```diff
@@ -279,6 +279,10 @@ namespace OpenMS
       */
       extern OPENMS_DLLAPI const std::string   FRAGMENT_ERROR_MEDIAN_PPM_USERPARAM;
 
+      extern OPENMS_DLLAPI const std::string   FRAGMENT_ERROR_PPM_METAVALUE_USERPARAM;
```
timosachsenberg, line 282: please add documentation and rename to FRAGMENT_ERROR_PPM_USERPARAM;

src/openms/include/OpenMS/CONCEPT/Constants.h
```diff
@@ -279,6 +279,10 @@ namespace OpenMS
...
+
+      extern OPENMS_DLLAPI const std::string   FRAGMENT_ERROR_DA_METAVALUE_USERPARAM;
```
timosachsenberg, line 284: - is this DA or m/z ?
- please also rename to remove redundant metavalue + userparam

src/openms/source/CONCEPT/Constants.cpp
```diff
@@ -169,6 +169,12 @@ namespace OpenMS
       // User parameter name for fragment mz error in ppm
       const std::string FRAGMENT_ERROR_MEDIAN_PPM_USERPARAM = "fragment_mz_error_median_ppm";
 
+      // User metavalue name for fragment mz error in ppm
+      const std::string FRAGMENT_ERROR_PPM_METAVALUE_USERPARAM = "fragment_mass_error_ppm";
```
timosachsenberg, line 173: please also add doc to .h file

timosachsenberg: see other comment

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,10 +109,118 @@ namespace OpenMS
...
+
+    if (!pep_id.metaValueExists("spectrum_reference"))
+    {
+      throw Exception::InvalidParameter(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No spectrum reference annotated at peptide identifiction!");
+    }
+    const MSSpectrum& exp_spectrum = exp[map_to_spectrum.at(pep_id.getMetaValue("spectrum_reference").toString())];
+
+    if (exp_spectrum.getMSLevel() != 2)
+    {
+      throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "The matching spectrum of the mzML is not an MS2 Spectrum.");
```
timosachsenberg, line 142:
```suggestion
      throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "Spectrum with wrong MS level provided. MS2 expected.");
```

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,10 +109,118 @@ namespace OpenMS
...
+    }
+    Precursor::ActivationMethod act_method;
+    if (exp_spectrum.getPrecursors().empty() || exp_spectrum.getPrecursors()[0].getActivationMethods().empty())
```
timosachsenberg, line 145: did you check if there is at least one precursor annotated?

src/openms/include/OpenMS/QC/FragmentMassError.h
```diff
@@ -90,11 +96,36 @@ namespace OpenMS
      */
     void compute(FeatureMap& fmap, const MSExperiment& exp, const QCBase::SpectraMap& map_to_spectrum, ToleranceUnit tolerance_unit = ToleranceUnit::AUTO, double tolerance = 20);
```
timosachsenberg, line 97: can this method be const?

src/openms/include/OpenMS/QC/FragmentMassError.h
```diff
@@ -90,11 +96,36 @@ namespace OpenMS
...
+     * @param map_to_spectrum Map to find index of spectrum given by meta value at PepID
+     * @param tolerance_unit Tolerance in ppm or Dalton (if auto was chosen, the unit and value will taken from FeatureMap metadata)
+     * @param tolerance Search window for matching peaks; distance has to be lower than tolerance value (Will be overwritten if tolerance_unit AUTO is chosen)
+     * @throws Exceptions::MissingInformation If fragment mass tolerance is missing in @search_params
+     * @throws Exception::InvalidParameter PeptideID is missing meta value 'spectrum_reference'
+     * @throws Exception::IllegalArgument Spectrum for a PepID has ms-level of 1
+     * @throws Exception::MissingInformation If no fragmentation method given in a MS2 precursor
+     * @throws Exception::InvalidParameter If the fragmentation method is not ECD, ETD, CID or HCD
+     */
+    void compute(std::vector<PeptideIdentification>& pep_ids, const ProteinIdentification::SearchParameters& search_params, const MSExperiment& exp, const QCBase::SpectraMap& map_to_spectrum, ToleranceUnit tolerance_unit = ToleranceUnit::AUTO, double tolerance = 20);
```
timosachsenberg, line 122: can this method be const?

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,10 +109,118 @@ namespace OpenMS
     return theo_spectrum;
   }
 
+  void calculateFME_(PeptideIdentification& pep_id, const MSExperiment& exp, const QCBase::SpectraMap& map_to_spectrum, bool& print_warning, double tolerance, FragmentMassError::ToleranceUnit tolerance_unit, double& accumulator_ppm, UInt32& counter_ppm, WindowMower& window_mower_filter)
```
timosachsenberg, line 112: make this a member function of the class not a free function in the OpenMS namespace
should probably be a static private member

src/openms/source/QC/FragmentMassError.cpp
```diff
@@ -108,10 +109,118 @@ namespace OpenMS
...
+    {
+      pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_PPM_METAVALUE_USERPARAM + "_variance", Math::variance(ppms.begin(), ppms.end()));
+    }
+    if (dalton.size() > 1)
+    {
+      pep_id.getHits()[0].setMetaValue(Constants::UserParam::FRAGMENT_ERROR_DA_METAVALUE_USERPARAM + "_variance", Math::variance(dalton.begin(), dalton.end()));
+    }
+  }
+
+  void calculateVariance_(FragmentMassError::Statistics& result, const PeptideIdentification& pep_id, const UInt num_ppm)
```
timosachsenberg, line 206: same here... should also be a static private member

timosachsenberg (changes requested):
sorry for being picky :) only minor things. good job

src/openms/include/OpenMS/QC/FragmentMassError.h
Waschi97: No, since it stores it's result in a member.

src/openms/include/OpenMS/QC/FragmentMassError.h
Waschi97: same here

timosachsenberg:
rebuild jenkins

jpfeuffer:
Oops, seems like the merge of the other PR created some conflicts with this one now

Issue #5164, opened 2021-02-11
OPENMS_PRECONDITION fails while running FeatureFinderMetabo


bertrandboudaud:
When running FeaturefinderMetabo with our data, we run into an exception, an OPENMS_PRECONDITION that is not satisfied in MassTrace::linearInterpolationAtY_:

```
OPENMS_PRECONDITION(yA < y_eval && y_eval < yB, "y_eval is not between yA and yB")
```
However, the values for yA and y_eval are equals: 284.00000000. shouldn't be considered in range in this case? or do you think it's an abnormal condition - would it be so uncommon to have a value that starts to the exact start value of the range?

The assertion passes by changing conditions to <= and everything gets executed until the end. however, since it's comparing doubles, not sure it's the best way to check this (or it should not be fixed at all, maybe).

we are launching FeatureFinderMetabo in a simple way:
FeatureFinderMetabo -in data.mzML -out out.featureXML

![OpenMSExceptionFeatureFinderMetabo](https://user-images.githubusercontent.com/1705849/107614171-11fb6580-6c4a-11eb-8169-f027a4de2023.png)

thank you for your feedback :)

jpfeuffer:
Hi! Thanks for reporting. I think you are absolutely right. An even better precondition would be to test for the more dangerous yB > yA so that it is not dividing by zero. But y_eval = yA or yB should be fine.

PR #5170, opened 2021-02-17
Extends precondition check bounds to equality

bertrandboudaud:
# Description

This refers to Issue #5164

When running FeaturefinderMetabo with our data, we run into an exception, an OPENMS_PRECONDITION that is not satisfied in MassTrace::linearInterpolationAtY_:

```
OPENMS_PRECONDITION(yA < y_eval && y_eval < yB, "y_eval is not between yA and yB")
```

However, the values for yA and y_eval are equals: 284.00000000 (screenshot attached). 

image

The assertion passes by changing conditions to <= and everything gets executed until the end. 

On the issue discussion, it is suggested only to check yB > yA to avoid division per 0, however, this situation seems to be already handled in that method.

An even better precondition would be to test for the then needed more dangerous yB > yA so that it is not dividing by zero
we are launching FeatureFinderMetabo in a simple way:
FeatureFinderMetabo -in data.mzML -out out.featureXML

# Checklist:
**-> This very minor fix, should I do a change in CHANGELOG?**

### DIFF
--- a/src/openms/source/KERNEL/MassTrace.cpp
+++ b/src/openms/source/KERNEL/MassTrace.cpp
@@ -311,12 +311,12 @@ namespace OpenMS
 
     double MassTrace::linearInterpolationAtY_(double xA, double xB, double yA, double yB, double y_eval) const
     {
-      OPENMS_PRECONDITION(yA < y_eval && y_eval < yB, "y_eval is not between yA and yB")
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Thanks a lot. Ah yes, the next line handles the yA = yEval = yB case implicitly right?

jpfeuffer:
Ignore the failing travis test.

jpfeuffer (approved):

PR #5191, opened 2021-02-25
added mzQC export for the QCCalculator

axelwalter:
# Description

Added a function for QCFile to export as mzQC in JSON format. 

# TODO:
- calculate SHA
- add option to store as qcML or mzQC
- add tests / test files

### DIFF
--- a/src/openms/source/FORMAT/QcMLFile.cpp
+++ b/src/openms/source/FORMAT/QcMLFile.cpp
@@ -39,6 +39,7 @@
 #include <boost/range/adaptor/map.hpp>
 #include <boost/range/algorithm/copy.hpp>
 
+#include <iostream>
 using namespace std;
 
 // TODO fix all the shadowed "const_iterator qpsit"
@@ -48,7 +49,7 @@ using namespace std;
 namespace OpenMS
 {
 
-  QcMLFile::QualityParameter::QualityParameter() :
+  QcFile::QualityParameter::QualityParameter() :
     name(),
     id(),
     value(),
@@ -60,7 +61,7 @@ namespace OpenMS
   {
   }
 
-  QcMLFile::QualityParameter::QualityParameter(const QualityParameter& rhs) :
+  QcFile::QualityParameter::QualityParameter(const QualityParameter& rhs) :
     name(rhs.name),
     id(rhs.id),
     value(rhs.value),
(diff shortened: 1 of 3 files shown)


src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -1202,6 +1272,150 @@ namespace OpenMS
...
+        << String(14, ' ') << "\"fileProperties\": [\n"
+        << String(16, ' ') << "{\n"
+        << String(18, ' ') << "\"accession\": \"MS:1000747\",\n" // exclude in qualityMetrics?
+        << String(18, ' ') << "\"name\": \"completion time\",\n"
+        << String(18, ' ') << "\"value\": \"" + completionTime + "\"\n"
+        << String(16, ' ') << "},\n"
+        << String(16, ' ') << "{\n"
+        << String(18, ' ') << "\"accession\": \"MS:1000569\",\n"
+        << String(18, ' ') << "\"name\": \"SHA-1\",\n"
+        << String(18, ' ') << "\"value\": \"" << "\"\n" // how to calculate SHA sum?
```
timosachsenberg, line 1376: from MSExperiment you can get the SourceFile https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/SourceFile_test.cpp#L150

timosachsenberg (changes requested):
Nice thanks for the contribution

cbielow:
interesting PR.
Can you motivate a bit why mzQC is added into the old qcML architecture?

The way I see it (and I think @mwalzer agrees): qcML is dead, so my first take would have been to start mzQC from scratch instead of trying to adapt old code. But that's just me.

We also have https://github.com/nlohmann/json already integrated into OpenMS, to immediate use.
To write a JSON string, you can just do (taken from the example section of `json`):

```
json j2 = {
  {"pi", 3.141},
  {"happy", true},
  {"name", "Niels"},
  {"nothing", nullptr},
  {"answer", {
    {"everything", 42}
  }},
  {"list", {1, 0, 2}},
  {"object", {
    {"currency", "USD"},
    {"value", 42.99}
  }}
};
```
This would avoid the need for manually stitching everything together and gives the correct indentations automatically etc...

The final remark: the official file ending of mzQC will be `.mzQC` (I think this is not part of the spec document yet, but we agreed upon that in the last mzQC meeting, if I remember correctly).

cbielow:
I'm happy to discuss more in a TC. maybe in the afternoon?

axelwalter:
Thanks for the feedback, we can discuss it in the afternoon meeting.

axelwalter:
Closing this pull request, revert changes and create new MzQCFile and MzQCCalculator independently of the old qcML stuff.

Issue #5182, opened 2021-02-21
How could I add user-defined RNA digestion enzyme?


mhjc01:
I would like to change the NASE module parameter which is an RNA digestion enzyme such as 'RNase T1'. 

If I want to add other enzymes, for example, like colicinE5(G|U) and mazF(N|ACA), from which file should I modify?

Thank you.

jpfeuffer:
Hi,

This would be:
https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/CHEMISTRY/Enzymes_RNA.xml

mhjc01:
Thank you for replying.

I could add the enzymes to `share/OpenMS/CHENISTRY/Enzymes_RNA.xml`.

And I want to share my implementation for your interest.

``` xml
<?xml version="1.0" encoding="ISO-8859-1"?>
<PARAMETERS>
  <NODE name="Enzymes">
    <NODE name="mazF">
      <ITEM name="Name" value="mazF" type="string" />
      <ITEM name="RegEx" value="(?&lt;=ACA)" type="string" />
      <ITEM name="RegExDescription" value="mazF cuts before ACA." type="string" />
      <ITEM name="CutsAfter" value=".*(?!m)$" type="string" />
      <ITEM name="CutsBefore" value="ACA" type="string" />
      <ITEM name="ThreePrimeGain" value="p" type="string" />
      <ITEM name="FivePrimeGain" value="" type="string" />
    </NODE>
    <NODE name="colicin_E5">
      <ITEM name="Name" value="colicin_E5" type="string" />
      <ITEM name="RegEx" value="(?&lt;=G)U" type="string" />
      <ITEM name="RegExDescription" value="colicin E5 cuts after G followed by U." type="string" />
      <ITEM name="CutsAfter" value="G(?!m)" type="string" />
      <ITEM name="CutsBefore" value=".*" type="string" />
      <ITEM name="ThreePrimeGain" value="p" type="string" />
      <ITEM name="FivePrimeGain" value="" type="string" />
    </NODE>
     ...
  </NODE>
</PARAMETERS>
```

timosachsenberg:
@poshul  would it make sense to add it to our default enzymes?

poshul:
@timosachsenberg  yup, I will add them.

poshul:
I've added the enzymes to our defaults and submitted a PR, thanks @mhjc01!

poshul:
[...]

PR #5200, opened 2021-03-01
[feature] added more RNA digestion enzymes

poshul:
# Description
Added new RNA digestion enzymes mentioned in https://github.com/OpenMS/OpenMS/issues/5182

# Checklist:
- [ X ] Updated or added python bindings for changed or new classes. (Tick if no updates were necessary.)


timosachsenberg (approved):

share/OpenMS/CHEMISTRY/Enzymes_RNA.xml
```diff
@@ -55,6 +55,24 @@
...
+      <ITEM name="RegExDescription" value="mazF cuts before ACA." type="string" />
+      <ITEM name="CutsAfter" value=".*(?!m)$" type="string" />
+      <ITEM name="CutsBefore" value="ACA" type="string" />
+      <ITEM name="ThreePrimeGain" value="p" type="string" />
+      <ITEM name="FivePrimeGain" value="" type="string" />
+    </NODE>
+    <NODE name="colicin_E5">
+      <ITEM name="Name" value="colicin_E5" type="string" />
+      <ITEM name="RegEx" value="(?&lt;=G)U" type="string" />
+      <ITEM name="RegExDescription" value="colicin E5 cuts after G followed by U." type="string" />
```
hendrikweisser, line 70: What about the "U" in "CutsAfter"/"CutsBefore"? (Depending on if it cuts "GU" in the middle or after the "U".)

hendrikweisser (changes requested):
Can you add some test cases to the unit tests, please? The best way to see whether these work as expected is with some positive and negative examples. "RNaseDigestion_test" (https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/RNaseDigestion_test.cpp) should be the right place.

share/OpenMS/CHEMISTRY/Enzymes_RNA.xml
poshul: According to https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1635277/ it is supposed to cut between the two. It also appears to cut between QU as well. I've pushed a commit that updates this

timosachsenberg: please update Description as well

poshul: @hendrikweisser I started adding tests, and quickly found that the multi-base specificity doesn't work with the current code. Specificially [this line](https://github.com/OpenMS/OpenMS/blob/55598cf39c73a2092365fbdb51e63defa9b29c6f/src/openms/source/CHEMISTRY/RNaseDigestion.cpp#L106) only looks at the nucleotide immediately after (which will never be equal to a multi-nucleotide string. Any suggestions on how to handle these sorts of enzymes?

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -103,7 +103,39 @@ namespace OpenMS
...
+          for (size_t j = 0; j < CAR_chars; ++j)
+          {
+            pre_String = pre_String + rna[ i - (CAR_chars - j)]->getCode() + ",";
+          }
+          for (size_t j = 0; j < CBR_chars; ++j)
+          {
+            post_String = post_String + "," + rna[ i + j ]->getCode();
+          }
+          //do the search
+          if (boost::regex_search(pre_String, cuts_after_regex_) && boost::regex_search(post_String, cuts_before_regex_))
```
hendrikweisser, line 132: I don't think this is equivalent to matching regexes (from a comma-separated list) individually against the nucleotides. E.g. you may have a regex in the beginning that reads through the whole string (but was only supposed to match one nucleotide).

My suggestion:
- change 'cuts_before_regex_' and 'cuts_after_regex_' to `vector<boost::regex>` ..._regexes_` (in RNaseDigestion and DigestionEnzymeRNA)
- fill them with regexes after comma-splitting the strings from the XML file
- when matching, iterate through the regexes and corresponding ribonucleotides
- in most cases there will be only one step, but no separate case needed for that

That's actually simpler than I thought it would be.

poshul: I will give this a shot monday and let you know. Thanks!

poshul: Okay, @hendrikweisser  I've implemented your suggestions. Let me know what you think.

share/OpenMS/CHEMISTRY/Enzymes_RNA.xml
```diff
@@ -55,6 +59,24 @@
       <ITEM name="ThreePrimeGain" value="" type="string" />
       <ITEM name="FivePrimeGain" value="p" type="string" />
     </NODE>
+    <NODE name="mazF">
+      <ITEM name="Name" value="mazF" type="string" />
+      <ITEM name="RegEx" value="(?&lt;=A,C,A)" type="string" />
```
hendrikweisser, line 64: I've been reminded that an important property of mazF is its modification-specificity: It doesn't cut before m6A! (Not sure about Am, m1A, m6Am etc.). So you need to adjust the first entry in the regexp list.

hendrikweisser (changes requested):
Code changes look good! Thanks for adding this functionality.

Just see my comment on the mazF cleavage definition.

share/OpenMS/CHEMISTRY/Enzymes_RNA.xml
hendrikweisser: Byron just mentioned that before "A(m5C)A" would *not* be a cutting site either. He didn't know about other A-methylations.

poshul: I've updated the regex as well as the test to handle those two exceptions. I've also included a big warning in the description for the enzyme that we don't consider other methylations.

hendrikweisser (approved):

hendrikweisser:
@poshul, I've approved this now. Thanks for the update. I'm still wondering whether it's more cautious to allow cleavages for methylations other than m6A/m5C or to prevent them. This includes handling of the "ambiguous mod. location" notation that we've introduced ("mA"). But in the absence of more information, either approach could be better/worse.

timosachsenberg:
rebuild jenkins

poshul:
rebuild jenkins

poshul:
@timosachsenberg  any idea what's going on with Jenkins here? It shows a missing file, and some weird binary file path.

timosachsenberg:
'/home/jenkins/ws/openms/PR/bldtst/a9154670/build/bin/ExecutePipelinee' extracted from Executable Path
and '/home/jenkins/ws/openms/PR/bldtst/a9154670/build/bin/ExecutePipelinee/je[NON-UTF-8-BYTE-0x80]y[NON-UTF-8-BYTE-0xBD][NON-UTF-8-BYTE-0xA5][NON-UTF-8-BYTE-0xFE]'
seems to be indeed an issue. Not sure what causes this though

timosachsenberg:
Test is called here:
https://github.com/OpenMS/OpenMS/blob/develop/src/tests/toppas/CMakeLists.txt#L58-L65

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -45,20 +45,36 @@ namespace OpenMS
     EnzymaticDigestion::setEnzyme(enzyme);
 
     const DigestionEnzymeRNA* rnase =
-      dynamic_cast<const DigestionEnzymeRNA*>(enzyme_);
+        dynamic_cast<const DigestionEnzymeRNA*>(enzyme_);
     String five_prime_code = rnase->getFivePrimeGain();
-    if (five_prime_code == "p") five_prime_code = "5'-p";
+    if (five_prime_code == "p")
```
timosachsenberg, line 50: please add curly brackets if you put it in a new line

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -69,24 +85,25 @@ namespace OpenMS
 
 
   vector<pair<Size, Size>> RNaseDigestion::getFragmentPositions_(
-    const NASequence& rna, Size min_length, Size max_length) const
+      const NASequence& rna, Size min_length, Size max_length) const
   {
-    if (min_length == 0) min_length = 1;
+    if (min_length == 0)
```
timosachsenberg, line 90: same here

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -69,24 +85,25 @@ namespace OpenMS
...
+      min_length = 1;
     if ((max_length == 0) || (max_length > rna.size()))
     {
       max_length = rna.size();
     }
 
     vector<pair<Size, Size>> result;
-    if (enzyme_->getName() == NoCleavage) // no cleavage
+    if (enzyme_->getName() == NoCleavage)// no cleavage
```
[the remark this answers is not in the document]
timosachsenberg: also below

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -98,13 +115,32 @@ namespace OpenMS
...
-            boost::regex_search(rna[i]->getCode(), cuts_before_regex_))
+        bool is_match = true;
+        // can't match if we dont have enough bases before or after
+        if (i < cuts_after_regexes_.size() || rna.size() - i < cuts_before_regexes_.size())
+        {
+          is_match = false;
+        }
+        for (auto it = cuts_after_regexes_.begin(); it != cuts_after_regexes_.end() && is_match; ++it)
+        {
+          if (!boost::regex_search(rna[i - cuts_after_regexes_.size() + (it - cuts_after_regexes_.begin())]->getCode(), *it))
```
timosachsenberg, line 131: please add comment what is checked here

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -98,13 +115,32 @@ namespace OpenMS
...
+          {
+            is_match = false;
+          }
+        }
+        for (auto it = cuts_before_regexes_.begin(); it != cuts_before_regexes_.end() && is_match; ++it)
+        {
+          if (!boost::regex_search(rna[i + (it - cuts_before_regexes_.begin())]->getCode(), *it))
```
timosachsenberg, line 138: also comment would be good here

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -118,7 +154,8 @@ namespace OpenMS
         for (Size offset = 0; offset <= missed_cleavages_; ++offset)
         {
           Size end_it = start_it + offset + 1;
-          if (end_it >= fragment_pos.size()) break;
+          if (end_it >= fragment_pos.size())
```
timosachsenberg, line 157: either put in same line or add curly brackets

timosachsenberg: see https://github.com/OpenMS/OpenMS/wiki/Coding-conventions

PR #5209, opened 2021-03-03
[INTERNAL] Remove non standard dependecies in Param class

### LINKED ISSUE #5090: Remove dependencies to non standard code in Param object
- replace usage of Map with map
- replace usage of String with string

from Param.h and Param.cpp

rubengruenberg:
# Description

Removed the dependecies for Map and String in the Param class in order to close #5090 
The class ParamValue was added to replace DataValue in Param, currently no specific tests for this class were added.
In contrast to DataValue the ParamValue class is a bit slimmer as it doesn't need to store unit and unit type.

TODO: discuss if DataValue could inherit from ParamValue to add e.g. units

# Checklist:


src/openms/include/OpenMS/APPLICATIONS/MapAlignerBase.h
```diff
@@ -84,7 +84,7 @@ class TOPPMapAlignerBase :
     Param params;
     params.setValue("type", default_model, "Type of model");
     // TODO: avoid referring to each TransformationModel subclass explicitly
-    StringList model_types = ListUtils::create<String>("linear,b_spline,lowess,interpolated");
+    std::vector<std::string> model_types = ListUtils::create<std::string>("linear,b_spline,lowess,interpolated");
```
timosachsenberg, line 87: maybe better to replace these constructs with:
= {"linear_b", "b_spline", "lowess", "interpolated"}

src/openms/include/OpenMS/APPLICATIONS/TOPPBase.h
```diff
@@ -362,7 +362,7 @@ namespace OpenMS
 
       where "some_key" == key in the examples.
     */
-    const DataValue& getParam_(const String& key) const;
+    const DataValue getParam_(const String& key) const;
```
timosachsenberg, line 365: hmm need to discuss if we can replace DataValue here.

src/openms/include/OpenMS/DATASTRUCTURES/Map.h
```diff
@@ -1,4 +1,5 @@
 // --------------------------------------------------------------------------
+// --------------------------------------------------------------------------
```
timosachsenberg, line 2: remove

src/openms/include/OpenMS/DATASTRUCTURES/ToolDescription.h
```diff
@@ -36,6 +36,7 @@
 
 #include <OpenMS/DATASTRUCTURES/Param.h>
 #include <OpenMS/DATASTRUCTURES/String.h>
+#include <OpenMS/DATASTRUCTURES/ListUtils.h>
```
timosachsenberg, line 39: are these includes of ListUtils.h always needed? I saw them in quite some files

src/openms/include/OpenMS/FILTERING/BASELINE/MorphologicalFilter.h
```diff
@@ -172,10 +172,10 @@ namespace OpenMS
       //structuring element
       defaults_.setValue("struc_elem_length", 3.0, "Length of the structuring element. This should be wider than the expected peak width.");
       defaults_.setValue("struc_elem_unit", "Thomson", "The unit of the 'struct_elem_length'.");
-      defaults_.setValidStrings("struc_elem_unit", ListUtils::create<String>("Thomson,DataPoints"));
+      defaults_.setValidStrings("struc_elem_unit", ListUtils::create<std::string>("Thomson,DataPoints"));
```
timosachsenberg, line 175: does replace by 
, {"Thomson", "DataPoints"})
work? would prefer - also in other files

rubengruenberg: yes that would work, that may also remove the need for the ListUtils include in some files

src/openms/include/OpenMS/DATASTRUCTURES/ToolDescription.h
rubengruenberg: The include often is only needed for the typedef of StringList, it could be removed by adding the typedef to each file or just writing vector<String> instead of StringList in the header files

timosachsenberg:
there seem to be some merge conflicts

timosachsenberg:
TODO:
- pyOpenMS bindings - @rubengruenberg can you take the ones from e.g. DataValue.pxd copy it int the same folder and adapt it to the new classes
- tests - @rubengruenberg can you add tests (most should be just copying the one from DataValue). See the developer documentation which cmake files you need to adapt

src/openms/include/OpenMS/DATASTRUCTURES/ParamValue.h
```diff
@@ -0,0 +1,396 @@
...
+
+private:
+
+    /// Clears the current state of the ParamValue and release every used memory.
+    void clear_() noexcept;
+
+    /// Convert a double to std::string
+    /// with full precision 15 decimal places are given, otherwise 3
+    /// numbers above 10000 or below 0.0001 are given in scientific notation (i.e. 1.0e04)
+    static std::string doubleToString(double value, bool full_precision = true);
```
timosachsenberg, line 392: I guess we can't use https://en.cppreference.com/w/cpp/utility/to_chars or https://en.cppreference.com/w/cpp/io/c/fprintf right?

rubengruenberg: I will look into your suggestions if they can be used instead, since i am not really happy with the current solution. I just wanted to make something that works for now.

timosachsenberg:
rebuild jenkins

timosachsenberg:
TODO: check if we need to adapt any pyOpenMS tutorials

timosachsenberg:
rebuild jenkins

timosachsenberg:
Currently pyOpenMS is failing with:
```
pyopenms/pyopenms_1.cpp: In function 'PyObject* __pyx_pf_8pyopenms_10pyopenms_1_16IntegerDataArray_18getMetaValue(__pyx_obj_8pyopenms_10pyopenms_1_IntegerDataArray*, PyObject*)':
pyopenms/pyopenms_1.cpp:40059:38: error: cannot convert 'OpenMS::DataValue::DataType' to 'OpenMS::ParamValue::DataType' in assignment
     __pyx_t_7 = __pyx_v__r.valueType();
```

timosachsenberg:
Strange that the error only occurs in pyOpenMS. @hroest do we need a wrap-attach here? 

Note that both enums are very similar:

https://github.com/OpenMS/OpenMS/blob/cd8dc3369b4ff5706d2d475b0d934c32d1f605fc/src/pyOpenMS/pxds/DataValue.pxd#L46 

https://github.com/OpenMS/OpenMS/blob/cd8dc3369b4ff5706d2d475b0d934c32d1f605fc/src/pyOpenMS/pxds/ParamValue.pxd#L31

timosachsenberg:
@rubengruenberg I think the problem is that both have an enum DataType which seems to conflict during wrapping.
Can you rename the DataType in ParamType to ValueType (also in the pxd files) and get everything to compile again? Could solve the issue.

timosachsenberg:
thanks. went much further. Now seems to be a missmatch between set of string and set of String
```
pyopenms/pyopenms_5.cpp: In function 'int __pyx_pf_8pyopenms_10pyopenms_5_10ParamEntry_4tags___set__(__pyx_obj_8pyopenms_10pyopenms_5_ParamEntry*, PyObject*)':
pyopenms/pyopenms_5.cpp:221161:0: error: no match for 'operator=' (operand types are 'std::set<std::__cxx11::basic_string<char> >' and 'std::set<OpenMS::String>')
   __pyx_v_self->inst.get()->tags = (*__pyx_v_v0);
 
In file included from /usr/include/c++/7/set:61:0,
                 from pyopenms/pyopenms_5.cpp:453:
/usr/include/c++/7/bits/stl_set.h:288:7: note: candidate: std::set<_Key, _Compare, _Alloc>& std::set<_Key, _Compare, _Alloc>::operator=(const std::set<_Key, _Compare, _Alloc>&) [with _Key = std::__cxx11::basic_string<char>; _Compare = std::less<std::__cxx11::basic_string<char> >; _Alloc = std::allocator<std::__cxx11::basic_string<char> >]
       operator=(const set&) = default;
```

timosachsenberg:
```
pyopenms/pyopenms_5.cpp: In function 'PyObject* __pyx_pf_8pyopenms_10pyopenms_5_10ParamEntry_2isValid(__pyx_obj_8pyopenms_10pyopenms_5_ParamEntry*, PyObject*)':
pyopenms/pyopenms_5.cpp:221978:0: error: cannot bind non-const lvalue reference of type 'std::__cxx11::string& {aka std::__cxx11::basic_string<char>&}' to an rvalue of type 'std::__cxx11::string {aka std::__cxx11::basic_string<char>}'
     __pyx_t_3 = __pyx_v_self->inst.get()->isValid(((std::string)__pyx_t_2));
 
In file included from /home/jenkins/ws/openms/PR/pyOMSDckr/OpenMS/src/openms/include/OpenMS/CONCEPT/UniqueIdGenerator.h:38:0,
                 from /home/jenkins/ws/openms/PR/pyOMSDckr/OpenMS/src/openms/include/OpenMS/CONCEPT/UniqueIdInterface.h:37,
                 from pyopenms/pyopenms_5.cpp:470:
/home/jenkins/ws/openms/PR/pyOMSDckr/OpenMS/src/openms/include/OpenMS/DATASTRUCTURES/Param.h:93:12: note:   initializing argument 1 of 'bool OpenMS::Param::ParamEntry::isValid(std::__cxx11::string&) const'
       bool isValid(std::string& message) const;
            ^~~~~~~
pyopenms/pyopenms_5.cpp: In function 'PyObject* __pyx_pf_8pyopenms_10pyopenms_5_10ParamEntry_8_init_2(__pyx_obj_8pyopenms_10pyopenms_5_ParamEntry*, PyObject*, __pyx_obj_8pyopenms_10pyopenms_5_ParamValue*, PyObject*, PyObject*)':
pyopenms/pyopenms_5.cpp:222608:0: error: no matching function for call to 'OpenMS::Param::ParamEntry::ParamEntry(boost::shared_ptr<OpenMS::String>::element_type&, boost::shared_ptr<OpenMS::ParamValue>::element_type&, std::__cxx11::string, std::vector<OpenMS::String>&)'
     __pyx_t_11 = new OpenMS::Param::ParamEntry((*__pyx_f_8pyopenms_10pyopenms_1_convString(__pyx_v_n).get()), (*__pyx_v_v->inst.get()), ((std::string)__pyx_t_10), (*__pyx_v_v3));
```

jpfeuffer:
I don't think autowrap can handle string references yet. Or there is a bug. It declares the function with input as python "bytes" and then passes an on-the-fly converted std::string to the function. It obviously will be an r-value.

jpfeuffer:
See https://github.com/uweschmitt/autowrap/blob/2dcf9724294f23f87ae686f7d5a5f193a5fab3e2/autowrap/ConversionProvider.py#L1505

timosachsenberg:
Thanks @jpfeuffer as we treat std::string as immutable python strings we probably need to change how the function is called in C++ if we want to wrap it.
So probably we can change the return type to std::pair<bool, std::string> and wrap this as 
from libcpp.pair cimport pair
libcpp_pair[bool, libcpp_string] .
Do you know if this could also be unpacked as a python tuple? let's say: error, message = p.isValid()  ?

jpfeuffer:
Should? work.

timosachsenberg:
nearly there ;) http://cdash.openms.de/viewTest.php?onlyfailed&buildid=137111

timosachsenberg:
probably also need to provide a conversion provider: https://github.com/OpenMS/OpenMS/blob/d9692da0d410c06b6cdc960f608a5c962360d09c/src/pyOpenMS/converters/special_autowrap_conversionproviders.py#L117

timosachsenberg:
TODO after merge: check if readthedocs needs to be updated

doc/code_examples/Tutorial_Param.cpp
```diff
@@ -28,6 +28,7 @@
 // ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 
+#include <OpenMS/CONCEPT/Types.h>
```
jpfeuffer, line 31: Why was this added?

src/openms/include/OpenMS/DATASTRUCTURES/ParamValue.h
```diff
@@ -0,0 +1,396 @@
...
+      Note: The implementation uses typedef double (as opposed to float, double, long double.)
+
+      @exception Exception::ConversionError is thrown if a cast from the the wrong type is requested
+    */
+    operator float() const;
+
+    /**
+      @brief conversion operator to short int
+
+      Note: The implementation uses typedef SignedSize.
```
jpfeuffer, line 179: Is this true? I see ptrdiff_t

src/openms/source/DATASTRUCTURES/ParamValue.cpp
```diff
@@ -0,0 +1,876 @@
...
+    case ParamValue::DOUBLE_VALUE:
+        os << p.data_.dou_;
+    break;
+    case ParamValue::EMPTY_VALUE:
+    break;
+    }
+    return os;
+  }
+
+    std::string ParamValue::doubleToString(double value, bool full_precision)
```
jpfeuffer, line 841: Hmm not sure if this is the right place here.
There is also more efficient code in StringConversions.
Indentation looks off.

src/openms/source/DATASTRUCTURES/ParamValue.cpp
```diff
@@ -0,0 +1,876 @@
...
+    return os;
+  }
+
+    std::string ParamValue::doubleToString(double value, bool full_precision) 
+    {
+        std::ostringstream os;
+        std::string s;
+        if (full_precision) os.precision(15);
+        else os.precision(3);
+        if (std::abs(value) >= 10000.0 ||
```
jpfeuffer, line 847: initialize abs

jpfeuffer (commented):
Only small thing

src/openms/source/ANALYSIS/MAPMATCHING/TransformationDescription.cpp
```diff
@@ -34,6 +34,7 @@
 
 #include <OpenMS/ANALYSIS/MAPMATCHING/TransformationDescription.h>
 #include <OpenMS/CONCEPT/LogStream.h>
+#include <OpenMS/DATASTRUCTURES/ListUtils.h>
```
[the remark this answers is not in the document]
timosachsenberg: used here:
```
  void TransformationDescription::getModelTypes(StringList& result)
  {
    result = ListUtils::create<String>("linear,b_spline,interpolated,lowess");
    // "none" and "identity" don't count
  }
```
before probably included  via DataValue

src/openms/source/CHEMISTRY/SpectrumAnnotator.cpp
```diff
@@ -395,7 +395,7 @@ namespace OpenMS
         //TODO add "FragmentArray"s
 
         Param sap = sa.getParameters();
-        pi.setMetaValue("fragment_match_tolerance", sap.getValue("tolerance"));
+        pi.setMetaValue("fragment_match_tolerance", DataValue(sap.getValue("tolerance")));
```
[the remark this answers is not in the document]
timosachsenberg: Yes, I think we could add that.
The other question is if MetaValue could just use ParamValue without all that unit stuff.

src/openms/source/DATASTRUCTURES/ParamValue.cpp
rubengruenberg: StringConversions is a non standard dependency and  can not be used here.
All the stl functions for string conversions I looked at can not replicate the required conversions (for example they won't remove trailing zeros or switch to scientific notation at a certain value).

timosachsenberg: I think Ruben had to come up with some custom code to get the same string format

jpfeuffer: Ok

cbielow:
why do we we need this huge change in the first place? Who needs this class to just depend on std types?

cbielow:
some changes here are really nice, e.g. using `{"a", "bb", "cc"}` instead of ListUtils::create..., but others really make the code a lot harder to read. e.g. when doing `String(....toString())`, and I dont even understand why this is needed.
In some other place one could certainly get away with overloading some conversions.

timosachsenberg:
We are exploring to move some parts e.g. Param + CTD (later CWL) into a library. We thus removed the dependencies on non STL stuff. The conversions happening now are mainly because meta value is also a bit clunky :P

jpfeuffer:
also wrapping is much harder with this super weird string inheritance. I fully support removal of String whenever possible

jpfeuffer:
also saves memory through removal of unnecessary unit

mr-c:
Many thanks @timosachsenberg @rubengruenberg !

PR #5235, opened 2021-03-15
Add some other redist finding methods and arbitrarily use first redist in case multiple are found

jpfeuffer:
@oliveralka


cmake/package_nsis.cmake
```diff
@@ -50,29 +50,44 @@ set(VC_REDIST_EXE "vcredist_${ARCH}.exe")
...
-	  #  message(FATAL_ERROR "Executing vswhere to find vsredist executable for win packaging failed. Either specify VC_REDIST_PATH or make sure vswhere works.")
-	  #endif()
-	  
-	  ## We have to glob recurse in the parent folder because there is a version number in the end.
-	  ## Unfortunately in my case the default version (latest) does not include the redist?!
-	  ## TODO Not sure if this environment variable always exists. In the VS command line it should! Fallback vswhere or VCINSTALLDIR/Redist/MSVC?
-	  get_filename_component(VC_ROOT_PATH "$ENV{VCToolsRedistDir}.." ABSOLUTE)
+	if("${OPENMS_MSVC_VERSION_STRING}" GREATER "17")
+	  ## according to https://docs.microsoft.com/de-de/cpp/windows/redistributing-visual-cpp-files?view=msvc-160
+	  get_filename_component(VC_ROOT_PATH "$ENV{VCINSTALLDIR}Redist/MSVC/v142" ABSOLUTE)
```
oliveralka, line 55: Should we check that it is really in the env? For abibuilder8 it was not automatically in the System Env.

oliveralka:
Fixes:
```warning: !warning: Deprecated: The MUI_LANGDLL_REGISTRY_ROOT, MUI_LANGDLL_REGISTRY_KEY and MUI_LANGDLL_REGISTRY_VALUENAME defines were replaced by UMUI_LANGUAGE_REGISTRY_VALUENAME, UMUI_LANGUAGE_REGISTRY_ROOT, UMUI_LANGUAGE_REGISTRY_KEY or you can use also the UMUI_PARAMS_REGISTRY_ROOT & UMUI_PARAMS_REGISTRY_KEY globals parameters. (macro:MUI_INTERFACE:15)```

`!define un.${StrFuncName} '${Un${StrFuncName}}'`
https://stackoverflow.com/questions/62081765/how-to-debug-nsis-script-3-05-2-gives-error-error-unterminated-string-parsing

cmake/package_nsis.cmake
jpfeuffer: Can you check on the VS command line? Or was it?

jpfeuffer: But yes, checking is better in any case

jpfeuffer: Could you do that for the next if, too?

oliveralka:
@jpfeuffer Let's hope that it works like this.

oliveralka (approved):

PR #5239, opened 2021-03-15
[FEATURE] Add OpenMSDatabasesInfo Util to list supported Enzymes and Mods in tsv

jpfeuffer:
# Description

See title. I think testing would be superfluous and tedious since we would need to adapt the test files every time the DB is updated.

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -209,7 +209,9 @@ public:
     /// Collects all modifications that can be used for identification searches
     void getAllSearchModifications(std::vector<String>& modifications) const;
 
-protected:
+    void writeCSV(const String& filename);
+
+  protected:
 
     /// Stores whether ModificationsDB was instantiated before
     static bool is_instantiated_;
@@ -272,6 +274,5 @@ private:
 
     /// Adds modifications from a given file in Unimod XML format
     void readFromUnimodXMLFile(const String& filename);
-    
   };
 }
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -209,7 +209,7 @@ public:
     /// Collects all modifications that can be used for identification searches
     void getAllSearchModifications(std::vector<String>& modifications) const;
 
-    /// Writes the contents to a TSV file
+    /// Writes tab separated entries: FullId,FullName,Origin,AA,TerminusSpecificity,DiffMonoMass (including header) to TSV file
     void writeTSV(const String& filename);
 
   protected:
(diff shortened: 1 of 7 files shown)


Leon-Bichmann (approved):
LGTM, as far as I can see.

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -209,7 +209,9 @@ namespace OpenMS
     /// Collects all modifications that can be used for identification searches
     void getAllSearchModifications(std::vector<String>& modifications) const;
 
-protected:
+    void writeCSV(const String& filename);
```
timosachsenberg, line 212: TSV oder CSV? Please add documentation here what fields are tab separated written.

src/openms/source/CHEMISTRY/ProteaseDB.cpp
```diff
@@ -105,4 +105,13 @@ namespace OpenMS
     }
   }
 
+  void ProteaseDB::writeCSV(String const& filename)
+  {
+    std::ofstream ofs(filename, std::ofstream::out);
+    ofs << "OpenMS_AllowedEnzymes" << "\n";
```
timosachsenberg, line 111: Just "Enzyme"?

src/utils/OpenMSDatabasesInfo.cpp
```diff
@@ -0,0 +1,103 @@
...
+  {
+  }
+
+protected:
+  // this function will be used to register the tool parameters
+  // it gets automatically called on tool execution
+  void registerOptionsAndFlags_() override
+  {
+    // Output CSV file
+    registerOutputFile_("enzymes_out", "<out>", "", "Enzymes as CSV", true, false);
```
timosachsenberg, line 78: says CSV type is TSV

timosachsenberg (approved):
Looks good. Only minor comments.
Some brainstorming: could this tool be also be used to conveniently export a json file with all information? Would writing it on std::cout also be an option? Could open a few other possibilities but for what you want to do now it is fine.

src/openms/source/CHEMISTRY/ProteaseDB.cpp
jpfeuffer: Want to make a clear distinction between header and options.

jpfeuffer:
Stdout is an option later but output files *have to be present* for some workflow engines.
Sure, if someone needs json, this can be done in the same tool.

src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
```diff
@@ -209,7 +209,10 @@ namespace OpenMS
     /// Collects all modifications that can be used for identification searches
     void getAllSearchModifications(std::vector<String>& modifications) const;
 
-protected:
+    /// Writes the contents to a TSV file
```
timosachsenberg, line 212:
```suggestion
    /// Writes tab separated entrie: FullId,FullName,Origin,AA,TerminusSpecificity,DiffMonoMass (including header) to TSV file
```

Issue #5231, opened 2021-03-12
TOPPViewer won't run on MacOs 11.0.1 ('Big sur')


arbi56:
After changing settings to allow the program to open (MacOS complains it can't check the SW) it starts but immediately exits. SwathWizard does exactly the same. SW was downloaded as a disk image from your web site.

jpfeuffer:
Hi!

Unfortunately, we do not yet have a Big Sur machine at hand and macOS gets stricter and stricter for distributed software. Can you try to open ToppView from the Terminal via the two different options:
- `open /Applications/TOPPView.app`
- `/Applications/TOPPView.app/Contents/MacOS/TOPPView`

to see if any of them give more output for debugging.

arbi56:
Hi Julianus - thanks for the response!

On my Mac the files are actually stored in Applications in another folder: OpenMS-2.6.0…in a terminal window I went to: /OpenMS-2.6.0/TOPPView.app/Contents/MacOS and typed ‘open TOPPView' - it opened! - I also navigated to the same place using the Finder, double-clicked and it also opened. In both cases a terminal window opened with the following:


I checked again and still can’t open the program by selecting TOPPView in the OPenMS-2.6.0 folder…

So, as far as I know it does work…but may not do what I want? I had the impression I could open simple text spectra or mgf files (I’m not in the proteomics business and find mzML et al overkill and too large) but it cannot open such files complaining that there are no survey scans (of course not, they are spectra!). Am I missing something?

I don’t know if this helps you - seems like the program is OK but the OS can’t find it properly? - let me know if I can do more to help.

Cheers

Ron

Ron Bonner, ***@***.***
Skype: bonnerrf, Mobile: +1 416 892 1505




.

jpfeuffer:
Hi Ron,

unfortunately, you are right, although we can read MGF it was strangely never added to the supported formats of TOPPView. It probably is a very minor addition that we are investigating now. We can let you know here, when you can download and try a fresh nightly build of the tool.

Best
Julianus

arbi56:
OK…thanks.

I was hoping I could say that TOPPView works with text/mgf/massbank file - seems strange that this isn’t such a tool! Please let me know if you do activate it

Cheers

R

jpfeuffer:
Hi,

viewing of MGF files should now be possible with the latest DMG on https://abibuilder.informatik.uni-tuebingen.de/archive/openms/OpenMSInstaller/nightly/

Feel free to give it a try.

[...]

PR #5240, opened 2021-03-15
add additional peakfiletypes in TOPPView

timosachsenberg:
# Description

Additional filetypes in TOPPView.
Addresses #5231 partly

# Checklist:

### DIFF
--- a/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
+++ b/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
@@ -134,7 +134,7 @@ namespace OpenMS
   /// supported types which can be opened with File-->Open
   const FileTypes::FileTypeList supported_types({ FileTypes::MZML, FileTypes::MZXML, FileTypes::MZDATA, FileTypes::SQMASS,
                                                   FileTypes::FEATUREXML, FileTypes::CONSENSUSXML, FileTypes::IDXML,
-                                                  FileTypes::DTA, FileTypes::DTA2D,
+                                                  FileTypes::DTA, FileTypes::DTA2D, FileTypes::MGF, FileTypes::MS2
                                                   FileTypes::BZ2, FileTypes::GZ });
 
   TOPPViewBase::TOPPViewBase(QWidget* parent) :
@@ -711,6 +711,16 @@ namespace OpenMS
         {
           data_type = LayerData::DT_PEAK;
         }
+      }
+      else if (file_type == FileTypes::MGF || file_type == FileTypes::MZXML ||
+        file_type ==  FileTypes::DTA || file_type ==  FileTypes::DTA2D ||
+        file_type ==  FileTypes::MZDATA || file_type ==  FileTypes::MS2
+        )
+      {
+        // Load all data into memory
+        fh.loadExperiment(abs_filename, *peak_map_sptr, file_type, ProgressLogger::GUI);
+        OPENMS_LOG_INFO << "INFO: done loading all " << std::endl;
+        data_type = LayerData::DT_PEAK;
       }
     }
     catch (Exception::BaseException& e)
--- a/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
+++ b/src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
@@ -134,7 +134,7 @@ namespace OpenMS
   /// supported types which can be opened with File-->Open
   const FileTypes::FileTypeList supported_types({ FileTypes::MZML, FileTypes::MZXML, FileTypes::MZDATA, FileTypes::SQMASS,
                                                   FileTypes::FEATUREXML, FileTypes::CONSENSUSXML, FileTypes::IDXML,
-                                                  FileTypes::DTA, FileTypes::DTA2D, FileTypes::MGF, FileTypes::MS2
+                                                  FileTypes::DTA, FileTypes::DTA2D, FileTypes::MGF, FileTypes::MS2,
                                                   FileTypes::BZ2, FileTypes::GZ });
 
   TOPPViewBase::TOPPViewBase(QWidget* parent) :
@@ -697,7 +697,7 @@ namespace OpenMS
           }
         }
 
-        // Load all data into memory
+        // Load all data into memory if e.g. no mzML file
         if (!parsing_success)
         {
           fh.loadExperiment(abs_filename, *peak_map_sptr, file_type, ProgressLogger::GUI);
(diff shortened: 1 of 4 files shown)


src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -134,7 +134,7 @@ namespace OpenMS
   /// supported types which can be opened with File-->Open
   const FileTypes::FileTypeList supported_types({ FileTypes::MZML, FileTypes::MZXML, FileTypes::MZDATA, FileTypes::SQMASS,
                                                   FileTypes::FEATUREXML, FileTypes::CONSENSUSXML, FileTypes::IDXML,
-                                                  FileTypes::DTA, FileTypes::DTA2D,
+                                                  FileTypes::DTA, FileTypes::DTA2D, FileTypes::MGF, FileTypes::MS2,
```
jpfeuffer, line 137: Cool! Can we add MSP for the NIST databases?
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/FileHandler.cpp#L369

Or was something speaking against?

timosachsenberg: It can not yet be read via FileHandler. I will take a look.

jpfeuffer:
Are you testing it on your machine? Otherwise, let me know when it is ready to test.

timosachsenberg:
Thread 1 "TOPPView" received signal SIGSEGV, Segmentation fault.
0x00007ffff6583ac0 in OpenMS::StringUtils::toInt(OpenMS::String const&) () from /home/sachsenb/OMS/OpenMS/openms-build/lib/libOpenMS.so
(gdb) bt
#0  0x00007ffff6583ac0 in OpenMS::StringUtils::toInt(OpenMS::String const&) () from /home/sachsenb/OMS/OpenMS/openms-build/lib/libOpenMS.so
#1  0x00007ffff69d7cb5 in OpenMS::MSPFile::load(OpenMS::String const&, std::vector<OpenMS::PeptideIdentification, std::allocator<OpenMS::PeptideIdentification> >&, OpenMS::MSExperiment&) () from /home/sachsenb/OMS/OpenMS/openms-build/lib/libOpenMS.so
#2  0x00007ffff795f6db in OpenMS::TOPPViewBase::addDataFile(OpenMS::String const&, bool, bool, OpenMS::String, unsigned int, unsigned long) ()
   from /home/sachsenb/OMS/OpenMS/openms-build/lib/libOpenMS_GUI.so
#3  0x00007ffff79632bb in OpenMS::TOPPViewBase::openFilesByDialog(OpenMS::String const&) () from /home/sachsenb/OMS/OpenMS/openms-build/lib/libOpenMS_GUI.so
#4  0x00007ffff7addd6c in QtPrivate::QSlotObject<void (OpenMS::TOPPViewBase::*)(OpenMS::String const&), QtPrivate::List<bool>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) () from /home/sachsenb/OMS/OpenMS/openms-build/lib/libOpenMS_GUI.so

timosachsenberg:
Works and supports files that can be read by MSPGenericFile. Still not ideal that peptide centric MSP files fail to load.

image

jpfeuffer:
Very nice. What is the issue with peptide-centric? Probably some ambiguity in the format, right? Some splitting of strings is empty or you have a string where an int is expected, right?

timosachsenberg:
yes basically a bit different strings and how we store the annotations (pep_id vs in spectrum)

timosachsenberg:
maybe something we could also discuss with @bertrandboudaud

bertrandboudaud:
Hi, i can have a look at that. do you have any example file that fails to load?

timosachsenberg:
There is e.g. https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/data/MSPFile_test.msp
that I think did not produce reasonable results.
Two things I found we might also look into:
- storing the m/z in the precursor structure of the spectrum
- finding a unified way to store proteomics and metabolomics peak annotations -> if we manage we could add code to visualize them in TOPPView

bertrandboudaud:
In this file MSPFile_test.msp there are 2 issues the MSPGenericFile::load does not handle well

1. "Num peaks" is written instead of "Num Peaks"

2. The way the spectra is organized.
A the moment, MSPGenericFile::load support only 2 kinds of organisation:

```
Num Peaks: 11
35 310; 36 1230; 37 27; 38 303; 47 5240;
66 203; 67 68; 68 77; 82 63; 83 240;
136 350;
```
or:
```
Num Peaks: 11
35:310 36:1230 37:27 38:303 47:5240
66:203 67:68 68:77 82:63 83:240
136:350
```

The test file show this:
```
Num peaks: 14
608.7	974	"y6/0.39 2/2 2.4"
730.4	407	"? 2/2 0.6"
764.4	1788	"b7/0.01 2/2 1.1"
832.0	814	"b8-46/0.53 2/2 1.3"
```

That is 3 fields, tab-separated, and new line for each peak. this organization is not supported at the moment. if I reorganize the file as described above, the file loads sucessfully. we then lose information about the labeling "y6/0.39 2/2 2.4" ...

The documentation about MSP specification I have found is quite old, but it seems that "Num Peaks" with upper case P is what is specified - I imagine we can be less case sensitive - and about the peak data it's not very specific. we can be one line = one peak, tab-separated, or the way it is already supported in MSPGenericFile for the moment which also already valid regarding the specification. However I don't see any information about the possibility of labeling the peak - which is currently not supported at all, it messes the parser as well.

jpfeuffer:
Nice insights. I agree that the specification/standardization is close to non-existent. However, I think it would be really cool if annotations would be supported in any kind of format. I assume when this file was generated, someone had some format in mind at least.

bertrandboudaud:
Looks like this test file was generated from the MSPFile class, which was specialized for proteomics. MSPGenericFile was added after to support metabolomics. It's a bit sad that one cannot read the file created by the other. maybe some factorization can be done there.

timosachsenberg:
I agree. Maybe we could refactor both classes into one MSPFile class.
e.g., you could make (if timer permits) parsing of 
- Num peaks case insensitive
- include something like this into your parser if it makes sense:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/MSPFile.cpp#L159-L169

And we could then give it a go and try to add the proteomics peak annotation format and storing of modifications.

Should already get us close to something that is usable for both proteomics and metabolomics.

jpfeuffer:
I agree. But to be clear, how was this annotation format designed? Is it documented somewhere? Is it use somewhere else than in OpenMS?

timosachsenberg:
I think some parts might be used in the spectrallibrarysearch.
So one way we could make this more universal is to fill MSSpectrum StringDataArray with the peak annotations.
Would work for both peptide and metabolite spectra.
But if we go for this we should discuss what the best way to store them would be.

PR #5241, opened 2021-03-16
Feature/mzqc export

axelwalter:
# add mzQC export to the QCCalculator

- QCCalculator stores data in mzQC or qcML format depending on output file extension
- using nlohmann/json thirdparty library to create mzQC files
- updated nlohmann/json to newest version
- all algorithms for data collection and storage are now in QcMLFile and MzQCFile
- QCCalculator is very compact
- until now only few and basic quality parameters for mzQC files, more will be added


src/openms/include/OpenMS/FORMAT/QcMLFile.h
```diff
@@ -166,8 +177,13 @@ namespace OpenMS
     void existsRunQualityParameter(const String filename, const String qpname, std::vector<String>& ids) const;
     ///Returns the ids of the parameter name given if found in given set, empty else
     void existsSetQualityParameter(const String filename, const String qpname, std::vector<String>& ids) const;
+    ///Calculation and collection of QC data
+    void collectQCData(const String& inputfile_id, const String& inputfile_feature,
+                       const String& inputfile_consensus, const String& inputfile_raw,
+                       const bool remove_duplicate_features, ControlledVocabulary cv, MSExperiment exp);
```
jpfeuffer, line 183: Do you need to copy the whole experiment? Looks a bit costly but you might have reasons.

src/utils/QCCalculator.cpp
```diff
@@ -131,1062 +118,63 @@ class TOPPQCCalculator :
...
-            row.push_back(exp[i].getAcquisitionInfo()[j].getMetaValue("MS:1000927"));
-            at.tableRows.push_back(row);
-          }
-        }
-      }
-    }
-    if (!at.tableRows.empty())
+      MzQCFile mzqcfile;
+      mzqcfile.store(inputfile_raw, outputfile_name, exp, cv, contactName, contactAddress, description, label);
+    } else
```
timosachsenberg, line 173: if you restrict the output filename in the register parameter section you don't need the else

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+// TODO fix all the shadowed "const_iterator qpsit"
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshadow"
+
+namespace OpenMS
+{ 
+  void MzQCFile::store(const String & inputFileName,
+                       const String & outputFileName,
+                       MSExperiment exp,
```
jpfeuffer, line 51: I think it is better to warn the developer that ranges are updated in the function description and to pass it by reference. But that is my opinion.

src/openms/include/OpenMS/FORMAT/MzQCFile.h
```diff
@@ -0,0 +1,80 @@
...
+  class OPENMS_DLLAPI MzQCFile
+  {
+  public:
+    // Default constructor
+    MzQCFile() = default;
+
+    // Store the QCFile
+    void store(const String & inputFileName,
+               const String & outputFileName,
+               MSExperiment exp,
```
jpfeuffer, line 70: see my comment in the cpp. I would personally prefer a reference here with a note in the method doc. Or require updated ranges from the beginning.

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+                       ControlledVocabulary cv,
```
jpfeuffer, line 52: From the usage, this looks like it could be passed const-ref but not sure.

src/openms/include/OpenMS/FORMAT/QcMLFile.h
```diff
@@ -37,6 +37,17 @@
...
+#include <OpenMS/KERNEL/MSExperiment.h>
+#include <OpenMS/KERNEL/FeatureMap.h>
+#include <QFileInfo>
+#include <OpenMS/FORMAT/IdXMLFile.h>
+#include <OpenMS/FORMAT/FeatureXMLFile.h>
+#include <OpenMS/FORMAT/FeatureXMLFile.h>
+#include <OpenMS/FORMAT/ConsensusXMLFile.h>
+#include <OpenMS/METADATA/ProteinIdentification.h>
+#include <OpenMS/MATH/MISC/MathFunctions.h>
+#include <OpenMS/MATH/STATISTICS/StatisticFunctions.h>
```
jpfeuffer, line 50: I think (all of?) those belong to the cpp file.

jpfeuffer (changes requested):
Minor comments.

src/openms/source/FORMAT/MzQCFile.cpp
timosachsenberg: Try to write it to accept a const MSExperiment& would be even better. Then we would need a public calculateRange in MSExperiment. What do you think @jpfeuffer

jpfeuffer: Yes since you actually do not need to update but only calculate this should work.

src/openms/include/OpenMS/FORMAT/MzQCFile.h
```diff
@@ -0,0 +1,80 @@
...
+
+      @ingroup FileIO
+  */
+  class OPENMS_DLLAPI MzQCFile
+  {
+  public:
+    // Default constructor
+    MzQCFile() = default;
+
+    // Store the QCFile
```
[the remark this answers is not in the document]
axelwalter: updated the new @param tags in the QCCalculator doxygen, do MzQCFile and QcMLFile need their own doxygen?

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+    if (!description.empty()) 
+    {
+      out["MzQC"]["description"] = description;
+    }
+
+    out["MzQC"]["runQualities"]["metadata"]["label"] = label;
+    out["MzQC"]["runQualities"]["metadata"]["inputFiles"] = {
+        {
+          {"location", String(QFileInfo(QString::fromStdString(inputFileName)).absoluteFilePath())},
```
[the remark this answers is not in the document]
axelwalter: done

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+          {"name", String(QFileInfo(QString::fromStdString(inputFileName)).baseName())},
```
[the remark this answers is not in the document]
axelwalter: done

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+    for (const auto& spectrum : exp)
+      {
+        const Size level = spectrum.getMSLevel();
+        ++counts[level];  // count MS level
+      }
+
+    exp.updateRanges();
+
+    // number of MS1 spectra
+    json qm;
```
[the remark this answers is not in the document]
axelwalter: it is now in an internal lambda function, which works well

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+    qm["accession"] = "QC:4000059";
+    try
+    {
+      qm["name"] = cv.getTerm(qm["accession"]).name;
```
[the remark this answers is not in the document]
axelwalter: yes there is a check, done

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    //open stream
+    ofstream os(outputFileName.c_str());
+    if (!os)
+    {
+      throw Exception::UnableToCreateFile(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, outputFileName);
+    }
+
+    os.precision(writtenDigits<double>(0.0));
+    os << out.dump(2);
```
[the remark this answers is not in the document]
axelwalter: it writes out the json in proper format and takes the indentation level as parameter, added a comment

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+}
+
+#pragma clang diagnostic pop
```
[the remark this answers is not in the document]
axelwalter: done

src/utils/QCCalculator.cpp
axelwalter: true, that was not necessary

src/openms/include/OpenMS/FORMAT/QcMLFile.h
axelwalter: The copy was not necessary, I changed it for for the mzQC file to const reference.
For the QcML file there seems to be still some issues with const reference, I will look into that.

src/openms/include/OpenMS/FORMAT/MzQCFile.h
axelwalter: updated the ranges before passing the exp as const reference worked

src/openms/source/FORMAT/MzQCFile.cpp
axelwalter: done

src/openms/include/OpenMS/FORMAT/QcMLFile.h
axelwalter: done

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Axel Walter $
+// $Authors: Axel Walter $
+// --------------------------------------------------------------------------
+
+#include <OpenMS/FORMAT/MzQCFile.h>
+
+#include <boost/range/adaptor/map.hpp>
```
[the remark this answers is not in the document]
axelwalter: I think they are still included here but not used, removing them did not result in errors.

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+    out["MzQC"]["runQualities"]["metadata"]["analysisSoftware"] = { // todo
+        {
+          {"accession", "MS:1009001" }, // create new qc-cv for QCCalculator: MS:1009001 quality control metrics generating software
+          {"name", "QCCalculator"},
+          {"version", "2.6.0"},
+          {"uri", "https://www.openms.de"}
+        }
+    };
+
+    out["MzQC"]["runQualities"]["qualityMetrics"] = {};
```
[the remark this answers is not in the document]
axelwalter: because it is an ordered json that contains the input order of elements. The idea was to build the scaffold of the json first and then add all the quality metrics afterwards.

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+              {"value", String(exp.getInstrument().getName())}
+            }
+          }
+    };
+
+    out["MzQC"]["runQualities"]["metadata"]["analysisSoftware"] = { // todo
+        {
+          {"accession", "MS:1009001" }, // create new qc-cv for QCCalculator: MS:1009001 quality control metrics generating software
+          {"name", "QCCalculator"},
+          {"version", "2.6.0"},
```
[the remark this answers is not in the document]
axelwalter: done

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,234 @@
...
+            {"name", "mzML format"}
+          }
+          }
+        },
+          "fileProperties",
+          {
+            {
+              {"accession", "MS:1000747"},
+              {"name", "completion time"},
+              {"value", String(exp.getDateTime().getDate() + " " + exp.getDateTime().getTime())}
```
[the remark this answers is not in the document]
axelwalter: the example mzQC files use a " " to separate date and time, however, the specification pdf uses "T" for separation. Corrected it to match the specification pdf.

src/openms/include/OpenMS/FORMAT/QcMLFile.h
axelwalter: now the experiment gets passed by reference in both mzQC and qcML files

src/openms/include/OpenMS/FORMAT/QcMLFile.h
```diff
@@ -166,8 +177,13 @@ namespace OpenMS
     void existsRunQualityParameter(const String filename, const String qpname, std::vector<String>& ids) const;
     ///Returns the ids of the parameter name given if found in given set, empty else
     void existsSetQualityParameter(const String filename, const String qpname, std::vector<String>& ids) const;
+    ///Calculation and collection of QC data
```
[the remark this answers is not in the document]
axelwalter: done

src/openms/include/OpenMS/FORMAT/MzQCFile.h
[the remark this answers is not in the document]
axelwalter: done

axelwalter:
in QcMLFile.cpp line 1063 calculateSNmedian() I had to pass the MSSpectrum by copy instead of reference. Perhaps because it gets sorted by intensity in the function? Any suggestions how to do this more efficiently?

jpfeuffer (approved):
Looks good now. Regarding sorting I think there is no easy way. You could only copy the values you need instead of the spectrum but not sure if it is worth the trouble.
But are there tests already? Also for developers  to see the expected outcome in JSON. Or is it not far enough for that already?

timosachsenberg:
I think tests for the algorithms are missing of course.
We still should add tests for the QCCalculator tool.
@axelwalter can you take a look at https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/CMakeLists.txt and try to add some test files?
Here you can run a tool and compare its output to a file. Already great way to ensure tools keep working / ouput doesn't change unexpected.

axelwalter:
Thanks for the comments. I will add some some tests and leave the spectrum as a copy for now.

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include <nlohmann/json.hpp>
+
+#include <vector>
+#include <map>
+#include <iostream>
+#include <fstream>
+
+using namespace std;
+
+// TODO fix all the shadowed "const_iterator qpsit"
```
timosachsenberg, line 52: is this still an issue? otherwise we could remove the clang pragmas

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,199 @@
...
+    out["MzQC"]["controlledVocabularies"] = {
+        {
+          {"name", "Proteomics Standards Initiative Quality Control Ontology"},
+          {"uri", "https://github.com/HUPO-PSI/mzQC/blob/master/cv/qc-cv.obo"},
+          {"version", "0.1.2"},
+        },
+        {
+          {"name", "Proteomics Standards Initiative Mass Spectrometry Ontology"},
+          {"uri", "https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo"},
+          {"version", "4.1.49"}
```
timosachsenberg, line 141: We could pull out all version numbers and put them in the header file as:
e.g,
constexpr char psi_ms_obo_version[] = "4.1.49";
so we have all version numbers at one place.
Not sure if worth it

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,199 @@
...
+    for (const auto& spectrum : exp)
+      {
+        const Size level = spectrum.getMSLevel();
+        ++counts[level];  // count MS level
+      }
+
+    // create qualityMetric in json format
+    auto addValue = [](const String& accession, const String& name, const auto& value) -> json
+      {
+        ControlledVocabulary cv;
```
timosachsenberg, line 157: only load the ControlledVocabulary once!  maybe directly in line 144
you can capture it in the lambda with [&cv]

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,199 @@
...
+    out["MzQC"]["runQualities"]["qualityMetrics"] += addValue("QC:4000138", "MZ acquisition range", tuple<UInt,UInt>{exp.getMinMZ(), exp.getMaxMZ()});
+    
+    //open stream
+    ofstream os(outputFileName.c_str());
+    if (!os)
+    {
+      throw Exception::UnableToCreateFile(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, outputFileName);
+    }
+
+    os.precision(writtenDigits<double>(0.0));
```
timosachsenberg, line 192: is this correct?

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -35,7 +35,18 @@
...
+#include <OpenMS/FORMAT/FeatureXMLFile.h>
+#include <OpenMS/FORMAT/FeatureXMLFile.h>
+#include <OpenMS/FORMAT/ConsensusXMLFile.h>
+#include <OpenMS/METADATA/ProteinIdentification.h>
+#include <OpenMS/MATH/MISC/MathFunctions.h>
+#include <OpenMS/MATH/STATISTICS/StatisticFunctions.h>
+
 #include <fstream>
+#include <map>
 #include <boost/range/adaptor/map.hpp>
```
timosachsenberg, line 50: what about these boost includes? can we get rid of them?

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -305,7 +316,7 @@ namespace OpenMS
 
   }
 
-  void QcMLFile::addRunQualityParameter(String run_id, QualityParameter qp)
+  void QcMLFile::addRunQualityParameter (String run_id, QualityParameter qp)
```
timosachsenberg, line 319: remove extra space

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -1038,8 +1049,1023 @@ namespace OpenMS
     }
   }
 
-  void QcMLFile::store(const String& filename) const
+  double getMassDifference(double theo_mz, double exp_mz, bool use_ppm)
```
timosachsenberg, line 1052: we have some methods in place. I think somewhere in the Math namespace.
Can you check and get rid of the duplication?

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -1038,8 +1049,1023 @@ namespace OpenMS
...
+    double error(exp_mz - theo_mz);
+    if (use_ppm)
+    {
+      error = error / (theo_mz * (double)1e-6);
+      //~ error = (1-exp_mz/theo_mz) * (double)1e6;
+    }
+    return error;
+  }
+
+  float calculateSNmedian(MSSpectrum spec, bool norm = true)
```
timosachsenberg, line 1063: I would use const MSSpectrum& spec here and copy the intensities into a vector<double> and use our median method in math or statistics namespace

timosachsenberg: actually I wonder if we don't have SNmedian estimation in OpenMS? can you check. I think we might have several implementations

src/openms/source/FORMAT/MzQCFile.cpp
axelwalter: copied that from the qcML code. perhaps this is not even necessary since the conversion to a string is handled by the json library?

src/openms/source/FORMAT/MzQCFile.cpp
axelwalter: nice!

src/openms/source/FORMAT/MzQCFile.cpp
axelwalter: personally i would leave it as it is since we have to edit everything here anyways when updating the cv

src/openms/source/FORMAT/QcMLFile.cpp
axelwalter: done

src/openms/source/FORMAT/QcMLFile.cpp
axelwalter: since this function was used only once in the code i removed it and did this simple calculation directly

src/openms/source/FORMAT/QcMLFile.cpp
axelwalter: I could not find an already existing SNmedian function, however, I used the Math::median() function here

src/openms/source/FORMAT/MzQCFile.cpp
axelwalter: removing them did not result in any errors

src/tests/topp/QCCalculator_2_output.mzQC
```diff
@@ -0,0 +1,92 @@
+{
+  "MzQC": {
```
timosachsenberg, line 2: can you check or validate somehow against the schema?
e.g., MzQC should be mzQC see https://github.com/HUPO-PSI/mzQC/blob/master/schema/mzqc_schema.json

src/utils/QCCalculator.cpp
```diff
@@ -131,1062 +121,56 @@ class TOPPQCCalculator :
...
     {
-      const ControlledVocabulary::CVTerm& term = cv.getTerm(at.cvAcc);
-      at.name = term.name; ///< Name
-    }
-    catch (...)
+      QcMLFile qcmlfile;
+      qcmlfile.collectQCData(inputfile_id, inputfile_feature,
+                    inputfile_consensus, inputfile_raw, remove_duplicate_features, exp);
+      qcmlfile.store(outputfile_name);
+    } else if (FileHandler::hasValidExtension (outputfile_name, FileTypes::MZQC))
```
timosachsenberg, line 168:
```suggestion
    } 
    else if (FileHandler::hasValidExtension (outputfile_name, FileTypes::MZQC))
```

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -1038,8 +1043,1007 @@ namespace OpenMS
     }
   }
 
-  void QcMLFile::store(const String& filename) const
+  float calculateSNmedian(const MSSpectrum& spec, bool norm = true)
```
timosachsenberg, line 1046: can you check if SignalToNoiseEstimatorMedian
would be a good replacement?
FILTERING/NOISEESTIMATION/SignalToNoiseEstimatorMedian.h

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -1038,8 +1043,1007 @@ namespace OpenMS
...
+        catch (...)
+        {
+          qp.name = "number of features"; ///< Name
+        }
+        addRunQualityParameter(base_name, qp);      
+      }
+
+      if (inputfile_feature != "" && !remove_duplicate_features)
+      {
+
```
timosachsenberg, line 1875:
```suggestion
```

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -1038,8 +1043,1007 @@ namespace OpenMS
...
+            for (UInt w = 0; w < var_mods.size(); ++w)
+            {
+              row.push_back(pep_mods[w]);
+            }
+            at.tableRows.push_back(row);
+          }
+        }
+        addRunAttachment(base_name, at);
+        
+
```
timosachsenberg, line 1773:
```suggestion
```

src/openms/source/FORMAT/QcMLFile.cpp
timosachsenberg: seems to be too far of. probably better to move this method into a QCBase derived metric at some point

src/openms/source/FORMAT/QcMLFile.cpp
```diff
@@ -1076,9 +1076,8 @@ namespace OpenMS
         sign_int += pt->getIntensity();
       }
     }
-    float sn_by_max2median_norm = (sign_int / sign_cnt) / (nois_int / nois_cnt);
-
-    return sn_by_max2median_norm;
+    if (sign_cnt == 0 | nois_cnt == 0 | nois_int <= 0) return 0;
```
timosachsenberg, line 1079:
```suggestion
    if (sign_cnt == 0 || nois_cnt == 0 || nois_int <= 0) return 0;
```

src/tests/topp/QCCalculator_2_output.mzQC
```diff
@@ -0,0 +1,123 @@
+{
+  "mzQC": {
```
cbielow, line 2: can you check that this file validates successfully using https://github.com/bigbio/mzqc-pylib/tree/master

with something like:
```
from mzqc import MZQCFile as qc
syn_check = qc.SyntacticCheck()
syn_check.validate("{ \"mzQC\": " + qc.JsonSerialisable.ToJson(mzqc) + " }")
```

cbielow (commented):
some minor things

src/openms/include/OpenMS/FORMAT/QcMLFile.h
```diff
@@ -37,6 +37,9 @@
 #include <OpenMS/CONCEPT/ProgressLogger.h>
 #include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
 #include <OpenMS/FORMAT/XMLFile.h>
+#include <OpenMS/FORMAT/ControlledVocabulary.h>
+#include <OpenMS/KERNEL/MSExperiment.h>
```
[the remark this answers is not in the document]
axelwalter: moved them to QcMLFile.cpp

src/tests/topp/QCCalculator_2_output.mzQC
axelwalter: Yes it validates, json.loads() loads the file content into a json object without errors.

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,244 @@
...
+    
+    for (const auto& p : tic)
+    {
+      intensities.push_back(p.getIntensity());
+      retention_times.push_back(p.getRT());
+    }
+
+    UInt tic_area = intensities[0];
+
+    for (UInt i = 1; i < intensities.size(); ++i)
```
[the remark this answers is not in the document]
axelwalter: I agree, separating the algorithm in classes is a good idea. Maybe we can discuss at the developer meeting how to implement it together QCBase so they can be used in your QualityControl and the QCCalculator.

src/tests/topp/QCCalculator_2_output.mzQC
axelwalter: nice tool, there were some issues that are fixed now and the output validates against this tool

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -0,0 +1,244 @@
...
+          {"uri", "https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo"},
+          {"version", "4.1.49"}
+        }
+    };
+    ControlledVocabulary cv;
+    cv.loadFromOBO("PSI-MS", File::find("/CV/psi-ms.obo"));
+    cv.loadFromOBO("QC", File::find("/CV/qc-cv.obo"));
+
+    // create qualityMetric in json format
+    auto addValue = [&cv](const String& accession, const String& name, const auto& value) -> json
```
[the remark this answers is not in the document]
axelwalter: I passed accession and name because it is a human readable annotation and if something changes in the OBO it can be changed in one place.  It is also possible to just add the name as a code comment and fetch the name from the CV by accession.
In this tool there will be only terms that are officially supported.

timosachsenberg:
ok looks like we have a mergable feature that can be extended on e.g. the user meeting

Issue #5223, opened 2021-03-11
Github actions issues


jpfeuffer:
- [x] It is not even building for PRs
- [x] Boost Debug is still on.
- [x] Names between travis and gh actions do not match (pr123 vs pr-123).
- [ ] Suffix for pyopenms builds missing.
- [x] CMake version needs to be changed (see #5222 )
- [ ] Too slow (at least play with threads param), caching would be nice (at least of contrib and whatever we download)
- [ ] Parallel level should be made consistent between building and testing
- [x] Thirdparty tests would be nice if they don't add too much runtime

jpfeuffer:
Ping @Leon-Bichmann

Leon-Bichmann:
ok will make another PR to fix these

oliveralka:
Why is the Github action shown as success, even if build failed on all platforms? 
e.g. https://github.com/OpenMS/OpenMS/actions/runs/921576251

tapaswenipathak:
can this be closed?

jpfeuffer:
I think this is still a problem

PR #5242, opened 2021-03-16
Fix/GitHub actions

Leon-Bichmann:
This PR tries to fix current encountered github_actions issues #5223


Leon-Bichmann:
Ok, so resetting the CTEST_BUILD_FLAG option makes the CI builds run on unix and osx in about 55min. Not sure how to include this option on windows though and I think that is why we tried to remove it.

About the THIRDPARTY tests : Im not sure how to include them.

Neither how to change the CMake version.

jpfeuffer:
MSBuild for Windows has just different options. I would try "/p:CL_MPCount=NUMBER".

Clone the THIRDPARTY submodule, flatten the directory structure and add them to the path:
https://github.com/OpenMS/OpenMS/blob/develop/tools/travis/lnx-cibuild.before.sh#L63
https://github.com/OpenMS/OpenMS/blob/develop/tools/travis/lnx-cibuild.sh#L63

You have to google for how to install different CMake version on GHactions Windows. You probably have to use either chocolatey or an action from the marketplace.

Leon-Bichmann:
Nice that the builds are working again.

oliveralka:
I don't get why the win build is so much slower than the others.

.github/workflows/openms-ci.yml
```diff
@@ -94,8 +102,10 @@ jobs:
           WITH_GUI: "ON"
           ADDRESS_SANITIZER: "Off"
           BUILD_TYPE: "Release"
-          OPENMP: "On"
+          OPENMP: "Off"
```
Leon-Bichmann, line 105: Not sure either - I thought its because of this. @jpfeuffer why did you disable it?

Leon-Bichmann:
Hey,
What is the status on this PR, anything still missing? Do we leave the windows build as it is?

.github/workflows/openms-ci.yml
jpfeuffer: I hoped that it would reduce compile times

jpfeuffer:
you can have a look at the associated issue. I think the only thing that would help on windows is a compiler cache.

Issue #5221, opened 2021-03-11
Optimize ProteomicsLFQ for single files


jpfeuffer:
I.e. do not do unnecessary stuff when only one file is given:
E.g. transfer_ids would requantify at the exact same places.

jpfeuffer:
The linked PR fixes the most obvious error but I think we should give it another check @timosachsenberg

PR #5243, opened 2021-03-17
[Fix] plfq cornercases and exp. design checks

### LINKED ISSUE #5224: ProteomicsLFQ exits with segfault *in the end* when experimental design numbering is not correct
I.e. not consecutive or not starting from 1.
We have seen cases where it worked but this was probably lucky.
This needs to be safer. And even if it cant be made much safer since we need to assume that ordering, then we need to check very carefully and very early, already during reading, or better the construction of an experimental design object.

jpfeuffer:
# Description

addresses #5221 partly
fixes #5224


src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -599,8 +599,55 @@ namespace OpenMS
...
+      {
+        throw Exception::InvalidValue(
+            __FILE__,
+            __LINE__,
+            OPENMS_PRETTY_FUNCTION, "Fraction groups do not start with 1.", String(msfile_section_[0].fraction_group));
+      }
+
+      Size last_fraction = 1;
+      Size last_fraction_group = 0;
       for (const MSFileSectionEntry& row : msfile_section_)
```
timosachsenberg, line 619: add comment above for loop:
// ensure that fraction group and fraction identifier are consecutive numbers

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -599,8 +599,55 @@ namespace OpenMS
       std::set< std::tuple< std::string, unsigned > > path_label_set;
       std::map< std::tuple< unsigned, unsigned >, std::set< unsigned > > fractiongroup_label_to_sample;
 
+      if (msfile_section_[0].fraction != 1)
```
timosachsenberg, line 602: add comment above:
// ensure that fraction group and fraction identifiers start with 1

src/openms/source/METADATA/ExperimentalDesign.cpp
jpfeuffer: isnt that obvious from the exception?

src/openms/source/METADATA/ExperimentalDesign.cpp
jpfeuffer: see other

src/utils/ProteomicsLFQ.cpp
```diff
@@ -1709,7 +1709,7 @@ class ProteomicsLFQ :
 
         if (e != EXECUTION_OK) { return e; }
         
-        if (getStringOption_("transfer_ids") != "false")
+        if (getStringOption_("transfer_ids") != "false" && ms_files.second.size() > 1)
         {
```
timosachsenberg, line 1713:
```suggestion
        {  // transfer IDs between multiple ms files
```

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -599,8 +599,55 @@ namespace OpenMS
...
+      if (msfile_section_[0].fraction_group != 1)
+      {
+        throw Exception::InvalidValue(
+            __FILE__,
+            __LINE__,
+            OPENMS_PRETTY_FUNCTION, "Fraction groups do not start with 1.", String(msfile_section_[0].fraction_group));
+      }
+
+      Size last_fraction = 1;
+      Size last_fraction_group = 0;
```
timosachsenberg, line 618:
```suggestion
      Size last_fraction_group = 0;
      // ensure that fraction group and fraction identifier are consecutive numbers
```

src/openms/source/METADATA/ExperimentalDesign.cpp
jpfeuffer: and the loop does not only do that

src/openms/source/METADATA/ExperimentalDesign.cpp
timosachsenberg, line 602:
```suggestion
      // ensure that fraction group and fraction identifiers start with 1
      if (msfile_section_[0].fraction != 1)
```

timosachsenberg (approved):

PR #5247, opened 2021-03-19
[FIX] Make ProgressLogger destructor virtual

jpfeuffer:
# Description

Avoid memory leaks in tests where we often create pointers of TOPPBase tools


timosachsenberg:
failed percolator probably related to thirdparty changes

timosachsenberg (approved):

PR #5249, opened 2021-03-20
[Fix] Memory leaks

jpfeuffer:
# Description

Especially in MzIdentML parser!

### DIFF
--- a/src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
+++ b/src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
@@ -113,9 +113,9 @@ private:
     // pointers of MassTraces contained in isotopic pattern
     std::vector<const MassTrace*> iso_pattern_;
 
-    double feat_score_;
+    double feat_score_{};
 
-    SignedSize charge_;
+    SignedSize charge_{};
   };
 
   class OPENMS_DLLAPI CmpMassTraceByMZ
--- a/src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
+++ b/src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
@@ -311,7 +311,7 @@ e conditions are fulfilled. Mainly the
     void findLocalFeatures_(const std::vector<const MassTrace*>& candidates, double total_intensity, std::vector<FeatureHypothesis>& output_hypotheses) const;
 
     /// SVM parameters
-    svm_model* isotope_filt_svm_;
+    svm_model* isotope_filt_svm_ = nullptr;
     std::vector<double> svm_feat_centers_;
     std::vector<double> svm_feat_scales_;
 
--- a/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
+++ b/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
@@ -72,9 +72,6 @@ namespace OpenMS
     template<typename T>
     class shared_xerces_ptr
     {
-      // The actual data we're holding
-      std::shared_ptr<T> item_;
-
       // Function to release Xerces data type with a release member function
       template<typename U>
       static void doRelease_(U* item)
@@ -83,17 +80,12 @@ namespace OpenMS
         if (nullptr == item->getOwnerDocument())
           item->release();
       }
-      // Specializations for character types, released by XMLString::release
-      template<>
-      static void doRelease_(char* item)
-      {
-        xercesc::XMLString::release(&item);
-      }
-      template<>
-      static void doRelease_(XMLCh* item)
-      {
-        xercesc::XMLString::release(&item);
-      }
+
+      static void doRelease_(char* item);
+      static void doRelease_(XMLCh* item);
+
+      // The actual data we're holding
+      std::shared_ptr<T> item_;
     public:
       // Default constructor
       shared_xerces_ptr() = default;
@@ -137,8 +129,8 @@ namespace OpenMS
     class unique_xerces_ptr
     {
     private:
-      // Function to release Xerces data type
-      template <typename U>
+
+      template<typename U>
       static void doRelease_(U*& item)
       {
         // Only release this if it has no parent (otherwise
@@ -147,20 +139,9 @@ namespace OpenMS
           item->release();
       }
 
-      // Specializations for character types, which needs to be
-      // released by XMLString::release
-      template <>
-      static void doRelease_(char*& item)
-      {
-        xercesc::XMLString::release(&item);
-      }
+      static void doRelease_(char*& item);
+      static void doRelease_(XMLCh*& item);
 
-      template <>
-      static void doRelease_(XMLCh*& item)
-      {
-        xercesc::XMLString::release(&item);
-      }
-      // The actual data we're holding
       T* item_;
 
     public:
@@ -226,6 +207,7 @@ namespace OpenMS
         return (nullptr == item_);
       }
     };
+
     //========================================================================================================
 
     /*
--- a/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
+++ b/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
@@ -70,7 +70,7 @@ namespace OpenMS
     //Under BSD license
     //========================================================================================================
     template<typename T>
-    class shared_xerces_ptr
+    class OPENMS_DLLAPI shared_xerces_ptr
     {
       // Function to release Xerces data type with a release member function
       template<typename U>
@@ -126,7 +126,7 @@ namespace OpenMS
     };
 
     template <typename T>
-    class unique_xerces_ptr
+    class OPENMS_DLLAPI unique_xerces_ptr
     {
     private:
 
--- a/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
+++ b/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
@@ -59,6 +59,174 @@ namespace OpenMS
   namespace Internal
   {
 
+    #define CONST_XMLCH(s) reinterpret_cast<const ::XMLCh*>(u ## s)
+
+    static_assert(sizeof(::XMLCh) == sizeof(char16_t),
+                  "XMLCh is not sized correctly for UTF-16.");
+
+    //Adapted from https://www.codeproject.com/articles/99551/redux-raii-adapter-for-xerces
+    //Copyright 2010 Orjan Westin
+    //Under BSD license
+    //========================================================================================================
+    template<typename T>
+    class shared_xerces_ptr
+    {
+      // The actual data we're holding
+      std::shared_ptr<T> item_;
+
+      // Function to release Xerces data type with a release member function
+      template<typename U>
+      static void doRelease_(U* item)
+      {
+        // Only release this if it has no owner
+        if (nullptr == item->getOwnerDocument())
+          item->release();
+      }
+      // Specializations for character types, released by XMLString::release
+      template<>
+      static void doRelease_(char* item)
+      {
+        xercesc::XMLString::release(&item);
+      }
+      template<>
+      static void doRelease_(XMLCh* item)
+      {
+        xercesc::XMLString::release(&item);
+      }
+    public:
+      // Default constructor
+      shared_xerces_ptr() = default;
+      // Assignment constructor
+      shared_xerces_ptr(T* item)
+          : item_(item, doRelease_ )
+      {}
+      // Assignment of data to guard
+      shared_xerces_ptr& operator=(T* item)
+      {
+        assign(item);
+        return *this;
+      }
+      // Give up hold on data
+      void reset()
+      {
+        item_.reset();
+      }
+      // Release currently held data, if any, to hold another
+      void assign(T* item)
+      {
+        item_.reset(item, doRelease_ );
+      }
+      // Get pointer to the currently held data, if any
+      T* get()
+      {
+        return item_.get();
+      }
+      const T* get() const
+      {
+        return item_.get();
+      }
+      // Return true if no data is held
+      bool is_released() const
+      {
+        return (nullptr == item_.get());
+      }
+    };
+
+    template <typename T>
+    class unique_xerces_ptr
+    {
+    private:
+      // Function to release Xerces data type
+      template <typename U>
+      static void doRelease_(U*& item)
+      {
+        // Only release this if it has no parent (otherwise
+        // parent will release it)
+        if (nullptr == item->getOwnerDocument())
+          item->release();
+      }
+
+      // Specializations for character types, which needs to be
+      // released by XMLString::release
+      template <>
+      static void doRelease_(char*& item)
+      {
+        xercesc::XMLString::release(&item);
+      }
+
+      template <>
+      static void doRelease_(XMLCh*& item)
+      {
+        xercesc::XMLString::release(&item);
+      }
+      // The actual data we're holding
+      T* item_;
+
+    public:
+
+      // Hide copy constructor and assignment operator
+      unique_xerces_ptr(const unique_xerces_ptr&) = delete;
+      unique_xerces_ptr& operator=(const unique_xerces_ptr&) = delete;
+
+      unique_xerces_ptr()
+          : item_(nullptr)
+      {}
+
+      explicit unique_xerces_ptr(T* i)
+          : item_(i)
+      {}
+
+      ~unique_xerces_ptr()
+      {
+        xerces_release();
+      }
+
+      // Assignment of data to guard (not chainable)
+      void operator=(T* i)
+      {
+        reassign(i);
+      }
+
+      // Release held data (i.e. delete/free it)
+      void xerces_release()
+      {
+        if (!is_released())
+        {
+          // Use type-specific release mechanism
+          doRelease_(item_);
+          item_ = nullptr;
+        }
+      }
+
+      // Give up held data (i.e. return data without releasing)
+      T* yield()
+      {
+        T* tempItem = item_;
+        item_ = nullptr;
+        return tempItem;
+      }
+
+      // Release currently held data, if any, to hold another
+      void assign(T* i)
+      {
+        xerces_release();
+        item_ = i;
+      }
+
+      // Get pointer to the currently held data, if any
+      T* get() const
+      {
+        return item_;
+      }
+
+      // Return true if no data is held
+      bool is_released() const
+      {
+        return (nullptr == item_);
+      }
+    };
+    //========================================================================================================
+
     /*
      * @brief Helper class for XML parsing that handles the conversions of Xerces strings
      *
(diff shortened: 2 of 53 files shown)


src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/InterpolationModel.h
```diff
@@ -94,6 +94,8 @@ namespace OpenMS
       interpolation_ = source.interpolation_;
       scaling_ = source.scaling_;
 
+      updateMembers_();
```
timosachsenberg, line 97: why is this needed?

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -329,7 +329,7 @@ namespace OpenMS
 
   FeatureFindingMetabo::~FeatureFindingMetabo()
   {
-
+    delete isotope_filt_svm_;
```
timosachsenberg, line 332: could we use a unique_ptr as well here?

src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp
```diff
@@ -449,16 +449,16 @@ namespace OpenMS
...
 
-    static const XMLCh* TAG_CV = xercesc::XMLString::transcode("cvParam");
-    static const XMLCh* TAG_binary = xercesc::XMLString::transcode("binary");
-    static const XMLCh* TAG_userParam = xercesc::XMLString::transcode("userParam");
-    static const XMLCh* TAG_referenceableParamGroupRef = xercesc::XMLString::transcode("referenceableParamGroupRef");
-    static const XMLCh* TAG_accession = xercesc::XMLString::transcode("accession");
-    static const XMLCh* TAG_unit_accession = xercesc::XMLString::transcode("unitAccession");
-    static const XMLCh* TAG_value = xercesc::XMLString::transcode("value");
-    static const XMLCh* TAG_name = xercesc::XMLString::transcode("name");
+    static const XMLCh* TAG_CV = CONST_XMLCH("cvParam");
```
timosachsenberg, line 454: can you add some comment here why the macros are used?

src/openms/source/FORMAT/HANDLERS/MzMLSpectrumDecoder.cpp
```diff
@@ -547,9 +547,11 @@ namespace OpenMS
   std::string MzMLSpectrumDecoder::domParseString_(const std::string& in, std::vector<BinaryData>& data)
   {
     // PRECONDITON is below (since we first need to do XML parsing before validating)
-    static const XMLCh* default_array_length_tag = xercesc::XMLString::transcode("defaultArrayLength");
-    static const XMLCh* id_tag = xercesc::XMLString::transcode("id");
-    static const XMLCh* binary_data_array_tag = xercesc::XMLString::transcode("binaryDataArray");
+    static constexpr XMLCh id_tag[] = {'i','d', 0};
+    static constexpr XMLCh default_array_length_tag[] = { 'd','e','f','a','u','l','t','A','r','r','a','y','L','e','n','g','t','h', 0};
```
timosachsenberg, line 551: can you add a comment why it is written like this? e.g., initializer list of uint16?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EGHTraceFitter.cpp
```diff
@@ -187,6 +185,7 @@ namespace OpenMS
     this->apex_rt_ = source.apex_rt_;
```
timosachsenberg, line 185: did you check if the whole function is just the default copy assignment and we could write = default in the header? same for copy constructor

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EmgFitter1D.cpp
```diff
@@ -231,7 +231,7 @@ namespace OpenMS
     {
       correlation = -1.0;
     }
-
+    delete model;
```
timosachsenberg, line 234: RAII?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderAlgorithmMRM.cpp
```diff
@@ -374,7 +374,7 @@ namespace OpenMS
               std::cerr << "An error occurred during the gnuplot execution" << std::endl;
             }
           }
-
+          delete model_rt;
```
timosachsenberg, line 377: use RAII?

src/tests/class_tests/openms/source/AverageLinkage_test.cpp
```diff
@@ -157,6 +157,7 @@ START_SECTION((static ClusterFunctor* create()))
 {
 	ClusterFunctor* cf = AverageLinkage::create();
   TEST_NOT_EQUAL( dynamic_cast<AverageLinkage*>(cf) , nullPointer)
+  delete cf;
```
timosachsenberg, line 160: RAII with unique_ptr?

src/tests/class_tests/openms/source/BiGaussFitter1D_test.cpp
```diff
@@ -122,6 +122,7 @@ START_SECTION((Fitter1D* create()))
   Fitter1D* ptr = BiGaussFitter1D::create();
   TEST_EQUAL(ptr->getName(), "BiGaussFitter1D")
 	TEST_NOT_EQUAL(ptr, nullPointer)
+	delete ptr;
```
timosachsenberg, line 125: same here?

src/tests/class_tests/openms/source/ControlledVocabulary_test.cpp
```diff
@@ -188,6 +188,7 @@ START_SECTION(([ControlledVocabulary::CVTerm] CVTerm()))
 {
   cvterm = new ControlledVocabulary::CVTerm();
   TEST_NOT_EQUAL(cvterm, cvtermNullPointer)
+  delete cvterm;
```
timosachsenberg, line 191: and here

src/topp/SpecLibSearcher.cpp
```diff
@@ -358,7 +358,7 @@ class TOPPSpecLibSearcher :
     OPENMS_LOG_INFO << "Time needed for preprocessing data: " << (end_build_time - start_build_time) << "\n";
 
     //compare function
-    PeakSpectrumCompareFunctor* comparor = Factory<PeakSpectrumCompareFunctor>::create(compare_function);
+    PeakSpectrumCompareFunctor* comparator = Factory<PeakSpectrumCompareFunctor>::create(compare_function);
```
timosachsenberg, line 361: wrap in unique ptr?

timosachsenberg:
nice cleanup

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/InterpolationModel.h
jpfeuffer: It says in the description of updateMembers that it should be at every derived copy constructor/assignment

timosachsenberg: ok makes sense. One thing to watch out as it is easy to just put a = default at a copy constructor/assignment

src/tests/class_tests/openms/source/AverageLinkage_test.cpp
jpfeuffer: I dont think it is worth for the tests where you delete 2 lines after. Feel free to do it if you think it is important.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EGHTraceFitter.cpp
jpfeuffer: it has update_members

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
jpfeuffer: No, needs special deleter.

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/InterpolationModel.h
jpfeuffer: I am actually not sure why updateMembers is needed. Only makes a difference if the params of the source do not match its own members right. But this prevents = default everywhere.

jpfeuffer:
By the way, I am giving up on SVMWrapper. It will continue leaking, especially during CV but the class is so incredibly badly designed that it is impossible to see where things get allocated. Additionally, there are two types of modes (one heap, one stack-allocated) which are only separated by if-cases in all of the methods.

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/FeatureFindingMetabo.h
```diff
@@ -311,7 +311,7 @@ e conditions are fulfilled. Mainly the
     void findLocalFeatures_(const std::vector<const MassTrace*>& candidates, double total_intensity, std::vector<FeatureHypothesis>& output_hypotheses) const;
 
     /// SVM parameters
-    svm_model* isotope_filt_svm_;
+    svm_model* isotope_filt_svm_ = nullptr;
```
timosachsenberg, line 314: just FYI: you can also use {}

src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h
```diff
@@ -59,6 +60,156 @@ namespace OpenMS
   namespace Internal
   {
 
+    #define CONST_XMLCH(s) reinterpret_cast<const ::XMLCh*>(u ## s)
```
timosachsenberg, line 63:
```suggestion
    // put u" in front of string literal to mark it as UTF-16 encoded, can then be directly convted to const XMLCh*
    #define CONST_XMLCH(s) reinterpret_cast<const ::XMLCh*>(u ## s)
```

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/ExtendedIsotopeFitter1D.h
```diff
@@ -73,7 +73,7 @@ namespace OpenMS
     }
 
     /// return interpolation model
-    QualityType fit1d(const RawDataArrayType & range, InterpolationModel * & model) override;
+    QualityType fit1d(const RawDataArrayType & range, std::unique_ptr<InterpolationModel>& model) override;
```
timosachsenberg, line 76: can we wrap that in pyOpenMS?

timosachsenberg: and would it make sense at all because of polymorphism ???

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/InterpolationModel.h
timosachsenberg: I think we would need to check if someone sets the param class in updateMembers (which I think he should not do).

src/openms/source/ANALYSIS/SVM/SVMWrapper.cpp
```diff
@@ -1658,6 +1656,9 @@ namespace OpenMS
       return nullptr;
     }
 
+    double temp = 0;
+    svm_problem* kernel_matrix;
```
timosachsenberg, line 1660: why does this need to be a pointer? because it otherwise allocates a lot of memory on the stack?

timosachsenberg: svm_problem_ could also be a unique_ptr AFAIK. Copy elision will allow returning it if I understood the concept

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
```diff
@@ -329,7 +304,10 @@ namespace OpenMS
 
   FeatureFindingMetabo::~FeatureFindingMetabo()
   {
-
+    if(isotope_filt_svm_ != nullptr)
```
timosachsenberg, line 307: coudl isotope_filt_svm_ be a unique_ptr and svm_free_and_destroy_model(&isotope_filt_svm_); be called in the lambda passed for destruction?  then this destructor could be default

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -187,25 +188,36 @@ namespace OpenMS
...
+        //          ostringstream errBuf;
+        //          errBuf << "Error parsing file: " << message << flush;
+        OPENMS_LOG_ERROR << "XERCES error parsing file: " << message << flush << endl;
+        XMLString::release(&message);
+      }
 
-        // no need to free this pointer - owned by the parent parser object
-        xercesc::DOMDocument* xmlDoc = mzid_parser_.getDocument();
+      // we adopt/own the document so we can free it in case this DOMHandler is reused on another file.
+      xercesc::DOMDocument* xmlDoc = mzid_parser_.adoptDocument();
```
timosachsenberg, line 202: does this imply any performance / memory hit?

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
```diff
@@ -1413,57 +1426,57 @@ namespace OpenMS
       {
         DOMElement* cl_sii = dynamic_cast<xercesc::DOMElement*>(siis->item(it->second));
         // Attributes
-        String peptide = XMLString::transcode(cl_sii->getAttribute(XMLString::transcode("peptide_ref")));
+        String peptide = StringManager::convert(cl_sii->getAttribute(CONST_XMLCH("peptide_ref")));
         peptides.push_back(peptide);
-        double exp_mz = String(XMLString::transcode(cl_sii->getAttribute(XMLString::transcode("experimentalMassToCharge")))).toDouble();
+        double exp_mz =StringManager::convert(cl_sii->getAttribute(CONST_XMLCH("experimentalMassToCharge"))).toDouble();
```
timosachsenberg, line 1431:
```suggestion
        double exp_mz = StringManager::convert(cl_sii->getAttribute(CONST_XMLCH("experimentalMassToCharge"))).toDouble();
```

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EGHTraceFitter.cpp
```diff
@@ -161,10 +161,7 @@ namespace OpenMS
...
   }
 
-  EGHTraceFitter::EGHTraceFitter()
-  {
-    //setName("EGHTraceFitter");
-  }
+  EGHTraceFitter::EGHTraceFitter() = default;
 
   EGHTraceFitter::EGHTraceFitter(const EGHTraceFitter& other) :
     TraceFitter(other)
```
timosachsenberg, line 167: could potentially also be deafult

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/GaussFitter1D.cpp
```diff
@@ -89,7 +89,7 @@ namespace OpenMS
 
 
     // build model
-    model = static_cast<InterpolationModel*>(Factory<BaseModel<1> >::create("GaussModel"));
+    model = std::unique_ptr<InterpolationModel>(dynamic_cast<InterpolationModel*>(Factory<BaseModel<1>>::create("GaussModel")));
```
timosachsenberg, line 92: maybe check if all these casts are needed.
I think the Factory could in principle directly return a std::unique_ptr<InterpolationModel>

src/tests/class_tests/openms/source/AverageLinkage_test.cpp
timosachsenberg: my argument would be: someone forgot it here :P

timosachsenberg (approved):

src/openms/source/FORMAT/HANDLERS/MzIdentMLDOMHandler.cpp
jpfeuffer: AFAIK not. It is only a difference in ownership of the memory. Either the parser has it, or the class that calls it.

src/openms/source/ANALYSIS/SVM/SVMWrapper.cpp
jpfeuffer: I don't know, just moved here. The svm_problems usually don't have to be pointers. But svm_models etc. you get back as pointers from libSVM. So, yes the correct way would be to use unique_ptrs for svm_model with custom deleter and do not use pointers for svm_problems until you pass it to libSVM I guess.

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
jpfeuffer: yes please

jpfeuffer: has to be done in all of SVMWrapper though.

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/ExtendedIsotopeFitter1D.h
jpfeuffer: Of course, you can. But not with autowrap.

jpfeuffer: Easy workaround is to offer a function that accepts a shared pointer since autowrap has a conversion provider for this (essentially passing the "inst" member of the already wrapped python class).

hroest: or just pass a smart_ptr ?

timosachsenberg: introducing a shared_ptr where a unique_ptr would suffice is bad design.
is there a possibility to wrap unique_ptr? this would be cleaner.

src/openms/source/ANALYSIS/SVM/SVMWrapper.cpp
jpfeuffer: This shoudl be addressed in a general SVM cleanup

src/openms/source/FILTERING/DATAREDUCTION/FeatureFindingMetabo.cpp
jpfeuffer: This should be addressed in a general SVM cleanup

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/EGHTraceFitter.cpp
jpfeuffer: update members?

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/GaussFitter1D.cpp
jpfeuffer: yes they are needed

PR #5251, opened 2021-03-24
FIX: Python Bindings SiriusAdapter

### LINKED ISSUE #5184: Segmentation fault (core dumped)
Hi,

I am trying to load mzML files with 500MB approximate size and I am constantly getting this segmentation fault error. However, when I use smaller files, this doesn't occur and my workflow runs as it is supposed to. This is not a memory issue since I am using a machine that has a lot of space. Is this a c++ error? 

import numpy as np
import pandas as pd
import pyopenms
from pyopenms import *

exp = MSExperiment()

import sys
print("Loading")
MzMLFile().load(sys.argv[1], exp)
print("Loaded")

oliveralka:
# Description
fixes  #5184

Issue: SIRIUSAdapter::store failed in pyopenms. 
The problem was that the BaseFeature* tried to access memory, which was altered by copying the FeatureMap internally in cython to read and write a python list. 
This was fixed by adding an additional class to wrap the vector<FeatureMap> and the KDTreeMaps. 

Also added a POSTCONDITION if the output directory of SIRIUS  is empty. 

# Checklist:


oliveralka:
rebuild jenkins

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/FeatureMapping.h
```diff
@@ -45,9 +46,17 @@ namespace OpenMS
       {
           public:
 
-          struct FeatureToMs2Indices
+          class FeatureMappingInfo
           {
-             Map<const BaseFeature*, std::vector<size_t>> assignedMS2;
+          public:
+            std::vector<FeatureMap> v_fp;
```
timosachsenberg, line 52: please add documentation and some description. 
maybe rename the variables to something more telling.
std::vector<FeatureMap> feature_maps; // the actual feature data
KDTreeFeatureMaps kd_tree; // kdTree references into feature_maps and provides fast spatial queries

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -508,12 +506,12 @@ namespace OpenMS
                                   });
           feature_map.erase(map_it, feature_map.end());
   
-          v_fp.push_back(feature_map);
-          fp_map_kd.addMaps(v_fp);
+          fm_info.v_fp.push_back(feature_map);
+          fm_info.fp_map_kd.addMaps(fm_info.v_fp);
```
timosachsenberg, line 510: add comment that kDTree references into feature_map

timosachsenberg:
good job in finding a solution 👍

oliveralka:
Thanks for the help with spotting the error! 👍

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/FeatureMapping.h
```diff
@@ -45,9 +46,19 @@ namespace OpenMS
       {
           public:
 
-          struct FeatureToMs2Indices
+          /// Stores information required for preprocessing
+          class FeatureMappingInfo
```
KyowonJeong, line 50: Well, this might not be a comment from a reviewer, but I am still confused why sometimes I use class and other structure.

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
```diff
@@ -72,10 +72,10 @@ String SiriusMzTabWriter::extractFeatureId(const String& path)
   return feature_id;
 }
 
-std::map< String, Size > SiriusMzTabWriter::extract_columnname_to_columnindex(CsvFile& csvfile)
+std::map< std::string, Size > SiriusMzTabWriter::extract_columnname_to_columnindex(CsvFile& csvfile)
 {
   StringList header_row;
-  std::map< String, Size > columnname_to_columnindex;
+  std::map< std::string, Size > columnname_to_columnindex;
```
KyowonJeong, line 78: I have been using String. Should I change them to std::string as well? I will.

KyowonJeong (commented):
Looks OK! (as far as my humble review ability concerned)...

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
timosachsenberg: In practice not that important I guess. I try to use std::string whenever it is sufficient.

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/FeatureMapping.h
timosachsenberg: There are no clear rules but in general it is good to use structs for simple records / PODs and switch to class if you have e.g. private or protected members.
See e.g., with some hints when not to use class 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c-classes-and-class-hierarchies for suggestions how to use them.

src/openms/source/FORMAT/DATAACCESS/SiriusMzTabWriter.cpp
oliveralka: Using `std::string` instead of `OpenMS::String` in the map is due to pyopenms wrapping. In general, you can use `OpenMS::String` without an issue:

```
Most complex STL constructs can be wrapped even if they are nested, however mixing them with user-defined types does not always work, see Limitations below. Nested std::vector constructs work well even with user-defined (OpenMS-defined) types. However, std::map<String, X> does not work (since String is user-defined, however, a primitive C++ type such as std::map<std::string, X> would work).
```
https://pyopenms.readthedocs.io/en/latest/wrap_classes.html

oliveralka:
rebuild jenkins

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/FeatureMapping.h
oliveralka: I have to admit that here it was simply personal preference, this could also be a struct.

oliveralka:
Since the PR was not merged yet, I added fix for the db parameter, since multiple databases are allowed.
The can now be specified as string "db1,db2,db3" - see #5255

KyowonJeong (approved):
The new changes looks fine!! Thanks!

PR #5256, opened 2021-03-27
[Feature] Adding TOPP tool/util params to .TOPPView.ini

dvdvgt:
# Description

This PR is a follow up to #5146 and aims to add the following:

- Instead of only showing TOPP tools in the toolsdialog also list TOPP utils
- Only listing tools/utils which are compatible with the current layer
- Saving tool/util params in the `~/.TOPPView.ini` file so that they can be reused at the next time unless there have been changes (or a reload was requested (still to be added)). For that reason a new section was added: `tool_params:` under which all tool/util params are saved. Checking whether the TOPPView.ini file has all tool/util params in it proved to be problem, since there is no way to check whether a section exists and there is no way to iterate over nodes as well as items (currently only items are iterable). Currently a special version item under `tool_params:version` is used to determine whether all tool/util params have to be generated or can be read from the ini by checking whether it exists (has a value) with `Param::exists`. This solution is the best I could come up without adding to much new code. If there's a better/safer solution I'm happy to incorporate it instead.
- A new class was added (`TVToolDiscovery`) which is responsible for generating a param for each tool/util in the background using `std::async` and saving them in a `unordered_map` for fast retrieval.
- `File::getUniqueName` was made thread-safe by using an `std::atomic<int>` instead of a normal int.
- `File::getExecutablePath` was made thread-safe by using a immediately lambda expression protecting a static function variable for concurrent access.

Node that I deliberately submitted this **PR prematurely** to discuss some issues with you first: 

### Too many threads (Critical)
- [x] By using `std::async` in `TVToolDiscovery::loadParams` too many threads are spawned which, I assume, leads to too many concurrently opened files so that while generating params the [https://ss64.com/bash/ulimit.html](ulimit) of maximum number of open file descriptors is exceeded when TOPPView is started via the shell. This causes TOPPView to crash so this definitely needs to be addressed though I'm not seeing a solution.
The default limit is ulimit is 1024 which should be plenty since there are only ~180 params to be generated. Why are some many threads created? Is it possible to just spawn one thread for generating one param?

Maybe use `boost::thread` instead? Do you have an idea?

### More robust way to check if params can be reused (Moderate)
- [x] As mentioned previously whether the params in TOPPView.ini can be reused is by checking whether the `tool_param:version` entry exists. Do you have a more robust idea in mind how this could also be accomplished?

### Adding tests for the new class
Tests have been omitted in case changes need to be made first.

# Checklist:


jpfeuffer:
Nice, thanks! I don't think there is a way to influence how async processes the tasks. This is up to the standard library and I read that the problem of too many threads is quite common. Why don't you just get the number of possible system threads, and spawn n std::threads (each with a nrtools/n chunk of the tools) on your own? I think boost::thread would be fine but prefer std::thread.

src/openms/source/SYSTEM/File.cpp
```diff
@@ -92,44 +94,44 @@ namespace OpenMS
...
+          } else {
+            std::cerr << "Path extracted from Executable Path does not exist! Returning empty string!\n";
+            rpath = "";
+          }
+        } else {
+          std::cerr << "Cannot get Executable Path! Not using a path prefix!\n";
+        }
 
-    path_checked = true; // enable short route for next run
+        path_checked = true; // enable short route for next run
```
cbielow, line 132: this bool is not needed anymore and could be removed entirely?

cbielow:
I think this was mentioned before but: the most elegant solution is to put all classes derived from TOPPBase into the library. Then you can simply query their parameters (requires slight modification to TOPPBase, but that should be easy). This way, its amazingly fast and there is no need to spawn any additional threads.
Copying the code for all 180 TOPP tools should not take more than 1 minute per tool. Granted, its a bit dull, but you can try with 1-2 tools, and see if it works.

dvdvgt:
> I think this was mentioned before but: the most elegant solution is to put all classes derived from TOPPBase into the library. Then you can simply query their parameters (requires slight modification to TOPPBase, but that should be easy). This way, its amazingly fast and there is no need to spawn any additional threads.
> 
> Copying the code for all 180 TOPP tools should not take more than 1 minute per tool. Granted, its a bit dull, but you can try with 1-2 tools, and see if it works.

I think there was miscommunication as this was not mentioned before or at least broad to me, in fact [you proposed](https://github.com/OpenMS/OpenMS/pull/5146#discussion_r567434828) to use std:async.

As this seems to be a point of friction and there seem to be many different ideas on how to do it most efficiently maybe we could get together to narrow it down to one solution.

timosachsenberg:
> I think this was mentioned before but: the most elegant solution is to put all classes derived from TOPPBase into the library. Then you can simply query their parameters (requires slight modification to TOPPBase, but that should be easy). This way, its amazingly fast and there is no need to spawn any additional threads.
Copying the code for all 180 TOPP tools should not take more than 1 minute per tool. Granted, its a bit dull, but you can try with 1-2 tools, and see if it works.

This idea has indeed been around a few times. Trying to remember the downsides of such an approach.
Obviously we would need to move tool code to the library most likely splitting it in header + cpp file.
Moving code to the library as an algorithm takes about a day or more per tool. (including tests, pyOpenMS binding). Just making tools instantiatable is much faster as everything except the main could be kept. It would require potentially some kind of factory mechanism and a registry to make them discoverable and accessible (in the easiest case one registry and "template factory method" to instantiate).
From a design/architecture point of view this is diluting the barrier between applications and library which might bite us at some other point in the future - not sure. What we would not gain is making TOPPView more pluggable so this would primarily be a speed  increase.  Is it worth it? If we only execute parsing once every new version comes out and already start when TOPPView is opened I am not sure if we need that change. But certainly something to discuss.

timosachsenberg:
Can we have this discussion soon? I don't want to hold this up for much longer as there has already been some back and forth.

src/openms_gui/include/OpenMS/VISUAL/DIALOGS/ToolsDialog.h
```diff
@@ -76,12 +75,13 @@ namespace OpenMS
...
 
         @param parent Qt parent widget
+        @param params Containing all TOPP tool/util params
         @param ini_file The file name of the temporary INI file created by this dialog
         @param default_dir The default directory for loading and storing INI files
         @param layer_type The type of data (determines the applicable tools)
         @param layer_name The name of the selected layer
     */
-    ToolsDialog(QWidget * parent, String ini_file, String default_dir, LayerData::DataType layer_type, String layer_name);
+    ToolsDialog(QWidget * parent, Param params, String ini_file, String default_dir, LayerData::DataType layer_type, String layer_name);
```
cbielow, line 84: const &

jpfeuffer:
I do not like the idea of "just" moving it to the library. 

a) It is quite some effort (although I don't think you would need to create algorithm classes out of it, just implement main_ and the TOPPBase object in a separate header but keep main in the tool) for a minor improvement

b) It is moving in the wrong direction. I would like to see way fewer things in a monolithic library. If at all, then make a ToolsBase library that can be compiled separately and will be linked from TOPPView only. For the tools we can still think if link to that library or just compile-in the tool-specific cpp again.

src/openms_gui/include/OpenMS/VISUAL/DIALOGS/ToolsDialog.h
dvdvgt: Currently the constructor is called with
```
ToolsDialog tools_dialog(this, param_.copy("tool_params:", true), topp_.file_name + "_ini", current_path_, layer.type, layer.getName());
```
So the copy does not live long enough to be passed as reference. Would you suggest just passing the whole `param_`?

cbielow: Temporaries should bind just fine to const references. They do not bind to non-const refs though. Did you use 'const Param& params'??

dvdvgt: You're right. I thought I had tried this before and it did not work.

dvdvgt:
I'm currently testing using `std::thread` instead of `std::async` so that only one thread is spawned per tool param, however the issue of too many threads remains. I suspect that each thread spawns multiple new threads somewhere in `TVToolDiscovery::getParamFromIni_`. Is there any threading involved in parsing XML files? Maybe in `XMLFile::parse_`?

jpfeuffer:
What does too many mean? There are around 180 tools right? So are there more than 180? To reduce them further you would need to split the list into batches and pass them to fewer threads.

dvdvgt:
> ... too many threads are spawned which, I assume, leads to too many concurrently opened files so that while generating params the https://ss64.com/bash/ulimit.html of maximum number of open file descriptors is exceeded when TOPPView is started via the shell.

This is what I meant by too many threads. By default as far as I know each program started from the shell is allowed to have 1024 concurrently opened file handles. However I suspect that this limit is exceeded with the current approach with the following error
```
QProcessPrivate::createPipe: Cannot create pipe 0x7fbaac002798 (Too many open files)
```
when ToppView is started from the shell.

I cannot understand where these +1024 file handles come from as there are only ~180 tools/utils and only one temporary ini file is created for each. Spawning 180 threads at once should be no problem but too many opened file handles is. So I figured that there have to be more than the 180 threads accessing these temporary ini files otherwise there's no way that the limit of 1024 opened file handles could be exceeded.

As I said, this is only an issue when ToppView is started from the shell and I only tested/experienced that on linux.

jpfeuffer:
Can you close the file here before removing:
https://github.com/OpenMS/OpenMS/blob/e5c93297cba4ef0615bb5628e827f677eb298818/src/openms/source/SYSTEM/File.cpp#L344

And can you use a counter or something instead of getUniqueFilename for your inis? Pretty dangerous with multiple threads starting at the same time.

I don't think the number of threads is too high but the number of filehandles each thread creates. I don't know what all the Qt File functions do. 1024/180 doesn't leave much room. Chunking is a must I would say.

timosachsenberg:
I think https://github.com/OpenMS/OpenMS/blob/e5c93297cba4ef0615bb5628e827f677eb298818/src/openms/source/SYSTEM/File.cpp#L450 is only thread safe initialized but access is not.
should be atomic<>

timosachsenberg:
Yes, probably some chunking needs to be done similar to
https://stackoverflow.com/questions/9942869/divide-container-into-chunks-c/9943807

timosachsenberg:
Another possibility would be to take a look at a c17 threadpool implementation like https://github.com/fbastos1/thread_pool_cpp17/ .
Usage is simple enough (though the implementation needs to rely on some newer C17 language features).

cbielow:
> 
> 
> > ... too many threads are spawned which, I assume, leads to too many concurrently opened files so that while generating params the https://ss64.com/bash/ulimit.html of maximum number of open file descriptors is exceeded when TOPPView is started via the shell.
> 
> This is what I meant by too many threads. By default as far as I know each program started from the shell is allowed to have 1024 concurrently opened file handles. However I suspect that this limit is exceeded with the current approach with the following error
> 
> ```
> QProcessPrivate::createPipe: Cannot create pipe 0x7fbaac002798 (Too many open files)
> ```
> 
> when ToppView is started from the shell.
> 
> I cannot understand where these +1024 file handles come from as there are only ~180 tools/utils and only one temporary ini file is created for each. Spawning 180 threads at once should be no problem but too many opened file handles is. So I figured that there have to be more than the 180 threads accessing these temporary ini files otherwise there's no way that the limit of 1024 opened file handles could be exceeded.
> 
> As I said, this is only an issue when ToppView is started from the shell and I only tested/experienced that on linux.

maybe try to list the current file handles when the process is in full swing (can be done internally I guess, If you can obtain the process id -- see File::getUniqueName() on how to do that) and then simply:  https://www.cyberciti.biz/faq/howto-linux-get-list-of-open-files/

dvdvgt:
https://github.com/OpenMS/OpenMS/blob/4d505a1c58cd13d74218745e0f911bc513bc49b3/src/openms_gui/source/VISUAL/TVToolDiscovery.cpp#L119
So apparently adding the file mode `QProcess::NotOpen` fixed the issue of too many file handles. Maybe because each QProcess no longer requests it's own file handle...? Not really sure why this works so I'm still a bit skeptical. 

> And can you use a counter or something instead of getUniqueFilename for your inis? Pretty dangerous with multiple threads starting at the same time. @jpfeuffer 

That should not be a problem because `File::getUniqueName` uses a static [atomic](https://en.cppreference.com/w/cpp/atomic/atomic) int counter for making sure each file name really is unique even when this function is called by multiple threads simultaneously.

jpfeuffer:
Ah, I did not see that. So many things this function is adding.
NotOpen should be fine. Probably only for the input and output streams.
Still, chunking might be good.
And the closing in File::writable: did you check that? Would be nice to fix this in any case. Might reduce some open file handles as well.

dvdvgt:
> And the closing in File::writable: did you check that? Would be nice to fix this in any case. Might reduce some open file handles as well.

I don't think that's necessary. It seems the file [is automatically closed](https://doc.qt.io/qt-5/qfile.html#remove) before being removed when QFile::remove is called.

dvdvgt:
The last two commits addressed some failed tests. I added an enum `TOOL_SCAN` to TOPPViewBase to offer more fine-grained control whether TVToolDiscovery is used to generate util/tool params.

When testing (i.e. CI on jenkis) no TOPPView ini file exists that's why TVToolDiscovery is used to generate tool/util params however because of how `File::getExecutablePath` works the executable path is not correctly found when executing tests. `File::getExecutablePath` checks where the current executable lies and therefore infers that that directory also has to be the executable directory where all tool, utils and so on are. When testing though the test binaries are in `OpenMS/build/src/tests/class_tests/bin` as a consequence the executable path is not found correctly and TVToolDiscovery::getToolFromIni does not work because `File::getSiblingExecutable` does not find any tools/utils. This leads to failing tests.

> Still, chunking might be good. @jpfeuffer 

Currently `std::async` is used and offers a working and robust solution. Chunking might be better however it's also less robust than just using `std::async` and addition to that launching all threads currently already takes about 1 second when TOPPView is started. I'm guessing chunking would increase the startup time even more which is why I'm advocating for `std::async` instead of some custom implementation with `std::thread`. What is your take on that?

src/tests/class_tests/openms/source/Param_test.cpp
```diff
@@ -655,6 +655,16 @@ START_SECTION((bool exists(const String& key) const))
 	TEST_EQUAL(p.exists("key:value"), false)
 END_SECTION
 
+START_SECTION((bool hasSection(const String& key) const))
+  Param p;
+	p.addSection("test", "");
+	p.addSection("test:test", "");
+	TEST_EQUAL(p.hasSection("test"), true)
```
cbielow, line 662: also test for `TEST_EQUAL(p.hasSection("test:"), true)`

src/openms_gui/source/VISUAL/TVToolDiscovery.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    const std::map<std::string, Param>& TVToolDiscovery::getToolParams()
+    {
+      // Make sure threads have been launched and waited for before accessing results
+      loadParams();
+      waitForParams();
+      return params_;
+    }
+
+    Param TVToolDiscovery::getParamFromIni_(const std::string& tool_name)
```
cbielow, line 98: this whole function looks very `manual`.
Look for higher order functions, such as `File::getTemporaryFile` or `ExternalProcess::run` to use a lot less code..

dvdvgt: I don't think `File::getTemporaryFile` is thread-safe. Having the path constructed in a "hacky" way is probably to be preferred over rewriting `File::getTemporaryFile` and all associated functions just for this one use case where it has to be thread-safe. What do you think?
You're right though, it is kind of an ugly function. I'll look into replacing some with preexisting functions.

timosachsenberg: hmm yes doesn't look thread safe to me but this can be easily fixed in that case as far as I see.
add:
std::lock_guard<std::mutex> lk(class_mutex_); 
as first line of every member function of TemporaryFiles_
and a private member std::mutex class_mutex_;

dvdvgt: I've looked into `ExternalProcess::run` as well. However the problem with too many open files handles I described previously reoccurred. I could fix that by adding an additional bool parameter with a default value (so that no existing calls need to be changed) of false to `ExternalProccess::run` like `read_only` so that
```cpp
ExternalProcess::RETURNSTATE ExternalProcess::run(const QString& exe, const QStringList& args, const QString& working_dir, const bool verbose, String& error_msg, bool read_only)
{
  // ...
  qp_->start(exe, args, read_only? QProcess::NotOpen : QProcess::ReadWrite) // opposed to qp_->start(exe, args)
  // ...
}
```

Would that be a possible workaround?

timosachsenberg: I think this should work. Maybe call the parameter different in our API e.g. noIO and add a comment/documentation that this leads to fewer threads opened by the process?

cbielow: concerning thread-safe `getTemporaryFile`: you should get away with `static std::atomic<int> number = 0;` in `File::getUniqueName` and a thread-local storage of `TemporaryFiles_::filenames_`

cbielow: concerning `ExternalProcess::run`: bool parameters are evil (because you never know what `someFunc(true, true, false)` is actually doing unless you look it up. Better is `someFunc(std::ios::in, MyEnum::truncate, Log::verbose)`, since its much more informative.
So, I'd suggest to just pass the parameter right through and make `QIODevice::OpenModeFlag mode = QIODevice::ReadWrite` the default. This incurs adding a Qt include which is not optimal. You could avoid this by defining your own enum with common modes and translate that to the Qt mode inside the cpp. This way you have the best of both worlds.

timosachsenberg: > concerning thread-safe `getTemporaryFile`: you should get away with `static std::atomic<int> number = 0;` in `File::getUniqueName` and a thread-local storage of `TemporaryFiles_::filenames_`

I think both works. Class level locking is simpler but slower. Though the overhead here is probably never an issue.

dvdvgt: > concerning thread-safe `getTemporaryFile`: you should get away with `static std::atomic<int> number = 0;` in `File::getUniqueName` and a thread-local storage of `TemporaryFiles_::filenames_`

Can you elaborate how you would go about making `TemporaryFiles_::filenames_` thread-local and how that would help with thread-safety while still maintaining the same functionality?

 As far as I understood it a `thread_local` leads to each thread having it's own copy of a variable as opposed to locking the `filenames_` vector behind a `mutex` where each thread still accesses the same vector. The first approach seems odd since this would mean that each thread creates new temporary files and pushes the filename onto a thread local copy of `filenames_` so that upon destruction not all temporary file will be deleted because all the created filenames are scattered across different copies of `filenames_`.
Also [a `thread_local` data member has also be declared static](https://stackoverflow.com/a/32365713). This will probably only complicate things even more. 

A different and probably much simpler idea is just to generate a unique folder in `/tmp` (or whatever the windows and macOS equivalent is) like `/tmp/OpenMS{CurrentDate}{CurrentTime}` and put all temporary files in there and just delete the whole folder upon destruction. That way it's inherently safe and no mutex is needed because it's not necessary to keep track of all filenames. What do you think of that?

dvdvgt:
> There should be no need to expose this, I.e. to bake the translation of flags into the interface. Can be done in cpp only. @cbielow 

What do you mean by "bake the translation of flags into the interface"? Can you elaborate? Which interface?

dvdvgt:
rebuild jenkins

timosachsenberg:
looks good from my side

timosachsenberg (approved):

jpfeuffer:
```
Processing script file: "C:/jenkins/ws/openms/ntly/TstPkg/9447518b/build/_CPack_Packages/win64/NSIS/project.nsi" (ACP)

File: failed opening file "C:/jenkins/ws/openms/ntly/TstPkg/9447518b/build/_CPack_Packages/win64/NSIS/OpenMS-2.6.0-pre-nightly-2021-05-04-Win64/share/doc\html\C_1_2jenkins_2ws_2openms_2ntly_2TstPkg_29447518b_2source_2src_2openms_gui_2include_2OpenMS_2VISUAL_2TVToolDiscovery_8h-example.html"

Error in script "C:/jenkins/ws/openms/ntly/TstPkg/9447518b/build/_CPack_Packages/win64/NSIS/project.nsi" on line 284 -- aborting creation process
```

Hi, unfortunately this broke packaging. Please double-check the doxygen documentation of TVToolDiscovery (especially the @ example)

timosachsenberg:
I can take a look

jpfeuffer:
Probably Windows path length. Good luck ;)

jpfeuffer:
But I am also not sure if you can use "@example" without a filename or what that implies.

timosachsenberg:
yeah that was what I also found odd. I removed it as we never used @example in that context before

PR #5263, opened 2021-04-01
[DOC] Extend description in FileMerger

### LINKED ISSUE #5257: mzTAB export fails after merging multiple fractions
When merging multiple fraction (consensusxml; collect->FileMerger-> MzTabExporter), I get the foloiwng error message:  "Error: Unexpected internal error (PSM controllerType=0 controllerNumber=1 scan=18249 does not map to an MS file registered in the quantitative metadata. Check your merging and filtering steps and/or report the issue, please.)" 

I think that this problem has been previously reported - has there been any progress?

Thanks

jpfeuffer:
Fixes #5257

### DIFF
--- a/src/topp/FileMerger.cpp
+++ b/src/topp/FileMerger.cpp
@@ -77,7 +77,9 @@ using namespace std;
   </table>
   </center>
 
-  Special attention should be given to the append_method for consensusXMLs (see parameter description).
+  Special attention should be given to the append_method for consensusXMLs. One column corresponds to one channel/label + raw file. Rows are quantified and linked features.
+  More details on the use cases can be found at the parameter description.
+  
   For non-consensusXML or consensusXML merging with append_rows, the meta information that is valid for the whole experiment (e.g. MS instrument and sample)
   is taken from the first file only.
 
--- a/src/topp/FileMerger.cpp
+++ b/src/topp/FileMerger.cpp
@@ -61,7 +61,7 @@ using namespace std;
 /**
   @page TOPP_FileMerger FileMerger
 
-  @brief Merges several files. Multiple output format supported, depending on input format.
+  @brief Merges several files. Multiple output formats supported, depending on the input format.
 
   <center>
   <table>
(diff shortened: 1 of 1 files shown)


timosachsenberg (approved):

src/topp/FileMerger.cpp
```diff
@@ -123,7 +124,10 @@ class TOPPFileMerger :
     setValidFormats_("out", ListUtils::create<String>("mzML,featureXML,consensusXML,traML,fasta"));
 
     registerFlag_("annotate_file_origin", "Store the original filename in each feature using meta value \"file_origin\" (for featureXML and consensusXML only).");
-    registerStringOption_("append_method", "<choice>", "append_rows", "Append consensusMaps rowise or colwise. (Please use colwise for the MSstatsConverter)", false);
+    registerStringOption_("append_method", "<choice>", "append_rows", "(ConsensusXML-only) Append quantitative information about features row-wise or column-wise.\n"
```
cbielow, line 127: the concept of rows and columns it a bit confusing here, because a user has probably no relation to it.
append_rows basically means treat the underlying featuremaps independently and simply append them?
Whereas append_rows basically means: merge the i
th featuremap from all maps.
alternatively: Maybe its enough to say that columns are featuremaps, and rows are input consensusmaps?

src/topp/FileMerger.cpp
```diff
@@ -71,16 +71,17 @@ using namespace std;
...
   </center>
 
-  The meta information that is valid for the whole experiment (e.g. MS instrument and sample)
-  is taken from the first file.
+  Special attention should be given to the append_method for consensusXMLs (see parameter description).
+  For non-consensusXML or consensusXML merging with append_rows, the meta information that is valid for the whole experiment (e.g. MS instrument and sample)
+  is taken from the first file only.
 
-  The retention times for the individual scans are taken from either:
+  For spectrum containing formats (no feature/consensusXML), the retention times for the individual scans are taken from either:
```
cbielow, line 84: `spectrum-containing`

src/topp/FileMerger.cpp
jpfeuffer: Hmm not sure if that is correct. I think both are appending to make one consensusMap. But row-wise will keep the map_index numbers (e.g. 1,2,3 and 1,2,3) and use the file information from the first consensusMap for the new extended map.  Col-wise will update the map_index numbers to (1,2,3,4,5,6) and add the corresponding file information to the ConsensusXML header.

jpfeuffer: @timosachsenberg implemented that mainly and might be able to comment more.

timosachsenberg:
Yes Julianus is right. One column corresponds to one channel/label + raw file. rows are quantified and linked features.

cbielow:
> 
> 
> Yes Julianus is right. One column corresponds to one channel/label + raw file. rows are quantified and linked features.

Ah ok. But this small sentence already hepls tremdendiously and is IMHO worth adding.

cbielow (approved):

PR #5264, opened 2021-04-03
[FEATURE] allow auto-deduction of special rules in PepIdx for ConsID

jpfeuffer:
# Description

Needed for consensusID where old search engines are annotated in the metavalues.
Also, in the case of many runs, you should use the loosest settings so the special rules are now activated as soon as one run is a "special" SE.

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/PeptideIndexing.h
@@ -249,23 +249,23 @@ public:
         enzyme.setEnzyme("Trypsin");
       } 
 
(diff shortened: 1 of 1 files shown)


cbielow:
looks good.
I'd be happy if the search engine terms (incl. the metavalues written by consensusID) were a bit more standardized, and maybe wrapped in an internal class instead of metavalues.... but that's probably a different can of worms.

cbielow (approved):

jpfeuffer:
They are standardized in a way that I do not know of any other way tools save this secondary information. I re-used how PercolatorAdapter did it. Most of that should be solved with the new ID data structure which has the processing steps.
But yes some global enums for possible processing steps and scores and score types would be nice.

PR #5267, opened 2021-04-08
[DOC] description on how to build a windows installer

cbielow:
...


doc/doxygen/install/install-win.doxygen
```diff
@@ -269,39 +269,73 @@
...
+      <li> Checkout the <a href="https://github.com/OpenMS/THIRDPARTY" target="_blank">THIRDPARTY GitHub Repository</a> and flatten the structure for the target platform, e.g. copy all subdirectories in <tt>./all/</tt> to <tt>./Windows/64bit/</tt>.
+
+      <li> Configure %OpenMS as usual with additional flags <tt>SEARCH_ENGINES_DIRECTORY</tt> (for the flattened THIRDPARTY directory) and <tt>PACKAGE_TYPE="nsis"</tt>, 
+      e.g.
+      \code
+      cd <build-tree>
+      cmake ... -DSEARCH_ENGINES_DIRECTORY=C:\dev\THIRDPARTY\Windows\64bit -DPACKAGE_TYPE="nsis" <path-to-source-tree>
+      \endcode
+      
+      <li> Build all targets (incl. 'doc' and 'doc_tutorials') in Release mode (copy the Qt5 plugin to the <code>[OpenMS_build]\bin\Release</code> and <code>[OpenMS_build]/doc/doxygen/parameters/Release/</code> directories first)
```
jpfeuffer, line 328: Would be cool if we could get rid of this at some point. Should be relatively easy to do with CMake right?
I.e. as a pre-build step for doc or doc_tutorials to copy the files. Or just call findQt there again. Or pass the location down to this included script.

jpfeuffer:
Looks good.

jpfeuffer (approved):

doc/doxygen/install/install-win.doxygen
jpfeuffer: Ah no, sorry, the problem is the tools to find the dependency at runtime before being packaged.
Hm for this only the auto-copy option or adding the Qt folder to the path via CMake are the remaining options.

PR #5280, opened 2021-04-13
NucleicAcidSearchEngine: support for RNase T1 cleavage with variable inosine mod.

hendrikweisser:
This is something I implemented over a year ago (time flies during the pandemic...), but forgot to merge.

RNase T1 is probably the most used digestion enzyme for oligonucleotide MS. It cuts after (unmethylated) G, but also after I (inosine). Inosine can be installed in RNA by A-to-I "editor" enzymes (e.g. ADAR).

When RNase T1 cleavage and "I" as a variable modification are combined in NucleicAcidSearchEngine (NASE), the problem is that the mod. introduces additional cleavage sites. In the implementation, however, digestion is performed before variable mods. are applied, so this adds a complicating constraint.
In this change set:
- the special case (RNase T1 + var. I) is detected in NASE,
- necessity for applying mods. is tracked during digestion (see new data structure `RNaseDigestion::DigestionFragment`),
- missed cleavages are tracked during application of mods. (in `ModifiedNASequenceGenerator`)

This unfortunately complicates the code. However, I have simplified the implementation of `ModifiedNASequenceGenerator` in the process, so overall I'd say it about evens out. (And the RNase T1 + inosine interaction is now handled correctly, of couse.)


src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
```diff
@@ -58,30 +60,55 @@ namespace OpenMS
...
-    /// Fast implementation of modification placement. No combinatorial placement is needed in this case
-    /// - just every site is modified once by each compatible modification.
-    /// Already modified residues are skipped
-    static void applyAtMostOneVariableModification_(
-      const std::set<ConstRibonucleotidePtr>& var_mods,
-      const NASequence& seq,
-      std::vector<NASequence>& all_modified_NASequences,
-      bool keep_original = true);
+      /// C'tor for convenience
+      ModSeqInfo(NASequence seq, Size var_mods_left,
```
timosachsenberg, line 90: do we need this? I think with newer c++ we already can initialize the struct with
ModSeqInfo m{NASequence("AATT"), 4, -1}

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
```diff
@@ -58,30 +60,55 @@ namespace OpenMS
...
   protected:
-    /// Recursively generate all combinatorial placements at compatible sites
-    static void recurseAndGenerateVariableModifiedSequences_(
-      const std::vector<int>& subset_indices,
-      const std::map<int, std::vector<ConstRibonucleotidePtr>>& map_compatibility,
-      int depth,
-      const NASequence& current_NASequence,
-      std::vector<NASequence>& modified_NASequences);
+    /// Sequence with modification information
+    struct ModSeqInfo
```
timosachsenberg, line 83: add _ suffix to indicate protected type?

src/openms/include/OpenMS/CHEMISTRY/RNaseDigestion.h
```diff
@@ -52,12 +52,24 @@ namespace OpenMS
...
+    /// Destructor
+    virtual ~RNaseDigestion() = default;
+
     /// Sets the enzyme for the digestion
     void setEnzyme(const DigestionEnzyme* enzyme) override;
 
     /// Sets the enzyme for the digestion (by name)
     void setEnzyme(const String& name);
 
+    /// Sets how many As may be modified to I
```
timosachsenberg, line 67: maybe extend documentation here with examples of enzymes / why this has been added

src/openms/source/CHEMISTRY/ModifiedNASequenceGenerator.cpp
```diff
@@ -90,209 +87,133 @@ namespace OpenMS
...
-    bool keep_unmodified)
+
+  void ModifiedNASequenceGenerator::addModToSequences_(
+    vector<ModSeqInfo>& temp_seqs, Size n_temp_seqs,
+    vector<NASequence>& finished_seqs,
+    const function<bool(NASequence&, Int&)>& applyMod)
   {
-    // no variable modifications specified or no variable mods allowed? no compatibility map needs to be build
-    if (var_mods.empty() || max_variable_mods_per_seq == 0)
+    for (Size i = 0; i < n_temp_seqs; ++i)
```
timosachsenberg, line 96: can you extend the documentation of this function?

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -100,31 +144,80 @@ namespace OpenMS
     }
     else // proper enzyme cleavage
     {
-      vector<Size> fragment_pos(1, 0);
+      // with variable A->I, would enzyme cut after/before I, but not unmod. A?
+      bool cuts_after_I = (variable_inosine_ &&
+                           boost::regex_search("I", cuts_after_regex_) &&
```
timosachsenberg, line 149: can we go without boost regex and use std regex?

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
```diff
@@ -100,31 +144,80 @@ namespace OpenMS
...
+          continue;
+        }
+        if (cuts_after_special && cuts_before_special)
+        {
+          fragment_pos.emplace_back(i, true, true);
+          continue;
         }
       }
-      fragment_pos.push_back(rna.size());
+      fragment_pos.emplace_back(rna.size(), false, false);
```
timosachsenberg, line 192: can you add a inline comment what the false, false stand for?

src/tests/class_tests/openms/source/ModifiedNASequenceGenerator_test.cpp
```diff
@@ -111,16 +111,39 @@ START_SECTION(static void applyVariableModifications(const std::set<ConstRibonuc
...
+  // two missed cleavage allowed:
+  ams.clear();
+  ModifiedNASequenceGenerator::applyVariableModifications(var_mods, sequence, 2, ams, false, 2);
+  TEST_EQUAL(ams.size(), 4 + 6);
+  // with a separate mod:
+  var_mods.insert(db->getRibonucleotide("Am"));
+  ams.clear();
+  ModifiedNASequenceGenerator::applyVariableModifications(var_mods, sequence, 2, ams, false, 1);
+  // sum below corresponds to: 1 Am, 2 Am, 1 I, 1 Am + 1 I:
+  TEST_EQUAL(ams.size(), 4 + 6 + 4 + 4 * 3);
```
timosachsenberg, line 143: can you also test the string representation?

src/tests/class_tests/openms/source/RNaseDigestion_test.cpp
```diff
@@ -184,6 +187,53 @@ START_SECTION((void digest(IdentificationData& id_data, Size min_length = 0,
...
+  rd.digest(id_data);
+
+  TEST_EQUAL(id_data.getIdentifiedOligos().size(), 7);
+
+  rd.setMissedCleavages(1);
+  rd.digest(id_data);
+
+  TEST_EQUAL(id_data.getIdentifiedOligos().size(), 11);
+
+  // get oligos ordered by occurrence in the RNA:
```
timosachsenberg, line 221: what about this comment? should this go into a Tutorial_ ?

timosachsenberg (approved):

poshul (commented):
I think this is a good improvement in functionality. I am curious whether we should address the broader issue of variable mods that change cleavage sites, but I think for now this covers the most common case.

Is overall NASE performance affected?

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
timosachsenberg: might have some speed impact though

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
timosachsenberg: no strong opinion though

hendrikweisser:
> I am curious whether we should address the broader issue of variable mods that change cleavage sites

I haven't gone to those lengths because we didn't have a compelling use case. There are numerous modifications that would prevent cleavage at a site (e.g. there are exceptions for methylated residues in several of the cleavage rules in our enzymes XML file), and I'm not sure how well we handle those. (I don't think we "refund" missed cleavages on sites that get subsequently modified in a way that would prevent cleavage.)
A-to-I with RNase T1 is the only case I know where a mod. adds a cleavage site.

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
hendrikweisser: Do we do that for structs/classes? It looks a bit weird.

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
hendrikweisser: Good to know for the future, but I'd leave it for the default value for `missed_cleavages_left`.

src/openms/include/OpenMS/CHEMISTRY/RNaseDigestion.h
hendrikweisser: Added: "This is relevant for RNase T1 cleavage when variable inosine modifications are considered."

src/openms/source/CHEMISTRY/ModifiedNASequenceGenerator.cpp
hendrikweisser: I've added a comment. There's already documentation in the header file.

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
hendrikweisser: We use `boost::regex` consistently in the "digestion" code, so I'd rather not change it just here.
If you want to switch to `std::regex`, I think that should be a separate pull request.

src/openms/source/CHEMISTRY/RNaseDigestion.cpp
hendrikweisser: Done.

src/openms/include/OpenMS/CHEMISTRY/RNaseDigestion.h
```diff
@@ -76,18 +88,32 @@ namespace OpenMS
...
+        req_mod_first(req_mod_first), req_mod_last(req_mod_last)
+      {
+      }
+    };
+
     const Ribonucleotide* five_prime_gain_; ///< 5' mod added by the enzyme
     const Ribonucleotide* three_prime_gain_; ///< 3' mod added by the enzyme
     boost::regex cuts_after_regex_; ///< reg. exp. for enzyme cutting pattern
     boost::regex cuts_before_regex_; ///< reg. exp. for enzyme cutting pattern
+    Size variable_inosine_; ///< max. number of As that may be modified to I
```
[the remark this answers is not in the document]
hendrikweisser: Possibly. But for now this is the only case, so I've used specific variable names.

src/tests/class_tests/openms/source/ModifiedNASequenceGenerator_test.cpp
hendrikweisser: Okay, but just for two examples.

src/tests/class_tests/openms/source/RNaseDigestion_test.cpp
hendrikweisser: I've used this for debugging and left it in there for when things break again in the future. ;-)
I've updated the comment accordingly.

src/openms/include/OpenMS/CHEMISTRY/ModifiedNASequenceGenerator.h
timosachsenberg: yep

hendrikweisser:
I've added the underscore to `ModSeqInfo_` as per Timo's request.

Good to merge?

poshul:
Looks like there is still a test failing?

hendrikweisser:
Hopefully the failing test is now fixed. (Looks like it was a system-specific difference.)

@timosachsenberg: Is the CI running Ubunut 18.04 but not 20.04? Would it be worth upgrading?

poshul:
rebuild jenkins

timosachsenberg (changes requested):
you proably also need to update the pyOpenMS bindings

poshul:
@hendrikweisser can you take a look at the python tests when you get a chance?

jpfeuffer:
What is the status of this PR?

hendrikweisser:
@jpfeuffer: Needs more work because other changes to RNA digestion were merged in the meantime, and pyOpenMS. Not a priority for me at the moment.

PR #5282, opened 2021-04-14
Feature/mzqc export

axelwalter:
# mzQC export for QCCalculator

- fix mzQC file validation
- fix other reviewer comments
- update tests


axelwalter:
Used the existing TIC QC class for the mzQC export and added a SpectrumCount class. TIC::compute() returns a struct now, therefore I added a return Value for it in QualityControl and TIC_test. However, maybe there is a more elegant solution, like an option to return the struct or void?

timosachsenberg:
contrib submodule also needs to be up to date

src/openms/include/OpenMS/QC/SpectrumCount.h
```diff
@@ -0,0 +1,74 @@
...
+//
+// --------------------------------------------------------------------------
+// $Maintainer: Axel Walter $
+// $Authors: Axel Walter $
+// --------------------------------------------------------------------------
+
+#pragma once
+
+#include <OpenMS/QC/QCBase.h>
+#include <OpenMS/KERNEL/MSExperiment.h>
```
cbielow, line 38: no needed

src/openms/include/OpenMS/QC/SpectrumCount.h
```diff
@@ -0,0 +1,74 @@
...
+  public:
+    /// Constructor
+    SpectrumCount() = default;
+
+    /// Destructor
+    virtual ~SpectrumCount() = default;
+
+    /**
+    @brief Compute number of spectra per MS level and returns them in a map
+
```
cbielow, line 60: so the mapping is `ms_level --> count`?

src/openms/include/OpenMS/QC/TIC.h
```diff
@@ -71,7 +71,16 @@ namespace OpenMS
...
+    struct Result
+    {
+      std::vector<UInt> intensities;
+      std::vector<float> retention_times;
+      UInt area = 0;
+      UInt fall = 0;
+      UInt jump = 0;
+    };
+
+    Result compute(const MSExperiment& exp, float bin_size=0);
```
cbielow, line 83: since result is returned here,
`std::vector<MSChromatogram> results_;` is not needed anymore?

cbielow: mhh... wait, its probably needed?
now its a bit confusing what the other `results_` is... maybe document that a little...

src/openms/include/OpenMS/QC/sources.cmake
```diff
@@ -16,6 +16,7 @@ set(sources_list_h
   RTAlignment.h
   DBSuitability.h
   TIC.h
+  SpectrumCount.h
```
cbielow, line 19: please sort by name

src/openms/source/QC/TIC.cpp
```diff
@@ -46,9 +46,36 @@ namespace OpenMS
...
-  void TIC::compute(const MSExperiment &exp, float bin_size)
+
+  TIC::Result TIC::compute(const MSExperiment& exp, float bin_size)
   {
     results_.push_back(exp.getTIC(bin_size));
+    struct Result result;
+    const MSChromatogram& tic = exp.getTIC();
+    if (!tic.empty())
+    {
+    for (const auto& p : tic)
```
cbielow, line 57: indentation

src/openms/source/QC/TIC.cpp
```diff
@@ -46,9 +46,36 @@ namespace OpenMS
...
+    {
+      result.intensities.push_back(p.getIntensity());
+      result.retention_times.push_back(p.getRT());
+    }
+
+    result.area = result.intensities[0];
+
+    for (UInt i = 1; i < result.intensities.size(); ++i)
```
cbielow, line 65: use `size_t` for comparison to `size()`, since UInt != size_t on some platforms (e.g. Windows)

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -153,92 +101,140 @@ namespace OpenMS
...
+      addMetric("QC:4000060", counts[2]);
+    }
+    // Number of chromatograms"
+    addMetric("QC:4000135", exp.getChromatograms().size());
+    // Run time (RT duration)
+    addMetric("QC:4000053", UInt(exp.getMaxRT() - exp.getMinRT()));
+    // MZ acquisition range
+    addMetric("QC:4000138", tuple<UInt,UInt>{exp.getMinMZ(), exp.getMaxMZ()});
+
+    if (tic.isRunnable(status) and !exp.getTIC().empty())
```
cbielow, line 130: `exp.getTIC()` is essentially computed twice, right?
Why not just call `tic.compute()` and see if result is empty?
Or, change the `compute()` to accept an externally computed TIC...

src/tests/topp/QCCalculator_2_output.mzQC
```diff
@@ -1,112 +1,113 @@
 {
   "mzQC": {
-    "creationDate": "2021-03-29T10:09:41",
+    "creationDate": "2021-04-14T12:11:57",
     "version": "1.0.0",
-    "contactName": "name",
-    "contactAddress": "address",
+    "contact_name": "name",
```
cbielow, line 5: should all be `camelCase`

src/utils/QCCalculator.cpp
```diff
@@ -144,11 +147,23 @@ class TOPPQCCalculator :
...
-    String contactName = getStringOption_("name");
-    String contactAddress = getStringOption_("address");
+    String contact_name = getStringOption_("name");
+    String contact_address = getStringOption_("address");
     String description = getStringOption_("description");
     String label = getStringOption_("label");
     bool remove_duplicate_features(getFlag_("remove_duplicate_features"));
+    
+    // ensure output file hase valid extension
+    FileTypes::Type out_type = FileTypes::nameToType(getStringOption_("out_type"));
```
cbielow, line 157: try the new `FileHandler::getConsistentOutputfileType()` instead.

src/utils/QCCalculator.cpp
```diff
@@ -144,11 +147,23 @@ class TOPPQCCalculator :
...
+    if (out_type == FileTypes::UNKNOWN)
+    {
+      FileHandler fh;
+      out_type = fh.getTypeByFileName(outputfile_name);
+    }
+    if (out_type != FileTypes::QCML and out_type != FileTypes::MZQC)
```
cbielow, line 163: this should not be needed, since the restriction is already enforced by TOPPBase
https://github.com/OpenMS/OpenMS/pull/5282/files#diff-4d40a77e3e4025cbf9759cb22f4559b8ca1441f0f02c6bd347f09dad9e342eb3R128

src/openms/include/OpenMS/QC/TIC.h
axelwalter: the Result struct does not contain the MSChromatogram returned by getTIC. Instead RTs and intensities are stores separately in vectors. Add std::vector<MSChromatogram> to the Result struct and get rid of results_? What happens with TIC::getResults()?

src/openms/include/OpenMS/QC/SpectrumCount.h
axelwalter: yes, result[ms_level] = count

src/openms/include/OpenMS/QC/SpectrumCount.h
axelwalter: moved #include <OpenMS/KERNEL/MSExperiment.h> to SpectrumCount.cpp and removed #include <map>

src/openms/include/OpenMS/QC/TIC.h
axelwalter: removed the results_ vector and getResults function. The QualityControl tool now collects TIC::Result structures in a vector, which is used by TIC::addMetaDataMetricsToMzTab.
Is there a case where multiple TICs are added to MzTab at once?

src/openms/source/FORMAT/MzQCFile.cpp
axelwalter: true, it is computed twice, removed that and checked if the result contains a vector of intensities.

src/openms/include/OpenMS/QC/TIC.h
```diff
@@ -71,7 +70,16 @@ namespace OpenMS
     @param bin_size RT bin size in seconds
     @return TIC Chromatogram
     **/
-    void compute(const MSExperiment &exp, float bin_size=0);
+    struct Result
```
cbielow, line 73: the struct is now annotated with the docs of `compute()`.
Intended? :)

src/openms/include/OpenMS/QC/TIC.h
```diff
@@ -71,7 +70,16 @@ namespace OpenMS
...
+    {
+      std::vector<UInt> intensities;
+      std::vector<float> retention_times;
+      UInt area = 0;
```
cbielow, line 77: can you document what these members indicate?

src/tests/class_tests/openms/source/TIC_test.cpp
```diff
@@ -73,25 +73,13 @@ START_SECTION(void compute(const MSExperiment &exp, float bin_size))
...
-  auto r = tic.getResults();
-  TEST_EQUAL(r.size(), 1);
-  ABORT_IF(r[0].size() != 1); // one intensity per input spectrum
-  ABORT_IF(r[0][0].getIntensity() != 0); // empty spectrum
-END_SECTION
-
-START_SECTION(vector<MSChromatogram> getResults() const)
-  NOT_TESTABLE // tested above
+  auto r = tic.compute(exp, 0);
+  ABORT_IF(!r.intensities.empty()); // empty spectrum
```
cbielow, line 77: no need to abort if you do not test anything else.
Simply TEST_EQUAL(...)

src/tests/class_tests/openms/source/TIC_test.cpp
```diff
@@ -73,25 +73,13 @@ START_SECTION(void compute(const MSExperiment &exp, float bin_size))
...
 END_SECTION
 
 START_SECTION(void clear())
```
cbielow, line 80: clear() was removed, right? so just delete the sub-test

src/utils/QCCalculator.cpp
```diff
@@ -144,11 +147,21 @@ class TOPPQCCalculator :
...
+    String contact_address = getStringOption_("address");
     String description = getStringOption_("description");
     String label = getStringOption_("label");
     bool remove_duplicate_features(getFlag_("remove_duplicate_features"));
+    
+    // ensure output file hase valid extension
+    FileTypes::Type out_type = FileHandler::getConsistentOutputfileType(outputfile_name, getStringOption_("out_type"));
+
+    // exit early if invalid file type
+    if (out_type != FileTypes::QCML or out_type != FileTypes::MZQC)
```
cbielow, line 160: not needed. checked by TOPPBASE

src/openms/include/OpenMS/QC/TIC.h
```diff
@@ -80,10 +88,10 @@ namespace OpenMS
     QCBase::Status requires() const override;
 
     /// append QC data for given metrics to mzTab's MTD section
-    void addMetaDataMetricsToMzTab(MzTabMetaData& meta);
+    void addMetaDataMetricsToMzTab(MzTabMetaData& meta, std::vector<Result>& tics);
 
   private:
     const String name_ = "TIC";
-    std::vector<MSChromatogram> results_;
+    Result result;
```
cbielow, line 95: not used anymore, right?

src/openms/include/OpenMS/QC/TIC.h
axelwalter: don't think so ;)

src/openms/include/OpenMS/QC/TIC.h
axelwalter: added comments in code and updated description for compute function

src/utils/QCCalculator.cpp
axelwalter: this covers the case when the user doesn't specify output_type and the output file name has a non-valid extension. In that case I don't want to read the mzML file and exit early. but this should not happen often, so it can be removed.

cbielow: in that case, `getConsistentOutputfileType` will throw an exception for you

PR #5287, opened 2021-04-16
Mixed stuff

cbielow:
# Description

* some better docs
* support for convenient output type determination in TOPP tools with less code
* example TIMS TOF file
* progress logging for FASTA file reading/writing


timosachsenberg (approved):

timosachsenberg:
hmm do we need support for pre 2019 compiler? what about V17 support in those?
@cbielow  @jpfeuffer

jpfeuffer:
I would assume VS2017 fails already with other stuff after migration to C17? Are we building with 2017 somewhere still?

timosachsenberg:
yep https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/bldtst/compiler=msvc-14.2,os_label=win10_64/972/testReport/projectroot.src.tests.class_tests/openms/MSExperiment_test/

jpfeuffer:
but 14.2 = 2019

timosachsenberg:
Visual Studio 2019 | Dev16 | 16.0
-- | -- | --
Visual Studio 2017 | Dev15 | 15.0
Visual Studio 2015 | Dev14 | 14.0

jpfeuffer:
We usually number it according to the underlying MSVCompiler version: see the table here https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B

jpfeuffer:
I think it is better for compatibility reasons (e.g. 14.x is compatible across each other)

jpfeuffer:
But yeah there are many 2019 version these days. Apparently with changes in their std lib implementation. So @cbielow seems to just have a newer 2019 version.

timosachsenberg:
hmm but even then is 14.2 still Visual Studio 2017 15.x according to the table? sorry don't get it

jpfeuffer:
Bildschirmfoto 2021-04-19 um 14 26 05
This is the table I am referring to.

cbielow:
so it seems my VS2019 supports the noexcept, the one for CI does not (even though its VS2019).

timosachsenberg:
aaah so 14.2 is > 14.16
interesting

jpfeuffer:
** Visual Studio 2019 Developer Command Prompt v16.8.4
Pretty recent. @cbielow Do you have 16.9?

jpfeuffer:
Maybe this? https://docs.microsoft.com/de-de/cpp/overview/cpp-conformance-improvements?view=msvc-160#explicit-noexcept-specifiers-on-destructor-templates

PR #5305, opened 2021-04-30
[pyOpenMS] convert list [str, ... ] to StringList

timosachsenberg:
# Description
Adapted the special_conversion_providers for autowrap.

now StringLists can be passed in python with both bytes and str
[b"string1", b"string2"] and ["string1", "string2"]

More similar to how user could write could before switch to python 3

# Checklist:


jpfeuffer:
Can you add a test for nonASCII strings?

timosachsenberg:
Is there an easy way to also support that for std::vector<std::string> ?
Our param object now uses stl only objects so we can't use a StringList here

jpfeuffer:
Why would you need this? You cannot construct a StringList anyway? Everything will be passed as python lists, no?

timosachsenberg:
Basically I want to use e.g.
param.setTags("key", ["s1", "s2"])
instead of
param.setTags("key", [b"s1", b"s2"])

PR #5306, opened 2021-04-30
[FEATURE] New Deisotoping Algorithm

emilpaulitz:
# Description

I added the function Deisotoper::deisotopeWithAveragineModel. It builds upon the already existing function using parts of Guo Ci Teo et al, DOI: 10.1021/acs.jproteome.0c00544. It considers more combinations of peaks as possible isotopic patterns and applies a more sophisticated model to verify the feasibility of found clusters than the already existing algorithm. Also, it removes low intensity peaks. I did some benchmarking using the B1.mzML dataset from https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Tutorials/Data/iPRG2016/ (iPRG2016 data from PRIDE entry PXD008425) to compare the new with the existing algorithm. Applying OpenMS's SimpleSearchEngine to the edited spectra, using the iPRG2016_shuff.fasta as protein database, and subsequent FDR estimation and filtering yielded the following results:

When configuring both functions to keep peaks that could not be assigned to an isotopic cluster (keep_only_deisotoped = false), the results were very similar and a bit worse for the new algorithm:

**old algorithm:** 18809 peptides found overall, but:
    match to target DB only: 11023 (58 %)
    match to decoy DB only : 7786 (41 %)
**After 1% FDR filtering:**
    1 identification runs with 1034 proteins,
    2950 spectra identified with 2950 spectrum matches.

**new algorithm:** 18766 peptides found overall,
    match to target DB only: 10962 (58 %)
    match to decoy DB only : 7804 (41 %) 
**After 1% FDR filtering:**
    1 identification runs with 1022 proteins,
    2935 spectra identified with 2935 spectrum matches.

When using keep_only_deisotoped = true, the deciding factor is min_isopeaks, as the following figure shows. If this is the same for both algorithms, the new algorithm unfortunately also performs slightly worse.

results

When disabeling preprocessing of spectra in the new algorithm, the pipeline could detect 2207 instead of 2206 peptides in one setting.

Time needed by the search engine was only really different in the upper setting, where the file generated by the new algorithm was quite a bit smaller because it removes low- and zero-intensity peaks, and subsequently the peptide search took a couple seconds less. 

An example: https://drive.google.com/drive/folders/1JAgFFOojfNAELmVlfql0Dma_yrHn2Rh3?usp=sharing

Both algorithms were applied to the Deisotoper_test_in.mzML file, with settings min_isopeaks = 3, keep_only_deisotoped = true and, for the new algorithm, rem_low_intensity = true. The result is in the files new_algo.mzML and old_algo.mzML respectively.
The old algorithm returned 446 peaks, and 357 of these had intensity 0. The new one only reported back 9 base peaks. Two of those (at m/z 407.97 and 358.98) were not reported back by the old algorithm, likely because another peak was identified as the base peak (e.g. 406.964 and 356.999), which the new algorithm rejected during its averagine check. 7 peaks were identified as base peaks by both algorithms.

# Checklist:
- [] Add relevant changes and new features to the CHANGELOG file


src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,292 @@
...
+    throw Exception::IllegalArgument(__FILE__,
+      __LINE__,
+      OPENMS_PRETTY_FUNCTION,
+      "Minimum/maximum number of isotopic peaks must be at least 2 (and min_isopeaks <= max_isopeaks).");
+  }
+
+  if (spec.empty()) { return; }
+
+  // discard low-intensity peaks
+  if (rem_low_intensity)
```
timosachsenberg, line 75: see https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp#L227
this could shorten the code

timosachsenberg: and maybe remove 0 intensiteis first with: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/ID/SimpleSearchEngineAlgorithm.cpp#L192

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,292 @@
...
+        {
+          // compare to averagine distribution
+          if (use_averagine_model)
+          {
+            // compute average weight (weighted by intensity)
+            total_weight += old_spectrum[p].getIntensity() * (old_spectrum[p].getMZ() * q - (q - 1) * Constants::PROTON_MASS_U);
+            total_intensity += old_spectrum[p].getIntensity();
+
+            // generate averagine distribution
+            CoarseIsotopePatternGenerator gen(extensions.size() + 1);
```
timosachsenberg, line 201: why only up to extensions.size() + 1?
do probabilities need to be normalized or not?

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,292 @@
...
+            IsotopeDistribution distr = gen.estimateFromPeptideWeight(total_weight / total_intensity);
+            distr.sortByMass();// necessary?
```
timosachsenberg, line 203: should not be necessary

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,292 @@
...
+              {
+                KL += Px * log(Px / distr[extensions.size()].getIntensity());
+              }
+
+              // choose threshold corresponding to cluster size
+              curr_threshold = (extensions.size() + 1 > 6) ? averagine_check_threshold[4] : averagine_check_threshold[extensions.size() - 1];
+            }
+
+
+            // compare to threshold and stop extension if distribution does not fit well enough
```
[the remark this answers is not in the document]
timosachsenberg: based on current KL

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,292 @@
...
+    has_precursor_data = true;
+    int precursor_charge = old_spectrum.getPrecursors()[0].getCharge();
+    precursor_mass = (old_spectrum.getPrecursors()[0].getMZ() * precursor_charge) - (Constants::PROTON_MASS * precursor_charge);
+  }
+
+  for (size_t current_peak = 0; current_peak != old_spectrum.size(); ++current_peak)
+  {
+    // only process peaks which are not already in a cluster. Would form clusters identical to the cluster they are assigned to,
+    // excluding its first peak(s), since peaks with lower mz are not considered for cluster-formation.
+    if (features[current_peak] != -1) continue;
```
timosachsenberg, line 145: note to myself: check in publication.
e.g., sometimes with deamidation it might make sense to take a smaller cluster with better KL

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: Yes, these classes are perfect. I think it is slightly less efficient now, but code is much shorter

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,292 @@
...
+
+    // Monoisotopic peaks with intensity 0.0 interfere with averagine check when normalizing the spectrum peaks height
+    if (use_averagine_model && old_spectrum[current_peak].getIntensity() == 0.0) continue;
```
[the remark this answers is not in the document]
emilpaulitz: If rem_low_intensity == false, ThresholdMower is not called and then the algorithm could break for two zero-peaks back to back because of divide by 0 in line 196. I added an "&& !rem_low_intensity" so the check for zero does not get called normally but I dont think we can dismiss this line completely, can we?

timosachsenberg:
TODO:
- evaluate with proper FDR filtering
- some comparison plots between old and new method

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: > why only up to extensions.size() + 1?
The peaks in the theoretical distribution without a counterpart in the cluster would all result in an addition to the KL of 0 since P(x) would be 0, so we do not need to generate those.

> do probabilities need to be normalized or not?
The cluster we want to test and the theoretical distribution should sum up to the same amount, or else the KL does not make sense, as it compares distributions. Does specification of max_isotope lead to an unnormalized distribution? In that case renormalization would be necessary

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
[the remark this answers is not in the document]
emilpaulitz: The paper states that any identified cluster has to pass the check for all subclusters, that is, without the n highest m/z peaks. Scoring the subclusters as we build up the full cluster should be the same thing

emilpaulitz:
> 
> 
> TODO:
> 
>     * evaluate with proper FDR filtering
> 
>     * some spectra comparison between old and new method

Did both and added to the original description. Any idea why the performance is worse? It could very well be that my pipeline is incorrect, or maybe the new algorithm is too strict and does not deisotope enough peaks?

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
timosachsenberg: I am not sure just guessing if something could go wrong here.
Wouldn't it make sense to generate the full isotope pattern for the current mass and then compare the distributions?
Right now it (at least seems to) calculates it for multiple extensions.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,285 @@
...
+    spec.getIntegerDataArrays().resize(spec.getIntegerDataArrays().size() + 1);
+    spec.getIntegerDataArrays().back().setName("iso_peak_count");
+    iso_peak_count_index = spec.getIntegerDataArrays().size() - 1;
+  }
+
+  // during discovery phase, work on a constant reference (just to make sure we do not modify spec)
+  const MSSpectrum& old_spectrum = spec;
+
+  // determine charge seeds and extend them
+  std::vector<size_t> mono_isotopic_peak(old_spectrum.size(), 0);
```
timosachsenberg, line 115: rename to mono_isotopic_peak_charge ?

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,285 @@
...
+
+      extensions.clear();
+      extensions.push_back(current_peak);
+
+      float total_weight = old_spectrum[current_peak].getIntensity() * (old_spectrum[current_peak].getMZ() * q - (q - 1) * Constants::PROTON_MASS_U);
+      float total_intensity = old_spectrum[current_peak].getIntensity();
+      for (unsigned int i = 1; i < max_isopeaks; ++i)
+      {
+        const double expected_mz = current_mz + static_cast<double>(i) * Constants::C13C12_MASSDIFF_U / static_cast<double>(q);
+        const int p = old_spectrum.findNearest(expected_mz, tolerance_dalton);
```
timosachsenberg, line 179: does the original publication only look for the nearest peak or consider all peaks in the tolerance window?

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,285 @@
...
+          has_min_isopeaks = (i >= min_isopeaks);
+          break;
+        }
+        else
+        {
+          // compare to averagine distribution
+          if (use_averagine_model)
+          {
+            // compute average weight (weighted by intensity)
+            total_weight += old_spectrum[p].getIntensity() * (old_spectrum[p].getMZ() * q - (q - 1) * Constants::PROTON_MASS_U);
```
timosachsenberg, line 191: I am not sure if I understand the weighting part.
The isotope pattern should reflect the averagine at the monoisotopic peak (imho).

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,292 @@
...
+    // if current_peak is possible monoisotopic peak for a cluster, pick the best of its clusters, annotate peaks with a feature number
+    if (!clusters.empty())
+    {
+      // pick cluster with largest size and highest charge (since all have the same monoisotopic peak)
+      unsigned int best_idx = 0;
+      Size largest_size = 0;
+      int highest_charge = min_charge;
+      for (unsigned int i = 0; i != clusters.size(); ++i)
+      {
+        if ((clusters[i].size() > largest_size) || ((clusters[i].size() == largest_size) && (charges_of_extensions[i] > highest_charge)))
```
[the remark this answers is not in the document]
timosachsenberg: seems legit :)

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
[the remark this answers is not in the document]
timosachsenberg: hmm probably depends a bit on what instrument vendors write out.
I would say we can call ThresholdMower in any case as these zero intensities are usually only artifacts.
Can you check if the performance goes down if you add the ThresholdMower irrespective of rem_low_intensity_ (e.g., at the very top)

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
timosachsenberg: Can you check the documentation and code of the generator? Maybe be need to extend it if it is not clear from the docs.
My concern is that if we only generate e.g. 1st+2nd and 1st+2nd+3rd isotopic peaks. These might contribute to e.g., 80% and 95% of the full isotope pattern. I could imagine that it might make a huge difference normalized vs. non-normalized.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
timosachsenberg: can you check that this is correct?

emilpaulitz: The paper does not explicitly say if peaks already in a cluster should be examined again. In my experience e.g. deamination peaks do not pass the averagine check anyway, as is tested as well:
https://github.com/OpenMS/OpenMS/blob/1b2aaa4d5e01376df3b9d002a4178c9764a9632c/src/tests/class_tests/openms/source/Deisotoper_test.cpp#L187

I do not believe that changing this would alter the output, as the averagine check should only be passed by the correct cluster, but it might be worth a try. Will give it a look.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: Yes

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: "A set of possible isotopic clusters is then obtained by checking the m/z distance between peaks, considering fragment charge states [...]

Does not really say. We could try to pick the peak in the tolerance window that achieves the best KL. Will give it a look as well.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: Yes this is probably a mistake, I was led by the average_weight parameter: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp#L119

But probably got that wrong. It would make much more sense to generate that pattern only once per monoisotopic peak, I will look into the code of the generator to figure out what average weight is needed to get the monoisotopic peak that we want, as you suggested.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: Yes that makes sense, will do

cbielow:
just a thought: if the algorithm requires a certain score when fitting the averagine model,  isn't this effectively ignoring the isotope peaks? I.e. the isotope pattern will stay in the spectrum if `keep_deisotoped_only = true`, thus adding "noise" peaks. Another (better?) option, would be to remove the entire pattern if it does not match the averagine (because its not a peptide).

timosachsenberg:
I think it might already be too conservative so removing a bad fit might make it worse

emilpaulitz:
I think this is already included, because if `keep_deisotoped_only = true`, **only** the monoisotopic peaks of confirmed patterns survive. So peaks in patterns that do not fit well enough to reach the threshold are removed per default.

In the `keep_deisotoped_only = false` setting, only isotopic peaks of confirmed clusters are removed, so I do not think it would be fitting to remove possibly bad fitting clusters...

If we find out the current algorithm is being too conservative, we could loosen the thesholds. But I am working on possible bug fixes, maybe these will resolve the discrepancy between these results and the papers.

emilpaulitz:
Alright, I looked into the three ideas I wrote down: 

1. I changed the generation of theoretical spetcra as discussed [here](https://github.com/OpenMS/OpenMS/pull/5306#discussion_r624060216) 
2. instead of using the peak as next extension returned by the findNearest function, calculate the KL of the cluster that would result from choosing every of the peaks in the tolerance window and keep the one with best KL. Now, that there were multiple non-zero peaks in the tolerance window only occured 1855 times in 37771 spectra (on average once every 20 spectra).
And these are the results:
3. the program also examines peaks that were previously already assigned to a cluster and resolves overlaps by KL divergence of the clusters 

results

The code for each experiment is in the branches of my fork: [spectra_generation](https://github.com/emilpaulitz/OpenMS/tree/spectra_generation), [best_nearest_peak](https://github.com/emilpaulitz/OpenMS/tree/best_nearest_peak), and [mult_clusters](https://github.com/emilpaulitz/OpenMS/tree/mult_clusters).

It seems like the spectra generation was actually an issue, but could not bridge the performance gap to the old algorithm. I used this mode of spectra generation in the best_nearest_peak experiment as well, as it was very compute intensive anyways. It thus seems like that change did not have an impact. Inspecting multiple clusters seems to have even worsened the results.

emilpaulitz:
Of Course! Should it be a new class, or integrated as a function into e.g. CoarseIsotopePatternGenerator? 

Currently, it only generates intensities (not peaks with m/z values) and it does not normalize the intensities. But I think both could be integrated as an option or overloaded functions.

Not 100% sure if I will have the time to implement that the next days though, and I am not available the next month, but I will give it a go.

timosachsenberg:
> Of Course! Should it be a new class, or integrated as a function into e.g. CoarseIsotopePatternGenerator?

Yeah maybe that is a good idea. Maybe as static function if possible? 

> Currently, it only generates intensities (not peaks with m/z values) and it does not normalize the intensities. But I think both could be integrated as an option or overloaded functions.

maybe as distinct static members (overloads are a bit more difficult to wrap in python) would be great

> Not 100% sure if I will have the time to implement that the next days though, and I am not available the next month, but I will give it a go.

No worries. Whenever that works for you

timosachsenberg (approved):
maybe also add python binding in Deisotoper.pxd

timosachsenberg:
ok is this good to merge in its current stage? apart from python bindings?

emilpaulitz:
From my side it's ready

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
```diff
@@ -37,6 +37,8 @@
 #include <OpenMS/KERNEL/StandardTypes.h>
 #include <OpenMS/CONCEPT/Types.h>
 #include <OpenMS/CONCEPT/Constants.h>
+#include <vector>
```
cbielow, line 40: please sort the includes lexicographically

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
```diff
@@ -47,6 +49,61 @@ class OPENMS_DLLAPI Deisotoper
...
+    This algorithm is in parts taken from Guo Ci Teo et al, DOI: 10.1021/acs.jproteome.0c00544
+    and is closely related to deisotopeAndSingleCharge by Timo Sachsenberg.
+
+    Deisotoping is based on C13 abundance and will try to identify isotopic
+    clusters fitting to an averagine model, taking in account the corresponding
+    charge state. This only makes sense for peptide fragment ion spectra.
+    The algorithm considers each peak in the spectrum and will try to form
+    isotopic clusters for every charge state from @p min_charge to @p max_charge.
+    Of the clusters that pass an averaginge check based on KL-divergence (for all 
+    subclusters starting at the base peak as well), the largest one with the
```
cbielow, line 63: at is `largest`? Number of peaks or total intensity explained?

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
```diff
@@ -47,6 +49,61 @@ class OPENMS_DLLAPI Deisotoper
...
+   * @param [keep_only_deisotoped] Only monoisotopic peaks of fragments with isotopic pattern are retained
+   * @param [min_isopeaks] The minimum number of isotopic peaks (at least 2) required for an isotopic cluster
+   * @param [max_isopeaks] The maximum number of isotopic peaks (at least 2) considered for an isotopic cluster
+   * @param [make_single_charged] Convert deisotoped monoisotopic peak to single charge, the original charge (>=1) gets annotated
+   * @param [annotate_charge] Annotate the charge to the peaks in the IntegerDataArray: "charge" (0 for unknown charge)
+   * @param [annotate_iso_peak_count] Annotate the number of isotopic peaks in a pattern for each monoisotopic peak in the IntegerDataArray: "iso_peak_count"
+   * @param [use_averagine_model] Compares observed patterns with theoretical averagine distributions using KL-divergence. If false, no intensity checks are applied.
+   * @param [add_up_intensity] Sum up the total intensity of each isotopic pattern into the intensity of the reported monoisotopic peak
+   * @param [rem_low_intensity] Only keep the 1000 or 5000 (see @p used_for_open_search ) most intense peaks in the spectrum.
```
cbielow, line 89: could you replace `rem_low_intensity` and `used_for_open_search` by an enum with
`NO_COUNT_FILTER`, `FILTER_TOP_1000` and `FILTER_TOP_5000`. This would make the parameter more readable and also reduce the number of parameters.
We could also simply use and `int number_of_final_peaks = 5000`.
(currently, the order of parameters in the docs is also different to the interface, but this problem would go away with the options proposed :) )

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
```diff
@@ -86,20 +143,35 @@ class OPENMS_DLLAPI Deisotoper
...
+  /*
+   * Return non-normalized approximation of the distribution, similar to Bellew et al, https://dx.doi.org/10.1093/bioinformatics/btl276
+   * As in most occasions, not all theoretical peaks are used in the calculation of KL divergence, normalization is done there anyway.
+   * This makes only the relative height of peaks in one distribution relevant, which is why the e-term is omitted as well.
+   * 
+   * This yields a 10-fold speed improvement over use of the OpenMS classes CoarseIsotopePatternGenerator + IsotopeDistribution. 
+   * 
+   * P(k) = (( (lambda*mass) ^ k) / k!) = (lambda^k / k!) * mass^k = precalculated[k] * mass^k 
+   */
+    static std::vector<MSSpectrum::PeakType::IntensityType> _approximateDistribution(MSSpectrum::PeakType::CoordinateType mass,
```
cbielow, line 172: a protected static method?
Suggestions: make this a free function in the cpp file. Then you can remove the declaration here and also forward-declare `MSSpectrum` which allows to remove the `#include` above.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,341 @@
 #include <OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h>
 #include <OpenMS/KERNEL/MSSpectrum.h>
 #include <OpenMS/MATH/MISC/MathFunctions.h>
+#include <OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h>
```
cbielow, line 41: please sort includes

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,341 @@
...
+  ThresholdMower threshold_mower_filter;
+  threshold_mower_filter.filterPeakSpectrum(spec);
+
+  // discard low-intensity peaks
+  if (rem_low_intensity)
+  { 
+    UInt max_num_peaks = used_for_open_search ? 1000 : 5000;
+
+    // only keep max_num_peaks highest peaks
+    NLargest nlargest_filter = NLargest(max_num_peaks);
```
cbielow, line 85: are you sure this is actually a good idea to filter the spectrum BEFORE finding isotope peaks? You might trim a isotope clusters.
Do you have some data on the effect of the filtering?

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,341 @@
...
+
+  // determine charge seeds and extend them
+  std::vector<size_t> mono_isotopic_peak_charge(old_spectrum.size(), 0);
+  std::vector<int> features(old_spectrum.size(), -1);
+  std::vector<double> mono_iso_peak_intensity(old_spectrum.size(), 0);
+  std::vector<Size> iso_peak_count(old_spectrum.size(), 1);
+  int feature_number = 0;
+
+  std::vector<size_t> extensions;
+  std::vector< std::vector<size_t> > clusters;
```
cbielow, line 122: this will be extremely inefficient in terms of runtime. If you have a maximum number of peaks inside the inner vector (max_isopeaks?) then flatten to vector<size_t> and manage the number of clusters using a counter.
May not dominate runtime though...

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,341 @@
...
+
+      bool has_min_isopeaks = true;
+
+      extensions.clear();
+      extensions.push_back(current_peak);
+
+      // Save frequently used values for performance reasons
+      std::vector<double> extensions_intensities = {current_intensity};
+      // This is essentially extensions.size() (which is needed a lot)
+      Int num_considered_peaks = 1;
```
cbielow, line 201: why not use `extension.size()` then?
The two variables are not always in sync below, which makes it a bit hard to see if the code is actually correct.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
```diff
@@ -38,9 +38,341 @@
...
+      {
+        spec[i].setMZ(spec[i].getMZ() * z - (z - 1) * Constants::PROTON_MASS_U);
+      }
+      select_idx.push_back(i);
+    }
+  }
+
+  // properly subsets all datapoints (incl. dataArrays)
+  spec.select(select_idx);
+  spec.sortByPosition();
```
cbielow, line 354: reasonably sure you dont need to sort, since input is sorted and `select` will not change order.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
timosachsenberg: I think it is described in the publication like that.

timosachsenberg:
@emilpaulitz do you have still time to revisit this PR with the suggestions from Chris?

emilpaulitz:
@timosachsenberg Yes definitely, but I will only be back in around 2-3 more weeks.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: if make_single_charged == true sorting should be necessary as the peaks' mz values change. I added an if clause

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
emilpaulitz: with respect to number of peaks. I changed the text 👍🏼

src/openms/include/OpenMS/FILTERING/DATAREDUCTION/Deisotoper.h
emilpaulitz: Yes, that makes it better. Decided for the int option and included the recommendation for 1000 in the docs.

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: Yes, this part, and also the numbers, are taken directly from the publication. I think most of my spectrums do not even have 5000 peaks but I will investigate with those that do in 2 weeks

src/openms/source/FILTERING/DATAREDUCTION/Deisotoper.cpp
emilpaulitz: Somehow thought this would improve runtime but probably does not as .size() is constant. I'll find a better solution

emilpaulitz:
This should be ready to merge now! 

For including the approximation into CoarsePatterngenerator I will open another PR.

PR #5308, opened 2021-05-03
Fasta Reader

noraw61:
# Description

Fasta Reader without seqan dependencies. According to our benchmarking with FileInfo the time loading the Fasta File is now about 2.08s (and before without our changes about 4.4s).

# Checklist:
- [x ] Make sure that you are listed in the AUTHORS file
- [x ] Add relevant changes and new features to the CHANGELOG file
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] New and existing unit tests pass locally with my changes
- [x ] Updated or added python bindings for changed or new classes. (Tick if no updates were necessary.)


src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
```diff
@@ -199,19 +200,19 @@ class FASTAContainer<TFI_File>
   }
 
   /// is the FASTA file empty?
-  bool empty() const
+  bool empty()
```
timosachsenberg, line 203: why do we loose the const here?

src/openms/include/OpenMS/FORMAT/FASTAFile.h
```diff
@@ -38,187 +38,180 @@
...
+        @brief Reads the next FASTA entry from file.
+        If you want to read all entries in one go, use load().
+        @return true if entry was read; false if eof was reached
+        @exception Exception::FileNotFound is thrown if the file does not exists.
+        @exception Exception::ParseError is thrown if the file does not suit to the standard.
+        */
+        bool readNext(FASTAEntry& protein);
+
+        /// current stream position
+        std::streampos position();
```
timosachsenberg, line 158: same here? why is position not const?

src/openms/include/OpenMS/FORMAT/FASTAFile.h
```diff
@@ -38,187 +38,180 @@
...
+
+        /// is stream at EOF?
+        bool atEnd();
```
timosachsenberg, line 161: and here

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -39,184 +39,247 @@
...
-    if (infile_.is_open()) infile_.close(); // precaution
+    bool FASTAFile::readEntry_(std::string& id, std::string& description, std::string& seq)
+    {
+        std::streambuf* sb = infile_.rdbuf();
+        bool keep_reading = true;
+        bool description_exists = true;
 
-    infile_.open(filename.c_str(), std::ios::binary | std::ios::in);
+        if (sb->sbumpc() != '>') return false; //was in wrong position for reading ID
+        while(keep_reading)// reading the ID
```
timosachsenberg, line 63: style polizei: space after while

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -39,184 +39,247 @@
...
+        {
+            int c = sb->sbumpc();// get and advance to next char
```
timosachsenberg, line 65: indentation = 2 spaces

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -39,184 +39,247 @@
...
+        {
+            return false;
+        }
+        seq_.clear();
+        id_.clear();
+        description_.clear();
+        if (!readEntry_(id_, description_, seq_))
+        {
+            if (entries_read_ == 0) seq_ = "The first entry could not be read!";
+            else seq_ = "Only " + String(entries_read_) + " proteins could be read. The record after failed.";
```
timosachsenberg, line 182:
```suggestion
            else seq_ = "Only " + String(entries_read_) + " proteins could be read. Parsing the next record failed.";
```

timosachsenberg (changes requested):
mainly style changes

timosachsenberg:
Thanks for the PR! Very useful.

cbielow (commented):
this I mentored this PR, I'm not a good reviewer ;)

src/openms/include/OpenMS/FORMAT/FASTAFile.h
```diff
@@ -38,187 +38,180 @@
...
+
+        /**
+        @brief Stores the data given by @p protein. Call writeStart() once before calling writeNext().
+        Call writeEnd() when done to close the file!
+        @exception Exception::UnableToCreateFile is thrown if the process is not able to write the file.
+        */
+        void writeNext(const FASTAEntry& protein);
+
+        /**
+        @brief Closes the file (flush). Called implicitly when FASTAFile object does out of scope.
```
jpfeuffer, line 180:
```suggestion
        @brief Closes the file (flush). Called implicitly when FASTAFile object goes out of scope.
```

timosachsenberg:
> this I mentored this PR, I'm not a good reviewer ;)

Still better than reviewing your own PR

src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
noraw61: because of the function atEnd() that is called there. The funcions peek(), tellg(), seekg() are not const :(

jpfeuffer (approved):
Only check the style a bit. Especially with comments at the end of the line. They often do not have a whitespace.

src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
timosachsenberg: without knowing the details... could the stream not be declared "mutable"?

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -39,179 +39,215 @@
...
-  {
+
+  FASTAFile::~FASTAFile() {
     // infile_ and outfile_ will close automatically when going out of scope. No need to do it explicitly here.
   }
 
-  void FASTAFile::readStart(const String& filename)
-  {
-    if (!File::exists(filename))
+  bool FASTAFile::readEntry_(std::string &id, std::string &description, std::string &seq) {
```
timosachsenberg, line 53: { in new line

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -39,179 +39,215 @@
...
+    bool description_exists = true;
+
+    if (sb->sbumpc() != '>') return false; // was in wrong position for reading ID
+    while (keep_reading) // reading the ID
+    {
+      int c = sb->sbumpc(); // get and advance to next char
+      switch (c) {
+        case ' ':
+        case '\t':
+          if (!id.empty()) {
```
timosachsenberg, line 65: here as well and othe rplaces

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -39,179 +39,215 @@
...
-    if (readRecord(id, s, *static_cast<FASTARecordReader*>(reader_.get()), seqan::Fasta()) != 0)
-    {
-      if (entries_read_ == 0) s = "The first entry could not be read!";
-      else s = "Only " + String(entries_read_) + " proteins could be read. The record after failed.";
-      throw Exception::ParseError(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "", "Error while parsing FASTA file! " + s + " Please check the file!");
+    seq_.clear();
+    id_.clear();
+    description_.clear();
+    if (!readEntry_(id_, description_, seq_)) {
+      if (entries_read_ == 0) seq_ = "The first entry could not be read!";
```
timosachsenberg, line 166: brackets and newlines here

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -39,179 +39,215 @@
...
 
-  void FASTAFile::store(const String& filename, const vector<FASTAEntry>& data) const
-  {
+  void FASTAFile::store(const String &filename, const vector<FASTAEntry> &data) const {
     startProgress(0, data.size(), "Writing FASTA file");
     FASTAFile f;
     f.writeStart(filename);
-    for (vector<FASTAEntry>::const_iterator it = data.begin(); it != data.end(); ++it)
-    {
+    for (vector<FASTAEntry>::const_iterator it = data.begin(); it != data.end(); ++it) {
```
timosachsenberg, line 250: prefer range for loops

timosachsenberg:
the non style TODO would be:
- to check if mutable would be an alternative and allow to keep the const at members
- ideally: remove the fasta headers in the thirdparty seqan folder as these are not used anymore

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -75,81 +155,72 @@ namespace OpenMS
...
 
-  bool FASTAFile::atEnd() const
+  bool FASTAFile::atEnd()
   {
-    return seqan::atEnd(*static_cast<FASTARecordReader*>(reader_.get()));
+    return (infile_.peek() == std::streambuf::traits_type::eof());
   }
 
-  void FASTAFile::load(const String& filename, vector<FASTAEntry>& data) const
+  void FASTAFile::load(const String &filename, vector<FASTAEntry> &data) const
```
timosachsenberg, line 223: prefer String&  to String &

timosachsenberg: others as well

timosachsenberg: I can do a quick fix

cbielow:
> 
> 
> the non style TODO would be:
> 
>     * to check if mutable would be an alternative and allow to keep the const at members

not sure this is worth the effort, since the class is unusable if its const anyway. To make a member mutable just to enable two(?) const member functions... not sure that justifies hiding a non-const filestream.

> 
>     * ideally: remove the fasta headers in the thirdparty seqan folder as these are not used anymore

To find out what exactly can be removed it probably not trivial. It's probably easier to remove seqan in one go once all dependencies are gone. The headers themselves don't hurt, since they do not influence build time etc.

PR #5316, opened 2021-05-12
GUI fixes

cbielow:
# Description

some refactoring towards integration of IMS into TOPPView


src/openms/include/OpenMS/KERNEL/MSSpectrum.h
```diff
@@ -227,6 +227,9 @@ namespace OpenMS
     */
     DriftTimeUnit getDriftTimeUnit() const;
 
+    /// returns the ion mobility drift time unit as string
+    String getDriftTimeUnitasString() const;
```
timosachsenberg, line 231: as->As?

src/openms/include/OpenMS/KERNEL/MSSpectrum.h
```diff
@@ -505,8 +541,13 @@ namespace OpenMS
     ConstIterator PosEnd(ConstIterator begin, CoordinateType mz, ConstIterator end) const;
 
     /// do the names of internal metadata arrays contain any hint of ion mobility data, e.g. 'Ion Mobility' 
+    /// (for spectra which represent an IM-frame)
     bool containsIMData() const;
 
+    /// get the Ion mobility data array (for spectra which represent an IM-frame)
+    /// @throws Exception::MissingInformation if IM data is not present
+    const MSSpectrum::FloatDataArray& getIMData() const;
```
timosachsenberg, line 549: is the name of the data array set (e.g., to a CV term)? if so I would document it

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -2165,65 +2168,115 @@ namespace OpenMS
...
+    @param im_frame Concatenated spectrum representing a frame
+    @param number_of_bins In how many bins should ion mobility values be sliced? Default(-1) assigns all peaks with identical ion-mobility values to a separate spectrum.
+    @return IM frame split into multiple bins (= 1 spectrum per bin)
 
-    // Get current spectrum
-    auto spidx = layer.getCurrentSpectrumIndex();
-    MSSpectrum tmps = layer.getCurrentSpectrum();
+    @throws Exception::MissingInformation if @p im_frame does not have IM data in floatDataArrays
+  */
+  MSExperiment splitByIonMobility(MSSpectrum im_frame, UInt number_of_bins = -1)
```
timosachsenberg, line 2184: sounds a bit like what we do for FAIMS. could this be moved out of TOPPView?

timosachsenberg (approved):

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
cbielow: yes!
but currently the `hack` which overwrites RT with IM is needed.
Will move when its fixed.

src/openms/include/OpenMS/METADATA/Precursor.h
```diff
@@ -60,7 +60,24 @@ namespace OpenMS
   {
 
 public:
-    // RULE of 0 for Ctor, Dtor, ...
+    /// Constructor
+    Precursor();
```
timosachsenberg, line 64: = default ?

src/openms/include/OpenMS/METADATA/Precursor.h
```diff
@@ -60,7 +60,24 @@ namespace OpenMS
...
+    Precursor(const Precursor&) = default;
+
+    // note: we implement the move constructor ourselves due to a bug in MSVS
+    // 2015/2017 which cannot produce a default move constructor for classes
+    // that contain STL containers (other than vector).
+
+    /// Move constructor
+    Precursor(Precursor&&) noexcept;
+    /// Destructor
+    ~Precursor() override;
```
timosachsenberg, line 75: + = default

src/openms/include/OpenMS/METADATA/Precursor.h
cbielow: no strong feelings here, since its either inlining it (potentially), thus making the binaries larger; or not...

Same with dtor...

timosachsenberg: Given that there is not much difference except of more concise code I am for defaulting as much as possible.

src/openms/source/SYSTEM/File.cpp
```diff
@@ -116,7 +116,7 @@ namespace OpenMS
             // ensure path ends with a "/", such that we can just write path + "ToolX", and to not worry about if its empty or a path.
             rpath.ensureLastChar('/');
           } else {
```
timosachsenberg, line 118:
```suggestion
          } 
          else 
          {
```

src/openms/include/OpenMS/KERNEL/MSSpectrum.h
cbielow: coming up in the next PR...

PR #5322, opened 2021-05-18
'DatabaseSuitability' from utils to topp tools in 'ToolHandler'

Waschi97:


timosachsenberg:
exception of type InvalidValue occurred in line 163, function OpenMS::TOPPBase::TOPPBase(const OpenMS::String &, const OpenMS::String &, bool, const std::vector<Citation> &, bool) of /Users/builder/jenkins/ws/openms/PR/bldtst/85c1a982/source/src/openms/source/APPLICATIONS/TOPPBase.cpp
error message: the value 'DatabaseSuitability' was used but is not valid; If 'DatabaseSuitability' is a Util, add it to the util list in ToolHandler. If it is not, set the 'official' flag of the TOPPBase constructor to true.

the flag in the constructor needs to be set to true indicating that it is an official tool

PR #5343, opened 2021-05-29
Include MSFragger engine in the PEPModel

tillenglert:
# Description

MSFragger engine is now included in the PEPModel.cpp including the score scaling

# Checklist:

### DIFF
--- a/src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
+++ b/src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
@@ -945,6 +945,10 @@ namespace OpenMS
       {
         return getScore_({"hyperscore"}, hit, current_score_type); //TODO evaluate transformations
       }
+      else if (engine == "MSFragger")
+      {
+        return (-1) * log10(getScore_({"expect"}, hit, current_score_type));
+      }
 
       throw Exception::UnableToFit(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "No parameters for chosen search engine", "The chosen search engine is currently not supported");
     }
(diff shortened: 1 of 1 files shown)


jpfeuffer (approved):
Thanks for the contribution. Of course one could do some research if that is the best score/score combination for discriminative distributions, but for now this is fine.

jpfeuffer:
By the way, feel free to add yourself to the Authors file in the next PR. And potentially add new features to the changelog.

src/openms/source/MATH/STATISTICS/PosteriorErrorProbabilityModel.cpp
```diff
@@ -961,6 +965,7 @@ namespace OpenMS
       const StringList search_engines = ListUtils::create<String>("XTandem,OMSSA,MASCOT,SpectraST,MyriMatch,"
                                                                   "SimTandem,MSGFPlus,MS-GF+,Comet,"
                                                                   "tide-search,SimpleSearchEngine,"
+                                                                  "MSFragger"
```
cbielow, line 968: I think this introduced a bug.
should be `"MSFragger,"` ... and since this way of specifying a list of strings is obviously not very robust, even better would be C++17 brace init:
`const StringList search_engines = { "XTandem", "OMSSA", ... }` :)

jpfeuffer: Yes @tillenglert please fix.

tillenglert: Sorry for that inconvenience, I haven't checked thorough enough and missed this. My new PR should fix it.
About that list: I tried to keep everything as it is and just include my changes, but I can include this in my next PR, thanks for that :)

PR #5352, opened 2021-06-04
IdentificationData updates, integration with tools, and SQLite-based file format

hendrikweisser:
edit by Timo (aka Timo's main concern/wishlist):

- [ ] check if copy constructor exists/works (might otherwise be an issue in pyOpenMS)
- [ ] check how we can get rid of the problematic inversion of dependency by adding IDFilter and MapTransformer as friends
- [ ] check if we have an example in Tutorial_ that covers the main use-cases of building up the IDDataStructure in C++
- [ ] check if we can export or visualize the IDDataStructure or (at least) provide a set of good queries/joins how to visualize e.g., directly with sqlite from the command line
- [x] Revisit the parts that keep an internal state (bad for parallelization) like the setCurrentProcesingStep vs. the ones that are stateless (e.g., registering a filename). The important difference is that one would forbid parallel processing, while the stateless only requires a lock for the method call (e.g., can we easily make everything stateless?). (Having a stateful builder on top of the stateless function could solve some usability implications later) -> don't use locks but merge() function on independent datastructures
- [ ] add support for the file options parameter to discard some information on loading
- [ ] add equality operator== test


oliveralka:
We have to fix the pyopenms issues and take a closer look at the changed tests (with close attention to changes in the PLFQ).

src/openms/include/OpenMS/FORMAT/OMSFile.h
```diff
@@ -0,0 +1,344 @@
...
+
+    // convenience functions:
+
+    static bool tableExists_(const String& db_name, const String& table_name);
+
+    static void raiseDBError_(const QSqlError& error, int line,
+                              const char* function, const String& context);
+
+    // store helper class:
+    class OMSFileStore: public ProgressLogger
```
hroest, line 103: @hendrikweisser can you move the helper class out of the main interface into a separate header file using the `OpenMS::Internal` namespace?

src/openms/source/FORMAT/OMSFile.cpp
```diff
@@ -0,0 +1,2675 @@
...
+    {
+      query.bindValue(":data_type", int(value.valueType()) + 1);
+    }
+    query.bindValue(":value", value.toQString());
+    if (!query.exec())
+    {
+      raiseDBError_(query.lastError(), __LINE__, OPENMS_PRETTY_FUNCTION,
+                    "error inserting data");
+    }
+    return query.lastInsertId().toLongLong();
```
hroest, line 230: @hendrikweisser do I see this correctly, this basically executes an SQL query for every single `DataValue`, isnt that a bit inefficient? Could you also do multiple binds and execute one query to enter many data values into the file? Have you looked at the performance of this?

src/openms/source/FORMAT/OMSFile.cpp
```diff
@@ -0,0 +1,2675 @@
...
+          raiseDBError_(query.lastError(), __LINE__, OPENMS_PRETTY_FUNCTION,
+                        "error inserting data");
+        }
+      }
+    } // else: use NULL for missing processing step reference
+    for (const auto& score_pair : step.scores)
+    {
+      query.bindValue(":score_type_id", Key(&(*score_pair.first)));
+      query.bindValue(":score", score_pair.second);
+      if (!query.exec())
```
hroest, line 397: also here, do you really have to execute a query each time? I think it would be faster to use `execBatch` and insert many values at once

hroest (changes requested):
This looks very useful, but I have some questions regarding the use of Qt SQL versus using Sqlite directly which we do in other parts of OpenMS. It seems redundant to both support Sqlite.h and Qt SQL and use both libraries in different areas of the code. What is the advantage of using the Qt abstractions?

Second I think you should utilize the batch execution methods more to improve performance when storing data, have you looked at the performance of this compared to feature XML? Using `execBatch` would probably improve performance quite a lot since it seems you are basically firing off an SQL query for each row in your data structure

hendrikweisser:
@hroest: I've moved the QtSql includes to the .cpp file as requested. Performance measurement and optimisation I would leave for the future. As you pointed out over email, let's not delay the integration of all this more than necessary.

> It seems redundant to both support Sqlite.h and Qt SQL and use both libraries in different areas of the code. What is the advantage of using the Qt abstractions?

I've commented on this before when I started using QtSql for this implementation. Qt offers a more complete and more C++-like interface for database interaction. In particular, it supports named placeholders in prepared statements, which makes the code much more readable.

hendrikweisser:
@hroest: Moved OMSFile helper classes to separate files and into the "Internal" namespace, as requested.

@oliveralka, @timosachsenberg: By popular demand (i.e. for ProteomicsLFQ and pyOpenMS) I've added the legacy version of `FeatureFinderIdentificationAlgorithm::run` (for `PeptideIdentification`/`ProteinIdentification`) back in. It does the conversion to the new format internally, but at least provides an interface without `IdentificationData`. Can you check if that works for you?

src/openms/include/OpenMS/ANALYSIS/MAPMATCHING/MapAlignmentAlgorithmIdentification.h
```diff
@@ -321,11 +343,17 @@ namespace OpenMS
...
-      If a reference file is supplied via the @p reference parameter, extract retention time
-      information and store it in #reference_.
+      If a reference file is supplied via the @p reference parameter, extract retention time information and store it in #reference_.
     */
     void getReference_();
 
+    /**
+      @brief Helper function to find/define the score type for processing IdentificationData
+
+      @return Reference to the score type denoted by parameter @p score_type_
```
timosachsenberg, line 353: adapt docu? score_type_ is no parameter?

src/openms/include/OpenMS/CHEMISTRY/AdductInfo.h
```diff
@@ -0,0 +1,97 @@
+// --------------------------------------------------------------------------
```
timosachsenberg, line 1: I made a diff: https://www.diffchecker.com/W9PdsUP4
that shows a few differences. 
Please check as these could be a result of something missed during the merge.
e.g., if a branch already moved code to a new header file and on the other branch changes come in to the old file things might get missed

src/openms/include/OpenMS/CHEMISTRY/EmpiricalFormula.h
```diff
@@ -271,12 +270,23 @@ namespace OpenMS
...
-    
+
     inline Iterator begin() { return formula_.begin(); }
 
     inline Iterator end() { return formula_.end(); }
     //@}
 
+    /** @name Static member functions
+     */
+    // @TODO: make these static member variables instead?
```
timosachsenberg, line 281: any opinions here?

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
```diff
@@ -93,21 +93,17 @@ namespace OpenMS
...
     : public IsotopePatternGenerator
   {
 
  public:
-    CoarseIsotopePatternGenerator();
-
-    CoarseIsotopePatternGenerator(const Size& max_isotope);
-
-    CoarseIsotopePatternGenerator(const Size& max_isotope, const bool round_masses);
+    CoarseIsotopePatternGenerator(const Size max_isotope = 0, const bool round_masses = false);
```
timosachsenberg, line 101: looks saner.
**TODO: we need to check if there are implications for the python bindings.**

src/openms/include/OpenMS/FILTERING/ID/IDFilter.h
```diff
@@ -1377,11 +1378,20 @@ namespace OpenMS
 
     /// @name Filter functions for class IdentificationData
     ///@{
-    static void keepBestMatchPerQuery(
+
```
timosachsenberg, line 1381: Hi Hendrik. Would be great if you could add here more documentation.
We might need to add quite a bit of new IDData code to IDFilter and this one looks super useful. 
Can you add some @brief documentation including the cleanup_affected parameter (e.g., what get's cleaned up)

hroest:
> Performance measurement and optimisation I would leave for the future.

can you give some ballpark estimate on how long it takes to read/write a reasonably large file with the Sql data structure vs writing the same data to an XML file? I just want to know if its fast enough to be useable, of course we can optimize later but at least it should be "reasonably fast"

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
```diff
@@ -0,0 +1,137 @@
...
+    public:
+      using Key = qint64;
+ 
+      OMSFileLoad(const String& filename, LogType log_type);
+
+      ~OMSFileLoad();
+
+      void load(IdentificationData& id_data);
+
+      void load(FeatureMap& features);
```
timosachsenberg, line 59: can you document the public members and note e.g., how complete the support is.

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
```diff
@@ -0,0 +1,137 @@
...
+{
+  namespace Internal
+  {
+    /// Helper class for loading .oms files (SQLite format)
+    class OMSFileLoad: public ProgressLogger
+    {
+    public:
+      using Key = qint64;
+ 
+      OMSFileLoad(const String& filename, LogType log_type);
```
timosachsenberg, line 53: Can you document why we need state here?
e.g., why a static load (filename ,...) is not sufficient

timosachsenberg: maybe this would be clearer with a bit more documentation

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
```diff
@@ -0,0 +1,137 @@
...
+#include <OpenMS/KERNEL/FeatureMap.h>
+#include <OpenMS/METADATA/ID/IdentificationData.h>
+
+class QSqlQuery;
+
+namespace OpenMS
+{
+  namespace Internal
+  {
+    /// Helper class for loading .oms files (SQLite format)
```
timosachsenberg, line 47: can you document how concurrent access is handled (not sure if this is the right place).
e.g., what happens if two threads load from the same file?
what happens if one stores and one loads (undefined behavior? error?)

src/openms/include/OpenMS/KERNEL/BaseFeature.h
```diff
@@ -36,11 +36,13 @@
 
 #include <OpenMS/KERNEL/RichPeak2D.h>
 #include <OpenMS/METADATA/PeptideIdentification.h>
+#include <OpenMS/METADATA/ID/IdentificationData.h>
+
+#include <boost/optional.hpp>
```
timosachsenberg, line 41: can you do a quick check if we can just use std::optional? this seems to predate the sqitch to C17

src/openms/include/OpenMS/KERNEL/BaseFeature.h
```diff
@@ -178,6 +192,34 @@ namespace OpenMS
     /// state of peptide identifications attached to this feature. If one ID has multiple hits, the output depends on the top-hit only
     AnnotationState getAnnotationState() const;
 
+    /// has a primary ID (peptide, RNA, compound) been assigned?
```
timosachsenberg, line 195: I think if we keep both PeptideIdentifications and the new IDDatastructure we need to properly group them - like that I think nobody will figure out what belongs together

src/openms/include/OpenMS/KERNEL/Feature.h
```diff
@@ -183,6 +183,9 @@ namespace OpenMS
       return assignments;
     }
 
+    /// update ID referenes (primary ID, input matches) for this feature and any subfeatures
```
timosachsenberg, line 186: can you add more documentation here for what this is used
e.g. use cases or example code.
like that I find it hard to make sense of that function here (without extensive researching into the ID datastructure)

src/openms/include/OpenMS/KERNEL/FeatureMap.h
```diff
@@ -306,6 +307,12 @@ namespace OpenMS
 
     OPENMS_DLLAPI AnnotationStatistics getAnnotationStatistics() const;
 
+    OPENMS_DLLAPI std::set<IdentificationData::ObservationMatchRef> getUnassignedIDMatches() const;
```
timosachsenberg, line 310: here as well... we might really need to group the api of the two different ID types clearly

src/openms/source/METADATA/DataProcessing.cpp
```diff
@@ -75,58 +75,57 @@ namespace OpenMS
   {
   }
 
-  bool DataProcessing::operator==(const DataProcessing & rhs) const
+  bool DataProcessing::operator==(const DataProcessing& rhs) const
   {
     return software_ == rhs.software_ &&
```
timosachsenberg, line 80: Is there something that might be faster to compare first? e.g., the completion_time_?

src/openms/source/KERNEL/FeatureMap.cpp
```diff
@@ -193,12 +202,21 @@ namespace OpenMS
...
     data_processing_.insert(data_processing_.end(), rhs.data_processing_.begin(), rhs.data_processing_.end());
 
+    Size n_old_features = size();
     // append features:
     this->insert(this->end(), rhs.begin(), rhs.end());
 
     // todo: check for double entries
     // features, unassignedpeptides, proteins...
 
+    // merge IDs (new format):
```
timosachsenberg, line 212: would it make sense to have a mergeAndUpdateAllIDReferences() sounds like this pattern could arise quite often?

src/openms/source/KERNEL/FeatureMap.cpp
```diff
@@ -124,10 +123,19 @@ namespace OpenMS
...
+    IdentificationData::RefTranslator trans = id_data_.merge(source.id_data_);
+    for (Feature& feature : *this)
+    {
+      feature.updateAllIDReferences(trans);
+    }
   }
 
+  FeatureMap::FeatureMap(FeatureMap&& source) = default;
+
   FeatureMap::~FeatureMap()
```
timosachsenberg, line 139: = default

src/openms/source/KERNEL/Feature.cpp
```diff
@@ -234,4 +234,13 @@ namespace OpenMS
     subordinates_ = rhs;
   }
 
+  void Feature::updateAllIDReferences(const IdentificationData::RefTranslator& trans)
+  {
+    updateIDReferences(trans);
```
timosachsenberg, line 239: can you add some comment here so it is easier to understand why updateID and updateAllID are used...

src/openms/source/KERNEL/BaseFeature.cpp
```diff
@@ -191,18 +188,84 @@ namespace OpenMS
...
-      }
-      else if (seqs.size() > 1)
-      {
-        return FEATURE_ID_MULTIPLE_DIVERGENT;                        // multiple different annotations ... probably bad mapping
-      }
-      else /*if (seqs.size()==0)*/
+      if (seqs.size() == 1) return FEATURE_ID_MULTIPLE_SAME; // hits have identical seqs
+      if (seqs.size() > 1)
+        return FEATURE_ID_MULTIPLE_DIVERGENT; // multiple different annotations ... probably bad mapping
+      else /*if (seqs.size()==0)*/ return FEATURE_ID_NONE;   // very rare case of empty hits
```
timosachsenberg, line 194: please use brackets here

src/openms/source/KERNEL/BaseFeature.cpp
```diff
@@ -191,18 +188,84 @@ namespace OpenMS
...
+
+  void BaseFeature::updateIDReferences(const IdentificationData::RefTranslator& trans)
+  {
+    if (primary_id_ != boost::none)
+    {
+      primary_id_ = trans.translate(*primary_id_);
+    }
+    set<IdentificationData::ObservationMatchRef> matches;
+    matches.swap(id_matches_);
+    for (auto item : matches)
```
timosachsenberg, line 266: does item need to be copied here? or would const auto& also work?

src/openms/source/KERNEL/BaseFeature.cpp
```diff
@@ -191,18 +188,84 @@ namespace OpenMS
...
+
+
+  void BaseFeature::addIDMatch(IdentificationData::ObservationMatchRef ref)
+  {
+    id_matches_.insert(ref);
+  }
+
+  void BaseFeature::updateIDReferences(const IdentificationData::RefTranslator& trans)
+  {
+    if (primary_id_ != boost::none)
```
timosachsenberg, line 260: Just based on this context I don't have a good intuition what it means of heaving this optional parameter
and what the translation does. 
can you add a comment here to guide the reader in the right direction?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -93,23 +93,41 @@ namespace OpenMS
...
   {
-    return input_files_.insert(file).first;
+    if (!no_checks_ && file.name.empty()) // key may not be empty
+    {
+      String msg = "input file must have a name";
+      throw Exception::IllegalArgument(__FILE__, __LINE__,
+                                       OPENMS_PRETTY_FUNCTION, msg);
+    }
+    auto result = input_files_.insert(file);
+    if (!result.second) // existing element - merge in new information
```
timosachsenberg, line 105: can you add a comment here what data get's merged?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -93,23 +93,41 @@ namespace OpenMS
...
-  IdentificationData::registerDataProcessingSoftware(
-    const DataProcessingSoftware& software)
+  IdentificationData::registerProcessingSoftware(
+    const ProcessingSoftware& software)
   {
-    for (ScoreTypeRef score_ref : software.assigned_scores)
+    if (!no_checks_)
     {
-      if (!isValidReference_(score_ref, score_types_))
+      for (ScoreTypeRef score_ref : software.assigned_scores)
```
timosachsenberg, line 123: I guess this is just an iterator? then the copy is fine

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -125,44 +143,47 @@ namespace OpenMS
...
 
   IdentificationData::ProcessingStepRef
-  IdentificationData::registerDataProcessingStep(
-    const DataProcessingStep& step, SearchParamRef search_ref)
+  IdentificationData::registerProcessingStep(
+    const ProcessingStep& step, SearchParamRef search_ref)
   {
-    // valid reference to software is required:
-    if (!isValidReference_(step.software_ref, processing_softwares_))
+    if (!no_checks_)
```
timosachsenberg, line 157: minor: I would have preferred if (check_inserts_) as opposed to the double negation

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -190,42 +212,59 @@ namespace OpenMS
...
+
+    // can't use "insertIntoMultiIndex_" because Observation doesn't have the
+    // "steps_and_scores" member (from ScoredProcessingResult)
+    auto result = observations_.insert(obs);
+    if (!result.second) // existing element - merge in new information
     {
-      String msg = "invalid reference to an input file - register that first";
-      throw Exception::IllegalArgument(__FILE__, __LINE__,
-                                       OPENMS_PRETTY_FUNCTION, msg);
+      observations_.modify(result.first, [&obs](Observation& existing)
```
timosachsenberg, line 241: can you document what information get's merged here?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -190,42 +212,59 @@ namespace OpenMS
...
+                           {
+                             existing.merge(obs);
+                           });
     }
-    DataQueryRef ref = data_queries_.insert(query).first;
-    data_query_lookup_.insert(ref);
-    return ref;
+
+    observation_lookup_.insert(uintptr_t(&(*result.first)));
```
timosachsenberg, line 247: add a short comment

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -190,42 +212,59 @@ namespace OpenMS
...
+
+    // @TODO: add processing step (currently not supported by Observation)
```
timosachsenberg, line 249: what could be an application of having a processing step at observation?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -251,154 +290,190 @@ namespace OpenMS
...
+    if (!no_checks_)
     {
-      checkScoreTypes_(group.scores);
+      checkAppliedProcessingSteps_(groups.steps_and_scores);
 
-      for (const auto& ref : group.parent_molecule_refs)
+      for (const auto& group : groups.groups)
       {
-        if (!isValidHashedReference_(ref, parent_molecule_lookup_))
+        checkScoreTypes_(group.scores);
```
timosachsenberg, line 341: add short comment what is checked

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -251,154 +290,190 @@ namespace OpenMS
...
 
     // add the current processing step?
     if ((current_step_ref_ != processing_steps_.end()) &&
-        (grouping.steps_and_scores.get<1>().find(current_step_ref_) ==
-         grouping.steps_and_scores.get<1>().end()))
+        (groups.steps_and_scores.get<1>().find(current_step_ref_) ==
+         groups.steps_and_scores.get<1>().end()))
     {
-      parent_molecule_groupings_.back().steps_and_scores.push_back(
+      parent_groups_.back().steps_and_scores.push_back(
```
timosachsenberg, line 362: this is... like many other functions - not thread safe.
Is this documented? e.g., that users need to ensure non-concurrent modification of the datastructure?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -251,154 +290,190 @@ namespace OpenMS
...
-  IdentificationData::registerMoleculeQueryMatch(const MoleculeQueryMatch&
-                                                 match)
+  IdentificationData::AdductRef
+  IdentificationData::registerAdduct(const AdductInfo& adduct)
   {
-    if (const IdentifiedPeptideRef* ref_ptr =
-        boost::get<IdentifiedPeptideRef>(&match.identified_molecule_ref))
+    // @TODO: require non-empty name? (auto-generate from formula?)
+    auto result = adducts_.insert(adduct);
+    if (!result.second && (result.first->getName() != adduct.getName()))
```
timosachsenberg, line 373: when does this case happen? e.g., what needs to equal? e.g. having two different adducts with same formula should be ok I guess

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -421,68 +496,82 @@ namespace OpenMS
...
-    vector<QueryMatchRef> results;
-    bool higher_better = score_ref->higher_better;
+    vector<ObservationMatchRef> results;
     pair<double, bool> best_score = make_pair(0.0, false);
-    QueryMatchRef best_ref = query_matches_.end();
-    for (QueryMatchRef ref = query_matches_.begin();
-         ref != query_matches_.end(); ++ref)
+    ObservationMatchRef best_ref = observation_matches_.end();
+    Size n_matches = 1; // number of matches for current observation
+    // matches for same observation appear consecutively, so just iterate:
```
timosachsenberg, line 521: How is this ensured and tested so we don't regress on that property by accident?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -628,126 +717,128 @@ namespace OpenMS
...
          it != identified_oligos_.end(); ++it)
     {
-      id_refs.insert(it);
+      id_vars.insert(it);
     }
-    removeFromSetIf_(query_matches_, [&](MoleculeQueryMatches::iterator it)
+    removeFromSetIf_(observation_matches_, [&](ObservationMatches::iterator it)
                      {
-                       return !id_refs.count(it->identified_molecule_ref);
+                       return !id_vars.count(it->identified_molecule_var);
```
timosachsenberg, line 739: potential performance problem? count will go through the whole thing

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -628,126 +717,128 @@ namespace OpenMS
...
 
-    // remove entries from parent molecule groups based on parent molecules:
+    // remove entries from parent sequence groups based on parent sequences:
     bool warn = false;
-    for (auto& grouping : parent_molecule_groupings_)
+    for (auto& groups : parent_groups_)
     {
-      for (auto group_it = grouping.groups.begin();
-           group_it != grouping.groups.end(); )
+      for (auto group_it = groups.groups.begin();
```
timosachsenberg, line 827: can we rename the variable a bit so we don't have groups.groups?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -628,126 +717,128 @@ namespace OpenMS
...
-      removeFromSetIfNotHashed_(parent_molecules_, parent_molecule_lookup_);
-      // update look-up table of parent molecule addresses (again):
-      updateAddressLookup_(parent_molecules_, parent_molecule_lookup_);
+      removeFromSetIfNotHashed_(parents_, parent_lookup_);
+      // update look-up table of parent sequence addresses (again):
+      updateAddressLookup_(parents_, parent_lookup_);
     }
 
-    // remove entries from parent molecule groups based on parent molecules:
+    // remove entries from parent sequence groups based on parent sequences:
```
timosachsenberg, line 823: can you add another line or two of comments what is happening here

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -788,7 +877,302 @@ namespace OpenMS
...
+    const IdentificationData::ScoredProcessingResult& other,
+    const RefTranslator& trans)
+  {
+    result.MetaInfoInterface::operator=(other);
+    for (const AppliedProcessingStep& applied : other.steps_and_scores)
+    {
+      AppliedProcessingStep copy;
+      if (applied.processing_step_opt)
+      {
+        copy.processing_step_opt = trans.processing_step_refs.at(*applied.processing_step_opt);
```
timosachsenberg, line 909: can you add a comment here regarding the trans

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -788,7 +877,302 @@ namespace OpenMS
...
+      IdentifiedPeptide copy(other_ref->sequence, ParentMatches());
+      // now copy steps/scores and parent matches while updating references:
+      mergeScoredProcessingResults_(copy, *other_ref, trans);
+      for (const auto& pair : other_ref->parent_matches)
+      {
+        ParentSequenceRef parent_ref = trans.parent_sequence_refs[pair.first];
+        copy.parent_matches[parent_ref] = pair.second;
+      }
+      // @TODO: with C++17, 'map::extract' offers a better way to update keys
+      trans.identified_peptide_refs[other_ref] = registerIdentifiedPeptide(copy);
```
timosachsenberg, line 1011: we ave C++17 now. how would this look like?

src/openms/source/METADATA/ID/IdentificationData.cpp
```diff
@@ -788,7 +877,302 @@ namespace OpenMS
...
+      result.addProcessingStep(copy);
+    }
+  }
+
+
+  IdentificationData::RefTranslator
+  IdentificationData::merge(const IdentificationData& other)
+  {
+    RefTranslator trans;
+    no_checks_ = true;
```
timosachsenberg, line 925: document why checks are disabled.
speed?

src/utils/NucleicAcidSearchEngine.cpp
```diff
@@ -161,16 +162,25 @@ class NucleicAcidSearchEngine :
     registerInputFile_("in", "<file>", "", "Input file: spectra");
     setValidFormats_("in", ListUtils::create<String>("mzML"));
 
-    registerInputFile_("database", "<file>", "", "Input file: sequence database");
+    registerInputFile_("database", "<file>", "", "Input file: sequence database. Required unless 'digest' is set.", false);
     setValidFormats_("database", ListUtils::create<String>("fasta"));
 
+    registerInputFile_("digest", "<file>", "", "Input file: pre-digested sequence database. Can be used instead of 'database'. Sets all 'oligo:...' parameters.", false);
+    setValidFormats_("digest", {"oms"});
```
timosachsenberg, line 169: is oms the file format? or fasta?

src/utils/NucleicAcidSearchEngine.cpp
```diff
@@ -783,104 +784,67 @@ class NucleicAcidSearchEngine :
...
-    // reference this step in all following ID data items, if applicable:
-    id_data.setCurrentProcessingStep(step_ref);
-  }
-
-
   void calculateAndFilterFDR_(IdentificationData& id_data, bool only_top_hits)
   {
     IdentificationData::ScoreTypeRef score_ref =
-      id_data.findScoreType("hyperscore").first;
+      id_data.findScoreType("hyperscore");
```
timosachsenberg, line 807: ident

src/utils/NucleicAcidSearchEngine.cpp
```diff
@@ -991,35 +976,134 @@ class NucleicAcidSearchEngine :
...
     spectra.sortSpectra(true);
 
+    // input file meta data:
+    String input_name = test_mode_ ? File::basename(in_mzml) : in_mzml;
+    IdentificationData::InputFile input(input_name);
+    vector<String> primary_files;
+    spectra.getPrimaryMSRunPath(primary_files);
+    input.primary_files.insert(primary_files.begin(), primary_files.end());
+    IdentificationData::InputFileRef file_ref =
+      id_data.registerInputFile(input);
```
timosachsenberg, line 1016: ident

src/topp/MapAlignerIdentification.cpp
```diff
@@ -467,6 +493,74 @@ class TOPPMapAlignerIdentification :
         progresslogger.endProgress();
       }
     }
+    break;
+
+    //-------------------------------------------------------------
+    // perform spectrum match alignment
+    //-------------------------------------------------------------
+    case FileTypes::OMS:
```
timosachsenberg, line 501: do we have a test for that new feature?

src/tests/topp/ProteomicsLFQ_7_out.mzTab
```diff
@@ -70,59 +70,59 @@ PRT	tr|A9FUD7|A9FUD7_SORC5	Putative membrane protein OS=Sorangium cellulosum (st
...
+PSM	KQTALVELLK	15	P02769|ALBU_BOVIN	1	18Protein_SoCe_Tr_detergents_trace_target_decoy	null	[, , OMSSA, 2.1.4]	0.0	null	2494.881591796880002	3	381.576354980468977	381.576298143604333	ms_run[2]:spectrum=3462	K	H	null	null	11275867043381640475	1	spectrum=3462	KQTALVELLK	internal	2.89327520051305e-10	0
+PSM	KQTALVELLK	16	P02769|ALBU_BOVIN	1	18Protein_SoCe_Tr_detergents_trace_target_decoy	null	[, , OMSSA, 2.1.4]	0.0	null	2492.93994140625	2	571.860717773437955	571.860808982021013	ms_run[2]:spectrum=3459	K	H	null	null	17447818880105247092	1	spectrum=3459	KQTALVELLK	internal	2.62096624298549e-05	0
+PSM	RHPEYAVSVLLR	17	P02769|ALBU_BOVIN	1	18Protein_SoCe_Tr_detergents_trace_target_decoy	null	[, , OMSSA, 2.1.4]	0.0	null	2412.920654296880002	3	480.608703613281023	480.608772593904405	ms_run[2]:spectrum=3369	R	L	null	null	1196373166564353753	1	spectrum=3369	RHPEYAVSVLLR	internal	9.077820128403891e-06	0
```
timosachsenberg, line 108: - is there a way to make the order of the column with "internal" the same in both test files?
- is there a way to order the PSMs in the same way?

timosachsenberg: at least we should have good reason to break potential down-stream scripts that hard-coded the column index for their data

src/tests/topp/ProteomicsLFQ_5_out.mzTab
```diff
@@ -108,164 +108,164 @@ PRT	tr|A9G8G1|A9G8G1_SORC5	Putative uncharacterized protein OS=Sorangium cellulo
...
+PEP	DGDIEAEISR	tr|A9EY18|A9EY18_SORC5	1	null	null	null	0.0	null	0.0	null	null	null	null	null	1523.370634117893133	null	3	368.843772909670918	ms_run[2]:spectrum=2311	null	null	null	2788.036376953125	null	null	null	null	null	DGDIEAEISR	2935923263525422257	internal	8.461750035304259e-04	0	368.843772909670918	1523.370634117893133	null	null	null	null
+PEP	SHCIAEVEK	P02769|ALBU_BOVIN	1	null	null	null	0.0	null	null	0.0	null	null	null	3-UNIMOD:4	1533.479042701062554	null	3	358.174576486337685	ms_run[3]:spectrum=2376	null	null	null	null	null	null	4.618461875e05	null	null	SHC(Carbamidomethyl)IAEVEK	14086301720075657476	internal	6.76118991622199e-03	0	null	null	358.174576486337685	1533.479042701062554	null	null
+PEP	EVEAICHSK	P02769|ALBU_BOVIN_rev	1	null	null	null	0.0	0.0	null	null	null	null	null	6-UNIMOD:4	1558.221639541818604	null	3	358.174576486337628	ms_run[1]:spectrum=2458	1.35815075e06	null	null	null	null	null	null	null	null	EVEAIC(Carbamidomethyl)HSK	5634796270667792442	internal	6.24514018064888e-04	1	null	null	null	null	358.174576486337628	1558.221639541818604
```
timosachsenberg, line 114: why was internal missing in the old version?
I see it in the PSM section though

src/tests/topp/IDMerger_6_output.idXML
```diff
@@ -0,0 +1,514 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" href="https://www.openms.de/xml-stylesheet/IdXML.xsl" ?>
+<IdXML version="1.5" xsi:noNamespaceSchemaLocation="https://www.openms.de/xml-schema/IdXML_1_5.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
timosachsenberg, line 3: nice that you also provide the human readable one.
is this also tested against e.g. the output from the old format? e.g., can we trust this output?

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -662,8 +858,27 @@ namespace OpenMS
...
+      {
+        element.setMetaValue(key, value);
+      });
+    }
+
 
-    // IDFilter needs access to do its job:
+    // these classes need access to manipulate data:
     friend class IDFilter;
+    friend class MapAlignmentTransformer;
```
timosachsenberg, line 882: **this is IMHO a design mistake that needs to be fixed.**
IDDataStructure needs a clean/safe and efficient interface so that IDFilter and MapAlignmentTransformer can implement the modifications.
Otherwise, we can't put IDDataStructure in a library on which others build to e.g. implement custom IDFilter etc.

src/openms/include/OpenMS/METADATA/ID/MetaData.h
```diff
@@ -64,20 +66,14 @@ namespace OpenMS
     {
       PROTEIN,
       COMPOUND,
-      RNA,
-      SIZE_OF_MOLECULETYPE
+      RNA
```
timosachsenberg, line 69: is the SIZE_OF_... entry not needed anymore?

src/openms/include/OpenMS/METADATA/ID/ParentGroup.h
```diff
@@ -44,35 +44,42 @@ namespace OpenMS
...
+    /** @brief: Group of ambiguously identified parent sequences (e.g. protein group)
     */
     // @TODO: derive from MetaInfoInterface?
-    struct ParentMoleculeGroup
+    struct ParentGroup
     {
       std::map<ScoreTypeRef, double> scores;
       // @TODO: does this need a "leader" or some such?
-      std::set<ParentMoleculeRef> parent_molecule_refs;
+      std::set<ParentSequenceRef> parent_refs;
```
timosachsenberg, line 54: I think we have some ordering in our code. Is the operator< stable (e.g., based on the sequence) or some address?

timosachsenberg: Maybe also a question for @jpfeuffer when he is back from vacation

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: The idea (in IdentificationData in general) is that the merging gets delegated to the respective data classes. So IdentificationData itself doesn't know, and shouldn't need to know, what gets updated.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: Yes, all the `...Ref` types are.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: It's this way around to have a flag that's off by default, i.e. you have to explicitly enable "no checks".

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: See https://github.com/OpenMS/OpenMS/pull/5352#discussion_r758288483.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: Not sure, which is why observations currently don't have processing steps. ;-)
I've added a question mark to the comment.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: I've added a warning to the top-level class documentation.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: Currently different adducts with the same sum formula and charge aren't allowed (i.e. they're considered the same adduct). This could be changed, but would introduce ambiguities because in practice adducts are matched by mass and charge. So if you get a match for one, does that also imply a match for the other (with same formula/charge)?

See `AdductCompare` in ObservationMatch.h (https://github.com/hendrikweisser/OpenMS/blob/98656f05c5be92a6be822e65f6c6f450f37e4a83/src/openms/include/OpenMS/METADATA/ID/ObservationMatch.h#L56-L66).

A possible alternative would be to keep a list of names for the "same" adduct.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: It's a consequence of how the matches are indexed in their multi-index container (ordered by observation first). I've extended the comment there.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: "The whole thing" is a set, so `count` will just test for existence.

src/utils/NucleicAcidSearchEngine.cpp
hendrikweisser: .oms is correct.

src/openms/source/METADATA/ID/IdentificationData.cpp
timosachsenberg: Ok but maybe we need a central point to divinest this for all data. Maybe at the level of the ID datastructure because users might look there first.

src/openms/source/METADATA/ID/IdentificationData.cpp
timosachsenberg: Then maybe just a generic comment like „set or overwrite values associated with current key“ ? If that is what it does …

src/openms/source/METADATA/ID/IdentificationData.cpp
timosachsenberg: How big would be the implications/changes needed if we would need to add this later?

src/openms/source/METADATA/ID/IdentificationData.cpp
timosachsenberg: Hmm … I could imagine that many structures that could be potential affects have the same formula

timosachsenberg: @oliveralka can you think about the implications?

src/topp/MapAlignerIdentification.cpp
hendrikweisser: Yes, "TOPP_MapAlignerIdentification_8".

src/openms/include/OpenMS/METADATA/ID/ParentGroup.h
hendrikweisser: All "...Ref" types are ordered by address, so not stable. See `IteratorWrapper::operator<` (https://github.com/hendrikweisser/OpenMS/blob/98656f05c5be92a6be822e65f6c6f450f37e4a83/src/openms/include/OpenMS/METADATA/ID/MetaData.h#L51-L55).

If we're revisiting the design of the references in the context of pyOpenMS, we could rethink this decision.

src/openms/include/OpenMS/METADATA/ID/MetaData.h
hendrikweisser: I found it just led to compiler warnings about unhandled cases in "switch" statements.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: This is a tricky one. IdentificationData purposely doesn't expose an interface for low-level manipulation (e.g. removing individual items of data), because consistency of the contents needs to be ensured and re-establishing that is costly (see `IdentificationData::cleanup`). Instead more high-level functions are provided (for filtering and transforming retention times) that however need privileged access to IdentificationData internals. It would be possible to move these functions into IdentificationData itself (to avoid the friend classes), but from a "separation of concerns" viewpoint I think that makes less sense.

> Otherwise, we can't put IDDataStructure in a library on which others build to e.g. implement custom IDFilter etc.

The point is that "user code" shouldn't implement filtering. It should just use the library functions that are implemented in a safe manner in IDFilter.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: Actually, RT transformation is a processing step that gets applied to observations, because that's where the RT information is stored. Currently that is not tracked explicitly, because the observations are considered "meta data", rather than "actual data". It could definitely be changed if we wanted to, but would make each observation "bigger".

src/tests/topp/ProteomicsLFQ_7_out.mzTab
hendrikweisser: I don't understand why the column order would change. Meta values of PSMs are just copied from IdentificationData's ObservationMatches to PeptideHits, so there's no special treatment (https://github.com/hendrikweisser/OpenMS/blob/98656f05c5be92a6be822e65f6c6f450f37e4a83/src/openms/source/METADATA/ID/IdentificationDataConverter.cpp#L371).

The order of PSMs in IdentificationData is essentially random (based on memory addresses), so to ensure a consistent order, you'd need to sort them before output. This is done when exporting IdentificationData to mzTab directly (https://github.com/hendrikweisser/OpenMS/blob/98656f05c5be92a6be822e65f6c6f450f37e4a83/src/openms/source/METADATA/ID/IdentificationDataConverter.cpp#L724).

Down-stream scripts should check the order of columns in the header to avoid surprises.

src/tests/topp/IDMerger_6_output.idXML
hendrikweisser: It's not currently tested with and without conversion, but I'm pretty sure I've checked the output file for completeness.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
timosachsenberg: IMHO That argument doesn’t fly for such a low level functionality. So how would this look like if we add functionality tools? E.g. PeptideIndexer? Or some of the inference algorithms? How can there be implemented without being a friend? OpenMS is a library and if that is our future proof data structure this should be possible for library users to implement this.

hendrikweisser: Any tool/algorithm that adds data (which should be the majority) is fine.

Removing data is more difficult because of the consistency requirements. That is a side effect of having complex data connected by iterators. You cannot change this without losing the advantage of O(1) dereferencing. (Also, IdentificationData keeps look-up tables of object addresses for fast checking of reference validity, which naturally need to be kept up-to-date.)

Modifying stored data is unfortunately also difficult, because data in a `set` (or equivalently: multi-index container) is `const` (has to be to uphold the invariants of how elements are ordered). Multi-index offers mechanisms for updating elements (see https://www.boost.org/doc/libs/1_77_0/libs/multi_index/doc/tutorial/basics.html#ord_updating), the most efficient of which is `modify` - however, for that you need to implement a functor and ensure that the "key" of the element doesn't change. So this is not exposed as a general interface, but used to implement specific modification functions (e.g. `IdentificationData::addScore`).

The gist is this:
- We cannot allow users to remove elements because it could break the consistency guarantees.
- We cannot allow users to make general updates to elements because it could break container invariants.

With these constraints in mind, what would you suggest to improve the interface?
We could still add a more general updating mechanism using multi-index's `replace` (after checking that the "key" part of an element does not change), but I haven't found a compelling use case for this (that wasn't solved more efficiently using `modify`).

src/openms/include/OpenMS/CHEMISTRY/AdductInfo.h
hendrikweisser: Nothing missed, just some additions.

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
hendrikweisser: Added documentation. Support is intended to be complete (but needs more testing).

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
timosachsenberg: I know it is a challenge. I am just looking at code I wrote and I often implement custom remove_if with lambda on PeptideHits and then call the usual removeUnreferenced updateGroups etc. to ensure that data is again consistent. This looks very C++ like. 
Brainstorming:
E.g could one one get additional unsafe access to delete and modify contents + mandatory cleanup/consistency check afterwards? Like e.g. unique_ptr that on destruction calls the consistency check / cleanup? Ideally wrappable in python

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
hendrikweisser: OMSFileLoad and OMSFileStore have internal variables (e.g. for the database name) that are used, hence no `static`.

For OMSFile itself, the only internal variable stores the log type. That could be moved to the `store`/`load` functions and those could become `static`, but to me it makes more sense to configure logging during object construction.

src/openms/include/OpenMS/KERNEL/BaseFeature.h
hendrikweisser: Yep, replaced and works.

src/openms/source/KERNEL/FeatureMap.cpp
hendrikweisser: I don't see where this would arise outside of the `operator+=` where it's currently implemented.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: I've added a sentence to the IdentificationData documentation.

src/openms/source/METADATA/DataProcessing.cpp
hendrikweisser: Premature optimisation? In any case, not my code.

src/openms/include/OpenMS/METADATA/ID/Observation.h
```diff
@@ -34,60 +34,71 @@
...
-
-      // ignore RT and m/z for comparisons to avoid issues with rounding:
-      bool operator==(const DataQuery& other) const
+      /// Merge in data from another object
+      Observation& merge(const Observation& other)
       {
-        return std::tie(input_file_opt, data_id) ==
-          std::tie(other.input_file_opt, other.data_id);
+        // merge meta info - existing entries may be overwritten:
+        std::vector<UInt> keys;
```
cbielow, line 78: use `void MetaInfoInterface::addMetaValues(const MetaInfoInterface &  from) `?

src/openms/include/OpenMS/METADATA/ID/ObservationMatch.h
```diff
@@ -0,0 +1,154 @@
...
+                     PeakAnnotations> PeakAnnotationSteps;
+
+    /// Comparator for adducts
+    // @TODO: this allows adducts with duplicate names, but requires different
+    // sum formulas/charges - is this what we want?
+    struct AdductCompare
+    {
+      bool operator()(const AdductInfo& left, const AdductInfo& right) const
+      {
+        return (std::make_pair(left.getCharge(), left.getEmpiricalFormula()) <
```
cbielow, line 63: this is a bit inefficient?!
use `std::tie`?

src/openms/include/OpenMS/METADATA/ID/ObservationMatch.h
```diff
@@ -0,0 +1,154 @@
...
+          boost::multi_index::composite_key<
+            ObservationMatch,
+            boost::multi_index::member<ObservationMatch, ObservationRef,
+                                       &ObservationMatch::observation_ref>,
+            boost::multi_index::member<
+              ObservationMatch, IdentifiedMolecule,
+              &ObservationMatch::identified_molecule_var>,
+            boost::multi_index::member<ObservationMatch, AdductOpt,
+                                       &ObservationMatch::adduct_opt>>>>
+      > ObservationMatches;
```
cbielow, line 150: easy. I could write this any time of the day...
:P

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: We could add `filterXYZ(predicate, ...)` functions for each of the main data types (`XYZ`). They would use a user-supplied predicate (functor, lambda, ...) to determine whether to keep/remove each element, then call `cleanup` once at the end (with additional arguments to `cleanup` passed from `...`).

src/openms/include/OpenMS/METADATA/ID/ObservationMatch.h
hendrikweisser: That's C++ template magic for you. If you follow the Boost documentation, it's not too hard to use, though.

src/openms/include/OpenMS/METADATA/ID/ObservationMatch.h
timosachsenberg: can you tie to a return type? not sure

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
timosachsenberg: Fine with me. Concerns by others?

src/tests/topp/ProteomicsLFQ_5_out.mzTab
timosachsenberg: Just looking at the diffs

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
timosachsenberg: The filter with predicate function is a step in the right direction. But we would need to consider:
- how to access the protected/private members?
- cleanup after every step seems a bit wasteful? Wouldn't that just allow AND combinations of filters between the different datatypes?
- how to e.g. filter PSMs based on properties in the different levels (e.g. remove all PSMs that match to a set of proteins with coverage > 90% OR are part of a protein group with some meta value )
- Shouldn't that be similar to a transaction? Do the checks only in the end?

src/openms/include/OpenMS/METADATA/ID/ObservationMatch.h
hendrikweisser: I'm pretty sure you can use `tie` with the comparison operator, but is it more efficient than `make_pair`?

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: You could copy the whole `parent_matches` map of the peptide, then change the keys (i.e. references to parent sequences) using this approach:
https://en.cppreference.com/w/cpp/container/map/extract#Notes
But actually that doesn't seem easier or more efficient in the end.

src/tests/topp/ProteomicsLFQ_7_out.mzTab
hendrikweisser: Do we need to change anything here before this can get merged?

src/tests/topp/ProteomicsLFQ_5_out.mzTab
hendrikweisser: Do we need to change anything here before this can get merged?

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: > how to access the protected/private members?

The filter functions would need to be in `IdentificationData` or a friend class (IDFilter).

> cleanup after every step seems a bit wasteful? Wouldn't that just allow AND combinations of filters between the different datatypes?

Much better than clean-up after removing a single element. Also, if you don't clean up between handling different data types, you may run into dangling references (e.g. if you remove proteins, then filter peptides by looking at their proteins).

> how to e.g. filter PSMs based on properties in the different levels (e.g. remove all PSMs that match to a set of proteins with coverage > 90% OR are part of a protein group with some meta value )

You always have read access to all the data, so complex filtering criteria aren't a problem.

In any case, can we get the merge done first, then build more filtering functionality?

src/openms/include/OpenMS/METADATA/ID/ObservationMatch.h
timosachsenberg: Tie with the comparison operator is not an issue but the parameters passed to tie. I think these might not be temporaries. Will look it up…

src/tests/topp/ProteomicsLFQ_7_out.mzTab
timosachsenberg: We need to be very sure that we don’t have (spike) memory usage or other regressions as we generate the packages for an ongoing large scale study using ProteomicsLFQ.

src/openms/source/METADATA/ID/IdentificationData.cpp
hendrikweisser: I'm removing the comment.

src/openms/include/OpenMS/METADATA/ID/Observation.h
hendrikweisser: Thanks. The code was written before this function existed. I've also changed it in `ScoreProcessingResult`.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
timosachsenberg: How would that access look like if we pass a lambda? E.g. how can you access private members in the lambda function that gets defined by a user and that gets passed to IDFilter?

hendrikweisser: The predicate (e.g. lambda) would be applied to each data element (e.g. `IdentifiedPeptide`) and have access to its public members (which is usually all of them). The filter function would have privileged access to the container holding the elements, in order to remove those for which the predicate returns true (or false?).

hendrikweisser:
@hroest: You asked about time needed for writing .oms files. I've tested with the biggest idXML I could find, which is 493.6 MB ("mergedIDs_Fido_qVals_noFilter_fdrNew.idXML", which I think I got from Julianus at some point). Converting this to .oms using IDFileConverter took 5:30 minutes, of which roughly 5 minutes were writing the .oms file. (Not exactly fast, but this file is orders of magnitude larger than typical idXMLs.)
Peak memory usage was reported to be 1014 MB, and the output file size is 486.2 MB.

(I had to make a change in `IdentificationDataConverter` to get the data converted to `IdentificationData` internally. The problem is that it's a merge of ID runs from multiple files, but all in the same `IdentificationRun` in the idXML. This means there are duplicate spectrum references from different files (same scan numbers), which can lead to conflicting PSMs. You get errors like this when trying to `registerObservationMatch` with a conflicting argument:
`InvalidValue: the value '3' was used but is not valid; Trying to overwrite ObservationMatch charge with conflicting value.`
I had to catch those exceptions and skip the problematic PSMs, of which there were 441. Not sure if this is a change we want to keep.)

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
timosachsenberg: Ok so it Would be possible to query in a PSM filter lambda e.g. properties of the group the current considered PSM maps to? E.g. does the public interface of the ID datastructure makes this possible and not too inefficient?

timosachsenberg:
Thanks Hendrik for doing the experiment.

> @hroest: You asked about time needed for writing .oms files. I've tested with the biggest idXML I could find, which is 493.6 MB ("mergedIDs_Fido_qVals_noFilter_fdrNew.idXML", which I think I got from Julianus at some point). Converting this to .oms using IDFileConverter took 5:30 minutes, of which roughly 5 minutes were writing the .oms file. (Not exactly fast, but this file is orders of magnitude larger than typical idXMLs.) Peak memory usage was reported to be 1014 MB, and the output file size is 486.2 MB.

Peak memory usage is great, but storing a 482MB in 5:30 min sounds a bit slow (e.g., I regularly need to write out 2gb of idXMLs for a single file so I would this call not untypical - also instruments are recently getting really fast). Are there any obvious optimizations? I was expecting something in the range of 10s. I am also a bit puzzled that the file size is nearly as big as the idXML? I thought we previously had a significant reduction?
 
> (I had to make a change in `IdentificationDataConverter` to get the data converted to `IdentificationData` internally. The problem is that it's a merge of ID runs from multiple files, but all in the same `IdentificationRun` in the idXML. This means there are duplicate spectrum references from different files (same scan numbers), which can lead to conflicting PSMs. You get errors like this when trying to `registerObservationMatch` with a conflicting argument: `InvalidValue: the value '3' was used but is not valid; Trying to overwrite ObservationMatch charge with conflicting value.` I had to catch those exceptions and skip the problematic PSMs, of which there were 441. Not sure if this is a change we want to keep.)

Maybe they have a file_origin or similar meta value annotated which allows splitting.

hroest:
> Peak memory usage is great, but storing a 482MB in 5:30 min sounds a bit slow (e.g., I regularly need to write out 2gb of idXMLs for a single file so I would this call not untypical - also instruments are recently getting really fast). Are there any obvious optimizations? I was expecting something in the range of 10s. I am also a bit puzzled that the file size is nearly as big as the idXML? I thought we previously had a significant reduction?

As I pointed out, the SQL queries used here to write out the data may be quite inefficient. When you are using an SQL query for every single CV term it will take a long time since the query needs to be parsed and executed individually. I agree that this is a major issue with useability

@hendrikweisser how long does it take to read in the file again? How long are read/write time for idXML?

hroest:
> (I had to make a change in IdentificationDataConverter to get the data converted to IdentificationData internally. The problem is that it's a merge of ID runs from multiple files, but all in the same IdentificationRun in the idXML. This means there are duplicate spectrum references from different files (same scan numbers), which can lead to conflicting PSMs. You get errors like this when trying to registerObservationMatch with a conflicting argument:
InvalidValue: the value '3' was used but is not valid; Trying to overwrite ObservationMatch charge with conflicting value.
I had to catch those exceptions and skip the problematic PSMs, of which there were 441. Not sure if this is a change we want to keep.)

@hendrikweisser does this mean that we currently cannot read idXML and convert it to the new data structure in a lossless fashion? And we cannot convert idXML to the SQL format in a lossless fashion?

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hendrikweisser: From a PSM you can follow a link to the peptide and from there to its protein(s). The protein groups you'd have to query separately using the public interface of IdentificationData (`getParentGroupSets`). Depending on your use case, you may want to populate a data structure that maps from protein to protein group for more efficient queries in that direction.

hendrikweisser:
> Peak memory usage is great, but storing a 482MB in 5:30 min sounds a bit slow

We can all agree that optimisation is necessary, but can this wait until after a merge? We have functionality that works and (at this point) doesn't replace anything that already exists in OpenMS (with a small caveat to due to the internal changes in FeatureFinderIdentification). This branch and "develop" are already diverging again, and the time in can invest here is limited.

> does this mean that we currently cannot read idXML and convert it to the new data structure in a lossless fashion?

Yes. idXML is "underspecified" in the sense that you can have conflicting data stored in the same file. Since IdentificationData doesn't allow this, there are cases that cannot be converted "losslessly".

hendrikweisser:
> how long does it take to read in the file again?

That's quick. Converting the .oms back to idXML takes 52 seconds (44 for loading, 7 for writing). Peak memory usage: 1139 MB. File size of the output idXML is 395 MB (down from 471 MB).
According to FileInfo, the size reduction is due to collapsing of PSMs under common spectra after the conversion. I.e. in the input idXML, you get 5 PeptideIdentifications for the same spectrum, with 1 PeptideHit each, but in the output you get 1 PeptideIdentification with 5 PeptideHits. The overall number of peptide hits and protein hits stays the same.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
timosachsenberg: and this would be possible by passing a lambda?

hendrikweisser: Sure, just bind all the relevant variables that you need to access.

timosachsenberg: ok but can I bind them? e.g. if they are private in the ID Datastructure.

hendrikweisser: We're starting to turn in circles now. Of course you can't bind private members, but you don't need to. There are public, but immutable (`const`) accessors for all the data. (There's not much point storing data if you can't read it again.) The limitations only apply when you want to change stored data.

hendrikweisser:
@hroest: Are you sure SQLite supports batch execution of queries? I can find some websites mentioning that it doesn't (e.g. https://docs.microsoft.com/en-us/dotnet/standard/data/sqlite/batching), but nothing in the SQLite docs one way or the other.

timosachsenberg:
> @hroest: Are you sure SQLite supports batch execution of queries? I can find some websites mentioning that it doesn't (e.g. https://docs.microsoft.com/en-us/dotnet/standard/data/sqlite/batching), but nothing in the SQLite docs one way or the other.

He is probably refering to transactions https://medium.com/@JasonWyatt/squeezing-performance-from-sqlite-insertions-971aff98eef2

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
timosachsenberg: Ok. Let's still wait for Julianus (back soon) as the affected LFQ workflow would have the biggest impact on his side

hendrikweisser:
> He is probably refering to transactions

In theory, those should be disabled as I'm calling `PRAGMA journal_mode = OFF` in the c'tor of `OMSFileStore` (see https://www.sqlite.org/pragma.html#pragma_journal_mode).  Still, worth testing that this actually works as intended.

hendrikweisser:
> Converting this to .oms using IDFileConverter took 5:30 minutes, of which roughly 5 minutes were writing the .oms file.

I've just tested one little change, where I prepared the query for inserting DataValues (used for meta values) just once, instead of every time `storeDataValue_` is called. That already shaved almost 1:30 min. off the execution time! I guess the thing to do then is to prepare all queries in advance and store them. (Note that this has to be done after the respective tables are created, or the prepared statements won't work.)

hroest:
> That's quick. Converting the .oms back to idXML takes 52 seconds (44 for loading, 7 for writing). Peak memory usage: 1139 MB. File size of the output idXML is 395 MB (down from 471 MB).

So this means that the same file takes 7 seconds to write in idXML and 5 minutes to write in SQL? So this is roughly 40 times slower? I do think we should address this.

@hendrikweisser what does take some time is to run the interpreter which has to interpret string statements, so if you prepare statements that should be faster.

@timosachsenberg yes you can batch multiple statements together in one transaction / one execution of a long SQL statement which is faster: 

https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/MzMLSqliteHandler.cpp#L1188

basically what you can do is this:

```
INSERT INTO tbl_name
    (a,b,c)
VALUES
    (1,2,3),
    (4,5,6),
    (7,8,9);
```
which is more efficient than starting the interpreter for each row. Preparing statements should also help (eg run the interpreter only once). Or similarly efficient: 
```
INSERT INTO tbl_name
    (a,b,c) VALUES (1,2,3);
INSERT INTO tbl_name
    (a,b,c) VALUES (4,5,6);
INSERT INTO tbl_name
    (a,b,c) VALUES (7,8,9);
```
in a single string / single transaction. See also [1], this is supported since 2012.

1. https://stackoverflow.com/questions/1609637/is-it-possible-to-insert-multiple-rows-at-a-time-in-an-sqlite-database/1609688#1609688

hroest:
See also this: https://stackoverflow.com/questions/1711631/improve-insert-per-second-performance-of-sqlite

@hendrikweisser I think that `PRAGMA journal_mode = OFF` should be used together with transactions as the pragma affects to-disk synchronization but transactions affect whether sqlite has to take into account asynchronious read/writes *in memory* as well (you could have multiple threads connecting to your database in memory). So probably doing both is the best option. See also the linked stackoverflow post. I think what you do is currently the 85  inserts per second (no transaction, one operation per row or the "worst-case scenario") indicated in the post while maximum performance is 96k inserts per second, so there is a 1000-fold performance improvement here

hendrikweisser:
> but transactions affect whether sqlite has to take into account asynchronious read/writes in memory as well

That's already disabled as well: `PRAGMA synchronous = OFF`. There should be no need for using transactions.

hendrikweisser:
Consistently storing prepared queries (that were used more than once) reduced the IDFileConverter runtime by 1-1.5 minutes. So no further improvement to my earlier test, but it made the code a bit cleaner. I'm using a `map` keyed by table names (strings) to store the queries, so look-up times could be improved by switching to a vector or to individual variables.

Contrary to my intuition, adding explicit transactions *did* make quite a big difference: It reduced the runtime down to 1:10 min., i.e. by about three quarters. That is despite the database configuration (pragmas) making transactions a no-op (at least in my understanding).
In my limited tests it didn't matter whether I used one transaction around everything or multiple, finer-grained transactions. I guess the point is only to avoid the implicit transactions that SQLite would otherwise add.

I haven't tried switching to multi-row inserts, because a) it would require more extensive changes to the code, and b) I'm already using prepared statements, so there may not be much benefit. I think further optimisation should be guided by a profiler.

In any case it now takes about one minute to write the .oms file, which is still ~10x slower than idXML, but hopefully acceptable at this stage.

hendrikweisser:
Where are we now with this? @timosachsenberg, @hroest, @oliveralka: Are you satisfied? Are we waiting for feedback from @jpfeuffer?

timosachsenberg:
> Where are we now with this? @timosachsenberg, @hroest, @oliveralka: Are you satisfied? Are we waiting for feedback from @jpfeuffer?

Yes I think feedback from @jpfeuffer and at least a few runs on larger datasets.

hendrikweisser:
> at least a few runs on larger datasets

Who's going to do that?

timosachsenberg:
> > at least a few runs on larger datasets
> 
> Who's going to do that?

If we ask nicely maybe Julianus on his UPS benchmarking dataset.

hendrikweisser:
As discussed before Christmas, I've taken the changes to FeatureFinderIdentification (internal use of new ID data structures) and ProteomicsLFQ out of this PR. Everything in here should be fully backwards compatible now.

jpfeuffer:
```
pyopenms/pyopenms_8.cpp:83385:189: error: no matching function for call to 'OpenMS::IdentificationData::setMetaValue(boost::shared_ptr<OpenMS::String>::element_type&, boost::shared_ptr<OpenMS::DataValue>::element_type&)'
     __pyx_v_self->inst.get()->setMetaValue((*__pyx_f_8pyopenms_10pyopenms_1_convString(__pyx_v_in_0).get()), (*((struct __pyx_obj_8pyopenms_10pyopenms_6_DataValue *)__pyx_t_4)->inst.get()));
```

was this always the case in this PR or just after the "partly revert"?

hendrikweisser:
> was this always the case in this PR or just after the "partly revert"?

The "revert" didn't touch IdentificationData itself or pyOpenMS, so this must have been the case already.

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -191,110 +264,132 @@ namespace OpenMS
       db_search_params_(std::move(other.db_search_params_)),
       db_search_steps_(std::move(other.db_search_steps_)),
       score_types_(std::move(other.score_types_)),
-      data_queries_(std::move(other.data_queries_)),
-      parent_molecules_(std::move(other.parent_molecules_)),
-      parent_molecule_groupings_(std::move(other.parent_molecule_groupings_)),
+      observations_(std::move(other.observations_)),
```
timosachsenberg, line 267: is the metainfointerface also moved?

timosachsenberg: also ::swap could be missing it

jpfeuffer (approved):

jpfeuffer:
@timosachsenberg ready to merge from my side

timosachsenberg:
Move and swap could be incomplete. @hendrikweisser can you check

hendrikweisser:
> Move and swap could be incomplete. @hendrikweisser can you check

Well spotted, Timo.

In order to implement `swap`, I've added `MetaInfoInterface::swap`, which didn't exist before. Then I had to make slight changes in `MSChromatoram` to avoid ambiguities on `swap` calls.
The oversight should be fixed now.

jpfeuffer:
Looks good

timosachsenberg (approved):

Issue #5357, opened 2021-06-07
MzTabExporter not backwards compatible to old files anymore


jpfeuffer:
I am pretty sure it was at some point.
E.g. running it on https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/FeatureLinkerUnlabeled_4_output.consensusXML

fails with:

```
Map descriptions (file name + label) in ConsensusMap are not unique:
  file: UNKNOWN label:   file: UNKNOWN label:   file: UNKNOWN label:
No MS file associated (primary MS run path).
No fractions annotated in consensusXML. Assuming unfractionated.
Error: Unexpected internal error (PSM  does not map to an MS file registered in the quantitative metadata. Check your merging and filtering steps and/or report the issue, please.)
<No fractions annotated in consensusXML. Assuming unfractionated.> occurred 3 times
<No MS file associated (primary MS run path).> occurred 3 times
```

This actually breaks some pipelines since not all tools correctly annotate everything, yet. I thought we had fallbacks for that
@timosachsenberg

timosachsenberg:
maybe we did that on purpose? I think I recall that we at some point introduced that hard check and fixed the obvious bugs.
Do you know where this file is from? Maybe we just missed some annotation in a tool

jpfeuffer:
It affects almost all consensusXMLs from all the linkers.

jpfeuffer:
Whenever you have only "unknown" and label "" in the consensusheaders

jpfeuffer:
@timosachsenberg can you check the status of this quickly? was this fixed by the additional checks we added. E.g. validating whenever something gets added.

PR #5371, opened 2021-06-11
add missing meta data to FFMultiplex output

lars20070:
The PR adds previously missing meta data (primary MS run path) to the `featureXML` output of the `FeatureFinderMultiplex` TOPP tool.


src/topp/FeatureFinderMultiplex.cpp
```diff
@@ -259,20 +258,24 @@ class TOPPFeatureFinderMultiplex :
     // write feature map, consensus maps and blacklist
     if (!(out_.empty()))
     {
-      writeFeatureMap_(out_, algorithm.getFeatureMap());
```
timosachsenberg, line 262: Don’t copy. Take reference if possible. Also below

src/topp/FeatureFinderMultiplex.cpp
```diff
@@ -259,20 +258,24 @@ class TOPPFeatureFinderMultiplex :
...
+      FeatureMap feature_map = algorithm.getFeatureMap();
```
timosachsenberg, line 261:
```suggestion
      FeatureMap& feature_map = algorithm.getFeatureMap();
```

src/topp/FeatureFinderMultiplex.cpp
```diff
@@ -259,20 +258,24 @@ class TOPPFeatureFinderMultiplex :
...
+      feature_map.setPrimaryMSRunPath({in_}, exp);
+      writeFeatureMap_(out_, feature_map);
     }
     if (!(out_multiplets_.empty()))
     {
-      writeConsensusMap_(out_multiplets_, algorithm.getConsensusMap());
+      ConsensusMap consensus_map = algorithm.getConsensusMap();
```
timosachsenberg, line 267:
```suggestion
      ConsensusMap& consensus_map = algorithm.getConsensusMap();
```

src/topp/FeatureFinderMultiplex.cpp
```diff
@@ -259,20 +258,24 @@ class TOPPFeatureFinderMultiplex :
...
-      writeFeatureMap_(out_, algorithm.getFeatureMap());
+      FeatureMap& feature_map = algorithm.getFeatureMap();
+      feature_map.setPrimaryMSRunPath({in_}, exp);
+      writeFeatureMap_(out_, feature_map);
     }
     if (!(out_multiplets_.empty()))
     {
-      writeConsensusMap_(out_multiplets_, algorithm.getConsensusMap());
+      ConsensusMap& consensus_map = algorithm.getConsensusMap();
+      consensus_map.setPrimaryMSRunPath({in_}, exp);
```
timosachsenberg, line 268: @lars20070 here you need to set the filename for every column in the consensus map. e.g., for standard silac you would set {in, in} for triple silac {in, in, in}.

lars20070: Ok. See latest changes.

timosachsenberg:
Please don’t use boost here, we try to remove a bit our dependency on it. String should have appropriate functions.
Also please move the determination of the number of channels that to a e.g. static member of your class

src/topp/FeatureFinderMultiplex.cpp
```diff
@@ -209,6 +209,28 @@ class TOPPFeatureFinderMultiplex :
     file.store(filename, blacklist);
   }
 
+  /**
+   * @brief determine the number of samples
+   * for example n=2 for SILAC, or n=1 for simple feature detection
+   *
+   * @param labels    string describing the labels
+   */
+  static size_t number_of_samples(String labels)
```
timosachsenberg, line 218: We don’t use _ here see default style

lars20070:
TOPP tests fail for example at this position. Not sure why.
```
<UserParam type="stringList" name="spectra_data" value="[!
	                                                         FeatureFinderMultiplex_2_input.mzML]"/>
```

timosachsenberg:
> TOPP tests fail for example at this position. Not sure why.
> 
> ```
> <UserParam type="stringList" name="spectra_data" value="[!
> 	                                                         FeatureFinderMultiplex_2_input.mzML]"/>
> ```

Isn’t that what you now added? The reference to the ms run?

lars20070:
Yes, it is. - I generated all test output with `-test`. Don't understand why the test fails and especially not why there.

src/tests/topp/FeatureFinderMultiplex_1_output.featureXML
```diff
@@ -1,172 +1,173 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
 <featureMap version="1.9" id="fm_7804704400743266335" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/OpenMS/OpenMS/develop/share/OpenMS/SCHEMAS/FeatureXML_1_9.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+	<UserParam type="stringList" name="spectra_data" value="[FeatureFinderMultiplex_1_input.mzML]"/>
```
lars20070, line 3: TOPP tests fail here. Does Jenkins use the absolute file path instead of the file name?

lars20070:
Juste checked that this PR fixes #5357. :+1:

jpfeuffer:
> Yes, it is. - I generated all test output with `-test`. Don't understand why the test fails and especially not why there.

@timosachsenberg Does the setPrimaryFilepath function even know about "-test"? Can one decide between absolute/relative/basename?
If not we usually just whitelist the spectraData metavalue in the FuzzyDiff command in the tests/topp/Cmakelists.txt

timosachsenberg:
> > Yes, it is. - I generated all test output with `-test`. Don't understand why the test fails and especially not why there.
> 
> @timosachsenberg Does the setPrimaryFilepath function even know about "-test"? Can one decide between absolute/relative/basename?
> If not we usually just whitelist the spectraData metavalue in the FuzzyDiff command in the tests/topp/Cmakelists.txt

No you would need to test for it in the tool and set a dummy path. Or just whitelist as we usually do.

lars20070:
How can/shall I fix the test? Whitelisting?

src/tests/topp/CMakeLists.txt
```diff
@@ -320,66 +320,66 @@ set_tests_properties("TOPP_FeatureFinderMRM_1_out1" PROPERTIES DEPENDS "TOPP_Fea
 #------------------------------------------------------------------------------
 # FeatureFinderMultiplex test
 add_test("TOPP_FeatureFinderMultiplex_1" ${TOPP_BIN_PATH}/FeatureFinderMultiplex -test -in ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_input.mzML -ini ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_parameters.ini -out FeatureFinderMultiplex_1.tmp -out_multiplets FeatureFinderMultiplex_2.tmp)
-add_test("TOPP_FeatureFinderMultiplex_1_out1" ${DIFF} -whitelist "id=" "href=" -in1 FeatureFinderMultiplex_1.tmp -in2 ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_output.featureXML )
+add_test("TOPP_FeatureFinderMultiplex_1_out1" ${DIFF} -whitelist "id=" "href=" "value=" -in1 FeatureFinderMultiplex_1.tmp -in2 ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_output.featureXML )
```
timosachsenberg, line 323: Please don’t whitelist all values. spectra_data!

lars20070:
I assume the Windows `StopWatch` is unrelated.

src/tests/topp/CMakeLists.txt
```diff
@@ -320,66 +320,66 @@ set_tests_properties("TOPP_FeatureFinderMRM_1_out1" PROPERTIES DEPENDS "TOPP_Fea
...
 # FeatureFinderMultiplex test
 add_test("TOPP_FeatureFinderMultiplex_1" ${TOPP_BIN_PATH}/FeatureFinderMultiplex -test -in ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_input.mzML -ini ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_parameters.ini -out FeatureFinderMultiplex_1.tmp -out_multiplets FeatureFinderMultiplex_2.tmp)
-add_test("TOPP_FeatureFinderMultiplex_1_out1" ${DIFF} -whitelist "id=" "href=" "value=" -in1 FeatureFinderMultiplex_1.tmp -in2 ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_output.featureXML )
+add_test("TOPP_FeatureFinderMultiplex_1_out1" ${DIFF} -whitelist "id=" "href=" "spectra_data" -in1 FeatureFinderMultiplex_1.tmp -in2 ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_output.featureXML )
 set_tests_properties("TOPP_FeatureFinderMultiplex_1_out1" PROPERTIES DEPENDS "TOPP_FeatureFinderMultiplex_1")
-add_test("TOPP_FeatureFinderMultiplex_1_out2" ${DIFF} -whitelist "id=" "href=" "value=" -in1 FeatureFinderMultiplex_2.tmp -in2 ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_output.consensusXML )
+add_test("TOPP_FeatureFinderMultiplex_1_out2" ${DIFF} -whitelist "id=" "href=" -in1 FeatureFinderMultiplex_2.tmp -in2 ${DATA_DIR_TOPP}/FeatureFinderMultiplex_1_output.consensusXML )
```
timosachsenberg, line 325: Why is no whitelist required here?

lars20070: `spectra_data` does not appear in `conensusXML`.
So [this](https://github.com/OpenMS/OpenMS/pull/5371/files#diff-f9fc730d9fb6e75e490158f83e6dd8f92b5521e358f001eb5ff64f8e4e4fa9abR291) is not sufficient?

timosachsenberg: I think it gets annotated at the map elements. Just search for the file Name and you should find it.

lars20070: Ok. - Only in `featureXML` the full path instead of the file name appears during the tests. `-test` seems to have a different effect in each case.

PR #5376, opened 2021-06-15
[Feature] AccurateMassSearch sqlite support

jpfeuffer:
# Description

Use an sqlite database for on-the-fly queries of compound masses in AMS.

### DIFF
--- a/src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
+++ b/src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
@@ -46,6 +46,7 @@
 #include <numeric>
 
 
+using namespace std;
 
 namespace OpenMS
 {
@@ -280,6 +281,7 @@ namespace OpenMS
   found_adduct_(),
   empirical_formula_(),
   matching_hmdb_ids_(),
+  matching_names_(),
   mass_trace_intensities_(),
   isotopes_sim_score_(-1.0)
   {
@@ -306,6 +308,7 @@ namespace OpenMS
     found_adduct_(source.found_adduct_),
     empirical_formula_(source.empirical_formula_),
     matching_hmdb_ids_(source.matching_hmdb_ids_),
+    matching_names_(source.matching_names_),
     mass_trace_intensities_(source.mass_trace_intensities_),
     isotopes_sim_score_(source.isotopes_sim_score_)
   {
@@ -330,6 +333,7 @@ namespace OpenMS
     found_adduct_ = rhs.found_adduct_;
     empirical_formula_ = rhs.empirical_formula_;
     matching_hmdb_ids_ = rhs.matching_hmdb_ids_;
+    matching_names_ = rhs.matching_names_;
     mass_trace_intensities_ = rhs.mass_trace_intensities_;
     isotopes_sim_score_ = rhs.isotopes_sim_score_;
 
@@ -476,6 +480,16 @@ namespace OpenMS
     matching_hmdb_ids_ = match_ids;
   }
 
+  const std::vector<String>& AccurateMassSearchResult::getMatchingNames() const
+  {
+    return matching_names_;
+  }
+
+  void AccurateMassSearchResult::setMatchingNames(const std::vector<String>& match_names)
+  {
+    matching_names_ = match_names;
+  }
+
   const std::vector<double>& AccurateMassSearchResult::getMasstraceIntensities() const
   {
     return mass_trace_intensities_;
@@ -539,18 +553,17 @@ namespace OpenMS
     defaults_.setValue("isotopic_similarity", "false", "Computes a similarity score for each hit (only if the feature exhibits at least two isotopic mass traces).");
     defaults_.setValidStrings("isotopic_similarity", {"false", "true"});
 
-    defaults_.setValue("db:mapping", std::vector<std::string>{"CHEMISTRY/HMDBMappingFile.tsv"}, "Database input file(s), containing three tab-separated columns of mass, formula, identifier. "
+    //TODO should we really support multiple databases? Not sure
+    defaults_.setValue("db:mapping", std::vector<std::string>{"CHEMISTRY/HMDB.sqlite"}, "Database input file(s), containing three tab-separated columns of mass, formula, identifier. "
                                                                       "If 'mass' is 0, it is re-computed from the molecular sum formula. "
                                                                       "By default CHEMISTRY/HMDBMappingFile.tsv in OpenMS/share is used! If empty, the default will be used.");
     defaults_.setValue("db:struct", std::vector<std::string>{"CHEMISTRY/HMDB2StructMapping.tsv"}, "Database input file(s), containing four tab-separated columns of identifier, name, SMILES, INCHI."
                                                                         "The identifier should match with mapping file. SMILES and INCHI are reported in the output, but not used otherwise. "
                                                                         "By default CHEMISTRY/HMDB2StructMapping.tsv in OpenMS/share is used! If empty, the default will be used.");
-    defaults_.setValue("positive_adducts", "CHEMISTRY/PositiveAdducts.tsv", "This file contains the list of potential positive adducts that will be looked for in the database. "
-                                                                                 "Edit the list if you wish to exclude/include adducts. "
-                                                                                 "By default CHEMISTRY/PositiveAdducts.tsv in OpenMS/share is used.", {"advanced"});
-    defaults_.setValue("negative_adducts", "CHEMISTRY/NegativeAdducts.tsv", "This file contains the list of potential negative adducts that will be looked for in the database. "
-                                                                                 "Edit the list if you wish to exclude/include adducts. "
-                                                                                 "By default CHEMISTRY/NegativeAdducts.tsv in OpenMS/share is used.", {"advanced"});
+    defaults_.setValue("positive_adducts", std::vector<std::string>{"M+H;1+","M+Na;1+","M+K;1+","M+3CH3CN+2H;2+","M+3H;3+"}, "This is the list of potential positive adducts that will be looked for in the database. "
+                                                                                 "Edit the list if you wish to exclude/include adducts.", {"advanced"});
+    defaults_.setValue("negative_adducts", std::vector<std::string>{"M-H;1-","M+Cl;1-","M+Br;1-"}, "This is the list of potential negative adducts that will be looked for in the database. "
+                                                                                 "Edit the list if you wish to exclude/include adducts. ", {"advanced"});
 
     defaults_.setValue("use_feature_adducts", "false", "Whether to filter AMS candidates mismatching available feature adduct annotation.");
     defaults_.setValidStrings("use_feature_adducts", {"false", "true"});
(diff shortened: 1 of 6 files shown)


src/openms/source/FORMAT/FileTypes.cpp
```diff
@@ -113,6 +113,7 @@ namespace OpenMS
     TypeNameBinding(FileTypes::NOVOR, "novor", "Novor custom parameter file"),
     TypeNameBinding(FileTypes::XQUESTXML, "xquest.xml", "xquest.xml file"),
     TypeNameBinding(FileTypes::SPECXML, "spec.xml", "spec.xml file"),
+    TypeNameBinding(FileTypes::SQLITE, "sqlite", "sqlite database file"),
```
timosachsenberg, line 116: would sqAMS be better?

src/utils/AccurateMassSearch.cpp
```diff
@@ -107,13 +107,12 @@ class TOPPAccurateMassSearch :
     Param p = AccurateMassSearchEngine().getDefaults();
     registerTOPPSubsection_("db", "Database files which contain the identifications");
     registerInputFileList_("db:mapping", "<file(s)>", ListUtils::toStringList<std::string>(p.getValue("db:mapping")), p.getDescription("db:mapping"), true, false, {"skipexists"});
-    setValidFormats_("db:mapping", {"tsv"});
+    setValidFormats_("db:mapping", {"sqlite"});
```
timosachsenberg, line 110: is your plan to keep both tsv and sqlite? WIP?

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -303,6 +308,7 @@ namespace OpenMS
     found_adduct_(source.found_adduct_),
     empirical_formula_(source.empirical_formula_),
     matching_hmdb_ids_(source.matching_hmdb_ids_),
+    matching_names_(source.matching_names_),
```
timosachsenberg, line 311: can't those constructors be defaulted?

timosachsenberg:
nice. I also like the file size reduction compared to the tsv files.
This is a breaking change so we would need to find an way to transition for users or keep backwards compatibility.

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -536,18 +553,17 @@ namespace OpenMS
     defaults_.setValue("isotopic_similarity", "false", "Computes a similarity score for each hit (only if the feature exhibits at least two isotopic mass traces).");
     defaults_.setValidStrings("isotopic_similarity", {"false", "true"});
 
-    defaults_.setValue("db:mapping", std::vector<std::string>{"CHEMISTRY/HMDBMappingFile.tsv"}, "Database input file(s), containing three tab-separated columns of mass, formula, identifier. "
+    //TODO should we really support multiple databases? Not sure
```
oliveralka, line 556: Yeah, I think support for multiple databases would be nice. E.g. on could search for metabolites in HMDB and for additional lipids in Lipid Maps.

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -536,18 +553,17 @@ namespace OpenMS
...
+    defaults_.setValue("db:mapping", std::vector<std::string>{"CHEMISTRY/HMDB.sqlite"}, "Database input file(s), containing three tab-separated columns of mass, formula, identifier. "
```
oliveralka, line 557:
```suggestion
    defaults_.setValue("db:mapping", std::vector<std::string>{"CHEMISTRY/HMDB.sqlite"}, "Database input file(s), containing a compounds sqlite table with six columns of Mz, Id, Formula, Name, SMILES, INCHI. "
```

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -1277,13 +1277,19 @@ namespace OpenMS
...
+    if (db_mapping_file_.empty()) db_mapping_file_ = param_.getValue("db:mapping").toStringVector();
+    db_struct_file_ = param_.getValue("db:struct").toStringVector();
+    if (db_struct_file_.empty()) db_struct_file_ = param_.getValue("db:struct").toStringVector();
 
+    /*
     pos_adducts_fname_ = param_.getValue("positive_adducts").toString();
     neg_adducts_fname_ = param_.getValue("negative_adducts").toString();
+    */
+
+    //TODO with empty adducts, do we fall back to defaults, or do we really assume no adducts should be searched?
```
oliveralka, line 1290: I do not know any use case "no adducts" would be used, since usually most of the compounds are protonated. Nonetheless, I would probably allow it, but throw a warning, that no adducts have been specified.

oliveralka:
As far as I have understood the "struct.tsv" file is not necessary if the sqlite format is used - is that correct? 

The other thing is - would we/should we provide all common databases (Lipid Maps, PubChem, HMDB) in our specific (Mz, Id ...) sqlite format? How would that be kept up to date and where would that be stored.

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
timosachsenberg: does the table need a specific name?

jpfeuffer:
Yes the struct mapping should be removed. I have a Todo there because I am not sure yet if I save the additional data during the first query or make a second query based on ID. (One might need to require indexing on the ID column then. For the mzTab tests mainly.

Query time is currently comparable on the mini tests. Need to test further.

There exist tools to parse pubchem into sqlite. We could make a job to update it monthly on our archive. Others could be added in the future. Or users do it themselves (e.g. very easy with KNIME).

Yes backwardscompat. is also a Todo. At least allowing one csv. But not sure if I would go much further. Otherwise we have many input files again.

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -633,44 +648,41 @@ namespace OpenMS
...
-        // std::cout << "****************************************************" << std::endl;
+        entry.setObservedMZ(observed_mz);
+        entry.setCalculatedMZ(theoretical_mz);
+        entry.setCharge(std::abs(adductinfo_it->getCharge())); // use theoretical adducts charge (is always valid); native charge might be zero
+        entry.setMZErrorPPM(error_ppm_mz);
+        //TODO What to do about that index. Is this really needed?
+        //ams_result.setMatchingIndex(i);
+        entry.setFoundAdduct(adductinfo_it->getName());
+
+        results.insert(results.end(), std::make_move_iterator(entries.begin()),
```
cbielow, line 684: shouldn't that be outside the loop?

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -778,12 +790,14 @@ namespace OpenMS
...
     // This loads additional properties like common name, smiles, and inchi key for each HMDB id
-    parseStructMappingFile_(db_struct_file_);
+    //parseStructMappingFile_(db_struct_file_);
 
-    parseAdductsFile_(pos_adducts_fname_, pos_adducts_);
-    parseAdductsFile_(neg_adducts_fname_, neg_adducts_);
+    //parseAdductsFile_(pos_adducts_fname_, pos_adducts_);
+    //parseAdductsFile_(neg_adducts_fname_, neg_adducts_);
+    //TODO allow one or multiple?
+    sqlite_connector_ = std::make_unique<SqliteConnector>(db_mapping_file_[0], SqliteConnector::SqlOpenMode::READONLY);
```
cbielow, line 800: why use a unique_ptr here and not a plain `SqliteConnector` member?

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
cbielow: Agreed.
One can imagine that the spectrum was preprocessed to only contain actual compound masses (no adduct, charge 1) and then searching without adducts would be meaningful.

src/openms/source/FORMAT/FileTypes.cpp
cbielow: I agree. Otherwise you don't really know what is inside unless you inspect it.
Thats like using '.xml' for all mzXML and mzML files

src/utils/AccurateMassSearch.cpp
```diff
@@ -143,8 +142,8 @@ class TOPPAccurateMassSearch :
     // copy top-level params to algorithm
     ams_param.setValue("db:mapping", ListUtils::create<std::string>(getStringList_("db:mapping")));
     ams_param.setValue("db:struct", ListUtils::create<std::string>(getStringList_("db:struct")));
-    ams_param.setValue("positive_adducts", getStringOption_("positive_adducts"));
-    ams_param.setValue("negative_adducts", getStringOption_("negative_adducts"));
+    ams_param.setValue("positive_adducts", ListUtils::create<std::string>(getStringList_("positive_adducts")));
```
cbielow, line 145: pheww...
this conversion is not really elegant.
Why is there no overload?

cbielow (commented):
Nice changes in general, but the new sqlite format is a bit underspecified IMHO.
How would a new user find out how to create his own DB?
Some documentation and a DB schema would be nice, and ideally a conversion script which converts from TSV to sqlite??

What was the motivation to move to sql in the first place?

src/openms/source/FORMAT/FileTypes.cpp
jpfeuffer: Ok!

src/utils/AccurateMassSearch.cpp
jpfeuffer: yes WIP

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
jpfeuffer: yes I think so, did not check thoroughly yet.

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
jpfeuffer: exactly. I will put quotes around 'compounds' in your doc addition

src/utils/AccurateMassSearch.cpp
jpfeuffer: hahaha had a long discussion with @timosachsenberg 
I'll see what I can do

jpfeuffer:
The motivation was basically to search huge databases like Pubchem. And to reduce the number of input files.

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
timosachsenberg: yep... e.g., if these are unknowns

timosachsenberg:
abandoned?

jpfeuffer:
Yes, I tried to do it for KNIME and general metabolomics. But since no one really was interested in testing/using, I abandoned.

I think the idea still would be useful but I don't know the state of metabolomics (if people still do simple accurate mass searches against larger compound  databases or if AI for untargeted took over).

timosachsenberg:
I think it is a mixture of multiple things. In practice it is still accurate mass and spectral libraries. But also IM plays quite a big role...

PR #5383, opened 2021-06-22
Fix class test macro

cbielow:
# Description

The `ABORT_IF` macro does abort the test, but does not make the test fail.
This is very surprising behaviour and this PR fixes that.


cbielow:
can someone with a working glass-bowl tell me why this test https://github.com/OpenMS/OpenMS/blob/develop/src/tests/class_tests/openms/source/ClassTest_test.cpp#L163 fails now?
It seems the FAILURE_IS_SUCCESS macro is not expanded anymore. I have no clue why that is, based on the changes made in this PR...

cbielow:
rebuild jenkins

cbielow:
ready for merge.
The StopWatch test still fails, because the usertime is 0 -- which should not be, but I have no idea why.... (so this must be postponed to another day)

Issue #3681, opened 2018-08-15
Config file on Linux


hroest:
According to https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html the config file on Linux should be in `~/.config` and not  in `~/.OpenMS` - can we please change that? Also leads to less clutter.

poshul:
Fixed in #5385

PR #5385, opened 2021-06-23
[fix] updated the config storage path for linux

poshul:
# Description
addresses https://github.com/OpenMS/OpenMS/issues/3681

# Checklist:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,5 @@
+
+
 Glossary:
 OpenMS - Name of the project and our C++ library
 TOPP - "The OpenMS Proteomics Pipeline", collection of chainable tools for flexible HPLC/MS workflows
@@ -92,7 +94,7 @@ Various:
 - Some further clarifications, standardization/consolidation of outputs formats/consistency
 - Continued fixes/improvement of documentation
 - Removed deprecated version of PeakPickerHiRes (LowMemPeakPickerHiResRandomAccess and LowMemPeakPickerHiRes). Their functionality can be accessed through options of PeakPickerHiRes
-- moved the config storage location for unix installations to comply with XDG guidelines.
+- moved the config storage location for unix installations to comply with X Display Group (freedesktop.org) guidelines.
 
 
 ------------------------------------------------------------------------------------------
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -92,6 +92,7 @@ Various:
 - Some further clarifications, standardization/consolidation of outputs formats/consistency
 - Continued fixes/improvement of documentation
 - Removed deprecated version of PeakPickerHiRes (LowMemPeakPickerHiResRandomAccess and LowMemPeakPickerHiRes). Their functionality can be accessed through options of PeakPickerHiRes
+- moved the config storage location for unix installations to comply with XDG guidelines.
 
 
 ------------------------------------------------------------------------------------------
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -621,9 +621,9 @@ namespace OpenMS
   Param File::getSystemParameters()
   {
     String home_path = File::getOpenMSHomePath();
+    String filename;
     //Comply with https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html on unix identifying systems
     #ifdef __unix__
-      String filename;
       if(getenv("XDG_CONFIG_HOME"))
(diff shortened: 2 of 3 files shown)


jpfeuffer:
Can you add that to the changelog before we forget?

poshul:
rebuild jenkins

poshul:
The Win 10 test failure looks like a VM issue, and not specific to this PR

CHANGELOG
```diff
@@ -92,6 +92,7 @@ Various:
 - Some further clarifications, standardization/consolidation of outputs formats/consistency
 - Continued fixes/improvement of documentation
 - Removed deprecated version of PeakPickerHiRes (LowMemPeakPickerHiResRandomAccess and LowMemPeakPickerHiRes). Their functionality can be accessed through options of PeakPickerHiRes
+- moved the config storage location for unix installations to comply with XDG guidelines.
```
cbielow, line 95: can we spell out `XDG` here, because many people might not know it

src/openms/source/SYSTEM/UpdateCheck.cpp
```diff
@@ -73,26 +73,33 @@ namespace OpenMS
     platform = "Win";
 #elif __APPLE__
     platform = "Mac";
-#else
+#elif __unix__
```
cbielow, line 76: not sure about that part.
XDG mentions KDE, GNOME, XServer and alike, and this sounds like Linux to me (`__linux__`).
However, `__unix__` seems to be defined on our debian as well, which I find very suprising (linux != unix, right? :-) )

src/openms/source/SYSTEM/File.cpp
```diff
@@ -621,8 +621,12 @@ namespace OpenMS
   Param File::getSystemParameters()
   {
     String home_path = File::getOpenMSHomePath();
-
+    //Comply with https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html on unix identifying systems
+    #ifdef __unix__
+    String filename = home_path + "/.config/OpenMS/OpenMS.ini";
```
cbielow, line 626: technically, you'd need to query `$XDG_CONFIG_HOME` first and use that as a full prefix. 
I'm not sure though if its usually defined or if the fallback is used in most cases anyway....

From the docs:
```
$XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used. 
```

src/openms/source/SYSTEM/UpdateCheck.cpp
poshul: I've added a check for __linux__ first and then __unix__ if its not linux. According to http://web.archive.org/web/20191012035921/http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system all linux variants have __unix__ set.

src/openms/source/SYSTEM/UpdateCheck.cpp
```diff
@@ -73,26 +73,38 @@ namespace OpenMS
...
 
     // write to tmp + userid folder
 
     // e.g.: OpenMS_Default_Win_64_FeatureFinderCentroided_2.0.0
     String tool_version_string;
+    String config_path;
+    //Comply with https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html on unix identifying systems
+    #ifdef __unix__
+    if(getenv("XDG_CONFIG_HOME"))
+      config_path = String(getenv("XDG_CONFIG_HOME")) + "/.config/OpenMS";
```
cbielow, line 92: from how I read the docs, `$XDG_CONFIG_HOME` already includes `..config/`, or am I misreading that?

src/openms/source/SYSTEM/File.cpp
cbielow, line 626: should this also check for XDG_CONFIG_HOME ?

src/openms/source/SYSTEM/UpdateCheck.cpp
poshul: Reading the doc again, I believe you are correct. I'll change that.

cbielow (approved):
thanks! Looks awesome :)

timosachsenberg:
is this also relevant for https://github.com/OpenMS/OpenMS/blob/c2555c3278de34a5f261df9a0915f42a8bfada3c/src/openms_gui/source/VISUAL/APPLICATIONS/GUITOOLS/TOPPView.cpp#L94 ?

PR #5390, opened 2021-07-03
[FIX] Class tests memory leaks

dvdvgt:
# Description

This PR includes numerous memory leaks fixes among all [class tests](http://cdash.openms.de/viewTest.php?buildid=167463&filtercount=1&showfilters=1&field1=status&compare1=63&value1=failed). Most tests had memory leaks because of "global pointers" (at least global within the test scope)
```cpp
START_TEST(SomeClass, "$Id$")
//...
SomeClass* ptr;
SomeClass* nullpointer = nullptr

// Tests
// ... multiple reassignments of ptr
// missing delete for ptr
```
and missing deletes afterwards. I think this could easily be avoidable if all pointers were declared within a test section so that it is harder to lose track of allocated pointers and each section includes a `delete` for each `new`.

In some cases though, the memory leaks seem to be rooting from the tested classes themselves.

# Leaking Classes
Because most of these leaks revealed some underlying issues I refrained from fixing them in this PR as this already quite a big PR already. In my opinion it would be best to open a dedicated PR for them so that the needed changes can be reviewed more thoroughly.  

## libSVMEncoder
One class that particularly seems to be leaking memory is `libSVMEncoder` with functions like `LibSVMEncoder::encodeLibSVMProblem` and `LibSVMEncoder::loadLibSVMProblem`.

https://github.com/OpenMS/OpenMS/blob/7b74d2c002aaf14c5bbfcc430ded009dc32cfb85/src/openms/source/FORMAT/LibSVMEncoder.cpp#L299
https://github.com/OpenMS/OpenMS/blob/7b74d2c002aaf14c5bbfcc430ded009dc32cfb85/src/openms/source/FORMAT/LibSVMEncoder.cpp#L300

Who is responsible for the cleanup of these pointers? The `const vector<svm_node*>& vectors` argument or the newly created `svm_problem* problem`? 
https://github.com/OpenMS/OpenMS/blob/7b74d2c002aaf14c5bbfcc430ded009dc32cfb85/src/openms/source/FORMAT/LibSVMEncoder.cpp#L172

Some Class tests and functions affected are
- DetectabilitySimulation_test
- PrecursorIonSelectionPreprocessing_test
- PSLPFormulation_test
- SimpleSVM::setup
- (There might be many more. I quite frankly lost track a bit by the sheer size of tests and files.)

## IMSAlphabet
Opposed to the other leaking classes this issues was simple to fix which is why I included it in this PR.
The function ` IMSAlphabet::load(const std::string & fname, IMSAlphabetParser<> * parser)` leaked memory because it had a pointer to a `IMSAlphabetParser<>` as parameter. I changed the signature so that it takes a reference instead. I checked all usages of the function and the change of signature does not seem to cause any issues.

https://github.com/OpenMS/OpenMS/blob/7b74d2c002aaf14c5bbfcc430ded009dc32cfb85/src/openms/include/OpenMS/CHEMISTRY/MASSDECOMPOSITION/IMS/IMSAlphabet.h#L300

## UnimodXMLHandler
`UnimodXMLHandler::endElement` leaks memory 
https://github.com/OpenMS/OpenMS/blob/7b74d2c002aaf14c5bbfcc430ded009dc32cfb85/src/openms/source/FORMAT/HANDLERS/UnimodXMLHandler.cpp#L192

## PeptideHit
https://github.com/OpenMS/OpenMS/blob/7b74d2c002aaf14c5bbfcc430ded009dc32cfb85/src/openms/source/METADATA/PeptideHit.cpp#L94

# Thirdparty leaks
Some dependencies that seem to have memory leaks are
- xercesc_3_2::MemoryManagerImpl::allocate in `MzMLFile_test`
- libsqlite3 e.g. in [OSWFile_test](http://cdash.openms.de/testDetails.php?test=47905017&build=167463)
- Various Qt class as can be seen in [TOPPView_test](http://cdash.openms.de/testDetails.php?test=47905350&build=167463)

# Checklist:


src/tests/class_tests/openms/source/IMSAlphabet_test.cpp
```diff
@@ -304,7 +304,8 @@ START_SECTION((virtual void load(const std::string &fname, IMSAlphabetParser<> *
   of << "nitrogen\t14.0" << std::endl;
   of.close();
 
-  load_copy.load(filename, new IMSAlphabetTextParser);
+  IMSAlphabetTextParser parser;
```
cbielow, line 307: nitpick: the section name should reflect the new function signature

src/tests/class_tests/openms/source/IdentificationData_test.cpp
```diff
@@ -50,13 +50,17 @@ START_TEST(IdentificationData, "$Id$")
...
-IdentificationData* null = 0;
+IdentificationData* ptr = nullptr;
+IdentificationData* null = nullptr;
 START_SECTION((IdentificationData()))
   ptr = new IdentificationData();
   TEST_NOT_EQUAL(ptr, null);
 END_SECTION
 
+START_SECTION((~IdentificationData()))
+  delete ptr;
```
cbielow, line 61: this will give a warning, since there is nothing to test here.
Either omit the section or add `NOT_TESTABLE`

cbielow:
thanks for that certainly very painful PR!

On a sidenote: can anyone remember WHY our tests actually do the new/delete in the first place? I can only recall we had a bug which was detected once we introduced new/delete (by accident), and from that time on it was put into every class test.
But I fail to remember why WHY (and in the meantime, that reason might be gone or there is an alternative solution),

src/tests/class_tests/openms/source/FlagSet_test.cpp
```diff
@@ -87,6 +87,10 @@ START_SECTION(FlagSet())
 }
 END_SECTION
 
+START_SECTION(~FlagSet())
```
jpfeuffer, line 90: Not_testable?

src/tests/class_tests/openms/source/ProteinHit_test.cpp
```diff
@@ -67,7 +67,6 @@ START_SECTION(ProteinHit())
 END_SECTION
 
 START_SECTION(~ProteinHit())
-	ptr = new ProteinHit();
```
jpfeuffer, line 70: No tests?

jpfeuffer (commented):
Thanks a lot!

timosachsenberg:
> thanks for that certainly very painful PR!
> 
> On a sidenote: can anyone remember WHY our tests actually do the new/delete in the first place? I can only recall we had a bug which was detected once we introduced new/delete (by accident), and from that time on it was put into every class test.
> But I fail to remember the `why` (and in the meantime, that reason might be gone or there is an alternative solution),

I think one reason is this script: https://github.com/OpenMS/OpenMS/blob/develop/tools/create_test.php
and another one copypasta

src/tests/class_tests/openms/source/FlagSet_test.cpp
timosachsenberg: I think in general, we just call the destructor in these tests which allows e.g., ASan to catch problems with missing virtual destructors etc.

src/tests/class_tests/openms/source/IdentificationData_test.cpp
timosachsenberg: is omitting a good idea? I though ASan could detect e.g., missing virtual destructors

timosachsenberg:
let's merge and check cdash

PR #5392, opened 2021-07-05
Feature/mzqc export

axelwalter:
# Description

Adding more QC metrics for mzQC export.

# Checklist:

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/QcMLFile.h
+++ b/src/openms/include/OpenMS/FORMAT/QcMLFile.h
@@ -37,17 +37,14 @@
 #include <OpenMS/CONCEPT/ProgressLogger.h>
 #include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
 #include <OpenMS/FORMAT/XMLFile.h>
-#include <OpenMS/FORMAT/ControlledVocabulary.h>
 #include <OpenMS/KERNEL/MSExperiment.h>
-#include <OpenMS/KERNEL/ConsensusMap.h>
-
 #include <vector>
-#include <map>
-#include <set>
-#include <algorithm>
 
 namespace OpenMS
 {
+  class ConsensusMap;
+  class FeatureMap;
+
   /**
       @brief File adapter for QcML files used to load and store QcML files
 
--- a/src/openms/include/OpenMS/FORMAT/QcMLFile.h
+++ b/src/openms/include/OpenMS/FORMAT/QcMLFile.h
@@ -28,7 +28,7 @@
 // ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 // --------------------------------------------------------------------------
-// $Maintainer: Mathias Walzer $
+// $Maintainer: Mathias Walzer, Axel Walter $
 // $Authors: Mathias Walzer $
 // --------------------------------------------------------------------------
 
@@ -39,8 +39,7 @@
 #include <OpenMS/FORMAT/XMLFile.h>
 #include <OpenMS/FORMAT/ControlledVocabulary.h>
 #include <OpenMS/KERNEL/MSExperiment.h>
-
-
+#include <OpenMS/KERNEL/ConsensusMap.h>
 
 #include <vector>
 #include <map>
@@ -173,16 +172,21 @@ public:
     ///Calculation and collection of QC data
     /**
       @brief Collects QC data in qualityParameters and qualityAttachments
-      @param inputfile_id identification file
-      @param inputfile_feature feature file
-      @param inputfile_consensus consensus file
+      @param prot_ids protein identifications from ID file (idXML)
+      @param pep_ids peptide identifications from ID file (idXML)
+      @param feature_map FeatureMap from feature file (featureXML)
+      @param consensus_map ConsensusMap from consensus file (consensusXML)
       @param inputfile_raw mzML input file name
       @param remove_duplicate_features removes duplicates in a set of merged features
       @param exp MSExperiment to extract QC data from, prior sortSpectra() and updateRanges() required
     */
-    void collectQCData(const String& inputfile_id, const String& inputfile_feature,
-                       const String& inputfile_consensus, const String& inputfile_raw,
-                       const bool remove_duplicate_features, const MSExperiment& exp);
+    void collectQCData(std::vector<ProteinIdentification>& prot_ids,
+                       std::vector<PeptideIdentification>& pep_ids,
+                       const FeatureMap& feature_map,
+                       const ConsensusMap& consensus_map,
+                       const String& inputfile_raw,
+                       const bool remove_duplicate_features,
+                       const MSExperiment& exp);
     ///Store the QCFile
     /**
       @brief Store the qcML file
--- a/src/openms/include/OpenMS/FORMAT/QcMLFile.h
+++ b/src/openms/include/OpenMS/FORMAT/QcMLFile.h
@@ -169,7 +169,7 @@ public:
     ///Calculation and collection of QC data
     /**
       @brief Collects QC data in qualityParameters and qualityAttachments
-      @param prot_ids protein identifications from ID file (idXML)
+      @param prot_ids protein identifications from ID file
       @param pep_ids peptide identifications
       @param feature_map FeatureMap from feature file (featureXML)
       @param consensus_map ConsensusMap from consensus file (consensusXML)
--- a/src/openms/include/OpenMS/FORMAT/QcMLFile.h
+++ b/src/openms/include/OpenMS/FORMAT/QcMLFile.h
@@ -170,7 +170,7 @@ public:
     /**
       @brief Collects QC data in qualityParameters and qualityAttachments
       @param prot_ids protein identifications from ID file (idXML)
-      @param pep_ids peptide identifications from ID file (idXML)
+      @param pep_ids peptide identifications
       @param feature_map FeatureMap from feature file (featureXML)
       @param consensus_map ConsensusMap from consensus file (consensusXML)
       @param inputfile_raw mzML input file name
@@ -229,4 +229,3 @@ protected:
   };
 
 } // namespace OpenMS
-
--- a/src/openms/include/OpenMS/KERNEL/MSExperiment.h
+++ b/src/openms/include/OpenMS/KERNEL/MSExperiment.h
@@ -537,15 +537,16 @@ public:
     //@}
 
     /**
-    @brief Compute Total Ion Count per MS spectrum and applies the resampling algorithm, if a bin size in RT seconds greater than 0 is given.
+    @brief Computes the total ion chromatogram (TIC) for a given MS level (use ms_level = 0 for all levels). 
 
     By default, each MS spectrum's intensity just gets summed up. Regular RT bins can be obtained by specifying @p rt_bin_size.
+    If a bin size in RT seconds greater than 0 is given resampling is used.
 
-    @param bin_size RT bin size in seconds
-    @param ms_level MS level of spectra for calculation
+    @param bin_size RT bin size in seconds (0 = no resampling)
+    @param ms_level MS level of spectra for calculation (0 = all levels)
     @return TIC Chromatogram
     **/
-    const MSChromatogram getTIC(float rt_bin_size = 0, UInt ms_level = 1) const;
+    const MSChromatogram calculateTIC(float rt_bin_size = 0, UInt ms_level = 1) const;
 
     /**
       @brief Clears all data and meta data
--- a/src/openms/include/OpenMS/KERNEL/MSExperiment.h
+++ b/src/openms/include/OpenMS/KERNEL/MSExperiment.h
@@ -537,14 +537,15 @@ public:
     //@}
 
     /**
-    @brief Compute Total Ion Count per MS1 spectrum and applies the resampling algorithm, if a bin size in RT seconds greater than 0 is given.
+    @brief Compute Total Ion Count per MS spectrum and applies the resampling algorithm, if a bin size in RT seconds greater than 0 is given.
 
-    By default, each MS1 spectrum's intensity just gets summed up. Regular RT bins can be obtained by specifying @p rt_bin_size.
+    By default, each MS spectrum's intensity just gets summed up. Regular RT bins can be obtained by specifying @p rt_bin_size.
 
     @param bin_size RT bin size in seconds
+    @param ms_level MS level of spectra for calculation
     @return TIC Chromatogram
     **/
-    const MSChromatogram getTIC(float rt_bin_size = 0, int ms_level = 1) const;
+    const MSChromatogram getTIC(float rt_bin_size = 0, UInt ms_level = 1) const;
 
     /**
       @brief Clears all data and meta data
--- a/src/openms/include/OpenMS/KERNEL/MSExperiment.h
+++ b/src/openms/include/OpenMS/KERNEL/MSExperiment.h
@@ -544,7 +544,7 @@ public:
     @param bin_size RT bin size in seconds
     @return TIC Chromatogram
     **/
-    const MSChromatogram getTIC(float rt_bin_size = 0) const;
+    const MSChromatogram getTIC(float rt_bin_size = 0, int ms_level = 1) const;
 
     /**
       @brief Clears all data and meta data
--- a/src/openms/include/OpenMS/QC/Features.h
+++ b/src/openms/include/OpenMS/QC/Features.h
@@ -32,67 +32,52 @@
 // $Authors: Axel Walter $
 // --------------------------------------------------------------------------
 
+#pragma once
 
-#include <OpenMS/QC/DetectedCompounds.h>
-#include <OpenMS/FORMAT/FeatureXMLFile.h>
-#include <OpenMS/KERNEL/FeatureMap.h>
+#include <OpenMS/QC/QCBase.h>
 
-using namespace std;
+/**
+ * @brief Detected Compounds as a Metabolomics QC metric
+ *
+ * Simple class to return the number of detected compounds
+ * from a given library of target compounds in a specific run.
+ *
+ */
 
 namespace OpenMS
-{ 
-
-  DetectedCompounds::Result DetectedCompounds::compute(const String& inputfile_feature)
+{
+  class OPENMS_DLLAPI Features : public QCBase
   {
-    DetectedCompounds::Result result;
-    FeatureMap map;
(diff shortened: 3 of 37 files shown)


src/openms/include/OpenMS/QC/Features.h
```diff
@@ -0,0 +1,83 @@
...
+ * @brief Detected Compounds as a Metabolomics QC metric
+ *
+ * Simple class to return the number of detected compounds
+ * from a given library of target compounds in a specific run.
+ *
+ */
+
+namespace OpenMS
+{
+  class OPENMS_DLLAPI Features : public QCBase
```
cbielow, line 49: this name is much too close to `OpenMS::Feature` and will create confusion.
Can you think of something more specific? `CompoundCount`?

src/openms/include/OpenMS/QC/Features.h
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+     /**
+    @brief computes the number of detected compounds in a featureXML file
+
+    @param inputfile_feature featureXML file created by FeatureFinderMetaboIdent based on a given library of target compounds
+    @return number of detected compounds and their mean absolute retention time shift
+
+    **/
+    Result compute(const String& inputfile_feature);
```
cbielow, line 74: this is probably not what most people want (imagine I already have a featuremap. Do you force me to store it first?)
Why not pass a FeatureMap?

src/openms/include/OpenMS/QC/Identifications.h
```diff
@@ -0,0 +1,87 @@
...
+ * @brief Detected Proteins/Peptides as a Proteomics QC metric
+ *
+ * Simple class to return the number of detected proteins/peptides
+ * from a given idXML file.
+ *
+ */
+
+namespace OpenMS
+{
+  class OPENMS_DLLAPI Identifications : public QCBase
```
cbielow, line 49: `IdentificationSummary`?

src/openms/include/OpenMS/QC/Identifications.h
```diff
@@ -0,0 +1,87 @@
...
+  {
+  public:
+    /// Constructor
+    Identifications() = default;
+
+    /// Destructor
+    virtual ~Identifications() = default;
+
+    // stores DetectedCompounds values calculated by compute function
+    struct OPENMS_DLLAPI Result
```
cbielow, line 59: please annotate the members.

src/openms/include/OpenMS/QC/Identifications.h
```diff
@@ -0,0 +1,87 @@
...
+    {
+      UInt peptide_spectrum_matches = 0;
+      std::tuple<UInt, float> unique_peptides = {0, 0.0};
```
cbielow, line 62: What is the `float` telling me?

src/openms/include/OpenMS/QC/Identifications.h
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+     /**
+    @brief computes the number of detected compounds in a featureXML file
+
+    @param inputfile_id idXML file 
+    @return ??
+
+    **/
+    Result compute(const String& inputfile_id);
```
cbielow, line 78: same here. Use a structure. Maybe even a template, so people can pass in featureMaps or consensusMaps as well?

src/openms/source/QC/Features.cpp
```diff
@@ -0,0 +1,98 @@
...
+    FeatureXMLFile f;
+    f.load(inputfile_feature, map);
+    float sum_rt_deviations = 0;
+    UInt rt_count = 0;
+    map.updateRanges();
+    result.detected_compounds = map.size();
+    for (const auto& f : map)
+    {
+      // if feature has peak_apex_position, get the meassured rt 
+      if (f.getSubordinates()[0].metaValueExists("peak_apex_position"))
```
cbielow, line 58: segfaults if no subordinates

src/openms/source/QC/Features.cpp
```diff
@@ -0,0 +1,98 @@
...
+      {
+        float rt_meassured = f.getSubordinates()[0].getMetaValue("peak_apex_position");
```
cbielow, line 60: not sure which FF computes this, but (loosely) by convention, the f.getRT() should be the apex anyway, right?

src/openms/include/OpenMS/QC/Features.h
axelwalter: thanks for the review!
the idea is to collect a bunch of qc metrics from a featureXML file and, similar, from an idXML file. To load these files only once I created a QC class for each. Maybe call them FeatureSummary and IdentfificationSummary like you suggested below?

src/openms/source/QC/Features.cpp
```diff
@@ -0,0 +1,98 @@
...
+        // if feature has native id, get substring with theoretical rt, convert to float
+        // and add absolute rt deviation for this feature to sum_rt_deviations, increment rt_count
+        if (f.getSubordinates()[0].metaValueExists("native_id"))
+        {
+          String native_id = f.getSubordinates()[0].getMetaValue("native_id");
```
cbielow, line 65: this looks very brittle and depends on how the nativeID actually looks like.

Did you have a look at `RTAlignment`? It does something very similar.

src/openms/include/OpenMS/QC/Features.h
axelwalter: for the QCCalculator people will pass an featureXML file, not a FeatureMap. I could load the content of the featureXML file into a FeatureMap before passing it to the compute function. I guess the same applies to the idXML file?

src/openms/include/OpenMS/QC/Identifications.h
axelwalter: sounds good

src/openms/source/QC/Identifications.cpp
```diff
@@ -0,0 +1,125 @@
...
+    vector<PeptideIdentification> pep_ids;
+    IdXMLFile().load(inputfile_id, prot_ids, pep_ids);
+    set<String> peptides;
+    set<String> proteins;
+    // PSMs and collect unique peptides in set
+    for (const auto& pep_id : pep_ids)
+    {
+      if (!pep_id.empty())
+      {
+        result.peptide_spectrum_matches += pep_id.getHits().size();
```
cbielow, line 57: the number of PSMs should be the number of PepIDs, not PepHits (because only one of those can be true). If this metric really counts 'number of possible explanations', that would seem a bit random?

src/openms/source/QC/Identifications.cpp
```diff
@@ -0,0 +1,125 @@
...
+        const auto& temp_hits = pep_id.getHits();
+        for (const auto& temp_hit : temp_hits)
```
cbielow, line 59: personally, I'd say count only the best hit.

cbielow: because only of the sequences can be true.

src/openms/include/OpenMS/QC/Identifications.h
axelwalter: That is the applied FDR threshold as requested by the CV.
https://github.com/HUPO-PSI/mzQC/blob/master/cv/qc-cv.obo#L1347

src/openms/source/QC/Identifications.cpp
```diff
@@ -0,0 +1,125 @@
...
+        {
+          peptides.insert(temp_hit.getSequence().toString());
+        }
+      }
+    }
+    // get sum of all peptide length for mean calculation
+    // get missed cleavages, check if Lys or Arg in the middle of sequence
+    UInt missed_cleavages = 0;
+    UInt peptide_length_sum = 0;
+    for (const auto& pep : peptides)
```
cbielow, line 69: see MissedCleavages metric, which goes some way to determine the enzyme IIRC.

Your code assumes Trypsin was used (not necessarily true).

src/openms/source/QC/Identifications.cpp
```diff
@@ -0,0 +1,125 @@
...
+      }
+    }
+    result.peptide_length_mean = (double)peptide_length_sum / peptides.size();
+    result.missed_cleavages_mean = (double)missed_cleavages / peptides.size();
+    // collect unique proteins in sets and scores mean
+    double protein_hit_scores_sum = 0;
+    UInt protein_hit_count = 0;
+    for (const auto& prot_id : prot_ids)
+    {
+      const auto& temp_hits = prot_id.getHits();
```
cbielow, line 87: this is tricky, because if the pepIDs contain multiple peptidehits, you'd probably want to exclusively retain the proteins which are references in the best peptideHit and ignore the rest.
Currently this is a vast overestimation.

src/openms/source/QC/Identifications.cpp
```diff
@@ -0,0 +1,125 @@
...
+      protein_hit_count += temp_hits.size();
+      for (const auto& temp_hit : temp_hits)
+      {
+        proteins.insert(temp_hit.getAccession());
+        protein_hit_scores_sum += temp_hit.getScore();
+      }
+    }
+    result.protein_hit_scores_mean = protein_hit_scores_sum/protein_hit_count;
+    // unique peptides and proteins with their significance threshhold (always the same in idXML file)
+    result.unique_peptides = {peptides.size(), pep_ids.front().getSignificanceThreshold()};
```
cbielow, line 97: this might not be an FDR???

src/openms/source/QC/TIC.cpp
```diff
@@ -54,6 +54,18 @@ namespace OpenMS
         result.retention_times.push_back(p.getRT());
       }
 
+      UInt max_int = *max_element(result.intensities.begin(), result.intensities.end());
+      for (const auto& i : result.intensities)
+      {
+        if (i == 0)
+        {
+          result.relative_intensities.push_back(0);
```
cbielow, line 62: dividing 0 by max_int should not be a problem, right?
(but maybe max_int should be checked for being non-zero?

src/openms/include/OpenMS/QC/Features.h
jpfeuffer: I think the tool should take the filenames but the internal algorithm class should be as selfcontained, reuseable and "chainable" as possible. Therefore I agree with @cbielow  it probably should take a reference to loaded data already.

src/openms/source/QC/Identifications.cpp
jpfeuffer: true, unfortunately this not really enforced. Most of the time this wont even be set since it is an ancient class member. I also don't think it is written out to idXML.
I did not think very long but I think you might need to recalculate the FDR to see which cutoff was applied.
Might be worthwhile to think about how to pass this through, though.

src/openms/include/OpenMS/QC/Features.h
axelwalter: that makes sense, will change that in the IdentificationSummary as well.

src/openms/include/OpenMS/QC/Identifications.h
timosachsenberg, line 62:
```suggestion
      std::tuple<UInt, float> unique_peptides = {0, -1.0}; // number of unique peptides and FDR threshold (-1.0 indicates no FDR threshold applied)
```

src/openms/include/OpenMS/QC/Identifications.h
timosachsenberg, line 49: IdentificationsSummary

src/openms/include/OpenMS/QC/TIC.h
```diff
@@ -81,10 +82,11 @@ namespace OpenMS
 
     @param exp Peak map to compute the MS1 tick from
     @param bin_size RT bin size in seconds
+    @param ms_level MS level of spectra for calculation
     @return result struct with with computed QC metrics: intensities, RTs (in seconds), area under TIC, 10x MS1 signal fall, 10x MS1 signal jump
 
     **/
-    Result compute(const MSExperiment& exp, float bin_size=0);
+    Result compute(const MSExperiment& exp, float bin_size=0, UInt ms_level=1);
```
timosachsenberg, line 89:
```suggestion
    Result compute(const MSExperiment& exp, float bin_size = 0, UInt ms_level = 1);
```

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -126,24 +140,74 @@ namespace OpenMS
...
+        chrom["Retention time"] = result.retention_times;
+        // MS1 Total ion current chromatogram
+        addMetric("QC:4000069", chrom);
         // MS1 signal jump (10x) count
         addMetric("QC:4000172", result.jump);
         // MS1 signal fall (10x) count
         addMetric("QC:4000173", result.fall);
       }
+      // MS2
+      result = tic.compute(exp,0,2);
```
timosachsenberg, line 173:
```suggestion
      result = tic.compute(exp, 0, 2);
```

src/openms/source/QC/Identifications.cpp
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  bool Identifications::Result::operator==(const Result& rhs) const
+  {
+    return peptide_spectrum_matches == rhs.peptide_spectrum_matches
+          && unique_peptides == rhs.unique_peptides
+          && unique_proteins == rhs.unique_proteins
+          && missed_cleavages_mean == rhs.missed_cleavages_mean
+          && protein_hit_scores_mean == rhs.protein_hit_scores_mean
+          && peptide_length_mean == rhs.peptide_length_mean;
```
timosachsenberg, line 109: min and max length are also important metrics

timosachsenberg (commented):
more in zoom

src/openms/include/OpenMS/KERNEL/MSExperiment.h
```diff
@@ -537,14 +537,15 @@ namespace OpenMS
...
 
-    By default, each MS1 spectrum's intensity just gets summed up. Regular RT bins can be obtained by specifying @p rt_bin_size.
+    By default, each MS spectrum's intensity just gets summed up. Regular RT bins can be obtained by specifying @p rt_bin_size.
 
     @param bin_size RT bin size in seconds
+    @param ms_level MS level of spectra for calculation
     @return TIC Chromatogram
     **/
-    const MSChromatogram getTIC(float rt_bin_size = 0) const;
+    const MSChromatogram getTIC(float rt_bin_size = 0, UInt ms_level = 1) const;
```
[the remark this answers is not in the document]
cbielow: sounds good.
Is there a CV Term or some kind of annotation for these chromatograms?
If yes, the above code could also look for them instead of recomputing the TIC ... and then the name would not be too bad.

src/openms/include/OpenMS/QC/Identifications.h
cbielow: [optional] maybe go for a small `struct`? then the member's name immediately tell the story.

axelwalter:
thanks for the useful reviews! Since peptide/protein identifications can also be in a featureXML I want to cover that as well.
Would you do that in FeatureSummary or perhaps it would be better to create a second compute function in IdentificationSummary where you pass a FeatureMap?
So the FeatureSummary be more like a Metabolomics Summary and the IdentificationSummary more like a Proteomics summary. Would that justify renaming the classes?

src/tests/topp/QCCalculator_2_output.mzQC
```diff
@@ -79,21 +79,39 @@
...
-                0,
-                120,
-                120
+                0.0,
+                100.0,
+                100.0,
+                91.66666412353516
               ],
               "Retention time": [
                 -1.0,
```
timosachsenberg, line 88: is -1 a valid value?

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -54,14 +57,15 @@ using namespace std;
 
 namespace OpenMS
 { 
-
   void MzQCFile::store(const String& input_file,
                        const String& output_file,
                        const MSExperiment& exp,
```
timosachsenberg, line 62: why do we have input_file and exp? Shouldn't exp be enough?

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -54,14 +57,15 @@ using namespace std;
...
                        const String& contact_name,
                        const String& contact_address,
                        const String& description,
-                       const String& label) const
+                       const String& label,
+                       const String& inputfile_feature,
```
timosachsenberg, line 67: could we just pass a const&FeatureMap and adapt the code that we don't need to read the featuremap in the QcML class? That way we could use it for featuremaps that are in memory

src/openms/source/FORMAT/MzQCFile.cpp
```diff
@@ -54,14 +57,15 @@ using namespace std;
...
+                       const String& inputfile_id) const
```
timosachsenberg, line 68: same here. using the identifications would make it usable also for in memory data

src/tests/topp/QCCalculator_2_output.mzQC
axelwalter: in this mzML test file there is a RT of -1.0 yes

src/openms/source/FORMAT/MzQCFile.cpp
axelwalter: we need to store the input file location and name

src/openms/include/OpenMS/FORMAT/QcMLFile.h
```diff
@@ -173,16 +172,21 @@ namespace OpenMS
     ///Calculation and collection of QC data
     /**
       @brief Collects QC data in qualityParameters and qualityAttachments
-      @param inputfile_id identification file
-      @param inputfile_feature feature file
-      @param inputfile_consensus consensus file
+      @param prot_ids protein identifications from ID file (idXML)
```
timosachsenberg, line 175:
```suggestion
      @param prot_ids protein identifications from ID file
```

src/openms/include/OpenMS/FORMAT/QcMLFile.h
```diff
@@ -173,16 +172,21 @@ namespace OpenMS
...
+      @param pep_ids peptide identifications from ID file (idXML)
```
timosachsenberg, line 176:
```suggestion
      @param pep_ids peptide identifications
```

timosachsenberg (approved):

PR #5401, opened 2021-07-07
[FIX] Correctly setup c17 for pyopenms

jpfeuffer:
It had a weird logic before. Leading to non-c17 builds on Linux gcc before.


oliveralka (commented):
Looks good - let's see if it works out.

PR #5417, opened 2021-07-14
MSFragger remove unused parameter

timosachsenberg:
# Description

# Checklist:

# Note:

### DIFF
--- a/src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
+++ b/src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
@@ -163,7 +163,7 @@ namespace OpenMS
       }
     }
 
-    void PercolatorFeatureSetHelper::addMSFRAGGERFeatures(vector<PeptideIdentification>& peptide_ids, StringList& feature_set)
+    void PercolatorFeatureSetHelper::addMSFRAGGERFeatures(StringList& feature_set)
     {
       feature_set.push_back("MS:1001330"); // expect_score
       feature_set.push_back("hyperscore");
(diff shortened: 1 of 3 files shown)


src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
```diff
@@ -163,7 +163,7 @@ namespace OpenMS
       }
     }
 
-    void PercolatorFeatureSetHelper::addMSFRAGGERFeatures(vector<PeptideIdentification>& peptide_ids, StringList& feature_set)
+    void PercolatorFeatureSetHelper::addMSFRAGGERFeatures(StringList& feature_set)
```
oliveralka, line 166: use "extra_features" as a variable name and change the ones below?

src/utils/PSMFeatureExtractor.cpp
```diff
@@ -257,11 +257,26 @@ class PSMFeatureExtractor :
...
+    {
+      PercolatorFeatureSetHelper::addXTANDEMFeatures(all_peptide_ids, feature_set);
+    }
+    else if (search_engine == "Comet") 
+    {
+      PercolatorFeatureSetHelper::addCOMETFeatures(all_peptide_ids, feature_set);
+    }
+    else if (search_engine == "MSFragger") 
+    {
+      PercolatorFeatureSetHelper::addMSFRAGGERFeatures(feature_set);
```
oliveralka, line 278: similar to the variable name above?

oliveralka (changes requested):
Looks good, two minor commets.

src/openms/source/ANALYSIS/ID/PercolatorFeatureSetHelper.cpp
timosachsenberg: these are not extra features like in the other case but the ones directly returned by MSFragger

PR #5441, opened 2021-07-19
[Add] Additional Python Wrapper (Release 2.7.0)

oliveralka:
# Description

As discussed, we would like to add additional wrappers with examples for Release 2.7.0

BasicProteinInferenceAlgorithm.pxd
BayesianProteinInferenceAlgorithm.pxd
MzQCFile.pxd
FeatureFinderIdentificationAlgorithm.pxd

# Checklist:

# Note:


oliveralka:
FeatureFinderIdentificationAlgorithm.pxd
`ProgressLogger& getProgressLogger() { return prog_log_; }`
`ProgressLogger getProgressLogger() nogil except + # wrap-ignore`
Is not wrapped, since it lead to a funky error:

```
Error compiling Cython file:
...
    
    def getProgressLogger(self):
        """Cython signature: ProgressLogger getProgressLogger()
        Return progress logger
"""
        cdef _ProgressLogger * _r = new _ProgressLogger(self.inst.get().getProgressLogger())
                                                      ^
```

jpfeuffer:
I don't think you can return references. You would need to return either a pointer or a copy. But I don't think anyone needs the ProgressLogger (since it's only for changing it). One can also just create a new one and overwrite the old. I don't even know if progress logging works reliably in python.

src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.h
```diff
@@ -82,6 +82,8 @@ class OPENMS_DLLAPI FeatureFinderIdentificationAlgorithm :
   PeakMap& getMSData() { return ms_data_; }
```
oliveralka, line 82: Should the definition of the getters and setter be moved to the source file?

timosachsenberg: yes please

poshul (approved):
Looks like a good positive addition of more wrappers.

timosachsenberg (approved):
Very nice addition and thanks for the fixes.

PR #5448, opened 2021-07-22
Try to add D and T to elements

jpfeuffer:
untested

### DIFF
--- a/src/openms/source/CHEMISTRY/ElementDB.cpp
+++ b/src/openms/source/CHEMISTRY/ElementDB.cpp
@@ -141,6 +141,16 @@ namespace OpenMS
 
   void ElementDB::storeElements_()
   {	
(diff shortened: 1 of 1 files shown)


timosachsenberg:
see **https://github.com/OpenMS/OpenMS/pull/5448/files#diff-8ebe022fac5a9f3c0eb06f940c6706d8d9bb652131218ee474d495451af8627dR154**
not sure if we should add it that way. Were do you need it? Parsing?

jpfeuffer:
Yes. A lot of DBs have D in chemical formulas.

timosachsenberg:
I would prefer the parser I think if it is easily feasible.

timosachsenberg:
But if it is well documented I guess it would be also fine to do it your way

timosachsenberg:
superseeded by https://github.com/OpenMS/OpenMS/pull/5453

PR #5455, opened 2021-07-23
Revert KNIME incompatible shortcut

jpfeuffer:
maybe should also go to release

### DIFF
--- a/src/topp/TextExporter.cpp
+++ b/src/topp/TextExporter.cpp
@@ -1303,7 +1303,7 @@ protected:
           throw Exception::InvalidParameter(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "'id:proteins_only' and 'id:peptides_only' cannot be used together");
         }
 
-        String what = peptides_only ? "" : "PEPTIDE";
+        String what = "PEPTIDE";
         if (!peptides_only)
         {
           writeRunHeader(output);


jpfeuffer:
I fixed it through incredibly complicated exception handling in KNIME. I still think it is not worth it to deviate from the schema only to save one column.
How should KNIME know which format to expect if we leave out the "section names". It should at least write it in the header.

poshul:
@jpfeuffer hows this looking now? Are the test errors downstream of the changes made?

jpfeuffer:
Yes every test needs to be changed. Including KNIME nodes to be tested again. And I did not really get feedback on this. And I would like to deprecate the textexporters.
In the end they are basically the same as mzTab (in my opinion even more complicated, but @timosachsenberg disagrees).

jpfeuffer:
Only the very basic exports of e.g. IDs are simpler.
But we do not even export proteins yet, let alone protein groups. Multiple runs will look very complicated, too.
The parsing of the hierarchical export of consensusXMLs especially with different runs etc. is even worse.

timosachsenberg:
We can deprecate them, but we should not remove them yet.

timosachsenberg:
some tests need an update I guess

poshul:
I'm fine with deprecating them. Does that solve the test issue?

jpfeuffer:
In that case we could close this PR and slowly replace occurrences in our tutorials.

timosachsenberg:
Or we keep the PR and update tests but still remove from the tutorials

PR #5464, opened 2021-07-28
fix bad words

axelwalter:
# Spell check

used Matteos tool to remove some bad words and misspellings


src/openms/source/CHEMISTRY/NASequence.cpp
```diff
@@ -174,7 +174,7 @@ namespace OpenMS
     our_form += (phosphate_form - internal_to_full) * (seq_.size() - 1);
     EmpiricalFormula local_three_prime, local_five_prime;
 
-    // Make local copies of the formulas for the terminal mods so we don't get into trouble dereferencing nullptrs
+    // Make local copies of the formulas for the terminal mods so we don't get into trouble dereferencing null ptrs
```
timosachsenberg, line 177: technically in c++ we write nullptr but I guess the fix is good

timosachsenberg (approved):
nice. Thanks!

oliveralka (approved):

PR #5451, opened 2021-07-22
Add missing elements

JangJang3:
# Description

Please include a summary of the change and which issue is fixed.

# Checklist:
- [x] Make sure that you are listed in the AUTHORS file
- [x] Add relevant changes and new features to the CHANGELOG file
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing unit tests pass locally with my changes
- [x] Updated or added python bindings for changed or new classes. (Tick if no updates were necessary.)

# How can I get additional information on failed tests during CI:
If your PR is failing you can check out 
- http://cdash.openms.de/index.php?project=OpenMS and look for your PR. If you click in the column that lists the failed tests you will get detailed error messages.

# Note:
- Once you opened a PR try to minimize the number of *pushes* to it as every push will trigger CI (automated builds and test) and is rather heavy on our infrastructure (e.g., if several pushes per day are performed).

PR #5466, opened 2021-07-28
Adding more Elements

Arraxx:
# Description

Added more elements with the reference below, excluding the elements present in PR #5451 

Reference:
https://pubchem.ncbi.nlm.nih.gov/
https://www.rsc.org/periodic-table/
https://www.ciaaw.org/

# Checklist:

# Note:

### DIFF
--- a/src/openms/source/CHEMISTRY/ElementDB.cpp
+++ b/src/openms/source/CHEMISTRY/ElementDB.cpp
@@ -556,8 +556,8 @@ namespace OpenMS
     buildElement_("Bismuth", "Bi", 83u, bismuth_abundance, bismuth_mass);
 
 
-    map<unsigned int, double> polonium_abundance = {{209u, 0.0}, {210u, 0.0}};
-    map<unsigned int, double> polonium_mass = {{209u, 208.9824308}, {210u, 209.9828741}};
+    map<unsigned int, double> polonium_abundance = {{210u, 0.99998}};
+    map<unsigned int, double> polonium_mass = {{210u, 209.9828741}};
     buildElement_("Polonium", "Po", 84u, polonium_abundance, polonium_mass);
 
 
@@ -566,12 +566,12 @@ namespace OpenMS
     buildElement_("Astatine", "At", 85u, astatine_abundance, astatine_mass);
 
   
-    map<unsigned int, double> radon_abundance = {{211u, 0.0}, {220u, 0.0}, {222u, 0.0}};
-    map<unsigned int, double> radon_mass = {{211u, 210.9906011}, {220u, 220.0113941}, {222u, 222.0175782}};
+    map<unsigned int, double> radon_abundance = {{220u, 0.09}, {222u, 0.900000001}};
+    map<unsigned int, double> radon_mass = {{220u, 220.0113941}, {222u, 222.0175782}};
     buildElement_("Radon", "Rn", 86u, radon_abundance, radon_mass);
 
 
-    map<unsigned int, double> francium_abundance = {{223u, 0.0}};
+    map<unsigned int, double> francium_abundance = {{223u, 0.0000001}};
     map<unsigned int, double> francium_mass = {{223u, 223.0197360}};
     buildElement_("Francium", "Fr", 87u, francium_abundance, francium_mass);
 
@@ -648,11 +648,6 @@ namespace OpenMS
     map<unsigned int, double> lawrencium_abundance = {{262u, 0.0000}};
     map<unsigned int, double> lawrencium_mass = {{262u, 262.10961}};
     buildElement_("Lawrencium", "Lr", 103u, lawrencium_abundance, lawrencium_mass);
-
-
-    map<unsigned int, double> rutherfordium_abundance = {{267u, 0.0}};
-    map<unsigned int, double> rutherfordium_mass = {{267u, 267.12179}};
-    buildElement_("Rutherfordium", "Rf", 104u, rutherfordium_abundance, rutherfordium_mass);
 
 
     map<unsigned int, double> rutherfordium_abundance = {{267u, 0.0}};
--- a/src/openms/source/CHEMISTRY/ElementDB.cpp
+++ b/src/openms/source/CHEMISTRY/ElementDB.cpp
@@ -441,6 +441,11 @@ namespace OpenMS
     buildElement_("Neodymium", "Nd", 60u, neodymium_abundance, neodymium_mass);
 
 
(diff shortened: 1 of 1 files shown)


poshul:
How are abundances handled for some of the heavier radioactive elements? Plutonium for example seems to have 0 abundances for all the isotopes. (Though if I ever run into organo-plutonium biomolecules I'd probably change jobs to something safer...)

Arraxx:
@poshul  as I was not able to find out the abundance of some elements that's why I added zero for them but on further searching i found that it can't be zero, so it will be great if you can suggest me something suitable for this.(or we can take any variable number)

jpfeuffer:
We have multiple options for non-natural elements:
- distribute equally
- use 1.0 at the first isotope
- fail/warn when its average weight is used

It does not matter much as long as it is documented well, since the use cases should be very rare. If someone uses them, they need to use the actual isotopes in their formulas anyway, since they most likely are looking to distinguish and quantify the different isotopes separately.

poshul:
I think we should leave out anything that is not naturally occurring. The isotope abundance is going to be dependent on how and where it was made. That applies to anything Transuranic, as well as Promethium. This leaves us with Po At Rn Fr and Ra, as naturally occurring elements that I cant find isotope abundance for. I'm open to suggestions as to handling these five.

Also I added Protactinium and Uranium.

jpfeuffer:
Polonium is here: https://www.chemie.de/lexikon/Polonium.html
You can check for others as well. Table on the right side (NH)

Arraxx:
rebuild jenkins

timosachsenberg:
looks good. does anyone know if relative abundacnes need to add up to 1.0 ? e.g., what if the sum deviates too much?

Arraxx:
any suggestions? @jpfeuffer @poshul

jpfeuffer:
@timosachsenberg most probably. How else would you calculate correct average masses? Unless we normalize anyway, but then it also yields wrong results if it was supposed to add up to 1 but doesn't.

jpfeuffer:
My preferred solution is failing/warning. So maybe we should remove/not add elements that do not have known abundances (although potentially monoisotopic mass could be used).
And document in the class documentation where we got the abundances from and what to do when an Element ist missing.

timosachsenberg:
I will check if we normalize after reading

timosachsenberg:
So what action should we implement?
In https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/ElementDB.cpp#L607
add up abundances and add an OPENMS_POSTCONDITION that sum of abundances is >= 0.95 ?

poshul:
I'd say setting an OPENMS_POSTCONDITION is the way to handle it, and just not include elements without abundances.

timosachsenberg:
@Arraxx can you take on that task?

Arraxx:
okay @timosachsenberg

timosachsenberg (approved):

timosachsenberg:
I think not all are addressed yet?

> So what action should we implement?
> In https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/ElementDB.cpp#L607
> 
> * add up abundances and add an OPENMS_POSTCONDITION that sum of abundances is >= 0.95 ?
> * document in class documentation where abundances are from
> * add a sentence to class doc "Note: if an element is missing, you can add it manually or open an issue on the OpenMS repository."

timosachsenberg:
@Arraxx do you plan to move forward here?

Arraxx:
Sure, I will try to complete it. @timosachsenberg

Arraxx:
@timosachsenberg , can you please explain it a little more to me.

jpfeuffer:
What is the status here? We have a usecase and it is getting urgent.

timosachsenberg:
I checked iridium and osmium. And it is safe to add them because the abundances add up to 1.0
For the other ones we are not sure if this has any negative impact (e.g., if total abundances are zero but we normalize internally through the sum of abundances)

timosachsenberg:
To get something actionable: I would suggest that we finish the PR with elements with abundances that add up to 1.0 and look into the others later.

jpfeuffer:
I would say we just fail if someone wants an average weight for this element.

jpfeuffer:
Creating compounds with a specific isotope would at least work.

timosachsenberg:
I think it is not only the average weight but a lot of other things (e.g., coarse isotope patterns)

jpfeuffer:
Then this fails, too. And the user is forced to provide a specific isotope.

timosachsenberg:
I removed the recently added ones that have zero abundance.
Still would be good to revisit them. E.g., adding them to enable e.g. monoisotopic mass calculation but failing if an unsupported operation is performed on them.

timosachsenberg:
rebuild jenkins

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -496,8 +497,8 @@ namespace OpenMS
     buildElement_("Hafnium", "Hf", 72u, hafnium_abundance, hafnium_mass);
 
 
-    map<unsigned int, double> tantalum_abundance = {{181u, 1.0}};
-    map<unsigned int, double> tantalum_mass = {{181u, 180.947995800000001}};
+    map<unsigned int, double> tantalum_abundance = {{180u, 0.0001176}, {181u, 1.0}};
```
hroest, line 500: these should add to 100%, no?

dlon450:
Hey, just wondering what factors are involved in the process for adding more (complicated) elements? I've got a list of missing elements that would help a lot for the tool I'm currently working on. I'm no chemistry expert but would be happy to provide further details/information and ask the chemists I'm working with how we can be of help. Thanks
[missing_elements.csv](https://github.com/OpenMS/OpenMS/files/8189879/missing_elements.csv)

hroest:
@dlon450 maybe we can have a discussion on discord, the main issues are conceptual as the current OpenMS only considers stable isotopes. We currently simply do not have proper modelling for non-stable isotopes that are not naturally occurring at a stable frequency. We would need to somehow incorporate radioactive isotopes, potentially with a modelling half lives which requires some discussion. 

Therefore, the process is a discussion with a proposal of proper modelling and then a PR

dlon450:
@hroest yes that would be great. My discord is
Derek!!#6699 
so maybe you can add me first and we can discuss this more there. Thanks

hroest:
I added you on Discord -- or just join us here at https://discord.gg/4TAGhqJ7s5

hroest:
I merged this with f3910ff6988b4910ced0787a964a8e48ced92

PR #5477, opened 2021-08-05
[FIX] PyOpenMS macOS 11.5.1 (10.16)

oliveralka:
# Description

Small fix for linking of PyOpenMS with system libs (now located in the SDK)
see https://developer.apple.com/forums/thread/655588?answerId=665804022#665804022

# Checklist:

# Note:


jpfeuffer:
Hmm shouldn't the isysroot take care of it? We are investigating. Do not merge yet

jpfeuffer:
Do you think this was needed in 10.14, too? Would it hurt?
We probably should switch to real version checking soon. The string comparison already fails for 10.9

oliveralka:
As far as I know, the library change to the SKD was introduced in Big Sur (10.16) see link in the PR description.

jpfeuffer:
To be honest, I think we were just lucky that only one SDK was installed all the time (or the most recent one was symlinked to /usr/lib). If the linker needs the sysroot, too to find the right libraries, then this was the case in earlier macOS', too. I would just try it with earlier macOS too. We have GitHub actions with 10.15 no?

oliveralka:
Ok, then let's test it via the GitHub actions. I think it should work. 

Works locally for Big Sur.

oliveralka:
Seems there are some other issues as well
```
(build_pyopenms_39) ➜  20210224_test pip install /Users/alka/Documents/work/software/openms_pyopenms/20210224_test/pyOpenMS/dist/pyopenms-2.7.0-cp39-cp39-macosx_11_0_x86_64.whl 
ERROR: pyopenms-2.7.0-cp39-cp39-macosx_11_0_x86_64.whl is not a supported wheel on this platform.
```
Have to investigate further. Was working before.

oliveralka:
Set `osx_deployment_target` was added to ZLIB in the contrib:
https://github.com/OpenMS/contrib/pull/107

I guess the minimum would now be defined by qt (10.13). Should be fine. 

@jpfeuffer What do you think?

jpfeuffer:
I think this PR is valid in any case. Makes sense that sysroot should go to the link args.

jpfeuffer (approved):

PR #5494, opened 2021-08-17
[Fix] Remaining memory leaks in tests

dvdvgt:
# Description
I had a look at the [remaining memory leaks](http://cdash.openms.de/viewTest.php?onlyfailed&buildid=170763) in the tests. Sadly almost all have them are due to leaks in third party dependencies and therefore hardly fixable.

| Name | Memory Leak source | Description |
|---|---|---|
| DetectabilitySimulation_test | libsvm | Unresolvable |
| FeatureFinderAlgorithmPicked_test | Raw pointer | Fixed by using a `std::shared_ptr`. Not sure if this introduces a relevant performance penalty. Thoughts? |
| FeatureFinderMetabo_test | libsvm | Unresolvable |
| LibSVMEncoder_test | libsvm | Unresolvable |
| MzMLFile_test | `xercesc_3_2::MemoryManagerImpl::allocate(unsigned long)` | Unresolvable |
| MzMLSpectrumDecoder_test | `xercesc_3_2::MemoryManagerImpl::allocate(unsigned long)`, ... | Unresolvable |
| OfflinePrecursorIonSelection_test | libsvm | Unresolvable |
| OSWFile_test | libsqlite | Unresolvable |
| PepXMLFile_test | probably a false positiv | Allocation [here](https://github.com/OpenMS/OpenMS/blob/bf46a11720e76a406addbaafdddaff9f6a48515f/src/openms/source/METADATA/PeptideHit.cpp#L94) and deletion [here](https://github.com/OpenMS/OpenMS/blob/bf46a11720e76a406addbaafdddaff9f6a48515f/src/openms/source/METADATA/PeptideHit.cpp#L116). Should be fine.  |
| PrecursorIonSelectionPreprocessing_test | libsvm | Unresolvable |
| PSLPFormulation_test | libsvm | Unresolvable |
| SimpleSVM_test | libsvm | Unresolvable |
| SVMWrapper_test | libsvm | Unresolvable |
| TheoreticalSpectrumGenerator_test | probably a false positiv | Allocation [here](https://github.com/OpenMS/OpenMS/blob/bf46a11720e76a406addbaafdddaff9f6a48515f/src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp#L151) and deletion [here](https://github.com/OpenMS/OpenMS/blob/bf46a11720e76a406addbaafdddaff9f6a48515f/src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp#L215). Should be fine. |
| TOPP_ExecutePipeline_1 | Qt | Unresolvable |
| TOPP_FeatureFinderCentroided_1 | Raw pointer in `FeatureFinderAlgorithmPicked` | Fixed. See prior entry `FeatureFinderAlgorithmPicked_test` |
| TOPP_PTModel_1 | libsvm | Unresolvable |
| TOPP_RTModel_1 | libsvm | Unresolvable |
| TOPP_RTModel_2 | libsvm | Unresolvable |
| TOPP_RTModel_3 | libsvm | Unresolvable |
| TOPP_RTModel_4 | libsvm | Unresolvable |
| TOPPView_test | Qt | Unresolvable |
| UnimodXMLFile_test | `xercesc_3_2::DefaultHandler` | Unresolvable |

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderAlgorithmPicked.h
+++ b/src/openms/include/OpenMS/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderAlgorithmPicked.h
@@ -232,7 +232,7 @@ protected:
 
       @return A pointer to the trace fitter that should be used.
      */
-    TraceFitter* chooseTraceFitter_(double& tau);
+    std::unique_ptr<TraceFitter> chooseTraceFitter_(double& tau);
 
     double intensityScore_(Size rt_bin, Size mz_bin, double intensity) const;
 
@@ -251,7 +251,7 @@ protected:
       @param traces Original mass traces found in the experiment.
       @param new_traces Mass traces created by cropping the original mass traces.
      */
-    void cropFeature_(TraceFitter* fitter,
+    void cropFeature_(std::shared_ptr<TraceFitter> fitter,
                       const MassTraces& traces,
                       MassTraces& new_traces);
 
@@ -278,7 +278,7 @@ protected:
 
       @return true if the feature is valid
      */
-    bool checkFeatureQuality_(TraceFitter* fitter,
+    bool checkFeatureQuality_(std::shared_ptr<TraceFitter> fitter,
                               MassTraces& feature_traces,
                               const double& seed_mz, const double& min_feature_score,
                               String& error_msg, double& fit_score, double& correlation, double& final_score);
@@ -296,7 +296,7 @@ protected:
       @param peak The Seed Peak
       @param path The path where to put the debug files (default is debug/features)
     */
-    void writeFeatureDebugInfo_(TraceFitter* fitter,
+    void writeFeatureDebugInfo_(std::shared_ptr<TraceFitter> fitter,
                                 const MassTraces& traces,
                                 const MassTraces& new_traces,
                                 bool feature_ok, const String error_msg, const double final_score, const Int plot_nr, const PeakType& peak,
--- a/src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp
+++ b/src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp
@@ -41,6 +41,7 @@
 #include <OpenMS/CHEMISTRY/AASequence.h>
 #include <OpenMS/CHEMISTRY/ResidueDB.h>
 #include <OpenMS/KERNEL/MSSpectrum.h>
+#include <OpenMS/CONCEPT/RAIICleanup.h>
 
 #include <unordered_set>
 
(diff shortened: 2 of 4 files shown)


jpfeuffer:
What does this do in case of actual deletion during assignment? Does not look right but I only gave it a quick thought.
```
    analysis_results_ = nullptr;
    if (source.analysis_results_ != nullptr)
    {
      // free memory first
      delete analysis_results_;
```

jpfeuffer:
In the TSG_test, is add_metainfo_ enabled? The move here looks suspicious, especially for a class that inherits from both std::vector and MetaInfoDescription with a default move ctor.
https://github.com/OpenMS/OpenMS/blob/bf46a11720e76a406addbaafdddaff9f6a48515f/src/openms/source/CHEMISTRY/TheoreticalSpectrumGenerator.cpp#L207?

I am also not sure if it is safe (or catchable by ASAN), if a moved_from pointer is deleted. Maybe make it a nullptr after the move?

jpfeuffer:
While I think you are correct with the rest of the leaks to be related to third-party libraries, I also think that probably some of them could be due to misuse of those libraries and could still be solved on our end.
If this can be done with a reasonable effort, is another question.

Thanks for investigating.

dvdvgt:
> What does this do in case of actual deletion during assignment? Does not look right but I only gave it a quick thought.

https://github.com/OpenMS/OpenMS/blob/bf46a11720e76a406addbaafdddaff9f6a48515f/src/openms/source/METADATA/PeptideHit.cpp#L129-L135

Yes, it does look a bit weird. `analysis_results_` is set to a `nullptr` which means if `analysis_results_` is deleted it is always a `nullptr` that gets deleted. A quick search revealed that deleting a `nullptr` is safe. However the memory `analysis_results_` points to before being assigned the `nullptr` is never freed. I'd change that so that `analysis_results_` is always deleted. What you think?

---

> In the TSG_test, is add_metainfo_ enabled? The move here looks suspicious, especially for a class that inherits from both std::vector and MetaInfoDescription with a default move ctor.

Form some tests activated for other it is not.

https://github.com/OpenMS/OpenMS/blob/34f271dee170e1f4cfc8a86565cc0d199e65850d/src/tests/class_tests/openms/source/TheoreticalSpectrumGenerator_test.cpp#L224-L227

https://github.com/OpenMS/OpenMS/blob/34f271dee170e1f4cfc8a86565cc0d199e65850d/src/tests/class_tests/openms/source/TheoreticalSpectrumGenerator_test.cpp#L251-L254

> Maybe make it a nullptr after the move?

I'll do that.

---

> If this can be done with a reasonable effort, is another question.

I figured that too. Fixing some of these memory leaks would require some major refactoring. That's probably not worth it considering only some bytes are leaking.

jpfeuffer:
Yep, give those a try. Do you have an ASAN setup on your hardware?

dvdvgt:
Regarding `TheoreticalSpectrumGenerator_test` I noticed that the leaks **only** occur here:
https://github.com/OpenMS/OpenMS/blob/34f271dee170e1f4cfc8a86565cc0d199e65850d/src/tests/class_tests/openms/source/TheoreticalSpectrumGenerator_test.cpp#L598
https://github.com/OpenMS/OpenMS/blob/34f271dee170e1f4cfc8a86565cc0d199e65850d/src/tests/class_tests/openms/source/TheoreticalSpectrumGenerator_test.cpp#L604
So I concluded that `getSpectrum` is generally safe except for a premature function exit like an exception. If an exception occurs the delete statement might not be reached and the memory is not freed. The only way I know how to prevent that is to use smart pointers because they'll clean up after you in any case. In this case `std::shared_ptr` because `ion_names` and `charges` need to have shared access since they can be passed to more than one function like `addPeaks_`, `addPrecursorPeaks_`, ....\
As always `std::shared_ptr` introduces some overhead but I think it should be negligible in this case. What do you think? 

---

Fixing the `PeptideHit::operator=` also seems to do the trick. Memory leaks are no longer occurring according to my ASAN.

jpfeuffer:
Yep, here safety would be more important than speed to me. I think the operations on the vectors will be more expensive than handling the pointers anyway.

timosachsenberg (approved):

PR #5498, opened 2021-08-23
[cython] link to openmp

jpfeuffer:
# Description

Tries to link to OpenMP also during cython compilation. Just linking to OpenMP during OpenMS lib compilation is seemingly not enough (i.e. when executing OpenMSBuildInfo().isOpenMPenabled() returns false).


oliveralka:
Should we set the osx_deployment_target to 10.13?

jpfeuffer:
I think it does not make a difference. If the rest is built for an older target it should not matter. Wheel will check all libs and pick the newest target.

jpfeuffer:
The buildfiles were for debugging but I think they cannot hurt. Maybe we should add them to other OSes as well at some point.

jpfeuffer:
For the public: The problem here is, that brew libraries are built with deployment target = current OS version = GitHub Actions macOS Runner OS version.
And libomp is downloaded from brew.
So we would need to build libomp ourselves if we want to lower the minimum target OS version.

oliveralka:
https://github.com/OpenMS/contrib/pull/108

oliveralka:
```
-rwxr-xr-x@ 1 alka  staff  35684056  9 Sep 09:09 libOpenMS.dylib
-rwxr-xr-x@ 1 alka  staff    169472  9 Sep 09:09 libOpenSwathAlgo.dylib
-rw-r--r--@ 1 alka  staff    617920  9 Sep 09:09 libomp.dylib
-rwxr-xr-x@ 1 alka  staff    105236  9 Sep 09:09 libz.dylib
```
I am not sure if the differences in permission matter. 

libomp.dylib (does not have a LC_VERSION_MIN_MACOSX)
```
cmd LC_BUILD_VERSION
    minos 10.15
      sdk 10.15.6
```
All other were build correctly

```
pyopenms otool -l *.dylib | grep -A 3 LC_VERSION_MIN_MACOSX
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.9
      sdk 11.1
--
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.9
      sdk 11.1
--
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.9
      sdk 10.15.4
```

tjeerdijk:
Hi Oliver,

Can you also look at the extended attributes with command “xattr”? I’ve been burned by these in another context.

Take care from Tjeerd

jpfeuffer:
I think the findopenmp module from cmake is pretty bad. With dozens of open issues..
https://gitlab.kitware.com/cmake/cmake/-/issues/21787

jpfeuffer:
Which means you probably have to give it many hints through parameters.

oliveralka:
Yes, it is pretty bad, but I tested it locally by adding the path to `CMAKE_PREFIX_PATH`

```
-DCMAKE_PREFIX_PATH="/usr/local/Cellar/qt@5/5.15.2;/Users/alka/Documents/work/software/contrib_test_build/lib/"
```

and it at least seems to have found the correct OpenMP.

```
//Path to a library.
OpenMP_libomp_LIBRARY:FILEPATH=/Users/alka/Documents/work/software/contrib_test_build/lib/libomp.a
```

jpfeuffer:
Not sure if Openmscontriblibs is added to prefix path early enough. Multithreading.cmake is included very early.

jpfeuffer:
Ok seems like it worked. The filename is 10_9 now. Need to test the installation now.

oliveralka:
There seems to be a problem with the linking:
```
>>> import pyopenms
Determination of memory status is not supported on this 
 platform, measuring for memoryleaks will never fail
OMP: Error #15: Initializing libomp.dylib, but found libomp.a already initialized.
OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://openmp.llvm.org/
[1]    66181 abort      python
```

jpfeuffer:
Can we build it dynamic? Maybe we should also just use the includes and require that openmp is installed? Probably through numpy anyway. 
https://gist.github.com/andyfaff/084005bee32aee83d6b59e843278ab3e

oliveralka:
https://github.com/OpenMS/contrib/pull/109

oliveralka:
In theory it works now:
```
(test_pyopenms_39) ➜  macOS-wheels pip install pyopenms-2.7.0-cp39-cp39-macosx_10_9_x86_64.whl 
Processing ./pyopenms-2.7.0-cp39-cp39-macosx_10_9_x86_64.whl
Requirement already satisfied: numpy in /usr/local/miniconda3/envs/test_pyopenms_39/lib/python3.9/site-packages (from pyopenms==2.7.0) (1.20.1)
Installing collected packages: pyopenms
Successfully installed pyopenms-2.7.0
(test_pyopenms_39) ➜  macOS-wheels python
Python 3.9.0 | packaged by conda-forge | (default, Oct 14 2020, 22:56:29) 
[Clang 10.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyopenms
Determination of memory status is not supported on this 
 platform, measuring for memoryleaks will never fail
>>> print(pyopenms.EmpiricalFormula('C6H12O6').getMonoWeight())
180.0633903828
```
libomp.dylib is now also transfered to the pyopenms directory. 

Unfortunately, this time I really had issue with calling pyopenms after installation:
Screenshot 2021-09-15 at 12 16 03
It seems the pyopenms*.whl is now added to quarantine due to the gatekeeper. 

Depending on the user premission that can be changed via:
`pip install sudo xattr -r -d com.apple.quarantine *`

jpfeuffer:
Can you check OpenMSInfo().numThreads() or sth like this?

oliveralka:
```
>>> omsinfo = pyopenms.OpenMSBuildInfo()
>>> omsinfo.isOpenMPEnabled()
True
>>> omsinfo.getOpenMPMaxNumThreads()
4
```

timosachsenberg:
hi what is the state of this PR?

oliveralka:
I am not sure what to do about the quarantine - Not sure if it is enought to just use 
`sudo xattr -r -d com.apple.quarantine *` on the .whl before uploading or if we somehow have to sign it differently.

jpfeuffer:
Haha you wish! It is quarantined when it is downloaded only ;)

jpfeuffer:
To be honest I think downloading from commandline (i.e. with pip) will not trigger quarantine.

jpfeuffer:
Only when you download via the browser from the artifacts.

oliveralka:
Then I guess - ready to merge!

PR #5509, opened 2021-08-31
[Fix] picked protein fdr

jpfeuffer:
from CDash on Gcc9


src/openms/include/OpenMS/ANALYSIS/ID/IDScoreGetterSetter.h
```diff
@@ -290,7 +290,7 @@ namespace OpenMS
                     std::vector<IDType> &ids,
                     const std::string &score_type,
                     bool higher_better,
-                    Args &... args)
+                    Args &&... args)
```
timosachsenberg, line 293: looks good to me... not sure if there is a better solution to having higher_better unused in one of the members

jpfeuffer: It probably should make an if case as well

Issue #5508, opened 2021-08-31
The question of the SiriusAdapter node 


hello-lei:
Hi,
I use the workflow to further identify the compound, then the mzTable is empty, but the data of the SIRIUS.ms is not empty![image](https://user-images.githubusercontent.com/83147334/131458472-fa5c09c6-f566-467a-a10f-c17ffc32febb.png)
![Uploading image.png…]()

Thank you very much!
@Oliver Alka

oliveralka:
Hi @hello-lei,
what is the purpose of having the PeakPickerHiRes and the HighResPrecursorMAssCorrector twice in your workflow? 

Which KNIME and OpenMS Version are you using? 

Could you please check the `View: SiriusAdapter Std Output` and `View: SiriusAdaper Error Output`
by right-clicking on the SiriusAdapter node?

MzTab-Reader has multiple output e.g. Metadata_Section, Protein_Section, ...
Did you check the correct section by using right-click and choosing `4: Small Molecule Section` on the MzTabReader Node?

hello-lei:
The PeakPickerHiRes convert the  profile file to centroid file and the HighResPrecursorMAssCorrector node corrects the mass and charge according the previous discussion.Because I try to change the empty data,but it doesn't work.
My KNIME is 4.4.0,OpenMS is 2.6.0.
![image](https://user-images.githubusercontent.com/83147334/131492738-3cca9d9c-5985-41ec-af11-73bfa1fd5e44.png)
 Since I update the Knime It became the ...and more.
![image](https://user-images.githubusercontent.com/83147334/131492902-1a308a1f-6432-403d-9a93-a82c7ea191cb.png)
Thank you very much.

oliveralka:
Could you please upgrade our KNIME Version to 4.4.1. The Std Output and Std Error work there as intended again. 
Please let me know the output of these.

hello-lei:
Ok,I will try it.Thank you!

hello-lei:
![image](https://user-images.githubusercontent.com/83147334/131515506-dfcdafda-5ece-45cd-af32-602183da0387.png)
![image](https://user-images.githubusercontent.com/83147334/131515615-501948b0-45d0-4b64-b8aa-18e5a1d69dd6.png)

oliveralka:
Thanks, could you upload the ones from the SiriusAdapter as well?

hello-lei:
I don't understand your meaning clearly? Is it?
[...]

PR #5511, opened 2021-09-01
[FIX] Sirius default executable names

oliveralka:
# Description

The default names were set incorrectly (windows), so KNIME was not able to get the path to the correct executable (see #5508). 

I also changed the "is_executable" to "skipexists". I hope with this step it is possible again to set the executable externally if KNIME is not able to find it. This should also be better for galaxy @timosachsenberg @jpfeuffer.

# Checklist:

# Note:


jpfeuffer:
Good find. What exactly does the change to skipexists do though? Or what are you hoping that it does?

oliveralka:
I hope that if the executable is not in the path, will allow to set it at the Node. 

Currently, you are not able to set the executable anymore for the SiriusAdapter (since it should be in the PATH). 
For the AssayGenerator it should be in the PATH as well, but if you want to use another version, or it is not found you can still set in at the KNIME Node - I use the "skipexists" for it - I hope that is the correct purpose. 

Screenshot 2021-09-01 at 11 16 20
Screenshot 2021-09-01 at 11 16 30

oliveralka:
@jpfeuffer I seems my thoughts are not correct and skipexists does not do what I intended to - is there a way to fix that?

jpfeuffer:
This happens here: https://github.com/OpenMS/OpenMS/blob/ddf175c5251b1ecaaed041474379489781f1eedd/cmake/knime_package_support.cmake#L138

oliveralka:
Ok, hopefully fixed now. 

The "optional" input "executable" should be available for the both nodes (SiriusAdapter, AssayGeneratorMetabo).

Do you think by setting the executable default name - this should work if it is in the PATH (KNIME)?
Or would you have to set it all the time?

jpfeuffer (approved):

PR #5523, opened 2021-09-07
validate experimental design

timosachsenberg:
# Description

Setting the file section should always trigger a validation.
Currents state:
whether an experimental design is valid depends on the tool capabilities/constraints

A solution should thus define constraints which are then checked for.

# Checklist:

# Note:


timosachsenberg:
Tests are failing because we allow subsets of input files in ProteomicsLFQ (which might not have consecutive fraction/fraction groups etc.).

timosachsenberg:
I think tools need to decide to what level they support deviation from the "canonical" design. Maybe we could fail (as we do now) as default and let tool disable (some) validy tests?

jpfeuffer:
Fine for me, although the ultimate goal for me would be to just have a factor based experimental design like almost every other package has.
I think it shouldn't be hard to keep another mapping when creating the experimental design.
Internally we can represent it as we want, consecutive starting from one, or whatever.

PR #5536, opened 2021-09-14
PLFQ Write chromatograms for debugging

jpfeuffer:
at certain debug level

### DIFF
--- a/src/utils/ProteomicsLFQ.cpp
+++ b/src/utils/ProteomicsLFQ.cpp
@@ -1158,7 +1158,7 @@ protected:
 
       if (debug_level_ > 670)
       {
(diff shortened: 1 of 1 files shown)


timosachsenberg:
I think that makes sense. In principle, one could even combine those into peak groups for several files and store them in an osw file

jpfeuffer:
Unfortunately I dont know how to do that. All I can say is that you can also write the transitions.

jpfeuffer:
Also, it sounds prohibitively expensive for large datasets to store all chromatograms across all files and then merge them.

jpfeuffer:
The only possibility I see is to fill the OSW db incrementally. But I have no idea if it is possible to create peakgroups with this.

PR #5544, opened 2021-09-20
[Feature] Try to add descriptions to Param class repr in pyopenms

jpfeuffer:
# Description

Untested. For an easy
`print(Algorithm().getDefaults())`
including descriptions


jpfeuffer:
I hope __str__ is not already defined by Cython/autowrap

Issue #4927, opened 2020-09-10
MSExperiment design correction: remove Identifications


timosachsenberg:

timosachsenberg:
slightly advanced for Hacktoberfest

timosachsenberg:
1. Removing getter and setter from: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/METADATA/SpectrumSettings.h#L153
as well as pyOpenMS wrapper.
2. Fix code to not store ID data there but keep ID and spectra separate

timosachsenberg:
Done

PR #5547, opened 2021-09-21
MSExperiment: Store idenfication data and spectra separately

dvdvgt:
This PR aims to tackle the issue #4927; to separate identification data from spectra data. This involves removing https://github.com/OpenMS/OpenMS/blob/d69467f0e22054957aad0425885ee05758ca6104/src/openms/include/OpenMS/METADATA/SpectrumSettings.h#L178
all `PeptideIdentification` from `SpectrumSettings` and maybe also
https://github.com/OpenMS/OpenMS/blob/d69467f0e22054957aad0425885ee05758ca6104/src/openms/include/OpenMS/METADATA/ExperimentalSettings.h#L149
all `ProteinIdentification` from `ExperimentSettings`. 

# Description

@timosachsenberg proposed to implement a new class that realizes the mapping between ID and spectra called [`AnnotatedMSRawData`](https://github.com/dvdvgt/OpenMS/blob/feature/AnnotatedMSExperiment/src/openms/include/OpenMS/KERNEL/AnnotatedMSRawData.h) (or something like that).

The idea is to offer an interface to conveniently get both `MSSpectrum` and `PeptideIdentification` as a `std::pair` while also maintaining the functionality offered by `MSExperiment`. I decided to go with inheritance instead of composition because it simplifies calling functions from `MSExperiment` since you don't have to access a `MSExperiment` member of `AnnotatedMSRawData` every time. \
 `AnnotatedMSRawData` implements iterator that returns a `std::pair` and can be used in e.g. *range-based for loops*.

Aside from the design of the new class `AnnotatedMSRawData`, the replacing of `MSExperiment` occurrences with `MSAnnotatedMSRawData` where identification data is needed, obliviously requires a considerable amount of refactoring. 

I already started to do some of the refactoring but I would like to hear your thoughts and suggestions before proceeding further so that issues of any kind can be found early on and fixed easily.

# Checklist:

# Note:


src/openms/include/OpenMS/KERNEL/AnnotatedMSRawData.h
```diff
@@ -0,0 +1,153 @@
...
+      T2 m_ptr2;
+  };
+
+  typedef AnnotatedMSRawData::PairIterator<std::vector<MSSpectrum>::iterator , std::vector<std::vector<PeptideIdentification>>::iterator> Iterator;
+  typedef AnnotatedMSRawData::PairIterator<std::vector<MSSpectrum>::const_iterator , std::vector<std::vector<PeptideIdentification>>::const_iterator> ConstIterator;
+
+  private:
+    std::vector<std::vector<PeptideIdentification>> peptide_ids;
+  };
+}
```
timosachsenberg, line 153:
```suggestion
}
```
```suggestion
}

```

src/openms_gui/source/VISUAL/LayerData.cpp
```diff
@@ -168,9 +168,9 @@ namespace OpenMS
 
   void LayerData::updateCache_()
   {
-    if (peak_map_->getNrSpectra() > current_spectrum_idx_ && (*peak_map_)[current_spectrum_idx_].size() > 0)
+    if (peak_map_->getNrSpectra() > current_spectrum_idx_ && (*peak_map_)[current_spectrum_idx_].first.size() > 0)
```
timosachsenberg, line 171: hmm not sure about the .first .second syntax.

src/openms/include/OpenMS/KERNEL/AnnotatedMSRawData.h
```diff
@@ -0,0 +1,153 @@
...
+#include <boost/range/combine.hpp>
+
+#include <vector>
+
+namespace OpenMS
+{
+  class PeptideIdentification;
+  class MSSpectrum;
+
+  class OPENMS_DLLAPI AnnotatedMSRawData : public MSExperiment
```
timosachsenberg, line 48: without inheritance might lead to a better design (but maybe some more changes)
e.g., one would need to write e.g. annot_ms_data.getMSExperiment()

src/openms/include/OpenMS/KERNEL/AnnotatedMSRawData.h
```diff
@@ -0,0 +1,153 @@
...
+    {
+      return PairIterator(spectra_.end(), peptide_ids.end());
+    }
+
+    inline auto cend() const
+    {
+      return PairIterator(spectra_.cend(), peptide_ids.cend());
+    }
+
+    inline Mapping operator[](size_t idx)
```
timosachsenberg, line 91: I kind of don't like the accession with .first .second that much as one always needs to remember what first and second is.
what would be the advantage for the user when he mainly writes something like: 
annot_ms.getExperiment()[idx]  or annot_ms.getIdentification()[idx]
but rarely needs both

src/openms/include/OpenMS/KERNEL/AnnotatedMSRawData.h
```diff
@@ -0,0 +1,153 @@
...
+    AnnotatedMSRawData() = default;
+    /// Move constructor for efficiently loading a MSExperiment without a deep copy.
+    explicit AnnotatedMSRawData(MSExperiment&& experiment) : MSExperiment(std::move(experiment)) {};
+    AnnotatedMSRawData(AnnotatedMSRawData&&) = default;
+    ~AnnotatedMSRawData() = default;
+
+    /// Get the peptide identifications for a single spectrum.
+    std::vector<PeptideIdentification>& getPeptideIds();
+    /// Get all peptide identifications for all spectra.
+    std::vector<std::vector<PeptideIdentification>>& getAllPeptideIds();
```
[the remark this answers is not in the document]
dvdvgt: I'm also not sure how to handle indexing. I was thinking to use a 2d vector

```
std::vector<std::vector<PeptideIdentification>>
```

where, as you mentioned, the indices of spectra in the MSExperiment correspond to the indices of the outter vector. That means however that in order to add or edit peptide ids of a specific spectrum, you'd have to know the index. I can imagine that there are cases where the index is unknown. As of now peptide identification is stored within the spectra itself so indices are not used for storing these IDs, at least not directly. Also using indices feels a bit unsafe since accessing the correct spectrum cannot be guarantied by the interface itself but must be ensured by the programmer himself.

Maybe it would be better to use a `std::map<MSSpectrum, std::vector<PeptideIdentification>>` so accessing IDs is more straight forward. This would come at the disadvantage that upon the construction of `AnnotatedMSRawData` all spectra would need to be copied of into the map and the spectra vector in `MSExperiment` might no longer be the same as the keys in the map as things are added to the map. That means you'd have to prohibit accessing the `MSExperiment::getSpectra` function somehow.\
Alternatively it might be possible to use a `std::reference_wraper` as key instead. This would complicate adding or removing spectra though because the map and actual vector of spectra always would need to be kept in sync. 

What are your thoughts on this?

timosachsenberg:
one note Julianus brought up.
It usually doesn't make sense to have multiple PeptideIdentifications per spectrum. Would it simplify the code if you would only assume one instead of a vector per spectrum?

src/openms/include/OpenMS/KERNEL/AnnotatedMSRawData.h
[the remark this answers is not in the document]
timosachsenberg: my take on what could work:
provide only std::vector<PeptideIdentification> - one PeptideIdentification (potentially empty) per spectrum
I think the vector of vector is not needed for all use cases we have (though the original design supported it)

Issue #5541, opened 2021-09-18
OpenMS-2.7.0 - Various test failed


sagitter:
Hi all.
Various tests are failed in Fedora 36 (devel branch) x86_64 ([openms-2.7.0-buildlog.zip](https://github.com/OpenMS/OpenMS/files/7190327/openms-2.7.0-buildlog.zip)) including **non-fatal errors** like those reported in #5538:

```
$ ctest --jobs 4 -VV --force-new-ctest-process --output-on-failure && exit 1
...

The following tests FAILED:
	519 - PrecursorIonSelectionPreprocessing_test (Failed)
	599 - DetectabilitySimulation_test (Failed)
	639 - MRMFeatureSelector_test (Failed)
	2282 - TOPP_OpenPepXL_1_out_2 (Failed)
	2287 - TOPP_OpenPepXLLF_1_out_2 (Failed)
	2334 - UTILS_AssayGeneratorMetabo_1 (Failed)
	2335 - UTILS_AssayGeneratorMetabo_1_out1 (Failed)
	2336 - UTILS_AssayGeneratorMetabo_2 (Failed)
	2337 - UTILS_AssayGeneratorMetabo_2_out1 (Failed)
	2338 - UTILS_AssayGeneratorMetabo_3 (Failed)
	2339 - UTILS_AssayGeneratorMetabo_3_out1 (Failed)
	2340 - UTILS_AssayGeneratorMetabo_4 (Failed)
	2341 - UTILS_AssayGeneratorMetabo_4_out1 (Failed)
	2342 - UTILS_AssayGeneratorMetabo_5 (Failed)
	2343 - UTILS_AssayGeneratorMetabo_5_out1 (Failed)
Errors while running CTest
```
For example:

```
test 519
          Start  519: PrecursorIonSelectionPreprocessing_test
519: Test command: /builddir/build/BUILD/OpenMS-Release2.7.0/build/src/tests/class_tests/bin/PrecursorIonSelectionPreprocessing_test
519: Test timeout computed to be: 1500
519: checking PrecursorIonSelectionPreprocessing() ... 
519:  +  line 56:  TEST_NOT_EQUAL(ptr,nullPointer): got '0x55570c32ec50', forbidden is '0'
519: : passed
519: 
519: checking ~PrecursorIonSelectionPreprocessing() ... : passed
519: Warning: no subtests performed in '~PrecursorIonSelectionPreprocessing()' (line 61)!
519: 
[...]

PR #5551, opened 2021-09-23
[FIX] AssayGeneratorMetabo Test  (wo SIRIUS)

oliveralka:
# Description

Due to the introduction of setting the default executable on all platforms, the non SIRIUS related tests failed. 

See Windows CI: https://github.com/OpenMS/OpenMS/runs/3678541178

I am not sure why this did not come up earlier. I guess, we have to add that to release 2.7.0 as well.

https://github.com/OpenMS/OpenMS/pull/5511

Part of current issue: #5541

# Checklist:

# Note:


jpfeuffer:
Related to the fedora issue. Please link.

jpfeuffer:
But this is not a great user experience to be forced to specify `-executable ""` to make the tool work.
Why is that exactly?

oliveralka:
So, the tool has to options - run with and without SIRIUS.

I switched the run methods to use SIRIUS as primary. That the executable can be found in KNIME automaticcaly, I added the executable sirius.bat/sirius as default. 

Now, when you want to run the tool without SIRIUS, but sirius is set as default, you get the following error:
```
2311: Input file 'sirius' could not be found (by searching on PATH). Either provide a full filepath or fix your PATH environment! Since this file is not strictly required, you might also pass the empty string "" as argument to prevent it's usage (this might limit the usability of the tool).
2311: Error: File not found (the file 'sirius' could not be found) 
```
https://github.com/OpenMS/OpenMS/blob/c2555c3278de34a5f261df9a0915f42a8bfada3c/src/openms/source/APPLICATIONS/TOPPBase.cpp#L1348

So basically this is now handeled by TOPPBase instead of the tool internally, which only wants to know which fragment annotation you use and if the executable is available. 

`else if (use_fragment_annotation && !executable.empty())`
https://github.com/OpenMS/OpenMS/blob/develop/src/utils/AssayGeneratorMetabo.cpp#L434

jpfeuffer:
How about leaving the parameter empty and if sirius is activated, fill it inside the code with the defaults (if it is still empty) and do the executable check (that TOPPBase would do)?

jpfeuffer:
And how about calling it sirius_executable?

oliveralka:
Would that work in KNIME then without specifying the executable?

jpfeuffer:
I don't think this is a KNIME specific issue. If no default executable name is given, it won't find Sirius in any case, right?
The problem is that you have to look in the PATH. And TOPPBase does that automatically.
If the executable is optional, you can of course not require it to be found during startup of the tool.
That means you have to defer the lookup until it is clear that the executable will be needed.

oliveralka:
Ok, thanks for the explanation, then I will try it similar to your suggestion above.

oliveralka:
rebuild jenkins

oliveralka:
Failed test unrelated: `projectroot.src.tests.class_tests.openms.FIAMSScheduler_test`

oliveralka:
@jpfeuffer I guess this has to go to release as well - will open another PR after review.

jpfeuffer:
I am honestly considering a 2.7.1 release instead.

oliveralka:
Ok, will add a branch and open an additional PR.

oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

oliveralka:
rebuild jenkins

PR #5562, opened 2021-09-27
Do not overwrite OPENMS_DATA_PATH in PyOpenMS if it is already set

radusuciu:
# Description

I've got a lot of customizations (mostly to `unimod.xml` ) that I've set in a new directory that is referenced by `OPENMS_DATA_PATH` but PyOpenMS clobbers that on initialization. This simple modification should prevent it from doing so, if there is a path already set.

# Checklist:

# Note:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -98,4 +98,4 @@ the authors tag in the respective file header.
  - Witold Wolski
  - Xiao Liang
  - Justin Sing
- 
+ - Radu Suciu
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@ General:
 - OpenMS now expects a compiler that supports C17
 - Config storage path on linux changed to ~/.config/
 - Some documentation improvements to pyOpenMS https://pyopenms.readthedocs.io/en/latest/
+- PyOpenMS checks if `OPENMS_DATA_PATH` environment variable is set, before setting to default value
 
 Adapters/Third-party support:
 - Added support for SIRIUS 4.9.0
--- a/src/pyOpenMS/pyopenms/__init__.py
+++ b/src/pyOpenMS/pyopenms/__init__.py
@@ -28,7 +28,9 @@ from .version import version as __version__
 
 import os
 here = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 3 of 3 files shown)


timosachsenberg:
@jpfeuffer looks reasonable. We often have both OpenMS and pyOpenMS installed so if versions differ this might lead to problems. E.g., new OpenMS version with Elements.xml removed and old version of pyOpenMS that expects the file to exist in share. Do you think we need to cover that case?

jpfeuffer:
Hmm yes, I am not sure. Maybe we should print an info message that the path is set from outside and will be used?

timosachsenberg:
I agree. Outputting a warning like:
"Warning: OPENMS_DATA_PATH environment variable already exists. pyOpenMS will use it ('" + path + "') to locate data in the OpenMS share folder (e.g., the unimod database)." 
Would suffice @radusuciu can you add this?

radusuciu:
That sounds very reasonable, apologies for the accidental merge commit - I didn't mean to push when I did. I have a slight edit to make to make the line length more reasonable if that's okay with you.

radusuciu:
Regardless of whether or not you accept my stylistic change, thank you very much for being open to this small contribution! This makes my life just a littttle bit easier :)

radusuciu:
An alternative that I just thought about would be to have a different environment variable that is specific to eg. `PYOPENMS_DATA_PATH` or similar.

```python
default_openms_data_path = os.path.join(here, "share/OpenMS")
pyopenms_openms_data_path = os.environ.get("PYOPENMS_DATA_PATH")

os.environ["OPENMS_DATA_PATH"] = pyopenms_openms_data_path or default_openms_data_path 
```

or

```python
default_openms_data_path = os.path.join(here, "share/OpenMS")
os.environ["OPENMS_DATA_PATH"] = os.environ.get("PYOPENMS_DATA_PATH", default_openms_data_path)
```

This probably wouldn't necessitate a warning, just some documentation about `PYOPENMS_DATA_PATH`.

timosachsenberg:
Nice. Thanks for the cleanup. From my side this PR is good to go!

oliveralka (approved):
Looks good!

PR #5563, opened 2021-09-28
[question] how to wrap rvalue reference version?

timosachsenberg:
# Description

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/IONMOBILITY/IMDataConverter.h
+++ b/src/openms/include/OpenMS/IONMOBILITY/IMDataConverter.h
@@ -67,9 +67,11 @@ namespace OpenMS
         @throws Exception::MissingInformation if @p exp is not FAIMS data
       */
       static std::vector<PeakMap> splitByFAIMSCV(PeakMap&& exp);
-      static std::vector<PeakMap> splitByFAIMSCV(PeakMap& exp) { return splitByFAIMSCV(std::move(exp)); } // PYOPENMS: needed for wrapping the r-value refernce version
-
       
+      #ifdef PYOPENMS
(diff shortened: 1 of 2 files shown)


timosachsenberg:
@cbielow can we just do something like this to be able to wrap rvalue reference functions using lvalue references?
Relevant to @dmccloskey and collaborators

cbielow:
I don't have a good solution for this either, just some thoughts:

if you provide the overload using a l-value reference, this is semantically different from the rvalue-ref version, because in C++ I would expect that in the l-value ref version, the data is still meaningful. However, since all the data will be moved away, this is a potential source of error, at least of confusion. The user will wonder why there are two overloads and what is the difference. Also, it allows to call the method without using an explicit `std::move`, so the user is not aware of what will happen to his data.

So, some thoughts:
1) we could only enable the l-value ref if `PYOPENMS=On` (so there is a chance you catch wrong usages in test, where python bindings are disabled).
2) do not use an overload, but another function name (append _PY, for example). Ideally, during the wrapping, can you actually expose it as `splitByFAIMSCV` (is this possible???)
3) will cython/autowrap (with some pushing) be able to handle r-value refs in the near future?

jpfeuffer:
Regarding 2:
Yes should be possible with "wrap-as"

Regarding 3:

Cython can: https://github.com/cython/cython/blob/daa0a4434fa3ef281f9df65f7bb6098d31e1e400/Cython/Compiler/Parsing.py#L2915

Autowrap cannot:
https://github.com/uweschmitt/autowrap/blob/2dcf9724294f23f87ae686f7d5a5f193a5fab3e2/autowrap/Types.py#L48

We would probably need to add support ourselves. For basic types this would be as simple as adding support in autowrap.Types and PXDParsing, then adding an "if type.isrval then wrap in move" in CodeGeneration. But nested types could be trickier. Potentially we also need to upgrade autowrap to Cython 3, where I cannot foresee the work.

PR #5565, opened 2021-09-30
[FEATURE] Export isotope errors when reading pepXML

jpfeuffer:
# Description

For use in Percolator, parse isotope error with round(neutralmassdiff/C12C13diff).
Also contains various fixes for terminal modifications once again.

Contains another try to make clangd work out of the box on Gitpod.

### DIFF
--- a/src/openms/include/OpenMS/FORMAT/PepXMLFile.h
+++ b/src/openms/include/OpenMS/FORMAT/PepXMLFile.h
@@ -296,9 +296,6 @@ private:
     /// RT and m/z of current PeptideIdentification (=spectrum)  
     double rt_{}, mz_{};
 
-    /// possible isotopes to consider and annotate
-    //std::vector<int> potential_isotopeerrors_;
-
     /// 1-based scan nr. of current PeptideIdentification (=spectrum). Scannr is usually from the start_scan attribute
     Size scannr_{};
 
--- a/src/openms/include/OpenMS/FORMAT/PepXMLFile.h
+++ b/src/openms/include/OpenMS/FORMAT/PepXMLFile.h
@@ -296,6 +296,9 @@ private:
     /// RT and m/z of current PeptideIdentification (=spectrum)  
     double rt_{}, mz_{};
 
+    /// possible isotopes to consider and annotate
+    //std::vector<int> potential_isotopeerrors_;
+
     /// 1-based scan nr. of current PeptideIdentification (=spectrum). Scannr is usually from the start_scan attribute
     Size scannr_{};
 
@@ -341,5 +344,7 @@ private:
     bool lookupAddFromHeader_(double modification_mass,
                               Size modification_position,
                               std::vector<AminoAcidModification> const& header_mods);
+
+    //static std::vector<int> getIsotopeErrorsFromIntSetting_(int intSetting);
   };
 } // namespace OpenMS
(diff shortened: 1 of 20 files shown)


src/openms/include/OpenMS/FORMAT/PepXMLFile.h
```diff
@@ -296,6 +296,9 @@ namespace OpenMS
     /// RT and m/z of current PeptideIdentification (=spectrum)  
     double rt_{}, mz_{};
 
+    /// possible isotopes to consider and annotate
+    //std::vector<int> potential_isotopeerrors_;
```
timosachsenberg, line 300: Unused?

src/openms/source/FORMAT/PepXMLFile.cpp
```diff
@@ -1296,6 +1317,10 @@ namespace OpenMS
           peptide_hit_.setMetaValue("num_matched_peptides", value);
         }
       }
+
+      double pepmassdiff = attributeAsDouble_(attributes, "massdiff");
```
timosachsenberg, line 1321: Is this mandatory or can it be optional in pepXML?

timosachsenberg (approved):
Maybe double check if we use the sign of the isotope error the same through openms

jpfeuffer:
The only time we use it is here: https://github.com/OpenMS/OpenMS/blob/develop/src/topp/PercolatorAdapter.cpp#L501
where it must be signed to yield correct results (if I am not mistaken).
The only other method that annotates it is MSGFPlus. I don't know if it can be negative there.

src/openms/source/FORMAT/PepXMLFile.cpp
jpfeuffer: required

```
<xs:attribute name="massdiff" type="xs:string" use="required">
<xs:annotation>
<xs:documentation>Mass(precursor ion) - Mass(peptide)</xs:documentation>
</xs:annotation>
</xs:attribute>
```

src/openms/include/OpenMS/FORMAT/PepXMLFile.h
jpfeuffer: yes

jpfeuffer:
I fixed the MyriMatch test errors. Problem is, that the heuristic does not work since Myrimatch calculates/reports wrong theoretical peptide masses:
```
    <spectrum_query spectrum="spectra.3.3.3" spectrumNativeID="spectrum=2" start_scan="3" end_scan="3" precursor_neutral_mass="2323.13979169311" assumed_charge="3" index="3" retention_time_sec="4923.77734375002">
      <search_result num_target_comparisons="1" num_decoy_comparisons="0">
        <search_hit hit_rank="1" peptide="RPGADSDIGGFGGLFDLAQAGFR" peptide_prev_aa="-" peptide_next_aa="A" protein="BSA3" num_tot_proteins="1" calc_neutral_pep_mass="2324.50666" massdiff="1.36686830689" num_tol_term="2" num_missed_cleavages="1" num_matched_ions="24" tot_num_ions="46">
```

calc_neutral_pep_mass should be 2323.13.. I have no idea what they are doing..

jpfeuffer:
If we disable isotopes in myrimatch it still finds them, so it is a problem of reporting.
What do we do? Calculate on our own? I guess so.

timosachsenberg:
Myrimatch seems to be not developed further so don’t worry

PR #5516, opened 2021-09-03
move templates to cpp file to reduce Seqan exposure

timosachsenberg:
# Description

Attempt to remove compile time and reduce exposure of seqan

# Checklist:
- [ ] Make sure that you are listed in the AUTHORS file
- [ ] Add relevant changes and new features to the CHANGELOG file
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes
- [ ] Updated or added python bindings for changed or new classes. (Tick if no updates were necessary.)

# How can I get additional information on failed tests during CI:
If your PR is failing you can check out 
- http://cdash.openms.de/index.php?project=OpenMS and look for your PR. If you click in the column that lists the failed tests you will get detailed error messages.

# Note:
- Once you opened a PR try to minimize the number of *pushes* to it as every push will trigger CI (automated builds and test) and is rather heavy on our infrastructure (e.g., if several pushes per day are performed).

PR #5603, opened 2021-10-20
Reduce OpenMS exposure to third-party headers

hroest:
# Description

Currently in OpenMS, we pull in a large number of header files from various libraries. The largest of these are 

- boost (300k LOC)
- seqan (150k LOC)
- Eigen (70k LOC)
- Qt (20k LOC)
- xerces (20k LOC)
- Wildmagic (5k LOC)

Note that boost has many components, many of which are template/header based such as regex, math functions, algorithms, shared ptr, random, spirit (a full parser, we use it for casting numbers) and data structures that we use(d) a lot including maps, bimaps, multi_index_containers etc.

This blows up files and creates longer compile times for downstream users. This impacts compilations such as pyOpenMS which include *all* the OpenMS public API in order to wrap it. Currently only the headers of the public OpenMS API yield a file after pre-processing of 762586 LOC (yes, 760k LOC), on top of which the Cython-generated code comes which is another 500k LOC, yielding files after the preprocessor that are 1-1.5M LOC.

This pull request moves libraries from the headers down into the cpp files or into header files that do not get included in general by downstream users. This decreases the size of the includes from 760k LOC down to 492k LOC, so it basically removes over 250k LOC from compilation. Depending on individual cpp files, this number may be smaller but it general this should lead to smaller files and potentially faster compilation. It does this by

- removing unnecessary includes
- moving functions from headers to cpp files
- forward declaring instead of including
- replacing members (as in BinnedSpectrum) with pointers and forward-declaring
- extracting functions from heavily used headers such as String.h and StringUtils.h and placing them into separate purpose headers
- replace iostream with iosfwd where appropriate and removing LogStream.h from header files

Also this does not yet address seqan, which is another 150k LOC and should be addressed in #5516, potentially bringing down our exposure below 350k LOC.

### DIFF
--- a/src/openms/include/OpenMS/COMPARISON/SPECTRA/BinnedSpectrum.h
+++ b/src/openms/include/OpenMS/COMPARISON/SPECTRA/BinnedSpectrum.h
@@ -138,7 +138,7 @@ public:
     bool operator!=(const BinnedSpectrum& rhs) const;
 
     /// returns the bin intensity at a given m/z position 
-    float getBinIntensity(double mz)
+    float getBinIntensity(double mz);
 
     /// return the bin index of a given m/z position
     size_t getBinIndex(float mz) const;


hroest:
@timosachsenberg can you have a look at BinnedSpectrum? This is required to remove Eigen from the header files but it does introduce a level of indirection

PR #5610, opened 2021-10-22
Feature/header cleanup inc

hroest:
This is a subset of of PR #5603 simply moving a bunch of code from header files to .cpp files

I will use an incremental PR to move headers into cpp and leave out some of the more controversial changes.


hroest:
@cbielow first guess was right: 9d01d505b41ef24af21f25e468074e81863b7b21 -- now it compiles and even on clang where it shows red there are no failures. 

merge?

PR #5615, opened 2021-10-26
[ADD] Spectrum Plotting and Spectrum Mirror Plotting Functions

matteopilz:
# Description

# Checklist:

# Note:

### DIFF
--- a/src/pyOpenMS/pyopenms/Plotting.py
+++ b/src/pyOpenMS/pyopenms/Plotting.py
@@ -194,7 +194,6 @@ zorders = {'a': 3, 'b': 4, 'c': 3, 'x': 3, 'y': 4, 'z': 3, '?': 2, 'f': 5,
 
 
 def plot_chromatogram(c: MSChromatogram):
-    import matplotlib.pyplot as plt
     """
     Plot chromatogram peaks.
 
@@ -203,6 +202,7 @@ def plot_chromatogram(c: MSChromatogram):
     c : MSChromatogram
         The chromatogram to be plotted.
     """
+    import matplotlib.pyplot as plt
     x, y = c.get_peaks()
     plt.plot(x, y)
     plt.xlabel("Retention time")


bittremieux:
Sorry that I'm only looking at it now. ASMS (and the week before to prepare) was hectic.

Nice work! Would it make sense to use spectrum_utils directly, rather than duplicating code? That way there's less chance for technical debt, and improvements would be available in both spectrum_utils and pyOpenMS.
Although handling dependencies might be a little trickier in that case (although I deliberately try to minimize spectrum_utils dependencies).

timosachsenberg:
Hi Wout no worries. We merged this already so we can try out a few things.
Managing dependencies is indeed a bit of an issue for us and we would try to keep them as minimal as possible.
I think by just including the single file we got away without any additional one as opposed to the ones from  https://github.com/bittremieux/spectrum_utils/blob/master/environment.yml .
Don't know if it is worth the trouble trying to handle that on the spectrum_utils side? Maybe when the plotting interface stabilizes on our side?

bittremieux:
Makes sense.

The minimal set of dependencies for spectrum_utils is numba, NumPy, Pyteomics, and matplotlib. Optional extensions for interactive plotting use Altair and Pandas, and ProForma support requires fastobo and lark-parser.
So out of the base dependencies that would add numba and Pyteomics I think? It's pretty challenging to make those optional, because they're essential for base spectrum_utils functionality.

We should discuss how we can integrate both packages. spectrum_utils was also recently added as a plotting backend to Pyteomics, which is cool. So what I was thinking is that we could maybe try to write a short application note with authors from all teams to describe some recent additions to spectrum_utils (USI, ProForma) and introduce a unified Python mass spec ecosystem (spectrum_utils + Pyteomics + pyOpenMS).

timosachsenberg:
I like the idea of having a unified Python mass spec ecosystem a lot. Let's brainstorm a bit.

Looking at spectrum_utils it is indeed closely tied to Pyteomics.
In this PR we managed without any additional dependency, so one could also discuss if a slight refactoring would make sense that makes the plotting part (let's call it spectrum_plots) independent of pyOpenMS and Pyteomics (e.g., just use lists/dicts or numpy arrays).
Both pyOpenMS and Pyteomics could then provide the glue code to make their data work with the plotting part of the library.

The other possibility would be to make Pyteomics also a dependency for pyOpenMS but:
1. this might cause additional confusion to novel users because they need to deal with two APIs (and we currently have this project to document and make the usage simpler :) )
2. might make Julianus not happy, who rightfully complains about every additional dependency (which is a bit more of an issue in a mixed C++ / python world).

bittremieux:
spectrum_utils uses Pyteomics to calculate theoretical fragment masses, which is needed to annotate peaks. So that's pretty important functionality, although not directly needed for the plotting itself.
I'd need to think a bit about how to refactor spectrum_utils and change the installation options to not have Pyteomics by default. It can definitely be done, but I also don't want to include (too much) friction for users if they need to install special extensions to have the normal spectrum_utils functionality.

I'm currently making some improvements to spectrum_utils. Let me work a bit more on that and then think about the best way to provide interoperability between spectrum_utils, pyOpenMS, and Pyteomics. We should probably also include @levitsky in that conversation then. (Tagging him here to just start thinking about it already.)

levitsky:
Hi everyone,

I like the idea of a universal spectrum plotting mechanism (or maybe several of them) that we could all potentially collaborate on, and supported by major libraries that deal with mass spec data. I think `spectrum_utils`, as a standalone project with clearly defined functionality, could become such a project.

When I learned about `spectrum_utils`, I had already implemented a primitive API for spectrum plotting, so I decided to keep the user-facing API but implement calling `spectrum_utils` as a "plotting backend", making it an optional dependency. I don't know if this approach makes sense for pyOpenMS or not, but for pyteomics users I see the benefit in that I encapsulate the boilerplate code that massages the data from `pyteomics` formats (spectra from parsers, peptide sequences with modX notation) into `spectrum_utils` format.

As for adding Pyteomics as a dependency, I tried to keep it as painless as possible. Pyteomics is pure Python and has no hard dependencies (and indeed no extras are needed for mass calculations), so hopefully it shouldn't cause any trouble on any platform.

hechth:
Hi all! I also strongly support the idea of creating some common mass spectrometry ecosystem for plotting. I've been working mostly on matchms and other MS-related projects with Galaxy, but would love to see some collaboration on visualization tools, as the resources seem scattered. Also, the requirements for plotting a quite minimal, so interfaces to the libraries at a higher level of abstraction would be easy to implement.

PR #5622, opened 2021-10-28
add pandas df conversion

axelwalter:
# Description

Add conversion for FeatureMap, ConsensusMap, MSExperiment and list of PeptideIdentification.

# Checklist:

# Note:


jpfeuffer:
Hmm wasn't enough.

jpfeuffer:
Feel free to add tests now. @timosachsenberg @axelwalter

timosachsenberg:
I think we can add code from the pyOpenMS examples on the readthedocs page here that also tests writing and reading of files as well as conversion of the resulting maps to pandas:

for featureXML
https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/tests/unittests/test000.py#L2098
for consensusXML
https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/tests/unittests/test000.py#L873
for mzML file
https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/tests/unittests/test000.py#L3662
and idXML:
https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/tests/unittests/test000.py#L2209

axelwalter:
I will work on a readthedocs tutorial and add the examples as test cases.

jpfeuffer:
Hmm but wouldn't it be better to have it merged first with a very basic test?

timosachsenberg:
yep I agree.e.g.  just adding one or two features and then test the df shape

jpfeuffer:
I added quite some Todos that should be addressed in the future, in addition to an idXML test. But from my side we could merge it after CI.

timosachsenberg (approved):
Feel free to merge if tests pass

axelwalter:
Still need to fix handling missing values in meta values.

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -235,33 +235,59 @@ def get_df(self, long : bool = False):
...
+    default_missing_values: default value for missing values for each data type
+    export_unidentified: export PeptideIdentifications without PeptideHit
 
     Returns:
     pandas.DataFrame: peptide identifications in a DataFrame
     """
     switchDict = {bool: '?', int: 'i', float: 'f', str: 'U100'}
+
+    # filter out PeptideIdentifications without PeptideHits if export_unidentified == False
+    if not export_unidentified:
```
jpfeuffer, line 255: Thanks for taking my suggestion, however, I would do it during the main iteration.
In C++ this would be fine, but in python this will do copies of each PeptideID.
I make a comment on where I would add it.

axelwalter: First I put it in the main iteration (line 302), however, returning None (when there are no peptide hits and export_unidentified == False) does not work with np.fromiter(). Therefore I filtered them out in advance.

jpfeuffer: In this case, I would just "continue" the loop but I think we can leave it for now. Maybe add a TODO.

jpfeuffer: Sorry, I might have gotten it wrong. Yes we would need to think about it.

axelwalter: Ok. Is it possible to continue here since we pass only one PeptideIdentification at a time and it is not really a loop?

jpfeuffer: I just checked and yes you can just do "return" without argument.

jpfeuffer: I added a commit

axelwalter: Nice!

jpfeuffer: Not sure if the iteration for counting is worth it, but we have to benchmark.

timosachsenberg:
woohhoo!

timosachsenberg:
I think peptide evidences are not exported (e.g, the protein accessions and start, end in the protein)

Issue #5619, opened 2021-10-27
OpenMS binaries are too large


hroest:
I realized that in the latest release (2.7) we produce 532 MB of binaries in the `bin` folder when statically linking. The reason for this is that the majority of our binaries statically link the xercesc library and include about 3.8 MB of code, in our latest release this affected 139 binaries accounting for around 500 MB of the size of the folder. This also affects the latest release OpenMS-2.7.0-Debian-Linux-x86_64.deb which *should* in principle have been dynamically linked but wasn't, looking at the file with `ar x` and extracting `data.tar.gz` I find 546 MB in the `bin` folder. 

I see two solutions to this
- dynamically link our binaries (when I do this, I get around 20 MB for the binaries)
- remove the need for linking `xercesc` by keeping it inside the `OpenMS::Internal` namespace

I dont know how feasible the first solution is, especially on Windows but we are clearly duplicating a lot of code by linking it into 140 different binaries.

hroest:
Yes, have a look https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h -- in its current implementation there is no way to separate XMLHandler from `xercesc` since it inherits from `xercesc::DefaultHandler` and that is how the parser interface works, you have to inherit from xerces. We therefore have to hide the implementation by either hiding the XMLFile itself through `FileHandler` or by hiding the implementation in a "Handler" class.

hroest:
> Totally agree on your second point. So the catches are xerces objects and full declarations in our public API right?

@jpfeuffer yes this is what gets included

https://github.com/OpenMS/OpenMS/blob/8fb42eca245b70bc79818a4f37e6243f67225c1a/src/openms/include/OpenMS/FORMAT/HANDLERS/XMLHandler.h#L45-L48

so a whole bunch of stuff in our public API and requiring downstream tools to link against xerces to make sure these symbols are defined. Also we inherit from `xercesc::DefaultHandler` so there is no way to hide this in the .cpp file

jpfeuffer:
Ok so switch from inheritance to composition and expose only functions that do not have xerces types?

And put some of the templated helper functions into cpp by specializing on the few datatypes we need?

jpfeuffer:
But I guess we *have to* inherit from the Xerces handler class, right? I found no other pattern to use Xerces. This means we have to hide our Handler completely.

jpfeuffer:
[...]

PR #5626, opened 2021-10-29
Feature/separate interfaces

hroest:
# Description

This is a subset of the changes from #5613 that I made when trying to separate OpenMS into different libraries. The main thing this PR does is to remove the dependency of various parts of OpenMS on xercesc which gets pulled in with `XMLHandler`. Unfortunately, including `XMLHandler.h` is quite common in OpenMS which leads to a large number of cross-dependencies and prevents us from using clean interfaces to interact with files. 

One part of this PR is to introduce a new interface for FeatureXML and ConsensusXML and move the implementation into the `OpenMS::Internal` namespace

Secondly, it uses the `FileHandler` as a clean interface to load and write files

# Checklist:


jpfeuffer:
Very nice. Is this enough to completely hide Xerces from the interface or is there more to be done?

The change in loading might break a lot of pyopenms workflows but in my opinion it is worth it. We probably should do an increase in the major version then.

hroest:
> Very nice. Is this enough to completely hide Xerces from the interface or is there more to be done?

there is more to be done, this is just a first step for the worst offenders:

```
$ grep XMLHandler.h OpenMS/FORMAT/*.h
OpenMS/FORMAT/CVMappingFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/IdXMLFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/MascotXMLFile.h:#include <OpenMS/FORMAT/HANDLERS/MascotXMLHandler.h>
OpenMS/FORMAT/OMSSAXMLFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/PepXMLFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/PepXMLFileMascot.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/ProtXMLFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/QcMLFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/TransformationXMLFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
OpenMS/FORMAT/XTandemXMLFile.h:#include <OpenMS/FORMAT/HANDLERS/XMLHandler.h>
```

basically we would have to do one of two things for all of these: write wrapper functions in FileHandler or separate implementation (with XMLHandler) from interface. Quite a bit of work but it should also help with #5619 and make sure that downstream consumers dont pull in stuff they dont need/want. 

> The change in loading might break a lot of pyopenms 

as long as we keep the old interfaces around, this should not actually be a problem

src/openms/source/ANALYSIS/DECHARGING/FeatureDeconvolution.cpp
```diff
@@ -1039,8 +1039,7 @@ namespace OpenMS
     FeatureMapType fm_missing;
     cl.suggestMissingFeatures(fm_out, cons_map, fm_missing);
 
-    FeatureXMLFile fmf;
-    fmf.store("fm_missing.featureXML", fm_missing);
+    FileHandler.storeFeatures("fm_missing.featureXML", fm_missing);
```
timosachsenberg, line 1042: maybe worth to dicuss: is it a better API for a user to have this moved to a FileHandler class? Wouldn't it be more natural to have such a method in FeatureMap? Separation is still be possible if I am not mistaken.

timosachsenberg: or e.g., keeping the old interface with FeatureXMLFile etc. that uses some internal FeatureXMLFileHandler

hroest: > e.g., keeping the old interface with FeatureXMLFile, so code doesn't break and call some internal FeatureXMLFileHandler

this is what we do for `MzMLFile` and `MzMLFileHandler` and what I implemented here in this PR as well, separating the interface `FeatureXMLFile` and `FeatureXMLFileHandler`. But the question that makes sense is whether we need 2 interfaces, `FileHandler` and `MzMLFile`

timosachsenberg: Ok I see. I agree - having a FeatureXMLFile interface looks a bit cleaner designwise compared to having only a "FileHandler" class. But no strong opinion here.

timosachsenberg: But let's keep the two interfaces for now

hroest:
@timosachsenberg per cdash all tests pass and builds work: http://cdash.openms.de/index.php?project=OpenMS&date=2021-10-29 should we merge?

timosachsenberg:
rebuild jenkins

timosachsenberg:
Thanks!

PR #5627, opened 2021-10-29
[FIX] use static inline for namespace

hroest:
trying to prevent some of the clang warnings on cdash


timosachsenberg:
I think it worked :)

PR #5629, opened 2021-11-01
change -use_ms1_traces to -enable_ms1 in OpenSwathWorkflow docs

jcharkow:
# Description
To my knowledge, -use_ms1_traces parameter has been deprecated and replaced with -enable_ms1. Update the documentation to reflect this. 

# Checklist:

# Note:


jpfeuffer:
What about https://github.com/OpenMS/OpenMS/blob/e66b2c45ae29fab5fb97e4c4d0e69e75e1849b91/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp#L480

Maybe we should change every occurrence of use_ms1_traces to make sure that internal parameter names and command line parameters match. Seems like they have a 1:1 correspondence in this case.

jcharkow:
Yes I agree, I'll look for other instances in which use_ms1_traces is used.

jcharkow:
I replaced all occurrences of `use_ms1_traces` with `enable_ms1` in the documentation and I also replaced the variable occurrences to stay consistent (`use_ms1_traces` and `use_ms1_traces_` with `enable_ms1`and `enable_ms1_`). I don't think this should cause any issues however I am not too experienced with C++.

PR #5630, opened 2021-11-02
[FIX] MetaboliteSpectralMatcher

jpfeuffer:


src/utils/MetaboliteSpectralMatcher.cpp
```diff
@@ -134,7 +134,7 @@ class TOPPMetaboliteSpectralMatcher :
 
     MzMLFile mz_file;
     mz_file.setLogType(log_type_);
-    std::vector<Int> ms_level(1,2);
+    std::vector<Int> ms_level{1,2};
```
timosachsenberg, line 137: is this the same as ms_level = {1,2} ?

timosachsenberg, line 137:
```suggestion
    std::vector<Int> ms_level{2};
```

jpfeuffer: Good question. I'm always confused. Maybe equals is clearer here.

If that was intended anyway.
If we want to load Ms2 only, then the error message should be changed and a less ambiguous code snippet used/ with comment

PR #5633, opened 2021-11-03
Isotopic distribution approximation

emilpaulitz:
# Description

Adds the static methods approximateFromPeptideWeight and approximateIntensities to the CoarseIsotopePatternGenerator class! They roughly approximate an isotope distribution or only a vector of corresponding intensities for a given monoisotopic weight. The approximation uses a Poisson distribution with λ = 1/1800, as proposed by [Bellew et al](https://dx.doi.org/10.1093/bioinformatics/btl276). 

Direct comparison of the first 6 intensities generated for a mass of 1000:
Comparison_Approximations2

The following KL divergences are observed when generating 20 peaks:
Comparison_Approximations
for comparison, when comparing an observed pattern with a theoretical ground truth, the observed pattern is said to be an isotopic pattern if the KL between the two is below 0.05 for 2 peaks and below 0.6 for >=6 peaks by [Guo Ci Teo et al](https://dx.doi.org/10.1021/acs.jproteome.0c00544).

A speed comparison yielded, that 100,000 iterations of generating 50 isotopic peaks takes 4067.8 ms with the estimateFromPeptideWeight function, 71.1 ms with the approximateFromPeptideWeight, and 40.4 ms with approximateIntensities.

I also changed the method in Deisotoper.cpp (my previous PR).

# Checklist:


timosachsenberg:
nice

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
```diff
@@ -179,6 +179,71 @@ namespace OpenMS
...
+    // for k=0, non-normalized value is always 1
+    result[0] = Peak1D(mass, 1.0);
+
+    float curr_intensity;
+    for (UInt k = 1; k < num_peaks; ++k)
+    {
+      curr_power *= factor;
+      curr_factorial *= k;
+      curr_intensity = curr_power / curr_factorial;
+      result[k] = Peak1D(mass + (k * OpenMS::Constants::NEUTRON_MASS_U),
```
timosachsenberg, line 206: maybe write how mass is calculated in the function description

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
```diff
@@ -154,6 +154,24 @@ namespace OpenMS
...
+    static IsotopeDistribution approximateFromPeptideWeight(double mass, int num_peaks = 20);
+
+    /**
+       @brief roughly approximate intensity distribution of peptidic isotope patterns from monoisotopic weight using Poisson distribution.
+       Foundation from: Bellew et al, https://dx.doi.org/10.1093/bioinformatics/btl276
+
+       @param average_weight: m/z of monoisotopic peak (with charge = 1) to approximate the distribution of intensities for
+       @param num_peaks: How many peaks should be generated (independent of this->max_isotope)
+    */
+    static std::vector<float> approximateIntensities(double mass, int num_peaks = 20);
```
hroest, line 173: why not use double here to be more accurate? or a template argument so that a client could use a more accurate type?

hroest:
nice, its 100x faster! 

But can you add some of what you wrote in your PR also into the documentation? Otherwise it is hard for people to understand when to use the approximations and when to use the full isotopic distribution functions; the tradeoffs with speed vs accuracy would be helpful if documented.

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
```diff
@@ -179,6 +179,71 @@ namespace OpenMS
...
+        std::isinf(curr_intensity) ? 0 : curr_intensity);// at some point, curr_intensity will become too small for float (which is the intensity type)
+    }
+
+    result.renormalize();
+
+    return result;
+  }
+
+  std::vector<float> CoarseIsotopePatternGenerator::approximateIntensities(double mass, int num_peaks)
```
cbielow, line 215: `num_peaks` should be unsigned, otherwise calling it with `-1` and funny things will happen :)
Maybe even make it `uint8_t`, if we feel that more than 255 peaks are never needed (and that many peaks are only useful for very large molecules, for which this approximation does not work anymore)

cbielow (commented):
nice!

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
timosachsenberg: I think you mean because of a potential float overflow? Accuracy-wise, I think float is fine (natural abundances fluctuate a bit on earth anyway). 
@emilpaulitz do you know if the differences are caused by assuming a Poisson distribution (as opposed to calculating convolution of binomial distributions + convolutions) or by numeric inaccuracies?
Nevertheless, I think we could add a template here and use e.g. the double version in the deisotoper?

timosachsenberg: @emilpaulitz does double make a noticeable speed difference? if not we could go with double instead I guess. Having a template is maybe a bit overkill here (also makes wrapping more difficult).

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
timosachsenberg: uint would be better, agree.

But maybe don't restrict it to 8 bit
1. one might actually want to use that method to compare the error between the approximation and the exact calculation for larger molecules
2. I don't know if we properly wrap the 8bit uint type in python. Anyone knows?

so going for uint would be a "practical" solution :)

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
emilpaulitz: Done

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.h
cbielow: I would suspect (tdb), that the approximation error is far greater than the loss of precision when comparing float to double.
But that is easy to check, I guess.

emilpaulitz: The differences are definitely caused by assuming a poisson distribution, as estimateFromPeptideWeight (which returns IsotopeDistribution) also uses float for the intensites. In fact, the differences seem to start at the 3rd digit the latest, and if I recall correctly, float has 6 accurate digits. But switching to double does not seem to make a difference time-wise, so I did.

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
emilpaulitz: Changed it to UInt

emilpaulitz:
Alright, also updated the description. Found an issue with overflowing int and shortened the calculations to avoid that, now it is even faster (but yields the same values). The values fluctuate on my machine, I got 60-75 and 100-175 times faster than estimateFromPeptideWeight.

It is ready from side.

src/tests/class_tests/openms/source/CoarseIsotopeDistribution_test.cpp
```diff
@@ -291,6 +291,64 @@ START_SECTION(IsotopeDitribution CoarseIsotopePatternGenerator::estimateFromPept
...
+  }
+}
+END_SECTION
+
+START_SECTION(IsotopeDitribution CoarseIsotopePatternGenerator::approximateIntensities(double mass, int num_peaks))
+{
+  std::vector<Int> masses_to_test = {20, 300, 1000, 2500};
+  for (auto mass = masses_to_test.begin(); mass != masses_to_test.end(); ++mass)
+  {
+    std::vector<double_t> approximation = CoarseIsotopePatternGenerator::approximateIntensities(*mass);
```
cbielow, line 328: `double_t` ?
Please don't....

timosachsenberg: I can change... gitpod is just click away :)

emilpaulitz: Was not aware that is an issue... Thanks @timosachsenberg for fixing!

cbielow: Its not an issue per se, but it violates "Do not write surprising code" :) and it potentially makes testing a lot harder (or fail), if some platform or someone decides to set this strange FP.. flag, which changes the underlying type of double.

timosachsenberg:
Thanks!

PR #5640, opened 2021-11-05
Fix tool scan in TOPPView

timosachsenberg:
# Description

Async may spawn one thread per topp tool. That's no problem but apparently 
spawning hundreds of processes (=tool execution for ini file reading, one for each thread) seems to hit
a hard limit in file handles or available pipes.

Simplest form of process pool implementation I can come up with :)

# Checklist:

# Note:

### DIFF
--- a/src/openms_gui/source/VISUAL/TVToolDiscovery.cpp
+++ b/src/openms_gui/source/VISUAL/TVToolDiscovery.cpp
@@ -100,6 +100,8 @@ namespace OpenMS
 
   Param TVToolDiscovery::getParamFromIni_(const std::string &tool_name)
   {
+    static std::atomic<int> running_processes{0}; // used to limit the number of parallel processes
+
     // Temporary file path and arguments
     String path = File::getTemporaryFile();
     String working_dir = path.prefix(path.find_last_of('/'));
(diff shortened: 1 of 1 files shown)


jpfeuffer:
Weren't there multiple suggestions when the PR was open? Why don't we set a fixed number of threads and just distribute the tasks on them (i.e. give them part of the tool list to check).

Your implementation might still spawn an arbitrary number of threads, with all but 6 waiting to access the int? Does it?
Maybe they don't open file handles anymore though.

timosachsenberg:
We previously used Qt for a threadpool implementation.
Then switched to std::async because of reviewer requests. Unfortunately, std c++ doesn't offer threadpools / has only very basic multi threading support.
Now we found out that the processes and open pipes/handles are a problem.
So we fixed that part with a process pool. Having hundreds or thousands of threads is not an issue though so I guess this solution is fine.

timosachsenberg:
In that case the problem occurred after about 80 processes were running in parallel.

timosachsenberg:
Also doesn’t seem detrimental to performance because XML parsing still happens in parallel

jpfeuffer:
I know. However, the last suggestions from Chris and me were such easy std::thread solutions.

timosachsenberg:
Hmm you mean the chunking proposed in https://github.com/OpenMS/OpenMS/pull/5256 ? I think it is possible but then you need another separate thread that fires up the chunks of threads right?  As you can’t do that in the gui thread. Not sure if easier but probably cleaner. On the other hand normal chunking is slower as it waits till a chunk is finished. Here we can fire the next process immediately when the last one was finished.

timosachsenberg:
Let’s come up with a proper thread pool solution at some point  and accept this as temporary fix?

timosachsenberg:
Let’s come up with a proper thread pool solution at some point  and accept this as temporary fix?

Issue #32, opened 2014-01-09
TOPPView save [32]


timosachsenberg:
Submitted by lars20070 on 2009-10-28 10:41:04

2D Bereich in TOPPview ausgeschnitten, dann mit Rechts Klick "save visible layer data" -> File Endung *.mzML wurde nicht angehaengen.

---

Commented by cbielow on 2009-11-09 13:40:58:
Not reproducible on Win64 with current svn version.
File ending is only omitted if "All files (*)" is selected as datatype.

---

Commented by andreas_bertsch on 2009-11-09 14:34:50:
Works on MacOSX and Windows variants. Does not work on Unix, because Qt uses a different dialog. 

Qt4.4 offers simple way, however we also support Qt4.3

-> Postponed to later release

---

Commented by andreas_bertsch on 2009-11-09 14:54:18:
moved  to 1.6

---

Commented by andreas_bertsch on 2009-11-13 14:48:24:
Finally fixed. If no correct file ending is given the correct one is appended, depending on the filter setting, mzXML, mzData and mzML (and also featureXML and consensusXML). If the filter setting of the peak save dialog is set to "All files (*)" mzML is written with .mzML as file ending

PR #5644, opened 2021-11-10
Perf mini

cbielow:
# Description

Mini example of performance opt.
This doubles the speed of the function...
see
https://quick-bench.com/q/eZKZ3QJM434xxYkUlJUMf5WuWps

# Note:


timosachsenberg:
I think I read that -ffast-math would allow the compiler to do that.
If that can be enabled for such performance critical code I think it might be better to offload that to the compiler.

cbielow:
just tried `-ffast-math` with OpenSwathWorkflow.
It crashes with `Error: Unexpected internal error (SWATH window #32 from swath_windows_file extends beyond the Swath window of the data. ` ... so there would be some work before even getting it to run.
:(

PR #5651, opened 2021-11-11
only detect features on MS1 (don't mix levels)

timosachsenberg:
# Description

Fixes issue 5650

# Checklist:

# Note:


jpfeuffer:
Nice, but depending on the access patterns, wouldn't it be better if those spectra are just skipped to avoid copying the full raw data?

timosachsenberg:
I think the data is handed down to the SWATH classes which probably don't (need to) implement that logic. If someone knows or can point me to a quick and clean way to add that I would be happy to do so.

timosachsenberg:
I will merge this as a quick fix.

jpfeuffer:
Also keep in mind that this copies everything, even though a user already provides filtered data.

timosachsenberg:
Yes, not great I admit, but copying was also done before. On the C++ side I think could add a move version.

hroest:
could we wrap a rvalue version of this function?

```
void setMSData(const PeakMap&& m);
```

timosachsenberg:
that would be one idea yes. I can add it along with the test for the other PR

PR #5656, opened 2021-11-15
Feature/pep evidences df

axelwalter:
# Description
Added the export of peptide evidences for the best peptide hit in peptide_identifications_to_df(). New columns for accession, start and end position of peptide evidences. They are comma separated similar to mzTab.

# Checklist:

# Note:

### DIFF
--- a/src/pyOpenMS/pyopenms/dataframes.py
+++ b/src/pyOpenMS/pyopenms/dataframes.py
@@ -294,7 +294,7 @@ def peptide_identifications_to_df(peps: List[PeptideIdentification], decode_onto
     clearMVs = [cv.getTerm(m).name if m.startswith("MS:") else m for m in decodedMVs]
     #cols = ["id", "RT", "mz", "score", "charge"] + decodedMVs
     clearcols = ["id", "RT", "mz", mainscorename, "charge", "protein_accession", "start", "end"] + clearMVs
-    coltypes = ['U100', 'f', 'f', 'f', 'i','U100', 'U100', 'U100'] + types
+    coltypes = ['U100', 'f', 'f', 'f', 'i','U1000', 'U1000', 'U1000'] + types
     dt = list(zip(clearcols, coltypes))
 
     def extract(pep):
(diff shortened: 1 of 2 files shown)


src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -293,20 +293,26 @@ def get_key(val):
     cv.loadFromOBO("psims", File.getOpenMSDataPath() + "/CV/psi-ms.obo")
     clearMVs = [cv.getTerm(m).name if m.startswith("MS:") else m for m in decodedMVs]
     #cols = ["id", "RT", "mz", "score", "charge"] + decodedMVs
-    clearcols = ["id", "RT", "mz", mainscorename, "charge"] + clearMVs
-    coltypes = ['U100', 'f', 'f', 'f', 'i'] + types
+    clearcols = ["id", "RT", "mz", mainscorename, "charge", "protein_accession", "start", "end"] + clearMVs
+    coltypes = ['U100', 'f', 'f', 'f', 'i','U100', 'U100', 'U100'] + types
```
timosachsenberg, line 297: I could imagine that these strings can be longer than 100 characters

src/pyOpenMS/tests/unittests/test000.py
```diff
@@ -2255,12 +2255,26 @@ def test_peptide_identifications_to_df():
...
+    p1.setRT(1243.56)
+    p1.setMZ(240.0)
+    p1.setScoreType("ScoreType")
+    p1.setHigherScoreBetter(False)
+    p1.setIdentifier("IdentificationRun2")
+
+    peps.append(p1)
+
+    assert pyopenms.peptide_identifications_to_df(peps).shape == (2,10)
+    assert pyopenms.peptide_identifications_to_df(peps, export_unidentified=False).shape == (1,10)
```
timosachsenberg, line 2277: awesome but can you add a test with two different peptide evidences and test the contents of the dataframe?
e.g. if the concatenation worked as expected

src/pyOpenMS/pyopenms/dataframes.py
axelwalter: Would getting the maximum string length in each pep evidence column and setting the dtype accordingly work?

timosachsenberg: yes that would work.
I just wonder if U100 saves memory in cases that we have e.g., only U20 strings.
e.g., would str be more memory efficient?

timosachsenberg:
Thanks! looks good. I think we can now think about an extension to the readthedocs page how to use the dfs

PR #5660, opened 2021-11-17
Patch/use td library

SGSSGene:
# Draft Note:
This is currently a draft and demonstrates how tool description library use would look like.
It is currently based on this PR https://github.com/deNBI-cibi/tool_description_lib/pull/5

# Description

This integrates the tool description library (tdl) as a replacement for the Param, ParamValue and ParamCTDFile data structures.

# Todos:
- [ ] Wait until tdl is actually ready to be shipped
- [x] Check that throwing exception result in the expected behavior. Catching mechanism might be broken...(R: added OpenMS exception base class to TDL)
- [ ] There might be a API problem for windows when shipped as an DLL since we use different macros. (Not sure, must be checked)
- [ ] Replace proper submodule URL with the correct one
- [ ] Output/Error streams, what should we do with those?
- [ ] Make sure the newest changes of OpenMS are in TDL


src/openms_gui/source/VISUAL/TVToolDiscovery.cpp
```diff
@@ -41,6 +41,7 @@
 #include <QCoreApplication>
 
 #include <iostream>
+#include <thread>
```
SGSSGene, line 44: This addition was accidental (But might be necessary and should have its own PR)

timosachsenberg: k thanks. I can add this manually to develop

SGSSGene: So when I compiled OpenMS directly from develop without any changes, I got an error. I am using gcc11. I havn't looked into it.

timosachsenberg: Yeah these include hierarchies inside STL seem to change once in a while.
I added that line on develop.

timosachsenberg:
rebuild jenkins

jpfeuffer:
```
CMake Error at src/openms/thirdparty/CMakeLists.txt:49 (add_subdirectory):
  The source directory

    /home/jenkins/ws/openms/PR/bldtst/a9154670/source/src/openms/thirdparty/tool_description_lib

  does not contain a CMakeLists.txt file.
```

jpfeuffer:
See this minimal CMakelists that we created for another h-only lib.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/thirdparty/evergreen/src/CMakeLists.txt

SGSSGene:
rebuild jenkins

.github/workflows/openms_clang_format.yml
```diff
@@ -29,5 +29,5 @@ jobs:
       if: steps.changed_files.outputs.any_modified == 'true'
       uses: DoozyX/clang-format-lint-action@v0.13
       with:
-        source: ${{ steps.changed_files.outputs.all_modified_files }}
+        source: ${{ steps.changed_files.outputs.all_changed_files }}
```
jpfeuffer, line 32: Does that include added files? I'm never sure.

SGSSGene: Yes. This is not clear from their names (they should rethink all the names) `all_modified_files` is "ACMRD" and `all_changed_files` is "ACMR"

A: Added
C: Copied
M: Modified
R: Renamed
D: Deleted

Reference: https://github.com/tj-actions/changed-files

jpfeuffer: Perfect, thank you.

SGSSGene:
This is being abandoned in favor of PR #6156

PR #5673, opened 2021-11-23
clang-tidy add const to methods and replace size() == 0 with empty

timosachsenberg:
# Description
readability-container-size-empty
readability-make-member-function-const

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
@@ -367,6 +367,4 @@ public:
       bool finalized_;
   };
 
-  // needed for the heap
-  bool operator<(const QTCluster& q1, const QTCluster& q2);
 } // namespace OpenMS
--- a/src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
@@ -251,7 +251,7 @@ public:
     Size size() const;
 
     /// Compare by quality
-    bool operator<(const QTCluster& cluster);
+    bool operator<(const QTCluster& cluster) const;
 
     /**
      * @brief Adds a new element/neighbor to the cluster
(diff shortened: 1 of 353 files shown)


src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
```diff
@@ -367,6 +367,4 @@ namespace OpenMS
       bool finalized_;
   };
 
-  // needed for the heap
-  bool operator<(const QTCluster& q1, const QTCluster& q2);
```
timosachsenberg, line 371: @cbielow this one failed with (ambigious method) after adding const to the unary operator< 
To me, it seems as it can be safely removed.

timosachsenberg:
also get a failed compile here:
E:\jenkins\ws\openms\PR\bldtst\9447518b\source\src\openms_gui\include\OpenMS/VISUAL/APPLICATIONS/TOPPViewBase.h(247,39): error C2556: 'OpenMS::LayerData *OpenMS::TOPPViewBase::getCurrentLayer(void) const': overloaded function differs only by return type from 'const OpenMS::LayerData *OpenMS::TOPPViewBase::getCurrentLayer(void) const'

src/openms/include/OpenMS/DATASTRUCTURES/QTCluster.h
cbielow: yes. Since you added `const` to `bool operator<(const QTCluster& cluster) const;`, this now takes part in overload resolution and there should even be an error - or at least a warning, since `operator<` is now ambiguous.

Strangely, I dont know when `bool operator<(const QTCluster& cluster);` would ever have been called...

timosachsenberg: after removal, the unary is now called when adding lazy to the cluster (which previously called the binary operator).

timosachsenberg:
compiles on gitpod... let's see who complains

timosachsenberg:
ok there seem still to be some of the .empty() vs default constructor tests... will fix them

timosachsenberg:
Tests passed now

cbielow (approved):
i cannot look at `... !empty()` anymore :)

timosachsenberg:
> i cannot look at `... !empty()` anymore :)

Better than size() > 0 though ;)

timosachsenberg:
Thanks for taking the time to go through it. It is tiring I admit…

PR #5674, opened 2021-11-24
Fix race condition in TOPPView

cbielow:
# Description

 fixes a race condition when multiple threads call newFile() and the reference gets invalidated (happened in TOPPView, when requesting a temporary file for INI's)

# Note:

### DIFF
--- a/src/openms/source/SYSTEM/File.cpp
+++ b/src/openms/source/SYSTEM/File.cpp
@@ -814,12 +814,13 @@ namespace OpenMS
   {
   }
 
-  const String& File::TemporaryFiles_::newFile()
+  String File::TemporaryFiles_::newFile()
   {
     String s = getTempDirectory().ensureLastChar('/') + getUniqueName();
     std::lock_guard<std::mutex> _(mtx_);
     filenames_.push_back(s);
-    return filenames_.back();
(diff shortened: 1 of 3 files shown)


src/openms/source/SYSTEM/File.cpp
```diff
@@ -814,12 +814,13 @@ namespace OpenMS
   {
   }
 
-  const String& File::TemporaryFiles_::newFile()
+  String File::TemporaryFiles_::newFile()
   {
     String s = getTempDirectory().ensureLastChar('/') + getUniqueName();
     std::lock_guard<std::mutex> _(mtx_);
     filenames_.push_back(s);
-    return filenames_.back();
```
timosachsenberg, line 822: Why is this not handled by the lock?

cbielow: because it returned a **reference**.
The function ends, the lock is released, but the reference still exists. Then comes another thread and reallocates the vector. Reference becomes invalid and boom...

timosachsenberg: ahhh right of course

timosachsenberg (approved):

timosachsenberg:
there seems to be an issue: http://cdash.openms.de/viewBuildError.php?buildid=210499 I can take a quick look

PR #5680, opened 2021-11-24
Clang tidy performance type-promotion-in-math-fn,move-const-arg,unnecessary-copy-initialization

timosachsenberg:
# Description

performance-type-promotion-in-math-fn
performance-move-const-arg

Valuable check but could bring up race conditions that got hidden by copying:
performance-unnecessary-copy-initialization

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
+++ b/src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
@@ -199,7 +199,7 @@ namespace OpenMS
       identified_oligos_(std::move(other.identified_oligos_)),
       query_matches_(std::move(other.query_matches_)),
       query_match_groups_(std::move(other.query_match_groups_)),
-      current_step_ref_(std::move(other.current_step_ref_)),
+      current_step_ref_(other.current_step_ref_),
       // look-up tables:
       data_query_lookup_(std::move(other.data_query_lookup_)),
       parent_molecule_lookup_(std::move(other.parent_molecule_lookup_)),
--- a/src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
+++ b/src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
@@ -505,7 +505,7 @@ namespace OpenMS
           new_hits.push_back(std::move(hit));
         }
       }
-      it->setHits(std::move(new_hits));
+      it->setHits(new_hits);
     }
   }
 
--- a/src/openms/source/DATASTRUCTURES/DataValue.cpp
+++ b/src/openms/source/DATASTRUCTURES/DataValue.cpp
@@ -236,10 +236,10 @@ namespace OpenMS
   }
 
   DataValue::DataValue(DataValue&& rhs) noexcept :
-    value_type_(std::move(rhs.value_type_)),
-    unit_type_(std::move(rhs.unit_type_)),
-    unit_(std::move(rhs.unit_)),
-    data_(std::move(rhs.data_))
+    value_type_(rhs.value_type_),
+    unit_type_(rhs.unit_type_),
+    unit_(rhs.unit_),
+    data_(rhs.data_)
   {
     // clean up rhs, take ownership of data_
     // NOTE: value_type_ == EMPTY_VALUE implies data_ is empty and can be reset
--- a/src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
+++ b/src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
@@ -91,6 +91,7 @@ namespace OpenMS
     }
 
     // convert to vector
+    annotated_spectra.reserve(native_ids_annotated_spectra.size());
     for (const auto& it : native_ids_annotated_spectra)
     {
       annotated_spectra.emplace_back(std::move(it.second));
--- a/src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
+++ b/src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
@@ -94,7 +94,7 @@ namespace OpenMS
     annotated_spectra.reserve(native_ids_annotated_spectra.size());
     for (const auto& it : native_ids_annotated_spectra)
     {
-      annotated_spectra.emplace_back(std::move(it.second));
+      annotated_spectra.emplace_back(it.second);
     }
 
     return annotated_spectra;
--- a/src/openms/source/FORMAT/FASTAFile.cpp
+++ b/src/openms/source/FORMAT/FASTAFile.cpp
@@ -198,9 +198,9 @@ namespace OpenMS
     }
     ++entries_read_;
 
-    protein.identifier = std::move(id_);
-    protein.description = std::move(description_);
-    protein.sequence = std::move(seq_);
+    protein.identifier = id_;
+    protein.description = description_;
+    protein.sequence = seq_;
 
     return true;
   }
--- a/src/openms/source/FORMAT/HANDLERS/ConsensusXMLHandler.cpp
+++ b/src/openms/source/FORMAT/HANDLERS/ConsensusXMLHandler.cpp
@@ -258,7 +258,7 @@ namespace OpenMS::Internal
             act_index_tuple.setCharge(charge);
           }
 
-          act_cons_element_.insert(std::move(act_index_tuple));
+          act_cons_element_.insert(act_index_tuple);
         }
       }
       act_cons_element_.getPosition() = pos_;
--- a/src/openms/source/FORMAT/IdXMLFile.cpp
+++ b/src/openms/source/FORMAT/IdXMLFile.cpp
@@ -603,7 +603,7 @@ namespace OpenMS
       // sequence
       String tmp;
       optionalAttributeAsString_(tmp, attributes, "sequence");
-      prot_hit_.setSequence(std::move(tmp));
+      prot_hit_.setSequence(tmp);
 
       last_meta_ = &prot_hit_;
 
--- a/src/openms/source/FORMAT/MzTabFile.cpp
+++ b/src/openms/source/FORMAT/MzTabFile.cpp
@@ -2918,10 +2918,12 @@ namespace OpenMS
     }
 
     vector<const PeptideIdentification*> pep_ids_ptr;
-    for (const PeptideIdentification& pi : peptide_identifications) { pep_ids_ptr.push_back(&pi); }
+    pep_ids_ptr.reserve(peptide_identifications.size());
+for (const PeptideIdentification& pi : peptide_identifications) { pep_ids_ptr.push_back(&pi); }
 
     vector<const ProteinIdentification*> prot_ids_ptr;
-    for (const ProteinIdentification& pi : protein_identifications) { prot_ids_ptr.push_back(&pi); }
+    prot_ids_ptr.reserve(protein_identifications.size());
+for (const ProteinIdentification& pi : protein_identifications) { prot_ids_ptr.push_back(&pi); }
 
     ofstream tab_file;
     tab_file.open(filename, ios::out | ios::trunc);
--- a/src/openms/source/FORMAT/OSWFile.cpp
+++ b/src/openms/source/FORMAT/OSWFile.cpp
@@ -304,7 +304,7 @@ namespace OpenMS
       {
         int id = Sql::extractInt(stmt, I_PROTID);
         accession = Sql::extractString(stmt, I_ACCESSION);
-        swath_result.addProtein(OSWProtein(std::move(accession), id, {}));
+        swath_result.addProtein(OSWProtein(accession, id, {}));
         rc = Sql::nextRow(stmt, rc); // next row
       }
     }
@@ -672,7 +672,7 @@ namespace OpenMS
           Sql::extractFloat(stmt, COLIDs::PRODUCT_MZ),
           Sql::extractChar(stmt, COLIDs::TYPE),
           Sql::extractInt(stmt, COLIDs::DECOY));
-        swath_result.addTransition(std::move(tr));
+        swath_result.addTransition(tr);
         rc = Sql::nextRow(stmt);
       }
       sqlite3_finalize(stmt);
--- a/src/openms/source/IONMOBILITY/IMDataConverter.cpp
+++ b/src/openms/source/IONMOBILITY/IMDataConverter.cpp
@@ -78,7 +78,7 @@ namespace OpenMS
     // fill up the PeakMaps by moving spectra from the input PeakMap
     for (const MSSpectrum& it : exp)
     {
-      split_peakmap[cv2index[it.getDriftTime()]].addSpectrum(std::move(it));
+      split_peakmap[cv2index[it.getDriftTime()]].addSpectrum(it);
     }
 
     return split_peakmap;
--- a/src/openms/source/KERNEL/Feature.cpp
+++ b/src/openms/source/KERNEL/Feature.cpp
@@ -67,7 +67,7 @@ namespace OpenMS
   Feature::Feature(Feature&& feature) noexcept :
     BaseFeature(std::move(feature)),
     convex_hulls_(std::move(feature.convex_hulls_)),
-    convex_hulls_modified_(std::move(feature.convex_hulls_modified_)),
+    convex_hulls_modified_(feature.convex_hulls_modified_),
     convex_hull_(std::move(feature.convex_hull_)),
     subordinates_(std::move(feature.subordinates_))
   {
@@ -204,7 +204,7 @@ namespace OpenMS
     BaseFeature::operator=(std::move(rhs));
     std::copy(rhs.qualities_, rhs.qualities_ + 2, qualities_);
     convex_hulls_           = std::move(rhs.convex_hulls_);
-    convex_hulls_modified_  = std::move(rhs.convex_hulls_modified_);
+    convex_hulls_modified_  = rhs.convex_hulls_modified_;
     convex_hull_            = std::move(rhs.convex_hull_);
     subordinates_           = std::move(rhs.subordinates_);
 
--- a/src/openms/source/METADATA/ExperimentalDesign.cpp
+++ b/src/openms/source/METADATA/ExperimentalDesign.cpp
@@ -289,6 +289,7 @@ namespace OpenMS
       for (unsigned u : sample_section_.getSamples())
       {
         std::vector<String> valuesToHash{};
+        valuesToHash.reserve(factors.size());
         for (const String& fac : factors)
         {
           valuesToHash.emplace_back(sample_section_.getFactorValue(u, fac));
@@ -349,7 +350,8 @@ namespace OpenMS
       for (unsigned u : sample_section_.getSamples())
       {
         std::vector<String> valuesToHash{};
-        for (const String& fac : nonRepFacs)
+        valuesToHash.reserve(nonRepFacs.size());
+for (const String& fac : nonRepFacs)
         {
           valuesToHash.emplace_back(sample_section_.getFactorValue(u, fac));
         }
--- a/src/openms/source/METADATA/MetaInfoInterface.cpp
+++ b/src/openms/source/METADATA/MetaInfoInterface.cpp
@@ -58,7 +58,7 @@ namespace OpenMS
 
   /// Move constructor
   MetaInfoInterface::MetaInfoInterface(MetaInfoInterface&& rhs) noexcept :
-    meta_(std::move(rhs.meta_))
+    meta_(rhs.meta_)
   {
     // take ownership
     rhs.meta_ = nullptr;
--- a/src/openms/source/METADATA/PeptideHit.cpp
+++ b/src/openms/source/METADATA/PeptideHit.cpp
@@ -100,7 +100,7 @@ namespace OpenMS
     MetaInfoInterface(std::move(source)), // NOTE: rhs itself is an lvalue
     sequence_(std::move(source.sequence_)),
     score_(source.score_),
-    analysis_results_(std::move(source.analysis_results_)),
+    analysis_results_(source.analysis_results_),
     rank_(source.rank_),
     charge_(source.charge_),
     peptide_evidences_(std::move(source.peptide_evidences_)),
--- a/src/openms_gui/source/VISUAL/TVIdentificationViewController.cpp
+++ b/src/openms_gui/source/VISUAL/TVIdentificationViewController.cpp
@@ -931,7 +931,7 @@ namespace OpenMS
     LayerData& current_layer = current_canvas->getCurrentLayer();
     const SpectrumType& current_spectrum = current_layer.getCurrentSpectrum();
 
-    AASequence aa_sequence = ph.getSequence();
+    const AASequence& aa_sequence = ph.getSequence();
 
     // get measured spectrum indices and spectrum
     Size current_spectrum_layer_index = current_canvas->getCurrentLayerIndex();
--- a/src/tests/class_tests/openms/source/DRange_test.cpp
+++ b/src/tests/class_tests/openms/source/DRange_test.cpp
@@ -90,7 +90,7 @@ DRange<2> r(p1,p2);
 //do not modify this range, it is used in many tests
 
 START_SECTION(DRange(const DRange& range))
-	DRange<2> r2(r);
+	const DRange<2>& r2(r);
 	TEST_REAL_SIMILAR(r2.minPosition()[0],-1.0f);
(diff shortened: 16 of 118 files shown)


src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -505,7 +505,7 @@ namespace OpenMS
           new_hits.push_back(std::move(hit));
         }
       }
-      it->setHits(std::move(new_hits));
+      it->setHits(new_hits);
```
timosachsenberg, line 508: we don't have setHits with &&

src/openms/source/FORMAT/FASTAFile.cpp
```diff
@@ -198,9 +198,9 @@ namespace OpenMS
     }
     ++entries_read_;
 
-    protein.identifier = std::move(id_);
-    protein.description = std::move(description_);
-    protein.sequence = std::move(seq_);
+    protein.identifier = id_;
+    protein.description = description_;
```
timosachsenberg, line 202: maybe worth to check why clang-tidy did remove the std::move here (aka has no effect)

src/openms/source/FORMAT/HANDLERS/ConsensusXMLHandler.cpp
```diff
@@ -258,7 +258,7 @@ namespace OpenMS::Internal
             act_index_tuple.setCharge(charge);
           }
 
-          act_cons_element_.insert(std::move(act_index_tuple));
+          act_cons_element_.insert(act_index_tuple);
```
timosachsenberg, line 261: also here... quick check might be good

src/openms/source/FORMAT/IdXMLFile.cpp
```diff
@@ -603,7 +603,7 @@ namespace OpenMS
       // sequence
       String tmp;
       optionalAttributeAsString_(tmp, attributes, "sequence");
-      prot_hit_.setSequence(std::move(tmp));
+      prot_hit_.setSequence(tmp);
```
timosachsenberg, line 606: also here

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2918,10 +2918,12 @@ namespace OpenMS
     }
 
     vector<const PeptideIdentification*> pep_ids_ptr;
-    for (const PeptideIdentification& pi : peptide_identifications) { pep_ids_ptr.push_back(&pi); }
+    pep_ids_ptr.reserve(peptide_identifications.size());
+for (const PeptideIdentification& pi : peptide_identifications) { pep_ids_ptr.push_back(&pi); }
```
timosachsenberg, line 2922:
```suggestion
    for (const PeptideIdentification& pi : peptide_identifications) { pep_ids_ptr.push_back(&pi); }
```

src/openms/source/FORMAT/MzTabFile.cpp
```diff
@@ -2918,10 +2918,12 @@ namespace OpenMS
...
 
     vector<const ProteinIdentification*> prot_ids_ptr;
-    for (const ProteinIdentification& pi : protein_identifications) { prot_ids_ptr.push_back(&pi); }
+    prot_ids_ptr.reserve(protein_identifications.size());
+for (const ProteinIdentification& pi : protein_identifications) { prot_ids_ptr.push_back(&pi); }
```
timosachsenberg, line 2926:
```suggestion
    for (const ProteinIdentification& pi : protein_identifications) { prot_ids_ptr.push_back(&pi); }
```

src/openms/source/FORMAT/OSWFile.cpp
```diff
@@ -304,7 +304,7 @@ namespace OpenMS
       {
         int id = Sql::extractInt(stmt, I_PROTID);
         accession = Sql::extractString(stmt, I_ACCESSION);
-        swath_result.addProtein(OSWProtein(std::move(accession), id, {}));
+        swath_result.addProtein(OSWProtein(accession, id, {}));
```
timosachsenberg, line 307: already temporary

src/openms/source/FORMAT/OSWFile.cpp
```diff
@@ -672,7 +672,7 @@ namespace OpenMS
           Sql::extractFloat(stmt, COLIDs::PRODUCT_MZ),
           Sql::extractChar(stmt, COLIDs::TYPE),
           Sql::extractInt(stmt, COLIDs::DECOY));
-        swath_result.addTransition(std::move(tr));
+        swath_result.addTransition(tr);
```
timosachsenberg, line 675: check?

src/openms/source/IONMOBILITY/IMDataConverter.cpp
```diff
@@ -78,7 +78,7 @@ namespace OpenMS
     // fill up the PeakMaps by moving spectra from the input PeakMap
     for (const MSSpectrum& it : exp)
     {
-      split_peakmap[cv2index[it.getDriftTime()]].addSpectrum(std::move(it));
+      split_peakmap[cv2index[it.getDriftTime()]].addSpectrum(it);
```
timosachsenberg, line 81: can't move const

src/openms/source/METADATA/ExperimentalDesign.cpp
```diff
@@ -349,7 +350,8 @@ namespace OpenMS
       for (unsigned u : sample_section_.getSamples())
       {
         std::vector<String> valuesToHash{};
-        for (const String& fac : nonRepFacs)
+        valuesToHash.reserve(nonRepFacs.size());
+for (const String& fac : nonRepFacs)
```
timosachsenberg, line 354:
```suggestion
        for (const String& fac : nonRepFacs)
```

src/openms/source/METADATA/MetaInfoInterface.cpp
```diff
@@ -58,7 +58,7 @@ namespace OpenMS
 
   /// Move constructor
   MetaInfoInterface::MetaInfoInterface(MetaInfoInterface&& rhs) noexcept :
-    meta_(std::move(rhs.meta_))
+    meta_(rhs.meta_)
```
timosachsenberg, line 61: check

src/openms/source/METADATA/PeptideHit.cpp
```diff
@@ -100,7 +100,7 @@ namespace OpenMS
     MetaInfoInterface(std::move(source)), // NOTE: rhs itself is an lvalue
     sequence_(std::move(source.sequence_)),
     score_(source.score_),
-    analysis_results_(std::move(source.analysis_results_)),
+    analysis_results_(source.analysis_results_),
```
timosachsenberg, line 103: check

src/openms_gui/source/VISUAL/TVIdentificationViewController.cpp
```diff
@@ -931,7 +931,7 @@ namespace OpenMS
     LayerData& current_layer = current_canvas->getCurrentLayer();
     const SpectrumType& current_spectrum = current_layer.getCurrentSpectrum();
 
-    AASequence aa_sequence = ph.getSequence();
+    const AASequence& aa_sequence = ph.getSequence();
```
timosachsenberg, line 934: thread safe?

src/tests/class_tests/openms/source/DataValue_test.cpp
```diff
@@ -223,16 +223,16 @@ START_SECTION((DataValue(const DataValue&)))
...
-  DataValue copy_of_p3(p3);
-  DataValue copy_of_p4(p4);
-  DataValue copy_of_p5(p5);
-  DataValue copy_of_p6(p6);
-  DataValue copy_of_p7(p7);
-  DataValue copy_of_p8(p8);
-  DataValue copy_of_p9(p9);
-  DataValue copy_of_p10(p10);
-  DataValue copy_of_p11(p11);
+  const DataValue& copy_of_p1(p1);
```
timosachsenberg, line 226: undo because we want to explicitly test copy

src/tests/class_tests/openms/source/ParamValue_test.cpp
```diff
@@ -200,16 +200,16 @@ START_TEST(ParamValue, "$Id$")
...
-                    ParamValue copy_of_p3(p3);
-                    ParamValue copy_of_p4(p4);
-                    ParamValue copy_of_p5(p5);
-                    ParamValue copy_of_p6(p6);
-                    ParamValue copy_of_p7(p7);
-                    ParamValue copy_of_p8(p8);
-                    ParamValue copy_of_p9(p9);
-                    ParamValue copy_of_p10(p10);
-                    ParamValue copy_of_p11(p11);
+                    const ParamValue& copy_of_p1(p1);
```
timosachsenberg, line 203: undo

src/tests/class_tests/openms/source/Param_test.cpp
```diff
@@ -934,7 +934,7 @@ p_src.setSectionDescription("test","sectiondesc");
 p_src.addTags("test:float", {"a", "b", "c"});
 
 START_SECTION((Param(const Param& rhs)))
-	Param p2(p_src);
+	const Param& p2(p_src);
```
timosachsenberg, line 937: undo

src/topp/ConsensusID.cpp
```diff
@@ -763,7 +763,7 @@ class TOPPConsensusID :
               auto acciter_inserted = iter_inserted.first->second.emplace(hit.getAccession());
               if (acciter_inserted.second)
               {
-                prot_ids[mzml_to_new_run_idx[original_file]].getHits().emplace_back(std::move(hit));
+                prot_ids[mzml_to_new_run_idx[original_file]].getHits().emplace_back(hit);
```
timosachsenberg, line 766: correct - can't move reference

src/topp/OpenSwathDecoyGenerator.cpp
```diff
@@ -280,12 +280,12 @@ class TOPPOpenSwathDecoyGenerator
       if (separate)
       {
         OPENMS_LOG_INFO << "Writing only decoys to file: " << out << std::endl;
-        targeted_merged = std::move(targeted_decoy);
+        targeted_merged = targeted_decoy;
```
timosachsenberg, line 283: not sure why move removed - check
also below

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
hroest: instead of removing this we should implement it
```suggestion
      it->setHits(std::move(new_hits)); // TODO implement setHits &&
```

src/openms/source/DATASTRUCTURES/DataValue.cpp
```diff
@@ -236,10 +236,10 @@ namespace OpenMS
...
 
   DataValue::DataValue(DataValue&& rhs) noexcept :
-    value_type_(std::move(rhs.value_type_)),
-    unit_type_(std::move(rhs.unit_type_)),
-    unit_(std::move(rhs.unit_)),
-    data_(std::move(rhs.data_))
+    value_type_(rhs.value_type_),
+    unit_type_(rhs.unit_type_),
+    unit_(rhs.unit_),
+    data_(rhs.data_)
```
hroest, line 242: not sure if I am a big fan of this: the type of these variables could change and `std::move` does not hurt here but it shows intent more clearly

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
```diff
@@ -91,9 +91,10 @@ namespace OpenMS
     }
 
     // convert to vector
+    annotated_spectra.reserve(native_ids_annotated_spectra.size());
     for (const auto& it : native_ids_annotated_spectra)
     {
-      annotated_spectra.emplace_back(std::move(it.second));
+      annotated_spectra.emplace_back(it.second);
```
hroest, line 97: @timosachsenberg @oliveralka this seems like it makes things worse?

src/openms/source/FORMAT/FASTAFile.cpp
hroest: @timosachsenberg I would not follow through with these, first it removes intent and secondly copying or moving-ing `std::string` clearly has a different effect! removing `std::move` may have a performance penalty here

src/openms/source/FORMAT/IdXMLFile.cpp
hroest: I guess there is no `void setSequence(const String && sequence);` but the solution is to implement this

https://github.com/OpenMS/OpenMS/blob/8594c2ad50b1e3d3263235ea592ed02049dbcb11/src/openms/include/OpenMS/METADATA/ProteinHit.h#L189

src/openms/source/FORMAT/OSWFile.cpp
hroest: ?? not really, the two lines are clearly different. The former could use a hypothetical move constructor `OSWProtein(String&&, int)` while the latter cannot. One would have to write one though

src/openms/source/FORMAT/OSWFile.cpp
hroest: requires implementation
```suggestion
        swath_result.addTransition(tr); // TODO implement
```

src/openms/source/METADATA/MetaInfoInterface.cpp
hroest, line 61:
```suggestion
    meta_(rhs.meta_) // TODO implement
```

src/openms_gui/source/VISUAL/TVIdentificationViewController.cpp
hroest: can `ph` be changed by another thread? otherwise I would say yes, it should be

src/tests/class_tests/openms/source/DRange_test.cpp
```diff
@@ -90,15 +90,15 @@ DRange<2> r(p1,p2);
 //do not modify this range, it is used in many tests
 
 START_SECTION(DRange(const DRange& range))
-	DRange<2> r2(r);
+	const DRange<2>& r2(r);
```
hroest, line 93: same here: undo as this will not test the copy constructor

src/topp/ConsensusID.cpp
hroest: @timosachsenberg I disagree, `hit` is not a reference but of type `ProteinHit` so it can be moved and here I think it *should* be moved 

https://github.com/OpenMS/OpenMS/blob/8594c2ad50b1e3d3263235ea592ed02049dbcb11/src/openms/include/OpenMS/METADATA/ProteinIdentification.h#L333

src/topp/OpenSwathDecoyGenerator.cpp
hroest: I guess because move assignment operator is missing?

https://github.com/OpenMS/OpenMS/blob/8594c2ad50b1e3d3263235ea592ed02049dbcb11/src/openms/include/OpenMS/ANALYSIS/TARGETED/TargetedExperiment.h#L103

hroest (changes requested):
this is overall very helpful and highlights current issues that need to be fixed, however this should NOT be merged as it would introduce problems and mask bugs / missing implementations

src/openms/source/FORMAT/FASTAFile.cpp
timosachsenberg: Sure. But according to clang it only removes the std::move if it has no effect. So it is worth to investigate why it gets removed while it shouldn’t

src/topp/ConsensusID.cpp
timosachsenberg: Then we need to check why clang things it can’t be moved

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
timosachsenberg: Move of const ref

timosachsenberg:
Yes this is not intended for merge but was the easiest way to generate a list of the clang tidy diagnostics

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
timosachsenberg: totally agree

timosachsenberg: how can we implement that?
    /// Sets the peptide hits
    void setHits(std::vector<PeptideHit>&& hits);
doesn't work with pyOpenMS
and having both will result in "cannot be overloaded".
But I thought that only applies to templates https://stackoverflow.com/questions/27835925/overload-between-rvalue-reference-and-const-lvalue-reference-in-template

src/openms/source/FORMAT/IdXMLFile.cpp
timosachsenberg: you mean void setSequence(String && sequence) ?

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
timosachsenberg: will address this is in  https://github.com/OpenMS/OpenMS/pull/5685

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
```diff
@@ -199,7 +199,7 @@ namespace OpenMS
       identified_oligos_(std::move(other.identified_oligos_)),
       query_matches_(std::move(other.query_matches_)),
       query_match_groups_(std::move(other.query_match_groups_)),
-      current_step_ref_(std::move(other.current_step_ref_)),
+      current_step_ref_(other.current_step_ref_),
```
timosachsenberg, line 202: check why

src/topp/OpenSwathDecoyGenerator.cpp
timosachsenberg: done in https://github.com/OpenMS/OpenMS/pull/5686/files

src/openms/source/FORMAT/IdXMLFile.cpp
timosachsenberg: will address this is in #5685

src/openms/source/KERNEL/Feature.cpp
```diff
@@ -67,7 +67,7 @@ namespace OpenMS
   Feature::Feature(Feature&& feature) noexcept :
     BaseFeature(std::move(feature)),
     convex_hulls_(std::move(feature.convex_hulls_)),
-    convex_hulls_modified_(std::move(feature.convex_hulls_modified_)),
+    convex_hulls_modified_(feature.convex_hulls_modified_),
```
timosachsenberg, line 70: this could be expensive... check

src/openms/source/METADATA/MetaInfoInterface.cpp
timosachsenberg: it's just a pointer MetaInfo* meta_;

src/openms/source/METADATA/PeptideHit.cpp
timosachsenberg: just pointer

src/openms/include/OpenMS/METADATA/ID/IdentificationData.h
hroest: I assume since this is a reference (eg an Iterator here), so this is probably ok?

src/openms/source/KERNEL/Feature.cpp
hroest: this is just a boolean, there is no need to use `std::move`. However if we remove it, the code looks a bit inconsistent

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
```diff
@@ -91,9 +91,10 @@ namespace OpenMS
     }
 
     // convert to vector
+    annotated_spectra.reserve(native_ids_annotated_spectra.size());
     for (const auto& it : native_ids_annotated_spectra)
```
hroest, line 95:
```suggestion
    for (auto& it : native_ids_annotated_spectra)
```

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
hroest: looking at this code, it seems that the const ref should not be const? @oliveralka

src/openms/source/FORMAT/HANDLERS/ConsensusXMLHandler.cpp
hroest: `ConsensusFeature::insert` are all missing a move version

src/openms/source/IONMOBILITY/IMDataConverter.cpp
```diff
@@ -78,7 +78,7 @@ namespace OpenMS
     // fill up the PeakMaps by moving spectra from the input PeakMap
     for (const MSSpectrum& it : exp)
```
hroest, line 79:
```suggestion
    for (MSSpectrum& it : exp)
```

src/openms/source/IONMOBILITY/IMDataConverter.cpp
hroest: @timosachsenberg this is clearly a bug as the intent was to move the spectra into the `split_peakmap`. Is there any sane way to check that something has been "moved from"? I am thinking here 

https://github.com/OpenMS/OpenMS/blob/3c4a078ad4687677393c138d4acb91bb444a7c7b/src/tests/class_tests/openms/source/IMDataConverter_test.cpp#L74

for example, a `MSSpectrum` and `MSExperiment` should always be empty after it has been moved-from, so we could check for that in the tests?

Also `addSpectrum(&&)` is missing

src/openms/source/IONMOBILITY/IMDataConverter.cpp
timosachsenberg: will add to https://github.com/OpenMS/OpenMS/pull/5685

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
timosachsenberg: will add to #5685

src/openms/source/KERNEL/Feature.cpp
timosachsenberg: agree - I would say we leave the std::move

src/openms/source/FORMAT/DATAACCESS/SiriusFragmentAnnotation.cpp
timosachsenberg: will add to #5685

src/openms/source/DATASTRUCTURES/DataValue.cpp
timosachsenberg: agree

src/openms/source/IONMOBILITY/IMDataConverter.cpp
timosachsenberg: addressed in https://github.com/OpenMS/OpenMS/pull/5685

src/openms/source/FORMAT/OSWFile.cpp
timosachsenberg: will address this is in #5685

src/openms/source/METADATA/MetaInfoInterface.cpp
timosachsenberg: just a pointer

src/openms/source/FORMAT/HANDLERS/ConsensusXMLHandler.cpp
timosachsenberg: will address this is in #5685

src/openms/source/FORMAT/OSWFile.cpp
timosachsenberg: ah right... maybe not worth the trouble?

src/openms/source/FORMAT/FASTAFile.cpp
timosachsenberg: not sure here. it is a normal std::string stored in a member.

src/openms/source/FORMAT/MzTabFile.cpp
timosachsenberg: will revisit those in othe rPR

PR #5681, opened 2021-11-24
Hide IsoSpec using pimpl and add some clang-tidy performance fixes on top

timosachsenberg:
# Description
clang-tidy fixes
cppcoreguidelines-virtual-class-destructor
misc-redundant-expression,
performance-trivially-destructible
performance-noexcept-move-constructor

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
@@ -241,7 +241,7 @@ public:
       **/
     IsoSpecTotalProbGeneratorWrapper(const EmpiricalFormula& formula, double p);
 
-    ~IsoSpecTotalProbGeneratorWrapper() = default;
+    ~IsoSpecTotalProbGeneratorWrapper();
 
     bool nextConf() final;
     Peak1D getConf() final;
@@ -306,7 +306,7 @@ public:
       **/
   IsoSpecThresholdGeneratorWrapper(const EmpiricalFormula& formula, double threshold, bool absolute);
 
-  ~IsoSpecThresholdGeneratorWrapper() = default;
+  ~IsoSpecThresholdGeneratorWrapper();
 
   bool nextConf() final;
   Peak1D getConf() final;
@@ -359,7 +359,7 @@ public:
       **/
   IsoSpecOrderedGeneratorWrapper(const EmpiricalFormula& formula);
 
-  ~IsoSpecOrderedGeneratorWrapper() = default;
+  ~IsoSpecOrderedGeneratorWrapper();
 
   inline bool nextConf() final;
   inline Peak1D getConf() final;
--- a/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
@@ -35,6 +35,7 @@
 #pragma once
 
 #include <vector>
+#include <memory>
 
 #include <OpenMS/CONCEPT/Types.h>
 #include <OpenMS/CONCEPT/Constants.h>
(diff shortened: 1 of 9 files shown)


src/openms/source/ANALYSIS/MRM/ReactionMonitoringTransition.cpp
```diff
@@ -145,7 +145,7 @@ namespace OpenMS
   }
 
   ReactionMonitoringTransition & ReactionMonitoringTransition::operator=(ReactionMonitoringTransition && rhs)
-  {
+ noexcept   {
```
hroest, line 148:
```suggestion
  {
```

hroest:
I see this is helpful, but how does clang know that something cannot throw?

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.cpp
```diff
@@ -214,6 +239,8 @@ namespace OpenMS
     return ID;
   }
 
+  IsoSpecThresholdWrapper::~IsoSpecThresholdWrapper() = default;
```
timosachsenberg, line 242: Today I learned that in case of these partial types we also need to put the default destructor into the cpp file, In header it fails.

timosachsenberg:
tests passed

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -236,14 +241,17 @@ namespace OpenMS
...
+
+    bool nextConf() final;
+    Peak1D getConf() final;
+    double getMass() final;
+    double getIntensity() final;
+    double getLogIntensity() final;
 
 protected:
-    IsoSpec::IsoLayeredGenerator ILG;
+    IsoSpecTotalProbGeneratorWrapper(const IsoSpecTotalProbGeneratorWrapper&) = delete;
```
cbielow, line 253: nitpicking here but:
would it make sense to move the copy ctor up into the area of all other ctors?! Just to have them all in one place, in case someone is checking them and sees that this one seems "missing"?

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -298,15 +306,18 @@ namespace OpenMS
...
+  bool nextConf() final;
+  Peak1D getConf() final;
+  double getMass() final;
+  double getIntensity() final;
+  double getLogIntensity() final;
 
 
 protected:
-  IsoSpec::IsoThresholdGenerator ITG;
+  IsoSpecThresholdGeneratorWrapper(const IsoSpecThresholdGeneratorWrapper&) = delete;
```
cbielow, line 319: same here

cbielow:
> I see this is helpful, but how does clang know that something cannot throw?

Do you mean noexcept? It cannot. We are just promising it does not (or bare the consequences).

cbielow:
looks fine to me

src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsoSpecWrapper.h
```diff
@@ -45,17 +46,21 @@
 #include <OpenMS/CHEMISTRY/EmpiricalFormula.h>
 #include <OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsotopeDistribution.h>
 
-
 // Override IsoSpec's use of mmap whenever it is available
 #define ISOSPEC_GOT_SYSTEM_MMAN false
 #define ISOSPEC_GOT_MMAN false
 #define ISOSPEC_BUILDING_OPENMS true
```
hroest, line 52: I assume you then also dont need these definitions in the header any more?

timosachsenberg: Yes I will remove them

timosachsenberg:
rebuild jenkins

timosachsenberg:
passed

hroest:
> > I see this is helpful, but how does clang know that something cannot throw?
> 
> Do you mean noexcept? It cannot. We are just promising it does not (or bare the consequences).

@cbielow @timosachsenberg yes exactly that was my point, how can clang-tidy automatically add `noexcept` -- shouldnt we check this manually?

timosachsenberg:
I think it checks all called functions for throw. If there is none it adds noexcept.

PR #5682, opened 2021-11-25
GUI Layer refactoring -- preparations

cbielow:
# Description

This PR adds a layer hierarchy in preparation of disentangling the LayerData datastructure for TOPPView GUI, which is currently a wild mixture of at least 5 different classes.
A lot more refactoring is needed, but to keep the PRs reviewable, this is a first chunk.

# Note:


src/openms_gui/include/OpenMS/VISUAL/LayerData.h
```diff
@@ -32,6 +32,8 @@
 // $Authors: Marc Sturm $
 // --------------------------------------------------------------------------
 
+#error No not use this header...
```
timosachsenberg, line 35: ?

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -2568,13 +2538,17 @@ namespace OpenMS
...
         // only the selected row can be dragged => the source layer is the selected layer
-        LayerData& layer = getActiveCanvas()->getCurrentLayer();
+        LayerDataBase& layer = getActiveCanvas()->getCurrentLayer();
 
         // attach feature, consensus and peak data
         FeatureMapSharedPtrType features = layer.getFeatureMap();
         ExperimentSharedPtrType peaks = layer.getPeakDataMuteable();
         ConsensusMapSharedPtrType consensus = layer.getConsensusMap();
-        vector<PeptideIdentification> peptides = layer.peptides;
+        vector<PeptideIdentification> peptides;
```
timosachsenberg, line 2547: maybe add some comment like
```suggestion
        // if the layer provides identification data -> retrieve it
        vector<PeptideIdentification> peptides;
```

src/openms_gui/source/VISUAL/LayerDataPeak.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <OpenMS/VISUAL/LayerDataPeak.h>
+
+using namespace std;
+
+namespace OpenMS
+{
+
+  namespace
+  {// internal linkage -- just to test instanciability
+    LayerDataPeak l_data;
```
timosachsenberg, line 44: will be removed before merge? (also in others)

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
```diff
@@ -887,40 +889,28 @@ namespace OpenMS
...
-    }
-    else if (layer.type == LayerData::DT_IDENT)
-    {
-      pep_begin = layer.peptides.begin();
-      pep_end = layer.peptides.end();
-    }
-    else
-      return;
+    bool show_labels = getLayerFlag(layer_index, LayerDataBase::I_LABELS);
+    auto p = dynamic_cast <IPeptideIds*>(&getLayer(layer_index));
```
timosachsenberg, line 893: add comment

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
```diff
@@ -3067,12 +3057,14 @@ namespace OpenMS
 
   void Plot2DCanvas::mergeIntoLayer(Size i, vector<PeptideIdentification> & peptides)
   {
-    LayerData& layer = layers_.getLayer(i);
-    OPENMS_PRECONDITION(layer.type == LayerData::DT_IDENT, "Plot2DCanvas::mergeIntoLayer(i, peptides) non-identification layer selected");
+    LayerDataBase& layer = layers_.getLayer(i);
+    OPENMS_PRECONDITION(layer.type == LayerDataBase::DT_IDENT, "Plot2DCanvas::mergeIntoLayer(i, peptides) non-identification layer selected");
```
timosachsenberg, line 3061: Just an observation: probably important to think about how
layer.type relates to the interfaces
There seems to be some redundancy as DT_IDENT implies IPeptideIds.

src/openms_gui/source/VISUAL/PlotCanvas.cpp
```diff
@@ -383,75 +388,69 @@ namespace OpenMS
...
-    LayerData new_layer;
-    new_layer.param = param_;
-    new_layer.filename = filename;
-    new_layer.setName(QFileInfo(filename.toQString()).completeBaseName());
-    new_layer.getFeatureMap() = map;
-    new_layer.type = LayerData::DT_FEATURE;
+    LayerDataBaseUPtr new_layer(new LayerDataFeature);
+    new_layer->getFeatureMap() = map;
 
+    augmentLayer(new_layer.get(), param_, filename);
```
timosachsenberg, line 433: maybe add coment as augment is pretty generic

src/openms_gui/source/VISUAL/PlotCanvas.cpp
```diff
@@ -383,75 +388,69 @@ namespace OpenMS
...
-    LayerData new_layer;
-    new_layer.param = param_;
-    new_layer.filename = filename;
-    new_layer.setName(QFileInfo(filename.toQString()).completeBaseName());
-    new_layer.peptides.swap(peptides);
-    new_layer.type = LayerData::DT_IDENT;
-
-    layers_.addLayer(std::move(new_layer));
+    LayerDataIdent* new_layer(new LayerDataIdent);
+    new_layer->getPeptideIds().swap(peptides);
```
timosachsenberg, line 451: std::move ?

timosachsenberg: or already an interface? setPeptideIds(...) ? could make transition to the new ID datastructure easier (if we store it internally in IDData to save memory)

timosachsenberg (approved):
a bit more comments would be great :)

src/openms_gui/include/OpenMS/VISUAL/LayerData.h
cbielow: I left the old header and cpp in the source tree (for now), just for reference.
To make sure noone is using the header until its removed, I added the macro.

cbielow: I added a TODO here, just to make sure its not forgotten:
https://github.com/OpenMS/OpenMS/projects/39#card-47792498

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
cbielow: this code will go away soon, but I can add the comment in the meantime :)

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
cbielow: Yes!
One of the last steps is to remove the `type` (at least from external view), so the user can solely use the much cleaner Interfaces which are about to be created.
But for now, I have to leave the type accessible, because without it everything breaks..

src/openms_gui/source/VISUAL/PlotCanvas.cpp
cbielow: good point. Will extend the interface.

src/openms_gui/include/OpenMS/VISUAL/LayerDataFeature.h
```diff
@@ -70,6 +70,14 @@ namespace OpenMS
       return getFeatureMap()->getUnassignedPeptideIdentifications();
     }
 
+    virtual void setPeptideIds(const PepIds& ids) override
+    {
+      getFeatureMap()->getUnassignedPeptideIdentifications() = ids;
```
timosachsenberg, line 75: Is this what we want here? Only the unassigned?

cbielow: this is only a refactoring.
And probably, the answer is `no`, but to also assign IDs to features one would need some IDMapping parameters. And I guess they are hard to guess, and thus would need to taken from the stored parameters in the featuremap, if they exist, or queried from the user... and I guess all of this was a bit too much when this was originally implemented...

PR #5685, opened 2021-11-25
add rvalue reference

timosachsenberg:
# Description

Based on results from clang-tidy scan  .

# Checklist:

# Note:


src/openms/source/METADATA/PeptideIdentification.cpp
```diff
@@ -134,6 +134,11 @@ namespace OpenMS
     hits_ = hits;
   }
 
+  void PeptideIdentification::setHits(std::vector<PeptideHit>&& hits)
+  {
+    hits_ = hits;
```
hroest, line 139:
```suggestion
    hits_ = std::move(hits);
```

otherwise this doesnt actually help?

hroest:
also this 

```
hits_.push_back(std::forward<PeptideHit>(hit));
```

should probably be 

```
hits_.push_back(std::move(hit));
```

It works, but is somewhat unconvential in an non-templated environment: https://stackoverflow.com/questions/28828159/usage-of-stdforward-vs-stdmove. Alternatively we could capture both cases using a template

```
template <typename T> void PeptideIdentification::setHits(T&& hits)
{
  hits_ = std::forward<T>(hits);
}
```

but doesnt seem very much "OpenMS-like"

src/openms/source/METADATA/ProteinHit.cpp
```diff
@@ -144,6 +144,14 @@ namespace OpenMS
     sequence_.trim();
```
cbielow, line 144: This is most likely a noop and should be removed.

src/openms/source/METADATA/ProteinHit.cpp
```diff
@@ -144,6 +144,14 @@ namespace OpenMS
...
   }
 
+  // sets the protein sequence
+  void ProteinHit::setSequence(String&& sequence)
+  {
+    sequence_ = std::move(sequence);
+    sequence_.trim();
```
cbielow, line 151: Not sure this is a good idea, since you do not know what will happen next.
Also: one function, one purpose. This is two purposes....

If one is, say, reading protein sequences from a file and creates ProtHits from them,
the more efficient interface is the old one, since you keep the original sequence lenght allocated in the reader and just need to reset its value without realloc.

timosachsenberg: What would you suggest? Only add an OPENMS_PRECONDITION that the sequence is trimmed?

I agree that moving could also make things less efficient depending on the context (like in the FASTA reader) but this can be handled in the FASTA reader (e.g., we could remove the move there).

src/openms/source/KERNEL/MSExperiment.cpp
```diff
@@ -672,12 +672,22 @@ namespace OpenMS
...
+
   /// adds a spectrum to the list
   void MSExperiment::addSpectrum(const MSSpectrum & spectrum)
   {
     spectra_.push_back(spectrum);
   }
 
+  void MSExperiment::addSpectrum(MSSpectrum && spectrum)
+  {
+    spectra_.emplace_back(std::move(spectrum));
```
cbielow, line 688: nitpicking (not important):
one convention (which I find reasonable), is that when moving complete objects, use `.push_back()`.
Only if constructing in place, use `emplace_back`.

src/openms/source/METADATA/ProteinHit.cpp
cbielow: I would suggest to simply remove the `trim()`.
The user can decide to trim beforehand if he decides it is reasonable.

timosachsenberg: My guess that this is there for a reason and removing it now will break things. If you volunteer to track that down to parsers etc. we can do ;)

timosachsenberg: I think what would work without the trim ist ph.getSequence().swap or ph.getSequence() = std::move(...)

cbielow: yes, that works, but it feels hacky.
I'd rather volunteer to check (and fix) the occurrences :)

removing the trim would per se not break anything, just use more memory.

cbielow: quickly looking though occurrences of ProteinHit::setSequence:
most of the time, the sequence is set on fresh objects and never touched again.
This means that trim() is a noop.
In a few cases, a proteinhit is used repeadedly (e.g. [in ](https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/FORMAT/HANDLERS/ConsensusXMLHandler.cpp#L454). Also there, the way the code is currently written, trimming does nothing. 
But, if the code were written better (e.g. reusing the allocated ProteinHit, instead of resetting it in every loop by doing `prot_hit_ = ProteinHit();`, then the trimming would actually be a pessimisation..

Please do not trim :)

timosachsenberg: I am just saying it might be there for a reason ;) e.g. parsing external files

src/openms/source/KERNEL/MSExperiment.cpp
timosachsenberg, line 688:
```suggestion
    spectra_.push_back(std::move(spectrum));
```

src/tests/class_tests/openms/source/IMDataConverter_test.cpp
```diff
@@ -72,6 +72,7 @@ START_SECTION((std::vector<PeakMap> splitByFAIMSCV(PeakMap& exp)))
   TEST_EQUAL(exp.getSpectra().size(), 19)
 
   vector<PeakMap> splitPeakMap = IMDataConverter::splitByFAIMSCV(std::move(exp));
+	TEST_EQUAL(exp.empty(), true) // moved out
```
timosachsenberg:
```suggestion
  TEST_EQUAL(exp.empty(), true) // moved out
```

src/openms/source/IONMOBILITY/IMDataConverter.cpp
```diff
@@ -185,7 +185,7 @@ namespace OpenMS
       // collapse for scans that actually have a float data array).
       if (in[k].containsIMData())
       {
-        MSExperiment frame = IMDataConverter::splitByIonMobility(in[k], number_of_bins);
+        MSExperiment frame = IMDataConverter::splitByIonMobility(in[k], number_of_bins); // TODO: @cbielow why no move version? e.g., for spec meta values
```
timosachsenberg, line 188: @cbielow any comments?

src/openms/source/IONMOBILITY/IMDataConverter.cpp
```diff
@@ -197,6 +197,7 @@ namespace OpenMS
         result.addSpectrum(std::move(in[k]));
       }
     }
+    in.clear(); // TODO: @cbielow. What about exp meta values?
```
timosachsenberg, line 200: @cbielow better to clear MSExperiment instead of having it in a mixed state? What about meta values? these are currently not transferred? is this correct

src/openms/source/KERNEL/ConsensusFeature.cpp
```diff
@@ -83,14 +87,31 @@ namespace OpenMS
...
+    {
+      String key = String("map") + handle.getMapIndex() + "/feature" + handle.getUniqueId();
+      throw Exception::InvalidValue(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, "The set already contained an element with this key.", key);
+    }
+  }
+
   void ConsensusFeature::insert(const HandleSetType& handle_set)
   {
-    for (ConsensusFeature::HandleSetType::const_iterator it = handle_set.begin(); it != handle_set.end(); ++it)
+    for (auto it = handle_set.cbegin(); it != handle_set.cend(); ++it)
```
hroest, line 101:
```suggestion
    for (auto& handle : handle_set)
```

src/openms/source/KERNEL/MSExperiment.cpp
```diff
@@ -696,12 +706,23 @@ namespace OpenMS
     chromatograms_ = chromatograms;
   }
 
+  /// sets the chromatogram list
+  void MSExperiment::setChromatograms(std::vector<MSChromatogram > && chromatograms)
```
hroest, line 710:
```suggestion
  void MSExperiment::setChromatograms(std::vector<MSChromatogram> && chromatograms)
```

src/openms/source/IONMOBILITY/IMDataConverter.cpp
cbielow: true.
one can probably do `std::move(in[k])`.

src/openms/source/IONMOBILITY/IMDataConverter.cpp
cbielow: ExperimentalSettings are currently not copied. This should probably be fixed.
Metadata of spectra are already copied in `splitByIonMobility`.

src/openms/source/IONMOBILITY/IMDataConverter.cpp
timosachsenberg, line 188:
```suggestion
        MSExperiment frame = IMDataConverter::splitByIonMobility(std::move(in[k]), number_of_bins);
```

src/openms/source/IONMOBILITY/IMDataConverter.cpp
timosachsenberg, line 200:
```suggestion
    result.ExperimentalSettings::operator=(in);
    in.clear(); // TODO: @cbielow. What about exp meta values?
```

timosachsenberg: do we need to update ranges?

cbielow: not at the moment since all data relevant to ranges is shuffled around but not modified.
However, once we include `IonMobility` in the range, we'd technically need to update, because there could be binning of IM values...

src/openms/include/OpenMS/KERNEL/ConsensusMap.h
```diff
@@ -174,7 +177,7 @@ namespace OpenMS
     OPENMS_DLLAPI ConsensusMap(const ConsensusMap& source);
 
     /// Destructor
-    OPENMS_DLLAPI ~ConsensusMap() override;
+    OPENMS_DLLAPI ~ConsensusMap() override = default;
```
timosachsenberg, line 180: hmm @cbielow do you know why this doesn't work after defaulting?

timosachsenberg: on mac and linux

jpfeuffer: Try putting it in the CPP: https://stackoverflow.com/questions/3065154/undefined-reference-to-vtable

timosachsenberg:
tests pass now.

src/openms/source/IONMOBILITY/IMDataConverter.cpp
```diff
@@ -197,6 +198,8 @@ namespace OpenMS
         result.addSpectrum(std::move(in[k]));
       }
     }
+    result.ExperimentalSettings::operator=(in);
```
cbielow, line 201: could also be `std::move(in)`?
Not crucial though

timosachsenberg, line 201:
```suggestion
    result.ExperimentalSettings::operator=(std::move(in));
```

PR #5689, opened 2021-11-30
Added new features in SpectraMerger

JeeH-K:
# Description

SpectraMerger modified so precursor mass filter can be used.

# Checklist:

# Note:


src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
```diff
@@ -297,22 +298,76 @@ namespace OpenMS
...
 
+    /**
+     * @brief check if the first and second mzs might be from the same mass
+     *
+     * @param mz1 the first m/z value
+     * @param mz2 the second m/z value
+     * @param tol_ppm tolerance in ppm
+     * @param max_c maximum possible charge value
+     */
+    bool areMassesMatched(double mz1, double mz2, double tol_ppm, int max_c)
```
cbielow, line 309: inputs should be `const`, just for good measure?!

KyowonJeong: Oh you are right.. Thanks. Should be const.

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
```diff
@@ -297,22 +298,76 @@ namespace OpenMS
...
+
+      const int min_c = 1;
+      const int max_iso_diff = 5; // maximum charge difference  5 is more than enough
+      const double max_charge_diff_ratio = 3.0; // maximum ratio between charges (large / small charge)
+
+      for (int c1 = min_c; c1 <= max_c; ++c1)
+      {
+        double mass1 = (mz1 - Constants::PROTON_MASS_U) * c1;
+
+        for (int c2 = min_c; c2 <= max_c; ++c2)
```
cbielow, line 324: do you really need a quadratic algorithm to do this?
You can compute the charge (`c`) ratio from `mz1/mz2 == c2/c1`. The solution must be in this range. There should be no need to test them all...

Are the charges annotated really that unreliable? And even if not, maybe its worth testing them and see if it matches?

How much does the current solution slow down the algorithm?

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
cbielow: `max_c` should be uint, or you should internally test for it being >=1, because otherwise the algorithm will always return false.

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
KyowonJeong: Currently this does not affect any running time on my macbook. But yes they can be improved as you mentioned. I will test with that as well. Very good advice.

src/openms/source/FILTERING/TRANSFORMERS/SpectraMerger.cpp
```diff
@@ -88,6 +96,8 @@ namespace OpenMS
     // same precursor MS/MS merging
     defaults_.setValue("precursor_method:mz_tolerance", 10e-5, "Max m/z distance of the precursor entries of two spectra to be merged in [Da].");
     defaults_.setMinFloat("precursor_method:mz_tolerance", 0);
+    defaults_.setValue("precursor_method:mass_tolerance", .0, "Max mass distance of the precursor entries of two spectra to be merged in [Da]. Active when set to a positive value.");
```
cbielow, line 99: this flag is never used.

Part of another PR?

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
```diff
@@ -566,31 +657,33 @@ namespace OpenMS
           Size spec_a = consensus_spec.size(), spec_b = exp[*sit].size(), align_size = alignment.size();
           for (auto pit = exp[*sit].begin(); pit != exp[*sit].end(); ++pit)
           {
-            if (alignment.empty() || alignment[align_index].second != spec_b_index)
+            if (alignment.size() == 0 || alignment[align_index].second != spec_b_index)
```
cbielow, line 660: I think @timosachsenberg will not like this :)

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
```diff
@@ -395,13 +466,25 @@ namespace OpenMS
...
+              if (add)
               {
-                weight = std::exp(factor * pow(it_rt_2->getRT() - it_rt->getRT(), 2));
+                double weight = 1;
+                if (average_type == "gaussian")
+                {
+                  weight = std::exp(factor * pow(it_rt_2->getRT() - it_rt->getRT(), 2));
+                }
+                std::pair<Size, double> p(m, weight);
+                spectra_to_average_over[n].push_back(p);
```
cbielow, line 486: if you feel like it, you can shorten this to
`spectra_to_average_over[n].emplace_back(m, weight);`

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
timosachsenberg: Haha. Clang-tidy complains. But yeah instead of size() > 0 it is considered slightly better to write !empty()

src/openms/source/FILTERING/TRANSFORMERS/SpectraMerger.cpp
KyowonJeong: Yes it is. This is the new parameter.  If this is used, mz_tolerance is deacitivated. Probably I need put that in the explanation.

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
KyowonJeong: Wow Is it? Thanks all. I fix them.

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
KyowonJeong: This type of sophisticated use is something I need to learn. Many thanks. Will try.

src/openms/source/FILTERING/TRANSFORMERS/SpectraMerger.cpp
cbielow: mhh.. but the flag is new, right?
Yet I dont see it being used anywhere else in this PR...
I am probably missing something but can you point me to the location where its value is queried?

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
```diff
@@ -297,22 +298,76 @@ namespace OpenMS
...
+
+          double mass2 = (mz2 - Constants::PROTON_MASS_U) * c2;
+
+          if (fabs(mass1 - mass2) > max_iso_diff)
+          {
+            continue;
+          }
+          for (int i = -max_iso_diff; i <= max_iso_diff; ++i)
+          {
+            if (fabs(mass1 - mass2 + i * Constants::ISOTOPE_MASSDIFF_55K_U) < mass1 * tol_ppm * 1e-6)
```
hroest, line 343: does this assume that mass2 has to be larger than mass1? can you document this?

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
```diff
@@ -297,22 +298,76 @@ namespace OpenMS
...
+      }
+      return false;
+    }
+
     /**
      * @brief average over neighbouring spectra
      *
      * @param exp experimental data to be averaged
      * @param average_type averaging type to be used ("gaussian" or "tophat")
+     * @param ms_level targe MS level. If it is -1, ms_level will be determined by ms_level parameter.
```
hroest, line 358:
```suggestion
     * @param ms_level target MS level. If it is -1, ms_level will be determined by ms_level parameter.
```

I dont understand what this means: "If it is -1, ms_level will be determined by ms_level parameter" -> it sounds like "X will be determined by X" ?

timosachsenberg:
any update here?

timosachsenberg:
@KyowonJeong any chance to revive this PR?

KyowonJeong:
> @KyowonJeong any chance to revive this PR?

This can be removed as I deleted this feature.

PR #5690, opened 2021-12-01
FLASHDeconvWizard

JeeH-K:
# Description

FLASHDeconvWizard split from FLASHDeconv PR #5015
To be merged after PR #5688 

# Checklist:

# Note:


src/openms_gui/source/VISUAL/DIALOGS/FLASHDeconvTabWidget.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+    void FLASHDeconvTabWidget::broadcastNewCWD_(const QString& new_cwd)
+    {
+      // RAII to avoid infinite loop (setCWD signals updatedCWD which is connected to slot broadcastNewCWD_)
+      QSignalBlocker blocker1(ui->input_mzMLs);
+      ui->input_mzMLs->setCWD(new_cwd);
+    }
+
+    /// custom arguments to allow for looping calls
+    struct Args
```
cbielow, line 375: a lot of code duplication from the original wizard.
Please move it into a separate class and just call it from both wizards

src/openms_gui/source/VISUAL/DIALOGS/FLASHDeconvTabWidget.cpp
```diff
@@ -0,0 +1,436 @@
...
+#include <algorithm>
+
+using namespace std;
+
+namespace OpenMS
+{
+  namespace Internal
+  {
+
+    FLASHDeconvGUILock::FLASHDeconvGUILock(FLASHDeconvTabWidget* ftw)
```
cbielow, line 62: this GUI lock class can most certainly be used standalone.
No need to copy it from the original wizard.
Just make it work with both.

src/openms_gui/source/VISUAL/DIALOGS/FLASHDeconvTabWidget.cpp
```diff
@@ -0,0 +1,436 @@
...
+      :
+      ftw_(ftw),
+      old_(ftw->currentWidget()),
```
timosachsenberg, line 65: is this order thread-safe?
e.g., old_(ftw->currentWidget()) get's executed before the lock is acquired?
Did not look into this in detail.

src/openms_gui/source/VISUAL/DIALOGS/FLASHDeconvTabWidget.cpp
timosachsenberg: were would you put it?
Some WizardHelper class?

cbielow: yep, sounds good

src/openms_gui/source/VISUAL/DIALOGS/FLASHDeconvTabWidget.cpp
cbielow: this should not matter because: 
a) this lock should only ever be acquired by the GUI thread, not a worker thread 
b) the lock happens in the Ctor, so only one thread can get there. And the lock is specific to this object.

timosachsenberg:
Apart from the potential code reuse as Chris suggested... look good.

timosachsenberg:
Maybe one suggestion for future updates:
it seems as if just the tool is called. In principle, you could also directly call your algorithm, which is a bit faster and would give you more control. But fine as it is now.

PR #5694, opened 2021-12-02
[fix] added 3' phosphate gain for unspecific cleavage of nucleotides.…

poshul:
… This is consistant with AbouHaidar and Ivanov 1999.

# Description

Previously unspecific cleavage of RNA resulted in the loss of the phosphate group at the cleavage site. Per https://www.degruyter.com/document/doi/10.1515/znc-1999-7-813/pdf it appears that the more common behavior is for the phosphate to stay on the 3' end of the "left" fragment.

# Checklist:

# Note:


poshul:
Jenkins seems to be on the fritz. @jpfeuffer any idea why we are getting the following:
`   Error when uploading file: /home/jenkins/ws/openms/PR/bldtst/a9154670/build/Testing/20211202-1027/Test.xml
   Error message was: Operation too slow. Less than 1 bytes/sec transferred the last 120 seconds
   Problems when submitting via HTTP`

timosachsenberg (approved):

Issue #5678, opened 2021-11-24
Question: ElutionModelFitter imputation


timosachsenberg:
On orbitrap data it happens that e.g. a single peak (one m/z peak in one spectrum) get's picked up but nothing else.
Fitting a model fails of course but imputation then put's in a reasonable looking value. But even without imputation the single value is kept.
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/ElutionModelFitter.cpp#L432

Would it make sense to discard the feature in the case of a single e.g. noise peak? I am asking because we see some elevated background intensity in orbitrap data. @oliveralka what do you think?

hroest:
I agree that a single point in RT should not be used and should probably be filtered out

timosachsenberg:
where would be the best place? in the ElutionModelFitter it is already too late I guess?

hroest:
It should be earlier, the feature should never be created in the first place

hroest:
see #5677 and #5668

that would probably be in the MRMTransitionGroupFinder::pickTransitionGroup where we should never pick such a group in the first place. I am a bit surprised since this uses the PeakPickerHiRes internally and it should not pick a single peak flanked by zeros or would it?

timosachsenberg:
> that would probably be in the MRMTransitionGroupFinder::pickTransitionGroup where we should never pick such a group in the first place. I am a bit surprised since this uses the PeakPickerHiRes internally and it should not pick a single peak flanked by zeros or would it?

PeakPickerHiRes also picks single peaks, as this can in rare cases happen (I think we tested it once, and we got slightly worse ID results).

Ok so setting min_peak_width_  to a very small value (1e-3) would probably solve the "single peak" issue.

What happens e.g., if we get for example: 0090000000900  (9= some high intensity, 0=zero in chromatogram) here? Would this also be removed (e.g. how is peak width here calculated)? 
What about 00000099000000 ? This would probably get a proper width and be picked?

timosachsenberg:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp#L55
checkend in
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h#L277

Ok, any objections to setting it to 0.001 as default?

[...]

PR #5702, opened 2021-12-13
fix imputation bug on single peak mass traces

timosachsenberg:
# Description

Might fix: https://github.com/OpenMS/OpenMS/issues/5678

# Checklist:

# Note:

### DIFF
--- a/src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
+++ b/src/openms/source/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.cpp
@@ -52,7 +52,7 @@ namespace OpenMS
   {
     defaults_.setValue("stop_after_feature", -1, "Stop finding after feature (ordered by intensity; -1 means do not stop).");
     defaults_.setValue("stop_after_intensity_ratio", 0.0001, "Stop after reaching intensity ratio");
-    defaults_.setValue("min_peak_width", -1.0, "Minimal peak width (s), discard all peaks below this value (-1 means no action).", {"advanced"});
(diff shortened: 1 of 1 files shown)


jpfeuffer:
To be honest, after looking into the code, I am not sure if this will solve the problem:
https://github.com/OpenMS/OpenMS/blob/2c5d58413f9720e87cf44de657ade5315ea2b6ed/src/openms/source/ANALYSIS/OPENSWATH/PeakPickerMRM.cpp#L54

It does not stop the picking or invalidates the transition group. It just makes sure that it at least picks up a peaks in a certain range around the "center" peak.

https://github.com/OpenMS/OpenMS/blob/2c5d58413f9720e87cf44de657ade5315ea2b6ed/src/openms/source/ANALYSIS/OPENSWATH/PeakPickerMRM.cpp#L200

Did you check this?

timosachsenberg:
I made the change based on @hroest comment:
```Also currently MRMTransitionGroupFinder::createMRMFeature uses a min_peak_width_ so in principle this can be prevent there by setting this to non-zero ```

jpfeuffer:
Ah sorry, didnt check the header. There is another usage of the member:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h#L277

Ok should be fine

jpfeuffer (approved):

timosachsenberg:
ok thanks for checking.
Maybe one could even be stricter using the information here:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMTransitionGroupPicker.h#L237-L238
to check if we have e.g. at least 3 non-zero peaks in the chromatogram - or similar

PR #5703, opened 2021-12-14
[REFACTOR] New RangeManager

cbielow:
# Description

Replace the old RangeManager with a modern implementation, which supports more operations on Ranges (RT, m/z, Intensity, Mobility etc), and is safer to use, e.g. instead of `map.getMin()[0]` which could mean anything, you now use `map.getMinRT()`.

The new RangeManager supports arbitrary, but explicitly specified dimensions by using variadic templates, e.g.
`RangeManager<RangeRT, RangeMZ>` can manage RT and m/z dimensions.
Ranges with different dimensions can interact, e.g. a range of a PeakMap (rt,mz,intensity) can be updated by the range of a spectrum (mz,intensity) by simply doing `map.extendRange(spec)`.
This usually leads to a reduction in code size, and much cleaner code, e.g. see
https://github.com/OpenMS/OpenMS/pull/5703/files#diff-c819ce9062138032f71bd4408e255d4cd375069116192fcc9a01a6fbcf0f6656L314

RangeManager is implemented for all core datastructures, such as Spectrum, PeakMap, FeatureMap, ConsensusMap etc
and is used by the GUI Layers (further refactoring is needed there for the plotting canvas, which still uses old-fashioned DIntervalBase<D>, but that change is rather large, and thus not part of this PR).

# Note:


src/openms_gui/include/OpenMS/VISUAL/LayerDataIdent.h
```diff
@@ -65,14 +65,16 @@ namespace OpenMS
...
-    float getMaxIntensity() const override
-    {
-      return -1;
+      RangeAllType r;
+      for (const PeptideIdentification& pep : peptides_)
+      {
+        r.extendRT(pep.getRT());
+        r.extendMZ(pep.getMZ());
+      }
+      r.extendIntensity(0);
```
timosachsenberg, line 76: is this needed?

src/openms_gui/source/VISUAL/Plot1DCanvas.cpp
```diff
@@ -2089,7 +2089,7 @@ namespace OpenMS
   {
     if (intensity_mode_ == IM_PERCENTAGE)
     {
-      percentage_factor_ = overall_data_range_.maxPosition()[1] / getLayer(layer_index).getCurrentSpectrum().getMaxInt();
+      percentage_factor_ = overall_data_range_.maxPosition()[1] / getLayer(layer_index).getCurrentSpectrum().getMaxIntensity();
```
timosachsenberg, line 2092: is here maxPosition()[1] still needed?

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
```diff
@@ -305,21 +305,9 @@ namespace OpenMS
...
-        percentage_factor_ = overall_data_range_.maxPosition()[2] / layer.getFeatureMap()->getMaxInt();
-      }
-      else if (layer.type == LayerDataBase::DT_CONSENSUS && layer.getConsensusMap()->getMaxInt() > 0.0)
-      {
-        percentage_factor_ = overall_data_range_.maxPosition()[2] / layer.getConsensusMap()->getMaxInt();
-      }
-      else if (layer.type == LayerDataBase::DT_CHROMATOGRAM && layer.getConsensusMap()->getMaxInt() > 0.0)
-      {
-        //TODO CHROM not sure if needed here
+        percentage_factor_ = overall_data_range_.maxPosition()[2] / layer.getMaxIntensity();
```
timosachsenberg, line 310: same here?

src/openms_gui/include/OpenMS/VISUAL/LayerDataIdent.h
cbielow: matter of taste I guess. Can be removed

src/openms_gui/source/VISUAL/Plot1DCanvas.cpp
cbielow: yes, because `overall_data_range_` is still the old-style range. Replacing that is part of an upcoming PR

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
```diff
@@ -2984,7 +2972,7 @@ namespace OpenMS
       }
 
       // update gradient if the min/max intensity changes
-      if (tmp.getIntensity() < current_layer.getFeatureMap()->getMinInt() || tmp.getIntensity() > current_layer.getFeatureMap()->getMaxInt())
+      if (!current_layer.getFeatureMap()->getRange().RangeIntensity::contains(tmp.getIntensity()))
```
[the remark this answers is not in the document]
cbielow: yes. Its a parent-class scope.
One could also add new methods to the RangeIntensity Class, i.e. `containsIntensity()`. That would avoid the need for the scoping, but its also more code to write in the RangeManager.

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -28,181 +28,574 @@
...
-    typedef DRange<1> IntensityRangeType;
+  public:
+    /// Ctor: initialize with empty range
+    RangeBase() = default;
+    /// set min and max; if min > max, then an exception will be thrown
+    RangeBase(const double min, const double max) :
+        min_(min), max_(max)
+    {
+      if (min_ > max_)
+        throw "Invalid initialization of range";
```
timosachsenberg, line 67: only std::runtime_error are propagated to python exceptions

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -28,181 +28,574 @@
...
 
-    /// Copy constructor
-    RangeManager(const RangeManager & rhs) :
-      int_range_(rhs.int_range_),
-      pos_range_(rhs.pos_range_)
-    {}
+    /// is the range empty (i.e. default constructed or cleared using clear())?
+    bool isEmpty() const
+    { // invariant: only possible when default constructed or clear()'ed
+      return min_ > max_;
```
timosachsenberg, line 79: if setMin(x) is called on the default constructed range then isEmpty() = false.
But wouldn't this be an empty range (because min = max)?

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -28,181 +28,574 @@
...
+  class RangeManager : public RangeBases...
+  {
+  public:
+    // rule of 0 -- no need for a virtual d'tor or anything fancy
+    // ...
+
+    bool operator==(const RangeManager& rhs) const
+    {
+      bool equal = true;
+      for_each_base([&](auto* base) {
```
timosachsenberg, line 408: can you add some comment here. for the for each base and the decay_t / decltype as we never used that before in the code base

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -28,181 +28,574 @@
...
+        out << *base;
+      });
+    }
 
-      int_range_.setMinX(it_min);
-      int_range_.setMaxX(it_max);
+  protected:
+    /// use fold expression to iterate over all RangeBases of RangeManager and apply a lambda for each one
+    template<typename Visitor>
+    void for_each_base(Visitor&& visitor)
```
timosachsenberg, line 550: maybe markt that with underscore so it is clear that this is a protected member

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -28,181 +28,574 @@
...
+#include <OpenMS/config.h>
+#include <OpenMS/CONCEPT/Exception.h>
+
+#include <algorithm> // for min/max
+#include <cassert>
+#include <iosfwd>  // for std::ostream
+
 
 namespace OpenMS
 {
```
timosachsenberg, line 46: can you add a few lines of comment (@brief ) so people understand the design decision.
blabla static polymorphism using visitor patterns blablaba maybe link to some stackoverflow or external resource

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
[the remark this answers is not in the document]
timosachsenberg: I would say it would be a bit more in line with the getMinIntensity() etc. methods.
How much code would you need to add?

src/openms/include/OpenMS/KERNEL/MSExperiment.h
```diff
@@ -90,13 +89,15 @@ namespace OpenMS
     /// Chromatogram peak type
     typedef ChromatogramPeakT ChromatogramPeakType;
     /// Area type
-    typedef DRange<2> AreaType;
+    //typedef DRange<2> AreaType;
```
hroest, line 92:
```suggestion
```

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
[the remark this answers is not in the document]
cbielow: one method per dimension. Four lines each.
I was acutally thinking about using an old-style preprocessor macro, just to avoid writing the same code 4 times (rt, mz, int, mobility). Not sure if thats the right way to go, but it would safe a lot of typing and testing.

src/openms/include/OpenMS/KERNEL/RangeManager.h
cbielow: sure, can do

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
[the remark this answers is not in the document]
timosachsenberg: What do others think? I think will not write those rangemanagers that often and a nice interface would outweight the extra work here.

cbielow:
all tests pass.
Ready for merge.

src/openms_gui/source/VISUAL/Plot2DCanvas.cpp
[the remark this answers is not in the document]
timosachsenberg: @jpfeuffer  what is your take on that?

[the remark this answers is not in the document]
jpfeuffer: Might be better for compilation to have less code, which I am valuing very highly these days.
On the other hand this indeed pretty weird and I wouldn't be able to write that kind of code without looking up namespace rules or finding other usages for the first few times.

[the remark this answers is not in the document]
timosachsenberg: Ah right :). @cbielow can you move some code from the header in the cpp file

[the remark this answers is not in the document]
cbielow: I can move some code, but what is moveable are mostly one-line functions.
The majority is templates and thus unmovable anyway.
I suspect compile time will not be affected in any way, but I can do it of course.

I also suspect that adding explicit functions will not influence compile time :)
So shall I add them?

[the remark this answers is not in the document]
jpfeuffer: Ok. Pretty please? 🥺

timosachsenberg:
Chris will add namespace change to other PR

PR #5708, opened 2021-12-15
pass calling process environment to children

timosachsenberg:
# Description

Passing the parent process environment reduces the danger of running a process with e.g. different environment variables set.

# Checklist:

# Note:

### DIFF
--- a/src/openms/source/SYSTEM/ExternalProcess.cpp
+++ b/src/openms/source/SYSTEM/ExternalProcess.cpp
@@ -81,6 +81,10 @@ namespace OpenMS
 
   ExternalProcess::RETURNSTATE ExternalProcess::run(const QString& exe, const QStringList& args, const QString& working_dir, const bool verbose, String& error_msg, IO_MODE io_mode)
   {
+    // pass environment variables to child process
+    QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
+    qp_->setProcessEnvironment(env);
+
     error_msg.clear();
     if (!working_dir.isEmpty())
     {


timosachsenberg:
@oliveralka reformat failed with https://github.com/OpenMS/OpenMS/runs/4537967682?check_suite_focus=true
do you know how to fix that?

src/openms/source/SYSTEM/ExternalProcess.cpp
```diff
@@ -81,6 +81,10 @@ namespace OpenMS
 
   ExternalProcess::RETURNSTATE ExternalProcess::run(const QString& exe, const QStringList& args, const QString& working_dir, const bool verbose, String& error_msg, IO_MODE io_mode)
   {
+    // pass environment variables to child process
+    QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
+    qp_->setProcessEnvironment(env);
```
cbielow, line 86: why is this required?
According to https://doc.qt.io/qt-5/qprocess.html#processEnvironment,
the calling process' environment is the default.

cbielow (commented):
not sure this is needed.
Did it not work as expected somewhere?

src/openms/source/SYSTEM/ExternalProcess.cpp
timosachsenberg: hmm in my case:

I ran a tool via a bash script.
- In the bash script I export the OPENMS_DATA_PATH
- the tool executed in the script starts an ExternalProcess and executes a python script
Somehow the OPENMS_DATA_PATH was not set in the python script

After this change it worked (I may double check but I did not get the error message again)

jpfeuffer: Hmm after reading the docs I agree with @cbielow that this should make no difference.

timosachsenberg: k thanks for checking. I will need to dig into this a bit more

oliveralka:
> @oliveralka reformat failed with https://github.com/OpenMS/OpenMS/runs/4537967682?check_suite_focus=true
> do you know how to fix that?

Please see #5712

We will have to test if that fixes everything.

PR #5722, opened 2021-12-22
[Feature] export MSExperiment to data frames for MassQL

axelwalter:
# Description
Added a function for pyOpenMS MSExperiment to export peak data to MS1 and MS2 data frames in a format required by MassQL.

# Checklist:

# Note:


src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+        The Python module massql allows queries in mass spectrometry data (MS1 and MS2
+        data frames) in a SQL like fashion (https://github.com/mwang87/MassQueryLanguage).
+        
+        Returns:
+        ms1_df (pandas.DataFrame): peak data of MS1 spectra
+        ms2_df (pandas.DataFrame): peak data of MS2 spectra with precursor information
+        """
+        self.updateRanges()
+
+        def get_polarity(spectrum):
```
timosachsenberg, line 319: add comment that this is an inner function for internal use?

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+            elif polarity == IonSource.Polarity.NEGATIVE:
+                return 2
+
+        def get_peak_arrays_from_spec(spec, scan_num):
+            if spec.getMSLevel() == 2:
+                ms2_data = (spec.getPrecursors()[0].getMZ(), self.getPrecursorSpectrum(scan_num)+1, spec.getPrecursors()[0].getCharge())
+            else:
+                ms2_data = ()
+            for i, peak in enumerate(spec):
+                peak_data = (peak.getIntensity(), peak.getMZ(), scan_num+1, spec.getRT()/60, get_polarity(spec))
```
timosachsenberg, line 334:
```suggestion
                peak_data = (peak.getIntensity(), peak.getMZ(), scan_num+1, spec.getRT()/60.0, get_polarity(spec))
```

timosachsenberg: I think we have a numpy function getPeakData... could this be used to make this faster?

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+        self.updateRanges()
+
+        def get_polarity(spectrum):
+            polarity = spectrum.getInstrumentSettings().getPolarity()
+            if polarity == IonSource.Polarity.POLNULL:
+                return 0
+            elif polarity == IonSource.Polarity.POSITIVE:
+                return 1
+            elif polarity == IonSource.Polarity.NEGATIVE:
+                return 2
```
timosachsenberg, line 326: could we use a python enum / would this make sense?

timosachsenberg: otherwise I might prefer -1, 0, +1

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+        def get_peak_arrays_from_spec(spec, scan_num):
+            if spec.getMSLevel() == 2:
+                ms2_data = (spec.getPrecursors()[0].getMZ(), self.getPrecursorSpectrum(scan_num)+1, spec.getPrecursors()[0].getCharge())
+            else:
+                ms2_data = ()
+            for i, peak in enumerate(spec):
+                peak_data = (peak.getIntensity(), peak.getMZ(), scan_num+1, spec.getRT()/60, get_polarity(spec))
+                yield peak_data + ms2_data
+
+        def get_spec_arr(spec, scan_num):
```
timosachsenberg, line 337: can you add some documentation to the methods (even if they are inner methods)

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+    def get_massql_df(self):
+        """Exports data from MSExperiment to pandas DataFrames to be used with MassQL.
+
+        The Python module massql allows queries in mass spectrometry data (MS1 and MS2
+        data frames) in a SQL like fashion (https://github.com/mwang87/MassQueryLanguage).
+        
+        Returns:
+        ms1_df (pandas.DataFrame): peak data of MS1 spectra
+        ms2_df (pandas.DataFrame): peak data of MS2 spectra with precursor information
+        """
```
timosachsenberg, line 316: maybe also document which columns are returned?

timosachsenberg:
I also added @enetz he has more experience with pandas and numpy so he might spot some low hanging performance optimizations.

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+            else:
+                ms2_data = ()
+            for i, peak in enumerate(spec):
+                peak_data = (peak.getIntensity(), peak.getMZ(), scan_num+1, spec.getRT()/60, get_polarity(spec))
+                yield peak_data + ms2_data
+
+        def get_spec_arr(spec, scan_num):
+            arr = np.asarray([peak_data for peak_data in get_peak_arrays_from_spec(spec, scan_num)], dtype='f')
+            arr = np.insert(arr, 1, arr[:,0]/np.amax(arr[:,0]), axis=1) # i_norm
+            arr = np.insert(arr, 2, arr[:,0]/np.sum(arr[:,0]), axis=1) # tic_i_norm
```
enetz, line 340: this can be made better readable by making a pd.DataFrame earlier on.
you can then use e.g.
`arr['i_norm'] = arr[: , 0]/...`
to add new columns and use pd.concat to stack them later on

src/pyOpenMS/pyopenms/dataframes.py
jpfeuffer: Yes, extract the full numpy arrays for mz and inty, generate equally sized numpy arrays for RT/60, scannum+1 and polarity (e.g. `a=np.empty(len(mzvals)); a.fill(RT/60)`), concatenate rowwise and then transpose.
Should be fastest, since all is in numpy

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+            if spec.getMSLevel() == 2:
+                ms2_data = (spec.getPrecursors()[0].getMZ(), self.getPrecursorSpectrum(scan_num)+1, spec.getPrecursors()[0].getCharge())
+            else:
+                ms2_data = ()
+            for i, peak in enumerate(spec):
+                peak_data = (peak.getIntensity(), peak.getMZ(), scan_num+1, spec.getRT()/60, get_polarity(spec))
+                yield peak_data + ms2_data
+
+        def get_spec_arr(spec, scan_num):
+            arr = np.asarray([peak_data for peak_data in get_peak_arrays_from_spec(spec, scan_num)], dtype='f')
```
jpfeuffer, line 338: fromiter didnt work or was slower? This will create a large python list first.

src/pyOpenMS/pyopenms/dataframes.py
jpfeuffer: Maybe add two more np.empty rows for i_norm and tic_i_norm. So you just have to fill it later and don't need to use insert.

src/pyOpenMS/pyopenms/dataframes.py
```diff
@@ -304,6 +304,61 @@ def get_df(self, long : bool = False):
...
+            arr = np.insert(arr, 1, arr[:,0]/np.amax(arr[:,0]), axis=1) # i_norm
```
jpfeuffer, line 339: could be slow. see my comments above. If you "preallocate" an empty row (or column after transpose, you can just calculate the max and the sum and use np.map to set the values.

jpfeuffer:
Cool idea, see my comments for optimization. In the best case, keep track of times after every change you make ;)
Mine are mostly theoretical ideas.

axelwalter:
Thanks for the review! Very interesting. I tried some suggestions for optimization, like using pandas earlier with this code:

```
    def get_peak_arrays_from_spec(spec, scan_num):
        if spec.getMSLevel() == 2:
            ms2_data = (spec.getPrecursors()[0].getMZ(), self.getPrecursorSpectrum(scan_num)+1, spec.getPrecursors()[0].getCharge())
        else:
            ms2_data = ()
        for peak in spec:
            peak_data = (peak.getIntensity(), peak.getMZ(), scan_num+1, spec.getRT()/60, get_polarity(spec))
            yield peak_data + ms2_data

    def get_spec_df(mslevel, dtypes):
        for scan_num, spec in enumerate(self):
            if spec.getMSLevel() == mslevel:
                df = pd.DataFrame(np.fromiter(get_peak_arrays_from_spec(spec, scan_num), dtype=dtypes))
                # df.insert() not slower then df['i_norm'] = ... (and have to reorder columns later)
                df.insert(1, 'i_norm', df['i']/df['i'].max()) # i_norm
                df.insert(2, 'i_tic_norm', df['i']/df['i'].sum()) # tic_i_norm
                yield df

    dtypes = [('i', 'float32'), ('mz', 'float64'), ('scan', 'int32'), ('rt', 'float32'), ('polarity', 'int32')]
    if 1 in self.getMSLevels():
        ms1_df = pd.concat(list(get_spec_df(1, dtypes)))
    else:
        ms1_df = pd.DataFrame(columns=[x[0] for x in dtypes])

    dtypes += [('precmz', 'float64'), ('ms1scan', 'int32'), ('charge', 'int32')]
    if 2 in self.getMSLevels():
        ms2_df = pd.concat(list(get_spec_df(2, dtypes)))
    else:
        ms2_df = pd.DataFrame(columns=[x[0] for x in dtypes])

    return ms1_df, ms2_df
```

However, this was significantly slower (about 10 seconds vs 2 seconds) so using numpy as much as possible seems to be working fine. Compared to file loading and exporting of data frames by the MassQL library the solution is also significantly faster (about 50%).

Added some more docs and changed the code to be more understandable, but was not able to get better run time compared to before.

But will still test some more of your suggestions now.

src/pyOpenMS/pyopenms/dataframes.py
axelwalter: Tried it first with fromiter, but that creates a 1D array and I need to access the columns to get intensity values and add the normalized values.
Is there an elegant way to do that with the 1D array generated with fromiter?

src/pyOpenMS/pyopenms/dataframes.py
axelwalter: Now I added some placeholder values (-1, since all values in the matrix are numeric numpy.empty does not work, np.nan would be an alternative) and change the values later to the normalized values, which works fine.
Btw, using numpy.insert does not make it slower, seems to be well optimized.

src/pyOpenMS/pyopenms/dataframes.py
axelwalter: Using pandas at this point was significantly slower then numpy, maybe you can check the code I posted below if that is what you had in mind?

src/pyOpenMS/pyopenms/dataframes.py
axelwalter: That was my first solution as well, since it is the most intuitive. However, the massql exports positive as 1 and negative as 2, that is the reason for this choice.

jpfeuffer:
By the way, do I see it correctly that we are basically are exporting every single peak in the experiment, only split into a Ms1 and an Ms2 table?

I think then the ultimate solution would be to add an Ms_level parameter to our new getPeakData function in C++ (the one that @timosachsenberg did). I would love that addition anyway, in case someone loads a full experiment but only wants Ms1 peaks for plotting. I can take a short look.

src/pyOpenMS/pyopenms/dataframes.py
axelwalter: Intersting, this would create the ndarray for the spectrum without numpy.asarray from a list of python lists. Will try...

axelwalter:
Yes that's basically it. Since we calculate the normalized intensities based on the spectrum, does it make sense to collect the data by iterating over the spectra? Because otherwise we have to split the larger array later for the calculations.

jpfeuffer:
Yeah exactly. i was thinking to just apply that function twice. Once for Ms1 and once for Ms2. I think in this case iteration is faster than memory management between c and python.

If we assume a fixed number of levels, we could even return a tuple of e.g. nine arrays:
Ms1 int,mz,rt, Ms2 int,mz,rt Ms3 int,mz,rt
This should be super fast 😂 but not sure if worth it.

jpfeuffer:
https://github.com/OpenMS/OpenMS/pull/5726

Then you can use getMS2PeakData for MS2 peaks and get2DPeakData (with the full range) for MS1 peaks.
The rest should be straightforward. No need for loops or anything.

axelwalter:
In this case you would still need that right? Because we need collect the MS2 data (precursor mz, ms1scan, ...) and calculate normalizations for each spectrum.

axelwalter:
With the new changes the computation takes only 0.24 s compared to 2 s with my test file.

timosachsenberg:
I think we might want to add a parameter for the level here: https://github.com/OpenMS/OpenMS/pull/5726

timosachsenberg (approved):

timosachsenberg:
would be good to have a test output to track regressions

src/pyOpenMS/tests/unittests/test000.py
```diff
@@ -2950,9 +2950,15 @@ def testMSExperiment():
 
     assert exp.get_df().shape == (3,3)
 
+    pyopenms.MzMLFile().load(os.path.join(os.environ['OPENMS_DATA_PATH'], 'examples/BSA/BSA1.mzML'), exp)
+
+    pyopenms.MzMLFile().load('/home/axel/dev/OpenMS/share/OpenMS/examples/BSA/BSA1.mzML', exp)
```
timosachsenberg, line 2955: relative path

timosachsenberg:
sorry but the test file is too big (10 mb tsv). Needs to be smaller than 1mb. e.g., one of the BSA fractions might be still result in too big tsv file (https://github.com/OpenMS/OpenMS/blob/develop/share/OpenMS/examples/FRACTIONS/BSA1_F1.mzML) . 
e.g. https://github.com/OpenMS/OpenMS/blob/develop/src/tests/topp/FileFilter_1_input.mzML could be of better size

timosachsenberg (approved):

PR #5726, opened 2021-12-23
[FEATURE] add function to quickly export MS2 peaks in rt,mz,int format

jpfeuffer:
For usage in pyopenms

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/MSExperiment.h
+++ b/src/openms/include/OpenMS/KERNEL/MSExperiment.h
@@ -377,6 +377,28 @@ public:
       }
     }
 
+    // for fast pyOpenMS access to MS2 peak data in format: [rt, mz, intensity]
+    void getMS2PeakData(
+      std::vector<float>& rt, 
+      std::vector<float>& mz, 
+      std::vector<float>& intensity) const
+    {
+      for (const auto& s : spectra_)
+      {
+        if (s.getMSLevel() == 2)
+        {
+          rt.reserve(rt.size() + s.size());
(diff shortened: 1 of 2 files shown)


src/openms/include/OpenMS/KERNEL/MSExperiment.h
```diff
@@ -377,6 +377,28 @@ namespace OpenMS
       }
     }
 
+    // for fast pyOpenMS access to MS2 peak data in format: [rt, mz, intensity]
+    void getMS2PeakData(
```
timosachsenberg, line 381: a level parameter could make sense here as we export all peaks (as opposed to the MS1 only area iterator version above)

timosachsenberg:
related to #5722

src/openms/include/OpenMS/KERNEL/MSExperiment.h
```diff
@@ -377,6 +377,28 @@ namespace OpenMS
...
+      std::vector<float>& rt, 
+      std::vector<float>& mz, 
+      std::vector<float>& intensity) const
+    {
+      for (const auto& s : spectra_)
+      {
+        if (s.getMSLevel() == 2)
+        {
+          rt.reserve(rt.size() + s.size());
```
cbielow, line 390: these reserves are probably a pessimization

jpfeuffer:
@timosachsenberg is this still needed? I guess not. @axelwalter was happy with the current speed.
If we continue we need to discuss the API of this function. this is just a suggestion.

timosachsenberg:
not sure... probably nice to have. not high priority though

PR #5740, opened 2022-01-10
[BUILD] Try to silence NPY Deprecated warnings in pyopenms

jpfeuffer:
see https://stackoverflow.com/questions/52749662
Might introduce new ones, so let's see.
Can be removed when we switch to Cython 3.


timosachsenberg (approved):

jpfeuffer:
Seems to need additional changes. Not sure if worth it.
https://gitlab.kwant-project.org/kwant/kwant/-/merge_requests/251

timosachsenberg:
maybe just switch to cython3 then?

jpfeuffer:
Yes but we need to update all CI

jpfeuffer:
Cython 3 does not have a stable release yet and is not picked up automatically by pip as the newest version.

jpfeuffer:
We would still need that PR to silence the warning I think. Cython3 just makes it not fail.

jpfeuffer:
This PR is actually not going to work with the CMake-based build that we have now. Closing.
We can remember the macros and add them once we use Cython3

PR #5744, opened 2022-01-13
[ADD] MzTabM support for AccurateMassSearch (based on IdentificationData)

oliveralka:
# Description

This is a draft for integrating MzTabM support based on the integration of the IdentificationData structure in AccurateMassSearch.

It contains the MzTabM data model, a function to fill the model based on feature maps with identification data attached. 
The MztabMFile and its store function.  

[1]
I am currently not happy with the handling of MzTab-M in AccurateMassSearch. The command would be:
```
AccurateMassSearch
-in /Users/alka/Desktop/mztab-m_example/MAD_2012_02_03_PStd_10_1.featureXML
-out /Users/alka/Desktop/mztab-m_example/mztabm/2022/MAD_2012_02_03_PStd_10_1_fmap_only_mztab-m_ID_H_NA_latest.mztab
-out_annotation /Users/alka/Desktop/mztab-m_example/mztabm/2022/MAD_2012_02_03_PStd_10_1_fmap_o_mztab_ID_H_NA_latest.oms
-db:mapping /Users/alka/Desktop/mztab-m_example/databases/HMDBMappingFile.tsv
-db:struct /Users/alka/Desktop/mztab-m_example/databases/HMDB2StructMapping.tsv
-positive_adduct /Users/alka/Desktop/mztab-m_example/databases/PositiveAdducts.tsv
-negative_adducts /Users/alka/Desktop/mztab-m_example/databases/NegativeAdducts.tsv
-algorithm:mass_error_value 1.0
-algorithm:id_format ID
-mztabm
```
Since it has the same suffix as MzTab (.mztab). I am not quite sure how to handle it in the FileHandler. Currently, it is handled directly in AccurateMassSearch. 

[2] 
Another issue is the "ordering" of the multi-index containers based on memory addresses. It is hard to compare files in the MzTabMFile_test, since the structures changes with each iteration. 

Hendrik's suggestion to sort the rows is not applicable to the current implementation. Since the ordering of identifications in the evidence section is needed to build the small molecule section. 

> Yes, or - depending on the design of the MzTabM implementation - you could sort the rows in the MzTab. That's how I've done it for the MzTab export in `IdentificationDataConverter`; see e.g.: https://github.com/hendrikweisser/OpenMS/blob/d7b6cb79d9e2abefc570122671298ffd91f76aa9/src/openms/source/METADATA/ID/IdentificationDataConverter.cpp#L737

Here, I would have to sort a temporary index of the "IdentificationDataInternal::ObservationMatchRef&" (as far as I understood). I played around with it, but I actually need some help to figure out how that works. 

# Checklist:

# Note:


src/openms/source/FORMAT/MzTabM.cpp
```diff
@@ -0,0 +1,738 @@
...
+        addMetaInfoToOptionalColumns(feature_user_value_keys, smf.opt_, String("global"), f);
+
+        smfs.emplace_back(smf);
+        ++feature_section_entry_counter;
+      }
+      else
+      {
+        // feature row based on number of individual identifications and adducts!
+        std::map<String, std::vector<int>> evidence_id_ref_per_adduct;
+        for (const IdentificationDataInternal::ObservationMatchRef& match_ref : match_refs) // iterate over all identifications of a feature
```
oliveralka, line 573: IMHO, the sorting of `IdentificationDataInternal::ObservationMatchRef&` should take place before the for loop. 

I tried:
`std::sort(match_refs.begin(), match_refs.end(), MzTabM::myCompare_);`

```
bool myCompare_(const IdentificationDataInternal::ObservationMatchRef& lhs, const IdentificationDataInternal::ObservationMatchRef& rhs) {
    return std::tie(lhs->identified_molecule_var.getIdentifiedCompoundRef()->identifier) < std::tie(rhs->identified_molecule_var.getIdentifiedCompoundRef()->identifier);
  }
```

But it does not seem to be compatible with the boost structure. Any ideas, suggestions here?

timosachsenberg:
> Since it has the same suffix as MzTab (.mztab). I am not quite sure how to handle it in the FileHandler. Currently, it is handled directly in AccurateMassSearch.

On reading? I think one can check e.g., the first lines one get's back and you could compare the MzTab version string.

src/openms/source/FORMAT/MzTabM.cpp
timosachsenberg: The issue 
is that the operator< is defined as the Address - which is not stable: https://github.com/OpenMS/OpenMS/blob/7f5009cfbabe74c126eece86f86cb9b7a5e4af62/src/openms/include/OpenMS/METADATA/ID/MetaData.h#L51

typedef IteratorWrapper<ObservationMatches::iterator> ObservationMatchRef;

One way this could be done is by adding a template specialization on ObservationMatches::iterator to IteratorWrapper that defines the operator< as you did. But this will probably not work with python. So I would suggest to create a 
```
struct ObservationMatchRef : public ObservationMatches::iterator
{
  bool operator<  ...
}

this should be wrappable in pyOpenMS and then provide your comparison operator

Comment (timosachsenberg) [reply src/openms/source/FORMAT/MzTabM.cpp:573]: I think we need to define both operator< and the equality operator in a meaningful way. Otherwise we can't properly compare ID data structures.

File: src/utils/AccurateMassSearch.cpp [hunk 2]
```diff
@@ -99,9 +102,10 @@ class TOPPAccurateMassSearch :
     setValidFormats_("in", {"featureXML", "consensusXML"});
     registerOutputFile_("out", "<file>", "", "mzTab file");
     setValidFormats_("out", ListUtils::create<String>("mzTab"));
```
oliveralka, line 104: @timosachsenberg: FileHandling not only on reading but more on writing MzTab, since only "mzTab" is registered in the FileHandler and not "mztabm" - due to the same suffix. So handling if MzTab or MzTab-M should be written in the end is currently handled by the tool e.g. via a flag (here -mztabm).

timosachsenberg: why is the flag needed?

timosachsenberg: you want to support both?

src/openms/source/FORMAT/MzTabM.cpp
timosachsenberg: I mean for IdentificationData as a whole as well. So basically we can check if two ID are the same (everything except for the Addresses)

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
```diff
@@ -572,19 +575,93 @@ namespace OpenMS
...
+      IdentificationData::SearchParamRef search_param_ref = id.registerDBSearchParam(search_param);
+
+      // file has been processed by software performing a specific processing action.      
+      std::set<DataProcessing::ProcessingAction> actions;
+      actions.insert(DataProcessing::IDENTIFICATION);
+      IdentificationData::ProcessingStep step(sw_ref, file_refs, DateTime::now(), actions);
+      step_ref = id.registerProcessingStep(step, search_param_ref);
+      id.setCurrentProcessingStep(step_ref); // add the new step
+    }
+
```
hendrikweisser, line 664: Indentation seems off below this point (lines 665-678).

src/openms/source/FORMAT/MzTabMFile.cpp
```diff
@@ -0,0 +1,678 @@
...
+
+    TextFile tmp_out;
+    for (TextFile::ConstIterator it = out.begin(); it != out.end(); ++it)
+    {
+      tmp_out.addLine(*it);
+    }
+    tmp_out.store(filename);
+  }
+
+  void MzTabMFile::addOptionalColumnsToSectionRow_(const std::vector<String>& column_names, const std::vector<MzTabOptionalColumnEntry>& column_entries, StringList& output) const
```
hendrikweisser, line 657: Is this duplicated between MzTab and MzTabM?

hendrikweisser (approved):
I'm not familiar with MzTabM, but overall the code looks good to me. Nice work!

src/utils/AccurateMassSearch.cpp
oliveralka: Yes, at least in a transition period - I think we should support both formats since a lot of older post-processing workflows are dependent on the MzTab format.

timosachsenberg: maybe I would prefer having a string option -mzTab_version with valid strings "mzTab-1.0.0" and "mzTab-M-v1.0.0" ? what do you think?

oliveralka: Yes, that would be fine as well :). I was only thinking if that could somehow be handled at the FileHandler level directly without needing the implementation for every tool (it applies to).

src/openms/source/FORMAT/MzTabM.cpp
oliveralka: Thanks for the help! 
In terms of general comparison operators - I think that goes a bit too far for this PR - probably should be added to an Issue?

timosachsenberg: I have it as an unchecked box in hendriks PR so we kind of know that it (likely) needs to be done

src/openms/source/FORMAT/MzTabMFile.cpp
oliveralka: Yes, correct, that is some minor duplication, but I wanted to keep the main implementation separate without adding another "base class", which IMHO would be overkill here. 

I probably could add MzTabMFile as a friend to MzTabFile, to allow the usage of the protected member function, but I think keeping them separate with some (minor) duplication is the cleaner option here.

hendrikweisser: You've already made the `MzTabBase` class. Wouldn't it fit there?

oliveralka: Hm, that would then be a mix of the mzTab specification data structure and the actual mztab file (reading/writing). I think I would not do that. @timosachsenberg What is your take on that?

src/openms/source/ANALYSIS/ID/AccurateMassSearchEngine.cpp
oliveralka: When I view the complete file (here or in my editor) it seems to be correct. I think it is an artifact of the representation in review mode.

src/openms/source/FORMAT/MzTabMFile.cpp
hendrikweisser: Alternatively, would it make sense to make the function public?

oliveralka:
```
-- Build files have been written to: /Users/alka/Documents/work/software/openms_test/pyopenms
(build_pyopenms_39) ➜  pyopenms make pyopenms
Consolidate compiler generated dependencies of target OpenSwathAlgo
[  0%] Built target OpenSwathAlgo
[  0%] Automatic MOC for target OpenMS
[  0%] Built target OpenMS_autogen
Consolidate compiler generated dependencies of target OpenMS
[100%] Built target OpenMS
[100%] Creating C++ extension with autowrap and Cython
Build type is:  Release
Number of submodules:  8
Number of concurrent threads:  4
ADDED __str__ method to DataFilter
ADDED __str__ method to MZTrafoModel
ADDED __str__ method to DataValue
ADDED __str__ method to EmpiricalFormula
ADDED __str__ method to NASequence
ADDED __str__ method to RANSACParam
ADDED __str__ method to MassDecomposition
ADDED __str__ method to AASequence
  File "/Users/alka/Documents/work/software/openms_test/pyopenms/pyOpenMS/create_cpp_extension.py", line 222, in <module>
    autowrap_include_dirs = doCythonCodeGeneration(modname, allDecl_mapping, instance_map, converters)
  File "/Users/alka/Documents/work/software/openms_test/pyopenms/pyOpenMS/create_cpp_extension.py", line 30, in doCythonCodeGeneration
    autowrap_include_dirs = autowrap.generate_code(allDecl_mapping[modname]["decls"], instance_map,
  File "/usr/local/miniconda3/envs/build_pyopenms_39/lib/python3.9/site-packages/autowrap/__init__.py", line 69, in generate_code
    gen.create_pyx_file(debug)
  File "/usr/local/miniconda3/envs/build_pyopenms_39/lib/python3.9/site-packages/autowrap/CodeGenerator.py", line 220, in create_pyx_file
    create_for(ResolvedEnum, self.create_wrapper_for_enum, self.enum_codes)
  File "/usr/local/miniconda3/envs/build_pyopenms_39/lib/python3.9/site-packages/autowrap/CodeGenerator.py", line 215, in create_for
    method(resolved, codes)
  File "/usr/local/miniconda3/envs/build_pyopenms_39/lib/python3.9/site-packages/autowrap/CodeGenerator.py", line 370, in create_wrapper_for_enum
    self.class_codes[class_name].add(code)
AttributeError: 'list' object has no attribute 'add'
make[3]: *** [pyOpenMS/pyopenms/pyopenms_1.cpp] Error 1
make[2]: *** [src/pyOpenMS/CMakeFiles/compile_pxds.dir/all] Error 2
make[1]: *** [src/pyOpenMS/CMakeFiles/pyopenms.dir/rule] Error 2
make: *** [pyopenms] Error 2
```
Pyopenms fails, all other tests pass. Is there an easier way to check the issue? The "error" is somewhat not expressive.

jpfeuffer:
Which autowrap version is it using? We just released a new one with major changes.

oliveralka:
I am using 0.22.5. (-- Looking for autowrap - found autowrap 0.22.5, version ok)

jpfeuffer:
Can you:
Replace the list arguments in all four defaultdicts to `lambda: Code()`? And try again?
https://github.com/OpenMS/autowrap/blob/master/autowrap/CodeGenerator.py#L170

In your `/usr/local/miniconda3/envs/build_pyopenms_39/lib/python3.9/site-packages/autowrap/CodeGenerator.py`

jpfeuffer:
Or rather `lambda: Code.Code()`

oliveralka:
Ok, thanks! Your the suggestion works - now the error output is generated accordingly. 
```
 170         self.enum_codes = defaultdict(lambda: Code.Code())
 171         self.class_codes = defaultdict(lambda: Code.Code())
 172         self.class_codes_extra = defaultdict(lambda: Code.Code())
 173         self.class_pxd_codes = defaultdict(lambda: Code.Code())
```
Errors:
```
Build type is:  Release
Number of submodules:  8
Number of concurrent threads:  4
ADDED __str__ method to DataFilter
ADDED __str__ method to MZTrafoModel
ADDED __str__ method to DataValue
ADDED __str__ method to EmpiricalFormula
ADDED __str__ method to NASequence
ADDED __str__ method to RANSACParam
ADDED __str__ method to MassDecomposition
ADDED __str__ method to AASequence
Autowrap has wrapped 76 classes, 697 methods and 14 enums
Autowrap has wrapped 72 classes, 785 methods and 10 enums
Autowrap has wrapped 82 classes, 791 methods and 11 enums
Autowrap has wrapped 79 classes, 767 methods and 12 enums
Autowrap has wrapped 74 classes, 894 methods and 11 enums
Autowrap has wrapped 79 classes, 668 methods and 11 enums
Autowrap has wrapped 89 classes, 814 methods and 10 enums
Autowrap has wrapped 86 classes, 771 methods and 10 enums
Cython compile pyopenms/pyopenms_2.pyx
Cython compile pyopenms/pyopenms_1.pyx
Cython compile pyopenms/pyopenms_3.pyx
Cython compile pyopenms/pyopenms_4.pyx

Error compiling Cython file:
...
    # simply holds the ptr and does not do any memory management.
    # see autowrap/data_files/autowrap/README.md for implementation
    # cdef AutowrapPtrHolder[_CrossLinksDB] inst

    def __init__(self):
      self.inst = AutowrapPtrHolder[_CrossLinksDB](_getInstance_CrossLinksDB())
                                                 ^

pyopenms/pyopenms_4.pyx:3122:50: Cannot assign type 'AutowrapPtrHolder[CrossLinksDB]' to 'shared_ptr[CrossLinksDB]'

Error compiling Cython file:
...

    # NOTE: using shared_ptr for a singleton will lead to segfaults, use raw ptr instead
    # cdef AutowrapPtrHolder[_RibonucleotideDB] inst

    def __init__(self):
      self.inst = AutowrapPtrHolder[_RibonucleotideDB](_getInstance_RibonucleotideDB())
                                                     ^

pyopenms/pyopenms_4.pyx:10330:54: Cannot assign type 'AutowrapPtrHolder[RibonucleotideDB]' to 'shared_ptr[RibonucleotideDB]'

Error compiling Cython file:
...
    
    # NOTE: using shared_ptr for a singleton will lead to segfaults, use raw ptr instead
    # cdef AutowrapPtrHolder[_ProteaseDB] inst

    def __init__(self):
      self.inst = AutowrapPtrHolder[_ProteaseDB](_getInstance_ProteaseDB()) 
                                               ^

pyopenms/pyopenms_3.pyx:8570:48: Cannot assign type 'AutowrapPtrHolder[ProteaseDB]' to 'shared_ptr[ProteaseDB]'
Cython compile pyopenms/pyopenms_5.pyx
Cython compile pyopenms/pyopenms_6.pyx
Cython compile pyopenms/pyopenms_7.pyx
Cython compile pyopenms/pyopenms_8.pyx

Error compiling Cython file:
...
    
    # NOTE: using shared_ptr for a singleton will lead to segfaults, use raw ptr instead
    # cdef AutowrapConstPtrHolder[_ElementDB] inst

    def __init__(self):
      self.inst = AutowrapConstPtrHolder[_ElementDB](_getInstance_ElementDB())
                                                   ^

pyopenms/pyopenms_6.pyx:2741:52: Cannot assign type 'AutowrapConstPtrHolder[ElementDB]' to 'shared_ptr[ElementDB]'

Error compiling Cython file:
...
    
    # NOTE: using shared_ptr for a singleton will lead to segfaults, use raw ptr instead
    # cdef AutowrapPtrHolder[_ModificationsDB] inst

    def __init__(self):
      self.inst = AutowrapPtrHolder[_ModificationsDB](_getInstance_ModificationsDB())
                                                    ^

pyopenms/pyopenms_7.pyx:9477:53: Cannot assign type 'AutowrapPtrHolder[ModificationsDB]' to 'shared_ptr[ModificationsDB]'

Error compiling Cython file:
...
    
    # NOTE: using shared_ptr for a singleton will lead to segfaults, use raw ptr instead
    # cdef AutowrapPtrHolder[_RNaseDB] inst

    def __init__(self):
      self.inst = AutowrapPtrHolder[_RNaseDB](_getInstance_RNaseDB())
                                            ^

pyopenms/pyopenms_7.pyx:13158:45: Cannot assign type 'AutowrapPtrHolder[RNaseDB]' to 'shared_ptr[RNaseDB]'

Error compiling Cython file:
...
    
    # NOTE: using shared_ptr for a singleton will lead to segfaults, use raw ptr instead
    # cdef AutowrapPtrHolder[_ResidueDB] inst

    def __init__(self):
      self.inst = AutowrapPtrHolder[_ResidueDB](_getInstance_ResidueDB())
                                              ^

pyopenms/pyopenms_7.pyx:13292:47: Cannot assign type 'AutowrapPtrHolder[ResidueDB]' to 'shared_ptr[ResidueDB]'
```

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+
+  struct CompareMzTabMMatchRef
+  {
+    bool operator() (const IdentificationDataInternal::ObservationMatchRef& lhs, const IdentificationDataInternal::ObservationMatchRef& rhs)  const
+    {
+      return lhs->identified_molecule_var.getIdentifiedCompoundRef()->identifier < rhs->identified_molecule_var.getIdentifiedCompoundRef()->identifier;
+    }
+  };
+
+  class OPENMS_DLLAPI MzTabMAssayMetaData
```
timosachsenberg, line 60: If possible add some basic documentation to the the public classes and structs

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+    std::map<Size, MzTabCVMetaData> cv; ///< Controlled Vocabulary details
+    std::map<Size, MzTabMDatabaseMetaData> database; ///< Database details
+    std::map<Size, MzTabParameter> derivatization_agent; ///<
+    MzTabParameter small_molecule_quantification_unit; ///< Description of the unit type used
+    MzTabParameter small_molecule_feature_quantification_unit; ///< Description of the unit type used
+    MzTabParameter small_molecule_identification_reliability; ///< Reliability of identification (4-level schema)
+    std::map<Size, MzTabParameter> id_confidence_measure; ///< Confidence measures / scores
+    // https://github.com/HUPO-PSI/mzTab/blob/master/specification_document-releases/2_0-Metabolomics-Release/mzTab_format_specification_2_0-M_release.adoc#6260-colunit-small_molecule_feature
+    // for a specific optional column?
+    // Should be in optional column metadata?
```
timosachsenberg, line 135: what about this comment?

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+    MzTabStringList smf_id_refs; ///< References to all the features on which quantification has been based.
+    MzTabStringList database_identifier; ///< Names of the used databases.
+    MzTabStringList chemical_formula; ///< Potential chemical formula of the reported compound.
+    MzTabStringList smiles; ///< Molecular structure in SMILES format.
+    MzTabStringList inchi; ///< InChi of the potential compound identifications.
+    MzTabStringList chemical_name; ///< Possible chemical/common names or general description
+    MzTabStringList uri; ///< The source entry’s location.
+    MzTabDoubleList theoretical_neutral_mass; ///< Precursor theoretical neutral mass
+    MzTabStringList adducts; ///< Adducts
+    // TODO: https://github.com/HUPO-PSI/mzTab/blob/master/specification_document-releases/2_0-Metabolomics-Release/mzTab_format_specification_2_0-M_release.adoc#6311-reliability
```
timosachsenberg, line 156: does this TODO refer to MzTabM or the tools?

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+    // Where can that be added appropriately - for each identification method individually?
+    MzTabString reliability; ///< Reliability of the given small molecule identification
+    MzTabParameter best_id_confidence_measure; ///< The identification approach with the highest confidence
+    MzTabDouble best_id_confidence_value; ///< The best confidence measure
+    std::map<Size, MzTabDouble> small_molecule_abundance_assay; ///<
```
timosachsenberg, line 161: empty doc lines

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+
+    static String getAdductString_(const IdentificationDataInternal::ObservationMatchRef& match_ref);
+
+    static void getFeatureMapMetaValues_(const FeatureMap& feature_map,
+                                         std::set<String>& feature_user_value_keys,
+                                         std::set<String>& observationmatch_user_value_keys,
+                                         std::set<String>& compound_user_value_keys);
+
+    template <class ForwardIterator>
+    static void replaceWhiteSpaces_(ForwardIterator first, ForwardIterator last)
```
timosachsenberg, line 291: why is this needed? do the standard stl algorithms or our String helper don't work?

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+    {
+      while (first!=last)
+      {
+        first->substitute(' ', '_');
+        ++first;
+      }
+    }
+
+    static void replaceWhiteSpaces_(std::set<String>& keys)
```
timosachsenberg, line 300: same here... seems redundant to the function above

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+        ++first;
+      }
+      std::swap(keys, tmp_keys);
+    }
+
+  };
+
+} // namespace OpenMS
+
+
```
timosachsenberg, line 318: remove additional newlines

src/openms/include/OpenMS/FORMAT/MzTabMFile.h
```diff
@@ -0,0 +1,78 @@
...
+    /// Default Destructor
+    ~MzTabMFile();
+
+    /// Store MzTabM file
+    void store(const String& filename, const MzTabM& mztab_m) const;
+
+  protected:
+    void generateMzTabMMetaDataSection_(const MzTabMMetaData& map, StringList& sl) const;
+
+    String generateMzTabMSmallMoleculeHeader_(const MzTabMMetaData& meta, const std::vector<String>& optional_columns, size_t& n_columns) const;
```
timosachsenberg, line 65: why does n_columns need to be passed? please add doc

src/tests/class_tests/openms/source/OMSFile_test.cpp
```diff
@@ -139,24 +140,52 @@ START_SECTION(void store(const String& filename, const FeatureMap& features))
...
   TEST_EQUAL(features[0].getSubordinates().size(), 2);
 
   IdentificationDataConverter::exportFeatureIDs(features);
-  FeatureXMLFile().store(OPENMS_GET_TEST_DATA_PATH("OMSFile_test_2.featureXML"),
-                         features);
+
+  features.sortByPosition();
+
+  // sort for reproducibility
+  auto proteins = features.getProteinIdentifications();
```
timosachsenberg, line 162: can this be an auto& ?

src/tests/class_tests/openms/source/OMSFile_test.cpp
```diff
@@ -139,24 +140,52 @@ START_SECTION(void store(const String& filename, const FeatureMap& features))
...
+  for (auto& protein : proteins)
+  {
+    protein.sort();
+  }
+  auto un_peptides = features.getUnassignedPeptideIdentifications();
```
timosachsenberg, line 167: same here can this bie auto&

src/tests/class_tests/openms/source/OMSFile_test.cpp
```diff
@@ -139,24 +140,52 @@ START_SECTION(void store(const String& filename, const FeatureMap& features))
...
+  for (auto& un_pep : un_peptides)
+  {
+    un_pep.sort();
+  }
+
+  features.setProteinIdentifications(proteins);
```
timosachsenberg, line 173: then these would not be necessary

src/tests/class_tests/openms/source/MzTabM_test.cpp
```diff
@@ -0,0 +1,368 @@
...
+    END_SECTION
+
+    START_SECTION(~MzTabM())
+        {
+          delete ptr;
+        }
+    END_SECTION
+
+    START_SECTION(Fill data structure)
+        {
```
timosachsenberg, line 68: seems to have some strange indentation here

src/tests/topp/CMakeLists.txt
```diff
@@ -2664,6 +2664,21 @@ set_tests_properties("UTILS_AccurateMassSearch_2_out2" PROPERTIES DEPENDS "UTILS
 add_test("UTILS_AccurateMassSearch_3" ${TOPP_BIN_PATH}/AccurateMassSearch -test -in ${DATA_DIR_TOPP}/AccurateMassSearch_2_input.featureXML -out AccurateMassSearch_3_output.tmp.mzTab -out_annotation AccurateMassSearch_2_output.tmp.featureXML -db:mapping ${DATA_DIR_TOPP}/AMS_test_Mapping.tsv -db:struct ${DATA_DIR_TOPP}/AMS_test_Struct.tsv -positive_adducts ${DATA_DIR_TOPP}/AMS_PositiveAdducts.tsv -negative_adducts ${DATA_DIR_TOPP}/AMS_NegativeAdducts.tsv)
 add_test("UTILS_AccurateMassSearch_3_out1" ${DIFF} -in1 AccurateMassSearch_3_output.tmp.mzTab -in2 ${DATA_DIR_TOPP}/AccurateMassSearch_3_output.mzTab -whitelist "")
 set_tests_properties("UTILS_AccurateMassSearch_3_out1" PROPERTIES DEPENDS "UTILS_AccurateMassSearch_3")
+# TODO: How to test the oms file?
```
timosachsenberg, line 2667: what about this test?

src/utils/AccurateMassSearch.cpp
```diff
@@ -146,11 +152,16 @@ class TOPPAccurateMassSearch :
     ams_param.setValue("positive_adducts", getStringOption_("positive_adducts"));
     ams_param.setValue("negative_adducts", getStringOption_("negative_adducts"));
 
+    if (file_ann.hasSuffix("oms"))
+    {
+      ams_param.setValue("id_format", "ID"); // enable novel ID data structure
```
timosachsenberg, line 157:
```suggestion
      ams_param.setValue("id_format", "ID"); // use IdentificationData to store id results
```

src/utils/AccurateMassSearch.cpp
```diff
@@ -193,15 +207,22 @@ class TOPPAccurateMassSearch :
...
-      // annotate output with data processing info
-      //addDataProcessing_(ms_feat_map, getProcessingInfo_(DataProcessing::IDENTIFICATION_MAPPING));
       if (!file_ann.empty())
       {
         ConsensusXMLFile().store(file_ann, ms_cons_map);
       }
     }
 
-    mztab_outfile.store(out, mztab_output);
+    if(mztabm && !(filetype == FileTypes::CONSENSUSXML))  // works for FeatureMap only (featureXML input)
```
timosachsenberg, line 216: output error/warning if applied to wrong type.
document

src/tests/topp/CMakeLists.txt
oliveralka: This is more a general thing - how can we compare OMS files directly. I think there is currently no way to do it. That is why we check if it is empty in the OMSFile_test. 

Here, we check the mztabm and the featureXML instead - which hints at a correct oms file. 

Still, in the future, I think we need something to test oms files directly.

src/utils/AccurateMassSearch.cpp
```diff
@@ -99,9 +102,11 @@ class TOPPAccurateMassSearch :
     setValidFormats_("in", {"featureXML", "consensusXML"});
     registerOutputFile_("out", "<file>", "", "mzTab file");
     setValidFormats_("out", ListUtils::create<String>("mzTab"));
+    registerStringOption_("mzTab_version", "<choice>", "mzTab-1.0.0", "mzTab version", false);
+    setValidStrings_("mzTab_version", ListUtils::create<String>("mzTab-1.0.0,mzTab-M-v1.0.0"));
```
timosachsenberg, line 106: I think in a file there was v2.0.0 ?

src/utils/AccurateMassSearch.cpp
```diff
@@ -1,3 +1,4 @@
+
```
timosachsenberg, line 1: remove extra newline

src/openms/include/OpenMS/FORMAT/MzTabFile.h
```diff
@@ -226,6 +226,8 @@ namespace OpenMS
                                   const std::map<String, Size>& map_run_to_num_sub
                                   );
 
+  private:
+    friend class MzTabMFile;
```
timosachsenberg, line 230: why is dependency needed?
if possible, get rid of friend

src/openms/include/OpenMS/FORMAT/MzTabM.h
```diff
@@ -0,0 +1,320 @@
...
+namespace OpenMS
+{
+  /**
+      @brief Data model of MzTabM files.
+      Please see the official MzTabM specification at
+      https://github.com/HUPO-PSI/mzTab/tree/master/specification_document-releases/2_0-Metabolomics-Release
+      @ingroup FileIO
+  */
+
+  struct CompareMzTabMMatchRef
```
timosachsenberg, line 52: move this to ID to replace iteratorwrapper

src/openms/include/OpenMS/ANALYSIS/ID/AccurateMassSearchEngine.h
```diff
@@ -228,7 +229,7 @@ namespace OpenMS
 
     /// main method of AccurateMassSearchEngine
     /// input map is not const, since it will get annotated with results
-    void run(FeatureMap&, MzTab&) const;
+    void run(FeatureMap&, MzTab&, MzTabM&) const;
```
timosachsenberg, line 232: it seems a bit wasteful to generate both files. especially for accuratemasssearch these could get pretty large.
could this be easily refactored to produce two run methods with minimal code dublication?

oliveralka: I did it as it is since I was thinking that it is not easily possible to provide two methods with minima code duplication. But we can take a look together if you like.

timosachsenberg: yeah that would be great.

src/tests/topp/CMakeLists.txt
timosachsenberg: yeah this is pretty bad but should be addressed on the OMS side in the future

src/openms/include/OpenMS/FORMAT/MzTabMFile.h
oliveralka: Add documentation. FYI: This stores the number of columns used in the header or the row, which is later used in a postcondition, to sanity check if the header and rows have the same length (columnwise).

hendrikweisser:
> Can we do an explicit specialization for the Ref types like this?

I don't think it's a good idea to this a) only for this one case, b) only to solve sorting of rows in MzTabM. In my opinion the IteratorWrappers are a separate issue that we're already considering for pyOpenMS, so I'd prefer a solution here that works independently of the IdentificationData internals.

jpfeuffer:
@oliveralka Can you try pyopenms with autowrap 0.22.6? defaultdict is also fixed there.

oliveralka:
@jpfeuffer autowrap 0.22.6 seems to have fixed all issues! Thanks!

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
```diff
@@ -146,20 +146,21 @@ namespace OpenMS {
     return output;
   }
 
-  void FIAMSDataProcessor::runAccurateMassSearch(FeatureMap& input, OpenMS::MzTab& output, OpenMS::MzTabM& output2) {
+  void FIAMSDataProcessor::runAccurateMassSearch(FeatureMap& input, OpenMS::MzTab& output) {
```
timosachsenberg, line 149:
```suggestion
  void FIAMSDataProcessor::runAccurateMassSearch(FeatureMap& input, OpenMS::MzTab& output)
  {
```

src/openms/source/FORMAT/MzTab.cpp
```diff
@@ -2394,7 +2394,7 @@ Not sure how to handle these:
       const Feature& f = feature_map[i];
       vector<String> keys;
       f.getKeys(keys); //TODO: why not just return it?
-      replaceWhiteSpaces_(keys.begin(), keys.end());
+      std::transform(keys.begin(), keys.end(), keys.begin(), [&](String& s) { return s.substitute(' ', '_'); });
```
timosachsenberg, line 2397:
```suggestion
      // replace whitespaces with underscore
      std::transform(keys.begin(), keys.end(), keys.begin(), [&](String& s) { return s.substitute(' ', '_'); });
```

timosachsenberg: You use this funcitonality a few times in the cpp file. You could just add it to the cpp file as a member function (not the header)

timosachsenberg: alternatively add a comment to each of them

src/openms/source/ANALYSIS/ID/FIAMSDataProcessor.cpp
oliveralka: The whole class (FIAMSDataProcessor) seems to follow the structure
```
function() {
}
```

oliveralka:
All tests pass (apart from clang-format linting).

jpfeuffer:
TBH I would have preferred to fix the pyopenms tests. The wrap-attach is correct to mimic the  C++ API

oliveralka:
@jpfeuffer You are right - confused me there for a second. ;)

PR #5750, opened 2022-01-17
Reduce warnings on unused functions

jpfeuffer:
with maybe_unused


timosachsenberg:
Looks good but I wonder why they are unused

jpfeuffer:
Not all of them were unused.

PR #5751, opened 2022-01-17
[FIX] pyopenms mac fix dependency order

jpfeuffer:
Still a minor error from the CMake restructue.
MacOS fix dependencies has to be run AFTER the pyopenms library build to fixup the DLLs in there too.


timosachsenberg (approved):
Awesome

PR #5765, opened 2022-01-26
Require OPENMS_TEST_VERBOSE_SUCCESS env variable set if output of successful test executions should be printed

timosachsenberg:
# Description

Output from successful tests clutter the log. In most cases this information is not needed. Optionally can now be enabled by setting the OPENMS_TEST_VERBOSE_SUCCESS environment variable

# Checklist:

# Note:

### DIFF
--- a/src/tests/class_tests/openms/CMakeLists.txt
+++ b/src/tests/class_tests/openms/CMakeLists.txt
@@ -81,7 +81,6 @@ endif()
 foreach(_class_test ${TEST_executables})
   add_executable(${_class_test} source/${_class_test}.cpp)
   target_link_libraries(${_class_test} ${OpenMS_LIBRARIES})
-  set_property(TEST ${_class_test} APPEND PROPERTY ENVIRONMENT "OPENMS_TEST_VERBOSE_SUCCESS=1")
   add_test(${_class_test} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${_class_test})
   # only add OPENMP flags to gcc linker (except Mac OS X, due to compiler bug
   # see https://sourceforge.net/apps/trac/open-ms/ticket/280 for details)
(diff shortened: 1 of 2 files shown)


src/tests/class_tests/openms/CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ endif()
 foreach(_class_test ${TEST_executables})
   add_executable(${_class_test} source/${_class_test}.cpp)
   target_link_libraries(${_class_test} ${OpenMS_LIBRARIES})
+  set_property(${_class_test} APPEND PROPERTY ENVIRONMENT "OPENMS_TEST_VERBOSE_SUCCESS")
```
timosachsenberg, line 84: remove? I think we want to pass the current environment

src/tests/class_tests/openms/CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ endif()
 foreach(_class_test ${TEST_executables})
   add_executable(${_class_test} source/${_class_test}.cpp)
   target_link_libraries(${_class_test} ${OpenMS_LIBRARIES})
+  set_property(TEST ${_class_test} APPEND PROPERTY ENVIRONMENT "OPENMS_TEST_VERBOSE_SUCCESS=1")
```
timosachsenberg, line 84:
```suggestion
```

timosachsenberg:
rebuild jenkins

cbielow:
is this only intended for strings?
And documenting the behaviour would be nice, I think.

timosachsenberg:
it is for all successful lines as these clutter our CI logs and are automatically truncated

jpfeuffer:
But I think he has a point. There is also "testRealSimilar" and others.

timosachsenberg:
ok I will exclude those lines as well.
were is the best place to document this?

cbielow:
No ideal place comes to mind, but maybe installation instructions, at the very end (`Testing your OpenMS/TOPP installation`)
?

May I suggest to query `getenv` only once in the ctor of Classtest and caching the result. Otherwise getenv will be called potentially hundreds of times, each time iterating though potentially hundreds of environment variables.

timosachsenberg:
Thanks for the input.
I am looking into alternatives now. E.g., if we can use the ctest verbosity level (which is passed to the test in OpenMS) to disable / enable output of successful tests.

PR #5773, opened 2022-01-31
[Fix] MS3 TMT IdMapping

jpfeuffer:
# Description

Before, the IsobaricAnalyzer was annotating the charge of the first fragment ion. Now it uses the MS1 precursor ion.
Also, since the exact settings, "precursor" and "correct ppm" are hard to guess during IDMapping, annotate and compare the scan_ids
used for ID instead.

### DIFF
--- a/src/openms/source/ANALYSIS/ID/IDMapper.cpp
+++ b/src/openms/source/ANALYSIS/ID/IDMapper.cpp
@@ -322,41 +322,22 @@ namespace OpenMS
           //check if we compare distance from centroid or subelements
           if (!measure_from_subelements)
           {
+            String ref_mv = "";
             if (map[cm_index].metaValueExists("id_scan_id")) // MS3 TMT
             {
-              if (ids[i].metaValueExists("spectrum_reference"))
-              {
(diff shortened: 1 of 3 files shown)


src/openms/source/ANALYSIS/ID/IDMapper.cpp
```diff
@@ -322,12 +322,41 @@ namespace OpenMS
           //check if we compare distance from centroid or subelements
           if (!measure_from_subelements)
           {
-            if (isMatch_(rt_pep - map[cm_index].getRT(), mz_pep, map[cm_index].getMZ()) && (ignore_charge_ || ListUtils::contains(current_charges, map[cm_index].getCharge())))
+            if (map[cm_index].metaValueExists("id_scan_id")) // MS3 TMT
```
timosachsenberg, line 325: Code seems to be duplicated 
- maybe first set MS3, MS2 or label-free in separate function. 
- set reference id once (either id_scan_id or scan_id)
- and use same code to push back ids?
especially as oder of if evaluation is critical here and not documented

PR #5795, opened 2022-02-07
Rewrite mutual information calculation and remove MIToolbox

khuenm:
# Description

Rewrite mutual information calculation based on implementation from MIToolbox.
Previous implementation used in its running peak 4 vectors of length n and 2 of length n^2. New implementation uses 2 vectors of length n and an unordered map of n entries. This is because the mutual information is calculated directly on the histogram of 2 input vectors instead of the probability distribution. 

Let `l` be the length of 2 input vectors `x` and `y`. `c(x)` denotes the histogram of `x` and `p(x)` denotes the probability distribution of elements in `x` (`p(x)` = `c(x)/l`). Same goes for `p(y)=c(y)/l` and `p(x,y)=c(x,y)/l`

Screenshot 2022-02-07 at 22 58 09

MIToolbox is no longer needed.

# Checklist:

# Note:


src/openswathalgo/include/OpenMS/OPENSWATHALGO/ALGO/Scoring.h
```diff
@@ -54,6 +54,18 @@ namespace OpenSwath
...
     //@{
+    typedef std::pair<unsigned int, unsigned int> pos2D;
+    /// Simple hash function for Scoring::pos2D
+    struct pair_hash 
+    {
+      template <class T1, class T2>
+      std::size_t operator () (const std::pair<T1,T2> &p) const {
+        auto h1 = std::hash<T1>{}(p.first);
+        auto h2 = std::hash<T2>{}(p.second);
+        return h1 ^ h2;
```
timosachsenberg, line 65: Would it make sense to take one of the “established” hash combine functions? E.g. does this collide more often

timosachsenberg:
Nice!

src/tests/class_tests/openswathalgo/Scoring_test.cpp
```diff
@@ -438,10 +438,10 @@ m1 = mi(x_ranking,y_ranking)
   std::vector<double> data2 = {15.8951349258423, 41.5446395874023, 76.0746307373047, 109.069435119629, 111.90364074707, 169.79216003418,
                                121.043930053711, 63.0136985778809, 44.6150207519531, 21.4926776885986, 7.93575811386108};
   std::vector<unsigned int> rank_vec1, rank_vec2;
-  Scoring::computeAndAppendRank(data1, rank_vec1);
-  Scoring::computeAndAppendRank(data2, rank_vec2);
+  unsigned int max_rank1 = Scoring::computeAndAppendRank(data1, rank_vec1);
+  unsigned int max_rank2 = Scoring::computeAndAppendRank(data2, rank_vec2);
```
cbielow, line 442: can you test the max_ranks as well?

src/openswathalgo/include/OpenMS/OPENSWATHALGO/ALGO/Scoring.h
timosachsenberg: @cbielow  We could then add this to our concept or math classes? 
```
inline void hash_combine(std::size_t& seed) { }

template <typename T, typename... Rest>
inline void hash_combine(std::size_t& seed, const T& v, Rest... rest) {
    std::hash<T> hasher;
    seed ^= hasher(v) + 0x9e3779b9 + (seed<<6) + (seed>>2);
    hash_combine(seed, rest...);
}
```

cbielow: yep sure.
I'll open a ticket so we do not forget.
Student projects are coming up here, so I can put that on the agenda.
I'd suggest to keep this PR as is and deal with hashing in one go in a new PR.

cbielow:
note: do not merge yet! There will be another PR which adds some tests for the correctness of mutual information results, so that we can be sure that the new implementation behaves identical to the MI toolbox.

cbielow:
rebuild jenkins

cbielow (approved):
looks good to me!

Speedup for calculating the mutual information was about x20, on that small test data set, right?
(just for reference).

cbielow:
rebuild jenkins

cbielow:
CI passes (even if some sporadically fail)

khuenm:
> looks good to me!
> 
> Speedup for calculating the mutual information was about x20, on that small test data set, right? (just for reference).

Yes! Here's a screenshot from quick-bench :D

Screenshot 2022-03-04 at 12 58 39

src/openswathalgo/include/OpenMS/OPENSWATHALGO/ALGO/Scoring.h
FreezB11: @timosachsenberg sir can i understand the requirement here i saw a issue also that was open so i just needed to understand what type of hashing is required

PR #5797, opened 2022-02-08
remove WM5 from headers

timosachsenberg:
# Description

To prevent symbols being exported.
RegressionUtils.h was removed for same reasons (did not safe code anyways).
As well as the test which was not run anyways.

# Checklist:

# Note:


src/openms/source/MATH/STATISTICS/QuadraticRegression.cpp
```diff
@@ -34,21 +34,120 @@
...
+      double weight = *wIter;
+
+      sumX += weight * x;
+      sumXX += weight * x * x;
+      sumXXX += weight * x * x * x;
+      sumXXXX += weight * x * x * x * x;
+
+      sumY += weight * y;
+      sumXY += weight * x * y;
+      sumXXY += weight * x * x * y;
```
jpfeuffer, line 121: this also screams for optimization, unless the compiler is very smart

jpfeuffer (changes requested):

src/openms/source/MATH/STATISTICS/QuadraticRegression.cpp
jpfeuffer, line 121:
```suggestion
      summand = weight;
      sumY += summand * y;
      summand *= x;
      sumX += summand;
      sumXY += summand * y;
      summand *= x;
      sumXX += summand;
      sumXXY += summand * y;
      summand *= x;
      sumXXX += summand;
      summand *= x;
      sumXXXX += summand;
```

jpfeuffer: but I don't know if worth it.

timosachsenberg: My guess is that the compiler is very smart here.

jpfeuffer:
https://stackoverflow.com/questions/60463014/cmath-error-on-macos-c17-call-to-abs-ambiguous
please leave fabs

timosachsenberg:
> https://stackoverflow.com/questions/60463014/cmath-error-on-macos-c17-call-to-abs-ambiguous please leave fabs

I thought this only holds for integral types and not for double. Are you sure this causes an error? Or is cmath missing?

jpfeuffer:
Yes it's in Jenkins.
Abs only works for floats since c17 and apparently not in all versions of Apples libc++

src/openms/source/FILTERING/SMOOTHING/LowessSmoothing.cpp
```diff
@@ -116,7 +117,7 @@ namespace OpenMS
     }
 
     // 0 <= u < t; u is regarded as 0.0 if fabs(u) falls below epsilon
-    if ((fabs(u) < std::numeric_limits<double>::epsilon() || (0.0 < u)) && (u < t))
+    if ((std::abs(u) < std::numeric_limits<double>::epsilon() || (0.0 < u)) && (u < t))
```
timosachsenberg, line 120:
```suggestion
    if ((fabs(u) < std::numeric_limits<double>::epsilon() || (0.0 < u)) && (u < t))
```

src/openms/source/FILTERING/SMOOTHING/LowessSmoothing.cpp
```diff
@@ -79,7 +80,7 @@ namespace OpenMS
       // Size inner_idx = 0;
       for (Size inner_idx = 0; inner_idx < input_size; ++inner_idx)
       {
-        distances[inner_idx] = std::fabs(input_x[outer_idx] - input_x[inner_idx]);
+        distances[inner_idx] = std::abs(input_x[outer_idx] - input_x[inner_idx]);
```
timosachsenberg, line 83:
```suggestion
        distances[inner_idx] = fabs(input_x[outer_idx] - input_x[inner_idx]);
```

jpfeuffer:
I think it has to be std::fabs

src/openms/source/MATH/STATISTICS/QuadraticRegression.cpp
cbielow: https://godbolt.org/z/9vnah9Gb9

it is (as expected), but only in `-O2`. So in debug mode, we're stuck with slow code.

jpfeuffer:
Should be fine. Windows out of space.

PR #5808, opened 2022-02-10
[CMake] Move non-optional simulation models to non-examples share and improve installation granularity

jpfeuffer:
mainly for conda


jpfeuffer:
win still full

PR #5776, opened 2022-02-01
Rewrite of PeptideIndexer

cbielow:
# Description

complete rewrite of AhoCorascik algorithm; 
* its about twice as fast as the previous implementation (which used SeqAn)
* uses less memory
* supports ambiguous AA's in peptides

# How can I get additional information on failed tests during CI:
If your PR is failing you can check out 
- http://cdash.openms.de/index.php?project=OpenMS and look for your PR. If you click in the column that lists the failed tests you will get detailed error messages.
- Or click on the action: e.g., for clang-format linting

# Note:
- Once you opened a PR try to minimize the number of *pushes* to it as every push will trigger CI (automated builds and test) and is rather heavy on our infrastructure (e.g., if several pushes per day are performed).

# Advanced commands (admins / reviewer only):
- /rebase will try to rebase the PR on the current develop branch.
- /reformat (experimental) applies the clang-format style changes as additional commit
- setting the label NoJenkins will skip tests for this PR on jenkins (saves resources e.g., on edits that do not affect tests)

PR #5809, opened 2022-02-11
Optimize PeptideIndexer

cbielow:
# Description

This PR removes the costly storage of all peptide hits in a std::map and uses std::vector instead.
Also each hit takes up less space now.

For an (extreme) dataset with 1e9 hits (!), the runtime reduces from 1h to 15min, while the peak memory usage reduces from 134 GB to 91 GB. The resulting idXML is 34 GB in size.

There is further potential (since there is another map) but that is for another PR.

This change also surfaced a small bug in the new AC implementation (#5776), which may report hits twice (which was not discovered since the results were stored in a map). This is also fixed now.

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
@@ -423,6 +423,9 @@ namespace OpenMS
     **/
     bool addHits_(Index i, const size_t text_pos, std::vector<Hit>& hits) const;
 
+    /// same as addHits_, but only follows the suffix chain until the spawn looses its prefix
+    bool addHitsSpawn_(Index i, const ACSpawn& spawn, const size_t text_pos, std::vector<Hit>& hits, const int current_spawn_depths) const;
+
     /// Starting at node @p i, find the child with label @p edge
     /// If no child exists, follow the suffix link and try again (until root is reached)
     /// Note: This operates on the BFS trie (after compressTrie()), not the naive one.
--- a/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
@@ -423,7 +423,7 @@ namespace OpenMS
     **/
     bool addHits_(Index i, const size_t text_pos, std::vector<Hit>& hits) const;
 
-    /// same as addHits_, but only follows the suffix chain until the spawn looses its prefix
+    /// same as addHits_, but only follows the suffix chain until the spawn loses its prefix
     bool addHitsSpawn_(Index i, const ACSpawn& spawn, const size_t text_pos, std::vector<Hit>& hits, const int current_spawn_depths) const;
 
     /// Starting at node @p i, find the child with label @p edge
--- a/src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
+++ b/src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
@@ -325,6 +325,7 @@ namespace OpenMS
     }
     return hits_before != hits.size();
   }
+  
   Index ACTrie::follow_(const Index i, const AA aa) const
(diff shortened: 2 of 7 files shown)


src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
```diff
@@ -423,6 +423,9 @@ namespace OpenMS
     **/
     bool addHits_(Index i, const size_t text_pos, std::vector<Hit>& hits) const;
 
+    /// same as addHits_, but only follows the suffix chain until the spawn looses its prefix
```
jpfeuffer, line 426:
```suggestion
    /// same as addHits_, but only follows the suffix chain until the spawn loses its prefix
```

src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
```diff
@@ -295,6 +295,36 @@ namespace OpenMS
...
+      return false;
+    };
+
+    // follow chain of suffix nodes until a node does not have hits anymore
+    while (collect())
+    {
+      i = trie_[i()].suffix;
+    }
+    return hits_before != hits.size();
+  }
```
jpfeuffer, line 327:
```suggestion
  }
  
```

src/openms/source/ANALYSIS/ID/PeptideIndexing.cpp
```diff
@@ -56,17 +56,24 @@ using namespace std;
   const std::array<std::string, (Size)PeptideIndexing::MissingDecoy::SIZE_OF_MISSING_DECOY> PeptideIndexing::names_of_missing_decoy = { "error" , "warn" , "silent" };
 
 
-  // internal datastructure to store match information (not exported)
+  // internal data structure to store match information (not exported)
   struct PeptideProteinMatchInformation
   {
-    OpenMS::Size protein_index; //< index of the protein the peptide is contained in
-    OpenMS::Int position; //< the position of the peptide in the protein
+    Hit::T peptide_index; ///< index of the peptide
```
jpfeuffer, line 62:
```suggestion
    Hit::T peptide_index; //< index of the peptide
```

jpfeuffer (approved):
with minor doc suggestions

src/openms/source/ANALYSIS/ID/PeptideIndexing.cpp
cbielow: documenting members is done with `///< ....`, so I guess it would be better to fix the other two members as well?
https://www.doxygen.nl/manual/docblocks.html#memberdoc

src/openms/source/ANALYSIS/ID/PeptideIndexing.cpp
```diff
@@ -56,17 +56,24 @@ using namespace std;
...
+    Hit::T protein_index; //< index of the protein the peptide is contained in
+    Hit::T position;      //< the position of the peptide in the protein
```
cbielow, line 64:
```suggestion
    Hit::T position;      ///< the position of the peptide in the protein
```

cbielow (commented):
apply suggestions from review

src/openms/source/ANALYSIS/ID/PeptideIndexing.cpp
```diff
@@ -56,17 +56,24 @@ using namespace std;
...
 
 
-  // internal datastructure to store match information (not exported)
+  // internal data structure to store match information (not exported)
   struct PeptideProteinMatchInformation
   {
-    OpenMS::Size protein_index; //< index of the protein the peptide is contained in
-    OpenMS::Int position; //< the position of the peptide in the protein
+    Hit::T peptide_index; ///< index of the peptide
+    Hit::T protein_index; //< index of the protein the peptide is contained in
```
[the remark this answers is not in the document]
jpfeuffer: ah, yes, then the others instead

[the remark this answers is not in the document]
cbielow: ;-)

PR #5822, opened 2022-02-17
A Painter1D class for Peaks

cbielow:
# Description

This PR is a first attempt to separate the plotting of layer data in Canvas1D from the layer type.
Painting is delegated to a separate class for the essential parts.
This will allow to add painters for chromatograms, mobilograms etc in the future (some more refatoring is needed).

Going in small pieces here, so this is the first PR or some more to come.
The Painter class is not fully fledged out yet. Expect some future changes there.

# Note:

### DIFF
--- a/doc/code_examples/Tutorial_GUI_Plot1D.cpp
+++ b/doc/code_examples/Tutorial_GUI_Plot1D.cpp
@@ -28,32 +28,36 @@
 // ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 
-#include <QApplication>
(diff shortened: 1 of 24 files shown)


jpfeuffer:
Should the 1D Painter also be responsible for drawing the ion ladder?

cbielow:
> > Should the 1D Painter also be responsible for drawing the ion ladder?
> 
> I think even cleaner would be putting the ion ladder into a separate widget e.g. at the top.

yes indeed. Its on the list, but one thing at a time :)

PR #5824, opened 2022-02-18
Update PULL_REQUEST_TEMPLATE.md

cbielow:
...


jpfeuffer:
"No Jenkins" label would be appropriate here by the way :D

cbielow:
> "No Jenkins" label would be appropriate here by the way :D

:)

I realized after opening the PR. But at that point it was too late...

PR #5825, opened 2022-02-18
[FIX] add move constructor and assignment operator

hroest:
# Description

adds move constructor / assignment operator to Targeted Experiment

- probably needs some testing

### DIFF
--- a/src/openms/source/ANALYSIS/TARGETED/TargetedExperiment.cpp
+++ b/src/openms/source/ANALYSIS/TARGETED/TargetedExperiment.cpp
@@ -98,6 +98,26 @@ namespace OpenMS
   {
   }
 
+  TargetedExperiment::TargetedExperiment(TargetedExperiment && rhs) noexcept :
+    cvs_(std::move(rhs.cvs_)),
+    contacts_(std::move(rhs.contacts_)),
+    publications_(std::move(rhs.publications_)),
+    instruments_(std::move(rhs.instruments_)),
+    targets_(std::move(rhs.targets_)),
+    software_(std::move(rhs.software_)),
+    proteins_(std::move(rhs.proteins_)),
+    compounds_(std::move(rhs.compounds_)),
+    peptides_(std::move(rhs.peptides_)),
+    transitions_(std::move(rhs.transitions_)),
+    include_targets_(std::move(rhs.include_targets_)),
+    exclude_targets_(std::move(rhs.exclude_targets_)),
+    source_files_(std::move(rhs.source_files_)),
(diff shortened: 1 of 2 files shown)


timosachsenberg:
does =default not work here?

timosachsenberg:
> does =default not work here?

ok I see we have some mutable members

timosachsenberg:
> does =default not work here?

ok I see we have some mutable members. What about the reference maps? they don't get moved. what happens to them?

timosachsenberg:
> does =default not work here?

ok I see we have some mutable members

timosachsenberg:
> does =default not work here?

ok I see we have some mutable members

hroest:
@timosachsenberg yes, they are not moved and that is why we mark these maps as dirty (need to be re-constructed before use). These are pointer-based maps, so they cannot be copied. We could argue that we should re-build them upon copy but that may be unnecessary work since we can do it when actually used

src/openms/source/ANALYSIS/TARGETED/TargetedExperiment.cpp
```diff
@@ -98,6 +98,26 @@ namespace OpenMS
...
+    proteins_(std::move(rhs.proteins_)),
+    compounds_(std::move(rhs.compounds_)),
+    peptides_(std::move(rhs.peptides_)),
+    transitions_(std::move(rhs.transitions_)),
+    include_targets_(std::move(rhs.include_targets_)),
+    exclude_targets_(std::move(rhs.exclude_targets_)),
+    source_files_(std::move(rhs.source_files_)),
+    protein_reference_map_dirty_(true),
+    peptide_reference_map_dirty_(true),
+    compound_reference_map_dirty_(true)
```
hroest, line 117: @timosachsenberg the maps are empty and marked as "dirty" so they will not be used but rather re-built

hroest:
@timosachsenberg merge?

timosachsenberg:
> I think we could assume that the pointers stay valid when we move so we _could_ move the maps as well. What do you think? See https://stackoverflow.com/questions/25347599/am-i-guaranteed-that-pointers-to-stdvector-elements-are-valid-after-the-vector it seems that it is not strictly guaranteed since std::move _could_ be implemented through a copy

Yes, this is also what I saw in STL implementation - move is implemented like swap to keep iterators and pointers valid. I think if it had significant benefit (e.g, simplified the code, improve speed in typical situations or logic around marking the maps as dirty) we could do it. I

cbielow:
You could remember the address of the first element before move and check if its identical after the move to avoid the dirty flag.

PR #5826, opened 2022-02-18
 add intensity-weighted average peak picker 

hroest:
# Description

When ion mobility is present, use intensity-weighted average in a peak picker. 

- This works well if the ion mobilities are close to each other, eg if the data is already binned
- This will fail when a full ion mobility frame is provided, for this a more advanced peak picker would be required that takes into account the full 2nd dimension. It is still an improvement compared to dropping the ion mobility completely


timosachsenberg:
alsp pining @enetz here

timosachsenberg:
looks good! Can you add a few lines to the documentation (e.g., in what format the data is expected).

hroest:
@timosachsenberg this is not so straight forward but I added a short explanation on how this would work.

timosachsenberg (approved):

PR #5829, opened 2022-02-18
[FEATURE] prepare for ion mobility support in FFCentroid

hroest:
- prepare for ion mobility support in FeatureFinderCentroided
- change data structure in mass trace to point to spectrum, allowing
access to RT and ion mobility

@timosachsenberg what do you think about this approach, adding the spectrum to the mass trace will allow us to access the ion mobility and other characteristics of the spectrum from which the peak was derived (including the RT). Maybe we could have a more generalized discussion on how to do feature finding in ion mobility data?

### DIFF
--- a/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/ElutionModelFitter.cpp
+++ b/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/ElutionModelFitter.cpp
@@ -92,24 +92,22 @@ double ElutionModelFitter::calculateFitQuality_(const TraceFitter* fitter,
 {
   double mre = 0.0;
   double total_weights = 0.0;
-  double rt_start = max(fitter->getLowerRTBound(), traces[0].peaks[0].first);
+  double rt_start = max(fitter->getLowerRTBound(), traces[0].peaks[0].first->getRT());
   double rt_end = min(fitter->getUpperRTBound(),
-                      traces[0].peaks.back().first);
+                      traces[0].peaks.back().first->getRT());
 
-  for (MassTraces::const_iterator tr_it = traces.begin();
-       tr_it != traces.end(); ++tr_it)
+  for (const auto& masstrace : traces)
   {
-    for (vector<pair<double, const Peak1D*> >::const_iterator p_it =
-           tr_it->peaks.begin(); p_it != tr_it->peaks.end(); ++p_it)
+    for (const auto& peak : masstrace.peaks)
     {
-      double rt = p_it->first;
+      double rt = peak.first->getRT();
       if ((rt >= rt_start) && (rt <= rt_end))
       {
         double model_value = fitter->getValue(rt);
-        double diff = fabs(model_value * tr_it->theoretical_int -
-                           p_it->second->getIntensity());
+        double diff = fabs(model_value * masstrace.theoretical_int -
+                           peak.second->getIntensity());
         mre += diff / model_value;
-        total_weights += tr_it->theoretical_int;
+        total_weights += masstrace.theoretical_int;
       }
     }
   }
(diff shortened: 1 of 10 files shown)


src/tests/class_tests/openms/source/EGHTraceFitter_test.cpp
```diff
@@ -64,174 +65,34 @@ mt2.theoretical_int = 0.2;
...
-p1_21.setMZ(1000);
-mt1.peaks.push_back(make_pair(683.1, &p1_21));
-Peak1D p2_21;
-p2_21.setIntensity(0.270670566473f);
-p2_21.setMZ(1001);
-mt2.peaks.push_back(make_pair(683.1, &p2_21));
+
+double intensities[] = { 1.08268226589f, 0.270670566473f, 1.58318959267f, 0.395797398167f, 2.22429840363f, 0.556074600906f, 3.00248879081f, 0.750622197703f, 3.89401804768f, 0.97350451192f, 4.8522452777f, 1.21306131943f, 5.80919229659f, 1.45229807415f, 6.68216169129f, 1.67054042282f, 7.38493077109f, 1.84623269277f, 7.84158938645f, 1.96039734661f, 8.0f, 2.0f, 7.84158938645f, 1.96039734661f, 7.38493077109f, 1.84623269277f, 6.68216169129f, 1.67054042282f, 5.80919229659f, 1.45229807415f, 4.8522452777f, 1.21306131943f, 3.89401804768f, 0.97350451192f, 3.00248879081f, 0.750622197703f, 2.22429840363f, 0.556074600906f, 1.58318959267f, 0.395797398167f, 1.08268226589f, 0.270670566473f};
+double rts[] = { 677.1, 677.1, 677.4, 677.4, 677.7, 677.7, 678, 678, 678.3, 678.3, 678.6, 678.6, 678.9, 678.9, 679.2, 679.2, 679.5, 679.5, 679.8, 679.8, 680.1, 680.1, 680.4, 680.4, 680.7, 680.7, 681, 681, 681.3, 681.3, 681.6, 681.6, 681.9, 681.9, 682.2, 682.2, 682.5, 682.5, 682.8, 682.8, 683.1, 683.1};
```
hroest, line 70: the tests are now much more readable and more easy to change

timosachsenberg:
@cbielow you have more experience with the FFC codebase. To me this looks reasonable. Also pinging @enetz because he will be working on IM soon.
One note though, for our proteomicsLFQ workflow we currently use a combination of FeatureFinderMultiplex (without labels = labe-free mode) to create seeds and the targeted FeatureFinderIdentification quite successfully. It seems as if it detects way more good/identified features than using FeatureFinderCentroided. Not tested on timsTOF data though.

src/openms/source/TRANSFORMATIONS/FEATUREFINDER/ElutionModelFitter.cpp
```diff
@@ -252,11 +255,16 @@ void ElutionModelFitter::fitElutionModels(FeatureMap& features)
     }
 
     vector<Peak1D> peaks;
+    vector<MSSpectrum> tmp_exp;
     // reserve space once, to avoid copying and invalidating pointers:
     Size points_per_hull = sub.getConvexHulls()[0].getHullPoints().size();
-    peaks.reserve(feat.getSubordinates().size() * points_per_hull +
-                  (add_zeros > 0.0)); // don't forget additional zero point
+    Size required_elem = feat.getSubordinates().size() * points_per_hull +
```
cbielow, line 261: can we actually be sure that all subordinates have the same number of points?

cbielow: at least we should add a check and throw if the vector reallocated

cbielow (commented):
looks good!

Before we merge, it would be nice to know if there is a performance penalty here (due to increase in active data size and pointer indirection). 
@hroest  Could you run a PeakPicker on a common-sized example dataset and compare?

cbielow:
any updates here?

timosachsenberg:
any chance to revive this PR or the discussion?

timosachsenberg:
Discontinued... feel free to revive...

Issue #4833, opened 2020-07-30
Error Reporting in DataValue.cpp


ekawaler:
For the sake of project debugging, I would love to see some more detailed error reporting in DataValue.cpp - for instance, while using IDFileConverter, I'm getting this error: "Error occurred in line 526 of file .../openms/2.5.0/raw/OpenMS-2.5.0/src/openms/source/DATASTRUCTURES/DataValue.cpp (in function: OpenMS::DataValue::operator int() const) !", but even with the debug option on I have no idea what file is causing it to throw the error, or what the value is that it's trying and failing to convert. (And unfortunately I'm using this on a cluster where I don't have root permissions, so I can't go in and add extra error reporting in my local version.) Thanks!

jpfeuffer:
I agree mostly:
1) The exceptions in DataValue.cpp should probably also include the value that was tried to be converted. This should be a rather quick fix.
2) There is a slightly more detailed exception message inside DataValue.cpp https://github.com/OpenMS/OpenMS/blob/02ec4d48fbd3ab6a3056ab4ce58a74d96b72b153/src/openms/source/DATASTRUCTURES/DataValue.cpp#L526. However, it might get swallowed by another try-catch. This could especially be the case during XML parsing since it is not so easy to pass through errors in the Xerces library. This would need to be investigated.
3) I am not sure if showing the file in the exception message is feasible but it definitely should show up in the logs with a message like:
"Loading/parsing file XYZ.ext..."
Such that when you encounter an exception later on, you will know which file caused it.

hroest:
maybe we should also write these to `std::cerr` to make sure its really written out and not swallowed somewhere?

timosachsenberg:
Fixed: item 1 in https://github.com/OpenMS/OpenMS/pull/5837

timosachsenberg:
Closing as resolved. Error reporting in DataValue was improved in two stages:

- PR #5837 (commit `086be4c000`, Feb 2022): Added type name context (`NamesOfDataType[]`) to all conversion error messages
- PR #8572 (commit `2fe4c47fd2`, Dec 2025): Added actual value context (`this->toString(true)`) to type-mismatch and negative-to-unsigned errors, plus 77 lines of `TEST_EXCEPTION_WITH_MESSAGE` tests

All conversion errors now include the DataValue's type name and (where applicable) the actual value, making debugging straightforward.

Identified during automated stale issue triage.

PR #5837, opened 2022-02-21
more error information in DataValue exception

timosachsenberg:
# Description

Add slightly additional output as requested in https://github.com/OpenMS/OpenMS/issues/4833
Ideally, we could also output the meta value name that caused an exception.

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/DataValue.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/DataValue.h
@@ -72,8 +72,12 @@ public:
       STRING_LIST, ///< string list
       INT_LIST, ///< integer list
       DOUBLE_LIST, ///< double list
-      EMPTY_VALUE ///< empty value
+      EMPTY_VALUE, ///< empty 
+      SIZE_OF_DATATYPE
     };
+
+    /// Names of activation methods
+    static const std::string NamesOfDataType[SIZE_OF_DATATYPE];
 
     /// Supported types for DataValue
     enum UnitType : unsigned char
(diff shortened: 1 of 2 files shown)


jpfeuffer:
What do you think about adding overloads like

int getMetaValue()
double getMetaValue()

to catch conversion errors together with the meta values names inside the MetaValueInterface class?
Otherwise, successful conversion always needs to be checked and handled by the developer.

timosachsenberg:
I don't know if I understand correctly. What would be the difference to the e.g. (int)getMetaValue() which uses the conversion operator?

jpfeuffer:
It is about catching the Exception and rethrowing with a clear error message including the name that was tried to be converted

timosachsenberg:
Hmm not sure. Right now, it basically returns (something similar to) a variant that can also store units etc. This could make such an interface a bit complex. Maybe it would be good to get rid of the unit at some point here first.

jpfeuffer:
I just mean that in those cases, where you know that you only need the internal value and you know that you want it to be a certain type, those overloaded functions could be used to provide a uniform way for error handling during those conversions.

timosachsenberg:
but you can't overload only by return type

jpfeuffer:
Hmm true, then we would need to change names.. I don't know, just an idea. It would also remove the need for toInt, toDouble, .. in pyopenms. Shifting the problem of course.

timosachsenberg:
rebuild jenkins

timosachsenberg:
will fix

hroest:
It seems this currently only fixes the problem to tell us which type was tried to convert to which other type (and failed) but it may be helpful to also output the *value* that was tried to convert, eg instead of "failed to convert string to int" it may be helpful to have "failed to convert string 'foo' to int"?

timosachsenberg:
Good point. I never planned to come up with a full solution to the issue ;) , but this could be easy to add and would help quite a bit.

timosachsenberg:
added via https://github.com/OpenMS/OpenMS/pull/5856

PR #5840, opened 2022-02-22
[FIX] scan number extraction

jpfeuffer:
# Description

In the IDFileConverter, when converting to pepXML with SpectrumLookup, we see a weird behaviour when the mzML
contains native IDs in the format "index=n".
Since spectrum metadata returns n as the scan number. Basically you end up with scan_number=n and index=n in the pepXML which is not correct.

Not sure if it is the right place to fix, but it should fix the problem.

### DIFF
--- a/src/openms/source/METADATA/SpectrumLookup.cpp
+++ b/src/openms/source/METADATA/SpectrumLookup.cpp
@@ -352,7 +352,7 @@ namespace OpenMS
           String value = String(matches[0]);
           if (native_id_type_accession == "MS:1000774")
           {
-            return ++value.toInt(); // if the native ID is index=.., the scan number is usually considered index+1 (especially for pepXML)
+            return ++(value.toInt()); // if the native ID is index=.., the scan number is usually considered index+1 (especially for pepXML)
           }
           else
           {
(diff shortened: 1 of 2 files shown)


src/openms/source/METADATA/SpectrumLookup.cpp
```diff
@@ -350,7 +350,14 @@ namespace OpenMS
         try
         {
           String value = String(matches[0]);
-          return value.toInt();
+          if (native_id_type_accession == "MS:1000774")
+          {
+            return ++(value.toInt()); // if the native ID is index=.., the scan number is usually considered index+1 (especially for pepXML)
```
timosachsenberg, line 355: ok then probably this needs update: https://github.com/OpenMS/OpenMS/blob/c2555c3278de34a5f261df9a0915f42a8bfada3c/src/tests/class_tests/openms/source/SpectrumLookup_test.cpp#L177

timosachsenberg: I think like you said. index is usually scan - 1.
"Index is the spectrum number in the file, starting from 0." From the mzTab-M specification.

timosachsenberg:
how should we go forward here? I think your approach is fine. Some tools might still differ in the way they use it, so we could expect some hiccups

jpfeuffer:
the thing is that this method looks much more general. It tries to get out the scan number from many different formats.
It has so many heuristics
E.g. are we sure that we should say the scan number was 42 if it is file=42?

timosachsenberg:
> the thing is that this method looks much more general. It tries to get out the scan number from many different formats. It has so many heuristics E.g. are we sure that we should say the scan number was 42 if it is file=42?

I don't know. If it is the 42th spectrum extracted into a file then yes ;)

PR #5844, opened 2022-02-23
[REFACTOR] replace Map with std::map

timosachsenberg:
# Description
Steps necessary:
- [x] Changed interface to std::map for access
- [x] Remove .has from Map interface and use modern constructs like:
```
if (auto m_it = mymap.find(key); m_it != mymap.end())
{
 // do something with m_it
}

to replace code like:
if (mymap.has(key))
{
  // do something with mymap.at(key)
}
```
- [x] replace Map with std::map C++
- [x] update python bindings / conversion providers

# Checklist:

# Note:


timosachsenberg:
rebuild jenkins

timosachsenberg:
hi, can anyone hold hands for the last mile? Python bindings and conversion providers are the last parts missing

jpfeuffer:
Let's just try find-replace Map with libcpp_map (including its include). Then we will see if anything made an assumption on the specific Map functions.

jpfeuffer:
We don't need any conversion providers anymore. If at all, we should remove the one for Map.

src/openms/source/ANALYSIS/QUANTITATION/ItraqConstants.cpp
```diff
@@ -209,7 +209,7 @@ namespace OpenMS
       info.description = "";
       info.name = channel_names[itraq_type].getValue(i, 0);
       info.id = (Int)i;
-      if (!reporter_mass_exact.has(info.name))
+      if (!(reporter_mass_exact.find(info.name) != reporter_mass_exact.end()))
```
timosachsenberg, line 212: Please replace check for non-existence with the simpler:
if (reporter_mass_exact.find(info.name) == reporter_mass_exact.end())

hroest:
> Let's just try find-replace Map with libcpp_map (including its include). Then we will see if anything made an assumption on the specific Map functions.

yes that would work, the only thing we will have to keep in mind is the Map special conversion providers which we need to adopt

https://github.com/OpenMS/OpenMS/blob/58ac2727f251f14febb1e2ddc063472f4113ee43/src/pyOpenMS/converters/special_autowrap_conversionproviders.py#L620

hroest:
with b19fa09aa01cb4c0b2f2e7db1df688e7dd8b9f4b now pyOpenMS compiles and all tests run. However I have not tested TOPP tools and unit tests

hroest:
PS: as  6aeecd7096d5d8e14b9b05d27681ae6993d39416 shows, we need to go over this by hand with a fine comb and cant just automatically replace things, if we create a map temporarily than we cannot compare `end()` of two different temporary maps. This just shows that this can lead to some subtle bugs...

src/openms/source/ANALYSIS/DECHARGING/ILPDCWrapper.cpp
```diff
@@ -97,13 +98,13 @@ namespace OpenMS
             g2f.erase(group2);
           }
         }
-        else if (f2g.has(f1) && !f2g.has(f2)) // only f1 is part of a group
+        else if (f2g.find(f1) != f2g.end() && !(f2g.find(f2) != f2g.end())) // only f1 is part of a group
```
hroest, line 101:
```suggestion
        else if ((f2g.find(f1) != f2g.end()) && !(f2g.find(f2) != f2g.end())) // only f1 is part of a group
```

src/openms/source/ANALYSIS/DECHARGING/ILPDCWrapper.cpp
```diff
@@ -72,15 +73,15 @@ namespace OpenMS
...
+      std::map<Size, Size> f2g; // feature id to connected group
+      std::map<Size, std::set<Size> > g2pairs; // group id to all pairs involved
+      std::map<Size, std::set<Size> > g2f; // group id to all features involved
 
       for (Size i = 0; i < pairs.size(); ++i)
       {
         Size f1 = pairs[i].getElementIndex(0);
         Size f2 = pairs[i].getElementIndex(1);
-        if (f2g.has(f1) && f2g.has(f2)) // edge connects two distinct groups
+        if (f2g.find(f1) != f2g.end() && f2g.find(f2) != f2g.end()) // edge connects two distinct groups
```
hroest, line 84:
```suggestion
        if ((f2g.find(f1) != f2g.end()) && (f2g.find(f2) != f2g.end())) // edge connects two distinct groups
```

hroest:
@timosachsenberg how far away are we from C++20 and std::map::contains ? https://en.cppreference.com/w/cpp/container/map/contains

timosachsenberg:
C20 support is still incomplete in most compilers but such simpler methods are well supported. VS caught up a bit though. Major point speaking against an upgrade is support by conda afaik. @jpfeuffer right?

jpfeuffer:
Conda is at GCC 9 I think. If that supports it, I think it would be fine.
I think we should target around Visual Studio 2019. And maybe MacOS Big Sur or the one before. Not sure about their compatibilities right now.

src/openms/source/ANALYSIS/QUANTITATION/ItraqConstants.cpp
timosachsenberg: @matteopilz

timosachsenberg:
Some python errors remain: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/pyOMSDckr/2404/console
otherwise passes

matteopilz:
Judging from the tests, the last error seems to come from https://github.com/OpenMS/OpenMS/blob/develop/src/pyOpenMS/tests/unittests/test000.py.
However, I cannot find out whats wrong.

src/pyOpenMS/addons/ADD_TO_ALL.pyx
```diff
@@ -1,7 +1,6 @@
 
 from libc.stddef cimport ptrdiff_t
 from libc.stdint cimport *
```
hroest, line 3:
```suggestion
from libc.stdint cimport *
from libcpp.map cimport map as libcpp_map
```

hroest:
> Some python errors remain: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/pyOMSDckr/2404/console

seems like a failure to compile, therefore all tests fail

matteopilz:
> > Some python errors remain: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/pyOMSDckr/2404/console
> 
> seems like a failure to compile, therefore all tests fail

Only two tests are failing now: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/pyOMSDckr/2408/console
test_pyopenms_unittests & pyopenms_unittest_test000
However, I am unable to find the issue here.

timosachsenberg:
having the output uploaded would be of great help

jpfeuffer:
Yep. There is probably just the upload command in the cmake script missing. You can also upload from the command line.

timosachsenberg:
@matteopilz can you check if you can add the upload command?

matteopilz:
> @matteopilz can you check if you can add the upload command?

Where do I insert the command?

jpfeuffer:
We are not using any scripts for the pyopenms build so you need to either do everything from the command line (not sure how to get all the setting from the CTestConfig.cmake inside there but probably possible).
Or you need to write a Cmake script like we do for GitHub actions.

jpfeuffer:
Click on details, go back to the job overview, configure, search for the bash script.

matteopilz:
[pyopenms_unittest_test000_stacktrace.txt](https://github.com/OpenMS/OpenMS/files/8204948/pyopenms_unittest_test000_stacktrace.txt)
[pyopenms_unittest_test000_std_output.txt](https://github.com/OpenMS/OpenMS/files/8204949/pyopenms_unittest_test000_std_output.txt)

This is the decompressed output from: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/pyOMSDckr/2408/testReport/projectroot/pyOpenMS/pyopenms_unittest_test000_py/

jpfeuffer:
So something wrong with AAseq getMonoWeight.
Either in the function or with the construction/retrieval of seq

src/openms/thirdparty/evergreen/src/Utility/graph_to_dot.hpp
```diff
@@ -2,6 +2,8 @@
 #define _GRAPH_TO_DOT_HPP
 
 #include <fstream>
+#include <map>
```
jpfeuffer, line 5: What happened here?

matteopilz: What do you mean?

timosachsenberg: I think he means that the thidparty library should not have any changes regarding the map

matteopilz: I've changed it back now.

matteopilz:
unittest still fails

hroest:
I quickly tested this, its not a Python issue. In C++ this also fails:

```
    AASequence seq = AASequence::fromString("PEPTIDESEKUEM(Oxidation)CER");
    TEST_EQUAL(seq.toUniModString(), "PEPTIDESEKUEM(UniMod:35)CER")
    TEST_REAL_SIMILAR(seq.getMonoWeight(), 1952.7200317517998)
```

```
371: checking AASequence fromString(const String& s, bool permissive = true) ...                                                                                                          371:  -  line 294:  TEST_REAL_SIMILAR(seq.getMonoWeight(),1952.7200317517998): got 1958.7140766518, expected 1952.7200317518 (absolute: 5.99404490000006 [1e-05], relative: 1.00306958744855 [1.00001], message: "ratio of numbers is large"                                                                                                                                        371: : failed                                                                                                                                                                             371:                            
```

hroest:
its probably this bug: #5904 and has nothing to do with this PR. We can wait for this to resolve first and then merge?

src/openms/include/OpenMS/DATASTRUCTURES/sources.cmake
```diff
@@ -31,8 +31,7 @@ KDTree.h
 ListUtils.h
 ListUtilsIO.h
 LPWrapper.h
-Map.h
-MassExplainer.h
+        MassExplainer.h
```
hroest, line 34:
```suggestion
MassExplainer.h
```

src/openms/include/OpenMS/FILTERING/TRANSFORMERS/SpectraMerger.h
```diff
@@ -47,6 +47,7 @@
 #include <OpenMS/CONCEPT/LogStream.h>
 #include <OpenMS/CONCEPT/ProgressLogger.h>
 #include <vector>
+#include <vector>
```
hroest, line 50:
```suggestion
```

hroest (approved):
It looks all good, all tests pass!

PR #5848, opened 2022-02-23
[FEATURE] strip files in installers with cpack

jpfeuffer:


timosachsenberg:
do you have an idea how much difference this makes?

cbielow (approved):
lets try by merging :)

jpfeuffer:
Bildschirmfoto 2022-02-24 um 14 02 21

jpfeuffer:
worked well on linux. not so much on the rest. Cmake says it's platform independent. And usually, the numbers are very stable. So it seems to works but just seems to not do so much on non-linux.

cbielow:
but still worth doing, I would say

jpfeuffer:
sure I'll leave it, unless we find out that the installers do not work anymore

PR #5881, opened 2022-03-02
TOPPView: fix log scale display (caused by negative intensity in visible area)

### LINKED ISSUE #5868: TOPPView large memory consumption when clicking on log scaling in 1D view
Memory consumption went up to 90gb on a small dataset. Tested on linux via x forwarding.

cbielow:
# Description

fixes #5868 which was introduced in #5703

# Note:

### DIFF
--- a/src/openms_gui/source/VISUAL/AxisPainter.cpp
+++ b/src/openms_gui/source/VISUAL/AxisPainter.cpp
@@ -125,7 +125,7 @@ namespace OpenMS
     // Painting tick levels
     for (Size i = 0; i != grid.size(); i++)
     {
-      // Just draw text on big intervalls
+      // Just draw text on big intervals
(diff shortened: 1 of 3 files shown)


timosachsenberg:
k thanks. might be indirectly related to the new way of scaling the y-axis
image
Which seems to introduce negative intensities? Hard to see from the screenshot but 1. the y-axis is not at zero (which looks a bit odd imho) 2. some peaks seem to be slightly below the zero line

cbielow:
> 1. the y-axis is not at zero (which looks a bit odd imho) 2. some peaks seem to be slightly below the zero line

this is also fixed by this PR :)
The plotting relies on the minimum visible area being set to 0, in order to plot the lower starting point correctly.
This is weird in itself and should be fixed, but for now, this is solved.

PR #5883, opened 2022-03-02
Fix TheoreticalSpectrumGenerationDialog - closes #5787

### LINKED ISSUE #5787: "Generate theoretical spectrum" in TOPPView is broken
image

- Not all items have checkboxes
- the checkboxes are not what they are labeled, i.e. checking b/y does not create b/y ions
- integration into TOPPView is not ideal (checks for missing peptide sequence etc should happen before the dialog is closed etc, i.e. the `TheoreticalSpectrumGenerationDialog` should be more feature-rich itself and return a spectrum plus other metadata, if desired).
- maybe also keep an instance of the class as member of toppview so re-opening the dialog preserves the previous settings.

PS: working on it. Please stand by :)

Waschi97:
Fixes the issues mentioned in #5787.

Additionally:

- Removed "Isotope Cluster" from the list widget in the dialog and added a new interface for generating isotope peaks that represents the settings of `TheoreticalSpectrumGenerator` better.

- A-b-Ions, D-Ions and W-Ions and their corresponding intensities are only shown if "RNA" is the chosen sequence type. The same is done for immonium ions, neutral losses and the isotope model for "Peptide" sequences.

- When "RNA" is the sequence type the dialog now actually computes the spectrum using `NucleicAcidSpectrumGenerator`.

- The dialog is resized automatically depending on what widgets are shown. This disables manual resizing by the user though.

- Added a check for empty chunks in `MSSpectrum::sortByPositionPresorted()` that otherwise caused a seq fault.

Dialog looks like this for "Peptide":

TheoSpecGenDialog_Peptide

And like this for "RNA":

TheoSpecGenDialog_RNA


timosachsenberg:
Nice. I personally consider the Intensities rather useless. I think they were originally introduced to distinguish ions but a better way would now be to use peak annotations as done in the identification view. But maybe that doesn't need to be part of this PR.

timosachsenberg:
Another thing that would be great if one could use empirical formulas e.g. to create the isotope pattern of e.g. metabolites instead of peptides.

src/openms_gui/include/OpenMS/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.h
```diff
@@ -60,6 +60,47 @@ namespace OpenMS
 
 public:
 
+    // Note: If an additional checkbox is added all of the following three objects have to be edited!
+    // 
+    // 
+    // enum to get ion checkbox index (Ordering has to be the same as in the ui!)
+    enum class Checkbox
```
cbielow, line 67: I like this approach, but there is no need to expose it in the header.
This can all be moved into the cpp to reduce dependencies (vector etc)

src/openms_gui/source/VISUAL/APPLICATIONS/TOPPViewBase.cpp
```diff
@@ -1988,74 +1988,11 @@ namespace OpenMS
...
-      {
-        generator.getSpectrum(spectrum, aa_sequence, charge, charge);
-      }
-      catch (Exception::BaseException& e)
-      {
-        QMessageBox::warning(this, "Error", QString("Spectrum generation failed! (") + e.what() + "). Please report this to the developers (specify what input you used)!");
-        return;
-      }
+      // spectrum is generated in the dialog, so just receive it here
+      PeakSpectrum spectrum = spec_gen_dialog_.getSpectrum();
```
cbielow, line 1995: very nice!

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -46,19 +50,28 @@ namespace OpenMS
   {
     ui_->setupUi(this);
 
-    connect(ui_->list_widget, SIGNAL(itemChanged(QListWidgetItem *)), this, SLOT(itemChanged(QListWidgetItem *)));
+    // if dialog is accepted, try generating a spectrum, only close dialog on success
+    connect(ui_->button_box, SIGNAL(accepted()), this, SLOT(calculateSpectrum()));
```
cbielow, line 54: please use the new syntax to connect
https://wiki.qt.io/New_Signal_Slot_Syntax

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -46,19 +50,28 @@ namespace OpenMS
...
+
+    // signals for changing isotope model interface
+    // there has to be a smarter way of doing this
+    connect(ui_->model_none, SIGNAL(toggled(bool)), this, SLOT(modelChanged()));
```
cbielow, line 58: i would not be aware of any :)

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -101,66 +131,86 @@ namespace OpenMS
...
-    p.setValue("has_Z", "false");
-    p.setValue("has_Precursor", "false");
-    p.setValue("has_abundantImmoniumIons", "false");
-
-    if (ui_->list_widget->item(0)->checkState() == Qt::Checked) // "A-ions"
+
+    return p;
+  }
+
+  MSSpectrum TheoreticalSpectrumGenerationDialog::getSpectrum() const
```
cbielow, line 138: return by const ref?

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -101,66 +131,86 @@ namespace OpenMS
...
+  {
+    return spec_;
+  }
+
+  void TheoreticalSpectrumGenerationDialog::calculateSpectrum()
+  {
+    String seq_string(this->getSequence());
+    if (seq_string == "")
```
cbielow, line 146: use !empty

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -101,66 +131,86 @@ namespace OpenMS
...
     {
-      p.setValue("has_A", "true");
+      QMessageBox::warning(this, "Error", "You must enter a peptide sequence!");
```
cbielow, line 148: could also be a dna seq?

cbielow:
> A-b-Ions, D-Ions and W-Ions

is this related to DNA seqs maybe?

timosachsenberg:
yes I think so https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp

timosachsenberg:
Thanks Tom!

timosachsenberg:
cool how does this look like - and how easy would it be to add e.g. a generator for empirical formulas? e.g., passing C6H12O6, and it creates e.g. the isotope pattern

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -36,29 +36,94 @@
...
+                                               Checkbox::X_Ions, Checkbox::Y_Ions,   Checkbox::Z_Ions, Checkbox::Precursor, Checkbox::Neutral_losses, Checkbox::Abundant_Immonium_Ions};
+
+  // map from checkbox (index) to corresponding parameter with description
+  const std::map<Checkbox, std::pair<String, String>> checkbox_to_param {
+    {Checkbox::A_Ions, {"add_a_ions", "Add peaks of a-ions to the spectrum"}},
+    {Checkbox::A_b_Ions, {"add_a-B_ions", "Add peaks of a-B-ions to the spectrum (nucleotide sequences only)"}},
+    {Checkbox::B_Ions, {"add_b_ions", "Add peaks of b-ions to the spectrum"}},
+    {Checkbox::C_Ions, {"add_c_ions", "Add peaks of c-ions to the spectrum"}},
+    {Checkbox::D_Ions, {"add_d_ions", "Add peaks of d-ions to the spectrum"}},
+    {Checkbox::W_Ions, {"add_w_ions", "Add peaks of w-ions to the spectrum"}},
```
jpfeuffer, line 81: Do they work for AAs?

Waschi97: not yet, `TheoreticalSpectrumGenerator` doesn't support them

jpfeuffer: just because it does not say (NA only) like the a-B ion

Waschi97: ahh, yes I can add that to the description - I just copied the parameter names and descriptions from the `NucleicAcidSpectrumGenerator` constructor

jpfeuffer: Should we maybe hide or delete them when not applicable? Especially if we do Metabolites too.

Waschi97: They are disabled and enabled depending on sequence type.
See line 111.

timosachsenberg:
Ok some ideas to make this more compatible with metabolomics.
- can we remove "None" from the Isotope Model? Then one could just switch between coarse and fine. Setting max isotopes to zero would disable it.
- would it be possible to remove instead of greying out the ion types? They don't exist at all for metabolites and also there is no good 1 to 1 correspondence between peptides and nucleic acid sequences.
- same for the intensities. Don't exist for metabolites (and are not useful for peptides / NA in my view anymore)
- should max isotope probability be "Min." ? Sounds like a threshold for the fine isotope structure generator
- just use the Coarse/Fine IsotopePatternGenerator to generate the spectra for metabolites

timosachsenberg:
but maybe we can first merge this PR. Seems rather feature complete to me.

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
jpfeuffer: Yes but disabling is cluttered. Can we hide?

Waschi97: Ahh that's what you mean. Yes, that is possible. I even tried that already but decided against it because I found it distracting when the dialog completely changed its appearance. Disabling just looked smoother for me.

timosachsenberg: if we add glycan or cross-links a fixed list could get out of hands :P
If you have the code still around for removing, I would actually also prefer that. And I guess it is not that distracting because a user will typically stick with one type (e.g., only switch once)

Waschi97:
"max isotope probability" should be fine, since the isotope peaks are calculated with `1.0 - p`

https://github.com/OpenMS/OpenMS/blob/58ac2727f251f14febb1e2ddc063472f4113ee43/src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/FineIsotopePatternGenerator.cpp#L46-L51

Waschi97:
For adding metabolomics to this dialog and the corresponding changes in the interface I would also prefer a second PR.

timosachsenberg:
ah ok so this is the total probability. got it. thanks

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
Waschi97: Alright, going to change to hiding instead of disabling.

src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp
```diff
@@ -66,10 +66,10 @@ namespace OpenMS
     defaults_.setValue("add_c_ions", "false", "Add peaks of c-ions to the spectrum");
     defaults_.setValidStrings("add_c_ions", {"true","false"});
 
-    defaults_.setValue("add_d_ions", "false", "Add peaks of d-ions to the spectrum");
+    defaults_.setValue("add_d_ions", "false", "Add peaks of d-ions to the spectrum (nucleotide sequences only)");
```
timosachsenberg, line 69: is this really needed here? I would assume that the options in this class are only relevant to NAs anyway

Waschi97: probably not honestly, just added it because the a-B ions already had that reminder

timosachsenberg: hmm it seems to me that this should probably just be a comment and not some description in the parameter class.
Would you mind just removing it while you are on it?

timosachsenberg (approved):

PR #5884, opened 2022-03-02
[Feature] Support for Ion Identity Molecular Networking by GNPS

axelwalter:
# Description
For IIMN by GNPS information about features that are related via different adducts needs to be exported by TextExporter.
During feature grouping the best ion annotation gets selected for the ConsensusFeature based on the best quality feature.
Also the Group IDs from a FeatureMap are annotated to later create a bipartite graph with ConsensusFeature indexes and Groups from Features using the boost graph library (described as comment in FeatureGroupingAlgorithm::annotateIonIdentityNetworks). For each ConsensusFeature the corresponding partners as well as a annotation network number get annotated and meta values and exported by TextExporter.

# Checklist:

# Note:


src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithmKD.cpp
```diff
@@ -492,12 +506,39 @@ namespace OpenMS
   void FeatureGroupingAlgorithmKD::addConsensusFeature_(const vector<Size>& indices, const KDTreeFeatureMaps& kd_data, ConsensusMap& out) const
   {
     ConsensusFeature cf;
+    Adduct adduct;
     float avg_quality = 0;
+    // determine best quality feature for adduct ion annotation (Constanst::UserParam::BEST_ION)
+    float best_quality = 0;
+    size_t best_quality_index = 0;
+    // collect the "Group" MetaValues of Features in a ConsensusFeature MetaValue (Constanst::UserParam::LINKED_GROUPS)
```
timosachsenberg, line 514:
```suggestion
    // collect the "Group" MetaValues of Features in a ConsensusFeature MetaValue (Constant::UserParam::LINKED_GROUPS)
```

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
```diff
@@ -43,6 +43,27 @@
...
+#include <map>
+
+#include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/connected_components.hpp>
+
+struct VertexLabel
+{
+    VertexLabel() = default;
+    VertexLabel(OpenMS::String u, bool is_feat):uid(u), is_feature(is_feat) {}
+    OpenMS::String uid = "0";
```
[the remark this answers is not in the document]
jpfeuffer: Also comment what it means when it is not a feature.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
```diff
@@ -170,6 +191,94 @@ namespace OpenMS
...
+        }
+      }
+    }
+
+    std::vector<int> components (boost::num_vertices (g));
+    boost::connected_components (g, &components[0]);
+
+    // annotate network number and create a map with feature ID and partner IDs
+    // partner feature vertexes are connected via a group vertex
+    map<size_t, set<size_t>> partner_map;
```
jpfeuffer, line 241: Cant you allocate a vector as big as the feature map? and then access by index with the uid?

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
```diff
@@ -170,6 +191,94 @@ namespace OpenMS
...
+            }
+        }
+        if (!already_set)
+        {
+        boost::add_edge(feature_vertex, add_vertex(VertexLabel(group, false), g), g);
+        }
+      }
+    }
+
+    std::vector<int> components (boost::num_vertices (g));
```
jpfeuffer, line 236: Comment what this vector represents.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
```diff
@@ -170,6 +191,94 @@ namespace OpenMS
...
+          if (g[i].uid == g[partner].uid) continue;
+          partner_map[stoi(g[i].uid)].insert(stoi(g[partner].uid));
+        }
+      }
+    }
+
+    // annotate partners
+    for (const auto& i : partner_map)
+    {
+      for (const auto& j : i.second)
```
jpfeuffer, line 261: Probably better to get all the IIMN_ROW_IDs first and set the metavalue only at the end of the inner loop.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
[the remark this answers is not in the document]
jpfeuffer: Maybe copy parts of the description in annotateIonIdentityNetworks

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
```diff
@@ -170,6 +191,94 @@ namespace OpenMS
...
+
+    for (size_t i = 0; i < out.size(); i++)
+    {
+      out[i].setMetaValue(Constants::UserParam::IIMN_ROW_ID, i+1);
+      if (!out[i].metaValueExists(Constants::UserParam::LINKED_GROUPS)) continue;
+      auto feature_vertex = add_vertex(VertexLabel(String(i), true), g);
+      for (const auto& group: out[i].getMetaValue(Constants::UserParam::LINKED_GROUPS).toStringList())
+      {
+        bool already_set = false;
+        for (auto i : boost::make_iterator_range(vertices(g)))
```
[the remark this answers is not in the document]
jpfeuffer: Yes, use and unordered_map from group index to vertex. Check if already inserted: if so, add edge to existing vertex, else add vertex, add vertex to unordered_map, add edge to new vertex.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
timosachsenberg: so do more or less all of them have partners or would we gain something from using the map (e.g., save memory because many don't have partners)

src/openms/include/OpenMS/CONCEPT/Constants.h
```diff
@@ -441,6 +441,42 @@ namespace OpenMS
...
+
+      /** User parameter name for a unique ConsensusFeature index in a ConsensusMap. (Required for IIMN)
+              String
+      */
+      inline const std::string   IIMN_ROW_ID = "row ID";
+
+      /** User parameter name for a ConsensusFeature to indicate a metabolite with different adduct states. (Required for IIMN)
+              String
+      */
+      inline const std::string   ANNOTATION_NETWORK_NUMBER = "annotation network number";
```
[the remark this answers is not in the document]
axelwalter: For IIMN they need to be lower case with spaces. The "Groups" meta value was already present and therefore the "LinkedGroups" should be upper case as well?!

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
axelwalter: Most of the features don't have partners, so this could be more efficient.

timosachsenberg: then I would keep it as is (or change to) unordered_map

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
```diff
@@ -170,6 +191,94 @@ namespace OpenMS
...
+    {
+      if (!g[i].is_feature) continue;
+      out[stoi(g[i].uid)].setMetaValue(Constants::UserParam::ANNOTATION_NETWORK_NUMBER, components[i]+1);
+      auto group_neighbours = boost::adjacent_vertices(i, g);
+      for (auto gn : make_iterator_range(group_neighbours))
+      {
+        auto feature_partners = boost::adjacent_vertices(gn, g);
+        for (auto partner : make_iterator_range(feature_partners))
+        {
+          if (g[i].uid == g[partner].uid) continue;
```
jpfeuffer, line 252: I think you could do "i == partner"?

axelwalter: yes or course!

axelwalter:
Thanks for the helpful comments! :)

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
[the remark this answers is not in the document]
axelwalter: It's a String because the original "Group" meta value from the FeatureMap is already a String. And because the VertexLabel needs to work for both the feature and group vertex, it is a String here.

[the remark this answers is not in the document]
axelwalter: Tried earlier to convert the String meta value to a size_t, but there were some issues. This should be possible without loss of information?

src/topp/TextExporter.cpp
```diff
@@ -213,6 +235,14 @@ namespace OpenMS
     {
       out << str + suffix;
     }
+    if (incl_IIMN & write_IIMN)
```
timosachsenberg, line 238: why are there two variables needed?
what does incl_IIMN mean here?
please don't use & but &&

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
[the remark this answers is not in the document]
jpfeuffer: @timosachsenberg should know. In theory yes, but not sure how complete the DataValue conversions are. I think I had problems before too.

src/topp/TextExporter.cpp
axelwalter: Right, the two variables are not needed.

axelwalter:
I will now add a IIMN class and move the FeatureGroupingAlgorithm::annotateIonIdentityMolecularNetworking method there. Also methods for the export of the FeatureQuantificationTable and SupplementaryPairsTable will be there and used in the GNPSExport tool. With this approach we don't need the TextExporter to include the IIMN meta values.

src/openms/source/ANALYSIS/MAPMATCHING/FeatureGroupingAlgorithm.cpp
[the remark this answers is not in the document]
jpfeuffer: I think the question is if Meta/DataValue can handle this.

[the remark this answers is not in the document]
timosachsenberg: ok if that is the question I agree. still would be possible to change the type in the vertex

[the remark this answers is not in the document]
jpfeuffer: @axelwalter This basically means you cannot use the automatic conversions from DataValue. You need to use stoi or something like this.
I prefer uints/size_t in the vertex

jpfeuffer:
rebuild jenkins

jpfeuffer:
`E:\jenkins\ws\openms\PR\bldtst\9447518b\source\src\openms\source\ANALYSIS\ID\IonIdentityMolecularNetworking.cpp(221,102): error C2668: 'std::basic_stringstream<char,std::char_traits<char>,std::allocator<char>>::basic_stringstream': ambiguous call to overloaded function [E:\jenkins\ws\openms\PR\bldtst\9447518b\build\src\openms\OpenMS.vcxproj]
`
same for macos

jpfeuffer:
And the test failed on linux
https://cdash.openms.de/test/7564735

axelwalter:
rebuild jenkins

src/openms/source/ANALYSIS/ID/IonIdentityMolecularNetworking.cpp
```diff
@@ -214,7 +215,7 @@ namespace OpenMS
     if (!consensus_map[0].metaValueExists(Constants::UserParam::IIMN_ROW_ID)) return;
 
     // generate unordered map with feature id and partner feature ids
-    std::unordered_map<size_t, std::set<size_t>> feature_partners; // map<feature_index, partner_feature_indices>
+    std::map<size_t, std::set<size_t>> feature_partners; // map<feature_index, partner_feature_indices>
```
timosachsenberg, line 218: this could change the algorithm speed a lot as it changes access from constant time to log time.
If it is not an issue I think it is fine.

axelwalter: How should we deal with the tests? They are not reproducible with the unordered map.

src/openms/include/OpenMS/CONCEPT/Constants.h
```diff
@@ -441,6 +441,42 @@ namespace OpenMS
               String
       */
       inline const std::string   XFDR_FDR = "XFDR:FDR";
+
+      /** User parameter name for best ion annotation in a ConsensusFeature, taken from the best quality feature. (Required for IIMN)
+              String
+      */
+      inline const std::string   IIMN_BEST_ION = "best ion";
```
cbielow, line 448: did anyone wonder why do we use strings?
stringview would be sufficient?

src/openms/include/OpenMS/ANALYSIS/ID/IonIdentityMolecularNetworking.h
```diff
@@ -0,0 +1,62 @@
...
+{
+  class OPENMS_DLLAPI IonIdentityMolecularNetworking
+  {
+    public:
+      /// Annotate ConsensusMap for ion identity molecular networking (IIMN) workflow by GNPS.
+      /// Adds meta values Constants::UserParams::IIMN_ROW_ID (unique index for each feature), Constants::UserParams::IIMN_ADDUCT_PARTNERS (related features row IDs)
+      /// and Constants::UserParams::IIMN_ANNOTATION_NETWORK_NUMBER (all related features with different adduct states) get the same network number).
+      /// This method requires the features annotated with the Constants::UserParams::IIMN_LINKED_GROUPS meta value.
+      //  If at least one of the features has an annotation for Constants::UserParam::IIMN_LINKED_GROUPS, annotate ConsensusMap for IIMN.
+      static void annotateConsensusMap(ConsensusMap& out);
```
cbielow, line 49: the parameter is input+output at the same time right? maybe name it just `map`. The function name does the rest...

src/openms/include/OpenMS/ANALYSIS/ID/IonIdentityMolecularNetworking.h
```diff
@@ -0,0 +1,62 @@
...
+namespace OpenMS
+{
+  class OPENMS_DLLAPI IonIdentityMolecularNetworking
+  {
+    public:
+      /// Annotate ConsensusMap for ion identity molecular networking (IIMN) workflow by GNPS.
+      /// Adds meta values Constants::UserParams::IIMN_ROW_ID (unique index for each feature), Constants::UserParams::IIMN_ADDUCT_PARTNERS (related features row IDs)
+      /// and Constants::UserParams::IIMN_ANNOTATION_NETWORK_NUMBER (all related features with different adduct states) get the same network number).
+      /// This method requires the features annotated with the Constants::UserParams::IIMN_LINKED_GROUPS meta value.
+      //  If at least one of the features has an annotation for Constants::UserParam::IIMN_LINKED_GROUPS, annotate ConsensusMap for IIMN.
```
cbielow, line 48: missing `/`?

src/openms/source/ANALYSIS/ID/IonIdentityMolecularNetworking.cpp
timosachsenberg: if memory is not a big deal you could just do a
map<...> sorted(feature_partners.begin(), feature_partners.end());
and then use this map instead of for (const auto& entry: feature_partners)
before writing out results.
If memory is an issue you could probably swap out the values so these don't get copied:
for (auto& [key, value] : feature_partners)
{
  sorted[key].swap(value);
}

src/topp/GNPSExport.cpp
```diff
@@ -112,9 +115,15 @@ class TOPPGNPSExport : public TOPPBase
     registerInputFileList_("in_mzml", "<files>", ListUtils::create<String>(""), "Original mzml files containing the ms2 spectra (aka peptide annotation). \nMust be in order that the consensusXML file maps the original mzML files.");
     setValidFormats_("in_mzml", ListUtils::create<String>("mzML"));
 
-    registerOutputFile_("out", "<file>", "", "Output MGF file");
+    registerOutputFile_("out", "<file>", "", "Output MGF file.");
     setValidFormats_("out", ListUtils::create<String>("mgf"));
 
+    registerOutputFile_("out_quantification", "<file>", "", "Output feature quantification table.");
+    setValidFormats_("out_quantification", ListUtils::create<String>("txt"));
```
cbielow, line 122: use `{"txt"}` for bevity and speed :)

src/openms/include/OpenMS/CONCEPT/Constants.h
cbielow: We may need a stringview overload to look up metavalues .... maybe thats why :P

src/openms/source/ANALYSIS/ID/IonIdentityMolecularNetworking.cpp
cbielow: I remember Julianus(?) adding some code to make the ordering consistent across platforms...
Cant find it ad hoc...

jpfeuffer: This was just for Random numbers I think.

src/openms/include/OpenMS/CONCEPT/Constants.h
timosachsenberg: How does this make a difference? There would still be string comparisons upon lookup?

cbielow: true, but no construction cost (at least the rhs would be constexpr automatically for SV's) and no heap allocation for the string if SSO does not kick in (better locality).

src/openms/source/ANALYSIS/ID/IonIdentityMolecularNetworking.cpp
cbielow: True.

does the result need to be sorted for the user, or merely the test?
If its just the test, then why not capture the output (as string?) and sort that instead of adding a conversion penalty in the production code. ??
Or get fancy and use a templated function, where you inject the type of container as template parameter.

axelwalter: For the user it doesn't matter if the result is sorted, so adjusting the tests might be an option.

axelwalter: Just trying the solution suggested by Timo as well to see if that works.

src/topp/GNPSExport.cpp
axelwalter: What do you mean exactly? Does this apply to the other formats as well?

src/openms/source/ANALYSIS/ID/IonIdentityMolecularNetworking.cpp
axelwalter: Thanks for the suggestions. Have created a sorted map while swapping the values to save memory. This is a good solution since even if the sorting is not strongly required for the user, it gives a better readable output.

src/topp/GNPSExport.cpp
cbielow: I meant:
`setValidFormats_("out_quantification", {"txt"});`

> Does this apply to the other formats as well?
Yes. If you find some others nearby, feel free to fix :)

timosachsenberg (approved):

timosachsenberg:
it failed in the nightlies because this was missing https://github.com/OpenMS/OpenMS/commit/c91d38cb1a5b9e3743ff2653cdc9bbd0972e11e3
Apparently needed for creating the install target

PR #5890, opened 2022-03-04
Feature/isotopes

### LINKED ISSUE #5851: ElementDB cannot be modified
We currently cannot modify ElementDB, so it is not possible to actually represents certain compounds, such as those with C14 as specified in our documentation: https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/release/latest/html/classOpenMS_1_1EmpiricalFormula.html#details

If one wants only use a specific isotope, it can be specified using "(",")" brackets. For example, to specify 14C a heavy isotope of carbon it is expressed as "(14)C". The isotope distribution of that instance contains only one isotope, 14C itself with a frequency of 100%.

I am not sure if we need to represent these (and other unstable) isotopes, but it may be useful if a user could at least add their own elements to ElementDB.

hroest:
# Description

Fixes  #5851 -- allows modification of ElementDB and drops the `const`

also adds the pyOpenMS bindings

However, there are some conceptual problems that remain with our approach, it seems what we call `Element` is actually used to represent *both* an element as well as an individual isotope. We may want to think about representing half lives for individual isotopes and separate the concepts of isotopes and elements (maybe with an extra class for isotopes?)

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
@@ -39,6 +39,7 @@
 #include <OpenMS/DATASTRUCTURES/Map.h>
 #include <OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsotopeDistribution.h>
 #include <OpenMS/CHEMISTRY/Element.h>
+#include <OpenMS/CONCEPT/Exception.h>
 
 #include <map>
 #include <string>
@@ -48,7 +49,7 @@ namespace OpenMS
 
   /** @ingroup Chemistry
 
-          @brief Singleton that stores elements.
+      @brief Singleton that stores elements and isotopes.
 
       The elements weights (in the default file) are taken from
       "Isotopic Compositions of the Elements 1997", Pure Appl. Chem., 70(1), 217-235, 1998.
@@ -59,11 +60,11 @@ namespace OpenMS
           Pure Appl. Chem., 2003, Vol. 75, No. 6, pp. 683-799
           doi:10.1351/pac200375060683
 
-          Specific isotopes of elements can be accessed by writing the atomic number of the isotope
-          in brackets followed by the element name, e.g. "(2)H" for deuterium.
+      Specific isotopes of elements can be accessed by writing the atomic number of the isotope
+      in brackets followed by the element name, e.g. "(2)H" for deuterium.
 
-    @improvement include exact mass values for the isotopes (done) and update IsotopeDistribution (Andreas)
-          @improvement add exact isotope distribution based on exact isotope values (Andreas)
+      @improvement include exact mass values for the isotopes (done) and update IsotopeDistribution (Andreas)
+      @improvement add exact isotope distribution based on exact isotope values (Andreas)
 */
 
   class OPENMS_DLLAPI ElementDB
@@ -75,7 +76,7 @@ public:
     //@{
     /// returns a pointer to the singleton instance of the element db
     /// This is thread safe upon first and subsequent calls.
-    static const ElementDB* getInstance();
+    static ElementDB* getInstance();
 
     /// returns a hashmap that contains names mapped to pointers to the elements
     const std::map<std::string, const Element*>& getNames() const;
@@ -95,6 +96,30 @@ public:
     /// returns a pointer to the element of atomic number; if no element is found 0 is returned
     const Element* getElement(unsigned int atomic_number) const;
 
+    /** Adds or replaces a new element to the database
+     *
+     * Adds a new element (or replaces an existing one if @em replace_existing is true). 
+     *
+     * @param name Common name of the element
+     * @param symbol Element symbol (one or two letter)
+     * @param an Atomic number (number of protons)
+     * @param abundance List of abundances for each isotope (e.g. {{12u, 0.9893}, {13u, 0.0107}} for Carbon)
+     * @param abundance List of masses for each isotope (e.g. {{12u, 12.0}, {13u, 13.003355}} for Carbon)
+    */
+    void addElement(const std::string& name,
+                    const std::string& symbol,
+                    const unsigned int an,
+                    const std::map<unsigned int, double>& abundance,
+                    const std::map<unsigned int, double>& mass,
+                    bool replace_existing)
+    {
+      if (hasElement(an) && !replace_existing)
+      {
+        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, String("Element with atomic number ") + an + " already exists");
+      }
+      buildElement_(name, symbol, an, abundance, mass);
+    }
+
     //@}
 
     /** @name Predicates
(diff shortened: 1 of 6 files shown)


timosachsenberg:
Can’t we find a proper solution. This will likely be misused to change the abundances (e.g. for labeling) leading to tons of problems in multi threaded context etc

timosachsenberg:
@cbielow what is your take on this?

hroest:
@timosachsenberg all other DBs are also not `const` and suffer from similar problems, so I dont see why this should be different (actually ElementDB is the only const one out of all those):

```
OpenMS/CONCEPT/UniqueIdGenerator.h:    static UniqueIdGenerator& getInstance_();
OpenMS/CONCEPT/LogConfigHandler.h:    static LogConfigHandler & getInstance();
OpenMS/CONCEPT/GlobalExceptionHandler.h:      static GlobalExceptionHandler & getInstance()
OpenMS/TRANSFORMATIONS/FEATUREFINDER/IsotopeWavelet.h:    static IsotopeWavelet * getInstance()
OpenMS/CHEMISTRY/RibonucleotideDB.h:    static RibonucleotideDB* getInstance();
OpenMS/CHEMISTRY/CrossLinksDB.h:    inline static CrossLinksDB* getInstance()
OpenMS/CHEMISTRY/ResidueDB.h:    static ResidueDB* getInstance();
OpenMS/CHEMISTRY/ModificationsDB.h:    static ModificationsDB* getInstance();
OpenMS/CHEMISTRY/DigestionEnzymeDB.h:    static InstanceType* getInstance()
OpenMS/CHEMISTRY/ElementDB.h:    static const ElementDB* getInstance();
```

so the answer is more likely: simply dont modify a singleton in a multithreaded context! 

The functionality is more likely to be used *before* the actual algorithm in a single threaded environment in order to prepare the database. How likely is it that someone will attempt to change the elements *during* the execution of an algorithm, that seems unlikely and unwise to me. So I see few benefits to keep it const and major drawbacks in terms of flexibility.

timosachsenberg:
Just saying it is a well known anti pattern to have non-const singletons - and that for good reason imho. By the way I am guilty of changing the relative abundances in MetaProSip as well - of which I am not proud of by const casting ;) so I know that we probably don’t have a good (quickly to implement) alternative right now to what you propose. It just bugs me :)

cbielow:
IMHO using non-const should be fine - it also allows users to change isotope abundances, although not in an ideal way.
We could go fancy and use two maps in the back, one for default elements, one for custom elements. But this would prevent a user from changing isotope abundances for common elements.

Why was parsing Elements.xml abandoned in the first place? Too slow? Did we measure? If yes, there should be a way to use a simpler format, which is fast to parse. I cannot see this taking longer than 0.1 seconds...
Using a file input at startup has two advantages: 
1) users can specify whatever they want
2) once loaded, it can be const and is inherently threadsafe

timosachsenberg:
Being able to deploy OpenMS.dll without a ton of files has been beneficial for interfacing with instrument software.
```
users can specify whatever they want
once loaded, it can be const and is inherently threadsafe
```
This is only one use case, but a more common is changing the isotope abundance (once or several times in an algorithm) for metabolic labeling analysis, stable isotope probing etc. Here the file solution does not help.

cbielow:
> Being able to deploy OpenMS.dll without a ton of files has been beneficial for interfacing with instrument software.

Interesting point. But we still need files from /share in order to run OpenMS (at least if you need a reasonable set of functionality), e.g. unimod,xml and PSI-mod.obo. So how did removing Elements.xml help?

> 
> ```
> users can specify whatever they want
> once loaded, it can be const and is inherently threadsafe
> ```
> 
> This is only one use case, but a more common is changing the isotope abundance (once or several times in an algorithm) for metabolic labeling analysis, stable isotope probing etc. Here the file solution does not help.

This just means that we do not have good interfaces to allow isotope distributions to be computed using custom distributions.
Can you link actual code so we discuss alternatives?

timosachsenberg:
> > Being able to deploy OpenMS.dll without a ton of files has been beneficial for interfacing with instrument software.
> 
> Interesting point. But we still need files from /share in order to run OpenMS (at least if you need a reasonable set of functionality), e.g. unimod,xml and PSI-mod.obo. So how did removing Elements.xml help?

This is right, but we already got rid of several files (Residues and Elements IIRC). And this helped for lower level algorithms.

Unimod and PSI-mod -> here I would prefer having the current version compiled in as well, ideally with an option to provide custom ones on top.

> > This is only one use case, but a more common is changing the isotope abundance (once or several times in an algorithm) for metabolic labeling analysis, stable isotope probing etc. Here the file solution does not help.
> 
> This just means that we do not have good interfaces to allow isotope distributions to be computed using custom distributions. Can you link actual code so we discuss alternatives?

We could add it to
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/CHEMISTRY/ISOTOPEDISTRIBUTION/IsotopePatternGenerator.h
or one of it's derived classes for Fine and Coarse

hroest:
> This just means that we do not have good interfaces to allow isotope distributions to be computed using custom distributions.
Can you link actual code so we discuss alternatives?

since we currently use a global singleton ElementDB from which these are parsed. The proper solution would be to be able to pass a custom ElementDB object with the appropriate settings for each algorithm, right?

hroest:
I agree, this could be addressed by re-computing weights through ResidueDB and (maybe) ModificationsDB though. This could be built in relatively easy through the databases but you need to remember to update them!

timosachsenberg:
Regarding your question: I also think that we should separate the Element vs isotope concept.

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -574,6 +570,11 @@ namespace OpenMS
 
   void ElementDB::addElementToMaps_(const string& name, const string& symbol, const unsigned int an, const Element* e)
   {
+    // delete existing element before adding a new one
+    if (atomic_numbers_.find(an) != atomic_numbers_.end())
+    {
+      delete atomic_numbers_[an];
```
hroest, line 576: there is a major issue here: if we delete the Element ptr, then any structure that uses that ptr has a problem, eg any `EmpiricalFormula` that already exists will run into a segfault...

timosachsenberg: ok so we could at least in principle modify it here (which will not update anything that was calculated in a lazy way though) - and did I mention that I hate those singletons ;)

hroest: yes, I think we could try 

```
*atomic_numbers_[an] = *e;
```

hroest:
I have now added support for in-place modification of Element (its not pretty, with a const_cast) and tests for that case. I also made some performance improvements when parsing `EmpiricalFormula` which is now faster:

```
  // 2.36 seconds for 1M formulae
  // after optimization -> 1.39 seconds for 1M formulae
  double res = 0; 
  for (int i = 0; i < 1e6; i++) 
  {
    EmpiricalFormula f("C3071H4826N816O927S40"); // BSA
    res += f.getMonoWeight();
  }
```
its not 42% faster than before, creating 1e6 complex `EmpiricalFormula` used as the benchmark.

hroest:
with the latest changes, the benchmark is now down to 0.78 seconds (783 nsec per iteration), which means an overall speed increase of 3x compared to the original.

783 nsec usec per iteration for BSA; 376 nsec for Glucose; 497 for PEPTIDE

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
```diff
@@ -35,20 +35,19 @@
 
 #pragma once
```
hroest, line 37:
```suggestion
#include <OpenMS/DATASTRUCTURES/Map.h>
```

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -574,17 +591,34 @@ namespace OpenMS
 
   void ElementDB::addElementToMaps_(const string& name, const string& symbol, const unsigned int an, const Element* e)
   {
-    names_[name] = e;
-    symbols_[symbol] = e;
-    atomic_numbers_[an] = e;
+    #pragma omp critical(OpenMS_ElementDB)
```
cbielow, line 594: why this critical region?
this is the only place that reading/writing is protected, so it does not really help a lot.
Adding elements should not be done in a threaded context anyway IMHO.

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -574,17 +591,34 @@ namespace OpenMS
...
+    {
+      // overwrite existing element if it already exists
+      // find() has to be protected here in a parallel context
+      if (atomic_numbers_.find(an) != atomic_numbers_.end())
+      {
+        // in order to ensure that existing elements are still valid and memory
+        // addresses do not change, we have to modify the Element in place
+        // instead of replacing it.
+        const Element* const_ele = atomic_numbers_[an];
```
cbielow, line 603: this is a duplicated lookup.
The result of the find() from above could be cached to avoid extra work.

src/openms/source/CHEMISTRY/ElementDB.cpp
hroest: @cbielow I agree, it does not prevent concurrent reading but it does help against multiple people writing at the same time which should be helpful

src/openms/source/CHEMISTRY/ElementDB.cpp
hroest: adding elements should never be performance critical, if it is then something is clearly wrong

hroest:
@timosachsenberg @cbielow any other concerns? I think we should be ready to merge

PR #5891, opened 2022-03-04
[FEATURE] fix monoisotopic definition

hroest:
fixes #5889

PR #5904, opened 2022-03-08
[FIX] test fix after #5891

hroest:
- after the change in monoisotopic mass, we have to adjust the test for
monoisotopic mass for selenocysteine from 74 to 80 (most abundant
isotope)

### DIFF
--- a/src/pyOpenMS/tests/unittests/test000.py
+++ b/src/pyOpenMS/tests/unittests/test000.py
@@ -236,7 +236,7 @@ def testAASequence():
         print("Error: Exception not triggered.")
         assert False
     assert seq.getFormula(pyopenms.Residue.ResidueType.Full, 0) == pyopenms.EmpiricalFormula("C75H122N20O32S2Se1")
-    assert abs(seq.getMonoWeight(pyopenms.Residue.ResidueType.Full, 0) - 1952.7200317517998) < 1e-5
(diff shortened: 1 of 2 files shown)


timosachsenberg (approved):

hroest:
seems like windows doesnt agree...

```

checking AASequence fromString(const String& s, bool permissive = true) ... 
 -  line 295:  TEST_EQUAL(EmpiricalFormula("C75H122N20O32S2").getMonoWeight(),1878.7975553518002): got '1878.8', expected '1878.8'
: failed
```

@jpfeuffer @timosachsenberg whats up with the output here? `got '1878.8', expected '1878.8'` is not really great?

timosachsenberg:
Agree but TEST_REAL_SIMILAR is the way to go here.

timosachsenberg:
I will add those changes

jpfeuffer:
Is there a default precision in our tests? Maybe we need to increase it then.

timosachsenberg:
> Is there a default precision in our tests? Maybe we need to increase it then.

we print with full precision in:
https://github.com/OpenMS/OpenMS/blob/08ed8a1c256c8f24a6008b85fb507aef06616f17/src/openms/include/OpenMS/CONCEPT/ClassTest.h#L297-L327

hroest:
> Agree but TEST_REAL_SIMILAR is the way to go here.

of course, my bad!

PR #5911, opened 2022-03-10
Support for overlapping IM and m/z windows (for diaPASEF data)

jcharkow:
# Description
This PR adds the -pasef flag to OpenSwathWorkflow which adds support for processing .mzML files with overlapping IM and m/z windows. The precursor will only be extracted from a single window that matches both in m/z and whose ion mobility (drift time) center is closest to the library ion mobility of the precursor.

# Checklist:
- [X ] Make sure that you are listed in the AUTHORS file
- [X ] Add relevant changes and new features to the CHANGELOG file
- [X ] I have commented my code, particularly in hard-to-understand areas

# Note:


hroest (changes requested):
I reviewed the changes together with Josh and he added a test. Currently we are trying to figure out why it fails on clang

jpfeuffer:
Also using libc++? Or clang with libstdc++?
I bet it is just some sorting or randomness issue.

src/openms/include/OpenMS/APPLICATIONS/OpenSwathBase.h
```diff
@@ -187,6 +188,8 @@ class TOPPOpenSwathBase :
       OPENMS_LOG_DEBUG << "Found swath map " << i
         << " with lower " << swath_maps[i].lower
         << " and upper " << swath_maps[i].upper
+        << " and im Lower bounds of " << swath_maps[i].imLower
+        << " and imUpper bounds of " << swath_maps[i].imUpper
```
timosachsenberg, line 192:
```suggestion
        << " and im Upper bounds of " << swath_maps[i].imUpper
```

src/openms/include/OpenMS/FORMAT/DATAACCESS/SwathFileConsumer.h
```diff
@@ -240,7 +256,8 @@ namespace OpenMS
         {
           // We group by the precursor mz (center of the window) since this
           // should be present in all SWATH scans.
-          if (std::fabs(center - swath_map_boundaries_[i].center) < 1e-6)
+          // also specify ion mobility, if ion mobility not present will just be -1
+          if ((std::fabs(center - swath_map_boundaries_[i].center) < 1e-6) && (std::fabs(lowerIm - swath_map_boundaries_[i].imLower) < 1e-6) && (std::fabs(upperIm - swath_map_boundaries_[i].imUpper < 1e-6)))
```
timosachsenberg, line 260: the bracket is wrong in the last expression. should close before the comparison

timosachsenberg: can you add a test for this path?

src/openms/include/OpenMS/FORMAT/DATAACCESS/SwathFileConsumer.h
```diff
@@ -266,11 +283,13 @@ namespace OpenMS
...
             boundary.upper = upper;
             boundary.center = center;
+            boundary.imLower = lowerIm;
+            boundary.imUpper = upperIm;
             swath_map_boundaries_.push_back(boundary);
 
             OPENMS_LOG_DEBUG << "Adding Swath centered at " << center
               << " m/z with an isolation window of " << lower << " to " << upper
-              << " m/z." << std::endl;
+              << " m/z." << " and IM lower limit of " <<  lowerIm << " and upper limit of " << upperIm << std::endl;
```
timosachsenberg, line 292:
```suggestion
              << " m/z and IM lower limit of " <<  lowerIm << " and upper limit of " << upperIm << std::endl;
```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -523,21 +526,59 @@ namespace OpenMS
...
 
           }
         }
       }
     }
 
-    // (iii) Perform extraction and scoring of fragment ion chromatograms (MS2)
+    if (pasef_)
+    {
+      // For PASEF experiments it is possible that have DIA windows with the same m/z however different IM.
```
timosachsenberg, line 545:
```suggestion
      // For PASEF experiments it is possible to have DIA windows with the same m/z however different IM.
```

tools/scripts/create_im_with_swathIm.py
```diff
@@ -0,0 +1,256 @@
...
+    sframe.setFloatDataArrays([fda])
+
+
+    sframe.sortByPosition()
+
+    exp.addSpectrum(sframe)
+
+
+
+############# EXPOERT MZML FILE ##########################
```
timosachsenberg, line 249:
```suggestion
############# EXPORT MZML FILE ##########################
```

tools/scripts/create_im_with_swathIm.py
```diff
@@ -0,0 +1,256 @@
...
+    createMS2Precursor(allmz, allint, allim, imCenterIdx=550, imWidth=10, base_int=base_int, imWinLim=(350, 600))
+
+    fda = pyopenms.FloatDataArray()
+    fda.setName("Ion Mobility")
+    fda.resize(len(allmz))
+
+    for k,val in enumerate(allim):
+        fda[k] = val
+
+
```
timosachsenberg, line 218: reduce duplicate new lines

src/tests/class_tests/openms/source/SwathFileConsumer_test.cpp
```diff
@@ -54,29 +54,62 @@ static bool SortSwathMapByLower(const OpenSwath::SwathMap left, const OpenSwath:
...
+        prec[0].setMZ(400 + i*25 + 12.5);
+        s.setPrecursors(prec);
+        Peak1D p; p.setMZ(101 + i); p.setIntensity(201 + i);
+        s.push_back(p);
+        s.setMetaValue("ion mobility lower limit", 0.6 + i *0.05 + scheme*0.1);
+        s.setMetaValue("ion mobility upper limit", 0.6 + i *0.05 + scheme*0.1);
+        exp.addSpectrum(s);
+      }
+    }
+  } else
```
timosachsenberg, line 96:
```suggestion
  } 
  else 
```

src/openms/source/FORMAT/SwathFile.cpp
```diff
@@ -367,15 +382,21 @@ namespace OpenMS
...
+            // set IM boundaries (if present)
+            boundary.imLower = lowerIm;
+            boundary.imUpper = upperIm;
+
             known_window_boundaries.push_back(boundary);
 
             OPENMS_LOG_DEBUG << "Adding Swath centered at " << center
               << " m/z with an isolation window of " << lower << " to " << upper
-              << " m/z." << std::endl;
+              << " m/z." <<  " and start of " << lowerIm << " and IM end of " << upperIm << std::endl;
```
timosachsenberg, line 399:
```suggestion
              << " m/z and start of " << lowerIm << " and IM end of " << upperIm << std::endl;
```

src/openms/include/OpenMS/FORMAT/DATAACCESS/SwathFileConsumer.h
jcharkow: Thanks for the catch. I fixed the bracket placement (https://github.com/OpenMS/OpenMS/pull/5911/commits/e4b7a5121360cd5d63bde4d3ab03fce3a701d340). What were you thinking in terms of a test here, because I think the window placement is already properly tested in the openswath workflow test?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -502,14 +502,17 @@ namespace OpenMS
...
+    // (iii) map transitions to individual DIA windows for cases where this is
+    // non-trivial (e.g. when there is m/z overlap and a transition could be
+    // extracted from more than one window
+    std::vector<int> tr_win_map; // maps transition k to dia map i from which it should be extracted
     if (prm_)
     {
       // Here we deal with overlapping PRM / DIA windows: we only want to extract
       // each peptide from a single window and we assume that PRM windows are
       // centered around the target peptide. We therefore select for each peptide
       // the best-matching PRM / DIA window:
```
timosachsenberg, line 514: comment how best-matching is defined?

timosachsenberg (approved):
looks good to me

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
jcharkow: "Best matching" is slightly different for PRM or PASEF,  for PASEF is described on lines 545-546.

timosachsenberg: Would it make sense to expose this information more prominent to the user?
E.g., 
https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1OpenSwathWorkflow.html#details

If it is just an unimportant implementation detail it is of course not needed, but I imagined someone might ask "How do you deal with overlapping m/z windows" etc.

jcharkow: Good idea! I've added a description (see  https://github.com/OpenMS/OpenMS/pull/5911/commits/61089c972d08eefa6da52fd8d5e27acda46bd79c)

I'm not sure if the description is too verbose so I've left this conversation open.

jpfeuffer:
I think on the big distributions you will by default not have the same stdlib as on MacOS but the one from g++.
You probably need to check every occurrence of unordered_maps and random number generators in your code path to make sure they are sorted afterwards and replace random with boost::random.

You can also look at your full test output here until the next PR job starts:
https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/bldtst/compiler=appleclang-11.0.0,os_label=mojave/ws/build/src/tests/

jpfeuffer:
I invited you to the Developers team also.

jcharkow:
Thank you but I don't think I received an invitation anywhere 

> I invited you to the Developers team also.

jcharkow:
> I think on the big distributions you will by default not have the same stdlib as on MacOS but the one from gnu/g++. You probably need to check every occurrence of unordered_maps and random number generators in your code path to make sure they are sorted afterwards and replace random with boost::random.
> 
> You can also look at your full test output here until the next PR job starts: https://abibuilder.informatik.uni-tuebingen.de/jenkins/job/openms/job/PR/job/bldtst/compiler=appleclang-11.0.0,os_label=mojave/ws/build/src/tests/

Thanks, I'll check this out.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -502,14 +502,17 @@ namespace OpenMS
       writeOutFeaturesAndChroms_(chromatograms, featureFile, out_featureFile, store_features, chromConsumer);
     }
 
-    std::vector<int> prm_map;
+    // (iii) map transitions to individual DIA windows for cases where this is
+    // non-trivial (e.g. when there is m/z overlap and a transition could be
+    // extracted from more than one window
+    std::vector<int> tr_win_map; // maps transition k to dia map i from which it should be extracted
     if (prm_)
```
jpfeuffer, line 509: what if both prm and pasef are true? the vector resizes will not clear the current mapping.

jpfeuffer:
Can you try with one thread? Does it change anything?

jcharkow:
I have been running with one thread. When I try multithreading the output does not change though.

jcharkow:
The strange order of interweaving peptideB and peptideC is because sorting is occurring at the fragment ion level (see ChromatogramExtractor.cpp:prepare_coordinates line 211: 
<code>std::sort(coordinates.begin(), coordinates.end(), ChromatogramExtractor::ExtractionCoordinates::SortExtractionCoordinatesByMZ);</code> Because peptideB and peptideC have the same fragment ions the order of chromatograms is ambiguous. 

I think changing the fragment ion m/z to be different should fix the issue. I am not sure if in the future we want to change the functionality so that we have sorting by m/z and then IM? But it seems unlikely that would run into a case like this in a non dummy test experiment.

timosachsenberg:
Not sure if this applies here but the less efficient https://en.cppreference.com/w/cpp/algorithm/stable_sort could potentially be used

jcharkow:
> Not sure if this applies here but the less efficient https://en.cppreference.com/w/cpp/algorithm/stable_sort could potentially be used

I got it working on mac but if we had another test file like this (with the same fragment ion mzs) we might run into this issue in a different context. So it might be worth considering changing the method?

jpfeuffer:
If you want, you can make use of the "-test" flag. And only use your slower solution when this is active.

timosachsenberg:
Some background:
[std::stable_sort](http://en.cppreference.com/w/cpp/algorithm/stable_sort) performs NlogN comparisons when sufficient memory is available. When insufficient memory is available, it degrades to N((logN)^2) comparisons. Therefore, it is roughly of the same efficiency as [std::sort](http://en.cppreference.com/w/cpp/algorithm/sort) (which performs O(NlogN) comparisons in both average and worst case) when memory is available.

In practice, stable_sort is about 5-20% slower. For many use cases this might not make much of a difference.

jcharkow:
I decided to change it to `std::stable_sort` for all use cases. I don't think it will have a big impact on performance in practice and I think this will be good for avoiding this issue in the long run. see https://github.com/OpenMS/OpenMS/pull/5911/commits/8226defd05e488eebac96e86a3d34a83c5c69b57

timosachsenberg:
can you take a look at https://cdash.openms.de/viewTest.php?onlyfailed&buildid=21054 maybe you or @hroest know why these tests are failing in debug mode

PR #5917, opened 2022-03-14
FeatureFinderIdentification add test for `-candidates_out`

bernt-matthias:
# Description

The `-candidates_out` option seems unused. 

# Checklist:

# Note:

### DIFF
--- a/src/tests/topp/CMakeLists.txt
+++ b/src/tests/topp/CMakeLists.txt
@@ -316,7 +316,7 @@ set_tests_properties("TOPP_FeatureFinderIdentification_3_out1" PROPERTIES DEPEND
 ##set_tests_properties("TOPP_FeatureFinderIdentification_4_out1" PROPERTIES DEPENDS "TOPP_FeatureFinderIdentification_4")
 
 ## with batch extraction size smaller than the nr of peptides (we need to whitelist feature ids, since they might be generated differently)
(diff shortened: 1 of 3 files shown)


timosachsenberg:
I think it is used in the original publication. Does is cause any troubles?

bernt-matthias:
> I think it is used in the original publication. Does is cause any troubles?

The Galaxy tests (which are generated from the CMakeFile) fail, because the output file is missing. 

This is because the `candidates_out` file is simply not written. 

In case the file is needed, the argument should be used somewhere ...

timosachsenberg:
k will check why it is not used

timosachsenberg:
should be used by algorithms by copying to it: https://github.com/OpenMS/OpenMS/blob/develop/src/topp/FeatureFinderIdentification.cpp#L228

bernt-matthias:
What is this line doing: https://github.com/OpenMS/OpenMS/blob/3cd3db31efef6241030830388dbceccb3d6ffd4e/src/topp/FeatureFinderIdentification.cpp#L209 ? 

If the parameter should be used it might be a good idea to have it covered by a test.

bernt-matthias:
Test now fails with `Error: File not found (the file 'FeatureFinderIdentification_5_candidates.tmp' could not be found)`. Which verifies my observation.

enetz:
this looks like the same issue as in #5920 (`.tmp` file ending), since the same function `FileHandler::storeFeatures` is used to write the file:
https://github.com/OpenMS/OpenMS/blob/62a701944098259fb7cebd4e6affc6c54f32c361/src/openms/source/TRANSFORMATIONS/FEATUREFINDER/FeatureFinderIdentificationAlgorithm.cpp#L612

timosachsenberg:
rebuild jenkins

timosachsenberg (approved):
Thanks! looks good.

PR #5920, opened 2022-03-14
Decharger: `-out_fm` not working? Add testfor  `-out_fm` and `-outpairs`

bernt-matthias:
# Description

`-out_fm` seems not to work

# Checklist:

# Note:


enetz:
The issue seems to be the `.tmp` file ending. `FileHandler::storeFeatures` does not write the file, if it isn't a feature file type:
https://github.com/OpenMS/OpenMS/blob/62a701944098259fb7cebd4e6affc6c54f32c361/src/openms/source/FORMAT/FileHandler.cpp#L634

bernt-matthias:
I see. 

Maybe one could introduce a `force_type` parameter like here (in this case the type is known in advance, or?) 

https://github.com/OpenMS/OpenMS/blob/62a701944098259fb7cebd4e6affc6c54f32c361/src/openms/source/FORMAT/FileHandler.cpp#L589

and/or at least write out some warning?

bernt-matthias:
> we could use `FeatureXMLFile::store` directly. I am not sure whether this is the right thing to do here

I think at the moment its the only allowed type for this output https://github.com/OpenMS/OpenMS/blob/62a701944098259fb7cebd4e6affc6c54f32c361/src/topp/Decharger.cpp#L115

But we could allow more if desired, I don't care :)

bernt-matthias:
Is there any way to do the clang reformatting automatically?

timosachsenberg (approved):

PR #5947, opened 2022-03-16
Fix/asan

hroest:
# Description

Currently we have 20 tests that fail the Address Sanitizer test: https://cdash.openms.de/viewTest.php?onlyfailed&buildid=16768

This fixes the majority of them, some are harder to fix such as the MzML and GUI tests so I left them for now. The majority of the issues came from the libsvm and code around it. The SVMWrapper still has some *highly unintuitive* interfaces with unclear ownership, but it has improved by quite a bit now.


cbielow (approved):
nice.

maybe we should switch to smart pointers at some point to avoid all this manual memory management.

timosachsenberg:
something seems of with jenkins... maybe my attempts to fix the zlib linking had some unwanted side effect.

timosachsenberg:
rebuild jenkins

src/tests/class_tests/openms/source/LibSVMEncoder_test.cpp
```diff
@@ -536,45 +548,72 @@ START_SECTION((void encodeOligo(const AASequence &sequence, UInt k_mer_length, c
...
+    nodes[i][count - 1].index = -1;
+    labels[i] = i * 2 / 3 + 0.03;
+    label_vector1.push_back(labels[i]);
+  }
+  problem->x = nodes;
+  problem->l = count;
+  problem->y = labels;
+
+  LibSVMEncoder::destroyProblem(problem); // cleanup
+  TEST_EQUAL(problem, nullptr)
```
timosachsenberg, line 614:
```suggestion
  TEST_EQUAL(problem, nullPointer)
```

timosachsenberg:
seems we get a segfault on mac now https://cdash.openms.de/test/8546502

hroest:
> seems we get a segfault on mac now https://cdash.openms.de/test/8546502

clearly this made things worse :-(

hroest:
strange, ASAN is fine with it but clearly OSX / clang is not ...

timosachsenberg:
now passes on mac

timosachsenberg:
we might need to watch this a bit. Could be that we have a segfault on mac again in https://cdash.openms.de/viewTest.php?onlyfailed&buildid=20718
Which I think is based on develop after the merge.
Could be unstable

hroest:
>  Could be unstable

seems like it, maybe we just leave those changes and tackle them at some other time

Issue #5940, opened 2022-03-15
fix compile warnings on mac


timosachsenberg:
Please mark (or copy the warning if you can't tick checkboxes) the warning that you are planning to fix. If you run into troubles or don't manage to fix it in e.g. one day please uncheck the box again so someone else can take the sub issue.

- [ ] src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h:122:10: warning: private field 'sonar_' is not used [-Wunused-private-field]
- [x] src/openms/include/OpenMS/DATASTRUCTURES/Utils/MatrixUtils.h:53:3: warning: unused function 'convertOpenMSMatrix2EigenMatrixXd' [-Wunused-function]
- [x] src/openms/include/OpenMS/DATASTRUCTURES/Utils/MatrixUtils.h:67:3: warning: unused function 'matrixIsIdentityMatrix' [-Wunused-function]
- [ ] src/openms/include/OpenMS/FILTERING/DATAREDUCTION/SplineInterpolatedPeaks.h:160:16: warning: private field 'pos_min_' is not used [-Wunused-private-field]
- [ ] src/openms/include/OpenMS/KERNEL/RangeUtils.h:187:23: note: place parentheses around the '!=' expression to silence this warning - warning: ^ has lower precedence than !=; != will be evaluated first [-Wparentheses]
- [ ] src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp:61:14: warning: unused function 'callback' [-Wunused-function]
- [ ] src/openms/source/FILTERING/CALIBRATION/TOFCalibration.cpp:198:23: warning: unused variable 'spec_iter' [-Wunused-variable]
- [x] src/openms/source/QC/DBSuitability.cpp:377:107: note: use array indexing to silence this warning - warning: adding 'OpenMS::Int' (aka 'int') to a string does not append to the string [-Wstring-plus-int]
- [x] src/openms/source/QC/DBSuitability.cpp:397:107: note: use array indexing to silence this warning - warning: adding 'OpenMS::Int' (aka 'int') to a string does not append to the string [-Wstring-plus-int]
- [ ] src/openms_gui/include/OpenMS/VISUAL/TOPPASVertex.h:191:10: warning: 'OpenMS::TOPPASVertex::paint' hides overloaded virtual function [-Woverloaded-virtual]
[...]

PR #5953, opened 2022-03-16
[FIX] Suppress -Wpessimizing-move in tests

matchy233:
# Description

This PR fixes the warnings related to `-Wpessimizing-move` in #5940. I specifically suppressed the warning in the two tests with intentional pessimistic move.

# Checklist:

# Note:


jpfeuffer:
Hmm I was rather thinking about actually avoiding all pessimizing moves altogether.

jpfeuffer:
Ah those tests have a note. Then it is fine! thanks

jpfeuffer (approved):

matchy233:
I observed that the build on Windows failed since MSVC [haven't supported pessimizing-move compiler warning yet](https://developercommunity.visualstudio.com/t/pessimizing-move-compiler-warning/1044200). I will add checks for compiler in `src/tests/class_tests/openms/CMakeLists.txt` soon.

jpfeuffer (changes requested):

src/tests/class_tests/openms/CMakeLists.txt
```diff
@@ -81,6 +81,12 @@ endif()
 foreach(_class_test ${TEST_executables})
   add_executable(${_class_test} source/${_class_test}.cpp)
   target_link_libraries(${_class_test} ${OpenMS_LIBRARIES})
+
+  # Suppress std::move warning for MSSpectrum_test and MSChromatogram_test
+  if (_class_test STREQUAL "MSSpectrum_test" OR _class_test STREQUAL "MSChromatogram_test")
+    target_compile_options(${_class_test} PUBLIC -Wno-pessimizing-move)
```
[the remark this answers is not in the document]
matchy233: yep! I noticed that just now. Will make adjustment in the next commit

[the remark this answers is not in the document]
matchy233: Also thank you for introducing genex :0...! I've never used genex before so my current workaround was just adding `NOT MSVC` in the `if()` statement...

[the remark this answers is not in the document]
jpfeuffer: I think in this case it should be equivalent, so you can leave it.

jpfeuffer (approved):

src/tests/class_tests/openms/CMakeLists.txt
```diff
@@ -81,6 +81,12 @@ endif()
 foreach(_class_test ${TEST_executables})
   add_executable(${_class_test} source/${_class_test}.cpp)
   target_link_libraries(${_class_test} ${OpenMS_LIBRARIES})
+
+  # Suppress std::move warning for MSSpectrum_test and MSChromatogram_test
+  if (NOT MSVC AND (_class_test STREQUAL "MSSpectrum_test" OR _class_test STREQUAL "MSChromatogram_test"))
+    target_compile_options(${_class_test} PUBLIC -Wno-pessimizing-move)
```
cbielow, line 87: sorry I'm a bit late to the party, but why not disable the warning in the test itself? e.g. using
`#pragma clang diagnostic ignored "-Wpadded"`
This way, it is easier to remember that the warning is disabled there (knowing this section of the build system seems only possible if your name is JP :))

matchy233: Yep I understand such way of suppressing warning and agree with your points. But I know that the `#pragma`  syntax for disabling warning is sorta difference for MSVC, so I'm not sure whether using `#pragma clang diagnostic ignored` will break Windows CI again...

matchy233: Maybe I can try building on a Windows machine using GitHub Actions to test the effect.

timosachsenberg: check out these lines https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/SYSTEM/SysInfo.cpp#L38 you can also put the pragma in such blocks if you want to only enable them on linux or mac

matchy233: Thanks for the suggestions! I've updated my code (and rebased to `develop`)

timosachsenberg:
awesome. just minor comment: could this be moved directly to the code that does the move?
doesn't need to be active for the full test

matchy233:
Sure @timosachsenberg, just modified in the new commit

src/tests/class_tests/openms/source/MSSpectrum_test.cpp
```diff
@@ -499,7 +494,15 @@ START_SECTION((MSSpectrum& operator= (const MSSpectrum&& source)))
 
   //Assignment of empty object
   //normal assignment
+#ifndef OPENMS_WINDOWSPLATFORM
+#pragma clang diagnostic push
+// Ignore -Wpessimizing-move, becuase it's intentional
```
timosachsenberg, line 499:
```suggestion
// Ignore -Wpessimizing-move, because we want to test the move assignment operator.
```

src/tests/class_tests/openms/source/MSChromatogram_test.cpp
```diff
@@ -904,7 +904,15 @@ START_SECTION((MSChromatogram& operator=(const MSChromatogram &&source)))
 
   //Assignment of empty object
   //normal assignment
+#ifndef OPENMS_WINDOWSPLATFORM
+#pragma clang diagnostic push
+// Ignore -Wpessimizing-move, becuase it's intentional
```
timosachsenberg, line 909:
```suggestion
// Ignore -Wpessimizing-move, because we want to test the move assignment operator.
```

PR #5954, opened 2022-03-16
[DOC] Require and document the actually required min Qt version

### LINKED ISSUE #5604: Qt min version
currently our cmake file says `SET(QT_MIN_VERSION "5.5.0")` however OpenMS doesnt compile any more with Qt 5.5 due to changes introduce over a year ago. Either we fix these changes or we figure out what the latest Qt version is that works again and check against current Linux distros what they ship.

jpfeuffer:
fixes #5604

for the most part. As discussed, I think this is reasonable.

### DIFF
--- a/doc/doxygen/install/install-linux.doxygen
+++ b/doc/doxygen/install/install-linux.doxygen
@@ -50,7 +50,7 @@
   <ul>
     <li>
       Essential are \b gcc/g++ (>= 7.0) or similar ANSI C++ compiler with full C++17 support,
-      \b CMake (>= 3.1; with >= 3.9 suggested), \b Qt5, \b patch, \b autoconf (> 2.60), \b automake (> 1.9), \b libtool (libtoolize/glibtoolize).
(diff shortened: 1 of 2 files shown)


doc/doxygen/install/install-linux.doxygen
```diff
@@ -50,7 +50,7 @@
   <ul>
     <li>
       Essential are \b gcc/g++ (>= 7.0) or similar ANSI C++ compiler with full C++17 support,
-      \b CMake (>= 3.1; with >= 3.9 suggested), \b Qt5, \b patch, \b autoconf (> 2.60), \b automake (> 1.9), \b libtool (libtoolize/glibtoolize).
+      \b CMake (>= 3.1; with >= 3.9 suggested), \b Qt5 (>= 5.6), \b patch, \b autoconf (> 2.60), \b automake (> 1.9), \b libtool (libtoolize/glibtoolize).
```
timosachsenberg, line 53: Is the cmake version still correct? I thought it might be higher now

jpfeuffer: Yes it's probably higher but I don't know how high. We can parse the Cmake minimum version statements but this does not help in checking since when Cmake has functional find modules

PR #5955, opened 2022-03-16
Fix/gui

hroest:
# Description

Several fixes that cause segfaults / memory leaks in TOPPView

- major bugfix for chromatogram view: use-after-free fixed for `layer` which caused random segfaults when viewing chromatograms.
- `Painter1DBase` needs a virtual destructor since its a base class, now the correct destructor is called (for derived, not base class)

### DIFF
--- a/src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
+++ b/src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
@@ -112,6 +112,13 @@ namespace OpenMS
     connect(this, &QTabWidget::tabBarDoubleClicked, this, &DataSelectionTabs::tabBarDoubleClicked);
   }
 
+  DataSelectionTabs::~DataSelectionTabs()
+  {
+    delete spectraview_controller_;
+    delete idview_controller_;
+    delete diatab_controller_;
+  }
+
   LayerDataBase* getCurrentLayerData(TOPPViewBase* tv)
   {
     PlotCanvas* cc = tv->getActiveCanvas();
--- a/src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
+++ b/src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
@@ -199,15 +199,16 @@ namespace OpenMS
       // only need to grab the one with the desired index
       ExperimentSharedPtrType exp_sptr = layer.getChromatogramData();
       auto ondisc_sptr = layer.getOnDiscPeakData();
+      auto annotation = layer.getChromatogramAnnotation();
+      String fname = layer.filename;
 
       widget_1d->canvas()->removeLayers();
 
       // fix legend and set layer name
-      String fname = layer.filename;
       String caption = fname + "[" + index + "]";
 
       // add chromatogram data as peak spectrum and update other controls
-      widget_1d->canvas()->addChromLayer(exp_sptr, ondisc_sptr, layer.getChromatogramAnnotation(), index, fname, caption, false);
+      widget_1d->canvas()->addChromLayer(exp_sptr, ondisc_sptr, annotation, index, fname, caption, false);
 
       tv_->updateBarsAndMenus(); // needed since we blocked update above (to avoid repeated layer updates for many layers!)
     }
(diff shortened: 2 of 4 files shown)


src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
```diff
@@ -112,6 +112,13 @@ namespace OpenMS
     connect(this, &QTabWidget::tabBarDoubleClicked, this, &DataSelectionTabs::tabBarDoubleClicked);
   }
 
+  DataSelectionTabs::~DataSelectionTabs()
+  {
+    delete spectraview_controller_;
+    delete idview_controller_;
+    delete diatab_controller_;
```
hroest, line 119: @cbielow can you check whether that is okay? I did some testing and these came up with `TOPPView_test` as leaking memory but can you have a look at what the ownership model is here and whether its safe to destroy them? or are these still needed later? in my tests it all ran fine

src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
```diff
@@ -236,13 +237,14 @@ namespace OpenMS
       // first get raw data (the full experiment with all chromatograms), we
       // only need to grab the ones with the desired indices
       ExperimentSharedPtrType exp_sptr = layer.getChromatogramData();
-      auto ondisc_sptr = layer.getOnDiscPeakData();
 
-      widget_1d->canvas()->removeLayers();
+      String fname = layer.filename;
+      auto annotation = layer.getChromatogramAnnotation();
+      auto ondisc_sptr = layer.getOnDiscPeakData();
+      widget_1d->canvas()->removeLayers(); // this actually deletes layer, make sure its not used any more
```
hroest, line 244: random segfaults result when `layer` is used after calling `widget_1d->canvas()->removeLayers();`

src/openms_gui/source/VISUAL/DataSelectionTabs.cpp
cbielow: +1
should be good.
I was wondering why I did not use a plain member, but its simply due to compilation time PIMPL.
We could simply make them members as well. Or switch to unique_ptr.

src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
```diff
@@ -199,15 +199,16 @@ namespace OpenMS
       // only need to grab the one with the desired index
       ExperimentSharedPtrType exp_sptr = layer.getChromatogramData();
       auto ondisc_sptr = layer.getOnDiscPeakData();
+      auto annotation = layer.getChromatogramAnnotation();
```
cbielow, line 202: wow. That this worked in the first place, is suprising.
Could you please, for increased safety, add a local scope for `layer`, so people do not use it after `widget_1d->canvas()->removeLayers();` was called?

src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
cbielow: nice fix

cbielow (approved):

src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
hroest: > That this worked in the first place, is suprising.

I dont think it did, hence the random segfaults...it *kinda* worked as the reference/ptr was still valid for a small while and so for `TVSpectraViewController::activate1DSpectrum(int index)` it actually worked since there is no memory allocation between deleting and `use-after-free` but for the second function with the loop it did not work since new memory was allocated in the mean time.

The scope is actually quite a bit of code to add so I am not sure if its worth it since surely, *surely* we wont make the same error twice?!
```
diff --git a/src/openms_gui/source/VISUAL/TVSpectraViewController.cpp b/src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
index 2aff7fb57d..e21e83896e 100644
--- a/src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
+++ b/src/openms_gui/source/VISUAL/TVSpectraViewController.cpp
@@ -181,12 +181,27 @@ namespace OpenMS
     if (widget_1d == nullptr) return;
     if (widget_1d->canvas()->getLayerCount() == 0) return;
 
-    LayerDataBase& layer = widget_1d->canvas()->getCurrentLayer();
+    ExperimentSharedPtrType exp_sptr;
+    boost::shared_ptr<OpenMS::OnDiscMSExperiment> ondisc_sptr;
+    boost::shared_ptr<OpenMS::OSWData> annotation;
+    String fname;
+    bool is_chromatogram;
+
+    // local scope for layer since we delete it later with removeLayers() and
+    // dont want to accidentially use it
+    {
+      LayerDataBase& layer = widget_1d->canvas()->getCurrentLayer();
+      exp_sptr = layer.getChromatogramData();
+      ondisc_sptr = layer.getOnDiscPeakData();
+      annotation = layer.getChromatogramAnnotation();
+      fname = layer.filename;
+      is_chromatogram = layer.chromatogram_flag_set();
+    }
 
     // If we have a chromatogram, we cannot just simply activate this spectrum.
     // we have to do much more work, e.g. creating a new experiment with the
     // new spectrum.
-    if (!layer.chromatogram_flag_set())
+    if (!is_chromatogram)
     {
       widget_1d->canvas()->activateSpectrum(index);
     }
@@ -197,10 +212,6 @@ namespace OpenMS
 
       // first get raw data (the full experiment with all chromatograms), we
       // only need to grab the one with the desired index
-      ExperimentSharedPtrType exp_sptr = layer.getChromatogramData();
-      auto ondisc_sptr = layer.getOnDiscPeakData();
-      auto annotation = layer.getChromatogramAnnotation();
-      String fname = layer.filename;
 
       widget_1d->canvas()->removeLayers();
```

Issue #5905, opened 2022-03-08
Properly model Elements and Isotopes


hroest:
# Problem

Currently both individual elements as well as Isotopes are modelled using `Element` which is a hacky conceptually, this implies that "an isotope  is the same as an element":

```
>>> ElementDB().getElement("O")
<pyopenms.pyopenms_2.Element object at 0x7f235032b670>
>>> ElementDB().getElement("(17)O")
<pyopenms.pyopenms_2.Element object at 0x7f235032b2b0>
```

with some strange side effects, eg (17)O is treated as a completely different Element than O and has its own isotope distribution and there is no link back from (17)O to the oxygen element (especially as we dont know that this is an isotope and not an element, we can use the atomic number though)

```
>>> ElementDB().getElement("(17)O").getIsotopeDistribution().getContainer()
[<pyopenms.pyopenms_4.Peak1D object at 0x7f235032b990>]
>>> ElementDB().getElement("O").getIsotopeDistribution().getContainer()
[<pyopenms.pyopenms_4.Peak1D object at 0x7f2370355090>, <pyopenms.pyopenms_4.Peak1D object at 0x7f235032b950>, <pyopenms.pyopenms_4.Peak1D object at 0x7f235032b670>]
```

It also means that we cannot store additional information that is isotope-specific, such as number of neutrons, half lives for radioactive elements, decay modes etc.

# Solutions

The main solution would be to add another class `Isotope` to properly model isotopes, however there are a few issues with this. We would then model the relationship with a `std::vector<Isotope>` as a member of `Element`. 
 
The main problem with this is that 

(i) many parts of OpenMS rely on `ElementDB().getElement("O").getIsotopeDistribution()` which should be reasonably fast and 
(ii) `ElementDB().getElement("(17)O")` should produce something that is of type `Element` since this can be used inside an `EmpiricalFormula`. 

[...]

PR #5956, opened 2022-03-16
Feature/isotope class

hroest:
# Description

Adds a new Isotope class that can store individual properties about isotopes, specifically

- mass
- abundance
- half-life
- decay mode

Currently, the `Element` has a list of `Isotope`s attached and the `ElementDB` has a function for both `getElement` and `getIsotope` to retrieve specific isotopes. Currently I have added a few isotopes to test, specifically Uranium and Carbon (C14). 

This addresses #5905


src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
```diff
@@ -113,6 +124,30 @@ namespace OpenMS
...
+     * Adds a new isotope (or replaces an existing one if @em replace_existing is true). 
+     *
+     * @param name Common name of the element
+     * @param symbol Element symbol (one or two letter)
+     * @param an Atomic number (number of protons)
+     * @param abundance natural abundance of the isotope (use 0 for unstable isotopes)
+     * @param mass mass of the isotope
+     * @param half_life Half life of the isotope in seconds (use -1 for stable isotopes)
+     * @param decay Main decay mode for unstable isotopes (use NONE for stable isotopes)
+     * @param replace_existing Whether to replace an existint isotope
```
timosachsenberg, line 139:
```suggestion
     * @param replace_existing Whether to replace an existing isotope
```
maybe document what happens if replace_existing = false and an existing one is passed (no-op? exeption?)

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
```diff
@@ -113,6 +124,30 @@ namespace OpenMS
...
+     *
+     * @note Do not use this function inside parallel code as it modifies a singleton that is shared between threads.
+    */
+    void addIsotope(const std::string& name,
+                    const std::string& symbol,
+                    const unsigned int an,
```
timosachsenberg, line 145: merely style but here we pass a const int and below double... either or would be fine

src/openms/include/OpenMS/CHEMISTRY/Isotope.h
```diff
@@ -0,0 +1,133 @@
...
+    /// Whether this is a stable isotope
+    bool isStable() const {return half_life_ < 0;}
+
+    /// writes the isotope to an output stream
+    friend OPENMS_DLLAPI std::ostream & operator<<(std::ostream & os, const Isotope & isotope);
+
+  protected:
+
+    int neutrons_ = -1;
+    double abundance_; // has to consistent with getIsotopeDistribution()
```
timosachsenberg, line 125: initialize it here as well?
```suggestion
    double abundance_ = -1; // has to consistent with getIsotopeDistribution()
```

src/openms/source/CHEMISTRY/Element.cpp
```diff
@@ -111,10 +114,33 @@ namespace OpenMS
...
+    updateIsotopeDistr(); // calculate new cached distribution
+  }
+
+  void Element::updateIsotopeDistr()
+  {
+    auto dist = isotope_distr_.getContainer();
+    dist.clear();
+    for (const auto& isotope : isotopes_)
+    {
+      dist.push_back(Peak1D(isotope->getMonoWeight(), isotope->getAbundance()));
```
timosachsenberg, line 138: in-place construction?
dist.emplace_back({isotope->getMonoWeight(), isotope->getAbundance()});

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
```diff
@@ -90,6 +90,15 @@ namespace OpenMS
     for (; it != formula.end(); ++it)
     {
       IsotopeDistribution tmp = it->first->getIsotopeDistribution();
+
+      // filter out unstable isotopes / those with zero natural abundance as
+      // they just blow up the computations we have do to but have zero
+      // abundance
+      auto data = tmp.getContainer();
```
timosachsenberg, line 97: do we need to copy here?
would auto&
work and omiting the tmp.set?

src/openms/source/CHEMISTRY/Isotope.cpp
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  std::ostream & operator<<(std::ostream & os, const Isotope & isotope)
+  {
+    os  << isotope.name_ << " "
+    << isotope.symbol_ << " Z="
+    << isotope.atomic_number_ << " N="
+    << isotope.neutrons_ << " : "
+    << isotope.mono_weight_ << " "
+    << isotope.abundance_ *100 << "% : ";
```
timosachsenberg, line 132:
```suggestion
    << isotope.abundance_ * 100 << "% : ";
```

src/openms/source/CHEMISTRY/Isotope.cpp
```diff
@@ -0,0 +1,146 @@
...
+    if (isotope.isStable())
+    {
+      os << "stable";
+    }
+    else
+    {
+      os << " half life: " << isotope.half_life_ << " s "
```
timosachsenberg, line 139:
```suggestion
      os << "half life: " << isotope.half_life_ << " s "
```

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
```diff
@@ -86,12 +87,21 @@ namespace OpenMS
     /// returns a hashmap that contains atomic numbers mapped to pointers of the elements
     const std::unordered_map<unsigned int, const Element*>& getAtomicNumbers() const;
 
+    /// returns a hashmap that contains atomic numbers mapped to pointers of the isotopes
```
timosachsenberg, line 90: I think the comment is off here

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
hroest: I think we would have to add a non-const function for this

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
hroest, line 145:
```suggestion
                    unsigned int an,
```

src/openms/source/CHEMISTRY/Element.cpp
hroest, line 138:
```suggestion
      dist.emplace_back({isotope->getMonoWeight(), isotope->getAbundance()});
```

the hope is that in the long term we wont need two redundant ways to store the isotope distribution anyways...

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
hroest, line 90:
```suggestion
    /// returns a hashmap that contains isotopic names mapped to pointers of the isotopes
```

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -125,6 +142,46 @@ namespace OpenMS
...
+  {
+    unsigned int mass_number = std::round(mass);
+    string iso_name = "(" + std::to_string(mass_number) + ")" + name;
+    string iso_symbol = "(" + std::to_string(mass_number) + ")" + symbol;
+
+    if (hasElement( iso_symbol )  )
+    {
+      if (!replace_existing)
+      {
+        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, String("Element with atomic number ") + an + " already exists");
```
hroest, line 162:
```suggestion
        throw Exception::IllegalArgument(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, String("Isotope with symbol ") + iso_symbol + " already exists and replace_existing is set to false.");
```

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
hroest, line 139:
```suggestion
     * @param replace_existing Whether to replace an existing isotope
     *
     * @exception Exception::IllegalArgument is thrown if the isotope already exists
```

src/openms/source/CHEMISTRY/ISOTOPEDISTRIBUTION/CoarseIsotopePatternGenerator.cpp
hroest: Btw: this is not the slow part of the function :-)

timosachsenberg:
some tests fail https://cdash.openms.de/viewTest.php?onlyfailed&buildid=19323

timosachsenberg:
Thanks! Test pass. I added a few potential reviewers because I think it doesn't hurt to have more eyes on this very central class.

src/pyOpenMS/tests/unittests/test000.py
```diff
@@ -5015,6 +5016,263 @@ def testHMMState():
...
+
+    #  not yet implemented
+    #
+    # const Map[ String, Element * ]  getNames() nogil except +
+    # const Map[ String, Element * ] getSymbols() nogil except +
+    # const Map[unsigned int, Element * ] getAtomicNumbers() nogil except +
+
+
+@report
+def testDPosition():
```
timosachsenberg, line 5081: why also move DPosition out of test000?

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
```diff
@@ -127,30 +166,42 @@ namespace OpenMS
...
-
-    **/
-    IsotopeDistribution parseIsotopeDistribution_(const std::map<unsigned int, double>& abundance, const std::map<unsigned int, double>& mass);
-
     /** calculates the average weight based on isotope abundance and mass
      **/
     double calculateAvgWeight_(const std::map<unsigned int, double>& abundance, const std::map<unsigned int, double>& mass);
 
-    /**_ calculates the mono weight based on the most abundant isotope 
+    /**_ calculates the mono weight based on the most abundant isotope
```
cbielow, line 173:
```suggestion
    /** Calculates the mono weight based on the most abundant isotope
```

src/openms/source/CHEMISTRY/Element.cpp
```diff
@@ -111,10 +114,34 @@ namespace OpenMS
...
+    return isotope_distr_;
+  }
+
+  void Element::setIsotopes(const std::vector<const Isotope*>& isotopes)
+  {
+    isotopes_ = isotopes;
+    updateIsotopeDistr(); // calculate new cached distribution
+  }
+
+  void Element::updateIsotopeDistr()
```
cbielow, line 132: can this method be private/protected?

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -35,14 +35,17 @@
...
+#include <OpenMS/CHEMISTRY/Isotope.h>
 #include <OpenMS/DATASTRUCTURES/String.h>
 #include <OpenMS/CONCEPT/Exception.h>
-#include <OpenMS/CHEMISTRY/Element.h>
 #include <iostream>
 #include <cmath>
 
 using namespace std;
 
+int SECONDS_PER_YEAR = 31556952; // Gregorian year
```
cbielow, line 47:
```suggestion
constexpr int SECONDS_PER_YEAR = 31556952; // Gregorian year
```

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
```diff
@@ -127,30 +166,42 @@ namespace OpenMS
...
+                        const std::string& symbol,
+                        const unsigned int an,
+                        const std::map<unsigned int, double>& abundance,
+                        const std::map<unsigned int, double>& mass,
+                        const std::map<unsigned int, double>& half_life,
+                        const std::map<unsigned int, Isotope::DecayMode>& decay);
 
     /// add element objects to documentation maps
-    void addElementToMaps_(const std::string& name, const std::string& symbol, const unsigned int an, const Element* e);
+    Element* addElementToMaps_(const std::string& name, const std::string& symbol, const unsigned int an, Element* e);
```
cbielow, line 201:
```suggestion
    const Element* addElementToMaps_(const std::string& name, const std::string& symbol, const unsigned int an, const Element* e);
```

cbielow: this can be made `const`, if I'm not mistaken.

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
```diff
@@ -127,30 +166,42 @@ namespace OpenMS
...
 
-    /// constructs isotope objects
-    void storeIsotopes_(const std::string& name, const std::string& symbol, const unsigned int an, const std::map<unsigned int, double>& Z_to_mass, const IsotopeDistribution& isotopes);
+    /// add element objects to documentation maps
+    Isotope* addIsotopeToMaps_(const std::string& name, const std::string& symbol, Isotope* e);
```
cbielow, line 204:
```suggestion
    const Isotope* addIsotopeToMaps_(const std::string& name, const std::string& symbol, const Isotope* e);
```

cbielow: same here

src/openms/source/CHEMISTRY/Element.cpp
cbielow: It seems to the used in `addIsotope`, but is this part really that performance critical that we burden the user with taking care of updating the isotope distribution if an isotope was replaced?

src/openms/source/CHEMISTRY/Element.cpp
```diff
@@ -111,10 +114,34 @@ namespace OpenMS
...
+  {
+    auto dist = isotope_distr_.getContainer();
+    dist.clear();
+    dist.reserve(isotopes_.size());
+    for (const auto& isotope : isotopes_)
+    {
+      dist.emplace_back(isotope->getMonoWeight(), isotope->getAbundance());
+    }
+    isotope_distr_.set(dist);
```
cbielow, line 141:
```suggestion
    isotope_distr_.set(std::move(dist));
```

src/tests/class_tests/openms/source/ElementDB_test.cpp
```diff
@@ -135,6 +152,70 @@ START_SECTION(void addElement(const std::string& name,
...
+  TEST_EXCEPTION(Exception::IllegalArgument, e_ptr->addElement("NewElement", "NE", 800u, oxygen_abundance, oxygen_mass, false))
+
+  // Test that we can add elements twice with replace=true 
+  e_ptr->addElement("NewElement", "NE", 800u, oxygen_abundance, oxygen_mass, true);
+}
+END_SECTION
+
+START_SECTION( void addIsotope(const std::string& name, const std::string& symbol, const unsigned int an, double abundance, double mass, double half_life, Isotope::DecayMode decay, bool replace_existing))
+{
+  const Isotope * iso1 = e_ptr->getIsotope("(238)U");
```
cbielow, line 171:
```suggestion
  const Isotope * iso1 = e_ptr->getIsotope("(238)U");
  TEST_REAL_SIMILAR(iso1->getAbundance(), 0.992742) // test natural abundance 
```

src/tests/class_tests/openms/source/ElementDB_test.cpp
```diff
@@ -135,6 +152,70 @@ START_SECTION(void addElement(const std::string& name,
...
+  TEST_EQUAL(e_ptr->getIsotope("(314)C") == nullptr, true)
+  int nr_isotopes = element->getIsotopes().size();
+  e_ptr->addIsotope("Uranium", "U", 92u, 0.3, 314.0, 1e5, Isotope::DecayMode::UNKNOWN, false);
+  const Isotope * new_iso = e_ptr->getIsotope("(314)U");
+  TEST_EQUAL( new_iso != nullptr, true)
+  TEST_EQUAL( new_iso->getElement(), element)
+  TEST_EQUAL( element->getIsotopes().size(), nr_isotopes+1) // increased number of isotopes by one
+  TEST_EQUAL( element->getIsotopeDistribution().getContainer().size(), nr_isotopes+1) // increased number of isotopes by one
+
+  // we have now managed to have 160% natural abundance for Uranium
```
cbielow, line 196: maybe this should be checked when updating the isotope distribution?!
Should be 1 +- eps, and exception otherwise?

src/tests/class_tests/openms/source/ElementDB_test.cpp
```diff
@@ -135,6 +152,70 @@ START_SECTION(void addElement(const std::string& name,
...
+  // NOTE: this is a major problem for average weight calculations etc
+  const Element* element = e_ptr->getElement(92);
+  double sum = 0;
+  for (auto& iso : element->getIsotopeDistribution()) {sum += iso.getIntensity();}
+  TEST_REAL_SIMILAR(sum, 1.30725795222315);
+
+  // new Uranium isotope added
+  TEST_EQUAL(e_ptr->getIsotope("(314)C") == nullptr, true)
+  int nr_isotopes = element->getIsotopes().size();
+  e_ptr->addIsotope("Uranium", "U", 92u, 0.3, 314.0, 1e5, Isotope::DecayMode::UNKNOWN, false);
```
cbielow, line 189: should adding a new isotope with a natural abundance only work if DecayMode == NONE?

cbielow (commented):
looks good to me.
Some comments though.
We better get this right the first time, since its a very central class

src/tests/class_tests/openms/source/ElementDB_test.cpp
hroest: yes, that would be a good idea

src/openms/source/CHEMISTRY/Element.cpp
hroest: yes `addIsotope` uses it, I guess you could trigger it with `setIsotopes(getIsotopes)` to achieve this?

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
hroest: `addElementToMaps_` cannot but `addIsotopeToMaps_` can, the reason is that we need to call a non-const function on the Element in order to update the isotopes

src/tests/class_tests/openms/source/ElementDB_test.cpp
hroest: actually thinking about it: I dont think so, you can clearly do MS measurements on molecules that have radioactive isotopes incorporated, such as Uranium and C14. There are samples that have a specific and well known composition of radioactive elements and maybe the user wants to model this. This is also where `addIsotope` would come in, if the user wants to adjust parameters to a sample they have at hand

src/pyOpenMS/tests/unittests/test000.py
hroest: moved back

src/tests/class_tests/openms/source/ElementDB_test.cpp
hroest: this was always possible in the past, I guess the question is whether we should allow users to do something that does not make sense. One issue could arise, as you update all isotopes with `addIsotope` you could end up in a situation where you are briefly over/below 100%, depending on the order in which you update the isotopes. I also think that any user who messes around with these isotopes (hopefully) knows what they are doing and does their own sanity checks?

cbielow: good point!
I would still argue that the check should be done by us.
What about a static function which creates an isotope, but does not add it to the DB. The user can then create a bunch of those and push them to the DB in one go.

src/tests/class_tests/openms/source/ElementDB_test.cpp
cbielow: ok, fair point. But then the documentation should not mention `natural` abundance.
https://github.com/OpenMS/OpenMS/pull/5956/files#diff-8b049d8c7ee5d2d85f2741157e4f6ae5cf91ff98c2247f000e5477f3f3389569R137
https://github.com/OpenMS/OpenMS/pull/5956/files#diff-db97d6be7dd38c6a135f55d80d9f03eb2bcea2d0577d5a449845a0f05821811eR51
https://github.com/OpenMS/OpenMS/pull/5956/files#diff-db97d6be7dd38c6a135f55d80d9f03eb2bcea2d0577d5a449845a0f05821811eR95

src/openms/include/OpenMS/CHEMISTRY/ElementDB.h
cbielow: Sorry, I still don't see it...
If you write `e = const_ele;` here https://github.com/OpenMS/OpenMS/pull/5956/files#diff-8ebe022fac5a9f3c0eb06f940c6706d8d9bb652131218ee474d495451af8627dR707, why would that not work?

src/openms/source/CHEMISTRY/Element.cpp
```diff
@@ -111,10 +114,34 @@ namespace OpenMS
 
   void Element::setIsotopeDistribution(const IsotopeDistribution & distribution)
   {
-    isotopes_ = distribution;
+    isotope_distr_ = distribution;
+    // TODO: remove this fxn
```
poshul, line 118: It looks at a glance like it's only used in `MetaProSIP` and `FeatureFInderAlgorithmPicked`. Should be pretty easy to get rid of

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -35,14 +35,17 @@
...
+#include <OpenMS/CHEMISTRY/Isotope.h>
 #include <OpenMS/DATASTRUCTURES/String.h>
 #include <OpenMS/CONCEPT/Exception.h>
-#include <OpenMS/CHEMISTRY/Element.h>
 #include <iostream>
 #include <cmath>
 
 using namespace std;
 
+constexpr int SECONDS_PER_YEAR = 31556952; // Gregorian year
```
poshul, line 47: Move to `Constants`

poshul (commented):
Other than the constant, this looks good to me.

poshul:
This looks ready to go, but should wait until the beginning of the next dev cycle.

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -640,53 +722,60 @@ namespace OpenMS
...
+      *iso = *isotope; // copy all data from input to the existing element
+      delete isotope;
+      return iso;
+    }
+    else
+    {
+      names_[iso_name] = isotope;
+      symbols_[iso_symbol] = isotope;
+      isotopes_[iso_symbol] = isotope;
+      isotopes_[iso_name] = isotope;
```
hroest, line 803: we should probably do the same things we do here as for the elements to make sure that the lookup still works

src/openms/source/CHEMISTRY/Element.cpp
```diff
@@ -80,10 +82,34 @@ namespace OpenMS
   {
     //force sortedness by mz. A lot of code relies on this.
     assert(std::is_sorted(distribution.begin(), distribution.end(), Peak1D::MZLess()));
-    isotopes_ = distribution;
+    isotope_distr_ = distribution;
+    // TODO: remove this fxn, the only way to change the isotope distribution should be through setIsotopes()
   }
```
timosachsenberg, line 89: remove now?

src/openms/source/CHEMISTRY/Element.cpp
```diff
@@ -80,10 +82,34 @@ namespace OpenMS
...
+  void Element::updateIsotopeDistr_()
+  {
+    auto dist = isotope_distr_.getContainer();
+    dist.clear();
+    dist.reserve(isotopes_.size());
+    for (const auto& isotope : isotopes_)
+    {
+      dist.emplace_back(isotope->getMonoWeight(), isotope->getAbundance());
+    }
+    isotope_distr_.set(std::move(dist));
```
timosachsenberg: do we need the assert is_sorted as used above?

src/tests/class_tests/openms/source/ElementDB_test.cpp
timosachsenberg: I agree that we should check for that and enforce 1 +- eps

timosachsenberg:
segfaults on mac

timosachsenberg:
```
checking void addIsotope(const std::string& name, const std::string& symbol, const unsigned int an, double abundance, double mass, double half_life, Isotope::DecayMode decay, bool replace_existing) ... 
DEBUG: Adding isotope Uranium with symbol U and atomic number 92, replace=1
DEBUG: Calculated iso_name=(238)Uranium, iso_symbol=(238)U
DEBUG: Checking if element exists with atomic number 92
DEBUG: Element with atomic number 92 exists
DEBUG: Isotope with symbol (238)U already exists
DEBUG: Creating new isotope to replace existing one
DEBUG: New isotope created successfully
DEBUG: Calling addIsotopeToMaps_ to replace existing isotope
DEBUG: addIsotopeToMaps_ called with iso_name=(238)Uranium, iso_symbol=(238)U
DEBUG: Found existing isotope in isotopes_ map
DEBUG: Existing isotope pointer: 0x600002a777b0
DEBUG: Selectively updating isotope properties (instead of using assignment operator)
DEBUG: Safe update completed, deleting new isotope instance
DEBUG: Returning existing isotope pointer
DEBUG: addIsotopeToMaps_ completed
DEBUG: Getting element for isotope
DEBUG: Isotope::getElement() called for  with atomic number 0
DEBUG: Isotope::getElement() returning pointer 0x0
DEBUG: Element pointer: 0x0
DEBUG: Getting isotopes from element
```
then segfault on mac

timosachsenberg (approved):
Should be fine to merge now

src/openms/include/OpenMS/CHEMISTRY/Isotope.h
```diff
@@ -0,0 +1,102 @@
...
+      expected to contain abundances while unstable isotopes generally do not
+      have natural abundances but contain information about half life and decay
+      mode.
+  */
+  class OPENMS_DLLAPI Isotope :
+    public Element
+  {
+    public:
+
+    enum DecayMode
```
cbielow, line 33:
```suggestion
    enum class DecayMode
```

src/openms/include/OpenMS/CHEMISTRY/Isotope.h
```diff
@@ -0,0 +1,102 @@
...
+    {
+      NONE = 0,     ///< No decay (stable isotope)
+      UNKNOWN,      ///< Unknown / Unspecified decay mode
+      ALPHA,        ///< Alpha decay
+      BETA_PLUS,    ///< Beta plus decay
+      BETA_MINUS,   ///< Beta minus decay
+      PROTON,       ///< Proton emission
```
cbielow, line 40: there are many more modes in theory.
Do we only need these?

src/openms/source/CHEMISTRY/Element.cpp
```diff
@@ -80,10 +83,47 @@ namespace OpenMS
...
+    dist.reserve(isotopes_.size());
+    
+    for (size_t i = 0; i < isotopes_.size(); i++)
+    {
+      const auto& isotope = isotopes_[i];
+      if (isotope == nullptr)
+      {
+        OPENMS_LOG_WARN << "Null isotope pointer at index " << i << std::endl;
+        continue;  // Skip null pointers to avoid crash
+      }
```
cbielow, line 114: this should never happen, right?
If it does, something is very fishy and warrants an exception

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -101,6 +119,62 @@ namespace OpenMS
...
+      const Isotope* isotope = new Isotope(iso_name, iso_symbol, an, mass_number - an, mass, abundance, half_life, decay);      
+      addIsotopeToMaps_(iso_name, iso_symbol, isotope);
+      
+      // we changed/updated the isotopes, so we need to update
+      const Element* const_ele = isotope->getElement();
+      
+      Element* element = const_cast<Element*>(const_ele);
+      auto current_isotopes = element->getIsotopes();
+      
+      element->setIsotopes(element->getIsotopes());
```
cbielow: wow, this is hacky.
How about a member `refreshIsotopeCache()`, to at least show the intention.

src/openms/source/CHEMISTRY/ElementDB.cpp
```diff
@@ -101,6 +119,62 @@ namespace OpenMS
...
+      isotope = addIsotopeToMaps_(iso_name, iso_symbol, isotope);
+  
+      // we added a new isotope, so we need to add it to the element
+      const Element* const_ele = isotope->getElement();
+      
+      Element* element = const_cast<Element*>(const_ele);
+      auto iso_list = element->getIsotopes();
+      
+      iso_list.push_back(isotope);
+      element->setIsotopes(iso_list);
```
cbielow, line 172: please at least document that this also updates the internal cache

src/tests/class_tests/openms/source/Element_test.cpp
```diff
@@ -99,6 +99,20 @@ START_SECTION(void setAverageWeight(double weight))
...
 END_SECTION
 
+START_SECTION(( void setIsotopes(const std::vector<const Isotope*>& isotopes) ))
+  std::vector<const Isotope*> isotopes = {new Isotope(), new Isotope()};
+  e_ptr->setIsotopes(isotopes);
+  // Clean up after the test
+  for (auto iso : isotopes)
+  {
+    delete iso;
+  }
```
cbielow: check that the cache is working as intended?
This test does not really test anything.

src/tests/class_tests/openms/source/Element_test.cpp
```diff
@@ -112,6 +126,10 @@ START_SECTION(double getMonoWeight() const)
 	TEST_REAL_SIMILAR(e_ptr->getMonoWeight(), 2.333)
 END_SECTION
 
+START_SECTION(virtual bool isIsotope())
+  TEST_EQUAL( e_ptr->isIsotope(), false )
```
cbielow, line 153: add a case for an isotope?

src/tests/class_tests/openms/source/Isotope_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+            double half_life,
+            Isotope::DecayMode dm)))
+	e_ptr = new Isotope(name, symbol, atomic_number, 10u, mono_weight, 0.6, 42, Isotope::DecayMode::UNKNOWN);	
+	TEST_NOT_EQUAL(e_ptr, e_nullPointer)
+END_SECTION
+
+START_SECTION( const Element* getElement() const)
+  const Element* el = e_ptr->getElement();
+  TEST_NOT_EQUAL(el, e_nullPointer)
+  TEST_EQUAL(el->getSymbol(), "P")
```
cbielow, line 63: this is confusing...
Why not make up a new (imaginary) isotope of "P"?

src/tests/class_tests/openms/source/Isotope_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+START_SECTION( virtual bool isIsotope() )
+  TEST_EQUAL(e_ptr->isIsotope(), true)
+END_SECTION
+
+START_SECTION(bool isStable() const)
+  TEST_EQUAL(e_ptr->isStable(), false)
+END_SECTION
+
+START_SECTION(Isotope& operator = (const Isotope& Isotope))
+	Isotope e = *e_ptr;
```
cbielow: you are not testing `operator=` here. You are testing the Ctor.
Use
```
 e = *e_ptr;
```
on an existing `e`

timosachsenberg:
diverged too far... was not able to rebase it

PR #5999, opened 2022-03-24
CI warning fixes for pr-5819

lennykovac:
# Description
WarmUp PR
pr-5819-macOS_10.14.6_x64-appleclang-11.0.0-Release-x64
/.../source/src/openms/source/ANALYSIS/OPENSWATH/TransitionPQPFile.cpp:61:14: warning: unused function 'callback' [-Wunused-function]
 This function never got called, so I removed it. It just printed the some information on th e

src/openms/include/OpenMS/KERNEL/RangeUtils.h:187:23: note: place parentheses around the ^ expression to evaluate it first

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -58,6 +58,7 @@ the authors tag in the respective file header.
  - Khue Nguyen
  - Knut Reinert
  - Lars Nilse
+ - Lenny Kovac
  - Leon Bichmann
  - Leon Kuchenbecker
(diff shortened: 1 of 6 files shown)


timosachsenberg:
thanks!

jpfeuffer:
So what's the deal with sonar and pos_min? Do we still get warning for them? Can we suppress these?

PR #6009, opened 2022-03-29
[FEATURE] package type stubs with pyopenms

### LINKED ISSUE #5821: Package new type stubs with pyopenms
- We have to somehow include the new pyi files in the pyopenms package. Either with setup.py package_data and/or MANIFEST.
- We have to add an empty py.typed file and also package it

See https://github.com/OpenMS/autowrap/pull/119/files#diff-dca49b38da57dcbac2184e1872b2df76946696a7a05ec6fddba412224adefef9R52-R62

- Then test it with e.g. PyCharm

jpfeuffer:
# Description

Fixes #5821

I am just having some problems with internal imports. I think in theory we somehow need to import each submodule in each submodule, such that the type engine finds all types from sibling modules. Maybe we need a __init__.pyi?
I tried that on PyCharm but it did not completely solve it. Trying mypy would be another TODO.


jpfeuffer:
With this, I hope people will try it out.

PR #6021, opened 2022-04-05
Fix/linear transform

### LINKED ISSUE #5978: FeatureLinkerUnlabeledKD produces empty consensus
I have two datasets that were run using the same method. The Map aligner works fine, all the preprocessing steps are great, but at the re-quantification step (FFmetaboIdent), even though each featureXML file generated is not empty, FeatureLinkerUnlabeledKD produces an empty consensus file. Why? When can this happen?

enetz:
# Description

Least-squares fitting with `Wm5::HeightLineFit2` can fail with two points that are too close to each other.
Added code to handle this specific case manually, since there is an exact solution with two points.

Fixes #5978

# Checklist:

# Note:

### DIFF
--- a/src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
+++ b/src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
@@ -60,7 +60,7 @@ namespace OpenMS
       symmetric_ = params_.getValue("symmetric_regression") == "true";
       // weight the data (if weighting is specified)
       TransformationModel::DataPoints data_weighted = data;
-      if ((params.exists("x_weight") && params.getValue("x_weight") != "x") 
+      if ((params.exists("x_weight") && params.getValue("x_weight") != "x")
        || (params.exists("y_weight") && params.getValue("y_weight") != "y"))
       {
         weightData(data_weighted);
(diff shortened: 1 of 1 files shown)


src/openms/source/ANALYSIS/MAPMATCHING/TransformationModelLinear.cpp
```diff
@@ -84,9 +84,22 @@ namespace OpenMS
         {
           points.push_back(Wm5::Vector2d(data_weighted[i].first, data_weighted[i].second));
         }
-        if (!Wm5::HeightLineFit2<double>(static_cast<int>(size), &points.front(), slope_, intercept_))
+        if (size == 2)
+        {
+          if (!Wm5::HeightLineFit2<double>(static_cast<int>(size), &points.front(), slope_, intercept_))
```
cbielow, line 89: why try fitting in the first place? Wouldn't computing the exact solution be enough?

enetz: yes, you are right. I will do the exact calculation directly and add another exception in case the slope or intercept are nan or inf

timosachsenberg (approved):
looks good... not sure though if the quotient could get super large and is useful ;)

enetz:
failing pyOpenMS test is unrelated

PR #6023, opened 2022-04-05
[Feature] Added Metabolite Isotope Calculation to TSGDialog of TOPPView

Waschi97:
# Description

`TheoreticalSpectrumGenerationDialog` now supports metabolite input for which it calculates the isotope pattern.

GUI looks like this for the new input mode:

TSGDialog_Metabolite

I also introduce a struct which stores the important information about an ion check box. Then I use an array of these to access the information for each ion. Previously multiple vectors were used for this.

# Checklist:

# Note:


src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -379,4 +418,48 @@ namespace OpenMS
...
+    else if (seq_type_ == "RNA")
+      input_type = 1;
+    else // Metabolite
+      input_type = 2;
+
+    for (size_t i = 0; i < check_boxes_.size(); ++i)
+    {
+      const CheckBox_* curr_box = &check_boxes_.at(i);
+
+      bool hidden = false;
```
cbielow, line 436: `bool hidden = (curr_box->state.at(input_type) == State::HIDDEN);`

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -95,16 +113,16 @@ namespace OpenMS
...
       {
         ui_->ion_types->item(i)->setCheckState(Qt::Checked);
-        continue;
       }
-      ui_->ion_types->item(i)->setCheckState(Qt::Unchecked);
+      else
+      {
+        ui_->ion_types->item(i)->setCheckState(Qt::Unchecked);
+      }
     }
```
cbielow, line 126: should this be part of updating the UI when switching between types?

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -126,34 +144,39 @@ namespace OpenMS
   {
     Param p;
 
-    bool peptide_input = ui_->seq_type->currentText() == "Peptide";
-
-    // add check boxes to parameters, i.e. ion types
-    for (size_t i = 0; i < int(CheckBox::NUMBER_OF_CHECK_BOXES); ++i)
+    if (seq_type_ != "Metabolite") // no ions for metabolite input
```
cbielow, line 147: using a string seems dangerous.
If someone changes the content of the dropdown, everything breaks.
Why not make this an enum?!

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
```diff
@@ -95,16 +113,16 @@ namespace OpenMS
     // select peptide sequence by default and update interface
     ui_->seq_type->setCurrentText("Peptide");
     seqTypeSwitch_();
```
cbielow, line 115: will this be called twice (because you call `setCurrentText()`) which is connected to seqTypeSwitch_?!

cbielow (changes requested):
Can you change
`Enter sequence` to `Enter empirical formula (e.g. ....)`
for metabolites?

and (for all)
`Max. Isotope Probability %` to `Show isotopes with probability (%) higher than: `

timosachsenberg:
wow! looks nice!

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
Waschi97: Probably not.. Since all other settings are preserved when switching between sequence types, changing the checked ions feels inconsistent.

src/openms_gui/source/VISUAL/DIALOGS/TheoreticalSpectrumGenerationDialog.cpp
Waschi97: That's what I also would've expected but I think the signal-slot connection only work after object construction. Therefore `seqTypeSwitch_` needs to be called here.

Waschi97: The slot `currentTextChanged` only gets called if the text actually changes. That's why calling the slot here is necessary. However I removed setting the current text since it is not really needed. (The ui-default for this combo box is already 'Peptide')

timosachsenberg:
https://cdash.openms.de/viewBuildError.php?buildid=27543

PR #6029, opened 2022-04-07
[FIX] BasicStatistics: reasonable results with vectors of zeroes

### LINKED ISSUE #5981: MapAlignerPoseClustering FATAL: uncaught exception!
This algorithm crashes from the develop-version with a specific dataset, but in TOPPAS there seems to be no problem at all (2.7.0). The error is :
```

Picking a reference (by size) ... done
Progress of 'Aligning input maps':
  20.44 %               
FATAL: uncaught exception!
exception of type InvalidValue occurred in line 1040, function virtual void OpenMS::PoseClusteringAffineSuperimposer::run(const std::vector<OpenMS::Peak2D>&, const std::vector<OpenMS::Peak2D>&, OpenMS::TransformationDescription&) of /home/eeko/data/OpenMS_build/OpenMS/src/openms/source/ANALYSIS/MAPMATCHING/PoseClusteringAffineSuperimposer.cpp
error message: the value '-nan' was used but is not valid; Superimposer could not compute an initial transformation!You can try to increase 'max_num_peaks_considered' to solve this.
Aborted
```

My parameters:

`MapAlignerPoseClustering -algorithm:max_num_peaks_considered -1 -algorithm:superimposer:mz_pair_max_distance 0.05 -algorithm:pairfinder:distance_MZ:max_difference 10.0 -algorithm:pairfinder:distance_MZ:unit ppm `

Exactly the same params were used in TOPPAS.

enetz:
`BasicStatistics` `mean_` and `variance_` returned `-nan` when given a vector of only zeroes, due to dividing by the sum.

This should fix #5981 and possibly a few other unexpected `nan` issues.

This is one possible solution, but more input would be appreciated.
Problems here are:
`sum_` and `mean_` would be zero for a vector like this as well: [-5, -3, 0, 3, 5], but the variance would not be zero.
Still we use `variance_ /= sum_` as a step in the calculation, which would cause a `nan`.

# Description

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/MATH/STATISTICS/BasicStatistics.h
+++ b/src/openms/include/OpenMS/MATH/STATISTICS/BasicStatistics.h
@@ -132,6 +132,12 @@ public:
           variance_ += *iter * diff;
         }
         variance_ /= sum_;
+
+        if (sum_ == 0 && (std::isnan(mean_) || std::isinf(mean_)) )
+        {
+          mean_ = 0;
+          variance_ = 0;
+        }
       }
 
       /// This does the actual calculation.
@@ -310,4 +316,3 @@ private:
   }   // namespace Math
 
 } // namespace OpenMS
-
--- a/src/tests/class_tests/openms/source/BasicStatistics_test.cpp
+++ b/src/tests/class_tests/openms/source/BasicStatistics_test.cpp
@@ -1,32 +1,32 @@
 // --------------------------------------------------------------------------
-//                   OpenMS -- Open-Source Mass Spectrometry               
+//                   OpenMS -- Open-Source Mass Spectrometry
 // --------------------------------------------------------------------------
 // Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
 // ETH Zurich, and Freie Universitaet Berlin 2002-2021.
-// 
+//
 // This software is released under a three-clause BSD license:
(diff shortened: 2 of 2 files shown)


src/openms/include/OpenMS/MATH/STATISTICS/BasicStatistics.h
```diff
@@ -132,6 +132,12 @@ namespace OpenMS
           variance_ += *iter * diff;
         }
         variance_ /= sum_;
+
+        if (sum_ == 0 && (std::isnan(mean_) || std::isinf(mean_)) )
```
timosachsenberg, line 136: there are two places where this could be added in the header

timosachsenberg: why is the sum_ == 0 required?
would std::isnan(mean_) || std::isinf(mean_) be enough?

enetz: I wanted to avoid fixing stuff that shouldn't be fixed. If those numbers are `nan` because there is a `nan` in the given input, then returning a `0` would be wrong. We should make sure we only set to `0` when it's obvious the `nan` was caused by the calculations in this class.

timosachsenberg:
rebuild jenkins

PR #6036, opened 2022-04-12
[BUILD] Use Cmake exports from boost

jpfeuffer:
# Description

Try to use the official exports from Boost Cmake instead of the find module.


timosachsenberg:
is this to prevent cmake picking up the wrong boost?

jpfeuffer:
No, just picking any boost. It does not work right now.

PR #6039, opened 2022-04-13
Test TOPPView Preferences Dialog

Waschi97:
# Description

Added a test for `TOPPViewPrefDialog`.
Tested is:
- if all GUI elements are constructed
- if the settings are correctly interpreted and exported
- (for some GUI elements) if they are interactable and behave expected

The GUI parts that are not tested are all instances of:
`Qt::ComboBox`, `OpenMS::MultiGradientSelector` or `OpenMS::ParamEditor`

Interacting with these objects with `QTest::mouseClick` proved to be quite difficult since the mouse position needs to be 100 % correct. I was unable to get the needed positions, so for now their interface is untested. (However, the correspondence between them and their depicted parameter is tested.)

I also added a label to the 'Theoretical Spectra Generation' tab of that dialog that reads:
"These settings are used to generate theoretical spectra of peptide identifications to align them and use them for spectra annotation. (i.e. "Tools" -> "Annotate with peptide identifications")"

# Note:


src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
```diff
@@ -0,0 +1,352 @@
...
+  // test color selectors
+  QTimer::singleShot(DELAY, this, &TestTVPrefDialog::checkColorDialog_);
+  QTest::mouseClick(UI->color_1D, Qt::LeftButton, 0, QPoint());
+  QTest::qWait(DELAY);
+
+  QTimer::singleShot(DELAY, this, &TestTVPrefDialog::checkColorDialog_);
+  QTest::mouseClick(UI->selected_1D, Qt::LeftButton, 0, QPoint());
+  QTest::qWait(DELAY);
+
+  QTimer::singleShot(DELAY, this, &TestTVPrefDialog::checkColorDialog_);
```
cbielow, line 165: Why do you need to create a timed event here?
Could you not also 
1) click
2) checkColorDialog_ 
??

And maybe providing the widget to test as a parameter would be more explicit (on first sight, the code looks like it tests the same thing three times, but it does not...)

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
```diff
@@ -0,0 +1,352 @@
...
+  QTest::mouseClick(tab_bar, Qt::LeftButton, 0, tab_bar->tabRect(TWO_D).center());
+  QTest::qWait(DELAY);
+
+  // test spin box
+  UI->feature_icon_size_2D->clear();
+  QTest::keyClicks(UI->feature_icon_size_2D, "5");
+  QTest::qWait(DELAY);
+  QVERIFY(5 == UI->feature_icon_size_2D->value());
+
+  // TODO:: 'peak_2D' (OpenMS::MultiGradientSelector), 'mapping_2D' (Qt::ComboBox), 'feature_icon_2D' (Qt::ComboBox)
```
cbielow, line 183: annotate WHY its not tested now

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
```diff
@@ -0,0 +1,352 @@
...
+  //               'TSG' tab                 //
+  /////////////////////////////////////////////
+
+  // click on tab
+  QTest::mouseClick(tab_bar, Qt::LeftButton, 0, tab_bar->tabRect(TSG).center());
+  QTest::qWait(DELAY);
+
+  // test spin box
+  UI->tolerance->clear();
+  QTest::keyClicks(UI->tolerance, "0,5");
```
cbielow, line 211: `0,5` may not work on all locales?!
I'm surprised it works now honestly, because we should be using a c-locale by default and I would have expected `0.5` to work.

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+  QVERIFY2(PARAM("2d:dot:gradient") == "Linear|0,#ffaa00;6,#ff0000;14,#aa00ff;23,#5500ff;100,#000000", "'2D peak gradient' param value not exported correctly.");
+  QVERIFY2(PARAM("2d:mapping_of_mz_to") == "x_axis", "'2D mapping of mz to' param value not exported correctly.");
+  QVERIFY2(PARAM("2d:dot:feature_icon") == "circle", "'2D feature icon' param value not exported correctly.");
+  QVERIFY2(int(PARAM("2d:dot:feature_icon_size")) == 5, "'2D feature icon size' param value not exported correctly.");
+
+  QVERIFY2(PARAM("3d:dot:gradient") == "Linear|0,#ffea00;6,#ffaa00;100,#ff0000", "'3D peak gradient' param value not exported correctly.");
+  QVERIFY2(int(PARAM("3d:dot:shade_mode")) == 1, "'3D shade mode' param value not exported correctly.");
+  QVERIFY2(int(PARAM("3d:dot:line_width")) == 2, "'3D line width' param value not exported correctly.");
+
```
cbielow, line 302: (you are going to hate this): technically you need to test two completely orthogonal sets of values, just to make sure the Dialog does not always return the same value for some widget (despite setting another value in the GUI), and you happen to test exactly for that value.

cbielow (changes requested):
looks good to me overall.
Some minor stuff to discuss though :)

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
```diff
@@ -0,0 +1,352 @@
...
+  QVERIFY2(UI->map_default, "'Default map view' combo box not created.");
+  QVERIFY2(UI->map_cutoff, "'Low intensity cutoff' combo box not created.");
+  QVERIFY2(UI->on_file_change, "'Action when file changes' combo box not created.");
+  QVERIFY2(UI->use_cached_ms1, "'Cache MS1 spectra to disk' check box not created.");
+  QVERIFY2(UI->use_cached_ms2, "'Cache MS2 spectra to disk' check box not created.");
+  QVERIFY2(UI->label_5, "'Default path:' label not created.");
+  QVERIFY2(UI->label_8, "'Default map view:' label not created.");
+  QVERIFY2(UI->label_9, "'Low intensity cutoff:' label not created.");
+  QVERIFY2(UI->label_14, "'Action when file changes:' label not created.");
+  QVERIFY2(UI->label, "'Caching Strategy:' label not created.");
```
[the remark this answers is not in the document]
Waschi97: sure, on it

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
```diff
@@ -0,0 +1,352 @@
...
+// higher values together with 'dialog_.show();' can be useful for debugging this test
+constexpr int DELAY {5};
+
+void testCheckBox(QCheckBox* cb)
+{
+  Qt::CheckState prev_state = cb->checkState();
+  QTest::mouseClick(cb, Qt::LeftButton, 0, cb->rect().bottomLeft());
+  QTest::qWait(DELAY);
+  QVERIFY2(prev_state != cb->checkState(), qPrintable("'" + cb->objectName() + "' check box didn't change its state by clicking on it."));
+}
```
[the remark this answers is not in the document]
Waschi97: I think it would... But maybe in another PR?

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
Waschi97: This is needed because when a dialog is opened the tests stops being executed until the dialog is closed.

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
Waschi97: `0,5` is the keyboard input for the spin box - it doesn't support typing a `.`
I'm unsure if this differs depending on system language...

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
Waschi97: hmmm I see...
Well, at least that's easy to do.

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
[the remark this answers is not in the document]
cbielow: sure

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
cbielow: ah,yes. 
Please add a comment, so other people know why its done this way without finding out the hard way

src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp
cbielow: could you give that some google/bing/duckduckgo time?

Waschi97: already on it 👍

Waschi97:
rebuild jenkins

cbielow:
I think this was good to go?

Waschi97:
It is/was. But for some reason CI kept failing. Let's see if it is still the case...

jpfeuffer:
@cbielow you also need to change your review status

jpfeuffer:
https://cdash.openms.de/viewBuildError.php?buildid=65196

Valid reason to fail

cbielow:
the build logs are deleted already.
I'll trigger another build
@Waschi97 could you have a look?

cbielow:
rebuild jenkins

cbielow (approved):

Waschi97:
Yeah, I was able to catch a look at the logs last week. It's just a naming issue. Just didn't have the time to fix it yet.

Waschi97:
rebuild jenkins

Waschi97:
rebuild jenkins

jpfeuffer:
rebuild jenkins

Waschi97:
Test seems to work on Linux.
Locally it also works fine on Windows.

For MacOS I think:
https://github.com/OpenMS/OpenMS/blob/0f1ac903d8faee2dcf38405caf7562e3d1317be9/src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp#L407-L408
For some reason doesn't close the dialog that opens during the test which results in a time out.

jpfeuffer:
Why would the Enter key close anything anyhow?

Waschi97:
Ohh ignore the comment on that line. Shouldn't have to do anything with the problem.

But to eleborate that comment:
A file dialog is opened by the test. Usually I / you can close such a dialog by pressing 'Enter'. For example you can manually press 'Enter' during the test to close that dialog, when the first, simulated 'Enter' didn't close the dialog.

jpfeuffer:
Maybe that assumption is not correct. I bet something (like the Exit button) needs to be in focus/selected.
Where is it documented that you can close Qt windows with Enter?

Waschi97:
Well, that might be the case. But the button is or should be in focus as you literally don't have to do anything other than press 'Enter' to close it.
For the QColorDialog the simulated 'Enter' works:
https://github.com/OpenMS/OpenMS/blob/0f1ac903d8faee2dcf38405caf7562e3d1317be9/src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp#L420-L421

Waschi97:
But all that shouldn't matter for the test failing on MacOS.
As `fd->close()` is the function that apparently doesn't work on MacOS. Which is a general member function of all QWidgets:
https://doc.qt.io/qt-6/qwidget.html#close

cbielow:
it seems that when the dialog is opened via https://doc.qt.io/qt-6/qfiledialog.html#getExistingDirectory on MacOS the native dialog is shown. And that may not react to `.close()`.
So you need to force it to use the QT dialog -- see above link.
This needs fixing for all (two?) file browser dialogs.

poshul:
@cbielow is this going to effect https://github.com/OpenMS/OpenMS/pull/6139? If not, can we go ahead and merge it?

cbielow:
the tests on MacOS still fail AFAIKS
https://cdash.seqan.de/testSummary.php?project=2&name=TVPrefDialog_test&date=2022-10-19

But once its fixed, no side effects on #6139 expected.

Waschi97:
Yeah, currently trying to figure out MacOS.

I would also not expect any effects on #6139.

cbielow:
could someone with a MacOS have a look here?
Figuring out GUI testing using reinforcement learning by CI is not very fun I would imagine :)

Waschi97:
Yes, it would be great if someone with MacOS could take a look. This would save a lot of time!

For debugging it can be helpful to increase the delay between simulated clicks:
https://github.com/Waschi97/OpenMS/blob/cf607b62a9358998a1973a47c70ba52070559965/src/tests/class_tests/openms_gui/source/GUI/TVPrefDialog_test.cpp#L43

It may also help to add a `dialog_.show()` in the contructor of `TestTVPrefDialog` in "src/tests/class_tests/openms_gui/include/GUI/TVPrefDialog_test.h"

timosachsenberg:
feel free to revive

PR #6046, opened 2022-04-18
[FIX] Suppress -Wunused-private-field warnings 

matchy233:
# Description

This PR (partly) fixes `-Wunused-private-field` warnings when compiling with `clang` on macOS.

There are two unused private fields:
1. `snoar_` in `OpenSwathOSWWriter.h`
2. `pos_min_` of `SplineInterpolatedPeaks::Navigator` in `SplineInterpolatedPeaks.h`

Both unused fields can be removed since they were not used anywhere in the code.

However, I feel that the first unused field should not be trivially removed since in `OpenSwathTSVWriter`, we utilized `sonar_` to determine whether we should write `var_sonar_<options>`  in the output or not:

```c++
if (sonar_)
{
  ofs << "\tvar_sonar_lag\tvar_sonar_shape\tvar_sonar_log_sn\tvar_sonar_log_diff\tvar_sonar_log_trend\tvar_sonar_rsq";
}
```
But in OSW writer, we append `VAR_SONAR<options>` to `sql_feature_ms2` regardless of `sonar_` provided or not. I'm curious whether we should add an `if` statement checking the value of `sonar_` like in TSV writer. (Hence `sonar_` is suppressed only using `[[maybe_unused]]`)

# Checklist:

# Note:

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
@@ -119,7 +119,7 @@ namespace OpenMS
     OpenMS::UInt64 run_id_;
     bool doWrite_;
     bool use_ms1_traces_;
-    [[maybe_unused]] bool sonar_;
+    bool sonar_;
     bool enable_uis_scoring_;
 
   public:
--- a/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
+++ b/src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathOSWWriter.h
@@ -119,7 +119,7 @@ namespace OpenMS
     OpenMS::UInt64 run_id_;
     bool doWrite_;
     bool use_ms1_traces_;
-    bool sonar_;
+    [[maybe_unused]] bool sonar_;
     bool enable_uis_scoring_;
 
   public:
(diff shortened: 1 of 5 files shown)


jpfeuffer:
It feels like you are right and an if-case in the writer would be better.

cbielow:
dumb question from my side: why is `sonar_` flagged as unused? It is clearly used in `OpenSwathTSVWriter.cpp` ...
Is this simply a false positive from Clang?

matchy233:
@cbielow sorry for not making it clear. OSW writer and TSV writer both have `sonar_` as a private flag. It's only unused in OSW writer. I mentioned TSV writer for comparison. :)

cbielow:
ah, I see.
thanks for the heads up

matchy233:
@jpfeuffer @timosachsenberg I've updated the PR with `if (sonar_)` branch included. Would be nice to have the latest changes reviewed!

jpfeuffer:
Looks good to me but maybe a swath person can have a look.

jcharkow (approved):
From the SWATH side these changes look good to me. However have not worked with SONAR data in the past.

jpfeuffer:
@matchy233 python bindings need to be updated still

timosachsenberg:
I can do a quick edit

matchy233:
Is this available to merge? @timosachsenberg  @jpfeuffer I think timo has already made the changes to the Python binding

jpfeuffer:
Unfortunately I don't see it. It is not in this PR, where it should be.

matchy233:
Hmm.. I probably have overwritten his commit to this PR by the force push
after I rebased my local branch to develop... I modified and made the
changes in the newest push.

Julianus Pfeuffer ***@***.***> 于2022年5月4日周三 17:25写道：

jpfeuffer:
Looks good. Will merge after CI.

PR #6053, opened 2022-04-20
[FIX] DecoyDatabase detects already decoy'd DBs

### LINKED ISSUE #4917: DecoyDatabase should detect already decoy'd DBs
our example FASTA files in `share` sometimes contain outdated Decoy symbols (e.g. `_rev` as suffix).
If one is not careful, this can lead to subtle errors which are a bit tricky to find when a downstream tool crashes.

At least our DecoyDatabase tool should check if its consuming a decoy'ed DB (the above example is currently happily reversed **again**).

This is already mentioned in  #4662, but I think its very important to fix and to prevent accidental errors with high potential for frustration.

hbeschorner:
# Description

Added check, so DecoyDatabase will terminate if input data is already decoy’d.
Also added tests for the function.

Fixes #4917

# Checklist:

# Note:


cbielow:
I'm biased for review because these are my students.
Review by @jpfeuffer or @timosachsenberg maybe? :)

src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
```diff
@@ -360,141 +372,150 @@ class DecoyHelper
...
-  
+
+  /**
+  @brief Function to count the occurrences of decoy strings in a given set of protein names
+
+  For tested decoy strings see DecoyHelper::affixes.
+  Returns all data needed for interpretation (see DecoyHelper::DecoyStatistics).
+  */
+  template<typename T>
+  static DecoyStatistics countDecoys(FASTAContainer<T>& proteins)
```
cbielow, line 456: one thing I just realized though: a class test for this function would be nice :)
add it here: https://github.com/cbielow/OpenMS/blob/develop/src/tests/class_tests/openms/source/FASTAContainer_test.cpp#L61

timosachsenberg (approved):
Nice first PR. Looks very clean.
As chris pointed out, a class test for countDecoys would be nice.

jpfeuffer:
How about allowing other decoy strings with the force parameter. If you want, undocumented. Just in case someone, even one of us, suddenly has a weird requirement. E.g. for entrapment decoys, or if you want multiple decoys for one peptide to make the FDR calculations more stable.

timosachsenberg:
But would you run the tool several times? Do you have a use care in mind where this would work?

jpfeuffer:
No, the tool is too inflexible for this anyway. Forcing a set of strings makes it even worse imho.

jpfeuffer:
If this is a needed in a pipeline for example, the development of the pipeline stalls for 6+ months until the next openms release.

timosachsenberg:
ah you mean just skipping the test if force flag is provided? Yeah that makes sense

src/openms/include/OpenMS/DATASTRUCTURES/FASTAContainer.h
vrossow: Class tests for countDecoys and and findDecoyString are now added.

hbeschorner:
Added force flag, so all custom decoy strings should be possible (warning if the custom decoy string will not be detected in check for decoys)

jpfeuffer (approved):

Issue #5653, opened 2021-11-12
Duplicated transition IDs in TargetedExperiment object from PQP source, caused by decoy peptides with same sequence but different gene


gureann:
Hi @hroest ,

I'm using OpenSwath to analyze DIA data acquired from QEHF, and an error occurred when I ran OSW with generated pqp file. 
The error was caused by some decoy peptides which got same sequence from different target peptide sequences (also belonged to different genes), and this lead to duplicated transition IDs at the pqp reading step for aggregation of gene table

If the input mzml was converted from thermo raw file by msconvert without peak picking, there will be no exception raised, and just stopped when searching, like this
```
Thread 3_0 will analyze 12911 compounds and 72038 transitions from SWATH 5 (batch 2 out of 12)
Thread 6_0 will analyze 14914 compounds and 82015 transitions from SWATH 4 (batch 2 out of 14)
Thread 0_0 will analyze 10545 compounds and 55144 transitions from SWATH 2 (batch 3 out of 10)
Thread 2_0 will analyze 14011 compounds and 74796 transitions from SWATH 3 (batch 3 out of 14)
(base) PS F:\>
```

When the input mzml was converted with peak picking, the error will be invalid ID
```
...
Thread 0_0 will analyze 10545 compounds and 55144 transitions from SWATH 2 (batch 3 out of 10)
Thread 4_0 will analyze 7360 compounds and 37058 transitions from SWATH 1 (batch 4 out of 7)

---------------------------------------------------
FATAL: uncaught exception!
---------------------------------------------------
last entry in the exception handler:
exception of type InvalidValue occurred in line 165, function void __cdecl OpenMS::MRMTransitionGroup<class OpenMS::MSChromatogram,struct OpenSwath::LightTransition>::addTransition(const struct OpenSwath::LightTransition &,const class OpenMS::String &) of C:\jenkins\ws\openms\RC\openms_release_packaging\9447518b\source\src\openms\include\OpenMS/KERNEL/MRMTransitionGroup.h
error message: the value '1747343' was used but is not valid; Internal error: Transition with nativeID was already present!
---------------------------------------------------
```

If I use TargetedFileConverter again, from pqp to tsv, the error will be raised correctly, in the checking step after reading database and generating TargetedExperiment
```
Progress of 'reading PQP file (SQL warmup)':
-- done [took 36.64 s (CPU), 36.82 s (Wall)] --
Progress of 'reading PQP file':
[...]

PR #6054, opened 2022-04-20
OpenSwathDecoyGenerator remove duplicates

jcharkow:
# Description

Addressing issue https://github.com/OpenMS/OpenMS/issues/5653

This commit adds a check to OpenSwathDecoyGenerator to ensure that a decoy sequence is not generated if it is the same as a target sequence or another decoy sequence. 

This check is done with modifications meaning that the same peptide sequence with different modifications can still have unique decoys generated

Tests are added and tests are changed to reflect the new behaviour. 

# Checklist:
- [ X ] Updated or added python bindings for changed or new classes. (Tick if no updates were necessary.)

# Note:


src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMDecoy.h
```diff
@@ -128,6 +128,14 @@ namespace OpenMS
                         const bool enable_unspecific_losses,
                         const int round_decPow = -4) const;
 
+    /**
+       @brief Swtich the final Amino Acid of a tryptic peptide.
```
timosachsenberg, line 132:
```suggestion
       @brief Switch the final Amino Acid of a tryptic peptide.
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMDecoy.h
```diff
@@ -128,6 +128,14 @@ namespace OpenMS
...
+       E.g. If the last Amino Acid is "K" switch to "R" (and visa versa).
```
timosachsenberg, line 133:
```suggestion
       E.g. If the last Amino Acid is "K" switch to "R" (and vice versa).
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMDecoy.h
```diff
@@ -216,6 +224,11 @@ namespace OpenMS
     OpenMS::TargetedExperiment::Peptide reversePeptide_(
       const OpenMS::TargetedExperiment::Peptide& peptide) const;
 
+    /**
+     @breif Convert a peptide to a string which contains the peptide sequence and modifications
```
timosachsenberg, line 228:
```suggestion
     @brief Convert a peptide to a string which contains the peptide sequence and modifications
```

timosachsenberg: can you also document in which format it is? is this the default AASequence we use in OpenMS

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -379,6 +378,27 @@ namespace OpenMS
     return false;
   }
 
+  String MRMDecoy::getModifiedPeptideSequence_(const OpenMS::TargetedExperiment::Peptide& pep) const
+  {
+          String full_peptide_name = "";
```
timosachsenberg, line 383: use two spaces as indentation

timosachsenberg, line 383:
```suggestion
          String full_peptide_name;
```

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -379,6 +378,27 @@ namespace OpenMS
...
+          for (int loc = -1; loc <= (int)pep.sequence.size(); loc++)
+          {
+            if (loc > -1 && loc < (int)pep.sequence.size())
+            {
+              full_peptide_name += pep.sequence[loc];
+            }
+            // C-terminal and N-terminal modifications may be at positions -1 or pep.sequence
+            for (Size modloc = 0; modloc < pep.mods.size(); modloc++)
```
timosachsenberg, line 391: Having this inner loop is potentially very inefficient. How often is this function called?

timosachsenberg (approved):
- some minor formatting / typos
- potential inefficent generation of modified strings

timosachsenberg:
Btw you can add suggestions to a batch so that only one commit is created. Each commit starts a CI run of which we only have a limited amount per month.

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
jcharkow: This is called once per peptide sequence so it might be called a lot. However the inner loop should only be called maybe maximum of 3 times because theoretically there are not too many modifications on a peptide. This code block is actually taken from OpenMS tsv writer. https://github.com/OpenMS/OpenMS/blob/c2555c3278de34a5f261df9a0915f42a8bfada3c/src/openms/source/ANALYSIS/OPENSWATH/OpenSwathTSVWriter.cpp#L145-L160
however this is probably not the most efficient way. Basically here I am just looking for a data structure which can hold modified peptide sequences to check for equality.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/MRMDecoy.h
jcharkow: > Btw you can add suggestions to a batch so that only one commit is created. Each commit starts a CI run of which we only have a limited amount per month.

Sorry and thanks for letting me know. I will do that next time.

jcharkow:
> Btw you can add suggestions to a batch so that only one commit is created. Each commit starts a CI run of which we only have a limited amount per month.

Sorry and thanks for letting me know. I will do this next time

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
jcharkow: Running with a big library with over 2000000 it generates decoys in a reasonable time for me (e.g. 20 sec) so I think this implementation is good enough for now

src/openms/source/ANALYSIS/OPENSWATH/MRMDecoy.cpp
```diff
@@ -379,6 +378,27 @@ namespace OpenMS
...
+      if (loc > -1 && loc < (int)pep.sequence.size())
+      {
+        full_peptide_name += pep.sequence[loc];
+      }
+      // C-terminal and N-terminal modifications may be at positions -1 or pep.sequence
+      for (Size modloc = 0; modloc < pep.mods.size(); modloc++)
+      {
+        if (pep.mods[modloc].location == loc)
+        {
+  	full_peptide_name += "(UniMod:" + String(pep.mods[modloc].unimod_id) + ")";
```
timosachsenberg, line 395: there are still some style errors. e.g. here no indentation. below it uses probably tabs instead of two spaces. maybe give it another quick check?

timosachsenberg: I will give try to give a quick fix to move this forward

timosachsenberg: can't push to Roestlab... I think you need to do the changes yourself. sorry

jcharkow: no problem I'll take a look

timosachsenberg (approved):

Issue #6052, opened 2022-04-20
Setup GitHub sponsoring button


jpfeuffer:
In the repo settings.

poshul:
This is now set-up

PR #6090, opened 2022-05-02
added funding.yml

poshul:
# Description

add FUNDING.yml with our opencollective account
https://github.com/OpenMS/OpenMS/issues/6052

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+open_collective: openms


timosachsenberg:
Cool. What does this imply? E.g. does this show up in the sponsor thingy etc?

poshul:
I believe that I also need to sign us up with Github Sponsors as well but I need to be an organization owner to do that.

timosachsenberg:
Done.

PR #6105, opened 2022-05-09
Support for enzyme term specificity (DB search params.) in OMS (SQLite) files

hendrikweisser:
Term specificity information for digestion enzymes was missing from the comparison operator in `DBSearchParam` and from input/output in OMS (SQLite) files. I've now added it.
Since there may be existing OMS files without the corresponding column in the database schema, I've increased the OMS file version number and added a version check to the file loading code. This makes it possible to still load files with the old schema (version 1).


timosachsenberg:
https://cdash.openms.de/viewBuildError.php?buildid=38412

hendrikweisser:
> https://cdash.openms.de/viewBuildError.php?buildid=38412

Should be fixed here:
https://github.com/OpenMS/OpenMS/pull/6105/commits/bec5d860a191378190c481e0cd4f570c33e26647

PR #6106, opened 2022-05-10
Feature/search spectrum extended

bertrandboudaud:
## Description

- Add extractSpectra to TargetedSpectraExtractor with support of MS1 features
- TargetedSpectraExtractor::searchSpectrum added an option to add unknown features to the result featureMap


src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -261,45 +302,81 @@ namespace OpenMS
...
+              s.setMetaValue("native_id", native_id);
+              s.setMetaValue("identifier", hit.getMetaValue("identifier"));
+              s.setMetaValue("description", hit.getMetaValue("description"));
+              s.setMetaValue("modifications", hit.getMetaValue("modifications"));
+              std::string adducts;
+              try
+              {
+                std::string str = hit.getMetaValue("modifications").toString();
+                std::string delimiter = ";";
+                adducts = str.substr(1, str.find(delimiter) - 1);
```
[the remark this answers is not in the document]
cbielow: otherwise this might be more compact:
```
adducts = hit.getMetaValue("modifications").toString().prefix(';')
```

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -261,45 +302,81 @@ namespace OpenMS
...
+              s.setMetaValue("adducts", adducts);
+              OpenMS::EmpiricalFormula chemform(hit.getMetaValue("chemical_formula").toString());
+              double adduct_mass = s.getMZ() * std::abs(hit.getCharge()) + static_cast<double>(hit.getMetaValue("mz_error_Da")) - chemform.getMonoWeight();
+              s.setMetaValue("dc_charge_adduct_mass", adduct_mass);
+              s.setMetaValue("chemical_formula", hit.getMetaValue("chemical_formula"));
+              s.setMetaValue("mz_error_ppm", hit.getMetaValue("mz_error_ppm"));
+              s.setMetaValue("mz_error_Da", hit.getMetaValue("mz_error_Da"));
+              s.setCharge(hit.getCharge());
+              std::vector<OpenMS::Feature> subs = {s};
+              f.setSubordinates(subs);
```
cbielow, line 348:
```suggestion
              f.setSubordinates({s});
```

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -261,45 +302,81 @@ namespace OpenMS
...
+              s.setMetaValue("description", "");
+              s.setMetaValue("modifications", "");
+              s.setMetaValue("adducts", "");
+              s.setMetaValue("dc_charge_adduct_mass", 0);
+              s.setMetaValue("chemical_formula", "");
+              s.setMetaValue("mz_error_ppm", 0);
+              s.setMetaValue("mz_error_Da", 0);
+              // s.setCharge(hit.getCharge()); // The polarity should have been set during peak picking
+              std::vector<OpenMS::Feature> subs = {s};
+              f.setSubordinates(subs);
```
cbielow, line 376:
```suggestion
              f.setSubordinates({s});
```

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -868,7 +968,21 @@ namespace OpenMS
...
         f.setUniqueId();
-        f.setMetaValue("PeptideRef", f_map.first);
+
+        // parse the identifier
+        std::string id_f;
+        try
+        {
+          std::string str = f_map.first;
+          std::string delimiter = "_";
+          id_f = str.substr(0, str.find(delimiter));
```
cbielow, line 978: String::prefix as above?

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -955,22 +1069,26 @@ namespace OpenMS
...
+        if (s.metaValueExists("PeptideRef") && s.metaValueExists("identifier"))
         {
-          auto found_s = fmapmap.emplace(s.getMetaValue("identifier").toStringList().at(0), std::vector<OpenMS::Feature>({s}));
+          std::string id = std::string(s.getMetaValue("PeptideRef")) + std::string("_") + std::string(s.getMetaValue("identifier").toStringList().at(0));
+          std::string id_s = id + std::string("_") + std::to_string(s.getRT());
+          auto found_s = fmapmap.emplace(id_s, std::vector<OpenMS::Feature>({s}));
           if (!found_s.second)
           {
-            fmapmap.at(s.getMetaValue("identifier").toStringList().at(0)).push_back(s);
+            fmapmap.at(id_s).push_back(s);
```
cbielow, line 1091: looks identical to the case above. Use a Lambda?

src/openms/source/FORMAT/MSPGenericFile.cpp
```diff
@@ -160,53 +166,57 @@ namespace OpenMS
...
-          output_file << "Synon: " << synonyms_str.substr(start, end - start) << '\n';
+          size_t start;
+          size_t end = 0;
+          std::string synonyms_str(synonyms);
+          std::string separator(synonyms_separator_);
+          while ((start = synonyms_str.find_first_not_of(separator, end)) != std::string::npos)
+          {
+            end = synonyms_str.find(separator, start);
+            output_file << "Synon: " << synonyms_str.substr(start, end - start) << '\n';
+          }
```
cbielow, line 184: I can see this is efficient, but a bit boilerplate.
If the `synonyms_separator_` where a char, one could do
```
  StringList list;
  synonyms.split(synonyms_separator_, list);
  for (const auto& syn : list) output_file << "Synon: " << syn << '\n';
```

but its a string unfortunately. But from taking a quick look at the param_:
`"The character that will separate the synonyms in the Synon metaValue."`
the synonyms_separator_ member could simply be a char.

src/openms/source/FORMAT/MSPGenericFile.cpp
```diff
@@ -160,53 +166,57 @@ namespace OpenMS
...
-      }
-      // Peaks
-      output_file << "Num Peaks: " << spectrum.size() << '\n';
-      int peak_counter = 0;
-      for (const auto& peak : spectrum)
-      {
-        output_file << peak.getPos() << ":" << peak.getIntensity() << " ";
-        if ((++peak_counter % 5) == 0)
+        // Other metadata
+        const std::vector<std::string> ignore_metadata = {"Synon", "CAS#", "NIST#", "Num Peaks"};
```
cbielow, line 191:
```suggestion
        constexpr std::array<std::string, 4> ignore_metadata = {"Synon", "CAS#", "NIST#", "Num Peaks"};
```

cbielow (commented):
Nice addition.
Thanks. 👍 

I have not looked at the algorithm much, but here are some suggestions :)

jpfeuffer:
Hi! Looks great. Can you explain in a few sentences what this allows one to do now? What would be an application for this?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.h
```diff
@@ -209,11 +209,13 @@ namespace OpenMS
       @brief Search accurate masses and add identification (peptide hits) as features/sub-features
 
       @param[in] feat_map The feature map to search in
-      @param[in] feat_map_output The output feature map, with peptide identifaction as sub features
+      @param[out] feat_map_output The output feature map, with peptide identifaction as sub features
+      @param[in] add_unknown_features Adds unknown features to the feature map
```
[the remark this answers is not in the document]
bertrandboudaud: renamed unknown to unidentified

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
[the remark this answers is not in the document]
bertrandboudaud: > starting at `1` is intentional?

yes, the first letter stands for the actual metabolite and then everything after are the abducts up until the ";"

image
(in the case the metabolite is "M")

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
bertrandboudaud: > String::prefix as above?

We are using std::string there, do you advice to use OpenMS String instead and profit from the prefix method?

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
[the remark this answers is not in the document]
cbielow: Maybe a comment with an example would be good to have here?!

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -261,45 +302,72 @@ namespace OpenMS
...
+              // Fill in accurateMassSearch metavalues
+              DataValue identifiers(std::vector<std::string>({mass_of_the_peak.str()}));
+              s.setMetaValue("identifier", identifiers);
+              s.setMetaValue("description", "");
+              s.setMetaValue("modifications", "");
+              s.setMetaValue("adducts", "");
+              s.setMetaValue("dc_charge_adduct_mass", 0);
+              s.setMetaValue("chemical_formula", "");
+              s.setMetaValue("mz_error_ppm", 0);
+              s.setMetaValue("mz_error_Da", 0);
```
cbielow, line 365: do you really need to fill metavalues which do not exist?
Especially `mz_error_Da = 0` could easily be misinterpreted downstream.
Better to omit them entirely?

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
cbielow: yes

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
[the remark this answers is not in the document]
bertrandboudaud: ok, done

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
bertrandboudaud: done

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
```diff
@@ -225,25 +225,65 @@ namespace OpenMS
...
+              annotated_spectrum.setName(peptide_ref_f);
+              annotated_spectra.push_back(annotated_spectrum);
+              // fill the ms2 features map
+              Feature ms2_feature;
+              ms2_feature.setRT(spectrum_rt);
+              ms2_feature.setMZ(spectrum_mz);
+              ms2_feature.setIntensity(feature.getIntensity());
+              ms2_feature.setMetaValue("native_id", native_id_f);
+              ms2_feature.setMetaValue("PeptideRef", peptide_ref_f);
+              ms2_features.push_back(ms2_feature);
```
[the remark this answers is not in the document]
bertrandboudaud: indeed, done

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
bertrandboudaud: indeed. done.

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
dmccloskey: The reasoning was to maintain these MetaValues for compatibility with the identified features but filled with "Null" values that downstream algorithms could then use to infer that the step was ran but no information is available for the MetaValues.

I definitely see your point about the case of `mz_error_Da = 0` and agree this is not optimal.  What would you recommend for that situation?  Something like `-1`? or something different altogether?

dmccloskey:
> Hi! Looks great. Can you explain in a few sentences what this allows one to do now? What would be an application for this?

These methods allow us to perform DDA for 1) library construction (e.g., for downstream DIA/SWATH/SRM), 2) spectral library generation (e.g., for downstream MS2 fragment library matching, and 3) MS2 fragment library matching/annotation of unidentified masses.

bertrandboudaud:
Hi @cbielow and @jpfeuffer i have fixed all issues raised - I think-, is anything else needed?

dmccloskey:
Hi @cbielow and @jpfeuffer Please let us know if there is anything else needed.

dmccloskey:
> Hi @cbielow and @jpfeuffer Please let us know if there is anything else needed.

Hi @timosachsenberg or @matteopilz Please let us know if anything else is needed.  We have not heard back from @cbielow or @jpfeuffer so maybe out on vacation?

src/openms/source/ANALYSIS/OPENSWATH/TargetedSpectraExtractor.cpp
jpfeuffer: Another option would be to add this algorithm as a DataProcessing step or a metavalue at the FeatureMap or Experiment level. This, plus a missing metavalue in the feature would then be enough.
But if you want you can think about it and we merge this PR first.

jpfeuffer (approved):

Issue #543, opened 2014-01-09
Add iBAQ support to ProteinQuantifier [543]


timosachsenberg:
Submitted by hendrikweisser on 2013-06-07 16:04:07

iBAQ is a popular approach for absolute quantification of proteins. It adds up the abundances of all peptides of a protein, and divides this by the number of theoretically observable peptides (fully tryptic, 6-30 amino acids, no missed cleavages).
See http://www.nature.com/nature/journal/v473/n7347/extref/nature10098-s2.pdf.

Support for this method should be added to ProteinQuantifier - it would, however, require access to the full protein sequences (FASTA file).

---

Commented by hroest on 2013-06-18 18:24:40:
another solution would be to create a generic script that uses the aLFQ package in R which has an "OpenMS" option to solve the issue:
###### 

library(aLFQ)
data_import <- import(system.file("data","UPS1_LFQ_openmslfq.csv",package="aLFQ"),  "openmslfq", system.file("data","UPS1_conc.csv",package="aLFQ"),mergeruns=FALSE)
data_proteins <- ProteinInference(data_import,peptide_method="iBAQ",fasta=system.file("data","UPS12.fasta",package="aLFQ"))
###### 

we could then add this to share/OpenMS/SCRIPTS/

---

Commented by hendrikweisser on 2013-06-18 22:33:19:
I agree with George and would prefer to have the functionality added directly to OpenMS, since it's pretty basic and aLFQ is not generally available yet (private communcation).

PR #6107, opened 2022-05-11
[FEATURE] Add iBAQ to ProteinQuantifier

matteopilz:
## Description

fixes https://github.com/OpenMS/OpenMS/issues/543
## Checklist

<details>

</details>

<details>
  
</details>

---


src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -34,10 +34,11 @@
 //
 
 #include <OpenMS/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.h>
-
 #include <OpenMS/CONCEPT/LogStream.h>
 #include <OpenMS/SYSTEM/File.h>
 #include <OpenMS/MATH/STATISTICS/StatisticFunctions.h>
+#include <OpenMS/CHEMISTRY/EnzymaticDigestion.h>
+#include "OpenMS/DATASTRUCTURES/StringView.h"
```
timosachsenberg, line 41: please use < instead of quotes

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -48,13 +49,17 @@ namespace OpenMS
     DefaultParamHandler("PeptideAndProteinQuant"), stats_(), pep_quant_(),
     prot_quant_()
   {
+    std::vector<std::string> true_false = {"true","false"};
+
+    defaults_.setValue("method", "TOP3", "- TOP3 - quantify based on three most abundant peptides (number can be changed in 'top').\n- iBAQ (intensity based absolute quantification), will override top with 0 and average with sum.");
```
timosachsenberg, line 54: I think using TOP3 as parameter name could be confusing. "top" would be better I think.

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -391,82 +396,93 @@ namespace OpenMS
...
-      for (auto const & sta : pep_q.second.total_psm_counts)
+      for (auto const& sta : pep_q.second.total_psm_counts)
       {
         prot_quant_[accession].psm_counts[peptide][sta.first] += sta.second;
       }
     }
 
+    std::string method = param_.getValue("method");
+
+    if (method == "iBAQ")  // TODO: parameter override too late - ProteinQuantifier reads parameters earlier, meaning output file contains incorrect parameters
```
timosachsenberg, line 426: I think this could be confusing for a developer if after setting to iBAQ, running the algorithm and then setting it to "top" again the param_ object has changed.
I think this needs to be changed.

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -391,82 +396,93 @@ namespace OpenMS
...
-        for (auto const & sa : sas)
+        for (auto const& sa : sas)
         {
           const Size& sample_id = sa.first;
           const Size& psms = sa.second;
-          if (psms > 0) prot_q.second.total_distinct_peptides[sample_id]++; // count this peptide sequence once if observed in sample
-          prot_q.second.total_psm_counts[sample_id] += psms; // count all PSMs of this protein in this sample
+          if (psms > 0)
+            prot_q.second.total_distinct_peptides[sample_id]++; // count this peptide sequence once if observed in sample
+          prot_q.second.total_psm_counts[sample_id] += psms;    // count all PSMs of this protein in this sample
```
timosachsenberg, line 451: indentation

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
timosachsenberg: e.g. if top or average is used at some point check there again for the method and use if / else for the two cases

timosachsenberg (changes requested):
nice addition. some smaller comments.

timosachsenberg:
TODO:
- add a test in peptideandproteinquant
- add a topp tool test
- check if the method is stored in the dataprocessing in the output file

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
matteopilz: This one is not from me, are both supposed to be in the if clause?

timosachsenberg: sorry forget it :)

jpfeuffer:
Very cool. Can you explain a little bit what ibaq does or link a paper/explanation? (In the class header docu)
Which of the other params apply to ibaq? For example I am never sure which/ how many peptidoforms per peptide are considered .

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -51,27 +51,27 @@ namespace OpenMS
...
     defaults_.setValidStrings("method", {"top","iBAQ"});
 
-    defaults_.setValue("top", 3, "Calculate protein abundance from this number of proteotypic peptides (most abundant first; '0' for all)");
-    defaults_.setMinInt("top", 0);
+    defaults_.setValue("top:N", 3, "Calculate protein abundance from this number of proteotypic peptides (most abundant first; '0' for all)");
+    defaults_.setMinInt("top:N", 0);
 
-    defaults_.setValue("average", "median", "Averaging method used to compute protein abundances from peptide abundances");
-    defaults_.setValidStrings("average", {"median","mean","weighted_mean","sum"});
+    defaults_.setValue("top:average", "median", "Averaging method used to compute protein abundances from peptide abundances");
```
jpfeuffer, line 60: I think if we start renaming, we should call this "top:aggregate" or something like that. A sum is not an average

timosachsenberg:
can you look into the failing tests so we can move this forward?

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -429,6 +429,11 @@ namespace OpenMS
 
     if (method == "iBAQ")
     {
+      if (proteins.getHits().empty())
+      {
+        OPENMS_LOG_ERROR << "No protein matches found, cannot calculate iBAQ!";
```
timosachsenberg, line 434: do we need a hard error? e.g., exception
will probably not make sense to continue here

timosachsenberg: or do we get a meaningful but empty output? then this would be fine I guess

matteopilz: The output will be like setting top:N to 0 and top:aggregate to sum. The results won't be divided by the number of theoretical peptides.

matteopilz: I guess we can either warn the user, that this happened or stop the execution entirely.

timosachsenberg: if the results don't make sense (e.g., are not iBAQ values) then we should fail. You will not read warnings in a log file when run on 1000 files

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -431,7 +431,7 @@ namespace OpenMS
     {
       if (proteins.getHits().empty())
       {
-        OPENMS_LOG_ERROR << "No protein matches found, cannot calculate iBAQ!";
+        OPENMS_LOG_FATAL_ERROR << "No protein matches found, cannot calculate iBAQ!";
```
timosachsenberg, line 434: add << std::endl;
Does this throw an exception / terminate the program?

matteopilz: It should: "Macro to be used if fatal error are reported (processing stops)"

timosachsenberg: can you double check? the documentation could just mean that this needs to be used if processing is stopped.
Ambiguous right now

matteopilz: I tried to test it, but it seems that in a standard workflow this will not be reached. I've added it to address an issue in the class test, where no protein hits were set. However, it seems, that this exception has already been catched.

timosachsenberg: Are you sure that an exception is thrown?

timosachsenberg: for me it looks like it is just a generic log stream that outputs to cerr by default:
https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CONCEPT/LogStream.cpp#L565

matteopilz: But when I remove the protein references in all peptide entries, the program runs through. However, the results are then empty.

timosachsenberg: can we zoom?

src/openms/source/ANALYSIS/QUANTITATION/PeptideAndProteinQuant.cpp
```diff
@@ -48,25 +49,29 @@ namespace OpenMS
...
+    defaults_.setMinInt("top:N", 0);
+
+    defaults_.setValue("top:aggregate", "median", "Aggregation method used to compute protein abundances from peptide abundances");
+    defaults_.setValidStrings("top:aggregate", {"median","mean","weighted_mean","sum"});
 
-    std::vector<std::string> true_false = {"true","false"};
 
-    defaults_.setValue("include_all", "false", "Include results for proteins with fewer proteotypic peptides than indicated by 'top' (no effect if 'top' is 0 or 1)");
-    defaults_.setValidStrings("include_all", true_false);
+    defaults_.setValue("top:include_all", "false", "Include results for proteins with fewer proteotypic peptides than indicated by 'N' (no effect if 'N' is 0 or 1)");
```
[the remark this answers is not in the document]
matteopilz: How is that done, with `registerSubsection` or `registerTOPPSubsection`?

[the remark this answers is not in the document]
timosachsenberg: @jpfeuffer do we have explicit setting of sections in Param?
I think you mean the TOPP tools.
but what we can do is setting a description *after* line 64 e.g., defaults_.setSectionDescription("top", "Parameter relevant to top N quantification.");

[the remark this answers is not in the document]
jpfeuffer: Yes maybe it does not work in algorithm classes. But yeah would be great if we have that section description. Maybe the section is implicitly created when you have ":" in the param names.

src/tests/class_tests/openms/source/PeptideAndProteinQuant_test.cpp
```diff
@@ -357,6 +357,32 @@ START_SECTION((const ProteinQuant& getProteinResults()))
 }
 END_SECTION
 
+START_SECTION()
```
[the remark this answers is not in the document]
matteopilz: Wouldn't setting a name be better? There are multiple sections with `(const ProteinQuant& getProteinResults())` already, so it seems a completely new test should also be in a new section.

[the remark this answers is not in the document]
timosachsenberg: Ah right. Yeah but it needs a proper name like the others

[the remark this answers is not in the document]
matteopilz: I have named it `(const ProteinQuant& getProteinResults())` as well, as it fits best.

PR #6114, opened 2022-05-12
JSON export for OMS (SQLite) files

hendrikweisser:
Implemented conversion of OMS files (SQLite-based format for new ID data structure) to JSON; useful e.g. to create reproducible output for testing. Added a corresponding tool ("JSONExporter"). (I decided against adding this functionality to TextExporter because none of the existing file formats/options overlap.)
Currently this only exports ID data, but could be extended to feature data (which OMS files can already store) as well.

The main idea behind making the output deterministic/reproducible was to replace foreign keys in the database (which use non-reproducible memory addresses) with properties that identify the row/object being referenced. This can be simple, but can also get complex across two dimensions:
1. An object references multiple other classes, i.e. there are multiple foreign keys per row. This requires multiple JOINs in the database to resolve.
2. The "identity" of an object is composed of several properties, e.g. observation match (=PSM): input file, spectrum ID, molecule type, sequence/identifier, charge, adduct formula, adduct charge. This makes it unwieldy to include all these properties where the object is being referenced. In such cases I generate an ordered index (using ROW_NUMBER in SQL) and use that for referencing.

In conclusion, this was trickier than I thought.


src/openms/source/FORMAT/OMSFileLoad.cpp
```diff
@@ -41,8 +41,12 @@
 #include <QtSql/QSqlDatabase>
 #include <QtSql/QSqlError>
 #include <QtSql/QSqlQuery>
+#include <QtSql/QSqlRecord>
 // strangely, this is needed for type conversions in "QSqlQuery::bindValue":
 #include <QtSql/QSqlQueryModel>
+// JSON export:
```
timosachsenberg, line 47: can we go with some additional json dependency - we already use the nlohmann json library.

timosachsenberg:
>The "identity" of an object is composed of several properties, e.g. observation match (=PSM): input file, spectrum ID, molecule type, sequence/identifier, charge, adduct formula, adduct charge. This makes it unwieldy to include all these properties where the object is being referenced. In such cases I generate an ordered index (using ROW_NUMBER in SQL) and use that for referencing.

Thanks. It looks like it could solve the testing problem for small files but I am a bit unsure on the performance for larger files. Did you try this on a big one already?

So basically this ROW_NUMBER is then an extra attribute in the JSON file right?
e.g. do I get it correctly that:
            "processing_step_index": 1,
is such an order row index.
and
            "processing_step_order": 1,
is an attribute?

Would it make sense to rename/mark it so it is clearly distinguished from normal attributes?

src/openms/source/FORMAT/OMSFileLoad.cpp
hendrikweisser: Can you elaborate?

timosachsenberg: #include <QtCore/QJsonArray> // for JSON export
We already have https://github.com/OpenMS/OpenMS/blob/develop/src/openms/thirdparty/json
and I would love to not have QJson here. Especially as I would love to see the use of Qt reduced in the core library (e.g. GUI would be a different story).

timosachsenberg:
I agree - my feeling is the that keeping the index explicit makes it easier to debug and follow references in the JSON.

timosachsenberg:
I would probably also export the version table. Should be easy to whitelist specific lines if needed. Or is there any other reason?

src/openms/source/FORMAT/OMSFileLoad.cpp
```diff
@@ -1130,4 +1134,237 @@ namespace OpenMS::Internal
...
+        sql = "SELECT software_name, software_version, score_type_accession, score_type_name, score_type_order FROM ID_ProcessingSoftware_AssignedScore JOIN (" + subquery_software + ") ON software_id = sw_id JOIN (" + subquery_score_ + ") ON score_type_id = st_id ORDER BY software_name, software_version, score_type_order";
+        json_data.insert("ID_ProcessingSoftware_AssignedScore", exportQueryToJSON_(sql, {}));
+      }
+    }
+    // processing steps:
+    if (tableExists_(db_name_, "ID_ProcessingStep"))
+    {
+      // processing steps potentially get referenced a lot, but their "identity" is complex
+      // (software + input files + timestamp) - create a view that includes an index number for referencing:
+      // @TODO: include input files in ordering (via "GROUP_CONCAT")
```
timosachsenberg, line 1252: how is the order determined right now?

src/openms/source/FORMAT/OMSFileLoad.cpp
```diff
@@ -1130,4 +1134,237 @@ namespace OpenMS::Internal
...
+                                                       "score_type_id", join_key}));
+    }
+
+    exportMetaInfoToJSON_(json_data, parent_table, join, order, join_key);
+  }
+
+
+  void OMSFileLoad::exportToJSON(ostream& output)
+  {
+    // @TODO: this constructs the whole JSON file in memory - write directly to stream instead?
```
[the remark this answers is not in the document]
hendrikweisser: I'd leave that for a future pull request.

src/openms/source/FORMAT/OMSFileLoad.cpp
hendrikweisser: Just software name, version, processing date/time. See the `ORDER BY ...` in the next line.

src/openms/source/FORMAT/OMSFileLoad.cpp
```diff
@@ -1130,4 +1134,237 @@ namespace OpenMS::Internal
...
+    }
+
+    QJsonArray array;
+    while(query.next())
+    {
+      QJsonObject record;
+      for (int i = 0; i < query.record().count(); ++i)
+      {
+        // @TODO: this will repeat field names for every row -
+        // avoid this with separate "header" and "rows" (array)?
```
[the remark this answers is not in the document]
timosachsenberg: If one would go this way would this give an advantage to a column format like:
"a": [1,3]
"b": [2,4]

jpfeuffer:
Not familiar with SQLite but it might be a version issue: https://stackoverflow.com/questions/71315554/sqlite-keep-getting-the-error-near-syntax-error

hendrikweisser:
Thanks, Julianus. I've added the query to the error message - let's see what happens.
I also agree that it must be an issue with the SQLite version. I think Ubuntu 18.04 uses Qt 5.9 with SQLite 3.16, while my 20.04 uses Qt 5.12 with SQLite 3.26. I haven't seen anything obvious in the release notes between 3.16 and 3.26 to explain the error.

By the way, is there a good reason we're still testing on 18.04 instead of 20.04?

hendrikweisser:
Actually the answer is in the Stack Overflow link Julianus posted - I just failed to realise that "ROW_NUMBER" is a window function, support for which was only added in SQLite 3.25. (See https://www.sqlite.org/windowfunctions.html.)

That it explains why the code isn't working on Ubuntu 18.04 (and I guess also Mac OS), but it means there isn't a straightforward fix.

hendrikweisser:
I've found a bit of time to transition the implementation of storing .oms files to use reproducible database keys. So instead of using memory addresses as database keys, now consecutive numbers (1-...) are used, preserving the ordering of objects in their containers. The storage (database creation) code got a bit more complex, because mappings between objects and their keys in the database need to be tracked and applied to references/foreign keys. On the flip side, the JSON export code became a lot simpler, because now database keys can be used to ensure a reproducible ordering in the output. (As a bonus, this should solve the problem with the older Ubuntu/Qt/SQLite version not supporting window functions, because they aren't used anymore.)

poshul (approved):
As long as all the tests pass this looks good to go.

poshul:
rebuild jenkins

timosachsenberg:
hmm do we need to change something in the deployment? e.g., now our core library also depends on qt json right?

poshul:
Qt JSON is part of qtCore so we should already have the required libraries.

timosachsenberg:
ah nice. that is strange but great!

PR #6117, opened 2022-05-13
[FEATURE, TEST] added test mode flag to xQuest spec.xml output format

enetz:
## Description

Updated TOPP tests for OpenPepXL and OpenPepXLLF .spec.xml output.
If in test mode, write out raw string of the spectrum, instead of the Base64 encoding. This allows FuzzyDiff to compare the numbers correctly and not fail due to slight differences in floating point number representation.
Tested successfully with the relevant boost versions, that caused the original issue.

## Checklist

<details>

</details>

<details>
  
</details>

---


src/openms/source/FORMAT/XQuestResultXMLFile.cpp
```diff
@@ -285,11 +285,18 @@ namespace OpenMS
...
+    if (!test_mode)
+    {
+      String out_encoded;
+      Base64().encodeStrings(in_strings, out_encoded, false, false);
+      String out_wrapped;
+      wrap_(out_encoded, 76, out_wrapped);
+      return out_wrapped;
+    }
+    else
+    {
```
timosachsenberg, line 297:
```suggestion
    { // skip base64 encoding in test mode
```

timosachsenberg:
I quickly checked cdash and it seems there is still an issue: [src/openms/source/FORMAT/XQuestResultXMLFile.cpp:298](https://github.com/OpenMS/OpenMS/blob/master/src/openms/source/FORMAT/XQuestResultXMLFile.cpp#L298):14: error: no viable conversion from returned value of type 'std::vector<String>' to function return type 'OpenMS::String'

timosachsenberg (approved):

Issue #5693, opened 2021-12-02
TMTpro 18plex


glormph:
Hi, I have been warned that TMTpro 18plex is coming (to our lab), and I wanted to ask if there is planned or existing support for it in OpenMS, as I could not find a mention of it in github issues. Thank you :)

https://www.thermofisher.com/order/catalog/product/A52045
https://pubmed.ncbi.nlm.nih.gov/33900084/

timosachsenberg:
Hi,
We currently don't have a project using TMTpro 18plex so we did not allocate resources to add support, yet.
If you feel familiar with C++ or want to get into it, you can take a look at: https://github.com/OpenMS/OpenMS/pull/4295/files
those were the changes applied to add TMT16plex.
Another possibility would be to ask the author of that PR @poshul if he could add it.

radusuciu:
Is adding support for this basically copying `TMTSixteenPlexQuantitationMethod` to `TMTEighteenPlexQuantitationMethod`, adding a few constants, new tests and making sure it builds? If so, I can probably do this!

timosachsenberg:
Yes exactly! That would be awesome

radusuciu:
Alright! I'm planning on using IsobaricAnalyzer for 18-plex data in the next couple of weeks so I can contribute to that if no one beats me to it.

PR #6131, opened 2022-05-20
Add support for TMT 18-plex

radusuciu:
## Description

This PR adds support for TMT-18 plex data - see issue #5693. It is a work in-progress but I wanted to get the process started and make sure I'm not completely off base given that I am not experienced in C++ or OpenMS development.

I believe that there is an issue with the tests currently that I'll be fixing locally before my next push.

## Checklist

<details>

</details>

<details>
  
</details>

---


timosachsenberg:
Awesome. Let us know if you hit any obstacles.

radusuciu:
So I have some questions about the correction matrix. For this PR, I just took the correction matrix from the 16-plex file and added values from a reagent kit that I have access to:

image
product number: A52046
lot number: WJ334386

If I understand correctly, currently all 18-plex kits are basically 16 + 2, and the default values for the 16-plex method are really averages of a few different 16-plex kits. If both of those things are true, then I think my approach makes sense, but it'd be great if someone could validate that.

For completeness, here is the product data sheet for 16-plex portion of my kit.

image
product number: A44520
lot number: WH327114

timosachsenberg:
Looks good to me though I would love to get some feedback from the guys who already messed with the correction matrix.

radusuciu:
Alright, sometime this week or next I plan on testing it with some actual data.

I see I'm still failing some linting rules - I can fix this as well, though I will note that this is mostly a copy + find/replace of the 16-plex code.

timosachsenberg:
The linting is not that important. Feel free to ignore for now

radusuciu:
Quick note, didn't quite get to it this week but it's still on my radar.

timosachsenberg:
no worries take any time you need :)

glormph:
Hi, my C++ skills are nonexistent, but if there's anything that can be done here to get it into the next release, I wouldn't mind helping. Lots of TMT18 to analyze in our lab already.

jpfeuffer:
I think the correction matrix and the channel map indeed needs to be double-checked. An average as default is fine but it looks like they differ significantly from your sheet.
Also see #6327  and related issues for why you cannot model your whole sheet currently.

radusuciu:
If we can ignore the higher interactions I can go over the sheet and attempt a correction, but I'd appreciate any and all feedback

src/openms/source/ANALYSIS/QUANTITATION/TMTEighteenPlexQuantitationMethod.cpp
```diff
@@ -0,0 +1,219 @@
...
+    //    "131C", 131.144500, 129C, 130C, x, x
+
+    // create the channel map                                                //-2  -1  +1  +2
+    // TODO: double check this
+    channels_.push_back(IsobaricChannelInformation("126",   0, "", 126.127726, -1, -1,  2,  4));
+    channels_.push_back(IsobaricChannelInformation("127N",  1, "", 127.124761, -1, -1,  3,  5));
+    channels_.push_back(IsobaricChannelInformation("127C",  2, "", 127.131081, -1,  0,  4,  6));
+    channels_.push_back(IsobaricChannelInformation("128N",  3, "", 128.128116, -1,  1,  5,  7));
+    channels_.push_back(IsobaricChannelInformation("128C",  4, "", 128.134436,  -1,  2,  6,  8));
+    channels_.push_back(IsobaricChannelInformation("129N",  5, "", 129.131471,  -1,  3,  7,  9)));
```
cbielow, line 73:
```suggestion
    channels_.push_back(IsobaricChannelInformation("128C",  4, "", 128.134436,  0,  2,  6,  8));
    channels_.push_back(IsobaricChannelInformation("129N",  5, "", 129.131471,  1,  3,  7,  9)));
```
The -2 channels are 0% in the example sheet (so filling in a `-1` serves the same purpose), but another sheet could contain value >0%, thus we should use the mapping.

src/openms/source/ANALYSIS/QUANTITATION/TMTEighteenPlexQuantitationMethod.cpp
```diff
@@ -0,0 +1,219 @@
...
+                                                                      "0.04/3.10/5.24/0.08",
+                                                                      "0.03/3.41/4.57/0.12",
+                                                                      "0.08/3.90/4.04/0.04",
+                                                                      "0.16/4.30/1.80/0.0",
+                                                                      "0.11/4.55/2.29/0.0",
+                                                                      "0.08/3.87/3.49/0.03",
+                                                                      "0.22/4.96/1.37/0.0",
+                                                                      "0.33/6.11/1.14/0.0",
+                                                                      "0.14/5.81/0.31/0.0",
+                                                                      "0.21/5.42/0.00/0.0"},
```
cbielow, line 135: not sure about the defaults there.
If we take the example sheet from above, we should only use the number from a single cell (the one which is mapped) and not add up both values from the sub-columns. The contribution from the channel which is currently not mapped should be independent of the mapped channel, hence our error is smaller if we do not sum up the two values .right?

src/openms/include/OpenMS/ANALYSIS/QUANTITATION/TMTEighteenPlexQuantitationMethod.h
```diff
@@ -0,0 +1,100 @@
...
+#include <OpenMS/config.h>
+
+#include <OpenMS/ANALYSIS/QUANTITATION/IsobaricQuantitationMethod.h>
+
+namespace OpenMS
+{
+  /**
+    @brief TMT 18plex quantitation to be used with the IsobaricQuantitation.
+
+    @htmlinclude OpenMS_TMTSixPlexQuantitationMethod.parameters
```
cbielow, line 46:
```suggestion
    @htmlinclude OpenMS_TMTEighteenPlexQuantitationMethod .parameters
```

but needs adding to `doc/doxygen/parameters/DefaultParamHandlerDocumenter.cpp` as well (search for `DOCME(TMTSixPlexQuantitationMethod)` in there and add the new `TMTEighteenPlexQuantitationMethod `)

src/openms/source/ANALYSIS/QUANTITATION/TMTEighteenPlexQuantitationMethod.cpp
jpfeuffer: Yes I realized that too. I was confused by the +1Da +2Da headers of the columns and went against my own suggestion of just using the C13 values.

src/openms/source/ANALYSIS/QUANTITATION/TMTEighteenPlexQuantitationMethod.cpp
```diff
@@ -0,0 +1,219 @@
...
+                       "Correction matrix for isotope distributions in percent (see documentation); use the following format: <-2Da>/<-1Da>/<+1Da>/<+2Da>; e.g. '0/0.3/4/0', '0.1/0.3/3/0.2'");
```
cbielow, line 136: This needs a better description. The user has two subcolumns to choose from per 1Da offset, so it should say something like

```suggestion
                       "Correction matrix for isotope distributions in percent (see documentation); use the following format: <-2Da>/<-1Da>/<+1Da>/<+2Da>; e.g. '0/0.3/4/0', '0.1/0.3/3/0.2'. Since there are two columns per 1Da offset in the Thermo data sheet, you should always pick the *C13* version, not the *N15* one, i.e. use these four columns: "-2x 13C , -13C, +13C, +2x 13C");
```

jpfeuffer:
Wouldn't it be easy to just not limit the entries to 4? Problem solved.
Since we are defining the channel IDs that are affected anyway, the code knows where to put it in the matrix.

cbielow:
Yes, can surely be done. But it requires some code changes. I'd say a day worth of work (including tests etc)

jpfeuffer:
Done here #6390 

The problem is: in e.g. 16plex we had channels that had a -1 for potential contributions and then had a non-zero value in the correction matrix. How? Why?
Pretty sure this was wrong.

jpfeuffer:
Thanks a lot @radusuciu your commits were included in my extended PR

PR #6132, opened 2022-05-21
add peptide level fdr/q-values

timosachsenberg:
## Description

add peptide level q-values or FDR in FalseDiscoveryRate as default

## Checklist

<details>

</details>

<details>
  
</details>

---


timosachsenberg:
Draft PR for adding peptide level q-values / FDR
TODO:
- tests need update
- documentation

jpfeuffer:
Ah there is a function in IDFilter to annotate best per peptide.

timosachsenberg:
But the. We would need to filter this separately and loose the option to do this charge state and run specific. Like this it should be compatible with the other options

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -166,15 +172,43 @@ namespace OpenMS
...
+            const String peptide_sequence = it->getHits()[i].getSequence().toUnmodifiedString();
+            const double score = it->getHits()[i].getScore();
+
             if (target_decoy == "target" || target_decoy == "target+decoy")
             {
-              target_scores.push_back(it->getHits()[i].getScore());
+              target_scores.push_back(score);
+
+              // store best score for peptide (unmodified sequence)
+              if (auto it = peptide_to_best_target_score.find(peptide_sequence); it == peptide_to_best_target_score.end())
```
jpfeuffer, line 183: I think emplace is better here.

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -166,15 +172,43 @@ namespace OpenMS
...
+              }              
             }
             else
             {
               if (target_decoy == "decoy")
               {
-                decoy_scores.push_back(it->getHits()[i].getScore());
+                decoy_scores.push_back(score);
+                // store best score for peptide (unmodified sequence)
+                if (auto it = peptide_to_best_decoy_score.find(peptide_sequence); it == peptide_to_best_decoy_score.end())
```
jpfeuffer, line 201: Here too

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -283,6 +317,28 @@ namespace OpenMS
...
+        for (const auto& ps : peptide_to_best_decoy_score)
+        {
+          decoy_peptide_scores.push_back(ps.second);
+        }
+        for (const auto& ps : peptide_to_best_target_score)
+        {
+          target_peptide_scores.push_back(ps.second);
+        }      
+        map<double, double> score_to_peptide_fdr;
+        calculateFDRs_(score_to_peptide_fdr, target_peptide_scores, decoy_peptide_scores, q_value, higher_score_better);
```
jpfeuffer, line 331: I think applyBasic is faster and more correct

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
```diff
@@ -310,6 +366,24 @@ namespace OpenMS
...
+              {
+                peptide_fdr = peptide_to_best_decoy_score[peptide_sequence];
+              }
+              else
+              {
+                peptide_fdr = peptide_to_best_target_score[peptide_sequence];
+              }
+              if (q_value)
+              {
+                hit.setMetaValue("peptide q-value", peptide_fdr);
```
jpfeuffer, line 381: I would actually vote for main score and put PSM qvalue in metavalue

jpfeuffer (changes requested):

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
timosachsenberg: Why? This is a property of the peptide level (which does not exist in the old id datastructure). So psm, peptide and protein score q-value should be imho main scores.

jpfeuffer: But you can only have one main score.
Until this is done in the new data structure I would prefer having peptide FDR as the main score if both exist for a PeptideHit.

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
timosachsenberg: puhhh this is how it i done for the others. maybe make a note or open an issue? I think FDR is  one of the classes that should get a simple API - there are now just so many ways to do things I completely lost track what is the best way to do things

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
timosachsenberg: Maybe we need to discuss htis in a quick video chat but my idea was to:
- calculate peptide level q-values/fdr by default without the need of setting any flag
- keep the PSM score as main score as otherwise we would break all downstream workflows (or how would you do that?)
- maybe already add a filtering option for peptide FDR in the FDR tool

I am open for suggestions but using the peptide FDR/q-value as main score seemed to break a lot of things (if done by default)

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
jpfeuffer: I just kept it for backwards compatibility. If someone can explain and comment the code, it can stay. It also needs a complexity analysis since it is a very central part. Otherwise, if I don't understand it, I would not use it.

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
jpfeuffer: I would like flags as for the other levels.
If both psm and peptide are active, peptide is going to be the main score and PSM is meta value.
I don't see any issues with it. As I said, I don't think any of our code requires q-values. The only thing that I think needs to be changed is export to mzTab because you need to change the CV term.

src/openms/source/ANALYSIS/ID/FalseDiscoveryRate.cpp
jpfeuffer: For now it is fine.

timosachsenberg:
I think that is as good as I can do right now.
Users can use the FDR tool to calculate peptide level FDR/q-value which get's stored in a meta value.
Not compete for everything we support but with IDScoreSwitcher, IDFilter etc. this could already be used to filter results

PR #6134, opened 2022-05-23
[FIX] DatabaseSuitability Tests only executed when CometAdapter is found

### LINKED ISSUE #6008: DatabaseSuitability tests fail on Github Actions
Linux: Seems like they are disabled on Linux. Third-party in path?
MacOS: On macos they fail completely? No idea why. Executable name universal? Executable rights on the runner (although comet test works I think)?

Waschi97:
resolves #6008

### DIFF
--- a/src/tests/topp/THIRDPARTY/third_party_tests.cmake
+++ b/src/tests/topp/THIRDPARTY/third_party_tests.cmake
@@ -345,8 +345,6 @@ endif()
 #------------------------------------------------------------------------------
 if (NOT (${SIRIUS_BINARY} STREQUAL "SIRIUS_BINARY-NOTFOUND"))
   # Note: Following test are performed without adduct/id information, since these are obtained by the MetaboliteAdductDecharger/AccurateMassSearch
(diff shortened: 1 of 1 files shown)


src/tests/topp/THIRDPARTY/third_party_tests.cmake
```diff
@@ -102,6 +102,10 @@ OPENMS_FINDBINARY(THERMORAWFILEPARSER_BINARY "ThermoRawFileParser.exe" "ThermoRa
 # LuciPhor2
 OPENMS_FINDBINARY(LUCIPHOR_BINARY "luciphor2.jar" "LuciPHOr2")
 
+#------------------------------------------------------------------------------
+# CometAdapter (Used in DatabaseSuitability)
+OPENMS_FINDBINARY(COMET_ADAPTER_BINARY "CometAdapter" "CometAdapter")
```
timosachsenberg, line 107: I think I am fine with this fix but would like to hear what @jpfeuffer thinks? Basically, checking if any OpenMS executable is found would work as well here?

timosachsenberg (approved):

timosachsenberg:
Thanks!

PR #6169, opened 2022-06-09
[Fix] GNPSMetaValueFile get mzML file names from ConsensusMap

axelwalter:
## Description

Small fix for GNPSMetaValueFile. Instead of passing a list of mzML file names, pass a ConsensusMap from which the mzML file names will be used to generate the meta value table.

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms/source/FORMAT/GNPSMetaValueFile.cpp
+++ b/src/openms/source/FORMAT/GNPSMetaValueFile.cpp
@@ -43,16 +43,21 @@
 namespace OpenMS
 {   
     /**
-    @brief Generates a meta value required for GNPS FBMN, as defined here: https://ccms-ucsd.github.io/GNPSDocumentation/metadata/
+    @brief Generates a meta value table required for GNPS FBMN, as defined here: https://ccms-ucsd.github.io/GNPSDocumentation/metadata/
     */
-    void GNPSMetaValueFile::store(const StringList& mzml_file_paths, const String& output_file)
-    {
+    void GNPSMetaValueFile::store(const ConsensusMap& consensus_map, const String& output_file)
+    {   
+        StringList mzML_file_paths;
+        for (const auto& header: consensus_map.getColumnHeaders())
+        {
+            mzML_file_paths.push_back(header.second.filename);
+        }
         std::ofstream outstr(output_file.c_str());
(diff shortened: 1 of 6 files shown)


src/openms/source/FORMAT/GNPSMetaValueFile.cpp
```diff
@@ -43,16 +43,21 @@
...
     /**
-    @brief Generates a meta value required for GNPS FBMN, as defined here: https://ccms-ucsd.github.io/GNPSDocumentation/metadata/
+    @brief Generates a meta value table required for GNPS FBMN, as defined here: https://ccms-ucsd.github.io/GNPSDocumentation/metadata/
     */
-    void GNPSMetaValueFile::store(const StringList& mzml_file_paths, const String& output_file)
-    {
+    void GNPSMetaValueFile::store(const ConsensusMap& consensus_map, const String& output_file)
+    {   
+        StringList mzML_file_paths;
+        for (const auto& header: consensus_map.getColumnHeaders())
```
timosachsenberg, line 51: can you check if https://github.com/OpenMS/OpenMS/blob/a686a995efc5d2f38519666039d856566a571654/src/openms/source/KERNEL/ConsensusMap.cpp#L570-L577
does the same?
then this would only be one line of code

jpfeuffer:
But isn't this strictly less powerful?
Or are you sure that users/devs have consensusXMLs 99% of the time?

eeko-kon:
>But isn't this strictly less powerful?
>Or are you sure that users/devs have consensusXMLs 99% of the time?

You cannot generate a FeatureQuantificationTable.txt or an MGF file without the consensusXML file, right? So there is no point of using the algorithm without a consensus.

The reason for generating a metadata table is to link the FeatureQuantificationTable MapIDs to filename and metadata information for visualization. If we try to generate the metadata from the input mzml files, this might lead to inconsistencies since the consensusXML keeps the whole path/to/original.mzML file.

jpfeuffer:
Ok, then it is fine.

src/openms/source/FORMAT/GNPSMetaValueFile.cpp
axelwalter: of course, that's much better!

timosachsenberg (approved):

Issue #5077, opened 2020-11-20
Profile compiling with clang -ftime-trace flag


timosachsenberg:
Might reveal some bottlenecks.
"Clang 9 (and newer) has a -ftime-trace flag, which makes it output a profiling report as JSON (in addition to an object file).
You can import this file into a profiler that comes with Chrome (chrome://tracing) to get a nice visualisation"

cbielow:
To put the numbers into perspective: the whole build took 11400 seconds (190min).

A whopping 2033 seconds of 11400 is simply the FrontEnd work of including `IdentificationData.h` (which is mainly used in BaseFeature.h). But it's not easy to get rid of :)

timosachsenberg:
sachsenb@DESKTOP-TPBSG5G:/mnt/d/development/WSL_OpenMS/build$ ../ClangBuildAnalyzer/ClangBuildAnalyzer --all src/ all_openms
Processing all files and saving to 'all_openms'...
  done in 116.0s. Run 'ClangBuildAnalyzer --analyze all_openms' to analyze it.
sachsenb@DESKTOP-TPBSG5G:/mnt/d/development/WSL_OpenMS/build$ ../ClangBuildAnalyzer/ClangBuildAnalyzer --analyze all_openms
Analyzing build trace from 'all_openms'...
**** Time summary:
Compilation (3429 times):
  Parsing (frontend):        12485.1 s
  Codegen & opts (backend):   5685.1 s

**** Files that took longest to parse (compiler frontend):
 41944 ms: src//tests/class_tests/openms/CMakeFiles/MSSim_test.dir/source/MSSim_test.cpp.o
 38194 ms: src//tests/class_tests/openms/CMakeFiles/EmgScoring_test.dir/source/EmgScoring_test.cpp.o
 37696 ms: src//tests/class_tests/openms/CMakeFiles/IonizationSimulation_test.dir/source/IonizationSimulation_test.cpp.o
 37034 ms: src//tests/class_tests/openms/CMakeFiles/DetectabilitySimulation_test.dir/source/DetectabilitySimulation_test.cpp.o
 34170 ms: src//tests/class_tests/openms/CMakeFiles/RTSimulation_test.dir/source/RTSimulation_test.cpp.o
 32742 ms: src//tests/class_tests/openms/CMakeFiles/EGHFitter1D_test.dir/source/EGHFitter1D_test.cpp.o
[...]

PR #6207, opened 2022-07-04
reduce compile time

cbielow:
## Description

Using some results from #5077, this PR removes the QDateTime dependency from OpenMS' DateTime (which has a huge include-time overhead) using PIMPL, and similarly reduces boost/regex.h dependencies.

**Results:**
Compile time for Clang 11 on Linux reduces by 10%
before: 11600 sec
after: 10500 sec 

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/DateTime.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/DateTime.h
@@ -214,7 +214,7 @@ public:
     void set(const String& date);
 
 private:
-    QDateTime* dt_;
+    QDateTime* dt_; // access through PImpl to avoid heavy include penalty
   };
 
 } // namespace OPENMS
--- a/src/openms/include/OpenMS/DATASTRUCTURES/DateTime.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/DateTime.h
@@ -37,7 +37,10 @@
 #include <OpenMS/CONCEPT/Types.h>
 #include <OpenMS/OpenMSConfig.h>
 
-#include <QtCore/QDate>
+#include <string>
+
+// foward declarations
+class QDateTime; 
 
 namespace OpenMS
 {
@@ -211,7 +214,7 @@ public:
     void set(const String& date);
 
 private:
-    QDateTime dt_;
+    QDateTime* dt_;
   };
 
 } // namespace OPENMS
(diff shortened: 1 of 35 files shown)


timosachsenberg (approved):

src/openms/include/OpenMS/DATASTRUCTURES/DateTime.h
```diff
@@ -66,17 +67,20 @@ namespace OpenMS
     DateTime();
 
     /// Copy constructor
-    DateTime(const DateTime& date) = default;
+    DateTime(const DateTime& date);
```
timosachsenberg, line 70: can this be noexcept?

src/tests/class_tests/openms/source/ProteaseDigestion_test.cpp
```diff
@@ -72,20 +72,6 @@ START_SECTION(([EXTRA] ProteaseDigestion(const ProteaseDigestion& rhs)))
     TEST_EQUAL(pd.getSpecificity(), pd2.getSpecificity());
 END_SECTION
 
-START_SECTION(([EXTRA] ProteaseDigestion& operator=(const ProteaseDigestion& rhs)))
```
timosachsenberg, line 75: what happened to this one?

src/openms/include/OpenMS/DATASTRUCTURES/DateTime.h
cbielow: technically no, since it allocates.

src/tests/class_tests/openms/source/ProteaseDigestion_test.cpp
cbielow: this operator was never explicitly implemented or used.
And now with unique_ptr it is implicitly deleted.

timosachsenberg: hmm maybe it is needed for pyOpenMS.

src/openms/include/OpenMS/DATASTRUCTURES/DateTime.h
cbielow: where is the advantage of noexcept here?
Its a copy anyway, not a move.

timosachsenberg: probably none... I thought it might have a slight performance impact but don't know for sure

cbielow: mhh.. adding `noexcept` might degrade performance actually. If the compiler cannot prove that no exception can be thrown, then it needs to wrap the function into a try-catch-block which ensures that terminate will be called upon throwing an exception. That is not cheap...

Issue #6003, opened 2022-03-25
Deprecated (doxygen) Doc pages notice does not work


jpfeuffer:
Should be like in the header of https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/release/2.4.0/html/index.html
but on every page.

Currently does not work. The idea was to have the following:
https://github.com/OpenMS/OpenMS/blob/develop/doc/doxygen/common/header.html#L42

So it gets automatically included if it is in the folder. Our jobs should add the "deprecation" file to old releases, whenever there is a new one.

Currently it seems like it needs a JS script to actually include that file. I thought it was an included HTML feature.
Probably requires an adaption of the above header.html.

Amit0617:
however if that's small html part which i expect it would be then it should be directly added there but if you really want it to include from external html file than embed or iframe tags can be used.
Js script can also be used if you want it to happen dynamically for all the html files at once.

jpfeuffer:
It cannot be added directly because the websites are generated automatically. We want to update the htmls only if there is a newer release. And we would need to touch EVERY html file because they all have the same header.

timosachsenberg:
Hi @jpfeuffer can you add some more details here.
Sounds a bit difficult to automatize.
Only way might be through jenkins that adds:
1. the deprecation file to the old releases in the release action
2. figure out how the JS needs to be added such that it gets actually included?

would it make sense to add this JS everywhere and let it query if it is the latest version?

jpfeuffer:
Yes everything works.

jpfeuffer:
Added a version selector instead:
https://github.com/OpenMS/OpenMS/commit/27ae821b1e163f18d44668dedba6b335fc6a47e0

PR #6243, opened 2022-07-14
Feature/awesome css

poshul:
## Description

Update doxygen theme to doxygen-awesome. MIT License for theme is compatible with ours.

## Checklist
- [ X ] Updated or added python bindings for changed or new classes (Tick if no updates were necessary.)

<details>

</details>

<details>
  
</details>

---


jpfeuffer:
Cool. Do you have a preview? Can it display everything in a way that is kind of compatible with the old style?

poshul:
Here's a preview. It works with all the content that I have tested it on. I'd definitely welcome more eyes in case I missed something.
image

jpfeuffer:
No, looks promising indeed. I don't think we have a too complex structure anyway. The only thing is:
I had this custom header HTML, that would inject a placeholder for a "Caution: not the latest docs" textbox (which did not work anyway due to a missing JS file).
If we could make that work, it would be great. We should keep that in mind.
It looks like this theme can also have a customizable header though.

jpfeuffer:
https://github.com/OpenMS/OpenMS/issues/6003

doc/doxygen/doxygen-awesome-css/.github/workflows/publish.yaml
```diff
@@ -0,0 +1,29 @@
+name: publish
```
timosachsenberg, line 1: are these workflows needed?

timosachsenberg: I think we are pulling in too many unneeded files. e.g. github actions, docs, img folder need to go. 
maybe also include/MyLibrary (did not check)

timosachsenberg:
there is some strange THIRDPARTY hash. Might be good to set this to THIRDPARTY master

timosachsenberg:
Looks good. Let’s try!

tapaswenipathak (commented):
wouldn't it be better to add it as a submodule: https://jothepro.github.io/doxygen-awesome-css/#autotoc_md10?

- it is doxygen theme, that has CSS.
- active development
- nicely supported

why does the front page have different colors than the remaining/default? 

Screenshot 2022-07-21 at 12 31 27 PM

poshul:
> wouldn't it be better to add it as a submodule: https://jothepro.github.io/doxygen-awesome-css/#autotoc_md10?
> 
>     * it is doxygen theme, that has CSS.
> 
>     * active development
> 
>     * nicely supported
> 
> 
> why does the front page have different colors than the remaining/default?
> 
> Screenshot 2022-07-21 at 12 31 27 PM

Color was changed to match our theme colors from our logo. We discussed whether to include it as a submodule or just statically and came to the conclusion that we'd rather not have an added external dependency, and that the overall workload of manually making changes if they are needed was minimal. If you very much disagree I would be open to revisiting the decision.

PR #6255, opened 2022-07-19
Several sirius fixes

### LINKED ISSUE #6247: SiriusAdapter sometimes fails with DataValue conversion error
`Error: Unexpected internal error (Could not convert DataValue::EMPTY to double)`
Unfortunately, it does not tell the line. Probably an if-case is missing somewhere, to check for emptyness.

### LINKED ISSUE #6250: Implement log output parsing or storing of log file for SiriusAdapter
I implemented it once but it got lost with the recent restructurings. The problem was last time that the startup script sirius.sh is kind of swallowing the logs.
Also, we need to add the log-level parameter, which is missing.

### LINKED ISSUE #6253: SiriusAdapter should pass the threads parameter as Sirius:processors
To be consistent and not having to specify two parameters for the same thing.

jpfeuffer:
Fixes #6253
fixes #6247 
fixes #6250 

turns out log parsing worked already but only was recorded when sirius completely fails.

### DIFF
--- a/src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
+++ b/src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
@@ -80,6 +80,9 @@ namespace OpenMS
       fingerid.parameters();
       passatutto.parameters();
 
+      defaults_.setValue("read_sirius_stdout", "false", "Read and print the standard output and error of the Sirius executable, even if it succeeds.", {"advanced"});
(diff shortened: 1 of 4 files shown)


poshul:
Any idea what's up with the windows test errors?

poshul (approved):

timosachsenberg:
does this depend on the SIRIUS 5 update?

jpfeuffer:
Yes. This cannot be tested since Sirius tests are broken due to update.
But axel will probably be happy to have no segfaults and debug output while testing.

PR #6278, opened 2022-07-28
fix compile bug

timosachsenberg:
## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms/include/OpenMS/KERNEL/AreaIterator.h
+++ b/src/openms/include/OpenMS/KERNEL/AreaIterator.h
@@ -276,7 +276,7 @@ private:
         {
           // skip over invalid MS levels and Mobility
           while (p_.current_scan_ != p_.end_scan_ &&
-                 (p_.current_scan_->getMSLevel() != (int)p_.ms_level_ ||
+                 ((int8_t)p_.current_scan_->getMSLevel() != p_.ms_level_ ||
                   !mb.containsMobility(p_.current_scan_->getDriftTime())))
           {
             ++p_.current_scan_;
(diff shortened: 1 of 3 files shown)


jpfeuffer:
Can you please leave the param names and just comment them out?

jpfeuffer:
We need a consistent rule about that in our Coding Conventions

timosachsenberg:
I added index back as this is the only one without a telling type

src/openms/include/OpenMS/KERNEL/AreaIterator.h
```diff
@@ -276,7 +276,7 @@ namespace OpenMS
         {
           // skip over invalid MS levels and Mobility
           while (p_.current_scan_ != p_.end_scan_ &&
-                 (p_.current_scan_->getMSLevel() != (int)p_.ms_level_ ||
+                 ((int8_t)p_.current_scan_->getMSLevel() != p_.ms_level_ ||
```
cbielow, line 279: we should maybe use a more robust solution for these kinds of conversion problems (since it will pop up again if we change the type of `current_scan_`.
I actually have a small PR of compile-warning fixes coming up which fixes this exact line. So maybe revert here?!

cbielow (approved):
looks good
Thanks for fixing!

src/openms/include/OpenMS/KERNEL/AreaIterator.h
timosachsenberg: Great. Please don't mind if I just merge this one now to get things working again. Your better fix can still be applied on top I guess.
BTW @jpfeuffer would pyOpenMS have problems with the int8_t?

jpfeuffer: As long as it is not in the function interface, there is no difference between pyopenms and OpenMS

PR #6319, opened 2022-08-30
[Feature] Sirius user account login

axelwalter:
## Description

Since Sirius 5 a login with a personal user account is required. Here, the SiriusAdapter gets new options for the user account email and password. If they are provided, the login command will be executed. The tests and pyopenms run locally with the new login options.

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
+++ b/src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
@@ -538,6 +538,40 @@ namespace OpenMS
     // Algorithm
     // ################
 
+    void SiriusAdapterAlgorithm::logInSiriusAccount(String& executable, const String& email, const String& password) const
+    {
+      if (!email.empty() && !password.empty())
+      {
+        // sirius login --email=email --password=password
+        QString executable_qstring = SiriusAdapterAlgorithm::determineSiriusExecutable(executable).toQString();
+        QStringList command_line{"login", String("--email="+email).toQString(), String("--password="+password).toQString()};
+        
+        // start QProcess with sirius login command
+        QProcess qp;
+        qp.start(executable_qstring, command_line);
+
+        // print executed command as info
+        std::stringstream ss;
+        ss << "Executing command: " << executable_qstring.toStdString();
+        for (const QString& it : command_line)
+        {
+          ss << " " << it.toStdString();
+        }
+        OPENMS_LOG_INFO << ss.str() << std::endl;
+
+        // wait until process finished
+        qp.waitForFinished(-1);
+
+        // always print process stdout (info) and stderror (warning)
+        const QString sirius_stdout(qp.readAllStandardOutput());
+        const QString sirius_stderr(qp.readAllStandardError());
+        OPENMS_LOG_INFO << String(sirius_stdout) << std::endl;
+        OPENMS_LOG_WARN << String(sirius_stderr) << std::endl;
+        
+        qp.close();
+      }
+    }
+
     // tmp_msfile (store), all parameters, out_dir (tmpstructure)
     const std::vector<String> SiriusAdapterAlgorithm::callSiriusQProcess(const String& tmp_ms_file,
                                                                          const String& tmp_out_dir,
--- a/src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
+++ b/src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
@@ -540,36 +540,33 @@ namespace OpenMS
 
     void SiriusAdapterAlgorithm::logInSiriusAccount(String& executable, const String& email, const String& password) const
     {
-      if (!email.empty() && !password.empty())
+      // sirius login --email=email --password=password
+      QString executable_qstring = SiriusAdapterAlgorithm::determineSiriusExecutable(executable).toQString();
+      QStringList command_line{"login", String("--email="+email).toQString(), String("--password="+password).toQString()};
+      
+      // start QProcess with sirius login command
+      QProcess qp;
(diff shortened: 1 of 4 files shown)


src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -538,6 +538,40 @@ namespace OpenMS
...
+    void SiriusAdapterAlgorithm::logInSiriusAccount(String& executable, const String& email, const String& password) const
+    {
+      if (!email.empty() && !password.empty())
+      {
+        // sirius login --email=email --password=password
+        QString executable_qstring = SiriusAdapterAlgorithm::determineSiriusExecutable(executable).toQString();
+        QStringList command_line{"login", String("--email="+email).toQString(), String("--password="+password).toQString()};
+        
+        // start QProcess with sirius login command
+        QProcess qp;
```
timosachsenberg, line 550: Do we have a command in OpenMS that encapsulates the qprocess?

jpfeuffer: No, only in TOPPBase

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -538,6 +538,40 @@ namespace OpenMS
...
+        qp.waitForFinished(-1);
+
+        // always print process stdout (info) and stderror (warning)
+        const QString sirius_stdout(qp.readAllStandardOutput());
+        const QString sirius_stderr(qp.readAllStandardError());
+        OPENMS_LOG_INFO << String(sirius_stdout) << std::endl;
+        OPENMS_LOG_WARN << String(sirius_stderr) << std::endl;
+        
+        qp.close();
+      }
```
timosachsenberg, line 572: Else? Throw exception?

src/utils/SiriusAdapter.cpp
```diff
@@ -227,6 +233,9 @@ class TOPPSiriusAdapter :
       return EXECUTION_OK;
     }
 
+    // log in to Sirius account if email and password are specified
```
timosachsenberg, line 236: Better to check here if password and email provided

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
timosachsenberg: It is unexpected that a function otherwise just does nothing

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
```diff
@@ -540,36 +540,33 @@ namespace OpenMS
 
     void SiriusAdapterAlgorithm::logInSiriusAccount(String& executable, const String& email, const String& password) const
     {
```
timosachsenberg, line 542: There should also be a check and e.g. throw of an exception because this is a public member.

src/utils/SiriusAdapter.cpp
```diff
@@ -234,7 +234,11 @@ class TOPPSiriusAdapter :
     }
 
     // log in to Sirius account if email and password are specified
-    algorithm.logInSiriusAccount(sirius_executable, sirius_user_email, sirius_user_password);
+    if (!sirius_user_email.empty() && !sirius_user_password.empty())
+    {
+      algorithm.logInSiriusAccount(sirius_executable, sirius_user_email, sirius_user_password);
+    }
```
timosachsenberg, line 240: Is this a warning or something we can‘t recover from? If we absolutely need a login then throwing an exception and terminating the program is bettet

jpfeuffer: Unfortunately, it is very unclear when login is needed. Even the Sirius developers don't know it (at least we get Exceptions for searching in specific libraries, although this should be caught when starting Sirius already.)

jpfeuffer: It is definitely not needed all the time.

axelwalter: You need to be logged in, however you can also do that before and independent of SiriusAdapter. A successful login will be saved and is not necessary on subsequent runs. So it should continue if no login information was given.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
axelwalter: Is that required? The email and password need to be passed in the function call anyways?

src/utils/SiriusAdapter.cpp
timosachsenberg: Ah ok thanks for the clarification.

src/openms/source/ANALYSIS/ID/SiriusAdapterAlgorithm.cpp
timosachsenberg: But it can be empty. This is not checked now

timosachsenberg: if this doesn’t crash/error I guess it is fine

timosachsenberg: If used without the check e.g. from pyopenms when empty strings are provided

axelwalter: In that case it will print the output from sirius login that user name and/or password was incorrect.

timosachsenberg (approved):

Issue #6160, opened 2022-06-06
Convert iterator for loops


timosachsenberg:
To range for loops for easier readability.
Many have already been converted but some still remain.

rishi2802:
Hello, Can I contribute to this?

rishi2802:
Hi, Can I contribute to this please?

timosachsenberg:
Sure. Maybe start with a few of them and make a small Pull request so we can give feedback if you are on the right track.

cbielow:
maybe a few things to consider:
1) prefer `const auto` over just `auto` if the variable does not need to be modified.
2) use `auto` if the type is small (roughly 16 bytes or less) , and you do not need to modify the container. Otherwise use `auto&`, i.e. take by reference, to avoid the copy (or enable modification).
Points 1) and 2) can be combined, e.g. `const auto&`.
3) use sensible names, e.g. if the former name of the variable using iterators was `it` (as it usually is), then you should find a better name (since `it` suggests an *iterator*, whereas the new range-based loop variable should be named like the type that underlies the variable, e.g. `spec` (for MSSpectrum) or `p` (for a Peak1D).

made-up example:
**Old**:
```
for (MSExperiment::Iterator it = exp.begin(); it != exp.end(); ++it)
{
  it->setRT(0.0);
}
```

becomes (new)
```
for (auto& spec : exp)
{
  spec.setRT(0.0);
}
```

Sainith123:
is this issue still open.can i have a look into it.

timosachsenberg:
There is already anboten PR. Thanks!

sa-shaheer-ali:
hello can i contribute

cbielow:
> hello can i contribute

sure. Have a look at what was done in https://github.com/OpenMS/OpenMS/pull/6526 and pick up from there.
There are some subtleties to the problem, i.e. blindly converting every loop may be counter productive. See some comments in the PR referenced above. Maybe start with a smaller PR, first.

GoutamIITP:
Hi @timosachsenberg,

I’m interested in contributing to this issue. I see that some work has already been done in #6526. Could you please guide me on which parts still need to be converted? I’d be happy to start with a small PR to ensure I’m following the right approach.

Looking forward to your response!

timosachsenberg:
I will close this one and open a new one

PR #6336, opened 2022-09-11
"auto" changed as "const auto&" as per issue - Convert iterator for loops #6160

rishi2802:
#6160  "auto" has been changed to "const auto&" based on the issue -  "auto" has been changed to "const auto&" based on the issue - Convert iterator for loops https://github.com/OpenMS/OpenMS/issues/6160

src/tests/class_tests/openms/source/ExperimentalDesign_test.cpp

### DIFF
--- a/src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimatorMeanIterative.h
+++ b/src/openms/include/OpenMS/FILTERING/NOISEESTIMATION/SignalToNoiseEstimatorMeanIterative.h
@@ -1,3 +1,66 @@
+------------
+//                   OpenMS -- Open-Source Mass Spectrometry
+// --------------------------------------------------------------------------
+// Copyright The OpenMS Team -- Eberhard Karls University Tuebingen,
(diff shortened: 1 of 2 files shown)


cbielow:
sorry to say but this PR misses the point of the referenced issue.
The issue was about **for-loop**s (`for (...) { ... }`).
There are also some random duplications/deletions in other parts of the file, which prevent this code from compiling.

I will close this PR. Please feel free to ask questions and give it another go.

rishi2802:
I'm so sorry, could you please elaborate the issue a bit, like what should I actually change in the iterator? I'll do it right the next time.
Thank you for your patience :)

cbielow:
no problem.
See the example here: https://github.com/OpenMS/OpenMS/issues/6160#issuecomment-1242942799
The old style is much less readable (there are other instances in real code which are even worse) and is potentially less performant -- depending on compiler and optimization settings.
Hence we'd like to replace the iterator-style for-loop with the new [range-based for-loop ](https://www.nextptr.com/tutorial/ta1208652092/how-cplusplus-rangebased-for-loop-works).

PR #6337, opened 2022-09-13
Feature/new thiols

poshul:
## Description

Add support for identification of ribonucleotides with thiol linkages.

## Checklist

<details>

</details>

<details>
  
</details>

---


timosachsenberg:
I think we need to disable reformatting. also not sure why SIRIUS stuff is affected

poshul:
rebuild jenkins

share/OpenMS/CHEMISTRY/Modomics.tsv
```diff
@@ -180,3 +180,7 @@ deoxycytidine	dC	dC	C			C9O4N3H13	227.0906	227.2176
 deoxyguanosine	dG	dG	G			C10O4N5H13	267.0968	267.2418
 deoxyuridine	dU	dU	U			C9O5N2H12	228.0746	228.2024
 thymidine	dT	dT	U			C10O5N2H14	242.0903	242.2290
+thio-adenosine	A*	A*	A	A	A	C10O4N5H13	267.0968	267.2449
```
timosachsenberg, line 183: - are these names from modomics?
- where is the S? does it need to be in the formula?

poshul: The names are not from modomics, and so if we keep them they should be moved to [Custom_RNA_modifications.tsv](https://github.com/OpenMS/OpenMS/blob/7185e2b4609fee952d60f90e7aae6b9a6c2cf091/share/OpenMS/CHEMISTRY/Custom_RNA_modifications.tsv).

The S is not part of the Nucleo**s**ide which is what populate the tsv's. The logic to add the phosphate linkages looks at the final character of the 'short name' and treats it as a thiol-linkage if that character is a star. 

The bigger question that I'm trying to decide is whether the NASequence parsing code should just populate the database with thiol versions of all of the entries in the tsvs or whether we should need to specify them. Any thoughts?

poshul:
@timosachsenberg I've moved the phosphorothioate versions to Custom_RNA_modifications instead of Modomics. I will be adding the RibonucleotideDB functions to automatically generate thio versions of normal ribos to another PR.

src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp
```diff
@@ -223,84 +215,90 @@ namespace OpenMS
...
+    if ((add_w_ions_ || add_x_ions_ || add_y_ions_ || add_z_ions_) && (oligo.size() > 1))
     {
       fragments_right.resize(oligo.size() - 1);
       fragments_right[0] = ribo_masses.back() + three_prime_mass;
       for (Size i = 1; i < oligo.size() - 1; ++i)
       {
         Size ribo_index = oligo.size() - i - 1;
-        fragments_right[i] = (fragments_right[i - 1] + ribo_masses[ribo_index] +
-                              backbone_mass);
+        fragments_right[i] = (fragments_right[i - 1] + ribo_masses[ribo_index] + backbone_mass + thiols[ribo_index]);
```
timosachsenberg, line 280: there is special treatment of the terminal one in 241. is this also needed here?

src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp
```diff
@@ -351,8 +346,7 @@ namespace OpenMS
     for (Size i = 0; i < size; ++i)
     {
       spectrum.push_back(uncharged_spectrum[i]);
-      spectrum.back().setMZ(abs(spectrum.back().getMZ() / charge +
-                                Constants::PROTON_MASS_U));
+      spectrum.back().setMZ(abs(spectrum.back().getMZ() / charge + Constants::PROTON_MASS_U));
```
timosachsenberg, line 349: don't use abs because of ambiguity with https://cplusplus.com/reference/cstdlib/abs/
use std::abs instead

timosachsenberg:
can you add tests that cover internal but also terminal linkage cases? there seems to be some special handling in the code that looks a bit fragile and could use a test

src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp
poshul: Nope, we use the same `thiols` object here, so the shortening doesn't need to be redone

src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp
poshul: Should I do that for all the instances of abs() in the old code as well?

timosachsenberg: yes it is a common source of errors

poshul: Actually it looks like there's just one other instance in this file where `abs()` is run on a floating point value. I'll just go ahead and clarify that one.

src/openms/source/CHEMISTRY/NucleicAcidSpectrumGenerator.cpp
timosachsenberg: then I don't get the logic. You check for ion type above and then perform the shortening based on what ions you want to generate. And then there is a separate block that reuses it. what if the first one is never executed?

poshul: You are correct, in the event that none of the 5' ion types were selected this would cause an error. I'll move the resize out of the if block.

poshul:
> can you add tests that cover internal but also terminal linkage cases? there seems to be some special handling in the code that looks a bit fragile and could use a test

Added tests to the NucleicAcidSpectrumGenerator class.

timosachsenberg (approved):

PR #6349, opened 2022-09-23
[Feature] MetaboliteSpectralMatcher optional spectra merging and export of native spectrum IDs

axelwalter:
## Description

In MetaboliteSpectralMatcher MS2 spectra with same precursor mass get merged by default. Result is a smaller peak map and reported spectrum indexes that don't match to the original peak map any more.
Added a parameter for optional spectrum merging and optional output of the merged peak map as mzML file.
During spectrum merging, the native IDs of merged spectra get combined in a comma separated consensus native ID instead of taking the native ID of only the first of the merged spectra.
MetaboliteSpectralMatcher now reports the native IDs of observed spectra in a new column.

## Checklist

<details>

</details>

<details>
  
</details>

---


eeko-kon (approved):
Tested the changed files and it works (merging spectra should be explained at documentation- possibly most would avoid it)

axelwalter:
Thanks for testing! Agreed, an explanation would be useful here.

src/tests/class_tests/openms/data/FIAMS_output/SerumTest_10.mzTab
```diff
@@ -1,4 +1,4 @@
-MTD	mzTab-version	1.0.0
-MTD	mzTab-mode	null
-MTD	mzTab-type	null
-MTD	description	null
+MTD	mzTab-version	1.0.0
```
timosachsenberg, line 1: I think this was accidently checked in?

axelwalter: yes, will remove it

src/utils/MetaboliteSpectralMatcher.cpp
```diff
@@ -103,6 +106,8 @@ class TOPPMetaboliteSpectralMatcher :
     setValidFormats_("database", {"mzML", "msp", "mgf"});
     registerOutputFile_("out", "<file>", "", "mzTab file");
     setValidFormats_("out", ListUtils::create<String>("mzTab"));
+    registerOutputFile_("out_merged_spectra", "<file>", "", "Output merged spectra as mzML file if algorithm:merge_spectra is set to true (default).", false);
```
jpfeuffer, line 109: What happens if the outputfile is empty but merge_spectra is true?

Keep in mind that the passing of outputfiles is not obvious in UI based workflow managers. By default, an output file is always given. If it is not created by the tool, the workflow will be in trouble.

axelwalter: In that case the spectra get merged but no output file is written.
The parameter `out_merged_spectra` is set as optional. Does this get handled as an optional output file like in other TOPP tools?

timosachsenberg: One possibility that is more consistent would be to make an optional "out_spectra"  parameter that gets written in both cases (with/without merging).

jpfeuffer: Yes, if wanting both is rare, this is a much better option

axelwalter: Right now it doesn't get written if merging is deactivated. But of course we could write in both cases as suggested. I will comment accordingly in the docs.

timosachsenberg (approved):

PR #6357, opened 2022-09-28
[Fix] SiriusAdapter adaptations to Sirius 5

axelwalter:
## Description

With Sirius 5 some adaptations to the SiriusAdapter where necessary.

Calling CSI:FingerID is done differently now, instead of the `fingerid` command the commands `fingerprint` and `structures` calculate molecular fingerprints and structures now. The number of candidates for CSI:FingerID can not be set individually any more, instead the number of candidates is taken from the sirius formula prediction parameters.

The output gets zipped by default, which lead to the failing of AssayGeneratorMetabo. In Sirius 5.6.0 a new `--no-compression` flag was added to not compress the output. This version is available as a snapshot right now and will be released soon.

Also changed the handling of Sirius parameters. The Sirius 5.6.0 original parameters were added as comments. By default, SiriusAdapter will use all the original parameters now and add parameters the the CLI call only when they have been changed by the user to declutter the CLI command.

All changes pass locally with the new sirius 5.6 snapshot, once the official version is released I will update the THIRDPARTY repository again.
 
## Checklist

<details>

</details>

<details>
  
</details>

---


eeko-kon:
One small comment: I would set the default compound timeout to 100 (s) instead of infinite (-1) like you used to have, otherwise Sirius runs for a day for very simple files (If you have many compounds with very high mass). This is also recommended by the Sirius team. Otherwise, you can mention it at the documentation perhaps as a suggestion?

jpfeuffer:
I think I agree with @eeko-kon 
@axelwalter Ready to review? Then convert to normal non-draft PR.

axelwalter:
Good point with the time outs!
Since the official Sirius 5.6. will be released soon I would wait for that to update our thirdparty repository. The changes in this branch are compatible only with the 5.6 version. What do you think @jpfeuffer ?

jpfeuffer:
Yes, I guess we have to wait. We still need to solve SQL problems for release anyway.

src/tests/topp/THIRDPARTY/third_party_tests.cmake
```diff
@@ -319,10 +319,10 @@ if (NOT (${SIRIUS_BINARY} STREQUAL "SIRIUS_BINARY-NOTFOUND"))
...
+    add_test("TOPP_SiriusAdapter_3" ${TOPP_BIN_PATH}/SiriusAdapter -test -sirius_executable "${SIRIUS_BINARY}" -in ${DATA_DIR_TOPP}/THIRDPARTY/SiriusAdapter_3_input.mzML -in_featureinfo ${DATA_DIR_TOPP}/THIRDPARTY/SiriusAdapter_3_input.featureXML -out_sirius SiriusAdapter_3_output.tmp -preprocessing:filter_by_num_masstraces 3 -sirius:candidates 5 -sirius:profile qtof -sirius:db all -out_annotated_spectra SiriusAdapter_3_output_spec.tmp -read_sirius_stdout)
     add_test("TOPP_SiriusAdapter_3_out" ${DIFF} -in1 SiriusAdapter_3_output.tmp -in2 ${DATA_DIR_TOPP}/THIRDPARTY/SiriusAdapter_3_output.mzTab -whitelist "MTD" "295.18707248574799") # ranks of the compound at "295.18707248574799" m/z shifts depending on OS (unix or windows)
     if (APPLE)  # mzMLs were created on macOS. Since Sirius is nondeterministic we can only test there.
```
jpfeuffer, line 324: Ah did you see my note? Maybe we can just test some of the weirdest tests on one platform. Better than whitelisting everything. Either you do a if (LINUX) or you copy the testfile from the mac CI machine.

poshul: Both Linux and Mac pass now, so either one would work (at least until the next phase of the moon)

axelwalter: Set the spec tests 1 and 3 for Linux only, so I'm able to test them locally.

poshul: Looks like 8 and 9 failed this time on windows.

axelwalter: Yes, there are too many exceptions between the operating systems with the tests 8, 9 and 10 so I removed lot's of whitelisting and made them Linux only.

jpfeuffer: Careful. BSD and macOS are Unix flavors. They will also run.

axelwalter: Hm, check for OS == "Ubuntu"? It get's determined with `lsb_release -si` in the build scripts.

poshul:
So, at this point I'm doubtful that these tests are adding probative value, given how stochastic the outputs are. I think it's time to re-evaluate whether we should keep the failing ones, rather than playing cat-and-mouse with whitelisting.

jpfeuffer:
I would like to see a diff of the files of CI and a local machine.
Is it really indeterministic on the same platform?? I would vote for checking the versions and the input that we write for Sirius then. And if it is Sirius to report it as a bug and disable the test for now.
Did anyone ever check if the results still make sense? E.g. if the compounds at least match the mass?

axelwalter:
rebuild jenkins

axelwalter:
Using the test files generated by the CI macOS works well and I guess it's okay to test the fluctuating tests only on one system.
With the passing tests it would be ready for merging from my side.

poshul (approved):

PR #6358, opened 2022-09-28
fixing docstrings

rahul799:
## Checklist

<details>

</details>

<details>
  
</details>

---


src/pyOpenMS/pxds/FIAMSDataProcessor.pxd
```diff
@@ -22,45 +22,52 @@ cdef extern from "<OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h>" namespace "OpenMS":
...
                 #   - the spectra are summed into one along the time axis with the bin size determined by mz and instrument resolution
                 #   - data is smoothed by applying the Savitzky-Golay filter
                 #   - peaks are picked
                 #   - the accurate mass search for all the picked peaks is performed
-                #
+                #   
                 #   The intermediate summed spectra and picked peaks can be saved to the filesystem. 
                 #   Also, the results of the accurate mass search and the signal-to-noise information 
                 #   of the resulting spectrum is saved.
+                #
```
jpfeuffer, line 37: No parameters??

src/pyOpenMS/pxds/FIAMSDataProcessor.pxd
```diff
@@ -22,45 +22,52 @@ cdef extern from "<OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h>" namespace "OpenMS":
...
+
         FeatureMap convertToFeatureMap(MSSpectrum & input_) nogil except +
             # wrap-doc:
-                #   Convert a spectrum to a feature map with the corresponding polarity
-                #   -----
+                #   Convert a spectrum to a feature map with the corresponding polarity\n
+                #   
                 #   Applies `SavitzkyGolayFilter` and `PeakPickerHiRes`
+                #   
                 #   -----
```
jpfeuffer, line 61: No parameters??

jpfeuffer:
I dont get it. Why is there sometimes Parameters and sometimes not?

src/pyOpenMS/pxds/FIAMSDataProcessor.pxd
```diff
@@ -22,45 +22,52 @@ cdef extern from "<OpenMS/ANALYSIS/ID/FIAMSDataProcessor.h>" namespace "OpenMS":
...
                 #   :param input: Input a picked spectrum
```
rahul799, line 62: Hey @jpfeuffer 
Are you talking about the above parameter?

rahul799: My bad, I understood. Actually both compile to same doc-strings. But I will add the parameter heading to make the template consistent.

axelwalter:
I would suggest either using the Parameter / Return header syntax or the :param / :return prefix and stick to one of them consistently.

jpfeuffer:
Definitely :param: :return:

jpfeuffer:
the keyword is return not returns

axelwalter (approved):

rahul799:
@jpfeuffer @axelwalter can we merge the PR now?

axelwalter:
Is the line with `# ----------` before the parameter section required?
Can we add parameter types?

jpfeuffer:
Was this all tested anyhow?
I think actually the bigger problem are the overloads. They will destroy those nice docstrings anyway.

rahul799:
> Was this all tested anyhow? I think actually the bigger problem are the overloads. They will destroy those nice docstrings anyway.

Yes I tested it in my local machine.

jpfeuffer:
Ok, so how does   BayesianProteinInferenceAlgorithm inferPosteriorProbabilities look like in your docs?

rahul799:
- inferPosteriorProbabilities:
Earlier ->  Screenshot from 2022-10-12 21-54-05
Now -> 
Screenshot from 2022-10-12 21-54-39

jpfeuffer:
Ah nice. What about the second overload? Does it pick up its parameters nicely as well?

rahul799:
Yes it does. I also added newline after the return statement.
Screenshot from 2022-10-13 01-58-05

jpfeuffer:
Ok but the second item/signature should be less indented (i.e. to the left).
We have to fix that in autowrap

axelwalter:
What do you think about omitting the Cython signatures if we have a wrap-doc?

This would require only minor changes in autowrap.
E.g. [here](https://github.com/OpenMS/autowrap/blob/master/autowrap/CodeGenerator.py#L611) for classes and similar for the methods, we could add the Cython stuff to the docstrings only if `extra_doc` (content of wrap-doc) is empty.

jpfeuffer:
But how would the user know which signature they are currently looking at? 
We need something to distinguish them.

jpfeuffer:
And it might help (advanced) users to see when a parameter is const (i.e. if the object will change when passing it to the function).

axelwalter:
True, it can be useful, but also confusing to some. And removing them would require a complete translation into the Python docstrings (which should be the case anyway).

src/pyOpenMS/pxds/BayesianProteinInferenceAlgorithm.pxd
```diff
@@ -93,12 +91,11 @@ cdef extern from "<OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h>" name
         void inferPosteriorProbabilities(ConsensusMap & cmap,
                                          bool greedy_group_resolution) nogil except +
           # wrap-doc:
-          #   Performs inference
-          #   -----
           #   Writes its results into protein and (optionally also) peptide hits (as new score)
```
jpfeuffer, line 94: Since lines seem to be concatenated without newlines in Sphinx, I think we need to add "." (colons) at the end of every line if a sentence ends there.

src/pyOpenMS/pxds/BayesianProteinInferenceAlgorithm.pxd
```diff
@@ -77,13 +76,12 @@ cdef extern from "<OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h>" name
                                          bool greedy_group_resolution,
                                          ExperimentalDesign exp_des) nogil except +
           # wrap-doc:
-          #   Performs inference
-          #   -----
           #   Writes its results into protein and (optionally also) peptide hits (as new score).
```
jpfeuffer, line 79: Too many whitespaces after the "#". Two are substracted for indentation with the wrap- statements. So you are left with another one.

jpfeuffer (changes requested):

src/pyOpenMS/pxds/BayesianProteinInferenceAlgorithm.pxd
jpfeuffer: Probably in many other places as well.

jpfeuffer:
Code snippets need to be escaped:
https://github.com/OpenMS/OpenMS/pull/6358/files#diff-2c24dd8b59090d3cd6ef007766de5c77e90e0b5f5e7df4ba2802ae5fcaae6436L37

jpfeuffer (approved):

PR #6377, opened 2022-10-10
[fix] fix segfault when passing data with no MS1 scans

poshul:
## Description
See above

## Checklist

<details>

</details>

<details>
  
</details>

---


timosachsenberg (commented):
The problem might not only be restricted to the util.
We probably should also output a warning / error in the library if the data is empty.
E.g., throwing an exception

cbielow:
> The problem might not only be restricted to the util. We probably should also output a warning / error in the library if the data is empty. E.g., throwing an exception

IMHO this is the ONLY way to go.
The warning belongs in the algorithm and the result should simply be empty.

src/utils/FeatureFinderMetaboIdent.cpp
```diff
@@ -259,7 +259,12 @@ class TOPPFeatureFinderMetaboIdent :
     mzml.setLogType(log_type_);
     mzml.getOptions().addMSLevel(1);
     mzml.load(in, ff_mident.getMSData());
-
+    if (ff_mident.getMSData().empty())
```
jpfeuffer, line 262: Hmm one could argue that an empty file leads to a successful run.
This might be important in pipeline situations

timosachsenberg:
- does it make sense to include the MS file name annotation (as we do in a successful run)?
- What about the tool? right now it fails for empty MS1. I see three options:
1. fail (I think most were against it)
2. continue with warning
3. fail with error, but continue with warning if force flag set

poshul:
> * does it make sense to include the MS file name annotation (as we do in a successful run)?
> 
>     * What about the tool? right now it fails for empty MS1. I see three options:
> 
> 
>     1. fail (I think most were against it)
> 
>     2. continue with warning
> 
>     3. fail with error, but continue with warning if force flag set

Inside the algorithm code we don't necessarily know the file name the `spectra_file` parameter is optional, and the loading the contents with `.load(...)` doesn't store file name the `FeatureFinderAlgorithmMetaboIdent`.

I still argue that the proper way to handle this in the tool is to error out. We have a specific error type for this `INCOMPATIBLE_INPUT_DATA` and the tool doesn't do anything useful without any MS1 spectra. I am happy to reconsider in the case someone can come up with a concrete use-case where overriding would be helpful.

timosachsenberg:
> Inside the algorithm code we don't necessarily know the file name the `spectra_file` parameter is optional, and the loading the contents with `.load(...)` doesn't store file name the `FeatureFinderAlgorithmMetaboIdent`.

see https://github.com/OpenMS/OpenMS/pull/6377/files#diff-648630a6214ce6708a8b5ec41232d0fbb4d11b9dd71682775585200ae1cd589dR174

> I still argue that the proper way to handle this in the tool is to error out. We have a specific error type for this `INCOMPATIBLE_INPUT_DATA` and the tool doesn't do anything useful without any MS1 spectra. I am happy to reconsider in the case someone can come up with a concrete use-case where overriding would be helpful.

I think @jpfeuffer is referring to situations of thousands of files analyzed in one go, potentially with post-processing scripts that expect output for all files. Not 100% what the correct handling is here. I would probably go with 3. Let the user overwrite the hard fail in such cases (and use the same annotation of file names, e.g. to make scripts work with empty features but with meta data - see above).

src/utils/FeatureFinderMetaboIdent.cpp
```diff
@@ -261,11 +261,11 @@ class TOPPFeatureFinderMetaboIdent :
...
     mzml.load(in, ff_mident.getMSData());
-    if (ff_mident.getMSData().empty() && !force)
+    if (ff_mident.getMSData().empty())
     {
-      OPENMS_LOG_ERROR << "Error: No MS1 scans in '"
-                       << in << "' - aborting." << endl;
-      return INCOMPATIBLE_INPUT_DATA;
+      OPENMS_LOG_WARNING << "Error: No MS1 scans in '"
+                         << in << "' - aborting." << endl;
+      if(!force) return INCOMPATIBLE_INPUT_DATA;
```
poshul, line 268: So now the problem is that we get the message twice in the force case, and its a warn rather than an error.

timosachsenberg: you are right.. I will undo this

cbielow:
IMHO this should work without FORCE, since failing without FORCE will:
1) crash the workflow -- imagine this happening in iteration 999 of 1000 ... very frustrating
2) for the user to restart the workflow and (depending on workflow engine) rerun all the previous data.... also frustrating

timosachsenberg:
I think having no MS1 at all is very rare. It is not like the FF did not find a feature - which is rare but happens. But having no MS1 I never encountered.

poshul:
I think the workflow crash issue is a wider issue with how workflow managers handle exceptions. Asking the tool to find features in a file with no MS1s and therefore no chromatograms is nonsensical in my mind the user implicitly violates a precondition of the program, which is that it has a valid input file. This is logically equivalent in my mind to the program throwing an error if one of the required inputs isn't specified.

I've seen this sort of data before in a couple of cases. Mainly direct infusion, where we know an analyte that we are interested in beforehand, but also in cases where someone upstream filtered for MS2 only (in this case so the MzML file would fit in an email).

poshul:
Can we move forward with this?

jpfeuffer (approved):

PR #6393, opened 2022-10-21
[feature] enable 3' cyclophosphate as a RNA mod

poshul:
## Description
Add support for 2',3' cyclophosphate 3' mods.
## Checklist

<details>

</details>

<details>
  
</details>

---


src/openms/source/ANALYSIS/ID/MetaboliteSpectralMatching.cpp
```diff
@@ -404,7 +404,7 @@ namespace OpenMS
       double db_intensity = 0.0;
       for (const auto& db_it : match.second)
       {
-        db_intensity = max(db_intensity, double(db_it->getIntensity()));
+        db_intensity = max(db_intensity, double(db_it->getIntensity())); //FIXME should this be the sum of all the intensities / number of intensities?
```
timosachsenberg, line 407: Maybe it gets normalized earlier?

timosachsenberg:
Is it possible to test whether all entries can be used? E.g. that changing an entry in the mod file does brake the parsing of hard coded string comparisons?

src/openms/source/ANALYSIS/ID/MetaboliteSpectralMatching.cpp
poshul: Not that I could find. The formal definition of Hyperscore just counts them once (https://www.inf.fu-berlin.de/lehre/WS14/ProteomicsWS14/LUS/lu7b/427/index.html), which isn't especially surprising as this case comes up way less commonly in peptides. I can change this to a less action oriented NOTE in the code.

timosachsenberg: ah I think this is correct. if a spectrum peak matches several db peaks it will take the highest intensity one in the spectra db as reference.

poshul:
> Is it possible to test whether all entries can be used? E.g. that changing an entry in the mod file does brake the parsing of hard coded string comparisons?
Just added this.

timosachsenberg (approved):
thanks for adding the test!

PR #6424, opened 2022-10-31
Make RibonucleotideDB initialization threadsafe

poshul:
## Checklist

<details>

</details>

<details>
  
</details>

---


src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -43,21 +43,34 @@ using namespace std;
...
 namespace OpenMS
 {
-  RibonucleotideDB::RibonucleotideDB():
-    max_code_length_(0)
+
+  RibonucleotideDB* RibonucleotideDB::getInstance()
   {
-    readFromFile_("CHEMISTRY/Modomics.tsv");
-    readFromFile_("CHEMISTRY/Custom_RNA_modifications.tsv");
+    static RibonucleotideDB* db_ = RibonucleotideDB::initializeRibonucleotideDB();
```
cbielow, line 49: this heap object will never be destroyed now (not even when the app closes). Is that intended?

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -43,21 +43,34 @@ using namespace std;
...
+    return db_;
   }
 
-  RibonucleotideDB* RibonucleotideDB::getInstance()
+  RibonucleotideDB* RibonucleotideDB::initializeRibonucleotideDB(OpenMS::String modomics_file, OpenMS::String custom_mods_file, OpenMS::String proprietary_mods_file)
+  {
+    static RibonucleotideDB* db_ = new RibonucleotideDB(modomics_file, custom_mods_file, proprietary_mods_file);
```
cbielow, line 55: this is not threadsafe.
i.e. multiple threads could still create a new object, but only one will succeed.
You either need https://en.cppreference.com/w/cpp/thread/call_once or make it a 
`static RibonucleotideDB db_(modomics_file, custom_mods_file, proprietary_mods_file);`

You also need to hide the copy and assignment operators (which are defaulted and broken at the moment). Otherwise a user can copy the object obtained using getInstance().

timosachsenberg: having a static RibonucleotideDB is better but 
wouldn't it also be fine to move get rid of the initialize call and directly do the new in getInstance() ?  
https://stackoverflow.com/questions/70643880/leaky-meyers-singleton-is-it-threadsafe

timosachsenberg: so we need to do something like:
https://www.modernescpp.com/index.php/thread-safe-initialization-of-data
```
class MySingleton{

  private:
    static std::once_flag initInstanceFlag;
    static MySingleton* instance;
    MySingleton()= default;
    ~MySingleton()= default;

  public:
    MySingleton(const MySingleton&)= delete;
    MySingleton& operator=(const MySingleton&)= delete;

    static MySingleton* getInstance(){
      std::call_once(initInstanceFlag,MySingleton::initSingleton);
      return instance;
    }

    static void initSingleton(){
      instance= new MySingleton();
    }
};

MySingleton* MySingleton::instance= nullptr;
std::once_flag MySingleton::initInstanceFlag;
```

timosachsenberg: or easier: don't use the pointer and static initialize

cbielow: > and this would also be wrong? https://github.com/OpenMS/OpenMS/blob/develop/src/openms/source/CHEMISTRY/ModificationsDB.cpp#L81-L97

yes. same problem

cbielow: > so we need to do something like: https://www.modernescpp.com/index.php/thread-safe-initialization-of-data
> 
> ```
> class MySingleton{
> 
>   private:
>     static std::once_flag initInstanceFlag;
>     static MySingleton* instance;
>     MySingleton()= default;
>     ~MySingleton()= default;
> 
>   public:
>     MySingleton(const MySingleton&)= delete;
>     MySingleton& operator=(const MySingleton&)= delete;
> 
>     static MySingleton* getInstance(){
>       std::call_once(initInstanceFlag,MySingleton::initSingleton);
>       return instance;
>     }
> 
>     static void initSingleton(){
>       instance= new MySingleton();
>     }
> };
> 
> MySingleton* MySingleton::instance= nullptr;
> std::once_flag MySingleton::initInstanceFlag;
> ```

why is 'static void initSingleton()' publicly accessible? That looks wrong...
just ditch the pointers and do
```
static const MySingleton& getInstance(){
      static MySingleton instance(a,b,c);
      return instance;
    }
```

poshul: What's the advantage of ditching the pointers? We have a couple of `*DB` classes, so I would want to keep the semantics between them similar.

cbielow: they are also not threadsafe.
But if you really want a pointer, then just do
```
static const MySingleton* getInstance(){
      static MySingleton instance(a,b,c);
      return *&instance;
    }
```
to get the old interface and thread-safety

poshul:
Since this is an issue with how we implement several of our DB classes Im closing this and opening a new PR that addresses all of them.

PR #6426, opened 2022-10-31
Escape underscores and other special sphinx characters

rahul799:
Ref - https://github.com/OpenMS/pyopenms-docs/issues/306

@jpfeuffer I have checked all the files and found only one case.


jpfeuffer:
Ok. Did you check ror other special characters as well?

rahul799:
Hi @jpfeuffer,
Yes, I checked but didn't found any.

PR #6428, opened 2022-10-31
Feature/clang performance

timosachsenberg:
## Checklist

<details>

</details>

<details>
  
</details>

---


timosachsenberg:
needs update with other branch...

timosachsenberg:
Tests look good. Not urgent but if one of you have time to also give it a second look I think it would be great.
@jpfeuffer do you see problems here with our pyOpenMS bindings?

src/openms/include/OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h
```diff
@@ -115,7 +115,7 @@ namespace OpenMS
         std::vector<ProteinIdentification>& proteinIDs,
         std::vector<PeptideIdentification>& peptideIDs,
         bool greedy_group_resolution,
-        std::optional<const ExperimentalDesign> exp_des = std::optional<const ExperimentalDesign>());
+        const std::optional<const ExperimentalDesign>& exp_des = std::optional<const ExperimentalDesign>());
```
cbielow, line 118: this can be shortened to 
```suggestion
        const std::optional<const ExperimentalDesign> exp_des;
```
Not sure how many of these exist...

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -183,7 +183,7 @@ namespace OpenMS
     void calculateDIAScores(OpenSwath::IMRMFeature* imrmfeature,
                             const std::vector<TransitionType>& transitions,
                             const std::vector<OpenSwath::SwathMap>& swath_maps,
-                            OpenSwath::SpectrumAccessPtr ms1_map,
+                            const OpenSwath::SpectrumAccessPtr& ms1_map,
```
cbielow, line 186: I think the general rule here is not to pass smartPtrs by const ref, but pass the underlying object as const ref instead.
But its an improvement, so let's leave it like it is for now.

src/openms/include/OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h
```diff
@@ -131,7 +131,7 @@ namespace OpenMS
     void inferPosteriorProbabilities(
         ConsensusMap& cmap,
         bool greedy_group_resolution,
-        std::optional<const ExperimentalDesign> exp_des = std::optional<const ExperimentalDesign>());
+        const std::optional<const ExperimentalDesign>& exp_des = std::optional<const ExperimentalDesign>());
```
cbielow, line 134:
```suggestion
        const std::optional<const ExperimentalDesign>exp_des;
```
turns out, only those two :)

cbielow (changes requested):
looks good!
Two minor things (just 2 clicks)

src/openms/include/OpenMS/ANALYSIS/ID/BayesianProteinInferenceAlgorithm.h
timosachsenberg: great! thanks. I will check. Maybe there is a clang-tidy fix for this as well :)

jpfeuffer:
Std::optional is funny right.. spent some time trying when I implemented it as well.

cbielow:
I completely missed that this a default parameter, not an inline expression...
Nothing else you can do in this case than leave it like it was. Sorry for the confusion

PR #6430, opened 2022-11-02
pLFQ keep only best scoring ID per consensus feature

timosachsenberg:
fixes a regression

## Description

pLFQ keep only best scoring ID per consensus feature. Otherwise downstream algorithms fail.

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/utils/ProteomicsLFQ.cpp
+++ b/src/utils/ProteomicsLFQ.cpp
@@ -1758,6 +1758,9 @@ protected:
 
     // clean up references (assigned and unassigned)
     IDFilter::removeUnreferencedProteins(consensus, true);
+    
+    // only keep best scoring ID for each consensus feature
+    IDConflictResolverAlgorithm::resolve(consensus);
 
     //-------------------------------------------------------------
     // Peptide quantification


jpfeuffer:
What does resolver do? Add as unassigned or remove? I think it might influence inference and FDR in different ways.

timosachsenberg:
I think it makes it unassigned

jpfeuffer:
Ok, this should actually be closer to the original way. I think FDR takes all PSMs per feature but I am not sure about inference. I think there only the best per feature.

jpfeuffer:
Just update the files with the script and have a look at the diff

timosachsenberg:
1. why should inference use only the quantified ones? I think no other software does it like that.
2. should I run it earlier e.g., before inferProteinGroups_

jpfeuffer:
1. It does not, unless you set the flag.
2. No everything is fine when it puts the PSMs to unassigned

jpfeuffer:
Unless.. does it remove every sequence or only non-matching ones?

timosachsenberg:
bool keep_matching = false is the default

so it will always reduce to one

jpfeuffer:
ok if everything is one and the rest is in unassigned, the behavior should be the same as before.

jpfeuffer:
you can run the update_test script under tools

jpfeuffer (approved):

PR #6431, opened 2022-11-02
Fix thread safety of DB classes

poshul:
## Description

This makes the singleton construction of the DB classes thread-safe.

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/ModificationsDB.h
@@ -79,9 +79,6 @@ public:
     /// Returns a pointer to the modifications DB (singleton)
     static ModificationsDB* getInstance();
 
-    /// Initializes the modification DB with non-default modification files (can only be done once)
-    static ModificationsDB* initializeModificationsDB(OpenMS::String unimod_file = "CHEMISTRY/unimod.xml", OpenMS::String psimod_file = "CHEMISTRY/PSI-MOD.obo", OpenMS::String xlmod_file = "CHEMISTRY/XLMOD.obo");
-
     /// Check whether ModificationsDB was instantiated before
     static bool isInstantiated();
 
--- a/src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
+++ b/src/openms/include/OpenMS/CHEMISTRY/RibonucleotideDB.h
@@ -58,9 +58,6 @@ namespace OpenMS
     /// replacement for constructor (singleton pattern)
     static RibonucleotideDB* getInstance();
 
-    /// destructor
-    virtual ~RibonucleotideDB();
-
     /// copy constructor not available
     RibonucleotideDB(const RibonucleotideDB& other) = delete;
 
@@ -102,8 +99,6 @@ namespace OpenMS
 
 
   protected:
-    /// default constructor
-    RibonucleotideDB();
 
     /// read (modified) nucleotides from input file
     void readFromFile_(const std::string& path);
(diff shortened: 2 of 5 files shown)


poshul:
Unfortunately `initializeModificationsDB` gets called directly in `OpenSwathAssayGenerator` due to support for custom unimod files. This is problematic, and I think it means that we have the same issue that @cbielow identified in my earlier PR. @timosachsenberg  and I spent some time yesterday brainstorming, but would welcome any input.

timosachsenberg:
one possibilty would be to make parsing of modifications dependent on an environment variable.
That way users could overwrite the use of modifications db on a global level but also down to single processes

cbielow:
> Unfortunately `initializeModificationsDB` gets called directly in `OpenSwathAssayGenerator` due to support for custom unimod files. This is problematic, and I think it means that we have the same issue that @cbielow identified in my earlier PR. @timosachsenberg and I spent some time yesterday brainstorming, but would welcome any input.

Where exactly is the problem?
This can still be done imho.

timosachsenberg:
I think it is about making initializeModificationsDB  private.
If it is private one needs to pass the custom modifications name before the singleton get's constructed.
One way would be (process level) global variables another one environment variables

cbielow:
this can be done.
Small codeup here:
https://godbolt.org/z/sdM578qEc

``` cpp
class ModificationsDB
{
    public:
        /// allows to pass the source files to create the singleton. Only honored upon first call to this function. See 'isInitialized()'
        static const ModificationsDB* getInstance(const std::string& source)
        {
            static ModificationsDB instance(source);
            ModificationsDB::is_initialized_ = true;
            return &instance;
        }
        /// query of getInstance() will create the singleton or simply return the existing instance
        static bool isInitialized()
        {
            return is_initialized_;
        }

    private:
        ModificationsDB(const std::string& source)
        {
            // ...
        }

        static inline bool is_initialized_ = false; ///< true if getInstance() was called before
};

int main() {
  
  std::cout << "Is intialized?" << ModificationsDB::isInitialized() << "\n";
  auto inst = ModificationsDB::getInstance("special.xml");
  std::cout << "Is intialized?" << ModificationsDB::isInitialized() << "\n";
  auto inst2 = ModificationsDB::getInstance("other.xml"); // ignores other.xml

  return 0;
}
```

This is not complete (needs hiding some move and assignment ctors etc). But it works :)

timosachsenberg:
sure that of course also works :) I have a bit of an aversion to this extra semantics and we would be stuck with the process level customizability

cbielow:
We can think of using environment variables, but this will make it very hard to reproduce the workflow results on another machine, since you never know what DB was loaded... its not part of the workflow config. And I'm not even sure that all workflow engines support this. This is a can I'm not sure we should open unless we have detailed information for at least the most prominent engines.

timosachsenberg:
@cbielow https://github.com/OpenMS/OpenMS/blob/develop/src/openms_gui/source/VISUAL/TVToolDiscovery.cpp#L56
here I used an immediate evaluated lambda. Am I right that this is indeed not thread-safe? Because I thought we also added it to protect for concurrent initialization.

cbielow:
mhh.. I invenstigated a bit more, and AFAIKT, even the pointer implementation **was** threadsafe! *!!*
https://godbolt.org/z/YaGbfh84o
At least modern compilers protect the whole expression, not just the assignment.
This is nice, but also means our old implementation of using pointers is actually fine, as is the lamdba.
Sorry for the confusion :/

timosachsenberg:
that is good news indeed :)
and thanks for checking in compiler explorer

PR #6446, opened 2022-11-07
restrict number of spawned processes

timosachsenberg:
## Description

In TOPPView tool discovery

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms_gui/include/OpenMS/VISUAL/APPLICATIONS/TOPPViewBase.h
+++ b/src/openms_gui/include/OpenMS/VISUAL/APPLICATIONS/TOPPViewBase.h
@@ -447,9 +447,12 @@ protected:
     LogWindow* log_;
 
     /// Determines TVToolDiscovery scans for tool/utils and generates new params.
-    TOOL_SCAN scan_mode_;
+    TOOL_SCAN scan_mode_;    
     /// Scans for tools/utils and generates a param for each.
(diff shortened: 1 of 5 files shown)


jpfeuffer:
Do you think it makes sense to add a debug mode for TOPPView, where it tells you, at least during loading, what is happening?

timosachsenberg:
I will check.

poshul:
It looks like we are seeing Sirius test failures here again. No idea why these changes would cause them. @axelwalter any thoughts?

timosachsenberg:
@jpfeuffer I merge this as a fix but I already added the flag to pass verbosity to TOPPView.
Currently it is not used in TOPPToolDiscovery but I could easily add some information.
Question is how we will output that (e.g., we probably can write to std::cout / use our OPENMS_ macros) writing to log files
is probably not that simple because the components are not tied to the logging mechanism.

jpfeuffer:
But didn't you say it failed because of file handles? This is not the same error that is encountered as in the issue.
So I would say the chances that this fixes anything are slim. And tomorrow after the user kindly downloads, extracts, tests etc the tool, we are not any smarter than before. So I wanted to take this chance to be able to see what is going on. Alternative is: you guide them through an lldb debugging session.

jpfeuffer:
I think cout is fine although I don't understand "the components are not tied to the logging mechanism".

Issue #4648, opened 2020-04-16
No distinction between double and float CVs possible


jpfeuffer:
This is very bad, since other libraries depend on the precision with which numbers are written
if specified specifically by a schema.
https://github.com/jpfeuffer/OpenMS/blob/380516006d6f7bd08a797768e4c92b0da4dbb165/src/openms/include/OpenMS/FORMAT/ControlledVocabulary.h#L67-L67

timosachsenberg:
Is this really a problem?
As I understood it, the precision of the value is independent from its lexical representation.
So even if the precision is low, the third-party library should still be able to parse a lexical representation with additional non-significant digits.

jpfeuffer:
If it's a good library, yes.

poshul:
My proposal is that we limit the length of `XSD_DECIMAL` output to be compliant with float (to handle library parsing issues), and circle back around to this for 3.0 with a proper fix. A little bit of digging shows that since `XS_FLOAT` and `XS_DOUBLE` are not derived from `XS_DECIMAL` making changes to how we output XML would break compatibility with current and past versions. I do believe that making this change eventually is the correct thing to do since at least MzML does differentiate between the two lengths.

timosachsenberg:
Maybe we could add the other types to ControlledVocabulary.h, query that information in our XML handlers and write them out accordingly.
BTW can you link to the max number of digits that may be written out?

timosachsenberg:
I would vote to make this a minor issue as it only caused a problem once in one specific deprecated java tool so far.
Or did I miss anything of importance?

jpfeuffer:
No, it was just marked for this release. Can be minor due to low practical importance.

PR #6454, opened 2022-11-08
Fix XSD value parsing in XML files

cbielow:
## Description

I stumbled upon this when loading a exotic mzML, which had something like
```
<userParam name="someInt" type="xsd:integer" value="9223372036854775807"/>
```

It turns out, our implementation only supported ranges covered by signed 32bit integers.
Any larger numbers would throw an Exception::ConversionError.

Also, anyone passing a "xsd:decimal" would also get an error, since we treated it as an integer, not a decimal number (float, double). Luckily, no one ever tried!

This PR fixes most of these things in **mzML**, **mzQuantML** and **traML**, but our internal representation of integers is an Int64, i.e. if an mzML file contains a large 64 bit unsigned int
```
<userParam name="aLargeUnsigned64bit" type="xsd:unsignedLong" value="9223372036854775809"/>
```
that's still an error. Fixing this would mean fixing `DataValue`... good luck.

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms/include/OpenMS/DATASTRUCTURES/ListUtils.h
+++ b/src/openms/include/OpenMS/DATASTRUCTURES/ListUtils.h
@@ -257,9 +257,9 @@ public:
     T convert(const String& s);
   
     template<>
-    inline Int convert(const String& s)
+    inline Int32 convert(const String& s)
     {
-      return s.toInt();
+      return s.toInt32();
     }
     template<>
     inline double convert(const String& s)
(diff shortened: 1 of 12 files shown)


jpfeuffer:
thanks. I had this problem before. Another problem is very related: https://github.com/OpenMS/OpenMS/issues/4648
In case you care about CVs, too.

timosachsenberg (approved):

PR #6461, opened 2022-11-10
Fix boolean logic

jpfeuffer:
## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/cmake/MacOSX/fix_dependencies.rb
+++ b/cmake/MacOSX/fix_dependencies.rb
@@ -432,7 +432,7 @@ Dir.chdir($lib_dir.to_s) do
   lib_files = Dir.glob(["*","*/*"])
   for content in lib_files
     # in case plugin dir is a subdir of lib dir
-    if not ($plugin_dir.nil? and ($lib_dir+content).to_s.start_with?($plugin_dir.to_s)) and fixable(content, $lib_dir)
+    if ($plugin_dir.nil? or not ($lib_dir+content).to_s.start_with?($plugin_dir.to_s)) and fixable(content, $lib_dir)
       if isFramework(content)
         handleFramework($lib_dir + content, $lib_dir, [])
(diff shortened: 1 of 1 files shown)


cmake/MacOSX/fix_dependencies.rb
```diff
@@ -432,7 +432,7 @@ def handleBinary(binaryPath)
   lib_files = Dir.glob(["*","*/*"])
   for content in lib_files
     # in case plugin dir is a subdir of lib dir
-    if not ($plugin_dir.nil? and ($lib_dir+content).to_s.start_with?($plugin_dir.to_s)) and fixable(content, $lib_dir)
+    if ($plugin_dir.nil? or not ($lib_dir+content).to_s.start_with?($plugin_dir.to_s)) and fixable(content, $lib_dir)
```
poshul, line 435: Isn't this also in case plugin dir is undefined now as well?

jpfeuffer: If plugin dir is null the first check will be skipped. Only if it is not null you additionally need to check that the files are not in the plugin dir in case plugin dir is a subfolder.

poshul: Which is the first check? My Ruby is a bit rusty, but this looks like it will enter the loop iff the content if fixable, and the plugin dir is null, or the plugin dir is not a subdir.

jpfeuffer: sounds correct, right? if the plugin dir is null, start_with will always be true. so you need that additional check.

poshul: Ah. Okay. That makes sense. I was thinking of null as in `/0` terminating C++ strings.

poshul (approved):

PR #6478, opened 2022-11-17
Switch from QtSql to SQLiteCPP

cbielow:
## Description

Rewrite of OMSFileLoad and OMSFileStore to use the internal SQLiteCPP library instead of Qt's.

Fixes a few things in the implementation which Qt just ignored (e.g. nested transactions) or creating indices on non-existing tables.
Performance has improved (see below). Also correctness seems better since the benchmark below crashes for the old version on Windows. The new implementation runs fine.

Here are some benchmarks on debian/gcc. Input is a 60 MB OMS file (which I converted from a 446 MB idXML)
`old`
```
(base) ...w/openms_build $ ./bin/IDFileConverter -in ../../../tmp/OMS_bench/bla_qt.oms -out ../../../tmp/OMS_bench/bla_qt2.oms
Progress of 'Loading...':
-- done [took 5.39 s (CPU), 5.40 s (Wall)] --
Progress of 'Storing...':
-- done [took 5.16 s (CPU), 5.71 s (Wall)] --
IDFileConverter took 11.26 s (wall), 10.70 s (CPU), 0.22 s (system), 10.48 s (user); Peak Memory Usage: 188 MB.
```

`new`:
```
(base) ...w/openms_build $ ./bin/IDFileConverter -in ../../../tmp/OMS_bench/bla_qt.oms -out ../../../tmp/OMS_bench/bla_new2.oms
Progress of 'Loading...':
-- done [took 1.99 s (CPU), 2.00 s (Wall)] --
Progress of 'Storing...':
-- done [took 3.66 s (CPU), 3.93 s (Wall)] --
IDFileConverter took 6.02 s (wall), 5.75 s (CPU), 0.21 s (system), 5.54 s (user); Peak Memory Usage: 186 MB.
```

Thus, **read performance at factor 2.5x better, write performance at factor 1.4x better!** comparing the old QtSql implementation vs. [SQLiteCPP](https://github.com/SRombauts/SQLiteCpp). (@SRombauts may be happy to hear :) )
Note: I did not change the algorithm (apart from using std::move where necessitated by the new interface). This is just a conversion from one interface to another at this point.

## Checklist

<details>

</details>

<details>
  
</details>

---


timosachsenberg:
@enetz can you run this through your benchmarking code you did when simplifying the schema?

poshul:
Do we need to include the example code in third_parties?

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
```diff
@@ -29,20 +29,19 @@
...
 
 #include <OpenMS/CONCEPT/ProgressLogger.h>
 #include <OpenMS/KERNEL/FeatureMap.h>
 #include <OpenMS/METADATA/ID/IdentificationData.h>
+#include <OpenMS/FORMAT/OMSFileStore.h>
 
-#include <QtCore/QJsonArray> // for JSON export
-#include <QString>
-#include <QStringList> // avoid compiler error on initialization using init. list
+#include <SQLiteCpp/Database.h>
```
timosachsenberg, line 42: could this be replaced by forward declarations?

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
```diff
@@ -119,40 +118,40 @@ namespace OpenMS
...
 
-      void createView_(const QString& name, const QString& select);
+      void createView_(const String& name, const String& select);
 
       QJsonArray exportTableToJSON_(const QString& table, const QString& order_by);
 
-      // store name, not database connection itself (see https://stackoverflow.com/a/55200682):
-      QString db_name_;
+      /// The database connection (read)
+      SQLite::Database db_;
```
timosachsenberg, line 154: would probably need a pointer type here

src/openms/include/OpenMS/FORMAT/OMSFileLoad.h
jpfeuffer: Was about to ask, then it could stay a private dependency.

cbielow:
yes, replaced with PiMPL. Benchmark results did not change fortunately.

timosachsenberg:
Is there a way to just fetch that code instead of including it?
I have no clue but some packages do something like
```
include(FetchContent)
FetchContent_Declare(
  xxx
  GIT_REPOSITORY xxxx
  GIT_TAG        vxxxx
)
FetchContent_MakeAvailable(xxxx)
```

cbielow:
> Is there a way to just fetch that code instead of including it? I have no clue but some packages do something like
> 
> ```
> include(FetchContent)
> FetchContent_Declare(
>   xxx
>   GIT_REPOSITORY xxxx
>   GIT_TAG        vxxxx
> )
> FetchContent_MakeAvailable(xxxx)
> ```

Never tried actually.
If we move in this direction it would make sense to be consistent here IMHO, and do that with the other thirdparty libs as well.
But this is out of scope of this PR.

cbielow:
> Do we need to include the example code in third_parties?

Does not hurt (apart from a few KB), and may be a nice reference for lookup in later sqlite implementation efforts.

cbielow:
ready for merge. Tests pass (even though Windows CI seems down).
Any more comments/TODO's?

timosachsenberg (approved):
so libsqlite is now also included through thirdparty (not the contrib anymore)?

cbielow:
> so libsqlite is now also included through thirdparty (not the contrib anymore)?

Yes. So we could remove it from contrib as well.

Other things which could be done (other PR):
 - remove our own sqliteadapter and replace with SQLiteCPP (mostly affects openswath code)

timosachsenberg:
k I think this is good to go. I can remove slite from contrib as well as I got a PR open anyways.

SRombauts:
Thanks @cbielow from pinging me, I am delighted to see your work has been merged, and as such SQLiteCpp

PR #6479, opened 2022-11-17
Speedup Ion Mobility OpenSwath Processing

jcharkow:
## Description

3-4 fold speedup of OpenSwathWorkflow.

Background: Filtering spectra by their ion mobility is very computationally expensive. Currently DIA scores are "blind" to the ion mobility dimension and thus require spectra to be filtered by ion mobility. Since the spectra are not sorted by ion mobility (rather by m/z) this is very computationally expensive. 
 
Solution:  Instead of filtering spectra examine only the regions of interest and compute scores from these. This takes advantage of the fact that the spectra are sorted by their m/z. 

This PR involves creating new functions and deprecating old ones however it should not affect the output. 

### Some of the functions augmented include:
- integrateWindow + integrateWindows = Now support integration of an ion mobility enhanced window. Accomplishes this adding drift_start and drift_end arguments. This family of function all rely on a new function _integrateWindowHelper(). This function can now integrate across a vector of spectra. This means that the spectra do not need to be added up by the expensive _addUpSpectra() function. 
- add DIA Scoring functions can now handle ion mobility
- computeIonMobilogram() (from IonMobilityScoring) = Computes ion mobilogram in more efficient manner than previous filterByDriftSpectrum
- fetchSpectrumSwath now does not automatically call _addUpSpectra() (which can be quite computationally expensive with ion mobility as it calls filterByDrift())

### Some Functions Removed include:
- DIAHelpers::integrateDriftSpectrum = not needed because the base function now supports integrating spectra.

### Tests:
- integrateWindow tests were added.
- OpenSwathTest 23_b was added to test the -add_up_spectra flag

### Issues
- With the new implementation this removes support for resampling spectra addition with ion mobility 
  - Possibly deprecate resampling spectra? 
- Currently SONAR still uses the old implementation of adding up spectra 
- 

## Checklist

<details>

</details>

<details>
  
</details>

---

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203581271718537


src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -58,143 +58,290 @@ namespace OpenMS::DIAHelpers
...
+        std::cerr << "Warning: Cannot integrate if spectrum is empty" << std::endl;
+        return;
+      }
+
+      // drift time checks
+      if ( drift_start >= 0 ) // integrate across im as well
+      {
+        // additional checks only relevant if ion mobility is present
+
+        //TODO this preconditions do not seem to be working? however the if statement below works
```
hroest, line 104: @jcharkow preconditions are only enabled in Debug mode, these are not supposed to be runtime checks but rather logical checks that are enabled during testing of the code only but are too expensive to be used in production code.

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -58,143 +58,290 @@ namespace OpenMS::DIAHelpers
...
+        OPENMS_PRECONDITION(spectrum->getDriftTimeArray() != nullptr, "Cannot integrate with drift time if no drift time is available.");
+        OPENMS_PRECONDITION(spectrum->getMZArray()->data.size() == spectrum->getDriftTimeArray()->data.size(), "MZ and Drift Time array need to have the same length.");
+
+        if (spectrum->getDriftTimeArray() == nullptr)
+        {
+          std::cerr << "Warning: Cannot integrate with drift time if no drift time is available.\n";
+          return;
```
hroest, line 111: is this a good way to deal with the error? what happens now to the function calling? should we instead throw an exception?

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -58,143 +58,290 @@ namespace OpenMS::DIAHelpers
       {
         left -= mz_extract_window / 2.0;
         right += mz_extract_window / 2.0;
+
+        // If the left value is now < 0, this is invalid correct it to be 0
+        if (left < 0)
+        {
+          left = 0;
+        }
```
hroest, line 66: why not move this outside the if statement?

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -58,143 +58,290 @@ namespace OpenMS::DIAHelpers
...
       else
       {
-        // not implemented
-        throw Exception::NotImplemented(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION);
+        // if (all_spectra.empty())
+        std::cerr << "WARNING: No Spectra provided" << std::endl;
+        im = -1;
+        mz = -1;
+        intensity = 0;
+        return false;
```
hroest, line 344: is this a good way to deal with errors?

hroest (changes requested):
looks good overall but there are some changes required

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -164,33 +164,30 @@ namespace OpenMS
...
-    OpenSwath::SpectrumPtr spectrum = fetchSpectrumSwath(used_swath_maps, imrmfeature->getRT(), add_up_spectra_, drift_lower, drift_upper);
+    std::vector<OpenSwath::SpectrumPtr> spectra = fetchSpectrumSwath(used_swath_maps, imrmfeature->getRT(), add_up_spectra_, drift_lower, drift_upper);
 
-    // calculate drift extraction width for current spectrum (with some extra for cross-correlation)
-    double drift_width = fabs(drift_upper - drift_lower);
-    double drift_lower_used = drift_lower - drift_width * im_drift_extra_pcnt_;
-    double drift_upper_used = drift_upper + drift_width * im_drift_extra_pcnt_;
+    // set the DIA parameters
+    double dia_extract_window_ = (double)diascoring.getParameters().getValue("dia_extraction_window");
+    bool dia_extraction_ppm_ = diascoring.getParameters().getValue("dia_extraction_unit") == "ppm";
```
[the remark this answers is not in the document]
jcharkow: Not quite sure what the best way to cache this would be. It is already in the outer loop of the function. Should I chance these to public/protected variables? Should they be arguments for the function?

jcharkow:
@hroest it seems that when running in debug mode OpenSwathWorkflow_17 test fails because of the OPENMS_PRECONDITION specifying that the drift_time_array cannot be empty when drift time is set. This is a problem because OpenSwathWorkflow_17 does not have drift time information for MS1. For now I've updated the test to disable ms1 (`enable_ms1 false`) as seen here [9b7f374](https://github.com/OpenMS/OpenMS/pull/6479/commits/9b7f3741a2e502cc4bf4d688dbb0b92c3642049e) however I don't know if that is the best fix for the long term.

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -58,143 +58,290 @@ namespace OpenMS::DIAHelpers
...
+                bool centroided)
+    {
+      // TODO are preconditions working?
+      OPENMS_PRECONDITION(spectrum != nullptr, "Spectrum cannot be nullptr");
+      OPENMS_PRECONDITION(spectrum->getMZArray() != nullptr, "Cannot integrate if no m/z is available.");
+      OPENMS_PRECONDITION(spectrum->getMZArray()->empty(), "Cannot integrate if no m/z is available.");
+      OPENMS_PRECONDITION(std::adjacent_find(spectrum->getMZArray()->data.begin(),
+        spectrum->getMZArray()->data.end(), std::greater<double>()) == spectrum->getMZArray()->data.end(),
+        "Precondition violated: m/z vector needs to be sorted!" );
+      OPENMS_PRECONDITION(spectrum->getMZArray()->data.size() == spectrum->getIntensityArray()->data.size(), "MZ and Intensity array need to have the same length.");
```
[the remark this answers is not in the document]
jcharkow: This has been addressed. Not letting me dismiss suggestion though

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
hroest: looks good

hroest: use LOG_WARN or LOG_ERROR instead of `cerr`

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -59,142 +59,282 @@ namespace OpenMS::DIAHelpers
...
+        std::cerr << "Warning: Cannot integrate if spectrum is empty" << std::endl;
+        return;
+      }
+
+      // check if drift time is set and if it is present
+      if (drift_start >= 0) // user wants integration across drift time
+      {
+        if (spectrum->getDriftTimeArray() == nullptr)
+        {
+            std::cerr << "Warning: Cannot integrate with drift time if no drift time is available, will integrate without drift time\n";
```
hroest, line 106: make an exception instead?

hroest: is it an empty spectrum? or is it a spectrum that lacks ion mobility?

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -59,142 +59,282 @@ namespace OpenMS::DIAHelpers
...
+
+      double mz(-1), intensity(0), im(-1);
+      if (windowsCenter.empty())
+      {
+        std::cerr << "Warning no windows supplied!" << std::endl;
+        return;
+      }
+      if (spectra.empty())
+      {
+        std::cerr << "WARNING: no spectra provided" << std::endl;
```
hroest, line 217: use LOG_WARN

src/tests/topp/CMakeLists.txt
```diff
@@ -1748,15 +1748,16 @@ ${TOPP_BIN_PATH}/OpenSwathDecoyGenerator -test -in ${DATA_DIR_TOPP}/OpenSwathDec
   set_tests_properties("TOPP_OpenSwathWorkflow_16_out2" PROPERTIES DEPENDS "TOPP_OpenSwathWorkflow_16")
 
   # Test with ion mobility
-  add_test("TOPP_OpenSwathWorkflow_17" ${TOPP_BIN_PATH}/OpenSwathWorkflow -in ${DATA_DIR_TOPP}/OpenSwathWorkflow_17_input.mzML -tr ${DATA_DIR_TOPP}/OpenSwathWorkflow_17_input.tsv -out_chrom OpenSwathWorkflow_17.chrom.mzML.tmp -out_features OpenSwathWorkflow_17.featureXML.tmp
+  add_test("TOPP_OpenSwathWorkflow_17" ${TOPP_BIN_PATH}/OpenSwathWorkflow -in ${DATA_DIR_TOPP}/OpenSwathWorkflow_17_input.mzML -tr ${DATA_DIR_TOPP}/OpenSwathWorkflow_17_input.tsv -out_chrom OpenSwathWorkflow_17.chrom.mzML.tmp -out_features OpenSwathWorkflow_17.featureXML.tmp -enable_ms1 false
```
hroest, line 1751: `use_ms1_ion_mobility` instead? or add ion mobility to the 17.mzML input file?

hroest: actually ion mobility should be there on the 17 mzML file, right?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -62,35 +61,50 @@ namespace OpenMS
...
+      @brief Integrate intensity in an array of spectra from start to end
+
+      This function will integrate the intensity in a spectrum between mz_start
+      and mz_end, returning the total intensity and an intensity-weighted m/z and im
+      value.
+
+      @note ion mobility is only computed if drift_start != -1
+      @note If there is no signal, mz will be set to -1 and intensity to 0
+      @return Returns true if a signal was found (and false if no signal was found)
+
```
cbielow, line 76:
```suggestion
      @brief Integrate intensity in an array of spectra from @p mz_start to @p mz_end

      This function will integrate the intensity in a spectrum between @p mz_start
      and @p mz_end, returning the intensity-weighted m/z and im
      values as well as the total intensity.

      @note ion mobility is only computed if drift_start != -1
      @note If there is no signal, mz will be set to -1 and intensity to 0
      @return Returns true if a signal was found (and false if no signal was found)

```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -532,92 +560,106 @@ namespace OpenMS
...
   }
 
-  OpenSwath::SpectrumPtr OpenSwathScoring::fetchSpectrumSwath(OpenSwath::SpectrumAccessPtr swath_map,
-                                                              double RT, int nr_spectra_to_add, const double drift_lower, const double drift_upper)
+  std::vector<OpenSwath::SpectrumPtr> OpenSwathScoring::fetchSpectrumSwath(OpenSwath::SpectrumAccessPtr swathmap, double RT, int nr_spectra_to_add, double drift_lower, double drift_upper)
   {
-    return getAddedSpectra_(std::move(swath_map), RT, nr_spectra_to_add, drift_lower, drift_upper);
+
+    std::vector<OpenSwath::SpectrumPtr> all_spectra = fetchMultipleSpectra_(swathmap, RT, nr_spectra_to_add);
+    if (spectra_addition_method_ == "simple")
```
cbielow, line 567: only a comment, since not part of this PR: should ideally be an `enum class`, not a string (easy to get the value wrong).

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -532,92 +560,106 @@ namespace OpenMS
     OpenSwath::Scoring::normalize_sum(&normalized_library_intensity[0], boost::numeric_cast<int>(normalized_library_intensity.size()));
   }
 
-  OpenSwath::SpectrumPtr OpenSwathScoring::fetchSpectrumSwath(OpenSwath::SpectrumAccessPtr swath_map,
-                                                              double RT, int nr_spectra_to_add, const double drift_lower, const double drift_upper)
+  std::vector<OpenSwath::SpectrumPtr> OpenSwathScoring::fetchSpectrumSwath(OpenSwath::SpectrumAccessPtr swathmap, double RT, int nr_spectra_to_add, double drift_lower, double drift_upper)
```
cbielow, line 563: looks like this function can be `const`?

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -532,92 +560,106 @@ namespace OpenMS
...
+      OpenSwath::BinaryDataArrayPtr im_arr = input->getDriftTimeArray();
 
-    OpenSwath::BinaryDataArrayPtr mz_arr_out(new OpenSwath::BinaryDataArray);
-    OpenSwath::BinaryDataArrayPtr intens_arr_out(new OpenSwath::BinaryDataArray);
-    OpenSwath::BinaryDataArrayPtr im_arr_out(new OpenSwath::BinaryDataArray);
-    im_arr_out->description = im_arr->description;
+      std::vector<double>::const_iterator mz_it = mz_arr->data.begin();
+      std::vector<double>::const_iterator int_it = int_arr->data.begin();
+      std::vector<double>::const_iterator im_it = im_arr->data.begin();
+      std::vector<double>::const_iterator mz_end = mz_arr->data.end();
```
cbielow, line 627:
```suggestion
      auto mz_it = mz_arr->data.cbegin();
      auto int_it = int_arr->data.cbegin();
      auto im_it = im_arr->data.cbegin();
      auto mz_end = mz_arr->data.cend();
```

cbielow (changes requested):
some initial comments... not done yet, but a start :)

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
cbielow: the comparison will also be a tad faster

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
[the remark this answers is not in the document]
cbielow: the former, making use of `DefaultParamHandler::updateMembers()`.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -532,92 +560,106 @@ namespace OpenMS
...
       OpenSwath::SpectrumPtr spectrum_ = SpectrumAddition::addUpSpectra(all_spectra, spacing_for_spectra_resampling_, true);
-      return spectrum_;
+      std::vector<OpenSwath::SpectrumPtr> spectrum_out;
+      spectrum_out.push_back(spectrum_);
+      return spectrum_out;
     }
   }
 
-  OpenSwath::SpectrumPtr filterByDrift(const OpenSwath::SpectrumPtr& input, const double drift_lower, const double drift_upper)
+  OpenSwath::SpectrumPtr OpenSwathScoring::filterByDrift_(const OpenSwath::SpectrumPtr& input, const double drift_lower, const double drift_upper)
```
cbielow, line 606: IMHO this function does not belong here.
Its a data manipulation function, and has little to do with scoring.
Hence, its hard to find and reuse.
we already have `IONMOBILITY/IMDataConverter.cpp`. Maybe this folder is a better place for this kind if function, so its more accessible to everyone.

Also, this function can be `static` (or a free function)

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -532,92 +560,106 @@ namespace OpenMS
...
+      output->setMZArray(mz_arr_out);
+      output->setIntensityArray(intens_arr_out);
+      output->getDataArrays().push_back(im_arr_out);
+      return output;
   }
 
-
-  OpenSwath::SpectrumPtr OpenSwathScoring::getAddedSpectra_(const OpenSwath::SpectrumAccessPtr& swath_map,
-                                                            double RT, int nr_spectra_to_add, const double drift_lower, const double drift_upper)
+  std::vector<OpenSwath::SpectrumPtr> OpenSwathScoring::fetchMultipleSpectra_(const OpenSwath::SpectrumAccessPtr& swath_map,
```
cbielow, line 654: same here. Not a scoring function.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
```diff
@@ -627,69 +669,79 @@ namespace OpenMS
...
-    else
+
+    return all_spectra;
+  }
+
+  OpenSwath::SpectrumPtr OpenSwathScoring::getAddedSpectra_(std::vector<OpenSwath::SpectrumPtr>& all_spectra, const double drift_lower, const double drift_upper)
+  {
+    OpenSwath::SpectrumPtr added_spec(new OpenSwath::Spectrum);
+    added_spec->getDataArrays().push_back( OpenSwath::BinaryDataArrayPtr(new OpenSwath::BinaryDataArray) );
+    added_spec->getDataArrays().back()->description = "Ion Mobility";
```
cbielow, line 692: magic keyword.
How do you know? :)
maybe https://github.com/OpenMS/OpenMS/blob/develop/src/openswathalgo/include/OpenMS/OPENSWATHALGO/DATAACCESS/DataStructures.h#L270?
But a common global var would be much safer, see https://github.com/OpenMS/OpenMS/blob/develop/src/openswathalgo/include/OpenMS/OPENSWATHALGO/DATAACCESS/DataStructures.h#L270t

src/openms/source/ANALYSIS/OPENSWATH/SwathMapMassCorrection.cpp
```diff
@@ -238,18 +238,22 @@ namespace OpenMS
       {
         if (ms1_im_)
         {
-          sp_ms1 = OpenSwathScoring().fetchSpectrumSwath(ms1_maps, bestRT, 1, 0, 0);
+          std::vector<OpenSwath::SpectrumPtr> fetchSpectrumArr = OpenSwathScoring().fetchSpectrumSwath(ms1_maps, bestRT, 1, 0, 0);
```
cbielow, line 241: this ` .. 1, 0, 0` looks like its easy to use wrong. Again, a DriftTime range as a single argument would be nice.

src/tests/class_tests/openms/source/DIAPrescoring_test.cpp
```diff
@@ -271,7 +284,103 @@ START_SECTION ( test score function with shifted first transition )
...
+    1, 3, 9, 15, 9, 3, 1,        // peak at 501
+    3, 9, 3,                     // peak at 502
+
+
+    10, 20, 50, 100, 50, 20, 10, // peak at 600
+    3, 7, 15, 30, 15, 7, 3,      // peak at 601
+    1, 3, 9, 15, 9, 3, 1,        // peak at 602
+    3, 9, 3                      // peak at 603
+  };
+  std::vector<double> intensity (arr1, arr1 + sizeof(arr1) / sizeof(double) );
```
cbielow, line 325: no need for these tricks since C++11
Just do: `std::vector<double> intensity { 1, 2, 3, 4, 5 ...};`

src/tests/class_tests/openms/source/DIAPrescoring_test.cpp
```diff
@@ -271,7 +284,103 @@ START_SECTION ( test score function with shifted first transition )
...
+      6, 8, 6                   // peak at 603
+  };
+  std::vector<double> im (arr3, arr3 + sizeof(arr3) / sizeof(double) );
+
+  data1->data = mz;
+  data2->data = intensity;
+  data3->data = im;
+  sptr->setMZArray( data1);
+  sptr->setIntensityArray( data2);
+  data3->description = "Ion Mobility";
```
cbielow, line 358: please dont

src/tests/class_tests/openms/source/DIAScoring_test.cpp
```diff
@@ -439,32 +452,32 @@ START_SECTION (void dia_massdiff_score(const std::vector< TransitionType > &tran
...
-  diascoring.dia_massdiff_score(transitions, sptr, normalized_library_intensity, ppm_score, ppm_score_weighted, ppm_errors);
+  diascoring.dia_massdiff_score(transitions, sptr, normalized_library_intensity, ppm_score, ppm_score_weighted, ppm_errors, -1, -1);
 
   TEST_REAL_SIMILAR(ppm_score, (15 + 10) / 2.0); // 15 ppm and 10 ppm
   TEST_REAL_SIMILAR(ppm_score_weighted, 15 * 0.7 + 10* 0.3); // weighted
 }
 END_SECTION
 
-START_SECTION ( bool DIAScoring::dia_ms1_massdiff_score(double precursor_mz, transitions, SpectrumType spectrum, double& ppm_score) )
+START_SECTION ( bool DIAScoring::dia_ms1_massdiff_score(double precursor_mz, transitions, std::vector<SpectrumType> spectrum, double& ppm_score, double drift_start, double drift_end) )
```
cbielow, line 462: the new arguments are never tested?!
Somewhere else maybe?
Add a comment where?

src/tests/class_tests/openms/source/SpectrumHelpers_test.cpp
```diff
@@ -98,19 +101,19 @@ START_SECTION ( [EXTRA] testscorefunction)
   // >> pearsonr(exp, theo)
   // (0.99463189043051314, 0.00047175434098498532)
   //
-  DIAHelpers::integrateWindow(sptr,499.6,501.4,mzres, intensityres);
+  DIAHelpers::integrateWindow(sptrArr,499.6,501.4,mzres, imres, intensityres, -1, -1);
```
cbielow, line 104: again, never testing with the new filtering?

cbielow (changes requested):
a few things here and there, but mainly missing tests for the new functionality.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
jcharkow: I've tried putting it constant and it throws a compilation error. I think because fetchMultipleSpectra_ is called?

src/tests/class_tests/openms/source/DIAPrescoring_test.cpp
jcharkow: Sorry @cbielow I'm confused what is wrong with this.

cbielow: Same problem as here:
https://github.com/OpenMS/OpenMS/pull/6479#discussion_r1071290079

;-)

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
jcharkow: Both of these links seem to be the same thing?

cbielow: https://github.com/OpenMS/OpenMS/blob/develop/src/openms/include/OpenMS/CONCEPT/Constants.h#L266 :)

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathScoring.cpp
jcharkow: Good point. Thought about this however since this uses OpenSwath specific data types I don't think it is appropriate to place in IONMOBILITY/IMDATACONVERTER. Perhaps this can be moved to DIAHelpers though since it is more of a helpter function than a scoring function

cbielow: DIAHelpers only improves the visibility by a small margin.
If you feel it does not belong in IONMOBILITY then create a new IONMOBILITY subfolder in openswath and put there.

jcharkow: I'm not a huge fan of creating a new IONMOBILITY subfolder yet, because I think in the long run it is better to make the downstream functions "ion mobility aware" Currently I'm thinking of placing it in `openswathalgo/dataaccess/iSpectrumAccess` with the idea that when the spectrum is fetched it can be filtered right away.

jcharkow: However this is in the openswathalgo folder which leads to some limitations

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -45,28 +45,44 @@
...
-   * 
+   *
   */
-  class OPENMS_DLLAPI OpenSwathScoring 
+  class OPENMS_DLLAPI OpenSwathScoring
   {
     typedef OpenSwath::LightCompound CompoundType;
     typedef OpenSwath::LightTransition TransitionType;
 
+    using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>; // a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target region
```
cbielow, line 70: use that in more places?!

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -45,28 +45,44 @@
...
+
+    enum SpectrumAdditionMethod
+    {
+      SIMPLE,
```
cbielow, line 74: find a better name?!
`ADDITION` (or whatever is actually done)

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -35,6 +35,7 @@
 #pragma once
 
 #include <OpenMS/CHEMISTRY/AASequence.h>
+#include <OpenMS/KERNEL/RangeManager.h>
```
cbielow, line 38: avoid includes (compile time cost) -- here you can get away with a forward declaration of `RangeMZ` etc, since they are simple classes.
Also in other headers.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -44,53 +45,59 @@ namespace OpenMS
   class TheoreticalSpectrumGenerator;
   namespace DIAHelpers
   {
+    using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>; // a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target region
```
cbielow, line 48:
```suggestion
    using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>; ///< a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target region
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -44,53 +45,59 @@ namespace OpenMS
...
-      @brief Integrate intensity in a spectrum from start to end
+      @brief Integrate intensity in a spectrum from in @p mz_range (and @p im_range if defined)
+      returning the intensity-weighted m/z and im values as well as the total intensity.
 
-      This function will integrate the intensity in a spectrum between mz_start
-      and mz_end, returning the total intensity and an intensity-weighted m/z
-      value.
-
-      @note If there is no signal, mz will be set to -1 and intensity to 0
+      @note If there is no signal, mz and im will be set to -1 and intensity to 0
```
cbielow, line 59:
```suggestion
      @note If there is no signal, @p mz and @p im will be set to -1 and intensity to 0
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -44,53 +45,59 @@ namespace OpenMS
...
-    OPENMS_DLLAPI bool integrateWindow(const OpenSwath::SpectrumPtr& spectrum, double mz_start,
-                                       double mz_end, double& mz, double& intensity, bool centroided = false);
+    OPENMS_DLLAPI bool integrateWindow(const OpenSwath::SpectrumPtr& spectrum,
+                                       double& mz, double& im, double& intensity, const RangeMZ& mz_range, const RangeMobility& im_range, bool centroided = false);
+
+    /**
+      @brief Integrate intensity in SpectrumSequence in range @p mz_range (and @p im_range if defined)
+      returning the intensity-weighted m/z and im values as well as the total intensity.
+
+      @note If there is no signal, mz and im will be set to -1 and intensity to 0
```
cbielow, line 70:
```suggestion
      @note If there is no signal, @p mz and @p im will be set to -1 and intensity to 0
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -162,8 +169,20 @@ namespace OpenMS
     OPENMS_DLLAPI void extractFirst(const std::vector<std::pair<double, double> >& peaks, std::vector<double>& mass);
     /// extract second from vector of pairs
     OPENMS_DLLAPI void extractSecond(const std::vector<std::pair<double, double> >& peaks, std::vector<double>& mass);
-    
+
+
+    /**
+      @brief Helper function for integrating a spectrum.
+    */
+    void _integrateWindowHelper(const OpenSwath::SpectrumPtr& spectrum,
```
cbielow, line 177:
```suggestion
    void integrateWindow_(const OpenSwath::SpectrumPtr& spectrum,
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAPrescoring.h
```diff
@@ -39,6 +39,8 @@
 #include <OpenMS/OPENSWATHALGO/DATAACCESS/TransitionExperiment.h>
 
 #include <OpenMS/DATASTRUCTURES/DefaultParamHandler.h>
+#include <OpenMS/KERNEL/RangeManager.h>
```
cbielow, line 42: use forward declare of `RangeMobility` instead

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
cbielow: this typedef occurs multiple times.
Can you define it in one common place?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -246,70 +259,70 @@ namespace OpenMS
...
-     * @param[in] swath_maps The map(s) containing the spectra
-     * @param[in] RT The target retention time
-     * @param[in] nr_spectra_to_add How many spectra to add up
-     * @param drift_lower Drift time lower extraction boundary
-     * @param drift_upper Drift time upper extraction boundary
+    /** @brief Prepares a spectrum for DIA analysis (single map)
      *
-     * @return Added up spectrum
+     * This function will fetch a vector of spectrum pointers to be used in DIA analysis.
+     * If nr_spectra_to_add == 1, than a vector of length 1 will be returned
```
cbielow, line 265:
```suggestion
     * If nr_spectra_to_add == 1, then a vector of length 1 will be returned
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -246,70 +259,70 @@ namespace OpenMS
...
-                                              const double drift_upper);
+    SpectrumSequence fetchSpectrumSwath(std::vector<OpenSwath::SwathMap> swath_maps, double RT, int nr_spectra_to_add, const RangeMobility& im_range);
 
-  protected:
 
-    /** @brief Returns an averaged spectrum
+   /** @brief Prepares a spectrum for DIA analysis (multiple map)
+     *
+     * This function will fetch a SpectrumSequence to be used in DIA analysis.
+     * If nr_spectra_to_add == 1, than a vector of length 1 will be returned.
```
cbielow, line 291:
```suggestion
     * If nr_spectra_to_add == 1, then a vector of length 1 will be returned.
```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
```diff
@@ -492,7 +494,7 @@ namespace OpenMS
     const TransformationDescription& trafo,
     const ChromExtractParams & cp,
     const ChromExtractParams & cp_ms1,
-    const Param & feature_finder_param,
+    Param & feature_finder_param,
```
cbielow, line 497: not sure this is needed.
Can you make an internal copy, or does the user need to get an updated Param object?

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
```diff
@@ -41,8 +41,46 @@
 namespace OpenMS
 {
 
-  OpenSwath::SpectrumPtr SpectrumAddition::addUpSpectra(const std::vector<OpenSwath::SpectrumPtr>& all_spectra,
-      double sampling_rate, bool filter_zeros)
+  void sortSpectrumByMZ(OpenSwath::Spectrum& spec)
```
cbielow, line 44: this method would be nice to have a member of the Spectrum?
I can see this being useful in a number of scenarios and people will probably not find this implementation.
btw: this kind of functionality already exists for MSSpectrum, so this code duplication is a bit unfortunate, but I can see why this was done.

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
```diff
@@ -41,8 +41,46 @@
...
+    }
+    std::stable_sort(sorted_indices.begin(), sorted_indices.end());
+
+    // extract list of indices
+    std::vector<Size> select_indices;
+    select_indices.reserve(sorted_indices.size());
+    for (const auto& sidx : sorted_indices)
+    {
+      select_indices.push_back(sidx.second);
+    }
```
cbielow, line 63: this can be replaced by something like: https://stackoverflow.com/a/43922758
which is shorter (no idea about performance though -- could be faster or slower... needs timing)

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
```diff
@@ -139,11 +183,78 @@ namespace OpenMS
...
+      for (Size k = 0; k < s->getDataArrays().size(); k++)
+      {
+        auto& v1 = added_spec->getDataArrays()[k]->data;
+        auto& v2 = s->getDataArrays()[k]->data;
+
+        v1.reserve( v1.size() + v2.size() );
+        v1.insert( v1.end(), v2.begin(), v2.end() );
+      }
+    }
+    sortSpectrumByMZ(*added_spec);
```
cbielow, line 253: I guess that all spectra are already sorted by m/z? If yes, you can make this much faster by simply doing an inplace_merge.

src/openswathalgo/include/OpenMS/OPENSWATHALGO/DATAACCESS/ISpectrumAccess.h
```diff
@@ -82,7 +87,68 @@ namespace OpenSwath
     virtual std::size_t getNrChromatograms() const = 0;
     /// Returns the native id of the chromatogram at the given id
     virtual std::string getChromatogramNativeID(int id) const = 0;
-  };
+
+    /* @breif Fetches a spectrumSequence (multiple spectra pointers) closest to the given RT
```
cbielow, line 91:
```suggestion
    /* @brief Fetches a spectrumSequence (multiple spectra pointers) closest to the given RT
```

src/openswathalgo/include/OpenMS/OPENSWATHALGO/DATAACCESS/ISpectrumAccess.h
```diff
@@ -82,7 +87,68 @@ namespace OpenSwath
...
     virtual std::string getChromatogramNativeID(int id) const = 0;
-  };
+
+    /* @brief Fetches a spectrumSequence (multiple spectra pointers) closest to the given RT
+     * @p RT = target RT
+     * @p nr_spectra_to_fetch = # spectra around target RT to fetch (length of the spectrum sequence)
+    */
+    SpectrumSequence getMultipleSpectra(double RT, int nr_spectra_to_fetch);
+
+    /* @breif Fetces a spectrumSequence (multiple spectra pointers) closest to the given RT. Filters all spectra by specified @p drift_start and @p drift_end
```
cbielow, line 97:
```suggestion
    /* @brief Fetches a spectrumSequence (multiple spectra pointers) closest to the given RT. Filters all spectra by specified @p drift_start and @p drift_end
```

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
jcharkow: This is added for line 527 where use_ms1_ion_mobility is updated based on class parameters. Is there a more elegant way to do this?

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
jcharkow: Seems that timing is equivalent so I replaced with the new implementation.

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
jcharkow: Not sure how to do an inplace is feasible considering that the  intensity as these will not be sorted. code might get quite messy...

cbielow: not sure what you mean.
Why would unsorted intensity be a problem when you trying to sort by m/z?

jcharkow: I am not sure what the best strategy would be to preserve the (mz, intensity) pairings with an inplace merge

jcharkow: can an in place merge be done without screwing up the mz-intensity pairing considering that mz and intensity are stored in independent arrays

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
jcharkow: @cbielow please see comment above and let me know your thoughts

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
cbielow: see [MSSpectrum::sortByPositionPresorted(...)](https://abibuilder.cs.uni-tuebingen.de/archive/openms/Documentation/nightly/html/classOpenMS_1_1MSSpectrum.html#a943f07f7e779e5a27bc6bc5f70b87942),  for an example.
It depends on how much time the sort actually takes and if its worth the trouble to port it to the swath spectrum.

jcharkow: Thank you for the example, that makes sense. From my experience the current method is not used too often in practice and is more experimental so I think that the current method is sufficient for the time being.

jcharkow:
Can someone resolve conflicts? I don't have write access so I cannot do it

cbielow:
These commandline instructions by Github are the wrong way around IMHO.
Just 
"pull upstream develop" (where upstream is the OpenMS/OpenMS repo)
within your branch and resolve conflicts locally. Then commit and push to your branch.

jcharkow:
Seems that there are some fails on windows and library linking not present on mac or ubuntu. Any ideas what might be causing this?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -162,8 +168,20 @@ namespace OpenMS
     OPENMS_DLLAPI void extractFirst(const std::vector<std::pair<double, double> >& peaks, std::vector<double>& mass);
     /// extract second from vector of pairs
     OPENMS_DLLAPI void extractSecond(const std::vector<std::pair<double, double> >& peaks, std::vector<double>& mass);
-    
+
+
+    /**
+      @brief Helper function for integrating a spectrum.
+    */
+    void integrateWindow_(const OpenSwath::SpectrumPtr& spectrum,
```
cbielow: missing `OPENMS_DLLAPI` prefix.
Same for OpenMS::DiaPrescore::score

cbielow: this  causes these build error https://cdash.openms.de/viewBuildError.php?buildid=148434

jcharkow: Great thank you!

src/tests/class_tests/openms/source/OpenSwathScoring_test.cpp
```diff
@@ -246,34 +285,300 @@ START_SECTION((OpenSwath::SpectrumPtr OpenSwathScoring::fetchSpectrumSwath(std::
...
+    // Test resampling, IM filtering should occur and the 4th spectrum should not be selected
+    {
+      sc.initialize(1.0, 1, 0.005, 0.0, su, "resample", true);
+
+      SpectrumSequence sp = sc.fetchSpectrumSwath(swath_ptr, 20.0, 3, im_range);
+
+      TEST_EQUAL(sp.size(), 1);
+      TEST_EQUAL(sp[0]->getMZArray()->data.size(), 3);
+      TEST_EQUAL(sp[0]->getIntensityArray()->data.size(), 3);
+      TEST_EQUAL(sp[0]->getDriftTimeArray(), nullptr); // for resampling we do not use IM array
```
jpfeuffer: There might be some errors in our testing code with "nullptr"

https://cdash.openms.de/viewBuildError.php?buildid=148443

@cbielow


Maybe try 0 for a quick fix? Not sure.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
jpfeuffer: OPENMS_DLLAPI missing?

src/tests/class_tests/openms/source/OpenSwathScoring_test.cpp
```diff
@@ -246,34 +285,300 @@ START_SECTION((OpenSwath::SpectrumPtr OpenSwathScoring::fetchSpectrumSwath(std::
...
+    // test resample - IM filtering should occur, also IM information is not needed so is cleared
+    {
+      sc.initialize(1.0, 1, 0.005, 0.0, su, "resample", true);
+      SpectrumSequence sp = sc.fetchSpectrumSwath(swath_ptr, 20.0, 3, im_range);
+
+      TEST_EQUAL(sp.size(), 1);
+
+      TEST_EQUAL(sp[0]->getMZArray()->data.size(), 3);
+      TEST_EQUAL(sp[0]->getIntensityArray()->data.size(), 3);
+      TEST_EQUAL(sp[0]->getDriftTimeArray(), nullptr); // for resampling we do not use IM array
```
cbielow: I guess this is the cause of https://cdash.openms.de/viewBuildError.php?buildid=148443
rewrite as `TEST_TRUE(  (sp[0]->getDriftTimeArray() == nullptr)` ??

cbielow:
Two are better than none :)))

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -71,7 +71,7 @@ namespace OpenMS::DIAHelpers
     // Helper for integrate window returns the sum of all intensities, sum of all ion mobilities and sum of all mz
     // no expensive division calls
     // assumes mz, im and intensity should already be initiated.
-    void integrateWindow_(const OpenSwath::SpectrumPtr& spectrum,
+    OPENMS_DLLAPI void integrateWindow_(const OpenSwath::SpectrumPtr& spectrum,
```
jpfeuffer: I think you need/should add this to the header file (DIAHelper.h) instead.

jcharkow:
Still seem to be having linker problems. Tried adding OPENMS_DLLAPI however still seems to be issues. Seems that DIAScoringHelper already contains OPENMS_DLLAPI in the class so I do not think it should be needed again?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -36,61 +36,67 @@
 
 #include <OpenMS/CHEMISTRY/AASequence.h>
 #include <OpenMS/OPENSWATHALGO/DATAACCESS/DataStructures.h>
-
 #include <vector>
-
 namespace OpenMS
 {
+  class RangeMZ;
+  class RangeMobility;
```
jpfeuffer:
```suggestion
  struct RangeMZ;
  struct RangeMobility;
```

jpfeuffer: class vs struct matters for Visual Studio!

jpfeuffer (changes requested):

src/tests/class_tests/openms/source/OpenSwathScoring_test.cpp
```diff
@@ -246,34 +285,300 @@ START_SECTION((OpenSwath::SpectrumPtr OpenSwathScoring::fetchSpectrumSwath(std::
...
+    // Test resampling, IM filtering should occur and the 4th spectrum should not be selected
+    {
+      sc.initialize(1.0, 1, 0.005, 0.0, su, "resample", true);
+
+      SpectrumSequence sp = sc.fetchSpectrumSwath(swath_ptr, 20.0, 3, im_range);
+
+      TEST_EQUAL(sp.size(), 1);
+      TEST_EQUAL(sp[0]->getMZArray()->data.size(), 3);
+      TEST_EQUAL(sp[0]->getIntensityArray()->data.size(), 3);
+      TEST_EQUAL(sp[0]->getDriftTimeArray(), nullptr); // for resampling we do not use IM array
```
jpfeuffer:
```suggestion
      TEST_TRUE( (sp[0]->getDriftTimeArray() == nullptr ) ); // for resampling we do not use IM array
```

jcharkow:
@jpfeuffer thank you for fixing this!

jpfeuffer:
No problem. pyopenms bindings seem to be affected, too and might need some fixes.

src/pyOpenMS/pxds/DIAScoring.pxd
```diff
@@ -17,18 +19,17 @@ cdef extern from "<OpenMS/ANALYSIS/OPENSWATH/DIAScoring.h>" namespace "OpenMS":
         # private
         DIAScoring(DIAScoring) except + nogil  # wrap-ignore
 
-        bool dia_ms1_massdiff_score(double precursor_mz, OSSpectrumPtr spectrum,
-                                    double& ppm_score) except + nogil  
+        bool dia_ms1_massdiff_score(double precursor_mz, libcpp_vector[shared_ptr[OSSpectrum] ] spectrum,
+                                    RangeMobility& im_range, double& ppm_score) except nogil +
```
jpfeuffer: I think this should be `except + nogil` everywhere.

jcharkow:
I think this PR might be ready for merging?

jcharkow:
@cbielow @hroest  can you please have a look? I think this is ready for merging

jpfeuffer (approved):
I did not do a fresh full review but since my requested changes were the last ones and they were successfully resolved, it is fine from my side to merge.

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
hroest: @jcharkow it is not really elegant, is there another way?

however, you need to consider that this changes the parameters here permanently. Is that what we want, or should we use `Param feature_finder_param` to only operate on a copy? Do we need to pass the changed parameter object back to the caller?

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
```diff
@@ -11,12 +11,48 @@
...
+    // sort indexes based on comparing values in v
+    // using std::stable_sort instead of std::sort
+    // to avoid unnecessary index re-orderings
+    // when v contains elements of equal values
+    auto mzArr = spec.getMZArray()->data.begin();
+    std::stable_sort(sorted_indices.begin(), sorted_indices.end(),
+       [mzArr](size_t i1, size_t i2) {return mzArr[i1] < mzArr[i2];});
+
+    // apply sorting across all arrays
+    for (auto& da : spec.getDataArrays() )
```
hroest, line 38: arent you applying this to the mz array again, effectively first sorting it and then reordering it in the same order you just ordered it ?

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
```diff
@@ -11,12 +11,48 @@
...
+    {
+      if (da->data.empty()) continue;
+      OpenSwath::BinaryDataArrayPtr tmp(new OpenSwath::BinaryDataArray);
```
hroest: is there a way to do this in-place?

src/openms/source/ANALYSIS/OPENSWATH/SpectrumAddition.cpp
```diff
@@ -11,12 +11,48 @@
...
+      tmp->description = da->description;
+      tmp->data.reserve(sorted_indices.size());
+      for (Size i = 0; i < sorted_indices.size(); ++i)
+      {
+        tmp->data.push_back( da->data[ sorted_indices[i] ] );
+      }
+      da = tmp;
```
hroest: would `std::swap` be faster?

hroest:
Yes, looks good from my end as well

jpfeuffer:
Can quickly summarize the changes from your lat commits?
Mainly documentation?

jcharkow:
> Can quickly summarize the changes from your lat commits? Mainly documentation?

More internal changes rather than documentation. All of them are around the `  void SpectrumAddition::sortSpectrumByMZ(OpenSwath::Spectrum& spec` in SpectrumAddition. Allowing this computation to be performed in place rather than creating a copy. This should increase efficiency. Tests are also updated to include testing this feature with ion mobility

jcharkow:
@cbielow  @hroest  can you please take another look at this and see if it is ok for merging?

jpfeuffer (approved):

hroest (approved):

jcharkow:
@cbielow can you please let me know if this branch is good for merging?

cbielow:
Down with  Corona ATM.
Will take a few days.  But feel free to merge if it's urgent.

Am 11.12.2023 18:41 schrieb Joshua Charkow ***@***.***>:

@cbielow<https://github.com/cbielow> can you please let me know if this branch is good for merging?

jcharkow:
No problem it is not super urgent so it can wait a bit. Hope you feel better soon.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/IonMobilityScoring.h
```diff
@@ -41,8 +42,22 @@ namespace OpenMS
...
+      MobilityPeak (double im_, double int_) :
+        im(im_),
+        intensity(int_)
+      {}
+    };
+    typedef std::vector< MobilityPeak > IonMobilogram;
+
   public:
 
+    using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>; // a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target region
```
cbielow: this is the 4th definition of the same thing.
Can you move that to a new header and use it where required?

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -23,24 +23,40 @@
...
   */
-  class OPENMS_DLLAPI OpenSwathScoring 
+  class OPENMS_DLLAPI OpenSwathScoring
   {
     typedef OpenSwath::LightCompound CompoundType;
     typedef OpenSwath::LightTransition TransitionType;
 
+    using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>; // a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target region
+
+    enum SpectrumAdditionMethod
```
cbielow:
```suggestion
    enum class SpectrumAdditionMethod
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -162,52 +179,48 @@ namespace OpenMS
...
     /** @brief Score a single chromatographic feature using the precursor map.
      *
-     * The scores are returned in the OpenSwath_Scores object. 
+     * The scores are returned in the OpenSwath_Scores object.
      *
      * @param ms1_map The MS1 (precursor ion map) from which the precursor spectra can be retrieved
      * @param diascoring DIA Scoring object to use for scoring
      * @param precursor_mz The m/z ratio of the precursor
      * @param rt The compound retention time
+     * @param compound the compond sequence
```
cbielow:
```suggestion
     * @param compound The compound sequence
```

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/OpenSwathScoring.h
```diff
@@ -220,70 +233,70 @@ namespace OpenMS
...
-                                            double RT,
-                                            int nr_spectra_to_add,
-                                            const double drift_lower,
-                                            const double drift_upper);
+    SpectrumSequence fetchSpectrumSwath(OpenSwath::SpectrumAccessPtr swath_map, double RT, int nr_spectra_to_add, const RangeMobility& im_range);
+
+
+  protected:
+
+    /** @breif converts a ion mobility enhanced spectrum to a non ion mobility spectrum by filtering by drift time
```
cbielow:
```suggestion
    /** @brief converts an ion mobility enhanced spectrum to a non-ion mobility spectrum by filtering by drift time
```

src/openms/source/ANALYSIS/OPENSWATH/DIAHelper.cpp
```diff
@@ -33,142 +35,295 @@ namespace OpenMS::DIAHelpers
...
+      OPENMS_PRECONDITION(std::adjacent_find(spectrum->getMZArray()->data.begin(),
+        spectrum->getMZArray()->data.end(), std::greater<double>()) == spectrum->getMZArray()->data.end(),
+        "Precondition violated: m/z vector needs to be sorted!" );
+      OPENMS_PRECONDITION(spectrum->getMZArray()->data.size() == spectrum->getIntensityArray()->data.size(), "precondition: MZ and Intensity array need to have the same length.");
+
+      // ion mobility specific preconditions
+      //OPENMS_PRECONDITION((im_range.isEmpty()) && (spectrum->getDriftTimeArray() != nullptr), "precondition: Cannot integrate with drift time if no drift time is available."); This is not a failure can handle this
+      OPENMS_PRECONDITION((spectrum->getDriftTimeArray() == nullptr) || (spectrum->getDriftTimeArray()->data.empty()) || (spectrum->getMZArray()->data.size() == spectrum->getDriftTimeArray()->data.size()), "precondition: MZ and Drift Time array need to have the same length.");
+      OPENMS_PRECONDITION(!centroided,  throw Exception::NotImplemented(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION));
```
cbielow, line 68: why do you add a bool where `false` works and `true` raises an exception?
You can omit the bool for now and save some headaches.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -10,61 +10,67 @@
...
-      @brief Integrate intensities in a spectrum from start to end
+      @brief Integrate intensities in a spectrum in range @p im_range (if defined) for multiple windows.
     */
+
     OPENMS_DLLAPI void integrateWindows(const OpenSwath::SpectrumPtr& spectrum, //!< [in] Spectrum
                                         const std::vector<double>& windows_center, //!< [in] center location
                                         double width,
                                         std::vector<double>& integrated_windows_intensity,
                                         std::vector<double>& integrated_windows_mz,
+					std::vector<double>& integrated_windows_im,
```
cbielow, line 65: please document parameters

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
```diff
@@ -212,29 +198,41 @@ namespace OpenMS
...
                                                    const double drift_extra)
   {
-    OPENMS_PRECONDITION(spectrum != nullptr, "Spectrum cannot be null");
+    OPENMS_PRECONDITION(!spectra.empty(), "Spectra cannot be empty")
+    OPENMS_PRECONDITION(!ms1spectrum.empty(), "MS1 spectrum cannot be empty")
     OPENMS_PRECONDITION(!transitions.empty(), "Need at least one transition");
 
-    if (ms1spectrum->getDriftTimeArray() == nullptr)
+    //TODO not sure what error format is best
+    for (auto s:spectra)
```
cbielow:
```suggestion
    for (const auto& s : spectra)
```

src/openms/source/ANALYSIS/OPENSWATH/IonMobilityScoring.cpp
```diff
@@ -212,29 +198,41 @@ namespace OpenMS
...
-      OPENMS_LOG_DEBUG << " ERROR: Drift time is missing in ion mobility spectrum!" << std::endl;
-      return;
+      if (s->getDriftTimeArray() == nullptr)
+      {
+        OPENMS_LOG_DEBUG << " ERROR: Drift time is missing in ion mobility spectrum!" << std::endl;
+        return;
+      }
+    }
+
+    for (auto s:ms1spectrum)
```
cbielow:
```suggestion
    for (const auto& s : ms1spectrum)
```

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -476,20 +472,33 @@ namespace OpenMS
...
       auto & prec = transition_group_detection.getChromatograms()[0].getPrecursor();
-      drift_lower = prec.getDriftTime() - prec.getDriftTimeWindowLowerOffset();
-      drift_upper = prec.getDriftTime() + prec.getDriftTimeWindowUpperOffset();
       drift_target = prec.getDriftTime();
+
+      if (drift_target > 0)
+      {
+        im_range.setMin(drift_target); // sets the minimum and maximum
+        // currently only works for symmetrical window
+        im_range.minSpanIfSingular(prec.getDriftTimeWindowLowerOffset() * 2);
```
cbielow: this is wrong for asymmetric windows?!

easier and correct:
```
 im_range = RangeMobility(drift_target - prec.getDriftTimeWindowLowerOffset(), drift_target + prec.getDriftTimeWindowUpperOffset());
```

src/openms/source/ANALYSIS/OPENSWATH/MRMFeatureFinderScoring.cpp
```diff
@@ -476,20 +472,33 @@ namespace OpenMS
...
+      }
```
cbielow: same code?!
maybe just operate on a pointer or call a lambda to avoid duplication

src/openms/source/ANALYSIS/OPENSWATH/OpenSwathWorkflow.cpp
cbielow: this still needs adressing

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -149,22 +150,22 @@ namespace OpenMS
...
     ppm_score = -1;
-    double mz, intensity;
+    double mz, intensity, im;
     {
       // Calculate the difference of the theoretical mass and the actually measured mass
-      double left(precursor_mz), right(precursor_mz);
-      DIAHelpers::adjustExtractionWindow(right, left, dia_extract_window_, dia_extraction_ppm_);
-      bool signalFound = DIAHelpers::integrateWindow(std::move(spectrum), left, right, mz, intensity, dia_centroided_);
+      RangeMZ mz_range(precursor_mz);
+      mz_range.minSpanIfSingular(dia_extract_window_, dia_extraction_ppm_);
```
cbielow: no!
RangeManager is agnostic of what unit the range has. Introducing ppm, with a default and some side effect behaviour is not a good idea.
Better to leave it separated:
```
 mz_range.minSpanIfSingular(  mzToPPMOptional(dia_extraction_window_, dia_extraction_ppm_) );
```

src/openms/source/ANALYSIS/OPENSWATH/DIAScoring.cpp
```diff
@@ -149,22 +150,22 @@ namespace OpenMS
...
+      bool signalFound = DIAHelpers::integrateWindow(spectrum, mz, im, intensity, mz_range, im_range, dia_centroided_);
 
       // Catch if no signal was found and replace it with the most extreme
       // value. Otherwise, calculate the difference in ppm.
       if (!signalFound)
       {
-        ppm_score = (right - left) / precursor_mz * 1000000;
+        ppm_score = (mz_range.getSpan()) / precursor_mz * 1000000;
```
cbielow: this kind of code can be made more readable by using our Math::ppm functions

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -39,7 +40,17 @@ namespace OpenMS
   public:
     /// C'tor: initialize with empty range
     RangeBase() = default;
-    
+    /*
+    RangeBase():
+     min_(0), max_(-1);
+    */
```
cbielow:
```suggestion
```

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -16,6 +16,7 @@
 #include <algorithm> // for min/max
 #include <cassert>
 #include <iosfwd>  // for std::ostream
+#include <iostream>
```
cbielow:
```suggestion
```

not needed

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -39,7 +40,17 @@ namespace OpenMS
   public:
     /// C'tor: initialize with empty range
     RangeBase() = default;
-    
+    /*
+    RangeBase():
+     min_(0), max_(-1);
+    */
+
+    /// Cutom C'tor which sets the range to a singular point
```
cbielow, line 43: ok, if its needed

src/openms/include/OpenMS/KERNEL/RangeManager.h
```diff
@@ -387,6 +396,28 @@ namespace OpenMS
...
+      else
+      {
+        RangeBase::minSpanIfSingular(initial_min_span);
+      }
+      if (min_ < 0)
+      {
+        min_ = 0;
+      }
+    }
+
```
cbielow:
```suggestion
```

This function must go!
It has side effects and should not be part of RangeManager in general

cbielow (changes requested):
there are some loose ends here, which need fixing

src/openms/include/OpenMS/KERNEL/RangeManager.h
jcharkow: Is there another place where this function could be or should I just implement this in OpenSwath classes? It seems like something more generally applicable than OpenSwath though

> This function must go! It has side effects and should not be part of RangeManager in general

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/IonMobilityScoring.h
jcharkow: Is it appropriate to create a new OpenMS namespace in the ISpectrumAccess.h and add the definition there or will that impact build times a lot?

jcharkow: Right now I can narrow it down to 4 declarations (I removed some) 
DIAHelper.h
OpenSwathScoring.h
SpectrumAddition.h
ISpectrumAccess.h

src/openms/include/OpenMS/KERNEL/RangeManager.h
jcharkow: Sorry did not see comment above

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/IonMobilityScoring.h
cbielow: > Is it appropriate to create a new OpenMS namespace in the ISpectrumAccess.h and add the definition there or will that impact build times a lot?

If the components are known anyway, build times won't be affected.
Even if, a definition in one place is much cleaner.

src/openms/include/OpenMS/KERNEL/RangeManager.h
cbielow: > Is there another place where this function could be or should I just implement this in OpenSwath classes? It seems like something more generally applicable than OpenSwath though
> 
> > This function must go! It has side effects and should not be part of RangeManager in general

You could create a new `RangeUtils.h` and put common conversion functions in there, e.g. the one from above:
```
  RangeMobility fromIsolationWindowDriftTime(const Precursor& pc);

  RangeMZ fromIsolationWindowMZ(const Precursor& pc);
```

jcharkow: it seems that there already is a RangeUtils.h and it does not seem super related. For now I've created a function in DIAHelpers which seems more appropriate for this specific usage

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -48,8 +49,16 @@ namespace OpenMS
 
     /**
       @brief Integrate intensities in a spectrum in range @p im_range (if defined) for multiple windows.
-    */
+      @p windows_center is a vector of the center location of the windows.
+      @p width is the width of the windows across mz
+      @p RangeMobility is the range of the IM dimension (if defined)
```
cbielow: `@p` is for inline emphasis.
You need `@param`

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -48,8 +49,16 @@ namespace OpenMS
...
+
+      Returns:
+      integrated_windows_intensity is a vector of the integrated intensity for each window
+      integrated_windows_mz is a vector of the integrated intensity-weighted m/z for each window
+      integrated_windows_im is a vector of the integrated intensity-weighted im for each window
```
cbielow: use `@param` or (even better) `@param[out] ...`  (note: no space between `@param` and `[out]`

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
cbielow:
```suggestion
      @param im_range The range of the IM dimension (if defined)
```

wrong variable name...

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
cbielow: also document `remove_zero`

src/openswathalgo/include/OpenMS/OPENSWATHALGO/DATAACCESS/ISpectrumAccess.h
```diff
@@ -15,10 +15,14 @@
 #include <string>
 #include <vector>
 
+namespace OpenMS
+{
+  using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>;  // a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target regions
```
cbielow:
```suggestion
  using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>;  ///< a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target regions
```

src/openswathalgo/include/OpenMS/OPENSWATHALGO/DATAACCESS/ISpectrumAccess.h
```diff
@@ -15,10 +15,14 @@
...
+}
 namespace OpenSwath
 {
 
-  using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>;
+  using SpectrumSequence = std::vector<OpenSwath::SpectrumPtr>;  // a vector of spectrum pointers that DIA scores can operate on, allows for clever integration of only the target regions
```
cbielow: nah... just reference the other, instead of defining it again

cbielow (commented):
very good changes.
Almost there! :)

jcharkow:
Thank you for the suggestions please let me know if there is anything else.

src/openms/include/OpenMS/ANALYSIS/OPENSWATH/DIAHelper.h
```diff
@@ -10,61 +10,74 @@
...
+    */
+    OPENMS_DLLAPI bool integrateWindow(const SpectrumSequence& spectrum,
+                                       double& mz, double& im, double& intensity, const RangeMZ& mz_range, const RangeMobility& im_range, bool centroided = false);
+
+    /**
+      @brief Integrate intensities in a spectrum in range @param im_range (if defined) for multiple windows.
+      @param windows_center is a vector of the center location of the windows.
+      @param width is the width of the windows across mz
+      @param im_range is the range of the IM dimension (if defined)
+      @param remove_zero is a flag indicating whether to remove zero intensity windows
```
cbielow:
```suggestion
      @brief Integrate intensities in a spectrum in range @p im_range (if defined) for multiple windows.
      @param windows_center is a vector of the center location of the windows.
      @param width is the width of the windows across mz
      @param im_range is the range of the IM dimension (if defined)
      @param remove_zero is a flag indicating whether to remove zero intensity windows
```
```suggestion
      @brief Integrate intensities in a spectrum in range @param im_range (if defined) for multiple windows.
      @param windows_center is a vector of the center location of the windows.
      @param width is the width of the windows across mz
      @param im_range is the range of the IM dimension (if defined)
      @param remove_zero is a flag indicating whether to remove zero intensity windows
```

PR #6482, opened 2022-11-18
Feature/modomics json

poshul:
## Description

The updated modomics API generated JSON or CSV files, rather than the older TSV. I've added support for parsing these, and updated our version.

## Checklist

<details>

</details>

<details>
  
</details>

---


src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -67,6 +73,141 @@ namespace OpenMS
...
+    }
+    else if (code.hasSuffix('?') || code.hasSuffix("?*")) // ambiguity code -> fill the map
+    {
+      if (!entry.contains("alternatives"))
+      {
+        String msg = "Ambiguous mod without alternative found in " + code;
+        //throw Exception::ParseError(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, code, msg);
+      }
+      String code1 = string(entry.at("alternatives").at(0)), code2 = string(entry.at("alternatives").at(1)); //we always have exactly two ambiguities
+      ambiguity_map_[code] = make_pair(getRibonucleotide(code1), getRibonucleotide(code2));
```
poshul, line 202: @timosachsenberg any idea why I get a `not declared in this scope` error for `ambiguity_map_` its a protected member of RibonucleotideDB, and works fine in the `parseRow` function. I assume its something to do with how `parseEntry` was forward declared in this file and `parseRow` is declared in the header.

timosachsenberg: parse entry is not a member

share/OpenMS/CHEMISTRY/Custom_RNA_modifications.tsv
```diff
@@ -34,6 +34,11 @@ ambiguous dimethylguanosine ("m2,2G" or "m_Gm")	mmG?		G	Ó	&Oacute;	C12O5N5H17	3
 ambiguous trimethylguanosine ("m2,2,7G" or "m2,7Gm")	mmmG?		G	Ô	&Ocirc;	C13O5N5H21	327.1543	327.3404	m2,2,7G m2,7Gm
 N4-acetylcytidine or 5-formyl-2'-O-methylcytidine	ac4C/f5Cm?		C	M	M	C11O6N3H15	285.0961	285.257	ac4C f5Cm
 3-(3-amino-3-carboxypropyl)uridine or 5-carboxymethylaminomethyl-2'-O-methyluridine	acp3U/cmnm5Um?		U	X	X	C13O8N3H19	345.1172	345.3096	acp3U cmnm5Um
+deoxyadenosine	dA	dA	A			C10O3N5H13	251.1018	251.2424
+deoxycytidine	dC	dC	C			C9O4N3H13	227.0906	227.2176
+deoxyguanosine	dG	dG	G			C10O4N5H13	267.0968	267.2418
+deoxyuridine	dU	dU	U			C9O5N2H12	228.0746	228.2024
+thymidine	dT	dT	U			C10O5N2H14	242.0903	242.2290
```
timosachsenberg, line 41: U?

poshul: Yup. We treat the DNA bases as modified (de-oxygenated) RNA ribobases. Since thymidine doesn't have a true analog in the four RNA ribobases we treat it as methyl-deoxy-uridine, hence U here.

timosachsenberg: makes sense... just wanted to confirm that it is ot a copy paste thign

src/tests/class_tests/openms/source/RibonucleotideDB_test.cpp
```diff
@@ -65,7 +65,7 @@ END_SECTION
 START_SECTION(ConstIterator begin())
 {
   RibonucleotideDB::ConstIterator it = ptr->begin();
-  TEST_STRING_EQUAL((*it)->getCode(), "m1Am");
+  TEST_STRING_EQUAL((*it)->getCode(), "io6A");
 }
 END_SECTION
```
timosachsenberg, line 71: can you add a test that checks if loading from tsv or json works?

poshul: will do.

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -131,44 +165,53 @@ namespace OpenMS
       OPENMS_LOG_WARN << "Average mass of " << code << " differs substantially from its formula mass.\n";
     }
 
-    if (!(entry.at("mass_monoiso").is_null()))
+    if (!(entry.find("mass_monoiso") == entry.cend()) && !(entry.at("mass_monoiso").is_null()))
```
timosachsenberg, line 202: borderline complexity of expression :)
the newer syntax for if expressions with initalization statement would make it easier/faster to parse for humans:
```
if (auto e = entry.find("mass_monoiso"); e != entry.cend() && !e->is_null)
{
}
```

src/tests/class_tests/openms/source/RibonucleotideDB_test.cpp
poshul: > can you add a test that checks if loading from tsv or json works?

Looking at these tests, the loading of the TSV and JSON occurs when the constructor for RibonucleotideDB is called the first time. What's the correct way to test them outside of constructor?

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
     }
   }
 
+  // All valid JSON ribonucleotides must at minimum have elements defining name, short_name, reference_moiety, and formula
+  bool entryIsWellFormed_(const nlohmann::json::value_type& entry)
```
cbielow, line 91: not the behaviour I would expect.
If you return a bool, which is never `false`, why return something in the first place?! :)

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
...
+    return true;
+  }
+  
+  EmpiricalFormula getBaseLossFormula_(const nlohmann::json::value_type& entry)
+  {
+    String code = entry.at("short_name");
+    // If we have an explicitly defined baseloss_formula
+    if (auto e = entry.find("baseloss_formula"); e != entry.cend() && !e->is_null())
+    {
+      return EmpiricalFormula(entry.at("baseloss_formula"));
```
cbielow, line 126:
```suggestion
      return EmpiricalFormula(*e);
```

don't look it up again. You have it right here.

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
...
+        ribo->setTermSpecificity(Ribonucleotide::THREE_PRIME);
+      }
+      else
+      {
+        ribo->setTermSpecificity(Ribonucleotide::ANYWHERE); //other nonspecific mods
+      }
+    }
+    else
+    {
+      OPENMS_LOG_ERROR << "Error: we don't support bases with multiple reference moieties or multicharacter moieties." << endl;
```
cbielow, line 188: prints an error, but keeps on going??

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
...
+    }
+    
+    if (!(entry.find("abbrev") == entry.cend()))
```
cbielow, line 191:
```suggestion
    if (entry.find("abbrev") != entry.cend())
```

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
...
+    {
+      ribo->setHTMLCode(entry.at("abbrev")); //This is the single letter unicode representation that only SOME mods have
+    }
+    ribo->setFormula(EmpiricalFormula(entry.at("formula")));
+    if ( !(entry.find("mass_avg") == entry.cend()) && !(entry.at("mass_avg").is_null()))
+    {
+      ribo->setAvgMass(entry.at("mass_avg"));
+    }
+    if (ribo->getAvgMass() - ribo->getFormula().getAverageWeight() >= 0.01)
```
cbielow, line 200: use std::abs() ?

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
...
+    {
+      OPENMS_LOG_WARN << "Average mass of " << code << " differs substantially from its formula mass.\n";
+    }
+
+    if (auto e = entry.find("mass_monoiso"); e != entry.cend() && !e->is_null())
+    {
+      ribo->setMonoMass(entry.at("mass_monoiso"));
```
cbielow, line 207: same here

src/tests/class_tests/openms/source/RibonucleotideDB_test.cpp
timosachsenberg: It is probably fine to test it in the test for the constructor.
But you should probably test at least one entry per file type (and add a comment what you are testing to communicate intend so it is not accidentally changed, e.g., back to only testing for entries in the TSV).

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
...
+      {
+        return EmpiricalFormula("C5H10O5");
+      }
+    }
+  }
+
+  ParsedEntry_ parseEntry_(const nlohmann::json::value_type& entry)
+  {
+    ParsedEntry_ parsed;
+    Ribonucleotide* ribo = new Ribonucleotide();
```
cbielow, line 157: this should really be a `unique_ptr<Ribonucleotide>`.
The exceptions below will cause memory leaks otherwise.

`ribonucleotides_` should also contain unique_ptr's. This avoids having a custom Dtor etc...

cbielow (changes requested):
cannot say anything about the modomics part, but the implementation would benefit from some changes :)

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -66,6 +87,209 @@ namespace OpenMS
...
+      throw Exception::FileNotReadable(__FILE__, __LINE__, OPENMS_PRETTY_FUNCTION, full_path);
+    }
+
+    QTextStream source(&file);
+    source.setCodec("UTF-8");
+    Size line_count = 0;
+    json mod_obj;
+    try
+    {
+      mod_obj = json::parse(String(source.readAll()));
```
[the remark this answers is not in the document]
poshul: `readAll()` happily returns whatever unicode is in there, the call to `String(...)` calls QT's toStdString(), which decay's it to ASCII

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
poshul: I've handled the exceptions here (freeing ribo before throwing.). I'll open an issue for the larger suggestion.

src/openms/source/CHEMISTRY/RibonucleotideDB.cpp
```diff
@@ -37,15 +37,36 @@
...
 #include <QTextStream>
+#include <nlohmann/json.hpp>
+
 
 using namespace std;
 
 namespace OpenMS
 {
+
+  struct ParsedEntry_
```
timosachsenberg, line 48: please add some documentation on the struct and data members
e.g. it should already be obvious why there are exactly two alternatives etc.

timosachsenberg (approved):
looks good... some more documentation would be great

timosachsenberg (approved):

timosachsenberg:
https://cdash.seqan.de/viewTest.php?onlyfailed&buildid=77235

PR #6485, opened 2022-11-21
fixes some (potential) copies

timosachsenberg:
## Description

no really performance critical things

## Checklist

<details>

</details>

<details>
  
</details>

---

### DIFF
--- a/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
@@ -227,6 +227,8 @@ namespace OpenMS
     /// allows to set the index, using `index.pos() = 3;` or simply read its value
     T& pos();
 
+    /// allows to set the index, using `index.pos() = 3;` or simply read its value
+    const T& pos() const;
   private:
     T i_ = std::numeric_limits<T>::max(); ///< internal number representation; invalid state by default
   };
--- a/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
+++ b/src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
@@ -227,8 +227,8 @@ namespace OpenMS
     /// allows to set the index, using `index.pos() = 3;` or simply read its value
     T& pos();
 
-    /// allows to set the index, using `index.pos() = 3;` or simply read its value
-    const T& pos() const;
+    /// allows to read the index, using `index.pos()` 
+    T pos() const;
   private:
     T i_ = std::numeric_limits<T>::max(); ///< internal number representation; invalid state by default
   };
--- a/src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
+++ b/src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
@@ -636,7 +636,7 @@ namespace OpenMS
     return i_;
   }
 
-  const Index::T& Index::pos() const
+  Index::T Index::pos() const
   {
     return i_;
   }
--- a/src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
+++ b/src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
@@ -80,7 +80,7 @@ namespace OpenMS
   void ACTrie::addNeedle(const std::string& needle)
   {
     Index cn {0}; // start at root
-    for (const auto& c : needle)
+    for (auto c : needle) // OMS_CODING_TEST_EXCLUDE
     {
       AA aa(c);
       // make sure invalid chars raise an exception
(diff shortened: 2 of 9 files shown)


src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
```diff
@@ -80,7 +80,7 @@ namespace OpenMS
   void ACTrie::addNeedle(const std::string& needle)
   {
     Index cn {0}; // start at root
-    for (auto c : needle)
+    for (const auto& c : needle)
```
cbielow, line 83: this is almost certainly a pessimisation, since its only a single byte.
We had a comment keyword to annotate that a simple `auto` is ok, right?

src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
```diff
@@ -526,7 +526,7 @@ namespace OpenMS
 
   Index ACTrie::findChildNaive_(Index parent, AA child_label)
   {
-    for (auto child : umap_index2children_naive_[parent])
+    for (const auto& child : umap_index2children_naive_[parent])
```
cbielow, line 529:
```suggestion
    for (auto child : umap_index2children_naive_[parent])  // OMS_CODING_TEST_EXCLUDE Note: only a 4byte type. Copy it!
```

src/openms/include/OpenMS/ANALYSIS/ID/AhoCorasickAmbiguous.h
```diff
@@ -227,6 +227,8 @@ namespace OpenMS
     /// allows to set the index, using `index.pos() = 3;` or simply read its value
     T& pos();
 
+    /// allows to set the index, using `index.pos() = 3;` or simply read its value
+    const T& pos() const;
```
cbielow, line 231:
```suggestion
    /// allows to read the index, using `index.pos()` 
    T pos() const;
```

src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
```diff
@@ -636,4 +636,9 @@ namespace OpenMS
     return i_;
   }
 
+  const Index::T& Index::pos() const
```
cbielow, line 639:
```suggestion
  Index::T Index::pos() const
```

src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
timosachsenberg: I checked this for a few cases in godbolt, and it didn't use a pointer/similar but was always able to generate the same asm code.

timosachsenberg: Alternatively, just use char c here instead of auto then it is also clear that a copy doesn't hurt

cbielow: > I checked this for a few cases in godbolt, and it didn't use a pointer/similar but was always able to generate the same asm code.

The asm generated highly depends on the context: the compiler must re-read an `a&` value from memory everytime it is mentioned, since it cannot know if the value changed in the meantime (keyword: aliasing). So copy is the safe bet and allows for more optimizations (potentially).

cbielow: > Alternatively, just use char c here instead of auto then it is also clear that a copy doesn't hurt

Not sure here, since this might incur a conversion, if some implementations use `unsigned char` vs `char`. Auto is the safe bet.

timosachsenberg: A std::string is a std::basic_string<char>

cbielow: yep, correct. The time it takes to figure out that this is guaranteed by the standard is probably not worth arguing over 'auto' vs. 'char'?  ;)

src/openms/source/ANALYSIS/RNPXL/RNPxlReport.cpp
```diff
@@ -267,18 +267,7 @@ namespace OpenMS
...
-            {
-              for (auto a : aa) // NOTE: performs copy, this cannot possibly work!
-              {
-                if (!a.isModified())
-                {
-                  a.setModification("[" + String(rna_weight) + "]");
-                  break;
-                }
-              }
-            }
```
cbielow, line 281: ok, so this never did anything, so removing it is safe, I guess.
But *should* it do something, i.e. is it actually a bug right now?

timosachsenberg: will get rid of the RNPxlReport anyway soon so this is not really relevant anymore

cbielow (approved):

src/openms/source/ANALYSIS/RNPXL/RNPxlReport.cpp
cbielow: ok. Makes sense.
PR looks good to me

src/openms/source/ANALYSIS/ID/AhoCorasickAmbiguous.cpp
timosachsenberg: agreed, ;) why I am nitpicking: it would have the nice side effect that we could just have the guideline to use mainly const auto& in the general case or the actual type for the build in ones. But not so important, I think this is done...

PR #6501, opened 2022-11-24
Peak1D and Peak2D: change intensity type to double as well

timosachsenberg:
## Description

K I did some tests how switching the intensity type from float to double (like for m/z) affects OpenMS.
Surprisingly, hick-ups were rather small and only impacted a few tests.

Note: in MSSpectrum, we did not save memory with the float anyway due to alignment of Peak1D structs
as array of structs.

Some relevant questions are: 
- is this helpful? e.g. 
- simplifies switching to AoS?
- use Eigen Map capabilities to directly process peaks and intensities. Is it worth it or just cool?
- should we even go from double to float for both m/z and intensity? This would approx. half memory usage.

## Checklist

<details>

</details>

<details>
  
</details>

---


cbielow:
not convinced this is the way to go.
Sure, the memory footprint is identical, due to padding in Peak1D and Peak2D.
However, doing non-trivial computations on doubles, like `log()` will take more time, since it takes longer to converge.
Trivial operations like addition should have the same speed.
Furthermore, if we ever move to SoA, then using `float` will allow twice the throughput for SIMD instructions.
However, using float for `m/z` may be harmful since there may not be enough precision for UltraHiRes data (in the future).
Bottom line: I would leave it as is: float for intensity, and double for m/z. You can probably make a good case for float in RT/IM.

PR #6523, opened 2022-12-01
FLASHDeconv (only)

timosachsenberg:
Reverts OpenMS/OpenMS#6518


timosachsenberg:
https://cdash.openms.de/viewBuildError.php?buildid=25106

JeeH-K (approved):

JeeH-K:
@timosachsenberg 
The only error we get from the current version, openms-ci/build-win, is not related to FLASHDeconv. It looks like there's a heap memory error or something. Do you think it's fine to merge now?

PR #6541, opened 2022-12-08
Some code to start discussing the ML interface a bit in more detail

timosachsenberg:
## Description

for discussion how pytorch could be integrated via an external library @KyowonJeong @Arslan-Siraj @jpfeuffer  @matteopilz 
(note: I did not include pytorch as e.g., ExternalProject_Add in the cmake file...)
this is first to discuss how we can get a proper separation in a library etc.
It is not complete but a first start...

The key idea is to optionally build a separate library (OpenMS_ML.dll similar to how we build our OpenMS_GUI.dll) if pytorch is enabled in cmake.
This library will then provide an interface to the models we use [src/openms_ml/include/OpenMSMLWrapper.h](https://github.com/OpenMS/OpenMS/pull/6541/files#diff-105b3e8c1afeec8e41659e0cb51329105eb79f28948b3a979c551c51aa5763f6) .

Because we don't including anything from pytorch in this header file, we precent an include include of pytorch in all of OpenMS.
We can of course use pytorch functionality by including the pytorch headers in the OpenMSMLWrapper.cpp file.

This ensures that everything of pytorch gets only linked to the OpenMS_ML library.

## Checklist

<details>

</details>

<details>
  
</details>

---


timosachsenberg:
Dumping here some examples how others include the actual pytorch library in CMake https://github.com/jolibrain/deepdetect/blob/master/CMakeLists.txt#L501-L515

CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ option(ENABLE_DOCS "Indicates whether documentation should be built." ON)
 option(ENABLE_TUTORIALS "Indicates whether tutorials should be build. Note that this also depends on the availability of (pdf)latex." ON)
 option(WITH_GUI "Build GUI parts of OpenMS (TOPPView&Co). This requires QtGui." ON)
 option(NO_WEBENGINE_WIDGETS "Do not use QtWebengineWidgets. Disables Javascript views in TOPPView." OFF)
+option(ENABLE_TORCH "Enable pyTorch" OFF)
```
timosachsenberg, line 89: not sure where the best location is to query ENABLE_TORCH.

src/openms_ml/include/OpenMSMLWrapper.h
```diff
@@ -0,0 +1,53 @@
...
+// --------------------------------------------------------------------------
+
+#pragma once
+
+#include <OpenMS/KERNEL/StandardTypes.h>
+
+#include <vector>
+#include <string>
+
+// Note: No torch headers may be included here! only in c++ file. otherwise torch will pollute whole OpenMS API
```
cbielow, line 42: maybe forward declarations are helpful, to make the API more powerful, yet avoid pollution?

src/openms_ml/include/OpenMSMLWrapper.h
```diff
@@ -0,0 +1,53 @@
...
+namespace OpenMS
+{
+	OPENMS_DLLAPI class DeepLCPredictor
+	{
+		explicit DeepLCWrapper(std::string filename);
```
cbielow, line 47:
```suggestion
		explicit DeepLCPredictor(const std::string_view filename);
```

Just a general thing, which we might want to do? I.e. switch to SVs more often?

timosachsenberg: We can’t wrap sv in python. Is there any advantage compared to a const& ?

src/openms_ml/include/OpenMSMLWrapper.h
timosachsenberg: Yeah definitely if we need to use pImpl or to manage some state e.g. a model*. We might get some info what is needed for @KyowonJeong use case.

CMakeLists.txt
jpfeuffer: Place looks fine but maybe call it WITH_OPENMS_ML? Or ENABLE_OPENMS_ML.

"Enabling torch" sounds kind of weird.

src/openms_ml/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+
+#------------------------------------------------------------------------------
+# add the library
+openms_add_library(TARGET_NAME OpenMS_ML
+                   SOURCE_FILES ${OpenMSML_sources}
+                   HEADER_FILES ${OpenMSML_sources_h}
+                   INTERNAL_INCLUDES ${PROJECT_SOURCE_DIR}/include ${PROJECT_BINARY_DIR}/include
+                   EXTERNAL_INCLUDES
+                   LINK_LIBRARIES ${OpenMS_ML_DEP_LIBRARIES}
+                   DLL_EXPORT_PATH "OpenMS/") # TODO ???
```
timosachsenberg, line 59: Not sure what should be here

src/openms_ml/include/OpenMSMLWrapper.h
cbielow: Well `std::string` forces an allocation (unless it can use SSO, but thats unlikely here). An SV works on any string-like type, e.g. you can pass a `char *`. For example when using argv[].
Its not performance critical for this parameter in any way, so not a concern, but it may be for other things.

CMakeLists.txt
timosachsenberg, line 89:
```suggestion
option(ENABLE_OPENMS_ML "Enable machine learning library to interface with e.g. pyTorch." OFF)
```

timosachsenberg: is the check for ENABLE_OPENMS_ML put in [src/openms_ml/CMakeLists.txt](https://github.com/OpenMS/OpenMS/pull/6541/files/66d987c2abaf02d87f712c78bd193cbffb999b7f#diff-b2c6494e225e03dc969755b7bffc33a5388a89a7066f318e7844bdf5e0be1c37)
? 
or already when add_subdirectory adds the CMakeLists ?

src/openms_ml/CMakeLists.txt
jpfeuffer: Will be the folder where TARGETNAMEConfig.h goes for the import/export TARGETNAME_DLL_API macros.
Should be the (include/)OpenMS_ML folder or how you will call it.

CMakeLists.txt
jpfeuffer: In the top-level. Then only add subdirectory if enabled.

src/openms_ml/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+# register relevant paths for the doxygen doc generation
+#openms_doc_path("${PROJECT_SOURCE_DIR}/include")
+#openms_doc_path("${PROJECT_SOURCE_DIR}/source/VISUAL/APPLICATIONS")
+
+# --------------------------------------------------------------------------
+# TODO: probably don't need to add executables / install tools
+# probably don't need a add_custom_target(TODO...) here
+
+# for linux/CPU only
+ExternalProject_Add(
```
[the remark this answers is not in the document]
jpfeuffer: if linux only, please add an if UNIX and not APPLE before you include all that stuff.

src/CMakeLists.txt
```diff
@@ -53,6 +53,13 @@ if(PYOPENMS)
   add_subdirectory(pyOpenMS)
 endif()
 
+if(ENABLE_OPENMS_ML)
```
jpfeuffer, line 56: and whatever restrictions it has. Linux, GPU, Arch?

src/openms_ml/CMakeLists.txt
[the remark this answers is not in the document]
jpfeuffer: and of course your own libraries need to come AFTER the creation of the target.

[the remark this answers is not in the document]
jpfeuffer: you probably also want FetchContent instead of ExternalProject. Former runs at configure time, the latter at build time.

src/openms_ml/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+# $Maintainer: Timo Sachsenberg $
+# $Authors: Timo Sachsenberg $
+# --------------------------------------------------------------------------
+
+################### Download libtorch
+
+# for libraries for linux/CPU only
+include(FetchContent)
+
+if (UNIX)
```
timosachsenberg, line 40: need to add others as well...

timosachsenberg:
libtorch CPU only can be downloaded for different platforms using this selector:
image

timosachsenberg:
superseeded by arslans PR

